PR #569, opened 2016-01-27
Rename DefaultLocation to TypeUseLocation

smillst:


smillst:
Pull requests typetools/checker-framework-inference#17 and typetools/sparta#3 adapt those code bases to this change and should be merged after this is merged.

smillst:
Also pull request typetools/checker-framework.demos#1 should be merged as well.  The Checker Framework test will fail if this demo branch isn't merged.

wmdietl:
I had a quick look through these changes and they looked good.
The only small detail I didn't like is the inconsistency in spacing.
You mostly write `@A({ X, Y })` whereas most existing code is `@A({X, Y})`. However, even within your changes you're inconsistent in whether you have a space or not. Also when an annotation takes a value, you sometimes introduce an extra space, like `@A(v= X)` instead of the previous `@A(v=Y)`.
In any way, adding the space sometimes adds even more inconsistency.

Issue #568, opened 2016-01-27
Inner classes in stubs files


cushon:
Is it possible to annotate qualified inner class types in stub files?

The manual [mentions](http://types.cs.washington.edu/checker-framework/current/checker-framework-manual.html#stub-limitations) that the stub parser can't handle annotations inside qualified names, and that `$`-qualified inner class names can be used as a workaround. I can't get it to work for parameter types.

Repro:

Lib.java

```java
import org.checkerframework.checker.nullness.qual.Nullable;
class Lib {
  static class Inner {}
  static void f(Lib.Inner e) {}
}
```

Test.java

```java
class Test {
  {
    Lib.f(null);
  }
}
```

stubs/lib.astub

```java
import org.checkerframework.checker.nullness.qual.Nullable;
class Lib {
  static void f(Lib.Inner e) {}
}
```

A few options I tried:

`static void f(Lib.@Nullable Inner e) {}`

Fails with:

```
  Underlying Exception: org.checkerframework.stubparser.ParseException: Encountered " "class" "class "" at line 3, column 1.
  Was expecting one of:
      <EOF>
      "import" ...
      "package" ...
      "@" ...
      "\u001a" ...
      "@" ...
      "@" ...
      "@" ...
      "@" ...
      ; Stack trace: org.checkerframework.stubparser.ASTParser.generateParseException(Unknown Source
```

`static void f(@Nullable Lib.Inner e) {}`

Fails with: `error: [argument.type.incompatible] incompatible types in argument.`

`static void f(@Nullable Lib$Inner e) {}`

Fails with: `error: [argument.type.incompatible] incompatible types in argument.`

smillst:
Thanks for reporting this.  The following seems to work:
```
static void f(@Nullable Inner e) {}
```
I'm not sure if this is a "bug" or an expected limitation.  If it's a limitation, we need to document it.  

The quote below is from [the stub file section of the manual](http://types.cs.washington.edu/checker-framework/current/checker-framework-manual.html#stub-limitations)
>The stub file reader does not handle nested classes. To work around this, it permits a top-level class to be written with a $ in its name, and applies the annotations to the appropriate nested class.

This limitation is about annotating an inner class or methods in an inner class in stubfiles.  For example:
````
@SomeAnnotation
class Lib$Inner{
}
````
We should clarify this.

cushon:
Thanks, clarifying the docs would be great.

[...]

Issue #407, opened 2015-07-03
Stub files cannot read annotations with class literals as attributes


GoogleCodeExporter:
```
I'm trying to write stubs for org.junit.runners.ParentRunner. My stub looks 
like the following:

class ParentRunner {
  void collectInitializationErrors(List<Throwable> errors) @UnknownInitialization(Runner.class);
}

but this fails with the exception:

error: StubParser: unhandled annotation attribute type: Runner.class class: 
class org.checkerframework.stubparser.ast.expr.ClassExpr
  Exception: java.lang.Throwable; Stack trace: org.checkerframework.framework.source.SourceChecker.errorAbort(SourceChecker.java:613)
  org.checkerframework.javacutil.ErrorReporter.errorAbort(ErrorReporter.java:28)
  org.checkerframework.framework.stub.StubParser.handleExpr(StubParser.java:1072)
  org.checkerframework.framework.stub.StubParser.getAnnotation(StubParser.java:945)
  org.checkerframework.framework.stub.StubParser.annotate(StubParser.java:652)
  org.checkerframework.framework.stub.StubParser.parseMethod(StubParser.java:501)
  org.checkerframework.framework.stub.StubParser.parse(StubParser.java:383)
  org.checkerframework.framework.stub.StubParser.parse(StubParser.java:319)
  org.checkerframework.framework.stub.StubParser.parse(StubParser.java:299)
  org.checkerframework.framework.stub.StubParser.parse(StubParser.java:294)
  org.checkerframework.framework.type.AnnotatedTypeFactory.buildIndexTypes(AnnotatedTypeFactory.java:2249)
  org.checkerframework.framework.type.GenericAnnotatedTypeFactory.postInit(GenericAnnotatedTypeFactory.java:148)
  org.checkerframework.checker.nullness.NullnessAnnotatedTypeFactory.<init>(NullnessAnnotatedTypeFactory.java:151)
  org.checkerframework.checker.nullness.NullnessVisitor.createTypeFactory(NullnessVisitor.java:109)
  org.checkerframework.checker.nullness.NullnessVisitor.createTypeFactory(NullnessVisitor.java:59)
  org.checkerframework.common.basetype.BaseTypeVisitor.<init>(BaseTypeVisitor.java:196)
  org.checkerframework.checker.initialization.InitializationVisitor.<init>(InitializationVisitor.java:71)
  org.checkerframework.checker.nullness.NullnessVisitor.<init>(NullnessVisitor.java:88)
[...]

Issue #297, opened 2015-07-03
Stub file annotation on fully qualified class names


GoogleCodeExporter:
```
The following stub file does not parse, but should.  Currently, the only way to 
annotate a fully qualified type is to put the annotation before the package 
name. 

    import checkers.nullness.quals.Nullable;

    package javax.net.ssl;

    class SSLContext
    {
        void init(javax.net.ssl.KeyManager @Nullable [] km,
                  javax.net.ssl.TrustManager @Nullable [] tm,
                  java.security. @Nullable SecureRandom random);
    }
```

Original issue reported on code.google.com by `Suzanne....@gmail.com` on 19 Dec 2013 at 6:19

mernst:
Fixed by 8997a34.

Issue #220, opened 2015-07-03
In stub files, annotations on receivers should be the first parameter of a method


GoogleCodeExporter:
```
Receivers in stub files must be written after the method signature.  It would 
make stub files easier to understand if they could be written as the first 
parameter of a method.

For example, rather than:
    void setImageResource(@PolySources @PolySinks int i) @PolySources @PolySinks;
We should be able to write:
    void setImageResource(@PolySources @PolySinks ImageView this, @PolySources @PolySinks int i);

Once this is implemented, the StubGenerator should be changed to output the 
receivers, too.



```

Original issue reported on code.google.com by `Suzanne....@gmail.com` on 9 May 2013 at 9:51

mernst:
Fixed by 8997a34.

PR #570, opened 2016-01-28
Stub limitations with Inner Classes

smillst:
Clarifies/adds limitations as discussed in Issue #568.

### DIFF
--- a/checker/manual/annotating-libraries.tex
+++ b/checker/manual/annotating-libraries.tex
@@ -401,7 +401,7 @@ following ways:
   The only required import statements are the ones to import type
   annotations.  Import statements for types are optional.
 
-  Enum constants in annotations need to be either fully qualified
+  Enum constants in annotations need to be either fully-qualified
   or imported.
   For example, one has to either write the enum constant \<ANY> in
   fully-qualified form:
@@ -602,9 +602,9 @@ in a stub file one has to write
 \end{Verbatim}
 
 \item
-  Annotations must be written before the package name on a fully qualified
+  Annotations must be written before the package name on a fully-qualified
   types rather than directly on the type it qualifies.  However, it is usually not
-  necessary to write the fully qualified name.
+  necessary to write the fully-qualified name.
 
   \begin{Verbatim}
           void init(@Nullable java.security.SecureRandom random);
@@ -612,7 +612,7 @@ in a stub file one has to write
 \item
    Annotations on types that are inner classes must be written before the
    package if the type is written as a
-   fully annotated type, or before the inner class name if the type is
+   fully-qualified type, or before the inner class name if the type is
    written as a simple type. For example,
 
 \begin{Verbatim}
--- a/checker/manual/annotating-libraries.tex
+++ b/checker/manual/annotating-libraries.tex
@@ -591,12 +591,12 @@ in a stub file one has to write
 
 \item
   The stub file reader
-  does not handle nested classes declarations.  To work around this, it permits a
+  does not handle nested class declarations.  To work around this, it permits a
   top-level class to be written with a \<\$> in its name, and applies the
   annotations to the appropriate nested class.
 
 \begin{Verbatim}
-    class Lib$Inner{
+    class Lib$Inner {
        // methods
     }
 \end{Verbatim}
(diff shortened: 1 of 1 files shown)


checker/manual/annotating-libraries.tex
```diff
@@ -603,6 +609,17 @@ \subsection{Limitations\label{stub-limitations}}
...
+   Annotations on types that are inner classes must written before the package on the
+   fully annotated type or before the inner class name. For example,
+
+\begin{Verbatim}
+        ProcessBuilder redirectError(@NonNull java.lang.ProcessBuilder.Redirect destination)
+\end{Verbatim}
+ or
+\begin{Verbatim}
+        ProcessBuilder redirectError(@NonNull Redirect destination)
+\end{Verbatim}
```
mernst: Is it possible to write the parameter's type as "ProcessBuilder.Redirect"?  Showing that example, or explaining that it is not permitted to be written, would be helpful.

smillst:
@mernst  I committed the fix you asked for.

mernst:
Looks good.  Merge when tests pass.

checker/manual/annotating-libraries.tex
```diff
@@ -603,6 +609,29 @@ \subsection{Limitations\label{stub-limitations}}
   \begin{Verbatim}
           void init(@Nullable java.security.SecureRandom random);
    \end{Verbatim}
+\item
+   Annotations on types that are inner classes must be written before the
+   package if the type is written as a
+   fully annotated type, or before the inner class name if the type is
```
wmdietl: Should the "fully annotated type" be "fully qualified type"?

checker/manual/annotating-libraries.tex
```diff
@@ -603,6 +609,29 @@ \subsection{Limitations\label{stub-limitations}}
...
+   written as a simple type. For example,
+
+\begin{Verbatim}
+        ProcessBuilder redirectError(@NonNull java.lang.ProcessBuilder.Redirect destination);
```
wmdietl: Can we mark the cases in the manual that violate Java 8 syntax, to highlight the discrepancies to source code?
Also, for each such mismatch, can we file an issue against the stub parser and reference the issue here.
(In Java 8 syntax, the example should be "java.lang.ProcessBuilder.@NonNull Redirect".)

checker/manual/annotating-libraries.tex
```diff
@@ -591,18 +591,47 @@ \subsection{Limitations\label{stub-limitations}}
...
+  necessary to write the fully-qualified name.
 
   \begin{Verbatim}
           void init(@Nullable java.security.SecureRandom random);
    \end{Verbatim}
+\item
+   Annotations on types that are inner classes must be written before the
+   package if the type is written as a
+   fully-qualified type, or before the inner class name if the type is
+   written as a simple type. For example,
```
mernst: Werner says (the comment got wiped out because of a subsequent commit):
Can we mark the cases in the manual that violate Java 8 syntax, to highlight the discrepancies to source code?
Also, for each such mismatch, can we file an issue against the stub parser and reference the issue here.
(In Java 8 syntax, the example should be java.lang.ProcessBuilder.@NonNull Redirect.)

smillst: >Can we mark the cases in the manual that violate Java 8 syntax, to highlight the discrepancies to source code?

All of the stub file limitations listed in this section are violations of Java 8 syntax.

>Also, for each such mismatch, can we file an issue against the stub parser and reference the issue here.

Since the right way to fix all of these limitations is to update the version of the JavaParser used by the stub parser, I don't think it makes sense to file an issue for each individual limitation.  Though we could file one that says "update JavaParser".

wmdietl: All of the stub file limitations listed in this section are violations of Java 8 syntax.

Sure, but it might be clearer if the manual states that more explicitly.

I hadn't realized that replacing all of the StubParser was your plan. If so, I agree that having one issue is enough.

smillst:
>From Werner:
>I hadn't realized that replacing all of the StubParser was your plan. If so, I agree that having one issue is enough.

I thought that was generally agreed upon plan for fixing these limitations.  It's not something on my agenda at the moment.

checker/manual/annotating-libraries.tex
mernst: This sounds like a good resolution to me.
It definitely isn't worth fixing any StubParser problems without updating to the latest version.

smillst: Turns out most of the limitations already have issues ( #568, #407, #297, #220).  I also submitted one for updating JavaParser #571.

PR #575, opened 2016-01-29
Move non-JDK annotations from jdk.astub files into their own stub files.

jthaine:
Also include the checker/lib directory in the distribution.

The original plan was to remove all apache-bcel annotations since there is a JAR file for them. However after speaking to Dan, we are not fully confident that all the annotations in all apache-bcel.astub files have been added to the jar. I think it would be confusing to users to remove some BCEL stub files and not others. So I think we should leave them all in for now. What I did remove is the temporary BCEL6 signature annotations that I had temporarily added just for Daikon to type check.


smillst:
I checked that each new stub file has the correct important statement and is in the same directory as the checker that references it.

checker/lib/README.developers
```diff
@@ -0,0 +1,11 @@
+Please read the README file in this directory first, as it is the file
+that is shipped in the distribution and is intended for users of the
+Checker Framework.
+
+When adding a .jar file to this directory, update the list in README and also
```
wmdietl: Are there instructions for how to create a new annotated JDK file?
If not, is this file the right place to add them?

jthaine: I am not aware of such instructions, but I don't think this is the right place to add them since the annotated JDK doesn't live in this directory.

changelog.txt
```diff
@@ -2,6 +2,12 @@
 
 Version 1.9.11, 1 February 2016
 
+The Checker Framework distribution now contains annotated versions
```
mernst: We missed the deadline for this release, right?  So shouldn't this text appear for the March 1 release, so that when we merge it after the release, it merges without a conflict?

checker/lib/README.developers
```diff
@@ -0,0 +1,11 @@
...
+the list in ../manual/annotating-libraries.tex .
```
mernst: The list in that file is commented out, and it's not reinstated by this pull request.

checker/lib/README.developers
```diff
@@ -0,0 +1,11 @@
+Please read the README file in this directory first, as it is the file
```
mernst: I'm not sure I understand the point of adding a new README.developers file.  If the file exists, it should be included in the distribution to minimize the difference between the distribution and the repository.  Furthermore, I would think that the comment about where the source code came from is interesting mostly to developers, but it wasn't moved to this file.  I would just keep all the information in a single README file to reduce confusion and make it easier to find and update information.

checker/lib/README.developers
```diff
@@ -0,0 +1,11 @@
...
+that is shipped in the distribution and is intended for users of the
+Checker Framework.
+
+When adding a .jar file to this directory, update the list in README and also
+the list in ../manual/annotating-libraries.tex .
+In general, old files need not be removed; that way, we accommodate users
+regardless of which version of a library they wish to use.
+
+To change which files from this directory are included in the distribution,
```
mernst: I find it confusing that there is both a checker-includes and a checker-excludes file, and that neither one is documented anywhere in the repository.  What is the relationship between them?  Why do we need both of them?  Why is checker-includes so large?

We should strive to minimize the differences between the repository and the release.  Therefore, anything that appears in either checker-includes or checker-excludes is suspect.  I propose that we attempt to make those files as small as possible, and that we add a comment (prefixed by a # sign permitting tools to strip those lines off before interpreting the rest) to every line to justify it.

checker/lib/README.developers
mernst: I don't understand Werner's question.  Could you please clarify?  Do you mean instructions for creating a new annotated (non-JDK) library, or for creating a new version of the JDK (one shouldn't do that; we should have one version of the JDK with all annotations in it), or for adding a new file to the annotated JDK (one shouldn't need to do that), or something else?  Thanks.

wmdietl: The instructions tell me where to put a new jar file, but not how to build one. 
If I have my own library, how do I create a jar file for inclusion here? 
For what type systems are these libraries annotated? If I annotated BCEL with a different type system, would it also go here?

changelog.txt
jthaine: Hi Mike. I just pushed a commit with some fixes. If you think they look good, I can create a clean branch with my changes in their current form and merge it tonight.
There isn't time tonight for me to make (and for us to agree on) the potential improvements to the README Werner asked about and the checker-includes/excludes related fix.
One thing I am unsure about is this line in the manual I uncommented:
\href{https://commons.apache.org/proper/commons-bcel/}{Apache Commons BCEL} (versions 5.2 and 6.0 pre-release)
Are these two still the versions that are checked in? Since you were involved with Dan and/or Mark in updating these, I thought you might know.

checker/manual/annotating-libraries.tex
```diff
@@ -1,30 +1,29 @@
 \htmlhr
 \chapter{Annotating libraries\label{annotating-libraries}}
 
-%% TODO: reinstate this once the checker/lib/ directory contains content.
```
jthaine: This is just uncommenting - I did not change any contents.

changelog.txt
mernst: I have updated the text and I think it's OK.  The lib/ directory does have both version 5.2 and 6.0 pre-release.

mernst:
I think this is OK to merge.

PR #477, opened 2015-09-23
Whole-program type inference using .jaif files

pbsf:
Added a behavior to the CFAbstractTransfer that enables a signature inference using .jaif files. 
To enable this feature the argument -AinferSignatures must be passed to javac.

With this feature on, method return types and field types are stored in .jaif files. The types in the .jaif files get updated at return statements and assignments. Note that this works for both private and non-private methods and fields.

After generating the .jaif files, the insert-annotations-to-source tool can be used to add annotations to .java files. I created a script that iteratively runs inference and adds annotations from .jaif files to the .java files on the program until the .jaif files from one execution are similar to the .jaif files from the previous execution.

Example of usage:
$CHECKERFRAMEWORK/checker/bin/infer-and-annotate.sh "LockChecker,NullnessChecker" \
     $JSR308/plume-lib/java/plume.jar -AprintErrorStack \
     `find $JSR308/plume-lib/java/src/plume/ -name "*.java"`

Testing done:
ant signature-inference-tests
ant javadoc
ant dist (Since AFU is now added to the classpath, it made sense to test this.)

PR #576, opened 2016-02-01
Feature: Signature inference

pbsf:
Follow-up from https://github.com/typetools/checker-framework/pull/477

Added a behavior to the CFAbstractTransfer that enables a signature inference using .jaif files. 
To enable this feature the argument -AinferSignatures must be passed to javac.

With this feature on, method return types and field types are stored in .jaif files. The types in the .jaif files get updated at return statements and assignments. Note that this works for both private and non-private methods and fields.

After generating the .jaif files, the insert-annotations-to-source tool can be used to add annotations to .java files. This can be done automatically using the infer-and-annotate.sh script. It iteratively runs inference and adds annotations from .jaif files to the .java files on the program until the .jaif files from one execution are similar to the .jaif files from the previous execution.

Example of usage:
```
$CHECKERFRAMEWORK/checker/bin/infer-and-annotate.sh "LockChecker,NullnessChecker" \
$JSR308/plume-lib/java/plume.jar -AprintErrorStack \
`find $JSR308/plume-lib/java/src/plume/ -name "*.java"`
```

Testing done:
ant signature-inference-tests
ant javadoc
ant dist (Since AFU is now added to the classpath, it made sense to test this.)


build-common.properties
```diff
@@ -46,6 +46,7 @@ dataflow.lib=${dataflow.loc}/dist/dataflow.jar
 dataflow.sources.lib=${dataflow.loc}/dist/dataflow-source.jar
 dataflow.javadoc.lib=${dataflow.loc}/dist/dataflow-javadoc.jar
 
+afu.lib=${annotation.tools}/annotation-file-utilities/annotation-file-utilities.jar
```
wmdietl: "afu.lib" should be with "annotation.tools" in the external dependency section.
For consistency, I would use a longer name.

framework/tests/src/tests/signatureinference/SignatureInferenceTestAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,162 @@
...
+            super(f);
+        }
+
+        @Override
+        public AnnotationMirror leastUpperBound(AnnotationMirror a1,
+                AnnotationMirror a2) {
+            if ((AnnotationUtils.areSameByClass(a1, Sibling1.class)
+                    && AnnotationUtils.areSameByClass(a2, Sibling2.class)) 
+                    || (AnnotationUtils.areSameByClass(a1, Sibling2.class)
+                    && AnnotationUtils.areSameByClass(a2, Sibling1.class))
```
mernst: This indentation is incorrect.  The entire expression should be to the right and below the open parenthesis.  You can fix that by moving "&&" under the "A" in "(AnnotationUtils.areSameByClass(a1, Sibling2.class)".  Same comment for other following lines.

framework/tests/src/tests/signatureinference/SignatureInferenceTestAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,162 @@
...
+            }
+
+            if (AnnotationUtils.areSameByClass(sub, Top.class)) {
+                return false;
+            }
+
+            if ((AnnotationUtils.areSameByClass(sub, Sibling1.class)
+                    || AnnotationUtils.areSameByClass(sub, Sibling2.class)
+                    || AnnotationUtils.areSameByClass(sub, SiblingWithFields.class))
+                    && AnnotationUtils.areSameByClass(sup, Parent.class)) {
```
mernst: This indentation is misleading too.  The "||" should be aligned under the "A" on the first line, but the "&&" should be aligned under the "(" that precedes the "A".

framework/tests/src/tests/signatureinference/SignatureInferenceTestAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,162 @@
...
+            if ((AnnotationUtils.areSameByClass(sub, Sibling1.class)
+                    || AnnotationUtils.areSameByClass(sub, Sibling2.class)
+                    || AnnotationUtils.areSameByClass(sub, SiblingWithFields.class)
+                    || AnnotationUtils.areSameByClass(sub, Parent.class)) 
+                    && AnnotationUtils.areSameByClass(sup, DefaultType.class)){
+                return true;
+            }
+
+            if (AnnotationUtils.areSameByClass(sub, SiblingWithFields.class)
+                    && AnnotationUtils.areSameByClass(sup, SiblingWithFields.class)) {
```
mernst: I find it clearer to align the && under the "A", that is one character further right than the "(" that encloses it.

pbsf:
Moved to: https://github.com/typetools/checker-framework/pull/612

PR #581, opened 2016-02-03
Add TYPE_PARAMETER to @Target on type qualifiers

smillst:
Type qualifiers should have ElementType.TYPE_PARAMETER in their`@Target` meta-annotation unless the qualifier should be forbidden on type parameter lower bounds and wildcard upper bounds.  

For example:
````
@TP T extends Object
````
The lower bound of `T` is `@TP null`

````
@TP ? super CharSequence
````
The upper bound of `?` is `@TP Object`

(When we reviewed and merged #533, we forgot that TYPE_PARAMETER was required.  This PR fixes that.)

### DIFF
--- a/checker/src/org/checkerframework/checker/fenum/qual/FenumBottom.java
+++ b/checker/src/org/checkerframework/checker/fenum/qual/FenumBottom.java
@@ -1,14 +1,18 @@
 package org.checkerframework.checker.fenum.qual;
 
-import java.lang.annotation.*;
-
-import com.sun.source.tree.Tree;
-
 import org.checkerframework.framework.qual.DefaultFor;
-import org.checkerframework.framework.qual.TypeUseLocation;
 import org.checkerframework.framework.qual.ImplicitFor;
 import org.checkerframework.framework.qual.SubtypeOf;
 import org.checkerframework.framework.qual.TargetLocations;
+import org.checkerframework.framework.qual.TypeUseLocation;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import com.sun.source.tree.Tree;
 
 /**
  * The bottom qualifier for fenums, its relationships are setup via the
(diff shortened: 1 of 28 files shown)

### CODE REVIEW

File: checker/src/org/checkerframework/checker/fenum/qual/FenumBottom.java [hunk 1]
```diff
@@ -1,14 +1,18 @@
...
-import com.sun.source.tree.Tree;
-
 import org.checkerframework.framework.qual.DefaultFor;
-import org.checkerframework.framework.qual.TypeUseLocation;
 import org.checkerframework.framework.qual.ImplicitFor;
 import org.checkerframework.framework.qual.SubtypeOf;
 import org.checkerframework.framework.qual.TargetLocations;
+import org.checkerframework.framework.qual.TypeUseLocation;
+
+import java.lang.annotation.Documented;
```
mernst: Minor: I would have expected to see java.* annotations at the beginning or end rather than in the middle, on the theory that import statements go from most general libraries to most specific libraries, but maybe we do something else?  It's not alphabetical, but I can't tell what the rule is.

mernst:
Modulo my confusion about the order of import statements, this looks good.  Thanks!

checker/src/org/checkerframework/checker/fenum/qual/FenumBottom.java
smillst: I followed the order of imports @wmdietl  uses in Eclipse which he exported and stored in [checker/eclipse/checker-framework.importorder](
https://github.com/typetools/checker-framework/blob/master/checker/eclipse/checker-framework.importorder)  

I don't have an opinion on the order of imports.  I just want to configure IntelliJ in such a way that imports on files under code review do not distract the reviewers.

smillst:
I'll merge once the tests pass.

checker/src/org/checkerframework/checker/fenum/qual/FenumBottom.java
mernst: Your diff (that we are commenting on) has java.* coming after org.checkerframework.*, but the file [checker/eclipse/checker-framework.importorder](
https://github.com/typetools/checker-framework/blob/master/checker/eclipse/checker-framework.importorder) has them in the opposite order.  So maybe your IntelliJ configuration isn't true to the Eclipse configuration file, or maybe I don't understand the Eclipse configuration file.

smillst: The order in the configuration file is specified via the integer at the beginning of each line rather than the order of the file.  So, org.checkerframework.* is before java.*

Issue #559, opened 2016-01-18
stub files and guava's Function


cushon:
The declaration of guava's Function interface is:

```java
package com.google.common.base;

import javax.annotation.Nullable;

public interface Function<F, T> {
  @Nullable T apply(@Nullable F input);
}
```

The `@Nullable` annotations are there for findbugs, which interprets `@Nullable` to mean "only that some subtypes *might* accept null" ([guava/issues/920](https://github.com/google/guava/issues/920#issuecomment-61330676)). They clash with the checker's semantics, where the type of `apply` should just depend on the instantiation of the type parameters and you can have any of:

* `Function<@Nullable F, @Nullable T>`
* `Function<@Nullable F, @NonNull T>`
* `Function<@NonNull F, @Nullable T>`
* `Function<@NonNull F, @NonNull T>`

I can't find a way to write stubs that override the existing annotations.

---

Repro

Passing `null` to a `Function<@NonNull String, @NonNull Integer>` should result in an error, but succeeds anyways:

Function.astub

```java
package com.google.common.base;

import org.checkerframework.checker.nullness.qual.Nullable;

public interface Function<F extends @Nullable Object, T extends @Nullable Object> {
  T apply(F from);
}
```

Test.java

```java
import com.google.common.base.Function;
import org.checkerframework.checker.nullness.qual.NonNull;

class Test {
  void f(Function<@NonNull String, @NonNull Integer> x) {
    x.apply((String) null);
  }
}
```

```
$ javac -version -processor org.checkerframework.checker.nullness.NullnessChecker \
-cp guava.jar -Astubs=Function.astub Test.java
javac 1.8.0-jsr308-1.9.10
...
succeeds, should fail
```

---

Adding explicit annotations to `apply` in the stub works as expected, but the formal parameter needs to be un-annotated for it to depend on the instantiation of the type parameter.

Is there a way to write a stub that causes the analysis to ignore the existing annotation? Or is there some other approach you'd recommend?

wmdietl:
Interesting problem. Removing the incorrect annotations would be a very simple solution, but I assume you don't want to do that :-)

There are four uses of field org.checkerframework.framework.stub.StubParser.atypeFactory in the StubParser that read the annotated type from the bytecode and then add the annotations from the stubfile.
In this situation, b/c the bytecode is already annotated and the stubfile contains no annotations, only the annotations from the bytecode persist.

[...]

PR #586, opened 2016-02-05
Add test and fix for Issue #559

wmdietl:
Add a test for Issue #559. Fix it by adding StubParser.handleExistingAnnotations.
Code can be improved once the StubParser understands '@AnnotatedFor`.

### DIFF
--- a/checker/tests/nullness-extra/issue559/Issue559.java
+++ b/checker/tests/nullness-extra/issue559/Issue559.java
@@ -1,0 +1,13 @@
+// Test case for Issue 559:
+// https://github.com/typetools/checker-framework/issues/559
+
+import java.util.Optional;
+
+class Issue559 {
+    void bar(Optional<String> o) {
+	// With myjdk.astub the following should fail with an
+	// argument.type.incompatible error.
+        o.orElse(null);
+	o.orElse("Hi");
+    }
+}
--- a/checker/tests/nullness-extra/issue559/Issue559.java
+++ b/checker/tests/nullness-extra/issue559/Issue559.java
@@ -5,9 +5,9 @@ import java.util.Optional;
 
 class Issue559 {
     void bar(Optional<String> o) {
-	// With myjdk.astub the following should fail with an
-	// argument.type.incompatible error.
+        // With myjdk.astub the following should fail with an
+        // argument.type.incompatible error.
         o.orElse(null);
-	o.orElse("Hi");
+        o.orElse("Hi");
     }
 }
--- a/framework/src/org/checkerframework/framework/stub/StubParser.java
+++ b/framework/src/org/checkerframework/framework/stub/StubParser.java
@@ -321,8 +321,9 @@ public class StubParser {
             parsePackage(cu.getPackage(), atypes, declAnnos);
         }
         if (cu.getTypes() != null) {
-            for (TypeDeclaration typeDecl : cu.getTypes())
+            for (TypeDeclaration typeDecl : cu.getTypes()) {
                 parse(typeDecl, packageName, packageAnnos, atypes, declAnnos);
+            }
         }
     }
 
@@ -404,8 +405,9 @@ public class StubParser {
         AnnotatedDeclaredType type = atypeFactory.fromElement(elt);
         annotate(type, decl.getAnnotations());
 
-        List<? extends AnnotatedTypeMirror> typeArguments = type.getTypeArguments();
-        List<TypeParameter> typeParameters = decl.getTypeParameters();
+        final List<? extends AnnotatedTypeMirror> typeArguments = type.getTypeArguments();
+        final List<TypeParameter> typeParameters = decl.getTypeParameters();
+
         // It can be the case that args=[] and params=null.
         // if ((typeParameters == null) != (typeArguments == null)) {
         //     throw new Error(String.format("parseType (%s, %s): inconsistent nullness for args and params%n  args = %s%n  params = %s%n", decl, elt, typeArguments, typeParameters));
@@ -446,7 +448,7 @@ public class StubParser {
              */
         }
 
-        annotateParameters(type.getTypeArguments(), decl.getTypeParameters());
+        annotateParameters(typeArguments, typeParameters);
         annotateSupertypes(decl, type);
         putNew(atypes, elt, type);
     }
@@ -456,7 +458,9 @@ public class StubParser {
             for (ClassOrInterfaceType superType : typeDecl.getExtends()) {
                 AnnotatedDeclaredType foundType = findType(superType, type.directSuperTypes());
                 assert foundType != null : "StubParser: could not find superclass " + superType + " from type " + type;
-                if (foundType != null) annotate(foundType, superType);
+                if (foundType != null) {
+                    annotate(foundType, superType);
+                }
             }
         }
         if (typeDecl.getImplements() != null) {
@@ -467,7 +471,9 @@ public class StubParser {
                 // this addition to be compatible with Java 6.
                 assert foundType != null || (superType.toString().equals("AutoCloseable") || superType.toString().equals("java.io.Closeable") || superType.toString().equals("Closeable")) :
                     "StubParser: could not find superinterface " + superType + " from type " + type;
-                if (foundType != null) annotate(foundType, superType);
+                if (foundType != null) {
+                    annotate(foundType, superType);
+                }
             }
         }
     }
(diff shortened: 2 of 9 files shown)


mernst:
The Travis build failed with:

        [junit] Testcase: tests.InterningTest:run[all-systems/java8/lambda/Lambda]:	Caused an ERROR

Can you reproduce that locally?  (I haven't tried to.)

checker/tests/nullness-extra/issue559/Issue559.java
```diff
@@ -0,0 +1,13 @@
+// Test case for Issue 559:
+// https://github.com/typetools/checker-framework/issues/559
+
+import java.util.Optional;
+
+class Issue559 {
+    void bar(Optional<String> o) {
+	// With myjdk.astub the following should fail with an
+	// argument.type.incompatible error.
+        o.orElse(null);
```
mernst: Extra space in indentation.

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -294,14 +297,17 @@ public void parse(Map<Element, AnnotatedTypeMirror> atypes, Map<String, Set<Anno
         parse(this.index, atypes, declAnnos);
     }
 
-    private void parse(IndexUnit index, Map<Element, AnnotatedTypeMirror> atypes, Map<String, Set<AnnotationMirror>> declAnnos) {
```
mernst: I personally keep the signature on one line for easier textual searching, but we can make the change if you prefer.  (It does avoid the super-long lines.)

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -376,63 +384,68 @@ private void parse(TypeDeclaration typeDecl,
...
     }
 
-    private void parseType(ClassOrInterfaceDeclaration decl, TypeElement elt, Map<Element, AnnotatedTypeMirror> atypes, Map<String, Set<AnnotationMirror>> declAnnos) {
+    private void parseType(ClassOrInterfaceDeclaration decl,
+            TypeElement elt, Map<Element, AnnotatedTypeMirror> atypes,
+            Map<String, Set<AnnotationMirror>> declAnnos) {
         annotateDecl(declAnnos, elt, decl.getAnnotations());
         AnnotatedDeclaredType type = atypeFactory.fromElement(elt);
         annotate(type, decl.getAnnotations());
-        {
```
mernst: I assume that the point of this curly brace was to emphasize the limited scope of the immediately following local variables.  (I think that's the standard reason for a code block.)  Did you extend their scope because it was necessary or just to avoid the extra indentation of the block that limited their scope?

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -505,6 +518,29 @@ private void parseMethod(MethodDeclaration decl, ExecutableElement elt,
         putNew(atypes, elt, methodType);
     }
 
+    private void handleExistingAnnotations(AnnotatedTypeMirror atype, Type typeDef) {
```
mernst: Could you please add documentation for this method, indicating what its purpose or effect is?

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -986,7 +1039,8 @@ private void handleExpr(AnnotationBuilder builder, String name,
                     builder.setValue(name, arr);
                 }
             } else {
-                ErrorReporter.errorAbort("StubParser: unhandled annotation attribute type: " + expr + " and expected: " + expected);
+                ErrorReporter.errorAbort("StubParser: unhandled annotation attribute type: " + expr +
+                        " and expected: " + expected);
```
mernst: I prefer to put operations at the beginning of a line rather than at the end of the previous one -- I find it makes the code easier to read, since one can tell immediately what operation is being performed.

wmdietl:
The complete failure is:

```
 Testcase: tests.InterningTest:run[all-systems/java8/lambda/Issue450]:	Caused an ERROR
    [junit] Forked Java VM exited abnormally. Please note the time in the report does not reflect the time until the VM exit.
    [junit] junit.framework.AssertionFailedError: Forked Java VM exited abnormally. Please note the time in the report does not reflect the time until the VM exit
```

I cannot reproduce this failure locally and it seems like the Interning Test failures you were having a few weeks ago.
The "master" branch is failing with a similar error message.

checker/tests/nullness-extra/issue559/Issue559.java
wmdietl: Will fix.

framework/src/org/checkerframework/framework/stub/StubParser.java
wmdietl: There were a few very long lines in this file and I tried to break all of them down to something more manageable.

framework/src/org/checkerframework/framework/stub/StubParser.java
wmdietl: To avoid the extra indentation. There are only three statements after the end of the code block and the benefit in limiting the scope didn't seem to outweigh the negative effect of the extra nesting.

framework/src/org/checkerframework/framework/stub/StubParser.java
wmdietl: Will do.
Can you suggest a good location in the Checker Framework manual where this changed behavior can be documented?

framework/src/org/checkerframework/framework/stub/StubParser.java
wmdietl: I don't like reaching the end of the line and not knowing what happens next.
This is the style used most commonly in javac and the Checker Framework.

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -376,63 +384,68 @@ private void parse(TypeDeclaration typeDecl,
...
-                                              decl, typeParameters.size(), typeParameters,
-                                              elt, elt.getClass(), type, type.getClass(), typeArguments.size(), typeArguments));
-                */
-            }
+                                          decl, typeParameters.size(), typeParameters,
+                                          elt, elt.getClass(), type, type.getClass(), typeArguments.size(), typeArguments));
+             */
         }
+
         annotateParameters(type.getTypeArguments(), decl.getTypeParameters());
```
mernst: Now that this line is in the scope of local variable typeArguments, should it use it instead of `type.getTypeArguments()`?  If not, then we should reinstate the scoping.

framework/src/org/checkerframework/framework/stub/StubParser.java
mernst: I meant that this method needs Javadoc.

However, you are right that the behavior should be documented in the Checker Framework manual as well.  I suggest a paragraph at the end of section "Using a stub file".

framework/src/org/checkerframework/framework/stub/StubParser.java
mernst: If the operators are all hidden at the end of the line, then I don't know what happens next.  If they are nicely lined up at the beginning, then I can tell.  I guess your preference is the opposite of mine.

Sun's style guide says to put the break after the operator.  The Google style guide, the Mozilla style guide, and Checkstyle say to put the break before the operator.  I agree with the rationale behind the latter style guides.  But there is no general consensus.

You could also leave the expressions on a single line and not break them, and then we don't have to have the discussion. :-)

framework/src/org/checkerframework/framework/stub/StubParser.java
wmdietl: Reusing variables now.

framework/src/org/checkerframework/framework/stub/StubParser.java
wmdietl: The Checker Framework currently uses the Sun style guide, as does javac on which the CF builds.
This pull request is consistent with the overall style of the rest of the code.
If you don't want to merge the PR just let me know. I don't see a point in discussing the coding style here and don't see a value in partially changing the coding style.

framework/src/org/checkerframework/framework/stub/StubParser.java
wmdietl: I added Javadoc and a short paragraph in the manual. Please have a look through the manual change.

smillst:
This is ready to be merged.  I did not merge it in case @wmdietl wish to condense this into a single commit. (Though, I don't think that is necessary.)

PR #587, opened 2016-02-09
Remove @ImplicitFor's dependency on tools.jar  

smillst:
Adds a new enum LiteralKind to use instead of Tree.Kind.


checker/manual/creating-a-checker.tex
```diff
@@ -516,23 +499,19 @@ \subsection{Declaratively specifying implicit annotations\label{declarative-type
...
-  @ImplicitFor(
-    types={TypeKind.PACKAGE},
-    typeClasses={AnnotatedPrimitiveType.class},
-    trees={
-      Tree.Kind.NEW_CLASS,
-      Tree.Kind.NEW_ARRAY,
-      Tree.Kind.PLUS,
+  @ImplicitFor(types = { TypeKind.PACKAGE, TypeKind.INT,
+                         TypeKind.BOOLEAN, TypeKind.CHAR,
+                         TypeKind.DOUBLE, TypeKind.FLOAT,TypeKind.LONG },
```
mernst: Missing space after comma.

As a more minor comment, I would put PACKAGE on its own line and then have lines just with primitives.

checker/manual/creating-a-checker.tex
```diff
@@ -516,23 +499,19 @@ \subsection{Declaratively specifying implicit annotations\label{declarative-type
...
 \begin{Verbatim}
   @SubtypeOf( { Nullable.class } )
-  @ImplicitFor(
-    types={TypeKind.PACKAGE},
-    typeClasses={AnnotatedPrimitiveType.class},
-    trees={
-      Tree.Kind.NEW_CLASS,
-      Tree.Kind.NEW_ARRAY,
-      Tree.Kind.PLUS,
+  @ImplicitFor(types = { TypeKind.PACKAGE, TypeKind.INT,
```
mernst: The primitive types TypeKind.BYTE and TypeKind.SHORT are missing from this list, and I'm not sure why.  (Is it because there aren't literals of those types, and this is only intended for literals?  A comment to that effect would be helpful, though given the "literals =" below I don't think that is the case.

checker/src/org/checkerframework/checker/interning/qual/Interned.java
```diff
@@ -26,16 +28,18 @@
  * annotated.
  * <p>
  *
- * This annotation is associated with the {@link InterningChecker}.
+ * This annotation is associated with the {@link org.checkerframework.checker.interning.InterningChecker}.
```
mernst: I don't think the fully-qualified names are necessary here (and I find them a bit distracting to read), if you retain the import statement.  But maybe this change is necessary in some version of Java?  Or maybe you made the change because your IDE incorrectly removes the import statement because it doesn't look in Javadoc comments, and you are using the fully-qualified name in the Javadoc comments to work around that problem?  I'm just wondering the rationale for the change.

checker/src/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
```diff
@@ -302,17 +304,26 @@ protected AnnotatedTypeMirror getDeclaredAndDefaultedAnnotatedType(Tree tree) {
 
     @Override
     protected TypeAnnotator createTypeAnnotator() {
+        ImplicitsTypeAnnotator implicitsTypeAnnotator = new ImplicitsTypeAnnotator(this);
+        implicitsTypeAnnotator.addTypeClass(AnnotatedTypeMirror.AnnotatedNoType.class, NONNULL);
+        implicitsTypeAnnotator.addTypeClass(AnnotatedTypeMirror.AnnotatedPrimitiveType.class, NONNULL);
         return new ListTypeAnnotator(
-                super.createTypeAnnotator(),
-                new NullnessTypeAnnotator(this)
+            new PropagationTypeAnnotator(this),
```
mernst: Changed indentation is inconsistent with code 15 lines below.

checker/src/org/checkerframework/checker/nullness/qual/NonNull.java
```diff
@@ -48,16 +45,13 @@
...
-        Tree.Kind.NEW_CLASS,
-        Tree.Kind.NEW_ARRAY,
-        Tree.Kind.PLUS, // for String concatenation
-        // All literals except NULL_LITERAL:
-        Tree.Kind.BOOLEAN_LITERAL, Tree.Kind.CHAR_LITERAL,
-        Tree.Kind.DOUBLE_LITERAL, Tree.Kind.FLOAT_LITERAL,
-        Tree.Kind.INT_LITERAL, Tree.Kind.LONG_LITERAL, Tree.Kind.STRING_LITERAL })
+@ImplicitFor(types = { TypeKind.PACKAGE, TypeKind.INT,
+                       TypeKind.BOOLEAN, TypeKind.CHAR,
+                       TypeKind.DOUBLE, TypeKind.FLOAT,TypeKind.LONG },
```
mernst: Similar question to above regarding BYTE and SHORT types.

checker/src/org/checkerframework/checker/interning/qual/Interned.java
smillst: My IDE made this change when auto-organizing the imports.  I think I have now changed the setting so that it will use simple names in JavaDoc and add the correct import.  (I'll also change this back when I address your other feedback.)

framework/src/org/checkerframework/framework/type/treeannotator/ImplicitsTreeAnnotator.java
```diff
@@ -104,6 +118,20 @@ public void addTreeClass(Class<? extends Tree> treeClass, AnnotationMirror theQu
                     treeClasses + " at " + treeClass + " with " +theQual);
         }
     }
+    public void addLiteralKind(LiteralKind literalKind, AnnotationMirror theQual) {
```
mernst: A brief Javadoc comment, for this method and the one above it, would be helpful.

checker/src/org/checkerframework/checker/interning/qual/Interned.java
mernst: OK.  If your IDE cannot handle this, it's OK to make the change, but I would do so in a different commit with a clear commit message.  Thanks!

checker/manual/creating-a-checker.tex
smillst: They should be in the list.  Good catch!

checker/src/org/checkerframework/checker/nullness/qual/NonNull.java
```diff
@@ -48,16 +45,14 @@
...
-        // All literals except NULL_LITERAL:
-        Tree.Kind.BOOLEAN_LITERAL, Tree.Kind.CHAR_LITERAL,
-        Tree.Kind.DOUBLE_LITERAL, Tree.Kind.FLOAT_LITERAL,
-        Tree.Kind.INT_LITERAL, Tree.Kind.LONG_LITERAL, Tree.Kind.STRING_LITERAL })
+@ImplicitFor(types = { TypeKind.PACKAGE,
+                       TypeKind.INT, TypeKind.BOOLEAN, TypeKind.CHAR,
+                       TypeKind.DOUBLE, TypeKind.FLOAT, TypeKind.LONG,
+                       TypeKind.SHORT, TypeKind.BYTE },
+    // All literals except NULL_LITERAL:
+    literals = { LiteralKind.BOOLEAN, LiteralKind.CHAR, LiteralKind.DOUBLE,
```
wmdietl: Why is there duplication between the "types" and the "literals" now?
Would it not be enough to keep the "types"? The only item in literals that is not in types is LiteralKind.STRING. Can that be replaced with a TypeKind.STRING instead?

checker/src/org/checkerframework/checker/oigj/OwnershipAnnotatedTypeFactory.java
```diff
@@ -47,10 +53,17 @@ public OwnershipAnnotatedTypeFactory(BaseTypeChecker checker) {
...
+        ImplicitsTreeAnnotator implicitsTreeAnnotator = new ImplicitsTreeAnnotator(this);
+        implicitsTreeAnnotator.addTreeKind(Tree.Kind.NEW_CLASS, MUTABLE);
+
+        implicitsTreeAnnotator.addTreeKind(Tree.Kind.CLASS, BOTTOM_QUAL);
+        implicitsTreeAnnotator.addTreeKind(Tree.Kind.ENUM, BOTTOM_QUAL);
+        implicitsTreeAnnotator.addTreeKind(Tree.Kind.ANNOTATION_TYPE, BOTTOM_QUAL);
+        implicitsTreeAnnotator.addTreeKind(Tree.Kind.NEW_ARRAY, BOTTOM_QUAL);
+
+
+        return new ListTreeAnnotator(implicitsTreeAnnotator, new PropagationTreeAnnotator(this),
```
wmdietl: It's not nice that there is no call of the super implementation anymore and instead you had to copy the current implementation here.
If at some point we change super.createTreeAnnotator(), this implementation will not benefit from the change.

framework/src/org/checkerframework/common/aliasing/qual/MaybeAliased.java
```diff
@@ -24,6 +24,6 @@
 @DefaultQualifierInHierarchy
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ ElementType.TYPE_PARAMETER, ElementType.TYPE_USE })
-@ImplicitFor(trees = { Tree.Kind.NULL_LITERAL }, typeNames = { java.lang.Void.class })
+@ImplicitFor(literals = { LiteralKind.NULL }, typeNames = { java.lang.Void.class })
```
wmdietl: Many times LiteralKind.NULL appears together with typeName j.l.Void, but not always.
Is that an inconsistency that we should fix? (This is unrelated to your changes, but would be a good change.)

framework/src/org/checkerframework/framework/qual/LiteralKind.java
```diff
@@ -0,0 +1,63 @@
...
+     */
+    BOOLEAN,
+    /**
+     * Corresponds to {@link com.sun.source.tree.Tree.Kind#CHAR_LITERAL} trees
+     */
+    CHAR,
+    /**
+     * Corresponds to {@link com.sun.source.tree.Tree.Kind#STRING_LITERAL} trees
+     */
+    STRING,
```
wmdietl: Should we add a PRIMITIVE constant for everything other than NULL and STRING?
TypeKind has an is isPrimitive method https://docs.oracle.com/javase/7/docs/api/javax/lang/model/type/TypeKind.html#isPrimitive()
but that cannot be called in an ImplicitFor specification. So this new constant could be used to shorten many of the ImplicitFor annotations affected by this change.

framework/src/org/checkerframework/framework/type/treeannotator/ImplicitsTreeAnnotator.java
```diff
@@ -59,6 +60,23 @@
     protected final QualifierHierarchy qualHierarchy;
 
     /**
+     * Map of {@link LiteralKind}s to {@link Tree.Kind}s.
+     * This is here and not in LiteralKinds because LiteralKind is in the checker-qual.jar
+     * which cannot depend on classes, such as Tree.Kind, that are in the tools.jar
+     */
+    private static final Map<LiteralKind,Tree.Kind> literalKindToTreeKind = new EnumMap<>(LiteralKind.class);
```
wmdietl: Space after ,

framework/src/org/checkerframework/framework/type/treeannotator/ImplicitsTreeAnnotator.java
```diff
@@ -59,6 +60,23 @@
...
+    {
```
wmdietl: literalKindToTreeKind is a static field. Should this be a static initializer?

framework/src/org/checkerframework/framework/type/treeannotator/ImplicitsTreeAnnotator.java
```diff
@@ -83,12 +101,8 @@ public ImplicitsTreeAnnotator(AnnotatedTypeFactory atypeFactory) {
                 continue;
 
             AnnotationMirror theQual = AnnotationUtils.fromClass(atypeFactory.getElementUtils(), qual);
-            for (Class<? extends Tree> treeClass : implicit.treeClasses()) {
-                addTreeClass(treeClass, theQual);
-            }
-
-            for (Kind treeKind : implicit.trees()) {
-                addTreeKind(treeKind, theQual);
+            for(LiteralKind literalKind:implicit.literals()){
```
wmdietl: Space after "for", before "{", and around ":"

framework/src/org/checkerframework/framework/type/treeannotator/ImplicitsTreeAnnotator.java
```diff
@@ -83,12 +101,8 @@ public ImplicitsTreeAnnotator(AnnotatedTypeFactory atypeFactory) {
...
+                addLiteralKind(literalKind, theQual);
```
wmdietl: Should there be a check whether ALL is used? Does it make sense to have a specific kind and ALL?

framework/src/org/checkerframework/framework/type/treeannotator/ImplicitsTreeAnnotator.java
```diff
@@ -105,6 +124,31 @@ public void addTreeClass(Class<? extends Tree> treeClass, AnnotationMirror theQu
...
+        if (literalKind == LiteralKind.ALL) {
+            for (LiteralKind iterLiteralKind : LiteralKind.valuesWithOutAll()) {
+               addLiteralKind(iterLiteralKind,theQual);
+            }
+        } else {
+            Tree.Kind treeKind = literalKindToTreeKind.get(literalKind);
+            if (treeKind != null) {
+                addTreeKind(treeKind, theQual);
+            } else {
+                ErrorReporter.errorAbort("LiteralKind " + literalKind + "is not mapped to a Tree.Kind.");
```
wmdietl: Space before "is".

checker/manual/creating-a-checker.tex
```diff
@@ -483,30 +489,7 @@ \subsection{Declaratively specifying implicit annotations\label{declarative-type
...
-  \item \code{typeClasses}: an array of class literals for classes
-    implementing \code{javax.lang.model.type.TypeMirror}, e.g.,
-    \code{javax.lang.model.type.PrimitiveType}.  Often you should use
-    a subclass of \refclass{framework/type}{AnnotatedTypeMirror}.
-  \item \code{stringPatterns}: an array of regular expressions that will
-    be matched against
-    string literals, e.g., \code{"[01]+"} for a binary number.  Useful
-    for annotations that indicate the format of a string.
-  \end{itemize}
+specified via any combination of four fields in \refclass{framework/qual}{ImplicitFor}.
```
wmdietl: Why did you remove all explanations instead of just removing the ones that are being removed?

framework/src/org/checkerframework/framework/type/treeannotator/ImplicitsTreeAnnotator.java
```diff
@@ -105,6 +124,31 @@ public void addTreeClass(Class<? extends Tree> treeClass, AnnotationMirror theQu
...
 
+    /**
+     * Added an implicit rule for a particular {@link LiteralKind}
+     * @param literalKind {@code LiteralKind} that should be implicited to {@code theQual}
+     * @param theQual the {@code AnnotationMirror} that should be applied to the {@code literalKind}
+     */
+    public void addLiteralKind(LiteralKind literalKind, AnnotationMirror theQual) {
+        if (literalKind == LiteralKind.ALL) {
+            for (LiteralKind iterLiteralKind : LiteralKind.valuesWithOutAll()) {
+               addLiteralKind(iterLiteralKind,theQual);
```
wmdietl: Spacing.

checker/src/org/checkerframework/checker/nullness/qual/NonNull.java
smillst: >Why is there duplication between the "types" and the "literals" now?

This was the previous behavior via `typeClasses = AnnotatedPrimitiveType.class`.  I think you are right that the primitive LiteralKinds are not necessary.  I'll make that change.

>The only item in literals that is not in types is LiteralKind.STRING. Can that be replaced with a TypeKind.STRING instead?

No, because local variables that are Strings should be defaulted to `@Nullable`

framework/src/org/checkerframework/framework/qual/LiteralKind.java
smillst: Good idea.  I'll make that change

checker/manual/creating-a-checker.tex
smillst: I thought this was just a copy of the declaration of `ImplicitFor` and I didn't want to maintain it in two places.  But there are actually more details here than in `ImplicitFor`'s Javadoc.  I'll move the details here into the Javadoc.

checker/src/org/checkerframework/checker/oigj/OwnershipAnnotatedTypeFactory.java
smillst: This is how other checkers add "implicits" procedurally, see UnitsAnnotatedTypeFactory for example.  The alternative is to call super.createTreeAnnotator, cast it to a ListTreeAnnotator, find the ImplicitTreeAnnotator -- that's not nice either.  Perhaps, we need another way to do this, but implemented that is orthogonal to this pull request.

framework/src/org/checkerframework/framework/qual/LiteralKind.java
smillst: I added this, but no checker currently uses it.

framework/src/org/checkerframework/common/aliasing/qual/MaybeAliased.java
smillst: I'm not sure when the j.i.Void type is used.  If it is for implicit lower bounds, that would effect how they would be defaulted.

smillst:
Need to add entry to change log.

changelog.txt
```diff
@@ -17,6 +17,8 @@ added under checker/examples/GradleExamples.
 
 Closed issues: 559.
 
+checker-qual.jar no longer uses classes, such as com.sun.source.tree.Tree, in tools.jar
```
wmdietl: I would formulate this more in what the effect on the user is.

checker/examples/subtyping-extension/qual/Encrypted.java
```diff
@@ -1,15 +1,16 @@
 package qual;
 
 import org.checkerframework.framework.qual.*;
+import org.checkerframework.framework.qual.LiteralKind;
```
wmdietl: There is both an import for `qual.*` and `qual.LiteralKind`. Can we make all imports explicit?

framework/src/org/checkerframework/framework/qual/LiteralKind.java
wmdietl: Why not? In several of your changes you list all the LiteralKinds for primitives. It would be nicer to use PRIMITIVE instead.

smillst: Which ones are you talking about? Several list all primitive TypeKinds, but none list all LiteralKinds.

wmdietl: I meant the ones that list all primitive TypeKinds should use PRIMITIVE. Your "I add this" was about the new PRIMITIVE and I was referring to that.

PR #588, opened 2016-02-09
Updated Gradle instructions to use Maven Central

smillst:
and added Gradle examples to checker/examples


checker/examples/GradleExamples/GradleJava8Example/src/main/java/org/checkerframework/example/GradleExample.java
```diff
@@ -0,0 +1,30 @@
...
+    public static @Nullable Object nullable = null;
+
+    public static void main(final String[] args) {
+        System.out.println("Hello World!");
+
+        StrBuilder stb = new StrBuilder();
+
+        @NonNull Object nn = nullable;  //error on this line
+        System.out.println(nn);
+    }
```
mernst: I expected the example files to be identical for the Java 7 and Java 8 Gradle examples.  Is there a reason that you made them different?

checker/examples/GradleExamples/GradleJava7Example/build.gradle
```diff
@@ -0,0 +1,60 @@
...
+
+repositories {
+    mavenCentral()
+}
+
+if (JavaVersion.current().isJava8Compatible()) {
+    throw new GradleException("This example requires Java 7, but you are using Java 8. Please see the Java 8 example in ../GradleJava8Example")
+}
+
+assert JavaVersion.current().isJava7()
```
mernst: Rather than asserting use of Java 7, could the Gradle instructions check whether Java 7 is being used, and adjust compile.options.compilerArgs and other build parameters in that case?  This would reduce duplication in the code, the examples, and the instructions.  It would also enable users to upgrade from Java 7 to Java 8 without having to change their Maven build files.

checker/examples/GradleExamples/GradleJava7Example/build.gradle
```diff
@@ -0,0 +1,60 @@
...
+
+    compile 'org.apache.commons:commons-lang3:3.1'
+
+    testCompile group: 'junit', name: 'junit', version: '4.11'
+}
+
+allprojects {
+    tasks.withType(JavaCompile).all { JavaCompile compile ->
+
+        compile.options.debug = true
```
mernst: Why is the debug flag set to true?  Is that something we want to recommend for all users?

checker/manual/external-tools.tex
```diff
@@ -444,37 +444,177 @@ \section{Maven\label{maven}}
...
+        ]
     options.fork = true
-    options.forkOptions.executable = "$CHECKERFRAMEWORK/checker/bin/javac"
+    options.forkOptions.jvmArgs += ["-Xbootclasspath/p:${configurations.checkerFrameworkJavac.asPath}"]
   }
 }
 \end{Verbatim}
 
-% extra $ to unconfuse Emacs's LaTeX mode
+Alternatively, create a new task to run the desired checkers:
```
mernst: Why is this text only in the Java 7 instructions?  Can it be added to the Java 8 instructions too?  (The merging of the Java 7 and Java 8 instructions that I suggested should take care of this.)

checker/examples/GradleExamples/GradleJava7Example/build.gradle
smillst: When I wrote the separate examples I was assuming two use cases (which I should make explicit):
1. Using JDK 7 on Java 7 source files to produce Java 7 class files and using annotations in comments.
2. Using JDK 8 on Java 8 source files to produce Java 8  class and **not** using annotations in comments.

The Gradle build files could be merged as you suggested, but the source code cannot.   

Would you prefer the Gradle examples demonstrate different use cases?  (Which ever the use cases demonstrated, I'll make it clear in the manual instructions and example code.)

mernst: Thanks for the explanation.  I didn't realize that the key distinction was whether annotations are in comments.

I suggest that the Gradle build files be merged, the example source code not be merged (but be identical except for the comment markers around annotations), and the instructions and possibly the directory names clarify the annotations-in-comments distinction.

checker/manual/external-tools.tex
```diff
@@ -444,37 +444,106 @@ \section{Maven\label{maven}}
...
+}
+\end{Verbatim}
+
+\item Using the Maven coordinates, declare the Checker Framework dependencies:
+
+\begin{Verbatim}
+dependencies {
+    ... extisting dependencies...
+    ext.checkerFrameworkVersion = '1.9.11'
+    ext.jdkVersion = JavaVersion.current().isJava7() ? 'jdk7' : 'jdk8'
```
mernst: Should this use `ext.targetJavaVersion` instead of `JavaVersion.current()`?  Same question in the example code.

checker/examples/GradleExamples/GradleJava7Example/build.gradle
```diff
@@ -0,0 +1,65 @@
...
+       description = 'a copy of JDK classes with Checker Framework type qualifers inserted'
+    }
+    checkerFramework {
+       description = 'The Checker Framework: custom pluggable types for Java'
+    } 
+}
+
+dependencies {
+    ext.checkerFrameworkVersion = '1.9.11'
+    ext.jdkVersion = JavaVersion.current().isJava7() ? 'jdk7' : 'jdk8'
```
mernst: Should this use `ext.targetJavaVersion` instead of `JavaVersion.current()`?  Same question in the example code.

mernst:
Modulo my one comment, this is ready to merge.

checker/examples/GradleExamples/GradleJava7Example/build.gradle
smillst: No, the annotated JDK version must match the `JavaVersion.current()` even if the user hard codes `targetJavaVersion` to something besides the current version.

PR #589, opened 2016-02-09
Update MavenExample

smillst:
Comment parts of the Maven pom that are not required when using Java 8
This makes the example consistent with the instructions in the manual.

### DIFF
--- a/checker/examples/MavenExample/pom.xml
+++ b/checker/examples/MavenExample/pom.xml
@@ -15,7 +15,10 @@
     <!-- These properties will be set by the Maven Dependency plugin -->
     <!-- Change to jdk7 if using Java 7 -->
     <annotatedJdk>${org.checkerframework:jdk8:jar}</annotatedJdk>
-    <typeAnnotationsJavac>${org.checkerframework:compiler:jar}</typeAnnotationsJavac>
+    <checkerFrameworkVersion><!-- checker-framework-version -->1.9.11<!-- /checker-framework-version --></checkerFrameworkVersion>
+      <!-- The type annotations compiler is required if using Java 7. -->
+      <!-- Uncomment the following line if using Java 7. -->
+      <!--<typeAnnotationsJavac>${org.checkerframework:compiler:jar}</typeAnnotationsJavac>-->
   </properties>
 
   <dependencies>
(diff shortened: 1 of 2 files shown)


checker/examples/MavenExample/pom.xml
```diff
@@ -15,7 +15,10 @@
     <!-- These properties will be set by the Maven Dependency plugin -->
     <!-- Change to jdk7 if using Java 7 -->
     <annotatedJdk>${org.checkerframework:jdk8:jar}</annotatedJdk>
-    <typeAnnotationsJavac>${org.checkerframework:compiler:jar}</typeAnnotationsJavac>
+    <checkerFrameworkVersion><!-- checker-framework-version -->1.9.11<!-- /checker-framework-version --></checkerFrameworkVersion>
+      <!-- The type annotations compiler is required if using Java 7. -->
```
mernst: I'm not sure why these 3 lines are indented.

checker/examples/MavenExample/pom.xml
```diff
@@ -15,7 +15,10 @@
     <!-- These properties will be set by the Maven Dependency plugin -->
     <!-- Change to jdk7 if using Java 7 -->
```
mernst: Here and below, there are instructions regarding what to do if using Java 7.  I suggest that instead we introduce a variable about JDK version.  That way, users can set that variable and change nothing else.  This will be less error-prone.

checker/examples/MavenExample/pom.xml
```diff
@@ -92,7 +95,10 @@
               <compilerArgs>
                   <!-- location of the annotated JDK, which comes from a Maven dependency -->
                   <arg>-Xbootclasspath/p:${annotatedJdk}</arg>
-                  <arg>-J-Xbootclasspath/p:${typeAnnotationsJavac}</arg>
+                  <!--treat Checker Framework errors as warnings-->
```
mernst: I see this is consistent with the previous behavior.  What is the rationale for this behavior?  I agree a user could set it if desired, but I don't see that we should be recommending that all Maven users always use -Awarns.  Therefore, I would omit this.

smillst: I think the previous behavior was to not stop the build because of Checker Framework errors.  It also did not stop the build if there were any javac errors, too; so I made this change.  

I just pushed a change that comments out the line below, and I'm also checking to make sure the release sanity checks don't depend on this behavior.

checker/examples/MavenExample/pom.xml
smillst: I agree.  Based on my limited Maven knowledge, I think we would want to use separate profiles that would be activated based on the version of Java used.  Because I don't know Maven well, this would be several hours of work.  We would also need to duplicate the example source file and put the annotations in comments.  I propose not doing this right now.

PR #591, opened 2016-02-10
Add java.util.concurrent.ExecutorService to the annotated JDK

atomicknight:
The base file was copied from the annotated-jdk8u-jdk Mercurial repository (analogous to `java.util.concurrent.BlockingQueue`). Implicit upper bounds were made explicit for clarity.

The mailing list discussion: https://groups.google.com/forum/#!topic/checker-framework-discuss/QK8USJrqkjA


mernst:
I have applied this pull request.  I didn't make upper bounds explicit, because I wanted to keep the differences from the upstream JDK as small as possible.

Thanks again for the pull request.

PR #593, opened 2016-02-11
@StubFiles should not be inherited

smillst:
If `@StubFiles` is inherited and a checker with this annotation is subclassed, then the `AnnotatedTypeFactory` looks for those stub files in the directory of that subclass (and the user directory) and issues an error if they do not exist.   This is currently causing the SPARTA Jenkins job to fail:

http://tern.cs.washington.edu:8080/job/sparta/4542/console

### DIFF
--- a/framework/src/org/checkerframework/framework/qual/StubFiles.java
+++ b/framework/src/org/checkerframework/framework/qual/StubFiles.java
@@ -7,9 +7,13 @@ import java.lang.annotation.*;
  * stub files that should be used in addition to jdk.astub.
  * This allows larger compound checkers to separate the annotations
  * into multiple files.
+ *
+ * This annotation is not inherited, that means if a checker with
+ * this annotation is subclassed, then this annotation must be copied
+ * to the subclass and the stub file must also be copied to the directory
+ * that contains the subclass.
  */
 @Documented
-@Inherited
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.TYPE)
 public @interface StubFiles {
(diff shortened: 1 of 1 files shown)


jthaine:
I think this makes it easier to reason about where annotations are coming from. Looks good to me.

wmdietl:
I agree that the annotation should not be inherited.
However, the logic to load stub files should now go up the class hierarchy and for each StubFile annotation on a Checker, load the corresponding file.
Otherwise, when looking at the logic in the superclass, one assumes that a particular stubfile is loaded. However, because a subclass exists, the stubfile suddenly doesn't get loaded.

jthaine:
Werner, I am concerned about the both the complexity and brittleness of having a search path for stub files. Right now we have to consider the -Astubs option and the stub files in the subchecker's directory. If we look at the super class stub files, this will create yet another place we look for a stub file. This will make it more difficult to reason about the annotations that are retrieved. I am worried that this will lead us down a path in the long term that will cause us to add more search paths for more scenarios, resulting in a situation similar to the classpath/bootclasspath complexity. I propose being strict about making the search path as simple as possible unless we have a critical scenario that makes it necessary to expand it.

wmdietl:
If you're subclassing a checker, you would expect to extend the behavior of the superclass.
The code in the superclass expects that its stub files are used.
I don't see what is brittle about this: every checker loads the stub files in its directory.
No duplication in subclasses, which to me is much less desirable.

smillst:
I've looked at adding this behavior in the past, and it's not straight forward.  So, I suggest submitting an issue and we can implement this feature when we have a need for it.  (Can we can discuss how best to implement it then.)

wmdietl:
Can you please add an issue? And instead of encouraging that people duplicate stub files, reference that issue?

Issue #345, opened 2015-07-03
Assignment in initializer between two uninitialized variables results in an initialized variable


GoogleCodeExporter:
```
What steps will reproduce the problem?
Run the NullnessChecker on the following code snippet (also attached):
public class Test {
    String f1;
    String f2;

    {
        f1 = f2;
        f2 = f1;
        f2.toString();   //Null pointer exception here
    }
    
    public static void main(String [] args) {
        Test a = new Test();
    }
}

What is the expected output? What do you see instead?
The first assignment should issue a warning since it would expect an 
initialized value.  EIther that, or the field f1 should not be considered 
initialized after the assignment.


What version of the product are you using? On what operating system?
Checker Framework 1.8.4
```

Original issue reported on code.google.com by `Jonathan...@gmail.com` on 8 Aug 2014 at 7:32

Attachments:
* [Test.java](https://storage.googleapis.com/google-code-attachments/checker-framework/issue-345/comment-0/Test.java)

smillst:
Test cases are checker/tests/nullness/AssignmentDuringInitialization.java and checker/tests/nullness/Issue345.java

smillst:
The errors are correctly issued when caching in the AnnotatedTypeFactory is turned off:
````
javacheck -processor nullness AssignmentDuringInitialization.java -AatfDoNotCache
AssignmentDuringInitialization.java:16: error: [assignment.type.incompatible] incompatible types in assignment.
        f1 = f2;
        ^
  found   : @Initialized @Nullable String
  required: @Initialized @NonNull String
AssignmentDuringInitialization.java:16: error: [assignment.type.incompatible] incompatible types in assignment.
        f1 = f2;
             ^
  found   : @Initialized @Nullable String
  required: @Initialized @NonNull String
AssignmentDuringInitialization.java:23: error: [assignment.type.incompatible] incompatible types in assignment.
        f3 = f4;
        ^
  found   : @Initialized @Nullable String
  required: @Initialized @NonNull String
AssignmentDuringInitialization.java:23: error: [assignment.type.incompatible] incompatible types in assignment.
        f3 = f4;
             ^
  found   : @Initialized @Nullable String
  required: @Initialized @NonNull String
````
(The -AatfDoNotCache is currently only available in the caching branch.)

smillst:
````
javacheck -processor nullness AssignmentDuringInitialization.java -AatfDoNotCache
[...]

Issue #601, opened 2016-02-19
Tests should be run with caching turned off (and on)


smillst:
All tests should be run with caching turned off to ensure the results are identical to the results when caching is turned on.  (Though the test will take longer without caching)

Turning off caching in the AnnotatedTypeFactory results in three test failures: 
- Value Checker: value/Repo.java
- Interning Checker: interning/CompileTimeConstants.java
- NullnessChecker: jtreg/rawtypes/RawTypeFail.java

smillst:
plume-lib: still type checks with caching turned off.  (plumb-lib with caching took 2.591 hours, without caching took 3.426 hours)

Daikon: have not tried to turn off caching, but I suspect that issue #565 is a caching bug.

smillst:
Issue #516 is a worked around the fact that raw types are not create the same way when caching is turned off.

smillst:
Raw Types:
The code below does not issue an error when caching is turned off.
````
Generic<@NonNull Object> generic = new Generic();
````
Currently, the Checker Framework attempts to infer type arguments for raw types.  (This is [document in the manual](http://types.cs.washington.edu/checker-framework/current/checker-framework-manual.html#generics-raw-types)).  During data flow, the raw type inference fails and the result is cached.  During type checking, the inference succeeds and the type the `new Generic()` expression is:
````
Generic<@NonNull Object>
```` 
Therefore the assignment succeeds.

PR #605, opened 2016-02-23
Don't cache or read caches when checking (some) fields in Initialization Checker

smillst:
The goal of this pull request is to fix Issue #345 -- not fix or improve caching generally.  Other places in the code set shouldCache or shouldReadCache to false (but not the other).  These are probably also bugs; I'm looking into them, too.

Also, as stated in the ToDo in the code, this is a hack around a larger issue with caching fields; see Issue #601.

I checked [plume-lib](https://travis-ci.org/smillst/plume-lib-typecheck/builds/111246124) and [daikon](https://travis-ci.org/smillst/daikon-typecheck-nullness/jobs/111244072) with this change an no new errors were issued.

### DIFF
--- a/checker/src/org/checkerframework/checker/initialization/InitializationVisitor.java
+++ b/checker/src/org/checkerframework/checker/initialization/InitializationVisitor.java
@@ -140,13 +140,28 @@ public class InitializationVisitor<Factory extends InitializationAnnotatedTypeFa
                 // (var2). Then, we take the child annotation from var2 and use it
                 // for var.
                 AnnotatedTypeMirror var = atypeFactory.getAnnotatedType(lhs);
-                boolean old = atypeFactory.HACK_DONT_CALL_POST_AS_MEMBER;
+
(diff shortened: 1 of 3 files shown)


wmdietl:
Is this really the right fix for this issue? In

https://github.com/typetools/checker-framework/issues/345#issuecomment-185401914

you mention:

```
javacheck -processor nullness AssignmentDuringInitialization.java -AatfDoNotCache
AssignmentDuringInitialization.java:16: error: [assignment.type.incompatible] incompatible types in assignment.
        f1 = f2;
        ^
```

However, I would expect an error about the initialization part, not the nullness part.
The RHS is f2, which hasn't been initialized. It's declared type is `NonNull`, so I wouldn't expect the `Nullable` in the error message. Instead, I would expect that the RHS isn't fully initialized.

Can you explain a bit why this is the right fix for this issue?
Thanks!

wmdietl:
Thinking some more about this, this seems right: the receiver `this` is under initialization, which makes the result of the field read `Nullable`.
Sorry about the confusion.

Issue #607, opened 2016-02-24
Index out of bounds exceptions with var args in byte code and lambda argument


Jianchu:
Will add test case as pull request once issue number given.

wmdietl:
I've discussed this issue with Jason and will help him with the pull request.

mernst:
OK, thanks to both of you!

PR #608, opened 2016-02-24
add test case for issue607

Jianchu:
Test case for issue #607

### DIFF
--- a/checker/tests/nullness-extra/Makefile
+++ b/checker/tests/nullness-extra/Makefile
@@ -1,7 +1,7 @@
-.PHONY: all Bug109 multiple-errors package-anno issue265 issue309 compat shorthand issue502 issue559 issue594
+.PHONY: all Bug109 multiple-errors package-anno issue265 issue309 compat shorthand issue502 issue559 issue594 issue607
 
 # This skips tests that are not currently passing
-all: Bug109 compat issue265 issue309 issue502 multiple-errors package-anno shorthand 
+all: Bug109 compat issue265 issue309 issue502 multiple-errors package-anno shorthand issue607
 
 Bug109:
 	$(JAVAC) -processor org.checkerframework.checker.nullness.NullnessChecker Bug109_A.java Bug109_B.java
@@ -34,4 +34,7 @@ package-anno:
 
 shorthand:
 	make -C shorthand
+
+issue607:
+	make -C issue607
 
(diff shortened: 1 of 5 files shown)


checker/tests/nullness-extra/Makefile
```diff
@@ -1,7 +1,7 @@
-.PHONY: all Bug109 multiple-errors package-anno issue265 issue309 compat shorthand issue502 issue559 issue594
+.PHONY: all Bug109 multiple-errors package-anno issue265 issue309 compat shorthand issue502 issue559 issue594 issue607
 
 # This skips tests that are not currently passing
-all: Bug109 compat issue265 issue309 issue502 multiple-errors package-anno shorthand 
+all: Bug109 compat issue265 issue309 issue502 multiple-errors package-anno shorthand issue607
```
wmdietl: Because the test is currently failing: add a new .PHONY target "skipped" that only contains issue607.
Then we can still run the test target, but the failure won't get executed with "all".

checker/tests/nullness-extra/issue607/Issue607SuperClass.java
```diff
@@ -0,0 +1,10 @@
+public class Issue607SuperClass {
+    static Issue607Interface issue;
+    public Issue607SuperClass(Issue607Interface... issue) {
+    }
+}
+
```
wmdietl: Remove trailing whitespace.

checker/tests/nullness-extra/issue607/Makefile
```diff
@@ -0,0 +1,9 @@
+.PHONY: all
+
+all: clean
+	$(JAVAC) -source 8 -target 8 Issue607SuperClass.java Issue607Interface.java
```
wmdietl: Is the `-source 8 -target 8` really necessary here? Isn't that the default for the `JAVAC` that's being used here?

checker/tests/nullness-extra/issue607/Makefile
```diff
@@ -0,0 +1,9 @@
...
+	$(JAVAC) -processor org.checkerframework.checker.nullness.NullnessChecker -source 8 -target 8 Issue607.java > Out.txt 2>&1
```
wmdietl: Having the `Out.txt` and `Expected.txt` is OK, but it's actually enough to just run the compiler. Removing these files would simplify the test a bit.

wmdietl:
Thanks! I've merged the PR.

PR #612, opened 2016-02-29
Feature: Whole-program inference

pbsf:
Follow-up from #576 

Added a behavior to the CFAbstractTransfer that enables a signature inference using .jaif files. 
To enable this feature the argument -AinferSignatures must be passed to javac.

With this feature on, method return types and field types are stored in .jaif files. The types in the .jaif files get updated at return statements and assignments. Note that this works for both private and non-private methods and fields.

After generating the .jaif files, the insert-annotations-to-source tool can be used to add annotations to .java files. This can be done automatically using the infer-and-annotate.sh script. It iteratively runs inference and adds annotations from .jaif files to the .java files on the program until the .jaif files from one execution are similar to the .jaif files from the previous execution.

Example of usage:

```
$CHECKERFRAMEWORK/checker/bin/infer-and-annotate.sh "LockChecker,NullnessChecker" \
$JSR308/plume-lib/java/plume.jar -AprintErrorStack \
`find $JSR308/plume-lib/java/src/plume/ -name "*.java"`
Testing done:
ant signature-inference-tests
ant javadoc
ant dist (Since AFU is now added to the classpath, it made sense to test this.)
```


Jianchu:
We tried running checkerframework/checker/bin/javac with -AinferSignatures, and got the following errors:
error: SourceChecker.typeProcess: unexpected Throwable (NoClassDefFoundError) while processing Test.java; message: annotations/el/DefException; invoke the compiler with -AprintErrorStack to see the stack trace.
1 error

updates to checker/bin/javac checker/bin-devel/javac and CheckerMain seems to be missing.

wmdietl:
It would be helpful if the script gave a useful error message about usage:

checker/bin/infer-and-annotate.sh 
Aborting infer-and-annotate.sh: Expected at least 3 arguments. Check the script's documentation.
Received the following arguments: .

wmdietl:
There should be a checker/bin-devel/infer-and-annotate.sh version of the script that uses the development version of all repositories. As much code as possible should be shared between the scripts.

checker/bin/infer-and-annotate.sh
```diff
@@ -0,0 +1,160 @@
...
+
+# Path that will contain .class files after running CF's javac. This dir will
+# be deleted after executing this script.
+TEMP_DIR=build/temp-signature-inference-output
+
+# Path to directory that will contain .jaif files from the previous iteration.
+PREV_ITERATION_DIR=build/prev-signature-inference
+
+# Path to annotation-file-utilities.jar
+AFU_JAR="${CHECKERFRAMEWORK}/../annotation-tools/annotation-file-utilities/annotation-file-utilities.jar"
```
wmdietl: This should be integrated into checker/bin/javac and checker/bin-devel/javac instead of trying to add this here.

Jianchu:
put this into Test.java, 
```
import org.checkerframework.checker.regex.qual.*;
public class Test {
    String y;
	public void foo() {
		y = "(y)";
	}
}
```

Then run 
`$CHECKERFRAMEWORK/checker/bin/infer-and-annotate.sh  "RegexChecker" -AprintErrorStack /home/jianchu/Desktop/Test.java`

we expected y to be inferred `@Regex(1)`, but the script didn't produce any output. Are we doing something wrong?

checker/bin/infer-and-annotate.sh
wmdietl: Instead or in addition to changes to the javac scripts, changes to CheckerMain/CheckerDevelMain might be necessary to add the AFU dependency. That dependency is currently only in the build and Eclipse files, but not in these scripts.

checker/manual/introduction.tex
```diff
@@ -662,6 +662,20 @@ \subsection{Summary of command-line options\label{checker-options}}
...
 \end{itemize}
 
+Inference
+\begin{itemize}
+\item \<-AinferSignatures>
+  Output suggested annotations for method signatures and fields.
+  These annotations may reduce the number of type-checking
+  errors when running type-checking in the future; see
+  Section~\ref{signature-inference}.
+\item \<-ArelaxedSignatureInference>
```
wmdietl: It would be more symmetric if these options were `-AinferSignatures` and `-AinferRelaxedSignatures` or, alternatively, `-AsignatureInference` and `-ArelaxedSignatureInference`. I like the first option better.

framework/src/org/checkerframework/common/signatureinference/SignatureInferenceScenes.java
```diff
@@ -0,0 +1,901 @@
...
+//  files. This is important otherwise developers might achieve different
+//  results (order of annotations) when running the signature inference for
+//  the same set of files.
+public class SignatureInferenceScenes {
+
+    /**
+     * Directory where .jaif files will be written to and read from.
+     * This directory is relative to where the CF's javac command is executed.
+     */
+    public final static String JAIF_FILES_PATH = "build" + File.separator +
```
wmdietl: Does this String have to be in sync with the values in `infer-and-annotate.sh`? Can we break that dependency?

framework/src/org/checkerframework/common/signatureinference/SignatureInferenceScenes.java
```diff
@@ -0,0 +1,901 @@
...
+            "signature-inference" + File.separator;
+
+    /** Maps .jaif file paths (Strings) to Scenes. */
```
wmdietl: The .jaif file paths are absolute? Or relative to `JAIF_FILES_PATH`?

framework/src/org/checkerframework/common/signatureinference/SignatureInferenceScenes.java
```diff
@@ -0,0 +1,901 @@
...
+    private static Map<String, AScene> scenes = new HashMap<>();
```
wmdietl: This class has static state and only static methods; this obscures who is responsible for the memory and doesn't allow multiple instances to work at the same time.
I would prefer a design where there are either:
- no static fields, only static methods, instantiation is forbidden (e.g. like ElementUtils or TreeUtils)
- instance fields, instance methods, and instantiation and cleanup by the Checker, ATF, or Visitor.

framework/src/org/checkerframework/common/signatureinference/SignatureInferenceScenes.java
```diff
@@ -0,0 +1,901 @@
...
+
+    /** Maps a TypeUseLocation to the name of all annotations that should not be
```
wmdietl: "to the name of all annotations" is unclear, as the value is a set. Is it fully qualified names?

framework/src/org/checkerframework/common/signatureinference/SignatureInferenceScenes.java
```diff
@@ -0,0 +1,901 @@
...
+                try {
+                    IndexFileParser.parseFile(jaifPath, scene);
+                } catch (IOException e) {
+                    ErrorReporter.errorAbort("Problem while reading file in: " + jaifPath + "."
+                            + " Exception message: " + e.getMessage());
+                }
+            }
+            scenes.put(jaifPath, scene);
+        }
+        return scenes.get(jaifPath);
```
wmdietl: The `scenes.get is redundant if you just added it. I would move the `AScene scene` before the if and initialize it in both branches, then return it.

framework/src/org/checkerframework/common/signatureinference/SignatureInferenceScenes.java
```diff
@@ -0,0 +1,901 @@
...
+
+    /**
+     * Removes all annotations that should be ignored from an ATypeElement.
+     * (See {@link #shouldIgnore}).
+     */
+    private static void removeIgnoredAnnosFromATypeElement(ATypeElement typeEl,
+            TypeUseLocation loc) {
+        Set<Annotation> annosToRemove = new HashSet<>();
+        Set<String> annosToIgnoreForLocation = annosToIgnore.get(loc);
+        if (annosToIgnoreForLocation == null) return; // No annotations to ignore for that position.
```
wmdietl: I find control like this hard to miss. I would make four lines out of this one line.

framework/src/org/checkerframework/common/signatureinference/SignatureInferenceScenes.java
```diff
@@ -0,0 +1,901 @@
...
+     * invoked.
+     * @param methodElt the element of the method being invoked.
+     * @param atf the annotated type factory of a given type system, whose
+     * type hierarchy will be used to update the method parameters' types.
+     */
+    public static void updateInferredMethodParametersTypes(
+            MethodInvocationNode methodInvNode, Tree receiverTree,
+            ExecutableElement methodElt, AnnotatedTypeFactory atf) {
+        if (receiverTree == null) {
+            // TODO: Method called from static context.
```
wmdietl: Is this just for static context? Isn't it also for invocations with implicit `this`?

framework/src/org/checkerframework/common/signatureinference/SignatureInferenceScenes.java
```diff
@@ -0,0 +1,901 @@
...
+            }
+        }
+
+        // Checks if am is an implicit annotation
+        // TODO: Handles cases of implicit annotations added via an
+        // org.checkerframework.framework.type.treeannotator.ImplicitsTreeAnnotator
+        ImplicitFor implicitFor = elt.getAnnotation(ImplicitFor.class);
+        if (implicitFor != null) {
+            TypeKind[] types = implicitFor.types();
+            for (TypeKind tk : types) if (tk == atm.getKind()) return true;
```
wmdietl: I would split this and other similar code over multiple lines and add more braces.

framework/src/org/checkerframework/common/signatureinference/SignatureInferenceScenes.java
```diff
@@ -0,0 +1,901 @@
...
+import annotations.field.BasicAFT;
+import annotations.field.ScalarAFT;
+import annotations.io.IndexFileParser;
+import annotations.io.IndexFileWriter;
+import annotations.util.JVMNames;
+
+import com.sun.source.tree.ClassTree;
+import com.sun.source.tree.MethodTree;
+import com.sun.source.tree.Tree;
+import com.sun.tools.javac.code.Attribute.Array;
```
wmdietl: We try to avoid dependencies on javac implementation classes from com.sun.tools.javac.code.
For many of these, could you either use the `Element` equivalent or a helper method from `javacutils`?

framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
```diff
@@ -74,6 +77,7 @@
 import com.sun.source.tree.Tree;
 import com.sun.source.tree.VariableTree;
 import com.sun.source.util.TreePath;
+import com.sun.tools.javac.code.Symbol.ClassSymbol;
```
wmdietl: Can you avoid this dependency on `ClassSymbol`?

framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
```diff
@@ -110,10 +114,38 @@
...
+     */
+    private final boolean inferSignatures;
+
     public CFAbstractTransfer(CFAbstractAnalysis<V, S, T> analysis) {
         this.analysis = analysis;
         this.sequentialSemantics = !analysis.checker.hasOption("concurrentSemantics");
+        this.inferSignatures = analysis.checker.hasOption("inferSignatures");
+        if (inferSignatures) {
+            checkInvalidOptionsInferSignature(
+                    new String[]{"useDefaultsForUncheckedCode"});
```
wmdietl: Making this an argument obscures the operation of the method. I would put that array somewhere cheap and refer directly to it in the method. Nobody calls this method with different arguments.

framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
```diff
@@ -718,12 +750,41 @@ protected void addInformationFromPreconditions(S info,
 
         S info = in.getRegularStore();
         V rhsValue = in.getValueOfSubNode(rhs);
+        Receiver expr = FlowExpressions.internalReprOf(
+                analysis.getTypeFactory(), lhs);
+        if (inferSignatures && expr instanceof FieldAccess &&
+                !analysis.checker.shouldSuppressWarnings(n.getTree(), null) &&
```
wmdietl: I would extract the `shouldSuppressWarnings` checks into a helper method. It appears in three duplicates and one partial duplicate.

framework/src/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -191,6 +191,8 @@
     // to be output.
     "resolveReflection",
 
+    // Whether to use .jaif files signature inference
+    "inferSignatures",
```
wmdietl: The manual also mentions `inferRelaxedSignature`.

framework/src/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -1257,7 +1259,7 @@ private boolean checkSuppressWarnings(String /*@Nullable*/ [] userSwKeys, String
      *         it is contained by a declaration with an appropriately-valued
      *         {@literal @}SuppressWarnings annotation; false otherwise
      */
-    private boolean shouldSuppressWarnings(Tree tree, String errKey) {
+    public boolean shouldSuppressWarnings(Tree tree, String errKey) {
```
wmdietl: You shouldn't make this method private. In your usecase you always pass null as `errKey`. Document a dedicated method for this. Maybe it can be a lot simpler without errKey.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -910,6 +911,13 @@ public AnnotatedTypeMirror getAnnotatedType(Tree tree) {
     protected void postProcessClassTree(ClassTree tree) {
         TypesIntoElements.store(processingEnv, this, tree);
         DeclarationsIntoElements.store(processingEnv, this, tree);
+        if (checker.getOptions().containsKey("inferSignatures")) {
```
wmdietl: As mentioned above, I would instantiate SignatureInferenceScenes in the ATF, write results here, and reclaim any unused memory.

framework/src/org/checkerframework/common/signatureinference/SignatureInferenceScenes.java
```diff
@@ -0,0 +1,901 @@
+package org.checkerframework.common.signatureinference;
```
wmdietl: This class is 900 lines long. Can you split it up into functional units?

framework/tests/signature-inference/non-annotated/InheritanceTest.java
```diff
@@ -0,0 +1,28 @@
+//@skip-test
+// TODO: Enable this test when signature inference is working correctly for cases
```
wmdietl: Do you plan to file an issue for this? Or will this be fixed before we merge this PR?

framework/tests/src/tests/signatureinference/SignatureInferenceTestAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,176 @@
...
+ * AnnotatedTypeFactory to test signature inference using .jaif
+ * files.
+ * <p>
+ * The used qualifier hierarchy is straightforward and only intended for test
+ * purposes.
+ * 
+ * @author pbsf
+ */
+public class SignatureInferenceTestAnnotatedTypeFactory
+        extends
```
wmdietl: Strange formatting.

framework/tests/src/tests/signatureinference/SignatureInferenceTestAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,176 @@
...
+import org.checkerframework.javacutil.AnnotationUtils;
+
+import tests.signatureinference.qual.DefaultType;
+import tests.signatureinference.qual.ImplicitAnno;
+import tests.signatureinference.qual.Parent;
+import tests.signatureinference.qual.Sibling1;
+import tests.signatureinference.qual.Sibling2;
+import tests.signatureinference.qual.SiblingWithFields;
+import tests.signatureinference.qual.SignatureInferenceBottom;
+import tests.signatureinference.qual.Top;
```
wmdietl: Empty line after last import.

framework/tests/src/tests/signatureinference/SignatureInferenceTestAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,176 @@
...
+        @Override
+        public AnnotationMirror getBottomAnnotation(AnnotationMirror start) {
+            return BOTTOM;
+        }
+
+        @Override
+        public AnnotationMirror leastUpperBound(AnnotationMirror a1,
+                AnnotationMirror a2) {
+            if ((AnnotationUtils.areSameByClass(a1, Sibling1.class)
+                 && AnnotationUtils.areSameByClass(a2, Sibling2.class))
```
wmdietl: Why is the indentation only different by a single character? What do the operators align with?

framework/tests/src/tests/signatureinference/SignatureInferenceTestAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,176 @@
...
+            if (AnnotationUtils.areSameByClass(sub, UnknownClass.class)
+                || AnnotationUtils.areSameByClass(sup, SignatureInferenceBottom.class)) {
+                return false;
+            }
+
+            if (AnnotationUtils.areSameByClass(sub, Top.class)) {
+                return false;
+            }
+
+            if ((AnnotationUtils.areSameByClass(sub, ImplicitAnno.class))
```
wmdietl: Why is that first clause in parentheses?

wmdietl:
Great job Paulo! I've added a bunch of comments about the code and design of the feature.
Unfortunately, we didn't manage to use it with the Regex Checker. Please let us know if we were doing something wrong.
Thanks!

pbsf:
@Jianchu, @wmdietl, thanks a lot for the feedback!

The Regex example didn't work because the `AnnotationMirror` of the variable on the RHS of the assignment is `@Key(..., desc="@Regex(1)")`, instead of `@Regex(1)`, and the `@Key` annotation is being ignored on this whole-program inference. On another words, the whole-program inference won't work on type systems that use a `CheckerAdapter`, which are the `Regex` and `Tainting` type systems. Before implementing the whole-program inference I wasn't aware of the `CheckerAdapter` implementation, so I didn't handle this scenario.

I intend to fix this problem together with the other refactorings suggested by @wmdietl.

framework/src/org/checkerframework/common/signatureinference/SignatureInferenceScenes.java
pbsf: I agree with your suggestion and I went with the second approach.

I tried to document my changes the best I could, but just to give you a quick overview of what changed in terms of implementation details:
Now we have a WholeProgramInference interface, which has all update* methods (e.g. updateInferredFieldType).  The AnnotatedTypeFactory now has a field of type WholeProgramInference. The ATF instantiates this field using the WholeProgramInferenceScenes implementation. That implementation no longer has static methods. The CFAbstractTransfer visit methods will retrieve the WholeProgramInference from the ATF (via a getter) to perform the whole-program inference.

framework/src/org/checkerframework/common/signatureinference/SignatureInferenceScenes.java
pbsf: I changed to: `"Maps a TypeUseLocation to a set of names of annotations that should not be added to .jaif files for that location."`

framework/src/org/checkerframework/common/signatureinference/SignatureInferenceScenes.java
pbsf: Relative. I improved the comment by mentioning that.

framework/src/org/checkerframework/common/signatureinference/SignatureInferenceScenes.java
pbsf: Yes, it is only for static context because of the following piece of code before calling this method in `CFAbstractTransfer`:

```
Tree receiverTree = n.getTarget().getReceiver().getTree();
if (receiverTree == null)  {
    receiverTree = analysis.getContainingClass(n.getTree());
    // receiverTree could still be null after the call above. That
    // happens when the method is called from a static context.
}
analysis.atypeFactory.getWholeProgramInference().updateInferredMethodParametersTypes(
                    n, receiverTree, method, analysis.getTypeFactory());
```

framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
pbsf: Given a `ClassTree`, is it possible to obtain the fully qualified name of the class? In case that is possible I could remove this dependency and clean up some code.

framework/src/org/checkerframework/common/signatureinference/SignatureInferenceScenes.java
pbsf: I split this into 3 classes:
`WholeProgramInferenceScenes`
`WholeProgramInferenceScenesHelper`
`AnnotationConverter`

wmdietl:
Hi Paulo, thanks for addressing so much of the feedback already. You are still assigned to the PR. Are there further changes you want to make or should someone have a look through the current state?

framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
wmdietl: You can probably ask for the Element corresponding to the ClassTree and then use one of the helpers in ElementUtils.

pbsf: I tried this before but faced problems when trying to get the fully qualified name of an inner class.

`ElementUtils.getQualifiedClassName(...).toString()` returns `packg.Class.InnerClass` instead of `packg.Class$InnerClass`.

That's the only reason I sticked with `ClassTree`, but it could be removed if this problem is fixed.

framework/src/org/checkerframework/framework/source/SourceChecker.java
pbsf: I think the new method would be identical, having the exact same number of lines. Only three lines would be different. Therefore I think keeping this one public is better than adding code duplication. Please let me know if you agree.

framework/tests/signature-inference/non-annotated/InheritanceTest.java
pbsf: The plan is to file an issue.

framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
```diff
@@ -723,12 +732,46 @@ protected void addInformationFromPreconditions(S info,
 
         S info = in.getRegularStore();
         V rhsValue = in.getValueOfSubNode(rhs);
+        Receiver expr = FlowExpressions.internalReprOf(
+                analysis.getTypeFactory(), lhs);
+        if (performWholePrograminference(
+                n.getTree(), InternalUtils.symbol(lhs.getTree()))) {
+            if (expr instanceof FieldAccess) {
```
smillst: Should this be `lhs` instead of `expr`? Or should `expr` be passed to updateInferredFieldType instead of `lhs`?

pbsf: You are right. I'm now using `lhs instanceof FieldAccessNode` and removed `expr` in both locations. I pushed the changes.

pbsf:
@wmdietl, feel free to have a look at the PR. Thanks!

smillst:
I'm going to review this now as we are close to the release.

wmdietl:
I think we should merge this after the May 1 release instead of so close before it.

mernst:
Our goal is to get it into the May 1 release.  I'd like to try to do that if possible.  Let's keep pushing on this for now.

framework/src/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
```diff
@@ -0,0 +1,432 @@
...
+     * @param methodElt the element of the method being invoked.
+     * @param atf the annotated type factory of a given type system, whose
+     * type hierarchy will be used to update the method parameters' types.
+     */
+    @Override
+    public void updateInferredMethodParameterTypes(
+            MethodInvocationNode methodInvNode, Tree receiverTree,
+            ExecutableElement methodElt, AnnotatedTypeFactory atf) {
+        if (receiverTree == null) {
+            // TODO: Method called from static context.
```
smillst: This happens when a static method is called from a static context correct?

smillst:
I've reviewed the code and made several changes (https://github.com/pbsf/checker-framework/pull/15).  I also added issues #682 and #683 to track some todos in the code. I still need to review the tests.

framework/src/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
pbsf: Correct.

smillst:
This branch was rebased.  New pull request is #687

PR #614, opened 2016-03-02
Add more classes to the annotated JDK

atomicknight:


checker/jdk/nullness/src/java/beans/PropertyDescriptor.java
```diff
@@ -0,0 +1,754 @@
...
+
+    /**
+     * Returns the property type that corresponds to the read and write method.
+     * The type precedence is given to the readMethod.
+     *
+     * @return the type of the property descriptor or null if both
+     *         read and write methods are null.
+     * @throws IntrospectionException if the read or write method is invalid
+     */
+    private Class<?> findPropertyType(Method readMethod, Method writeMethod)
```
mernst: Per the documentation, I would mark both parameters and the return type as `@Nullable`, even though the method is private.

checker/jdk/nullness/src/org/w3c/dom/Document.java
```diff
@@ -0,0 +1,847 @@
...
+     *   if the <code>qualifiedName</code> has a prefix that is "xml" and
+     *   the <code>namespaceURI</code> is different from "<a href='http://www.w3.org/XML/1998/namespace'>
+     *   http://www.w3.org/XML/1998/namespace</a>" [<a href='http://www.w3.org/TR/1999/REC-xml-names-19990114/'>XML Namespaces</a>]
+     *   . Also raised, when the node being renamed is an attribute, if the
+     *   <code>qualifiedName</code>, or its prefix, is "xmlns" and the
+     *   <code>namespaceURI</code> is different from "<a href='http://www.w3.org/2000/xmlns/'>http://www.w3.org/2000/xmlns/</a>".
+     * @since DOM Level 3
+     */
+    public Node renameNode(Node n,
+                           String namespaceURI,
```
mernst: The Javadoc documentation indicates that `namespaceURI` is permitted to be `null`.
`namespaceURI` is also `@Nullable` in most other places in the pull request.  I suggest reviewing all uses of `namespaceURI` to see whether more should be `@Nullable`.  For example, despite the lack of documentation one way or the other, should it be `@Nullable` in `lookupPrefix` and in `isDefaultNamespace` elsewhere in this pull request?

checker/jdk/nullness/src/org/w3c/dom/Document.java
```diff
@@ -0,0 +1,847 @@
...
+     * <br>To create an element with a qualified name and namespace URI, use
+     * the <code>createElementNS</code> method.
+     * @param tagName The name of the element type to instantiate. For XML,
+     *   this is case-sensitive, otherwise it depends on the
+     *   case-sensitivity of the markup language in use. In that case, the
+     *   name is mapped to the canonical form of that markup by the DOM
+     *   implementation.
+     * @return A new <code>Element</code> object with the
+     *   <code>nodeName</code> attribute set to <code>tagName</code>, and
+     *   <code>localName</code>, <code>prefix</code>, and
```
mernst: This documentation, and nearly-identical documentation in `createAttribute`, suggests that all of `localName`, `prefix`, and `namespaceURI` should be `@Nullable` throughout unless there is an indication otherwise.  Are there such indications?

checker/jdk/nullness/src/org/w3c/dom/Document.java
```diff
@@ -0,0 +1,847 @@
...
+     *  objects supporting a version of the "XMLVersion" feature must not
+     * raise a <code>NOT_SUPPORTED_ERR</code> exception for the same version
+     * number when using <code>Document.xmlVersion</code>.
+     * @exception DOMException
+     *    NOT_SUPPORTED_ERR: Raised if the version is set to a value that is
+     *   not supported by this <code>Document</code> or if this document
+     *   does not support the "XML" feature.
+     * @since DOM Level 3
+     */
+    public void setXmlVersion(String xmlVersion)
```
mernst: The documentation above indicates that the argument may be `@Nullable`.

checker/jdk/nullness/src/org/w3c/dom/Document.java
```diff
@@ -0,0 +1,847 @@
...
+     * performed when setting this attribute; this could result in a
+     * <code>null</code> value returned when using <code>Node.baseURI</code>
+     * .
+     * <br> Beware that when the <code>Document</code> supports the feature
+     * "HTML" [<a href='http://www.w3.org/TR/2003/REC-DOM-Level-2-HTML-20030109'>DOM Level 2 HTML</a>]
+     * , the href attribute of the HTML BASE element takes precedence over
+     * this attribute when computing <code>Node.baseURI</code>.
+     * @since DOM Level 3
+     */
+    public void setDocumentURI(String documentURI);
```
mernst: The documentation above indicates that the parameter should be `@Nullable`.

mernst:
Thanks very much for the pull request, Abraham.  Overall this looks good.  I have noted a few places where it seems to me that annotations are missing; can you take a look?  Thanks again.

atomicknight:
Thanks for the review - I'll take another look at the places you've noted.

Related to one of the comments - is there a convention regarding annotating or not annotating private methods? I skipped all private methods during the first pass but can go back through them if that's desirable. Please let me know. Thanks!

mernst:
> is there a convention regarding annotating or not annotating private methods?

Yes.  The convention is to annotate an entire class at a time, including annotating private methods/fields.
If a class is only partially annotated, the code is confusing to read, since it isn't clear whether an an unannotated type means you determined that the default is desirable, or it means you didn't yet examine that type.

In addition, we eventually plan to type-check the annotated parts of the JDK, and having private methods annotated will help with that.

mernst:
Abraham-
A quick ping:  how is this coming along?  The code freeze for the April 1 release will be in the first half of next week.  Thus, it's not extremely urgent, but it would be great to get these in the release.

atomicknight:
Sorry, I was swamped with work last week and didn't get around to it. I'll definitely have time to work on it this week, so it should be done before the code freeze.

mernst:
Sounds great.  Thanks a lot.

checker/jdk/nullness/src/java/beans/PropertyDescriptor.java
atomicknight: Done.

checker/jdk/nullness/src/org/w3c/dom/Document.java
atomicknight: I think the documentation is misleading in this instance. The JavaDoc seems to be a translated version of the DOM specifications, which are somewhat more lenient than the actual JAXP implementations. Notably, the specifications assume that write and read operations are symmetric, whereas the JAXP implementations do not necessarily abide by such assumptions.

In this particular case, the default JAXP implementation bundled with the JVM will throw a `NullPointerException` if the argument is `null`. I'll add a comment here noting that. I'll also go through the other instances and check what behavior the default implementation exhibits.

checker/jdk/nullness/src/org/w3c/dom/Document.java
atomicknight: Done.

checker/jdk/nullness/src/org/w3c/dom/Document.java
atomicknight: I agree that `prefix` and `namespaceURI` should generally be `@Nullable`. However, `localName` cannot be declared `@Nullable` generally because certain XML node types require a non-null name (e.g. tag names for elements). From testing the various methods using the default JAXP implementation, I've found that `localName` can really only be `null` if the node was created by certain factory methods (like this one).

checker/jdk/nullness/src/org/w3c/dom/Document.java
atomicknight: Done.

atomicknight:
I've reviewed all the classes again and have amended the pull request accordingly (including annotations on non-public methods/fields).

checker/jdk/nullness/src/org/w3c/dom/Document.java
atomicknight: To clarify the previous comment, the name `localName` is used somewhat inconsistently in the APIs. For example, `Document#getElementsByTagNameNS` takes two parameters: the namespace URI and the local name. For this particular API, the local name is actually the element tag name, which is defined to be non-null. Attempting to pass a null argument will seemingly work until you try to iterate over the resulting node list, at which point a `NullPointerException` is thrown.

mernst: I hadn't noticed that `localName` is used inconsistently.  In that case, my comments might not apply to all of them.

> However, localName cannot be declared `@Nullable` generally because certain XML node types require a non-null name (e.g. tag names for elements).

`@Nullable` does not mean "`null` is always a legal argument".
`@NonNull` does mean "`null` is never a legal argument".
Therefore, if there is any invocation of a method that passes `null`, then the correct annotation for that method's parameter is `@Nullable`.

(I think this addresses your comment, but let me know if I am misunderstanding.)

For specific APIs where `null` is never permitted, then `@NonNull` is appropriate, of course.

atomicknight: Sorry, my wording was imprecise. What I meant was: not every occurrence of `localName` can be treated as `@Nullable` because the variable name is not consistently used across different APIs; in certain cases, the name `localName` is used to reference properties that can never be `null`.

mernst: Got it, that makes sense.

PR #620, opened 2016-03-09
Don't cache raw types

smillst:
Sometimes when getAnnotatedType is called on a particular raw type, the assignment context is null and other times it is nonnull. So, don't cache raw types so that getAnnotatedType computes the correct type based on the current assignment context.

(This is really a partial fix to a more general problem of the effect of assignment context on the result of getAnnotatedType.  I'm looking for other places this causes a problems and how to correct cache types that vary based on assignment context.)


smillst:
This fix is not sufficient because the assignment context should really only be used for newClassTrees.  So, I'm closing this PR and will open another one with more changes.

PR #621, opened 2016-03-11
Rawtypes

smillst:
Make raw types have the same types when caching is on and off


checker/build.xml
```diff
@@ -798,6 +798,15 @@
             <param name="param" value="tests.NullnessReflectionTest"/>
         </antcall>
     </target>
+    <target name="nullness-rawtypes-tests" depends="jar,build-tests"
```
wmdietl: Add empty line before.

checker/tests/interning/Raw3.java
```diff
@@ -54,7 +54,7 @@
   // with raw types
   List<@Interned String> foo6() {
     List<String> sl = new ArrayList<>();
-    //:: warning: [unchecked] unchecked conversion
+    //:: warning: [unchecked] unchecked conversion :: error: (return.type.incompatible)
```
wmdietl: We previously wanted that we can always assign a raw type to a generic type - the programmer already get's the javac warning about the unchecked conversion. Are we sure we want to impose this possibly large numbers of errors on people stuck with raw types?

checker/tests/nullness-rawtypes-nocaching/RawTypeAssignments.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@SuppressWarnings("unchecked")
+public class RawTypeAssignments {
+    Map rawMap = new HashMap();
+    Map<String, List<String>> notRawMapDiamondRec = new HashMap<>();
+    //:: error: (assignment.type.incompatible)
+    Map<String, List<String>> notRawMapRawHashMapRec = new HashMap();
```
wmdietl: Same comment as above: we previously wanted to allow such assignments.

checker/tests/nullness-rawtypes-nocaching/RawTypesBounded.java
```diff
@@ -0,0 +1,66 @@
...
+
+    void useTypeArg(){
+        BoundedGeneric rawLocal = new BoundedGeneric<String>();
+        BoundedGeneric<String> generic1 = rawReturn();
+        //:: error: (assignment.type.incompatible)
+        BoundedGeneric<String> generic2 = rawField;
+        //:: error: (assignment.type.incompatible)
+        BoundedGeneric<String> generic3 = rawLocal;
+    }
+
```
wmdietl: Inconsistent space between methods.

checker/tests/src/tests/NullnessRawTypesNoCachingTest.java
```diff
@@ -0,0 +1,27 @@
...
+ * Tests for the Nullness Checker on raw types with caching turned off.
+ */
+public class NullnessRawTypesNoCachingTest extends CheckerFrameworkTest {
+
+    public NullnessRawTypesNoCachingTest(File testFile) {
+        super(testFile,
+                org.checkerframework.checker.nullness.NullnessChecker.class,
+                "nullness",
+                "-Anomsgtext",
+              "-AatfDoNotCache" ,"-AatfDoNotReadCache");
```
wmdietl: Inconsistent indentation.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1106,36 +1106,23 @@ public AnnotatedTypeMirror fromTypeTree(Tree tree) {
         }
 
         AnnotatedTypeMirror result = TypeFromTree.fromTypeTree(this, tree);
-
+
```
wmdietl: Spurious space.

checker/tests/interning/Raw3.java
mernst: I tend to agree that if javac already issues a warning, the Checker Framework probably doesn't also need one.  (Except that we really want people not to use raw types when using the Checker Framework!)

wmdietl:
The code previously used the assignment context to determine arguments to use for raw types. This change always uses uninferred wildcards, resulting in many more warnings.
The previous policy was that the compiler already gives an "unchecked" warning for these locations and that we don't want to generate all these additional errors.
We should discuss whether we want to change this.

checker/tests/src/tests/NullnessRawTypesNoCachingTest.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    @Parameters
+    public static String[] getTestDirs() {
+        return new String[]{"nullness-rawtypes-nocaching"};
```
wmdietl: Why do we have two different test cases for caching and no caching? Isn't our goal to make the behavior the same between the two? I.e. have one directory of tests that get's used by the checker with and without caching.

checker/tests/interning/Raw3.java
smillst: If the  [ignoreRawTypeArguments](http://types.cs.washington.edu/checker-framework/current/checker-framework-manual.html#generics-raw-types) option is supplied, then this error is not issued.

smillst: Also, in this particular case, the type arguments for the raw type are correctly inferred.

wmdietl: Then why is there an additional error?

smillst: Because the type of `s1` is `List<@UnknownInterned String>`,  so the type of `(List) s1` should also be `List<@UnknownInterned String>`.  The return type of the method is `List<@Interned String>`, so the returned expression is not a subtype of the return type.

checker/tests/src/tests/NullnessRawTypesNoCachingTest.java
smillst: I'll make them use the same directory.  They previously issued different warnings (d0b246abb842eb) and I wanted to track the differences.

smillst:
@wmdietl  commented:
>The code previously used the assignment context to determine arguments to use for raw types. This change always uses uninferred wildcards, resulting in many more warnings.

The assignment context is always null the first time `fromTypeTree(Tree tree)` is called, so the uninferred wildcards were previously cached.  So, except in case of a cache miss, no more errors are issued after this change than before.  d0b246abb842ebb shows which errors where issued with and without caching before the changes to `fromTypeTree` in this pull request.

smillst:
@wmdietl @mernst  I addressed the white space issues, removed the test with caching turned off, and added tests using the ignoreRawTypesArguments option.  The new tests (10b99c9) show that using this option prevents the assignment.type.incompatible errors when a raw type is assigned.  The assignment.type.incompatible errors are still issued at uses of raw types (see [checker/tests/nullness-rawtypes-ignore/RawTypeUses.java](https://github.com/typetools/checker-framework/blob/10b99c92acf989ea8cdff77db1d2b4355a541112/checker/tests/nullness-rawtypes-ignore/RawTypeUses.java)  for examples.)

checker/tests/interning/Raw3.java
wmdietl: You are saying that we should prefer propagating the type from the cast expression instead of preferring the assignment context, right?
Was the previous behavior that the raw type in the cast used the assignment context? Then because a cast from `List<@UnknownInderned String>` (the type of `s1`) to `List<@Interned String>` (the return type) is legal, the previous behavior allowed this?

checker/tests/src/tests/NullnessIgnoreRawTypesTest.java
```diff
@@ -0,0 +1,27 @@
...
+import org.junit.runners.Parameterized.Parameters;
+
+/**
+ * Tests for the Nullness Checker on raw types using the ignoreRawTypeArguments option
+ */
+public class NullnessIgnoreRawTypesTest extends CheckerFrameworkTest {
+
+    public NullnessIgnoreRawTypesTest(File testFile) {
+        super(testFile,
+                org.checkerframework.checker.nullness.NullnessChecker.class,
```
wmdietl: Both of these tests only use the freedom-before-commitment system. Should they also test with rawness?

checker/tests/src/tests/NullnessRawTypesTest.java
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Tests for the Nullness Checker on raw types
+ */
+public class NullnessRawTypesTest extends CheckerFrameworkTest {
+
+    public NullnessRawTypesTest(File testFile) {
+        super(testFile,
+                org.checkerframework.checker.nullness.NullnessChecker.class,
+                "nullness",
+                "-Anomsgtext");
```
wmdietl: This test suite doesn't pass any special compiler flags. Is it worth adding yet another top-level nullness test directory for these three files?
By moving to the normal nullness/rawness directory, you would also get both FbC and rawness coverage.

smillst:
Closed in favor of #637.

PR #623, opened 2016-03-14
getDefaultedAnnotatedType and other clean ups

smillst:
This pull request covers included several related refactorings.
1. getDefaultedAnnotatedType renamed to getAnnotatedTypeLhs (Name and use of method is up for debate.)
2. getAnnotatedTypeLhs now calls getAnnotatedType with `useFlow` set to false, and caching turned off.
3. New method getAnnotatedTypeLhsNoTypeVarDefault that returns the type of a type variable with out dataflow refinement and without local variable defaulting.  (Used during type argument inference)
4. shouldCache and shouldReadCache merged into one variable, shouldCache
5. "postAsMemberHack" in the Initialization and Nullness Checkers has been simplified and documented.

(Note, this is a pull request to the rawtypes branch because without the rawtypes changes, the changes in this pull request will cause tests to fail.)

### DIFF
--- a/checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
+++ b/checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
@@ -78,8 +78,30 @@ public abstract class InitializationAnnotatedTypeFactory<
         Flow extends CFAbstractAnalysis<Value, Store, Transfer>>
     extends GenericAnnotatedTypeFactory<Value, Store, Transfer, Flow> {
 
-    /** Annotation constants */
-    protected final AnnotationMirror COMMITTED, FREE, FBCBOTTOM, NOT_ONLY_COMMITTED, UNCLASSIFIED;
+    /**
+     * UnknownInitialization or Raw
+     */
+    protected final AnnotationMirror UNCLASSIFIED;
+
+    /**
+     * Initialized or NonRaw
+     */
+    protected final AnnotationMirror COMMITTED;
+
+    /**
+     * UnderInitialization or null
+     */
+    protected final AnnotationMirror FREE;
+
+    /**
+     * NotOnlyInitialized or null
+     */
+    protected final AnnotationMirror NOT_ONLY_COMMITTED;
+
+    /**
+     * FBCBottom or NonRaw
+     */
+    protected final AnnotationMirror FBCBOTTOM;
 
     /**
      * Should the initialization type system be FBC? If not, the rawness type
@@ -589,19 +611,12 @@ public abstract class InitializationAnnotatedTypeFactory<
                 // The receiver is initialized for this frame.
                 // Change the type of the field to @UnknownInitialization so that
                 // anything can be assigned to this field.
-                type.replaceAnnotation(qualHierarchy.getTopAnnotation(UNCLASSIFIED));
+                type.replaceAnnotation(UNCLASSIFIED);
 
             } else if(computingAnnotatedTypeMirrorOfLHS) {
                 // The receiver is not initialized for this frame, and the type of a lhs is being computed.
-                // Only replace initialization annotations.
-                Set<AnnotationMirror> set = AnnotationUtils.createAnnotationSet();
-                set.addAll(type.getAnnotations());
-                for (AnnotationMirror anno : set) {
-                    if (isInitializationAnnotation(anno)) {
-                        type.removeAnnotation(anno);
-                        type.addAnnotation(qualHierarchy.getTopAnnotation(anno));
-                    }
-                }
+                // Only replace initialization annotation.
+                type.replaceAnnotation(UNCLASSIFIED);
             } else {
                 // The receiver is not initialized for this frame and the type being computed is not a LHS.
                 // Replace all annotations with the top annotation for that hierarchy.
--- a/checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
+++ b/checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
@@ -345,7 +345,6 @@ public abstract class InitializationAnnotatedTypeFactory<
     /**
      * Controls which hierarchies' qualifiers are changed based on the
      * receiver type and the declared annotations for a field.
-     * (committed-only).
      * @see #computeFieldAccessType
      * @see #getAnnotatedTypeLhs(Tree)
      */
@@ -353,13 +352,10 @@ public abstract class InitializationAnnotatedTypeFactory<
 
     @Override
     public AnnotatedTypeMirror getAnnotatedTypeLhs(Tree lhsTree) {
-        boolean oldOnlyComputeFieldCommittedHeirachry = computingAnnotatedTypeMirrorOfLHS;
+        boolean oldComputingAnnotatedTypeMirrorOfLHS = computingAnnotatedTypeMirrorOfLHS;
         computingAnnotatedTypeMirrorOfLHS = true;
-
         AnnotatedTypeMirror result = super.getAnnotatedTypeLhs(lhsTree);
-
-        computingAnnotatedTypeMirrorOfLHS = oldOnlyComputeFieldCommittedHeirachry;
-
+        computingAnnotatedTypeMirrorOfLHS = oldComputingAnnotatedTypeMirrorOfLHS;
         return result;
     }
 
@@ -539,7 +535,7 @@ public abstract class InitializationAnnotatedTypeFactory<
             return false;
 
         Name when = AnnotationUtils.getElementValueClassName(unused, "when",
-                false);
+                                                             false);
         for (AnnotationMirror anno : receiverAnnos) {
             Name annoName = ((TypeElement) anno.getAnnotationType().asElement())
                     .getQualifiedName();
@@ -560,8 +556,7 @@ public abstract class InitializationAnnotatedTypeFactory<
 
     /**
      * Determine the type of a field access (implicit or explicit) based on the
-     * receiver type and the declared annotations for the field
-     * (committed-only).
+     * receiver type and the declared annotations for the field.
      *
      * @param type
      *            Type of the field access expression.
@@ -591,27 +586,25 @@ public abstract class InitializationAnnotatedTypeFactory<
                     .asType();
             boolean isInitializedForFrame = isInitializedForFrame(receiverType, fieldDeclarationType);
             if (isInitializedForFrame) {
-                // receiver is initialized for this frame.
-                // change the type of the field to @UnknownInitialization so that
+                // The receiver is initialized for this frame.
+                // Change the type of the field to @UnknownInitialization so that
                 // anything can be assigned to this field.
                 type.replaceAnnotation(qualHierarchy.getTopAnnotation(UNCLASSIFIED));
 
             } else if(computingAnnotatedTypeMirrorOfLHS) {
-                // receiver is not initialized for this frame and the type of a lhs is being computed
-                // so replace all annotations with top expect for the field invariant annotation hierarchy
-                // (Nullness, for example)
-                AnnotationMirror childTypeSystemAnnotation = type.getAnnotationInHierarchy(getFieldInvariantAnnotation());
+                // The receiver is not initialized for this frame, and the type of a lhs is being computed.
+                // Only replace initialization annotations.
                 Set<AnnotationMirror> set = AnnotationUtils.createAnnotationSet();
                 set.addAll(type.getAnnotations());
                 for (AnnotationMirror anno : set) {
-                    if (!AnnotationUtils.areSame(childTypeSystemAnnotation, anno)) {
+                    if (isInitializationAnnotation(anno)) {
                         type.removeAnnotation(anno);
                         type.addAnnotation(qualHierarchy.getTopAnnotation(anno));
                     }
                 }
             } else {
-                // receiver is not initialized for this frame and the type being computed is not a LHS
-                // so replace all annotations with the top annotation for that hierarchy
+                // The receiver is not initialized for this frame and the type being computed is not a LHS.
+                // Replace all annotations with the top annotation for that hierarchy.
                 type.clearAnnotations();
                 type.addAnnotations(qualHierarchy.getTopAnnotations());
             }
--- a/checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
+++ b/checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
@@ -321,8 +321,6 @@ public abstract class InitializationAnnotatedTypeFactory<
         return true;
     }
 
-    protected boolean HACK_DONT_CALL_POST_AS_MEMBER = false;
-
     /**
      * {@inheritDoc}
      *
@@ -337,14 +335,32 @@ public abstract class InitializationAnnotatedTypeFactory<
             AnnotatedTypeMirror owner, Element element) {
         super.postAsMemberOf(type, owner, element);
 
-        if (!HACK_DONT_CALL_POST_AS_MEMBER) {
-            if (element.getKind().isField()) {
-                Collection<? extends AnnotationMirror> declaredFieldAnnotations = getDeclAnnotations(element);
-                AnnotatedTypeMirror fieldAnnotations = getAnnotatedType(element);
-                computeFieldAccessType(type, declaredFieldAnnotations, owner,
-                        fieldAnnotations, element);
-            }
+        if (element.getKind().isField()) {
+            Collection<? extends AnnotationMirror> declaredFieldAnnotations = getDeclAnnotations(element);
+            AnnotatedTypeMirror fieldAnnotations = getAnnotatedType(element);
+            computeFieldAccessType(type, declaredFieldAnnotations, owner, fieldAnnotations, element);
         }
+    }
+
+    /**
+     * Controls which hierarchies' qualifiers are changed based on the
+     * receiver type and the declared annotations for a field.
+     * (committed-only).
+     * @see #computeFieldAccessType
+     * @see #getAnnotatedTypeLhs(Tree)
+     */
+    private boolean computingAnnotatedTypeMirrorOfLHS = false;
+
+    @Override
+    public AnnotatedTypeMirror getAnnotatedTypeLhs(Tree lhsTree) {
+        boolean oldOnlyComputeFieldCommittedHeirachry = computingAnnotatedTypeMirrorOfLHS;
+        computingAnnotatedTypeMirrorOfLHS = true;
+
+        AnnotatedTypeMirror result = super.getAnnotatedTypeLhs(lhsTree);
+
+        computingAnnotatedTypeMirrorOfLHS = oldOnlyComputeFieldCommittedHeirachry;
+
+        return result;
     }
 
     @Override
@@ -575,8 +591,27 @@ public abstract class InitializationAnnotatedTypeFactory<
                     .asType();
             boolean isInitializedForFrame = isInitializedForFrame(receiverType, fieldDeclarationType);
             if (isInitializedForFrame) {
+                // receiver is initialized for this frame.
+                // change the type of the field to @UnknownInitialization so that
+                // anything can be assigned to this field.
                 type.replaceAnnotation(qualHierarchy.getTopAnnotation(UNCLASSIFIED));
+
+            } else if(computingAnnotatedTypeMirrorOfLHS) {
+                // receiver is not initialized for this frame and the type of a lhs is being computed
+                // so replace all annotations with top expect for the field invariant annotation hierarchy
+                // (Nullness, for example)
+                AnnotationMirror childTypeSystemAnnotation = type.getAnnotationInHierarchy(getFieldInvariantAnnotation());
+                Set<AnnotationMirror> set = AnnotationUtils.createAnnotationSet();
+                set.addAll(type.getAnnotations());
+                for (AnnotationMirror anno : set) {
+                    if (!AnnotationUtils.areSame(childTypeSystemAnnotation, anno)) {
+                        type.removeAnnotation(anno);
+                        type.addAnnotation(qualHierarchy.getTopAnnotation(anno));
+                    }
+                }
             } else {
+                // receiver is not initialized for this frame and the type being computed is not a LHS
+                // so replace all annotations with the top annotation for that hierarchy
                 type.clearAnnotations();
                 type.addAnnotations(qualHierarchy.getTopAnnotations());
             }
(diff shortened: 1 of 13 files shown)


framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
```diff
@@ -150,13 +150,15 @@ protected V getValueFromFactory(Tree tree, Node node) {
...
-            boolean oldFlow = factory.getUseFlow();
-            factory.setUseFlow(false);
-            Element element = node.getAssignmentContext().getElementForType();
-            if (element != null) {
-                AnnotatedTypeMirror assCtxt = factory.getAnnotatedType(element);
+            Tree contextTree = node.getAssignmentContext().getContextTree();
+            AnnotatedTypeMirror assCtxt = null;
+            if (contextTree != null) {
+                assCtxt = factory.getAnnotatedTypeLhs(contextTree);
+            } else if( node.getAssignmentContext().getElementForType() != null) {
```
wmdietl: Space after `if` here and on line 161. No space after `(`.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -334,13 +331,12 @@ public AnnotatedTypeFactory(BaseTypeChecker checker) {
 
         this.cacheDeclAnnos = new HashMap<Element, Set<AnnotationMirror>>();
 
-        int cacheSize = getCacheSize();
+        this.shouldCache = !checker.hasOption("atfDoNotCache");
+        int cacheSize = shouldCache ? getCacheSize() : 0;
```
wmdietl: I would make this more explicit: when shouldcache is false, set fields to `null` otherwise determine cache size and create them.

wmdietl:
It's great that this removes more lines than it adds and that HACK-mode is removed! I'll finish my review tomorrow.

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -337,16 +335,34 @@ public void postAsMemberOf(AnnotatedTypeMirror type,
...
+     * receiver type and the declared annotations for a field.
+     * (committed-only).
+     * @see #computeFieldAccessType
+     * @see #getAnnotatedTypeLhs(Tree)
+     */
+    private boolean computingAnnotatedTypeMirrorOfLHS = false;
+
+    @Override
+    public AnnotatedTypeMirror getAnnotatedTypeLhs(Tree lhsTree) {
+        boolean oldOnlyComputeFieldCommittedHeirachry = computingAnnotatedTypeMirrorOfLHS;
```
wmdietl: I would name the local variable like the field, prepending "old". If nothing else, change Heirarchry -> Hierarchy.

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -337,16 +335,34 @@ public void postAsMemberOf(AnnotatedTypeMirror type,
...
+            Collection<? extends AnnotationMirror> declaredFieldAnnotations = getDeclAnnotations(element);
+            AnnotatedTypeMirror fieldAnnotations = getAnnotatedType(element);
+            computeFieldAccessType(type, declaredFieldAnnotations, owner, fieldAnnotations, element);
         }
     }
 
+    /**
+     * Controls which hierarchies' qualifiers are changed based on the
+     * receiver type and the declared annotations for a field.
+     * (committed-only).
```
wmdietl: It's not clear what this parenthesized expression is referring to.

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -575,8 +591,27 @@ private void computeFieldAccessType(AnnotatedTypeMirror type,
                     .asType();
             boolean isInitializedForFrame = isInitializedForFrame(receiverType, fieldDeclarationType);
             if (isInitializedForFrame) {
+                // receiver is initialized for this frame.
+                // change the type of the field to @UnknownInitialization so that
+                // anything can be assigned to this field.
                 type.replaceAnnotation(qualHierarchy.getTopAnnotation(UNCLASSIFIED));
+
+            } else if(computingAnnotatedTypeMirrorOfLHS) {
```
wmdietl: Space after "if".

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -575,8 +591,27 @@ private void computeFieldAccessType(AnnotatedTypeMirror type,
                     .asType();
             boolean isInitializedForFrame = isInitializedForFrame(receiverType, fieldDeclarationType);
             if (isInitializedForFrame) {
+                // receiver is initialized for this frame.
```
wmdietl: "receiver" is not referring to a variable, so I would make it "The receiver". I would also capitalize "change" on the next line to make it a proper sentence.

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -575,8 +591,27 @@ private void computeFieldAccessType(AnnotatedTypeMirror type,
...
+                // change the type of the field to @UnknownInitialization so that
+                // anything can be assigned to this field.
                 type.replaceAnnotation(qualHierarchy.getTopAnnotation(UNCLASSIFIED));
+
+            } else if(computingAnnotatedTypeMirrorOfLHS) {
+                // receiver is not initialized for this frame and the type of a lhs is being computed
```
wmdietl: Same comment about capitalization as above. Also full stop or comma at the end of this line?

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -575,8 +591,27 @@ private void computeFieldAccessType(AnnotatedTypeMirror type,
...
+                // so replace all annotations with top expect for the field invariant annotation hierarchy
```
wmdietl: I don't think this is a "field invariant annotation hierarchy". A type hierarchy might have a single field invariant annotation, which is returned by getFieldInvariantAnnotation(). For Nullness, I think it's MonotonicNonNull, but you should check.

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -575,8 +591,27 @@ private void computeFieldAccessType(AnnotatedTypeMirror type,
...
+                // (Nullness, for example)
+                AnnotationMirror childTypeSystemAnnotation = type.getAnnotationInHierarchy(getFieldInvariantAnnotation());
```
wmdietl: Can you explain the logic of this code? By doing the type.getAIH(getFieltInvariantA) you get whatever qualifier from the child hierarchy the type contains. Then you only replace annotations that are not in that hierarchy. So for Nullness, you don't replace the annotation with the top annotation. I would have expected from the documentation that you replace everything other than the field invariant annotation with top.

framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
wmdietl: getElementForType https://github.com/typetools/checker-framework/blob/master/dataflow/src/org/checkerframework/dataflow/cfg/node/AssignmentContext.java#L40 doesn't look particularly cheap. I would compute it once and put it in a local, instead of computing it twice.

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -896,31 +910,59 @@ protected String dotOutputFileName(UnderlyingAST ast) {
...
-            annotateImplicit(tree, res, false);
+        boolean oldUseFlow = useFlow;
+        boolean oldShouldCache = shouldCache;
+        useFlow = false;
+        // Don't cache the result because getAnnotatedType(lhsTree) could
+        // be called and would expect a different result.
+        shouldCache = false;
+        if (lhsTree instanceof VariableTree) {
+            res = getAnnotatedType(lhsTree);
+        } else if (lhsTree instanceof AssignmentTree) {
```
wmdietl: I think the documentation is inconsistent with the implementation: the lhsTree is the whole AssignmentTree, not just the LHS. Should the lhstree be just the "Variable" part and callers strengthened? Why is there an arbitrary expression case? What does it mean to call this function on "foo(bar())"?

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -337,16 +335,34 @@ public void postAsMemberOf(AnnotatedTypeMirror type,
...
+     * @see #computeFieldAccessType
+     * @see #getAnnotatedTypeLhs(Tree)
+     */
+    private boolean computingAnnotatedTypeMirrorOfLHS = false;
+
+    @Override
+    public AnnotatedTypeMirror getAnnotatedTypeLhs(Tree lhsTree) {
+        boolean oldOnlyComputeFieldCommittedHeirachry = computingAnnotatedTypeMirrorOfLHS;
+        computingAnnotatedTypeMirrorOfLHS = true;
+
```
wmdietl: I wouldn't have an empty line after each statement.

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
smillst: The documentation was misleading. By "field invariant annotation hierarchy" I meant "child hierarchy".  I refactored this code to check whether an annotation is in an initialization hierarchy rather than checking for annotations in the child hierarchy.

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -575,8 +586,25 @@ private void computeFieldAccessType(AnnotatedTypeMirror type,
                     .asType();
             boolean isInitializedForFrame = isInitializedForFrame(receiverType, fieldDeclarationType);
             if (isInitializedForFrame) {
+                // The receiver is initialized for this frame.
+                // Change the type of the field to @UnknownInitialization so that
+                // anything can be assigned to this field.
                 type.replaceAnnotation(qualHierarchy.getTopAnnotation(UNCLASSIFIED));
+
+            } else if(computingAnnotatedTypeMirrorOfLHS) {
+                // The receiver is not initialized for this frame, and the type of a lhs is being computed.
```
wmdietl: Isn't this whole branch equivalent to just doing `type.replaceAnnotation(qualHierarchy.getTopAnnotation(UNCLASSIFIED));`?
The type can only have a single initialization annotation, so the above should replace that one with the top for that hierarchy.

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -575,8 +586,25 @@ private void computeFieldAccessType(AnnotatedTypeMirror type,
                     .asType();
             boolean isInitializedForFrame = isInitializedForFrame(receiverType, fieldDeclarationType);
             if (isInitializedForFrame) {
+                // The receiver is initialized for this frame.
+                // Change the type of the field to @UnknownInitialization so that
+                // anything can be assigned to this field.
                 type.replaceAnnotation(qualHierarchy.getTopAnnotation(UNCLASSIFIED));
```
wmdietl: Do you understand why this uses `qualHierarchy.getTopAnnotation(UNCLASSIFIED)` instead of just `UNCLASSIFIED`?
UNCLASSIFIED is either UnknownInitialization or Raw, which are both the tops of their hierarchies.

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -896,31 +910,63 @@ protected String dotOutputFileName(UnderlyingAST ast) {
...
-        } else if (TreeUtils.isExpressionTree(tree)) {
-            res = fromExpression((ExpressionTree) tree);
-            annotateImplicit(tree, res, false);
-        } else {
-            assert false;
+        boolean oldUseFlow = useFlow;
+        boolean oldShouldCache = shouldCache;
+        useFlow = false;
+        // Don't cache the result because getAnnotatedType(lhsTree) could
+        // be called and would expect a different result.
```
wmdietl: This is a bit confusing, because line 957 calls `getAnnotatedType`. Maybe add something like: later calls to getAnnotatedType will expect flow-sensitive type refinements.

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
smillst: Yes, good point.  I fixed this.

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -78,8 +78,30 @@
         Flow extends CFAbstractAnalysis<Value, Store, Transfer>>
     extends GenericAnnotatedTypeFactory<Value, Store, Transfer, Flow> {
 
-    /** Annotation constants */
-    protected final AnnotationMirror COMMITTED, FREE, FBCBOTTOM, NOT_ONLY_COMMITTED, UNCLASSIFIED;
+    /**
+     * UnknownInitialization or Raw
```
wmdietl: These comments would be more useful if they were links to the respective classes.

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -575,8 +608,18 @@ private void computeFieldAccessType(AnnotatedTypeMirror type,
                     .asType();
             boolean isInitializedForFrame = isInitializedForFrame(receiverType, fieldDeclarationType);
             if (isInitializedForFrame) {
-                type.replaceAnnotation(qualHierarchy.getTopAnnotation(UNCLASSIFIED));
+                // The receiver is initialized for this frame.
+                // Change the type of the field to @UnknownInitialization so that
+                // anything can be assigned to this field.
+                type.replaceAnnotation(UNCLASSIFIED);
+
+            } else if(computingAnnotatedTypeMirrorOfLHS) {
```
wmdietl: Space after if.

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -575,8 +608,18 @@ private void computeFieldAccessType(AnnotatedTypeMirror type,
                     .asType();
             boolean isInitializedForFrame = isInitializedForFrame(receiverType, fieldDeclarationType);
             if (isInitializedForFrame) {
-                type.replaceAnnotation(qualHierarchy.getTopAnnotation(UNCLASSIFIED));
+                // The receiver is initialized for this frame.
+                // Change the type of the field to @UnknownInitialization so that
```
wmdietl: UNCLASSIFIED is either `@UnknownInitialization` or `@Raw`.

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -575,8 +608,18 @@ private void computeFieldAccessType(AnnotatedTypeMirror type,
...
+                // anything can be assigned to this field.
+                type.replaceAnnotation(UNCLASSIFIED);
+
+            } else if(computingAnnotatedTypeMirrorOfLHS) {
+                // The receiver is not initialized for this frame, and the type of a lhs is being computed.
+                // Only replace initialization annotation.
```
wmdietl: It would be nice to explain the modifier change in this branch similarly to the previous branch.
If the explanation is the same, would it be easier to merge the branches?

framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
```diff
@@ -150,13 +150,19 @@ protected V getValueFromFactory(Tree tree, Node node) {
...
+                assCtxt = factory.getAnnotatedTypeLhs(contextTree);
+            } else {
+                Element assCtxtElement = node.getAssignmentContext().getElementForType();
+                if (assCtxtElement != null) {
+                    // if contextTree is null, use the element to get the type
+                    assCtxt = factory.getAnnotatedType(assCtxtElement);
+                }
+            }
+
+            if(assCtxt != null) {
```
wmdietl: Space after if.

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -896,31 +910,63 @@ protected String dotOutputFileName(UnderlyingAST ast) {
         return null;
     }
 
+    /**
+     * Returns the type of the left-hand side of an assignment without
+     * applying local variable defaults to type variables.
+     *
+     * The type variables that are types of local variables are defaulted to
```
wmdietl: I thought refinement for type variables worked differently - they aren't assigned top, like for other reference types.
If they are also assigned top, why do we need any special handling for type variables?

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -896,31 +910,63 @@ protected String dotOutputFileName(UnderlyingAST ast) {
...
-        } else if (TreeUtils.isExpressionTree(tree)) {
-            res = fromExpression((ExpressionTree) tree);
-            annotateImplicit(tree, res, false);
-        } else {
-            assert false;
+        boolean oldUseFlow = useFlow;
+        boolean oldShouldCache = shouldCache;
+        useFlow = false;
+        // Don't cache the result because getAnnotatedType(lhsTree) could
+        // be called from elsewhere and would expect expect flow-sensitive type refinements.
```
wmdietl: Duplicate `expect`.

wmdietl:
Nice job! For a few of the changes we should run more tests to ensure everything works correctly.
I don't particularly like `getAnnotatedTypeLhs`, but have run out of alternative suggestions.

smillst:
> For a few of the changes we should run more tests to ensure everything works correctly.

What tests would you like to see?  I've run all the Travis Daikon type-check jobs and the plume-lib type-check.

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
smillst: Type variables are [defaulted to top (or whatever the default for local variables is)](https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/util/defaults/QualifierDefaults.java#L461).  This special handling is for type argument inference.

smillst:
The code review is complete.  The code will be merged after #637 is merged.  It will also have to be rebased, so closing the pull request.

wmdietl:
The pull request didn't specify what testing was already performed. If daikon and plume-lib tests pass, that sounds good.

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
wmdietl: I have to admit that I find the behavior for type variables rather confusing. "@Nullable T" is not a supertype of "T", so defaulting local type variables to the top type would prevent simple code like:

```
<T> T foo(T p) {
  T local = p;
  return local;
}
```

So isn't there more going on than just using the top type for local variables? If we are doing the same thing for type variables and reference types, why would type variables need special treatment?

smillst: In your code example, `T`'s lower bound is `@NonNull Void` and its upper bound is `@Nullable Object`.  `@Nullable T`'s lower bound is `@Nullable Void` and its upper bound is `@Nullable Object`. So `@Nullable T` is a super type of all possible values of `T`.  (And this is true whatever the annotations are on the declaration of `T`.)

So the refined type of `local` after the assignment is `T`  ( with lower bound `@NonNull Void` and upper bound `@Nullable Object`), so the return is legal.

Type variables are treated differently during type argument inference.  For example:
  ````
<S> S bar () {...}
<T> T foo(T p) {
  T local = bar();
  return local;
}
````
During type argument inference of `bar`, the assignment context is  `local`.  If the local variable default is used, then the type of assignment context type is `@Nullable T` and the type argument inferred for `bar()` is `@Nullable T`.  And an incompatible types in return error is issued.

If instead, the local variable default is not applied, then the assignment context type is `T` (with lower bound `@NonNull Void` and upper bound `@Nullable Object`) and the type argument inferred for `bar()` is `T`.  During dataflow, the type of `local` is refined to `T` and the return is legal. 

If the assignment context type was a declared type, for example:
  ````
<S> S bar () {...}
Object foo() {
   Object local = bar();
  return local;
}
````
The local variable default must be used or else the assignment context type is missing an annotation.  So, an incompatible types in return error is issued in the above code.  We could improve type argument inference in this case and by using the lower bound of `S` instead of the local variable default.

smillst: (I don't think this is currently document, I had to figure this out based on failing test cases.  I'll put this explantation some where and add a comment on this method referencing this explantation.)

wmdietl: Thanks for writing this up, this is very helpful! It would definitely be nice to have this somewhere easier to find than this discussion.
Can the solution for type variables be applied to reference types also? Is the problem that for reference types we don't know what to use instead? Couldn't we just use the default that we would use for parameters/returns/fields then?
I think at least opening an issue for this would be good.

smillst:
Rebased version merged: 6e84ba0

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
smillst: >Is the problem that for reference types we don't know what to use instead?  

Yes.

>Couldn't we just use the default that we would use for parameters/returns/fields then?

Something like this would probably work, but it would probably have to be type system dependent.

I've added a comment explaining this:  3c050a6.

wmdietl: We can add a new "LOCAL_VARIABLE_DEFAULT_FOR_TYPE_INFERENCE" location :-)
Anyways, let's open an issue to not forget about this.

PR #626, opened 2016-03-17
Method type argument inference fix for null literals

jyluo:
In method type argument inference, null literals should be inferred and kept as an AnnotatedNullType, and not be replaced by a AnnotatedWildcardType.

If the null literal was passed as the argument to a method such as the following,
```
<T extends Object> T toInfer(T input) {
        return input;
}
```
the type argument is currently inferred as a AnnotatedNullType, but then an AnnotatedWildcardType is created and return with the AnnotatedNullType's annotations applied to its extends bound. It should remain as an AnnotatedNullType.

From the included InferNullType.java test case (and with the SuppressWarnings commented out for demonstration purposes), running it with the Nullness checker produces the following outputs:
```
// Output prior to this fix
$ javac -processor Nullness InferNullType.java
InferNullType.java:9: error: [type.argument.type.incompatible] incompatible types in type argument.
        Object m = toInfer(null);
                          ^
  found   : @Initialized @Nullable ? extends @Initialized @Nullable Object
1 error

// Output with this fix
$ javac -processor Nullness InferNullType.java
InferNullType.java:9: error: [type.argument.type.incompatible] incompatible types in type argument.
        Object m = toInfer(null);
                          ^
  found   : null
1 error
```
The ```type.argument.type.incompatible``` error is expected when using the Nullness checker due to the implicit upper bound of ```@NonNull``` in the declaration of ```<T extends Object>```.

### DIFF
--- a/framework/src/org/checkerframework/framework/type/DefaultRawnessComparer.java
+++ b/framework/src/org/checkerframework/framework/type/DefaultRawnessComparer.java
@@ -2,6 +2,7 @@ package org.checkerframework.framework.type;
 
 import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedArrayType;
 import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedDeclaredType;
+import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedNullType;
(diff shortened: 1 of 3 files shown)


wmdietl:
Thanks for the fix! Looks good to me.

smillst:
Thanks for the fix!

PR #629, opened 2016-03-17
Fixes 624 and related bug in Value Checker

smillst:


framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -175,6 +175,27 @@ private AnnotationMirror createBooleanAnnotationMirror(List<Boolean> values) {
...
+    public TransferResult<CFValue, CFStore> visitStringConcatenateAssignment(
+            StringConcatenateAssignmentNode n, TransferInput<CFValue, CFStore> p) {
+        TransferResult<CFValue, CFStore> result = super.visitStringConcatenateAssignment(
+                n, p);
+        List<String> lefts = getStringValues(n.getLeftOperand(), p);
+        List<String> rights = getStringValues(n.getRightOperand(), p);
+        List<String> concat = new ArrayList<>();
+        for (String left : lefts) {
+            for (String right : rights) {
+                concat.add(left + right);
```
jthaine: This can result in duplicate strings. Should a set be used before converting it back to a list?

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -175,6 +175,27 @@ private AnnotationMirror createBooleanAnnotationMirror(List<Boolean> values) {
...
+            }
+        }
+        AnnotationMirror stringVal = createStringValAnnotationMirror(concat);
```
jthaine: Can the common code between this and the method below be factored out into a private method?

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1481,7 +1481,12 @@ public Void visitUnary(UnaryTree node, Void p) {
      */
     @Override
     public Void visitCompoundAssignment(CompoundAssignmentTree node, Void p) {
-        commonAssignmentCheck(node.getVariable(), node.getExpression(),
+        // Given a compound assignment:
+        // s += expr;
+        // Should be whether s + expr can be assigned to s,
+        // but the "s + expr" tree does not exist.  So instead, check that
+        // s += expr can be assigned to s.
+        commonAssignmentCheck(node.getVariable(), node,
```
jthaine: Recommend making it extra clear in the comment that `s += expr` is `node` in this case. This is very easy to understand when seeing the diff, but once the diff is gone, it may not be immediately obvious.

jthaine:
Looks good. I have only minor comments.

framework/src/org/checkerframework/common/value/ValueTransfer.java
smillst: ValueAnnotatedTypeFactory#createStringAnnotation removes duplicate strings.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1480,10 +1480,15 @@ public Void visitUnary(UnaryTree node, Void p) {
      * {@link #checkAssignability(AnnotatedTypeMirror, Tree)}.
      */
     @Override
-    public Void visitCompoundAssignment(CompoundAssignmentTree node, Void p) {
-        commonAssignmentCheck(node.getVariable(), node.getExpression(),
+    public Void visitCompoundAssignment(CompoundAssignmentTree compoundAssignmentTree, Void p) {
```
wmdietl: All other methods use `node` for the first parameter. I'm not sure using the long name is helpful. (I see this addresses Javier's comment in an indirect way.)

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1480,10 +1480,15 @@ public Void visitUnary(UnaryTree node, Void p) {
...
+        // Given a compound assignment:
+        // s += expr;
+        // Should be whether s + expr can be assigned to s,
```
wmdietl: I had to read the `should be whether` part multiple times. Maybe the sentence can be clarified.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -169,18 +169,27 @@ private AnnotationMirror createBooleanAnnotationMirror(List<Boolean> values) {
...
     }
 
     @Override
-    public TransferResult<CFValue, CFStore> visitStringConcatenate(
-            StringConcatenateNode n, TransferInput<CFValue, CFStore> p) {
-        TransferResult<CFValue, CFStore> result = super.visitStringConcatenate(
-                n, p);
-        List<String> lefts = getStringValues(n.getLeftOperand(), p);
-        List<String> rights = getStringValues(n.getRightOperand(), p);
+    public TransferResult<CFValue, CFStore> visitStringConcatenateAssignment(StringConcatenateAssignmentNode n, TransferInput<CFValue, CFStore> p) {
```
wmdietl: The method signature for the three methods here are very long; I found the previous line breaks easier to read. (I hate horizontal scrolling in github and all the screen space that is wasted...)

wmdietl:
Some small comments. Otherwise I'm fine with merging.

PR #633, opened 2016-03-28
Issue 557: Corrects assignment context for expression in conditional

### LINKED ISSUE #557: Poor type inference for generic method type parameters in conditional expression
Consider following code:

```java
Optional<String> opt()
{
    boolean flag = true;
    return flag ? Optional.of("Hello") : Optional.empty();
}
```

Checker framework raises an error:

> `incompatible types in return.`
> `found: @Initialized @NonNull Optional<@Initialized @Nullable String>`
> `required: @Initialized @NonNull Optional<@Initialized @NonNull String>`

If you add an explicit type parameter, no error is raised:

```java
Optional<String> opt()
{
   boolean flag = true;
   return flag ? Optional.of("Hello") : Optional.<String> empty();
}
```

But in Java 8, you are not required to specify type parameter here because type inference has been improved for complex expressions.

Checker Framework should be as smart as Java compiler on such expressions.

smillst:
Fixes #557

### DIFF
--- a/javacutil/src/org/checkerframework/javacutil/TreeUtils.java
+++ b/javacutil/src/org/checkerframework/javacutil/TreeUtils.java
@@ -313,7 +313,7 @@ public final class TreeUtils {
      * a parameter or a receiver expression to a receiver.)
      *
      * The assignment context for the {@code treePath} is the leaf of its parent,
-     * if the leaf is on of the following trees:
+     * if the leaf is one of the following trees:
      * <ul>
      *   <li>AssignmentTree </li>
      *   <li>CompoundAssignmentTree </li>
--- a/javacutil/src/org/checkerframework/javacutil/TreeUtils.java
+++ b/javacutil/src/org/checkerframework/javacutil/TreeUtils.java
@@ -309,10 +309,11 @@ public final class TreeUtils {
 
     /**
      * Returns the tree with the assignment context for the treePath
-     * leaf node.
+     * leaf node.  (Does not handle pseudo-assignment of an argument to
+     * a parameter or a receiver expression to a receiver.)
      *
-     * The assignment context for the treepath is the most enclosing
-     * tree of type:
+     * The assignment context for the {@code treePath} is the leaf of its parent,
+     * if the leaf is on of the following trees:
      * <ul>
      *   <li>AssignmentTree </li>
      *   <li>CompoundAssignmentTree </li>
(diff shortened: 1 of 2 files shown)


javacutil/src/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -336,8 +337,11 @@ public static Tree getAssignmentContext(final TreePath treePath) {
                 (node instanceof NewArrayTree) ||
                 (node instanceof NewClassTree) ||
                 (node instanceof ReturnTree) ||
-                (node instanceof VariableTree))
+                (node instanceof VariableTree)) {
             return node;
+        } else if (node instanceof ConditionalExpressionTree) {
```
wmdietl: It would be good to also update the documentation of the method.
In particular, it says "The assignment context for the treepath is the most enclosing tree of type xyz"; however, the current implementation only returns a Tree if the parent already is one of those Tree kinds. If it is not, it simply returns null at the bottom of the method.
Should the change be to change the last line to "return getAssignmentContext(path);" to always handle the recursive case if it's not one of the leaf cases?

wmdietl: (I mean changing the return statement instead of adding the branch for the `ConditionalExpressionTree`)

smillst: No, that would return incorrect assignment contexts.  For example:
````
String s = m.toString();
```
Calling this method on the expression `m`  should not return `s`'s tree. 
I've improved the documentation to make this clear.

javacutil/src/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -308,10 +309,11 @@ public static ExpressionTree skipParens(final ExpressionTree tree) {
...
     /**
      * Returns the tree with the assignment context for the treePath
-     * leaf node.
+     * leaf node.  (Does not handle pseudo-assignment of an argument to
+     * a parameter or a receiver expression to a receiver.)
      *
-     * The assignment context for the treepath is the most enclosing
-     * tree of type:
+     * The assignment context for the {@code treePath} is the leaf of its parent,
+     * if the leaf is on of the following trees:
```
wmdietl: `on` -> `one`

javacutil/src/org/checkerframework/javacutil/TreeUtils.java
wmdietl: Looks good.

smillst:
The review of these changes is complete.  I've merged this into the mergeAfterAprilRelease, which I'll merge after the April release.

smillst:
Merged here: 5df9e10

PR #634, opened 2016-03-28
Validate the declared type without refinement

topnessman:
There are two changes here:
1. We need method getDefaultedAnnotatedType to get declared type instead of refined type, which reduces the number of error messages from type refinement during type validation.
2. Second parameter of method getDefaultedAnnotatedType is removed. This change hasn't been reflected in BaseTypeVisitor yet.

### DIFF
--- a/framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
+++ b/framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
@@ -219,7 +219,7 @@ public class BaseTypeVisitor<Factory extends GenericAnnotatedTypeFactory<?, ?, ?
     /**
      * Constructs an instance of the appropriate type factory for the
      * implemented type system.
-     *
+     *eof
      * The default implementation uses the checker naming convention to create
      * the appropriate type factory.  If no factory is found, it returns
      * {@link BaseAnnotatedTypeFactory}.  It reflectively invokes the
@@ -3106,7 +3106,7 @@ public class BaseTypeVisitor<Factory extends GenericAnnotatedTypeFactory<?, ?, ?
             }
             break;
         default:
-            type = atypeFactory.getAnnotatedType(tree);
+        	type = atypeFactory.getDefaultedAnnotatedType(tree);
         }
         return validateType(tree, type);
     }


framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -219,7 +219,7 @@ protected BaseTypeVisitor(BaseTypeChecker checker, Factory typeFactory) {
     /**
      * Constructs an instance of the appropriate type factory for the
      * implemented type system.
-     *
+     *eof
```
mernst: What is the point of this change?

mernst:
I don't see the second change in your changeset, so the commit message is misleading:
> 2. Second parameter of method getDefaultedAnnotatedType is removed. This change hasn't been reflected in BaseTypeVisitor yet.

Can you please adjust either the commit or the commit message, so that they are consistent?  Thanks.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -3106,7 +3106,7 @@ public boolean validateTypeOf(Tree tree) {
             }
             break;
         default:
-            type = atypeFactory.getAnnotatedType(tree);
+        	type = atypeFactory.getDefaultedAnnotatedType(tree);
```
mernst: Please don't make gratuitious whitespace changes.  Source code files should never contain tab characters.

Can you please provide a test case that fails before this change and passes after the change?  That will motivate the need for the change.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
wmdietl: This change should be undone.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
wmdietl: I agree on the whitespace change, it should be undone.

Suzanne and I discussed the motivation for this change here:

https://groups.google.com/d/msg/checker-framework-dev/L321YcG3srY/6wwnC8XKBgAJ

I don't think any of the standard type systems can be used to reproduce this. Once the updated Universe type checker implementation is finished, we'll have a test.

smillst: I agreed that getDefaultedAnnotatedType should be used to validate the LHS of an assignment, but this code uses it for all expressions.  Given my proposed changes to getDefaultedAnnotatedType (#623), a test case is absolutely necessary to prevent a regression.  

I tried to reproduce the error with the PolyAllChecker (framework/tests/poly-all/polyall).
````
import polyall.quals.H1S1;
import polyall.quals.H1S2;
import polyall.quals.H1Top;

public @H1S2 class NewClass {
    public @H1S2 NewClass(){}
    void foo() {
        @H1S2 NewClass classA = new @H1Top NewClass();
        @H1S2 NewClass classA2 = classA;
        @H1S1 NewClass classA3;
    }
}
````

````
javacheck -cp ../../tests/build/ -processor polyall.PolyAllChecker NewClass.java
NewClass.java:8: error: [type.invalid] [@H1Top, @H2Top] may not be applied to the type "@H1Top @H2Top NewClass"
        @H1S2 NewClass classA = new @H1Top NewClass();
                                ^
NewClass.java:10: error: [type.invalid] [@H1S1, @H2Top] may not be applied to the type "@H1S1 @H2Top NewClass"
        @H1S1 NewClass classA3;
                       ^
2 errors
````
But there is no error on the refined type of classA.  So I'm not sure why the Universe type system is issuing an error.

Comment (wmdietl) [inline framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java]: At the moment, without PR #623 , the default branch is used for all kinds of trees, other than type trees and methods.
To me using getDefaultedAnnotatedTypes here makes sense: take the type the programmer wrote for validation, without any refinements. This makes sense both for the LHS and for any other expression.
Let me try to produce a test case with the PolyAll Checker and compare the the UTS.

Comment (smillst) [inline framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java]: >To me using getDefaultedAnnotatedTypes here makes sense: take the type the programmer wrote for validation, without any refinements. This makes sense both for the LHS and for any other expression. 

I don't think this makes sense for any expression for example:

````
void foo(@D MyClass dParam, @C MyClass cParam, boolean flag) {
    @A MyClass d = dParam;
    @A MyClass c = cParam;
    @A MyClass cond = flag ? bRefined : cRefined;
}
````
Assuming `@B <: @A`, `@C <: @B`, `@D <: @B`, what type should getDefaultedAnnotatedType return given the conditional expression tree?

Comment (smillst) [inline framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java]: If we assume that `@B MyClass` is not a valid type, then we have to assume that `@D MyClass` and `@C MyClass` are also not valid types because they are subtypes of `@B MyClass`.  Right?
 
````
void foo(@D MyClass dParam, @C MyClass cParam, boolean flag) {
    @A MyClass d = dParam;
    @A MyClass c = cParam;
    @A MyClass cond = flag ? bRefined : cRefined;
}
````
Currently, the checker would issue an error for `@D MyClass dParam`, `@C MyClass cParam`, both places `dParam` and `cParam` appear, and also  for the expression `flag ? bRefined : cRefined`.

>There is no explicit B in the program, so what would the developer do?

The developer would see the errors on `@D MyClass dParam` and `@C MyClass cParam` and correct the annotations.

> should type validation only validate the explicitly written annotations 

Does this mean that errors would only be issued at `@D MyClass dParam` and `@C MyClass cParam`?

Comment (wmdietl) [inline framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java]: I imagined a system in which C and D would be valid.
I would only expect an error for the B that is introduced as LUB of the conditional expression; I don't want an error for the type of `cond`.
I prefer checking all types for validity, even though this might be redundant. At some point we can have a look and see whether it's worth trying to optimize this to only check explicit types.
I'll try reproducing this error in a standard type system and will get back to you.

Comment (smillst) [inline framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java]: >I'll try reproducing this error in a standard type system and will get back to you.
 Sounds good.  It might also be helpful if I knew the rules of the Universe type system.

Comment (wmdietl) [inline framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java]: Mier (topnessman) is hopefully finished with cleaning up the repo soon, then he can show the current status.
The type rule is that `@Rep` is forbidden in static contexts, e.g. in a static method/field.
Then, we had `@Any Object o = new @Rep Object();` where `@Any` is a supertype of `@Rep`.
We expect one error for the use of `Rep` in the object creation, but not a second one for the local variable or field type.

Comment (wmdietl) [inline framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java]: Your first example (https://github.com/typetools/checker-framework/pull/634#discussion_r57753925) wasn't a valid counterexample, because it didn't involve a type refinement: the RHS was a supertype of the LHS.

To reproduce the issue, do:
- clone or create a branch of the current checker-framework, set up project
- copy the attached PolyAllVisitor.java to checker-framework/framework/tests/test-polyall/polyall
- copy the attached TypeRefinement.java to checker-framework/framework/tests/polyall
- compile the tests
- in checker-framework/framework do something like:

`javac -cp ./build:./tests/build -processor polyall.PolyAllChecker tests/polyall/TypeRefinement.java`

You should see:

```
tests/polyall/TypeRefinement.java:6: error: [polyall.h1s2.forbidden] (polyall.h1s2.forbidden)
	@H1Top Object o2 = new @H1S2 Object ();
	              ^
tests/polyall/TypeRefinement.java:6: error: [polyall.h1s2.forbidden] (polyall.h1s2.forbidden)
	@H1Top Object o2 = new @H1S2 Object ();
	                   ^
2 errors
```

The goal would be to get the error only on the explicit use of `@H1S2` not on the one that is the result of a type refinement.
Is the way we wrote the type validator not the recommended approach?

[BugReport.tar.gz](https://github.com/typetools/checker-framework/files/219724/BugReport.tar.gz)

smillst: Thanks for the test case!  I think the BaseTypeValidator validates all types, even if not explicitly written.  For example:
````
@Nullable T extends Number
````
The annotation on Number is defaulted to `@NonNull`, so the bounds of `T` are not valid.  The base type visitor issues a warning.  

Given this code, which might be in byte code.
```` 
@H1S2 Object getH1S2() {
   return new @H1S2 Object();
}
````
Should `o1 = getH1S2()` issue a warning, too?  

PR #663 is my preferred way of avoiding the error on `o2`.

wmdietl:
Fix in #663 properly integrates the test and fixes this issue.

Issue #635, opened 2016-03-28
Improve inference of type arguments for raw types


smillst:
Type argument inference for raw types should be improved and the "ignoreRawTypeArguments" option should be made optional again.

smillst:
See pull request #637 .

smillst:
Also, checker/jtreg/rawtypes/RawTypeFail.java should be re-enabled and expanded.

smillst:
The following code crashes when -AignoreRawTypeArguments=false is used:

```
import java.util.Collections;
import java.util.Vector;

public class SortVector {
    void test() {
        Vector vector = new Vector();
        Collections.sort(vector);
    }
}
``` 

javacheck -processor tainting SortVector.java -AprintErrorStack -AignoreRawTypeArguments=false

smillst:
Additional test case for raw types:
https://github.com/typetools/checker-framework/issues/2880

PR #637, opened 2016-03-29
Rawtypes (continuation of #621)

smillst:
Type arguments for raw types are inferred using assignment context -- even if the raw type isn't begin assigned.  For example:
```` 
class MyGeneric<T extends @Nullable Object> {
    MyGeneric<T> id() {...}
}
void use(MyGeneric raw1) {
    // no error 
    MyGeneric<String> myGeneric = raw1.id();
}
````
2 problems with this (as it is currently implemented )
 
1.. It's doesn't work for all raw types the same way.  For example:
`````
MyGeneric getRaw() { ... }
void use(MyGeneric raw1) {
    // no error 
    MyGeneric<String> myGeneric = raw1.id();
    // assignment.type.incompatible 
    MyGeneric<String> myGeneri5 = getRaw().id();
}
````
2.. During data flow, the assignment context is null, so raw type inference fails and the result is cached.  This means that when caching is turned on, (more) errors are issued because of raw types. (d0b246abb842ebba648f420bf1d8fbb05333cb42 shows the different errors.)

This pull request removes the code that infers type arguments of raw types in order to avoid the above problems.  Type argument inference for raw types should be improved in the future (see #635). In order to avoid issuing errors about raw types, the "ignoreRawTypeArugments" option is always assumed to have been used.

### DIFF
--- a/framework/src/org/checkerframework/framework/source/SourceChecker.java
+++ b/framework/src/org/checkerframework/framework/source/SourceChecker.java
@@ -1680,12 +1680,6 @@ public abstract class SourceChecker
     // TODO I would like to rename getLintOption to hasLintOption
     @Override
     public final boolean hasOption(String name) {
-
-        // Always ignore errors from raw type arguments.
-        // See https://github.com/typetools/checker-framework/issues/635
-        if (name.equals("ignoreRawTypeArguments")) {
-            return true;
-        }
         return getOptions().containsKey(name);
     }
 
--- a/framework/src/org/checkerframework/framework/source/SourceChecker.java
+++ b/framework/src/org/checkerframework/framework/source/SourceChecker.java
@@ -1680,6 +1680,12 @@ public abstract class SourceChecker
     // TODO I would like to rename getLintOption to hasLintOption
     @Override
     public final boolean hasOption(String name) {
+
+        // Always ignore errors from raw type arguments.
+        // See https://github.com/typetools/checker-framework/issues/635
+        if (name.equals("ignoreRawTypeArguments")) {
+            return true;
+        }
         return getOptions().containsKey(name);
     }
 
--- a/framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
+++ b/framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
@@ -539,7 +539,7 @@ public class AnnotatedTypeFactory implements AnnotationProvider {
      */
     protected TypeHierarchy createTypeHierarchy() {
         return new DefaultTypeHierarchy(checker, getQualifierHierarchy(),
-                                        checker.hasOption("ignoreRawTypeArguments"),
+                                        checker.getOption("ignoreRawTypeArguments", "true").equals("true"),
                                         checker.hasOption("invariantArrays"));
     }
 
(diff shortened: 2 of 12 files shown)

### CODE REVIEW

File: framework/src/org/checkerframework/framework/source/SourceChecker.java [hunk 1]
```diff
@@ -1680,6 +1680,12 @@ public final String getOption(String name) {
     // TODO I would like to rename getLintOption to hasLintOption
     @Override
     public final boolean hasOption(String name) {
+
+        // Always ignore errors from raw type arguments.
+        // See https://github.com/typetools/checker-framework/issues/635
+        if (name.equals("ignoreRawTypeArguments")) {
```
wmdietl: I don't like this way of overriding the default. In the places where we have `getOption("ignoreRawTypeArguments", false)` (hopefully only one) we should change the default to true.
That way, someone very eager can still call the checker with `-AignoreRawTypeArguments=false`

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1114,31 +1114,18 @@ public AnnotatedTypeMirror fromTypeTree(Tree tree) {
...
-                        typeArgs = null;
-                    }
-                } else {
-                    // TODO: the context is null, use uninstantiated wildcards instead.
-                    typeArgs = new ArrayList<AnnotatedTypeMirror>();
-                    AnnotatedDeclaredType declaration = fromElement((TypeElement)dt.getUnderlyingType().asElement());
-                    for (AnnotatedTypeMirror typeParam : declaration.getTypeArguments()) {
-                        AnnotatedWildcardType wct = getUninferredWildcardType((AnnotatedTypeVariable) typeParam);
-                        typeArgs.add(wct);
-                    }
```
wmdietl: I would add a reference to Issue #635 here, to make it easier to find this location again.

framework/src/org/checkerframework/framework/source/SourceChecker.java
smillst: 4 checkers call `hasOption("ignoreRawTypeArguments')`.  Also, AnnotatedTypeFactory and QualifiedTypeFactoryAdapter call `hasOption("ignoreRawTypeArguments')`.  (There are no places that call `getOption("ignoreRawTypeArguments", false)`.)  By adding this code here those call sites don't have to be changed.

smillst:
>There is also a method DefaultTypeHierarchy.fixUpRawTypes is it still needed if we don't mess with raw types anymore?

Yes, because raw type arguments are still used.  For example:
````
List rawList = ...;
Object s = rawList.get(0);
````
The raw type's type argument is used to get the return type of get.

File: framework/src/org/checkerframework/framework/source/SourceChecker.java (hunk 1 above)
Comment (wmdietl) [inline framework/src/org/checkerframework/framework/source/SourceChecker.java]: Yes, but by doing it like this those call sites are confusing and there is no way to provide the alternative behavior.

Comment (smillst) [inline framework/src/org/checkerframework/framework/source/SourceChecker.java]: I don't understand how you are proposing I change the default.

Comment (wmdietl) [inline framework/src/org/checkerframework/framework/source/SourceChecker.java]: Change

```
if (hasOption("ignoreRawTypeArguments") ...
```

into

```
if (getOption("ignoreRawTypeArgument", "true").equals("true")) ...
```

Then when somebody wants the stricter checks, they can pass `-AignoreRawTypeArguments=false`.

smillst:
I've merged this branch to mergeAfterAprilRelease

smillst:
merged here: 	dbaa1b4

Issue #478, opened 2015-09-25
Intersection types fail functional interface test


dbrosoft:
See [gist](https://gist.github.com/dbrosoft/81a013dbcba44b5cd471) for source code and reported errors. `errors0.txt` shows the output of `javac -processor signature -AprintErrorStack IntersectionBug.java`, and `errors1.txt` shows the same after I added a case for intersection types to `org.checkerframework.framework.type.TypeFromTypeTreeVisitor` to get past the constructor invocation.  Oracle's `javac` (1.8.0_51) compiles the source without error, and so does the Checker Framework's compiler if not given a `-processor` argument, but at least `igj`, `javari`, `nullness`, and `signature` cause the compiler to crash.

JonathanBurke:
So, pre-Java8 I believe the only location that an intersection could appear is in a type parameter declaration.  I am sure there are locations within the Checker Framework where we assume this is the case.

dbrosoft:
Ah, that makes sense.  I will have a go at this one, as it is a blocking
issue for compiling the merged annotated JDK.

-- Dan.

dbrosoft:
I've created an 'issue-478' branch with the aforementioned change to `TypeFromTypeTreeVisitor` for anyone who wants to pick this one up.

mernst:
Dan, can you add the test case in the gist to the issue-478 branch, so that there is a test case available in an obvious place?

dbrosoft:
Done.

-- Dan.

PR #638, opened 2016-03-30
IntersectionTypeTree: Issue #573 

### LINKED ISSUE #573: Lambdas: TypeFromTree.defaultAction: conversion undefined for tree type INTERSECTION_TYPE
[MyChronology.java](https://gist.github.com/dbrosoft/594b3a377435e683cf77) is the smallest example I have that produces the above message; strangely, taking any of the definitions out eliminates the error.  Compilation (using the CF compiler) succeeds, but nullness checking crashes.

smillst:
IntersectionTypeTrees should be handled like other TypeTrees.  This fixes #573.  Also fixes the crash in #478 that occurred when caching was turned off.   

Below is a longer explanation of why caching was masking this bug:
The type of the cast type tree, TypeCastTree.getType(), was cached when creating the type of the cast tree, TypeCastTree. (See TypeFromExpressionVisitory#visitTypeCast.)  So, when AnnotatedTypeFactory#getFunctionalInterfaceType called getAnnotatedType on the cast type tree, the cached expression was returned.   If no cached expression was present, the getAnnotatedType called fromExpression on the cast type tree, but the implementation was missing and thus a crash occurred.

### DIFF
--- a/checker/tests/nullness/java8/lambda/Issue573FullTest.java
+++ b/checker/tests/nullness/java8/lambda/Issue573FullTest.java
@@ -1,0 +1,244 @@
+// full test case for Issue 573
+// a short testcase for this issue is Issue573.java in framework/tests/all-systems/java8/lambda
+// https://github.com/typetools/checker-framework/issues/573
+// @below-java8-jdk-skip-test
+
+import java.time.chrono.*;
+
+import static java.time.temporal.ChronoField.ALIGNED_DAY_OF_WEEK_IN_MONTH;
+import static java.time.temporal.ChronoField.ALIGNED_DAY_OF_WEEK_IN_YEAR;
+import static java.time.temporal.ChronoField.ALIGNED_WEEK_OF_MONTH;
+import static java.time.temporal.ChronoField.ALIGNED_WEEK_OF_YEAR;
+import static java.time.temporal.ChronoField.DAY_OF_MONTH;
+import static java.time.temporal.ChronoField.DAY_OF_WEEK;
+import static java.time.temporal.ChronoField.DAY_OF_YEAR;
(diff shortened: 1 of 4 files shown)


checker/tests/nullness/java8/lambda/Issue573FullTest.java
```diff
@@ -0,0 +1,244 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Objects;
+import java.util.ServiceLoader;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+public abstract class Issue573FullTest implements Chronology {
```
smillst: It's the test case from the issue.  The reference is at the top of the file.

wmdietl: Yes, but that class looks very much like:
http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/f0851bc0e7bf/src/share/classes/java/time/chrono/Chronology.java
which has a big fat Oracle copyright on top.

smillst: Good point. I'll either copy the whole file including the copyright or just remove this test case.

smillst:
Merged into mergeAfterAprilRelease

smillst:
Rebased version merged here: 3727a87

wmdietl:
The test fails with OIGJ. I suppressed the test for now:
https://github.com/typetools/checker-framework/commit/b526d3b7d892cb23aa97141bf0777b427e219623
Please have a look whether there is something better to do.

PR #639, opened 2016-03-30
Remove Plug-in Development Environment dependencies from the Eclipse plug-in

jthaine:
My only concern with these changes is that we don't have an Eclipse plug-in test suite to test them. However, they did pass the sanity test that is done before each release.


smillst:
A user tested this to right? https://groups.google.com/forum/#!msg/checker-framework-discuss/J6XzhEiQZ3k/uHtMvD6-AgAJ

jthaine:
Yes. I'm not sure how their testing is different from mine however.

PR #641, opened 2016-03-30
Issue #572: Skip parenthesized trees.

### LINKED ISSUE #572: Lambdas: Functional interface not found for parenthesized lambda expressions
The Checker Framework compiler has no problem with the sample file [ParenthesizedLambda.java](https://gist.github.com/dbrosoft/437c22645955e5d81241), but the nullness checker crashes due to a missing case in the switch statement beginning at line 2921 of [AnnotatedTypeFactory.java](https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java#L2921).

smillst:
Fixes #572

### DIFF
--- a/checker/tests/nullness/java8/lambda/Issue572.java
+++ b/checker/tests/nullness/java8/lambda/Issue572.java
@@ -1,6 +1,5 @@
-// @skip-test
-
 // Test case for issue #572: https://github.com/typetools/checker-framework/issues/572
+// @below-java8-jdk-skip-test
 
 import java.util.function.BiFunction;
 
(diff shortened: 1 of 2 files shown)


checker/tests/nullness/java8/lambda/Issue572.java
```diff
@@ -13,6 +12,7 @@
   }
 
   public <A, B> A konst(@NonNull A a, @Nullable B b) {
+    //:: error: (argument.type.incompatible)
```
wmdietl: Can we also add a passing test case?
Instead of having A and B used as type parameters in two methods, I would use unique names.
Do we even need method type parameters for konst?

PR #645, opened 2016-04-02
Properly include XML classes in annotated JDK binaries

atomicknight:


checker/jdk/Makefile.jdk
```diff
@@ -30,8 +30,8 @@ build: classes
 
 ############## Validate annotated jdk against missing methods ##############
 
-SOURCES = $(shell find src/java src/javax -name '*.java')
-# CLASSES = $(shell find src/java src/javax -name '*.class')
+SOURCES = $(shell find src/java src/javax src/org/w3c src/org/xml -name '*.java')
```
wmdietl: Shouldn't this be "src/org src/com" to include the two directories that are currently not included?

checker/jdk/Makefile.jdk
```diff
@@ -62,7 +62,7 @@ validate: classes $(DIFFS)
 # does not exist.  So long as the variables are not used on such a make
 # invocation, this doesn't cause a problem, but it is irritating (and a bit
 # sloppy).
-BUILD_CLASSES = $(shell find build/java build/javax -name '*.class')
+BUILD_CLASSES = $(shell find build/java build/javax build/org/w3c build/org/xml -name '*.class')
```
wmdietl: Can we have a single variable that contains the packages "java javax com org" and then use Makefile magic to either prepend "src" or "build" to each one? That would reduce the duplication in these two locations.

wmdietl:
Thanks for the PR! I added two comments.

checker/jdk/Makefile.jdk
atomicknight: `src/com` doesn't seem to contain JDK classes, so I didn't include it here. I would have used `src/org`, but I wasn't sure if other non-JDK classes would be placed here as well.

wmdietl: Good point about `com/google`! They will soon be moved to a dedicated repository, if I'm not mistaken here:
https://github.com/typetools/guava
So we don't need to worry about them.

atomicknight: Good to know. What about the `com.sun.*` packages? These seem to be part of the `javac` API as opposed to the JDK itself.

checker/jdk/Makefile.jdk
atomicknight: Depending on whether there are any non-JDK classes in these directories, it might be easier just to use the top-level directory (e.g. `src` or `build`) instead? I'm not too familiar with `make`, so I'm not seeing an obvious way to implement this. Maybe something with sub-shells or `env` will work?

checker/jdk/Makefile.jdk
wmdietl: The JDK, including langtools (= javac, javadoc, etc), will also be annotated in an external repo soon.

checker/jdk/Makefile.jdk
wmdietl: As the whole build system for the annotated jdk will be replaced shortly, how about just including 'org' for now?

atomicknight: I suppose the `-regex` option for `find` would also work:
```bash
PKGS_EXPR = '[^/]+/\(java\|javax\|org/w3c\|org/xml\)/.+'
SOURCES = $(shell find src $(PKGS_EXPR) -name '*.java')
BUILD_CLASSES = $(shell find build $(PKGS_EXPR) -name '*.class')
```

atomicknight:
Updated PR with suggested changes. I don't have `make` installed on my machines right now though, so I haven't been able to test these changes yet.

wmdietl:
Thanks, looks good!

checker/jdk/Makefile.jdk
smillst: The `-regex` options does not work the same with BSD find.  I switched back to duplicating the folders, so that the annotated JDK can be built on Macs without installing GNU find.

PR #646, opened 2016-04-02
Add javax.net.ssl.X509ExtendedKeyManager to the annotated JDK

atomicknight:


mernst:
Thanks for the addition.

PR #648, opened 2016-04-04
Move code that creates AnnotatedNullType to ATF

smillst:
This change allows AnnotatedTypeFactory#toAnnotatedType to be marked protected.  (It's not changed to protected in this pull request because there are additional calls that have to be change first.)

### DIFF
--- a/framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
+++ b/framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
@@ -1306,14 +1306,14 @@ public class AnnotatedTypeFactory implements AnnotationProvider {
     }
 
     /**
-     * Return an AnnotatedNullType with {@code annosToAdd} added
-     * @param annosToAdd annotations to add to the AnnotatedNullType
-     * @return AnnotatedNullType with {@code annosToAdd} added
+     * Creates and returns an AnnotatedNullType qualified with {@code annotations}.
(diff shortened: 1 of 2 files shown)


framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1305,6 +1306,18 @@ protected void annotateInheritedFromClass(/*@Mutable*/ AnnotatedTypeMirror type,
     }
 
     /**
+     * Return an AnnotatedNullType with {@code annosToAdd} added
```
wmdietl: I find the `added` confusing, because I looked for the parameter to which the annotations would be added.
Are there any other annotations already present on the null type? Or will those be the only annotations on it?
I would clarify the parameter name and documentation.

wmdietl:
Can you make all the related changes to make AnnotatedTypeFactory#toAnnotatedType  protected?
Why only this single change instead of all of them together?

smillst:
>Can you make all the related changes to make AnnotatedTypeFactory#toAnnotatedType protected?
Why only this single change instead of all of them together?

Yes, the other call is changed in #649 (2f5444c) and the method is marked protected in #652.  I thought it would be easier to review refactoring separately.

wmdietl:
Sorry, I went through the pull requests bottom-up and didn't see forward references to those upcoming PRs. It makes sense to split them up into small, consistent PRs.

wmdietl:
Looks good to merge.

PR #649, opened 2016-04-04
Refactor calls to AnnotatedTypeFactory methods

smillst:
These changes are so that various AnnotatedTypeFactory methods can be marked non-public.


wmdietl:
These changes look good to me (modulo the tests passing).

Issue #515, opened 2015-11-10
Annotations that should never be used in source code -- except on bounds


smillst:
Several checkers' top and/or bottom annotations are not supposed to be written in source code, but on occasion it is necessary to write the top or bottom annotation on a bound. See [this discussion](https://groups.google.com/forum/#!topic/checker-framework-discuss/u42B7k3hb10) for an example.

We should correct the manual, so that it mentions these possible uses.

wmdietl:
We should also update those qualifiers I assume. Just updating the manual will not fix the issue.

At the moment most type systems use the same defaults for upper and lower bounds of wildcards, that is:
`List<? extends Object>`
`List<? super Object>`
both use the same default qualifier. This is nice and consistent.
Would it be easier to use to use the bottom type for "super" bounds? We can express this using org.checkerframework.framework.qual.DefaultLocation.LOWER_BOUNDS.

mernst:
I agree that using the bottom type for super bounds makes sense -- probably as a default for all type systems.

mernst:
I have separated two other concerns into issue #520 and issue #521.

What remains for this issue is to provide a new meta-annotation (say, called `@TargetTrees`) to indicate locations a particular qualifier is not intended to be written, even though such locations are compatible with the annotation's `@Target` meta-annotation.

An example is that many type systems' bottom qualifier (those with "Bottom" in their name, including `@BottomAll` from issue #521) should only be written on a wildcard super bound.

The `@TargetTrees` should take as an argument a list of `Tree.Kind` locations at which the annotation is permitted.

If a programmer writes an annotation at any other location that is not listed in the `@TargetTrees` meta-annotation, then the Checker Framework issues a warning.  This warning may be suppressed.

The `@TargetTrees` meta-annotation should be added to a number of existing annotations with "Bottom" in their name, and likewise for some with "Top" or "Unknown" in their name.

smillst:
Parts 1 and 2 were completed in pull request #533.

jthaine:
Part 3 is now critical to support the Lock Checker's @GuardSatisfied annotation, which may only be written in certain locations.

jthaine:
When resolving this issue, remove the following method from the Lock Checker: LockVisitor.issueErrorIfGuardSatisfiedAnnotationInUnsupportedLocation

wmdietl:
@topnessman started working on enforcing TargetLocations.

We have a question about compound types.
[...]

PR #650, opened 2016-04-04
Initialization: get explicit annotations from Tree

smillst:
AnnotatedTypeFactory#fromMember should not  be used to get explicit annotations.

### DIFF
--- a/checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
+++ b/checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
@@ -49,12 +49,10 @@ import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.TypeMirror;
 import javax.lang.model.util.Types;
 
-import com.sun.source.tree.AnnotationTree;
 import com.sun.source.tree.ClassTree;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.LiteralTree;
 import com.sun.source.tree.MethodTree;
-import com.sun.source.tree.ModifiersTree;
 import com.sun.source.tree.NewClassTree;
 import com.sun.source.tree.Tree;
 import com.sun.source.tree.Tree.Kind;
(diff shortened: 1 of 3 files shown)


checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -633,6 +635,24 @@ private void computeFieldAccessType(AnnotatedTypeMirror type,
...
 
+    /**
+     * Returns the annotations explicitly written on a constructor result.
+     * Callers should check that {@code constructorDeclaration} is in fact a declaration
+     * of a constructor.
+     *
+     * @param constructorDeclaration Declaration tree of constructor
+     * @return set of annotations explicit on the resulting type of the constructor
+     */
+    public Set<AnnotationMirror> getExplicitAnnotationsOnConstructorResult(MethodTree constructorDeclaration) {
```
wmdietl: Should this method be moved to `AnnotationUtils`? Do we have `getExplicitAnnotationsOnXXX` for other XXX?
It looks like the method could be static and doesn't seem initialization specific.

smillst: Good point.  I moved the method to AnnotationUtils.  InternalUtils has these methods which are similar:
````
List<? extends AnnotationMirror> annotationsFromTree(AnnotatedTypeTree node)
List<? extends AnnotationMirror> annotationsFromTree(TypeParameterTree node)
List<? extends AnnotationMirror> annotationsFromArrayCreation(NewArrayTree node, int level)  
````

File: javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java [hunk 2]
```diff
@@ -556,4 +559,22 @@ public static Name getElementValueClassName(AnnotationMirror anno, CharSequence
...
+
+    /**
+     * Returns the annotations explicitly written on a constructor result.
+     * Callers should check that {@code constructorDeclaration} is in fact a declaration
+     * of a constructor.
+     *
+     * @param constructorDeclaration Declaration tree of constructor
+     * @return set of annotations explicit on the resulting type of the constructor
+     */
+    public static Set<AnnotationMirror> getExplicitAnnotationsOnConstructorResult(MethodTree constructorDeclaration) {
```
wmdietl: Would it make sense to move this to https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/util/ConstructorReturnUtil.java ?
That class only has a single method and seems rather underused.

smillst: I would prefer to leave it in AnnotationUtils where it is more likely to be discovered and used.  I also think when `@TargetLocations` are enforced (Issue #515), we will want to add methods to get explicit annotations from other trees.

wmdietl: Can we just move the single method from ConstructorReturnUtil to AnnotationUtils? It seems counter-intuitive to only have one of the methods about constructor returns in a separate class.

wmdietl:
Looks good to merge.

PR #651, opened 2016-04-04
Refactor member reference code

smillst:
- Move code that gets the return type of a member reference constructor to
AnnotatedTypeFactory.
- Move code that checks is method type argument inference is required to private method.

### DIFF
--- a/framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
+++ b/framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
@@ -44,7 +44,6 @@ import org.checkerframework.framework.type.QualifierHierarchy;
 import org.checkerframework.framework.type.TypeHierarchy;
 import org.checkerframework.framework.type.VisitorState;
 import org.checkerframework.framework.util.AnnotatedTypes;
-import org.checkerframework.framework.util.ConstructorReturnUtil;
 import org.checkerframework.framework.util.ContractsUtils;
 import org.checkerframework.framework.util.FlowExpressionParseUtil;
 import org.checkerframework.framework.util.FlowExpressionParseUtil.FlowExpressionContext;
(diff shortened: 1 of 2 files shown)


framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -2286,40 +2285,13 @@ protected boolean checkMethodReferenceAsOverride(MemberReferenceTree overriderTr
         // ========= Overriding Executable =========
         // The ::method element
         ExecutableElement overridingElement = (ExecutableElement)InternalUtils.symbol(overriderTree);
+        AnnotatedExecutableType overrider =
```
wmdietl: Why is it sometimes `overrider` and sometimes `overriding`? Can we make that more consistent?

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -539,6 +541,35 @@ protected void postDirectSuperTypes(AnnotatedTypeMirror type,
...
+     * Gets the type of the resulting constructor call of a MemberReferenceTree.
+     *
+     * @param memberReferenceTree MemberReferenceTree where the member is a constructor
+     * @param constructorType     AnnotatedExecutableType of the declaration of the constructor
+     * @return AnnotatedTypeMirror of the resulting type of the constructor
+     */
+    public AnnotatedTypeMirror getResultingTypeOfConstructorMemberReference(MemberReferenceTree memberReferenceTree,
+                                                                            AnnotatedExecutableType constructorType) {
+        if (memberReferenceTree.getMode() != MemberReferenceTree.ReferenceMode.NEW) {
+            ErrorReporter.errorAbort("");
```
wmdietl: Some useful message in case this ever gets hit. If this is unlikely, should it be an assert instead?

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -539,6 +541,35 @@ protected void postDirectSuperTypes(AnnotatedTypeMirror type,
...
+
+        // The return type for constructors should only have explicit annotations from the constructor
+        // Recreate some of the logic from TypeFromTree.visitNewClass here.
+
+        // The return type of the constructor will be the type of the expression of the member reference tree.
+        AnnotatedDeclaredType constructorReturnType = (AnnotatedDeclaredType)
+                fromTypeTree(memberReferenceTree.getQualifierExpression());
+
+        // Keep only explicit annotations and those from @Poly
+        ConstructorReturnUtil.keepOnlyExplicitConstructorAnnotations(this, constructorReturnType, constructorType);
```
wmdietl: How is this method related to getExplicitAnnotationsOnConstructorResult in PR #650 ?

smillst: If the constructor result is annotated with `@Poly*`, getExplicitAnnotationsOnConstructorResult returns `@Poly*` where as ConstructorReturnUtil.keepOnlyExplicitConstructorAnnotations substitutes `@Poly*` with whatever annotation it should be given the call site.

I'll change ConstructorReturnUtil.keepOnlyExplicitConstructorAnnotations to use getExplicitAnnotationsOnConstructorResult, once that pull request is merged.

wmdietl: How about putting more constructor-related methods into ConstructorReturnUtil? That would make that class more useful and slim down some of our fatter utility classes.

smillst: I like the fatter utility classes because it means few places to look for methods. Also, see comment my comment here: https://github.com/typetools/checker-framework/pull/650#discussion_r59768859

wmdietl: As noted there I would just move the single method from ConstructorReturnUtil to AnnotationUtils and remove that useless class.

wmdietl:
Fine with merge.

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
smillst: Yes, I'll do this after I merge this pull request and #650.

PR #655, opened 2016-04-05
Save code from old LUB branch.

smillst:
See code comment for more details.


smillst:
(I opened this PR so that GitHub will save the code changes.)

smillst:
The old lub branch 5,268 commits.  All but 5 were from pull master.   This PR includes all the changes in those commits as a single commit.  So, I've deleted the old lub branch.

PR #659, opened 2016-04-07
Add java.lang.Object as supertype, add test for Issue 658, and code cleanups.

wmdietl:


checker/src/org/checkerframework/checker/nullness/NullnessVisitor.java
```diff
@@ -60,7 +60,7 @@
 public class NullnessVisitor extends InitializationVisitor<NullnessAnnotatedTypeFactory,
         NullnessValue, NullnessStore> {
     // Error message keys
-    private static final /*@CompilerMessageKey*/ String ASSIGNMENT_TYPE_INCOMPATIBLE = "assignment.type.incompatible";
+    // private static final /*@CompilerMessageKey*/ String ASSIGNMENT_TYPE_INCOMPATIBLE = "assignment.type.incompatible";
```
smillst: Why did you comment this out? Should it just be deleted?

checker/src/org/checkerframework/checker/regex/RegexUtil.java
```diff
@@ -2,16 +2,16 @@
 
 package org.checkerframework.checker.regex;
 
-import java.util.regex.Pattern;
```
smillst: This needs to be synced to https://github.com/mernst/plume-lib/blob/master/java/src/plume/RegexUtil.java.

smillst:
Looks good modulo my comments. I would have preferred that commits fdb31fb9e01c2b8bd0a3de176939ca15d9b1bafd to eab8a40552948599e5c1f0dca8ff3c549acae719 not been included in the PR.  It's makes it so much harder to review code when formatting and unrelated changes are included.

checker/src/org/checkerframework/checker/nullness/NullnessVisitor.java
wmdietl: It was an unused field and I commented it out to get rid of the warning. I kept it in case a use case arises.

checker/src/org/checkerframework/checker/regex/RegexUtil.java
wmdietl: This was an unused import when I made the pull request. I think trunk already removed this, too.

wmdietl:
Thanks for the comments!

PR #662, opened 2016-04-08
Generalize CFG visualization.

wmdietl:
Instead of having code to produce a DOT visualization of the CFG spread throughout the source code, introduce a new CFGVisualizer abstraction, with the DOTCFGVisualizer being the standard implementation.
Cleans up many of the internal classes and allows adding different visualizations.

Some changes in:
framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
are unrelated and general cleanup.
Changes in checker/jdk/Makefile.jdk produce less noise when building the annotated JDK.


smillst:
Looks good.

wmdietl:
Thanks, merging it.

PR #663, opened 2016-04-14
Validate declared type of LHS, not refined type.

smillst:
(This is an alternative to #634.)

````
 @H1Top Object o = new @H1Invalid Object();
````
Two invalid type errors used to be issued for this code. One for the new Object and one for `o`.


wmdietl:
Thanks for properly integrating the test case and fixing the issue!
The fix looks good to me and I'm fine with merging it. I'll close #634.

PR #664, opened 2016-04-14
Release scripts fixes

jthaine:


release/release_utils.py
```diff
@@ -390,21 +399,26 @@ def hg_push(repo_root):
     else:
         execute('hg -R %s push' % repo_root)
 
-# Pull the latest changes and update
-def update_project(path):
-    if git_bare_repo_exists_at_path(path):
-        execute('git -C %s fetch' % path)
-    elif is_git(path):
-        execute('git -C %s pull' % path)
+# Clean repo, pull the latest changes and update
```
mernst: Can you clarify what "clean repo" means?  It's alarming that this silently discards changes.  Why not just fail if there are discrepancies?

release/release_utils.py
```diff
@@ -526,22 +540,12 @@ def purge(repo, ignored_files=False):
...
-        purge(repo, all)
-        revert(repo) # avoids the issue of purge deleting ignored files we want to get back
-    print ''
-
-def clean_repos(repos, prompt):
-    if maybe_prompt_yn('Remove all modified files, untracked files and outgoing commits from:\n%s ?' % '\n'.join(repos), prompt):
-        for repo in repos:
-            if repo_exists(repo):
-                clean_repo(repo, False)
+def clean_repo(repo):
```
mernst: Every routine in the file should contain a documentation string.

release/release_utils.py
jthaine: The intended meaning of a "clean repo" is a repo that cannot be distinguished from a freshly cloned repo. This is what the script was previously doing - discarding the repos altogether and making a fresh clone each time (without a warning the user could respond to when the --auto option was provided).
The "Verify repositories" step will fail if there are discrepancies, so I'll change the "Clone/update repositories" step so that it doesn't discard changes, allowing the verify step to just fail.

jthaine: Of course, the merge portion of the update step could fail if the repository is unclean, so the script might not get to the verification step.

release/release_utils.py
jthaine: Over 100 documentation strings are missing across the release scripts. I'm working on adding them (I think it's a good idea to do this before the next release while these scripts are fresh in my head), but I think the Lock Checker tasks have a higher priority so I plan to set this aside for right now.

PR #673, opened 2016-04-21
Don't insert an annotation if one with the same name already exists.

smillst:
Normally, the Checker Framework only writes an annotation into an element if an annotation with the same class and value is already there.  This works for most checkers because a duplicated annotation is not allowed.  However, the Lock Checker replaces explicitly written @GuardedBy annotations with a canonicalized versions in the AnnotatedTypeMirrors.  So, TypesIntoElements writes the canonicalized version into the element even though an annotation with the same name is already in the element.  This pull request fix that.  The Lock Checker will contain a test case that tests this.


smillst:
This change is causing test failures.

PR #674, opened 2016-04-22
Don't insert an annotation if one with the same name already exists.

smillst:
Normally, the Checker Framework only writes an annotation into an element if an annotation with the same class and value is already there. This works for most checkers because a duplicated annotation is not allowed. However, the Lock Checker replaces explicitly written @GuardedBy annotations with a canonicalized versions in the AnnotatedTypeMirrors. So, TypesIntoElements writes the canonicalized version into the element even though an annotation with the same name is already in the element. This pull request fix that. The Lock Checker will contain a test case that tests this.


javacutil/src/org/checkerframework/javacutil/TypeAnnotationUtils.java
```diff
@@ -65,6 +65,25 @@ public static boolean isTypeCompoundContained(Types types, List<TypeCompound> li
         return false;
     }
 
+    /**
+     * Check whether a TypeCompound with the same name and position is contained in a list of TypeCompounds.
+     *
```
wmdietl: Some additional contrasting with the existing method would be good.
Add a note to the existing method.

framework/src/org/checkerframework/framework/type/TypesIntoElements.java
```diff
@@ -244,7 +244,8 @@ private static void addUniqueTypeCompounds(Types types, Symbol sym, List<TypeCom
         List<Attribute.TypeCompound> res = List.nil();
 
         for (Attribute.TypeCompound tc : tcs) {
-            if (!TypeAnnotationUtils.isTypeCompoundContained(types, raw, tc)) {
+            // Don't insert an annotation if one with the same name already exists.
+            if (!TypeAnnotationUtils.isTypeCompoundContainedByName(types, raw, tc)) {
```
wmdietl: Would it make sense to replace an existing annotation with the canonical version?
Let's say the programmer wrote `@A("foo")` which get's normalized to `@A("this.foo")` in the ATM.
However, this change seems to leave the annotation in the Element as `@A("foo")`.
Wouldn't it be better to find existing `@A` annotations and replace them with the canonical version?
Aren't you afraid of inconsistent results, in particular when looking at Elements for Trees that might have been thrown out?

smillst: Because the Lock Checker can't differentiate between annotations in elements that have ben canonicalized and those that haven't, canonicalization is always performed by annotateImplicit(Tree, ATM) or annotateImplicit(Element, ATM).  So, it doesn't matter if the annotation in the element is changed to the canonicalized version.  I think it makes more sense to write out the annotation that was written.

mernst: My inclination would have been to canonicalize, too.  But I don't have a deep enough understanding of this code to make a strong pronouncement about it.

smillst: If we want to write out the canonicalized version of an annotation that appears in source code, then we will have to remove the existing annotation from the element.  I don't know how to do that. Werner, is that something you could do?

wmdietl: I will not have time to look at this in the next few days. You should look at sym to see whether it has alternatives to appending lists of TypeCompound. Building up the right `res` should be easy.

wmdietl:
Where does the Lock Checker add the canonical version of the annotation?
There is a `sym.setTypeAttributes(a);` that you can use instead of the `sym.appendUniqueTypeAttributes(res);`.
However, at this point, do we know which one is the canonical version? They are in order, so maybe later annotations should overwrite earlier ones with the same name?
If this is just relevant for the Lock Checker, it might be cleanest to provide a simple API to allow replacing annotations on an Element.
@jthaine , can you point me to where you canonicalize?

javacutil/src/org/checkerframework/javacutil/TypeAnnotationUtils.java
```diff
@@ -65,6 +65,25 @@ public static boolean isTypeCompoundContained(Types types, List<TypeCompound> li
...
+     *
+     * @param list The input list of TypeCompounds.
+     * @param tc The TypeCompound to find.
+     * @return true, iff a TypeCompound with the same name and position as tc is contained in list.
+     */
+    public static boolean isTypeCompoundContainedByName(Types types, List<TypeCompound> list, TypeCompound tc) {
+        for (Attribute.TypeCompound rawat : list) {
+            if (contentEquals(rawat.type.tsym.name, tc.type.tsym.name) &&
+                isSameTAPositionExceptTreePos(rawat.position, tc.position)) {
```
jthaine: The tests pass on my viewpoint adaptation branch if I comment out the following line below:
`p1.location.equals(p2.location) &&`

smillst:
I've add a test case (in this branch) framework/tests/polyall/ReplacingAnnos.java.  The output of it is:
````
/Users/smillst/jsr308/checker-framework/framework/tests/polyall/ReplacingAnnos.java:12: warning: [type.invalid] [@H1S1, @H1S2, @H2Top] may not be applied to the type "@H1S1 @H1S2 @H2Top Object"
        foo(addH1S2);
            ^
1 warning
````
This test case is trying to insert an annotation that is a different class but in the same hierarchy as the one in source code.  The Lock Checker needs to insert an annotation that is the same class but has a different value than the one written in source code.  Should we support both use cases?

javacutil/src/org/checkerframework/javacutil/TypeAnnotationUtils.java
smillst: Commenting out that line cause failures when building the annotated JDK.

smillst:
This is on hold until we want to implement viewpoint adaption for the Lock Checker.  #770

PR #676, opened 2016-04-23
add above-java8-jdk-skip-test annotaion

CharlesZ-Chen:
add a new fake-annotation "@above-java8-jdk-skip-test" in TestUtilities in order to ignore IFlowxxxTest on jdk8


mernst:
If the name is `@above-java8-jdk-skip-test`, then I would expect the test to be active for JDK 8 and only skipped for 9 and more (since it is skipped "above Java 8").  Should the name be changed, or is the pull request comment wrong?

Also, this feature must be documented (as must every feature).  You could document it in `checker-framework/checker/tests/README`.

CharlesZ-Chen:
Thanks for the reviewing. I should changed the name. I'll make the pull request after changing it and adding the documentation. Thanks!

PR #677, opened 2016-04-23
add above-java7-jdk-skip-test annotation

CharlesZ-Chen:
1. Add "@above-java7-jdk-skip-test" in TestUtilities class.
2. Add documentation of this new feature in checker/tests/README.

### DIFF
--- a/checker/tests/README
+++ b/checker/tests/README
@@ -177,6 +177,10 @@ need runtime support (e.g. lambdas) or depends on Java 8 APIs
 Tests that only exercise the static type annotation API do not need
 this marker, as the tests will be executed with a Java 8 compiler.
 
+Write @above-java7-jdk-skip-test anywhere in a test file to disable that
+test if the executing JDK version is higher than 7.
+This is useful if a test has a mismatch between Java 7 and Java 8 (e.g. Tests of Information Flow Checker), and need to ignore this test on Java 8.
+
 To disable all tests for a given type system, temporarily move away a file;
 for example,
   mv $CHECKERFRAMEWORK/checker/tests/src/tests/OIGJTest.java $CHECKERFRAMEWORK/checker/tests/src/tests/OIGJTest.java-SAVE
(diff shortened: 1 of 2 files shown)


checker/tests/README
```diff
@@ -177,6 +177,10 @@ need runtime support (e.g. lambdas) or depends on Java 8 APIs
 Tests that only exercise the static type annotation API do not need
 this marker, as the tests will be executed with a Java 8 compiler.
 
+Write @above-java7-jdk-skip-test anywhere in a test file to disable that
+test if the executing JDK version is higher than 7.
+This is useful if a test has a mismatch between Java 7 and Java 8 (e.g. Tests of Information Flow Checker), and need to ignore this test on Java 8.
```
mernst: Please wrap this line at 80 characters (better, fill the paragraph).
Why is "Tests" capitalized?
Missing word "you" before "need".
Out of curiosity, why is there a mismatch between Java 7 and Java 8 for the Information Flow Checker?  Would it be useful to explain this briefly here?

CharlesZ-Chen:
Thanks for the quick reviewing again!

I will make the documentation to be better.

There is no reason for the capitalised Tests.. I will change it to lower case.

I'm actually new to the checker framework and I've told by Prof. Dietl that there is a mismatch between Java 7 and Java 8 for the IFlow Checker, but I don't know the detail reason of that. I'll check it with Prof. Dietl.

Thanks!

mernst:
FYI, you don't have to close a pull request in order to enhance it.  Just make more commits, push them to the "from" repository for the pull request, and the pull request is automatically updated with your new commits.

CharlesZ-Chen:
Wow, I don't know that before... Thanks!

wmdietl:
Suzanne investigated the Java 8 build failures of the Information Flow Checker (see https://travis-ci.org/typetools/checker-framework-inference/jobs/124549995) and concluded that the errors:

```
Exception: java.lang.AssertionError: Mismatch in type argument size between Set<K> (1) and KeySetView<K extends Object, V extends Object> (2); Stack trace: org.checkerframework.framework.stub.StubParser.annotate(StubParser.java:661)
```

are caused by a mismatch between Java 7 as used by Android and Java 8 libraries.
Disabling all Java 8 tests is undesirable, as other type systems are supposed to be used with Java 8. See discussion here https://github.com/typetools/checker-framework-inference/commit/3d27e684b450bb171f7f6d81a76e31c715c08e8a

However, I don't like the solution of introducing this new skip-test annotation. I'll suggest an alternative in the checker-framework-inference PR at https://github.com/wmdietl/checker-framework-inference/pull/6
I think we can close and abandon this PR.

CharlesZ-Chen:
Ok, I'll re-think about the implementation. Thanks!

PR #678, opened 2016-04-24
Issue warning when @Target ElementTypes are inconsistent

mernst:
This pull request makes the Checker Framework issue an explicit warning rather than silently ignoring an annotation that is probably intended to be a type annotation.

However, I'm not sure exactly how to issue a warning for which there is no source location because no source file has started to be analyzed yet.  See the TODO comments in the code.


framework/src/org/checkerframework/framework/util/AnnotatedTypes.java
```diff
@@ -1440,16 +1444,22 @@ public static boolean hasTypeQualifierElementTypes(ElementType[] elements) {
             } else if (!element.equals(ElementType.TYPE_PARAMETER)) {
                 // if there's an ElementType with a enumerated value of something other than
                 // TYPE_USE or TYPE_PARAMETER then it isn't a valid annotation
-                hasOtherElementTypes = true;
+                otherElementType = element;
+            }
+            if (hasTypeUse && otherElementType != null) {
+                checker.report(Result.warning("inconsistent.target.metaannotation", cls.getName(), otherElementType),
+                               // TODO: what should I pass as the source (an Element or Tree) here?
```
smillst: No tree or element seems appropriate here.  Maybe we should add a version of SourceChecker#report that does not expect an element or tree?

framework/src/org/checkerframework/framework/util/AnnotatedTypes.java
```diff
@@ -1407,14 +1409,14 @@ private static boolean containsModifierImpl(AnnotatedTypeMirror type, Annotation
 
     private static Map<TypeElement, Boolean> isTypeAnnotationCache = new IdentityHashMap<>();
 
-    public static boolean isTypeAnnotation(AnnotationMirror anno) {
+    public static boolean isTypeAnnotation(AnnotationMirror anno, Class<?> cls, BaseTypeChecker checker) {
```
smillst: I could see calling this method for annotations that may not be apart of a checker.  In those cases, the error shouldn't be issued.  In general, I would not expect an method in AnnotatedTypes to issue a warning.

wmdietl:
I would suggest to make this an error instead of a warning.

A type system user has no way of handling or suppressing this warning. It will just me an annoyance.

A type system developer might ignore the warning and ship a potentially broken checker. The error is easy to deal with for the developer and prevents problems with users.

If the error is too annoying for type system developers, which I doubt, we can add an option to either only raise the error while in "development" mode or a way to suppress the error.

framework/src/org/checkerframework/framework/util/AnnotatedTypes.java
smillst: Change this to use `ErrorReporter.errorAbort(...)` to crash right away.

mernst:
Once this pull request is approved, it should be collapsed into a single commit before being merged.

framework/src/org/checkerframework/framework/util/AnnotatedTypes.java
```diff
@@ -1440,16 +1441,19 @@ public static boolean hasTypeQualifierElementTypes(ElementType[] elements) {
...
+                ErrorReporter.errorAbort("@Target meta-annotation should not contain both TYPE_USE and " + otherElementType + ", for annotation " + cls.getName());
             }
         }
 
-        return (hasTypeUse && !hasOtherElementTypes);
+        return hasTypeUse;
     }
 
-    public static boolean containsTypeAnnotation(Collection<? extends AnnotationMirror> annos) {
+    public static boolean containsTypeAnnotation(Collection<? extends AnnotationMirror> annos, Class<?> cls) {
```
smillst: This method now issue an error if any of the AnnotationMirrors in `annos` is not a type annotation.

framework/src/org/checkerframework/framework/util/AnnotatedTypes.java
```diff
@@ -1425,13 +1425,14 @@ public static boolean isTypeAnnotation(AnnotationMirror anno) {
      *
      * @param elements
      *            an array of {@link ElementType} values
+     * @param cls the annotation class being tested; used for diagnostic messages only
      * @return true if the array only has {@link ElementType#TYPE_USE} (and
      *         optionally {@link ElementType#TYPE_PARAMETER}), false if it
```
smillst: This method never returns false.  Instead it throws an exception via ErrorReport and an error is reported to the user.  Would you please document this?

smillst:
After my minor comments are addressed, this can be merged.  (Note: the Travis error could be resolved by merging master to this branch, but that isn't really necessary since continuous-integration/travis-ci/pr is passing.)

mernst:
Replaced by #748

PR #697, opened 2016-05-02
Fix Issue 696

### LINKED ISSUE #696: Array of primitive in wildcard bound
I know, that some may consider the one below as bad coding-style, still sometimes this kind is unavoidable:
```
public static void test(final @Untainted List<@Untainted ? extends @Untainted Entry<byte @Untainted [], byte @Untainted []>> input) {
    ... 
}
```
The compiler error is the following:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.3:compile (default-compile) on project ucdaa-api: Compilation failure: Compilation failure:
[ERROR] error: Unexpected type in Wildcard bound:
[ERROR] kind=BYTE
[ERROR] atm=byte
```

If you remove the `@Untainted ? extends` from before the `Entry` (which you shouldn't, according to Effective Java Item 28: Use bounded wildcards to improve API flexibility), then it compiles successfully, i.e., this works:
```
public static void test(final @Untainted List<@Untainted Entry<byte @Untainted [], byte @Untainted []>> input) {
    ... 
}
```

Full stack-trace for compilation error attached.

[checker-bytearray-extends.txt](https://github.com/typetools/checker-framework/files/244669/checker-bytearray-extends.txt)

smillst:
The component type of an array in a wildcard bound can be a primitive type. (Fixes #696)


jthaine:
Looks good to me.

framework/src/org/checkerframework/framework/type/BoundsInitializer.java
```diff
@@ -211,10 +211,13 @@ public Void visitUnion(AnnotatedUnionType type, Void aVoid) {
 
         @Override
         public Void visitArray(AnnotatedArrayType type, Void aVoid) {
-            final BoundPathNode componentNode = addPathNode(new ArrayComponentNode());
-            type.setComponentType(replaceOrVisit(type.getComponentType()));
-            removePathNode(componentNode);
-
+            if (!TypesUtils.isPrimitive(type.getComponentType().getUnderlyingType())) {
```
wmdietl: It's impossible to tell from the class description whether this is the right behavior or not.
Is the intent to add annotations to all types in the bounds? Then why wouldn't you want to apply defaults to primitives that appear as array components?
If you understand the purpose of this class, could you expand on its javadoc?

smillst: I improved the Javadoc on BoundsInitializer and BoundsInitializer.InitializerVisitor

wmdietl:
Looks good!

Issue #691, opened 2016-04-28
Format Checker: must implement GLB


dbrosoft:
Save [Issue691.zip](https://github.com/typetools/checker-framework/files/241144/Issue691.zip) and proceed as follows:
```sh
$ unzip Issue691.zip
$ cd Issue691
$ java -jar $CHECKERFRAMEWORK/checker/dist/checker.jar -processor formatter -AprintErrorStack java/util/Collections.java
```
Stack trace is provided in `error.txt`.

smillst:
Stack trace:
````
error: SourceChecker.typeProcess: unexpected Throwable (NullPointerException) while processing java/util/Collections.java
  Compilation unit: java/util/Collections.java
  Exception: java.lang.NullPointerException; Stack trace: org.checkerframework.checker.formatter.FormatterTreeUtil.formatAnnotationToCategories(FormatterTreeUtil.java:394)
  org.checkerframework.checker.formatter.FormatterAnnotatedTypeFactory$FormatterQualifierHierarchy.isSubtype(FormatterAnnotatedTypeFactory.java:120)
  org.checkerframework.framework.util.MultiGraphQualifierHierarchy.findGlb(MultiGraphQualifierHierarchy.java:782)
  org.checkerframework.framework.util.MultiGraphQualifierHierarchy.calculateGlbs(MultiGraphQualifierHierarchy.java:762)
  org.checkerframework.framework.util.MultiGraphQualifierHierarchy.greatestLowerBound(MultiGraphQualifierHierarchy.java:413)
  org.checkerframework.framework.type.QualifierHierarchy.greatestLowerBounds(QualifierHierarchy.java:210)
  org.checkerframework.framework.util.AnnotatedTypes.addAnnotationsImpl(AnnotatedTypes.java:848)
  ````

smillst:
@dbrosoft  The title of this issue was "AnnotatedTypeFactory.fromNewClassContextHelper: unexpected context (formatter)" but that's not the error I got (nor the error in error.txt).   Do you have another test case with that error?

smillst:
Because the Format Checker has annotations that take arguments, the default implementation of GLB is wrong.  (The test case for this issue should be added to all-systems, with `@SuppressWarnings("")` so that new checkers will crash if they do not implement GLB when that should.)

dbrosoft:
[...]

PR #708, opened 2016-05-10
Fix issue #691

jthaine:


checker/src/org/checkerframework/checker/formatter/FormatterTreeUtil.java
```diff
@@ -386,12 +386,17 @@ public AnnotationMirror categoriesToFormatAnnotation(ConversionCategory[] args)
 
     /**
      * Takes a syntax tree element that represents a {@link Format} annotation,
-     * and returns its value.
+     * and returns its value. Returns null (instead of an empty array) if the
+     * {@code Format} annotation's value is empty.
      */
     public ConversionCategory[] formatAnnotationToCategories(AnnotationMirror anno) {
+        AnnotationValue av = AnnotationUtils.getElementValuesWithDefaults(anno).get(formatArgTypesElement);
+        if (av == null) {
```
jthaine: This was actually causing the crash (I can add a comment if appropriate).

checker/tests/src/tests/FormatterGlbChecker.java
```diff
@@ -0,0 +1,255 @@
+package tests;
```
jthaine: This test the GLB implementation, but not the crash.

framework/tests/all-systems/Issue691.java
```diff
@@ -0,0 +1,9 @@
+// Test case for Issue 691:
+// https://github.com/typetools/checker-framework/issues/691
+
+interface MyInterface<T> {
+}
+@SuppressWarnings("")
```
jthaine: This ensures the crash does not occur.

checker/src/org/checkerframework/checker/formatter/FormatterAnnotatedTypeFactory.java
```diff
@@ -114,14 +114,18 @@ public FormatterQualifierHierarchy(MultiGraphFactory f) {
...
-                AnnotationUtils.areSameIgnoringValues(lhs, FORMAT))
-            {
+                AnnotationUtils.areSameIgnoringValues(lhs, FORMAT)) {
                 ConversionCategory[] rhsArgTypes =
                         treeUtil.formatAnnotationToCategories(rhs);
                 ConversionCategory[] lhsArgTypes =
                         treeUtil.formatAnnotationToCategories(lhs);
 
-                if (rhsArgTypes.length != lhsArgTypes.length) {
+                if (rhsArgTypes == null && lhsArgTypes == null) {
```
jthaine: I changed formatAnnotationToCategories to return null below, so these cases are necessary to handle that.

checker/src/org/checkerframework/checker/formatter/FormatterAnnotatedTypeFactory.java
```diff
@@ -147,5 +151,51 @@ public boolean isSubtype(AnnotationMirror rhs, AnnotationMirror lhs) {
...
+                        treeUtil.formatAnnotationToCategories(anno1);
+                ConversionCategory[] anno2ArgTypes =
+                        treeUtil.formatAnnotationToCategories(anno2);
+
+                if (anno1ArgTypes == null && anno2ArgTypes == null) {
+                	return FORMAT;
+                }
+
+                if (anno1ArgTypes == null || anno2ArgTypes == null) {
+                    return FORMATBOTTOM; // TODO: Should this be InvalidFormat instead?
```
jthaine: I am not sure about this and the case below. Perhaps Konstantin should also have a look as he is probably most familiar with this type system.

checker/src/org/checkerframework/checker/formatter/FormatterTreeUtil.java
```diff
@@ -386,12 +386,17 @@ public AnnotationMirror categoriesToFormatAnnotation(ConversionCategory[] args)
...
+        	return null;
```
smillst: This is just pushing the NPE elsewhere.  The format annotation should always have a value, so if it does not, there's a bug somewhere.  Instead of returning null, Error.abort with a descriptive message.

framework/tests/all-systems/Issue691.java
smillst: It would be helpful adding a commit about why `@SuppressWarnings("")` is used.  And that the code below cause GLB to be executed, which results in a crash if the default implementation isn't correct for a given checker.

checker/src/org/checkerframework/checker/formatter/FormatterAnnotatedTypeFactory.java
```diff
@@ -147,5 +151,51 @@ public boolean isSubtype(AnnotationMirror rhs, AnnotationMirror lhs) {
...
+                }
+
+
+                if (anno1ArgTypes.length != anno2ArgTypes.length) {
+                    return FORMATBOTTOM; // TODO: Should this be InvalidFormat instead? (Different situation from the if statement above)
```
smillst: Can you explain what case this is?  I think it's `GLB(@Format(GENERAL), @FORMAT({GENERAL, CHAR})) = @FormatBottom`.  Which seems to be correct given that `@Format(GENERAL)` is not a subtype of `@FORMAT({GENERAL, CHAR})`

checker/src/org/checkerframework/checker/formatter/FormatterTreeUtil.java
smillst: It's a javac error to write `@Format()` because there is not default for value.  The only way that av is null is if the FormatChecker itself created it.

checker/src/org/checkerframework/checker/formatter/FormatterAnnotatedTypeFactory.java
jthaine: Yes, that is the case. And I agree it should be `@FormatBottom`, for the reason you stated.

smillst:
I made a few small tweaks -- everything else looks good.  @jthaine if you agree with the tweaks, then merge.

jthaine:
Merged via commit 13138dd5c5fa404f07882e9d86f5d2c63f8c0224

PR #709, opened 2016-05-10
Rename TypeHierarchy#isSubtype to TypeHierarchy#isSubtypeOrConvertible 

smillst:
`TypeHierarchy#isSubtype(AnnotatedTypeMirror, AnnotatedTypeMirror)` is implemented to accept combinations of AnntotatedTypeMirrors where the underlying Java types are not subtypes, but are convertible.  For example AnnotatedDeclaredType, AnnotatedPrimitiveType, see implementation [here]( https://github.com/typetools/checker-framework/blob/isSubTypeRenaming/framework/src/org/checkerframework/framework/type/DefaultTypeHierarchy.java#L527-535). So the name should really be isSubtypeOrConvertible.

It would make the transition easier if isSubtype could be deprecated, but because TypeHierarchy is an interface, then all implementations would have to implement both isSubtype and isSubtypeOrConvertible.    And then once isSubtype is removed, the implementations would have to change again.  It's really a question of whether we think more clients call TypeHierarchy#isSubtype than implement TypeHierarchy.  I would appreciate input on this. 

Most of the changes are from automatic refactoring.  I also improved the Javadoc on [isSutypeOrConvertible](https://github.com/typetools/checker-framework/pull/709/files#diff-97ff8d0c334410fa4bacbb5a77782f5b), which should be reviewed more carefully than the other changes.

### DIFF
--- a/checker/src/org/checkerframework/checker/fenum/FenumVisitor.java
+++ b/checker/src/org/checkerframework/checker/fenum/FenumVisitor.java
@@ -31,8 +31,8 @@ public class FenumVisitor extends BaseTypeVisitor<FenumAnnotatedTypeFactory> {
 
             AnnotatedTypeMirror lhs = atypeFactory.getAnnotatedType(node.getLeftOperand());
             AnnotatedTypeMirror rhs = atypeFactory.getAnnotatedType(node.getRightOperand());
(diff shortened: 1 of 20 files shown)


smillst:
Instead of renaming the method, I'll document which conversion are performed by isSubtype and how they are preformed

PR #710, opened 2016-05-10
Add Intersection_Primitive and Typevar_Primitive combinations to isSubtype

smillst:
Example of when these combinations are used:
````
public class TypeVarPrimitives {
    <T extends Long> void method(T tLong) {
        long l = tLong;
    }
    <T extends Long & Cloneable> void methodIntersection(T tLong) {
        long l = tLong;
    }
````
(This covers 2/5 cases in issue #701)

### DIFF
--- a/checker/tests/nullness/TypeVarPrimitives.java
+++ b/checker/tests/nullness/TypeVarPrimitives.java
@@ -1,0 +1,26 @@
+// Unannotated version in framework/tests/all-systems/Primitives.java
+import org.checkerframework.checker.nullness.qual.*;
+public class TypeVarPrimitives {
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: checker/tests/nullness/TypeVarPrimitives.java [hunk 1]
```diff
@@ -0,0 +1,26 @@
+// Unannotated version in framework/tests/all-systems/Primitives.java
```
wmdietl: Should this be `framework/tests/all-systems/TypeVarPrimitives.java`? At least that's the file being added below.

wmdietl:
Looks good.

PR #719, opened 2016-05-12
Add docstrings to and remove dead code from the release scripts

jthaine:


release/release_build.py
```diff
@@ -59,15 +59,19 @@ def clone_or_update_repos(auto):
...
 
 def get_afu_date(building_afu):
+    """If the AFU is being built, return the current date, otherwise return the date of the last AFU release as indicated in the AFU home page."""
     if building_afu:
         return get_current_date()
     else:
         afu_site = os.path.join(HTTP_PATH_TO_LIVE_SITE, "annotation-file-utilities")
+        # TODO: these tags no longer exist in the AFU home page. Fix this to
+        # extract the date from the afu-version tags or remove this code altogether.
         return extract_from_site(afu_site, "<!-- afu-date -->", "<!-- /afu-date -->")
```
mernst: If this line doesn't work, should it be replaced by throwing an exception?

release/release_build.py
```diff
@@ -255,6 +266,8 @@ def build_checker_framework_release(version, afu_version, afu_release_date, chec
     return
 
 def commit_to_interm_projects(jsr308_version, afu_version, projects_to_release):
+    """Commit the changes for each project from its build repo to its corresponding intermediate repo"
+    "in preparation for running the release_push script, which does not read the build repos."""
```
mernst: The single quotes at the end of the previous line and the beginning of this one look extraneous.  Python doesn't require them within a triple-quoted string.

It's good to break the lines, and all other very long lines (> 100 or >80 characters) that were added in this pull request should also be broken for readability, both in the GitHub pull request and in the final code.

Issue #688, opened 2016-04-28
I18n Checker: must implement GLB


dbrosoft:
Save [Issue685.zip](https://github.com/typetools/checker-framework/files/239674/Issue685.zip) and proceed as follows:
```sh
$ unzip Issue685.zip
$ cd Issue685
$ java -jar $CHECKERFRAMEWORK/checker/dist/checker.jar -processor i18n -AprintErrorStack java/lang/invoke/MemberName.java
```
Stack trace is provided in `error.txt`.  Source for classes in `sun` is in the [Annotated JDK](https://bitbucket.org/typetools/annotated-jdk8u-jdk).

dbrosoft:
(Now why did the number skip ahead and make my choice of zipfile name look silly?)

mernst:
Because issues and pulls share a single index.  There are pull requests 686 and 687.

smillst:
Because the I18N Checker has annotations that take arguments, the default implementation of GLB is wrong.

PR #726, opened 2016-05-13
Fix issue #688

jthaine:


smillst:
Wow, what a simple error!

Issue #689, opened 2016-04-28
Assertion failure in GuiEffectVisitor.visitMethodInvocation()


dbrosoft:
Save [Issue689.zip](https://github.com/typetools/checker-framework/files/239799/Issue689.zip) and proceed as follows:
```sh
$ unzip Issue689.zip
$ cd Issue689
$ java -jar $CHECKERFRAMEWORK/checker/dist/checker.jar -processor guieffect -AprintErrorStack java/util/AbstractList.java
```
It appears (based on experiments commenting out method code) that the problematic method in `AbstractList.java` is `listIterator()` in class `SubList`.  The stack trace of the assertion failure is provided in `error.txt`.

csgordon:
It looks like ```TreeUtils.enclosingMethod(getCurrentPath())``` may be misbehaving when called in the middle of a visit to a field initializer.

The specific code the checker is tripping on is in the anonymous inner class inside SubList that implements its iterator, starting at line 715 of the file included in the report:
```java
        return new ListIterator<E>() {
            private final ListIterator<E> i = l.listIterator(index+offset);
...
```
The compiler of synthesizes a constructor for the anonymous class, which is visited by the ```GuiEffectVisitor```.  On entry to that method, the ```MethodTree``` argument to ```visitMethod``` looks as expected, we push it on a stack of method trees, and visit the body.

When we get to the call in that field initializer, we look up the enclosing MethodTree via ```TreeUtils.enclosingMethod(getCurrentPath())```.  I'm not sure what happens, but when I print the resulting MethodTree, I see instead the tree for the enclosing **anonymous class**, not the method.  I would expect the call above to produce the enclosing synthesized method... @wmdietl?

But that's not the whole story.  I mentioned that I push the method tree onto a stack when we enter it; when we get to the initializer, we've pushed an extra null on the stack...

Back at this in a bit...

mernst:
Closed by PR #729

PR #729, opened 2016-05-15
Candidate for #689

csgordon:
This is a candidate fix for #689.  It appears the assertion was too strong, because apparently none of the work on the ECOOP paper encountered anonymous inner classes with non-trivial (i.e., method-invoking) field initializers, which don't get visited in the context of an enclosing method structure.

A test for this is necessary as well before I close the issue, but I want this to go through CI to maybe unstick things for Monday.

### DIFF
--- a/checker/src/org/checkerframework/checker/guieffect/GuiEffectVisitor.java
+++ b/checker/src/org/checkerframework/checker/guieffect/GuiEffectVisitor.java
@@ -156,7 +156,9 @@ public class GuiEffectVisitor extends BaseTypeVisitor<GuiEffectTypeFactory> {
         }
 
         Effect callerEffect = atypeFactory.getDeclaredEffect(callerElt);
-        assert (callerEffect.equals(effStack.peek()));
+        // Field initializers inside anonymous inner classes show up with a null current-method ---
+        // the traversal goes straight from the class to the initializer.
+        assert (currentMethods.peek() == null || callerEffect.equals(effStack.peek()));
 
         if (!Effect.LE(targetEffect, callerEffect)) {
             checker.report(Result.failure("call.invalid.ui", targetEffect, callerEffect), node);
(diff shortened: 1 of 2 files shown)


checker/src/org/checkerframework/checker/guieffect/GuiEffectVisitor.java
```diff
@@ -156,7 +156,9 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
         }
 
         Effect callerEffect = atypeFactory.getDeclaredEffect(callerElt);
-        assert (callerEffect.equals(effStack.peek()));
+        // Field initializers inside anonymous inner classes show up with a null current-method ---
+        // the traversal goes straight from the class to the initializer.
+        assert (currentMethods.peek() == null || callerEffect.equals(effStack.peek()));
```
wmdietl: Would `currentMethods.empty()` be simpler for the first condition? I hope we don't expect a null value on the stack.

wmdietl:
Is there a minimal test case that you can also add?
I'll look at your comment in the issue tomorrow.
Thansk!

checker/src/org/checkerframework/checker/guieffect/GuiEffectVisitor.java
csgordon: We do; ```currentMethods``` is a stack of ```MethodTree```s, where ```null``` is used as a sentinel to indicate that there is no surrounding method context.  Previously I thought this only applied to static initializers (which is why the sentinel exists in the first place), but apparently instance initializers also show up in this context.

If there's a more idiomatic way to handle the sentinel (some kind of non-null dummy method tree) I'm open to a bit of refactoring.

csgordon:
There is, I just haven't had a chance to sit down and work through the plumbing yet.

csgordon:
@wmdietl there is now a test

Issue #724, opened 2016-05-13
AnnotatedDeclaredType#getEnclosingType returns ATM with missing annotations


smillst:

smillst:
QualifiedDeclaredType needs to keep track of the enclosing type like AnnotatedDeclaredType.

smillst:
See test case in framework/all-systems/Issue577.java

smillst:
I think this is the root cause of issues #437 and #134.

wmdietlGC:
From #1099: #437 and #134 are fixed. Some mentions of #724 also work now.
A comment in AnnotatedTypes.asOuterSuper mentions #724 and is about nested classes; that hack should also be fixed. This issue stays open to fix this.

PR #730, opened 2016-05-16
asMemberOf should substitute type variables declared in outer classes too

### LINKED ISSUE #577: Comparing type variable to int[] fails
The framework occasionally (thrice in the merged annotated JDK) throws an exception with a message about "Incomparable types".  For instance, `javac -processor nullness -AprintErrorStack `[java/util/stream/IncomparableTypes.java](https://gist.github.com/dbrosoft/25391fc28ad992be0d1c) twice prints
```
error: Incomparable types ( T_ARR extends Object, int [])visitHistory = VisitHistory(  )
```
followed by a stack trace.

smillst:
Fixes #577

AnnotatedDeclaredType#getEnclosingType is missing annotations and the qualifier framework does not implement QualifiedDeclaredType#getEnclosingType (see #724).   Because of this, this pull request causes the Regex and Tainting Checker to issue (new) false positive warnings.

### DIFF
--- a/checker/tests/nullness/Issue577.java
+++ b/checker/tests/nullness/Issue577.java
@@ -1,0 +1,82 @@
+// Test case for issue #577 with expected errors:
+// Also see test case in framework/tests/all-systems/Issue577.java
+// https://github.com/typetools/checker-framework/issues/577
+
+import org.checkerframework.checker.nullness.qual.*;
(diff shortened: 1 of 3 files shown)


smillst:
I'll open another pull request for others to review this code change.

PR #731, opened 2016-05-16
asMemberOf should substitute all type variables declared in outer class, too.

### LINKED ISSUE #577: Comparing type variable to int[] fails

smillst:
Fixes #577

AnnotatedDeclaredType#getEnclosingType is missing annotations and the qualifier framework does not implement QualifiedDeclaredType#getEnclosingType (see #724).   Because of this, this pull request causes the Regex and Tainting Checker to issue (new) false positive warnings.  

(This code has been merged (#730), but not reviewed.)

### DIFF
--- a/checker/tests/nullness/Issue577.java
+++ b/checker/tests/nullness/Issue577.java
@@ -1,0 +1,82 @@
+// Test case for issue #577 with expected errors:
+// Also see test case in framework/tests/all-systems/Issue577.java
+// https://github.com/typetools/checker-framework/issues/577
+
+import org.checkerframework.checker.nullness.qual.*;
+
+class Banana<T extends Number> extends Apple<int[]> {
+    @Override
+    void fooOuter(int[] array) {}
+    class InnerBanana extends InnerApple<long[]> {
+        @Override
+        //:: error: (override.param.invalid)
+        <F2 extends Object> void foo(int[] array, long[] array2, F2 param3) {
+        }
+    }
+}
+
+class Apple<T> {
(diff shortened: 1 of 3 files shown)


checker/tests/nullness/Issue577.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+}
+
+class IntersectionAsMemberOf {
+    interface MyGenericInterface<F> {
+        F getF();
+    }
+
+    <T extends Object & MyGenericInterface<@NonNull String>> void foo(T param) {
+      //TODO: this error is due to a bug in asSuper
```
wmdietl: I'm a bit weary of TODOs in test files. Would it make sense to extract these into separate test files that are skipped? That is a bit more explicit about work still being required.

smillst: No, it makes more sense to keep the as member of test together.  I've removed this todo in my asSuper branch and will search for references to 717 before closing the issue.

wmdietl:
I just had a quick look. As it's already merged and working, this should be fine :-)

PR #737, opened 2016-05-20
Fix underlying type of anonymous classes

smillst:
The underlying type of anonymous classes was the type of it's super class.  For example, the underlying type of `new Object(){}` was Object, but it should be `<OuterClass$1>`.  (daikon and plume-lib type check with this change.)


framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1934,23 +1934,23 @@ private boolean isSyntheticArgument(Tree tree) {
...
-            // If newClassTree does not create anonymous class,
-            // newClassTree.getIdentifier includes the explicit annotations in this location:
-            //   new @HERE Class()
-            type = (AnnotatedDeclaredType) fromTypeTree(newClassTree.getIdentifier());
-        } else {
-            type = (AnnotatedDeclaredType) toAnnotatedType(InternalUtils.typeOf(newClassTree), false);
-        }
 
         if (newClassTree.getClassBody() != null) {
+            type = (AnnotatedDeclaredType) toAnnotatedType(InternalUtils.typeOf(newClassTree), false);
```
wmdietl: Can an anonymous class use diamonds?
I don't quite understand the three cases and why diamonds are handled again on line 1956.

wmdietl:
Is it possible to write test cases for this?

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
smillst: > Can an anonymous class use diamonds?

No
>I don't quite understand the three cases and why diamonds are handled again on line 1956.

I refactored the code to clarify this.

smillst:
@wmdietl I would like to rebase this pull request into one commit before merging.

checker/src/org/checkerframework/checker/fenum/FenumVisitor.java
```diff
@@ -58,7 +59,7 @@ public Void visitSwitch(SwitchTree node, Void p) {
 
     @Override
     protected boolean checkConstructorInvocation(AnnotatedDeclaredType dt,
```
wmdietl: Most of the other `checkXXX` methods have a void return type. Is anyone actually using the boolean returned by this function?

wmdietl:
Looks good.

smillst:
merged in ae78785

PR #739, opened 2016-05-25
Release scripts fixes

jthaine:
Clean the repos on every release_build run unless told otherwise by the user, and
move files from the zipfilesets tag in release.xml to the includes/excludes files.


mernst:
Why is file CFLogo redundantly added to the repository rather than being copied from checker/manual/logo/Logo/CFLogo.png when needed?

jthaine:
I thought this was the way we agreed to remove references to CFLogo.png from the zipfilesets - I misunderstood then. If we keep only one copy of CFLogo.png in the repository, then I think we should continue to use zipfileset entries as the mechanism to copy CFLogo.png into multiple locations, since it does not require making a physical copies of the file that then need to be cleaned up.
I restored the CFLogo.png zipfileset entries and removed the duplicate copies of CFLogo.png. I can squash these commits into one if they look good to you, so as not to pollute the history.

mernst:
Looks good!  Feel free to squash the history and then merge.

jthaine:
Merged via commit ec77a9f15035c5672f8abb5b5c8bf2f8d6289ddd

Issue #738, opened 2016-05-20
ClassCastException: AnnotatedIntersectionType to AnnotatedArrayType


Kiena:
This may be a consequence, thus duplicate, of one or more already reported type bound related bugs, though I couldn't find one which would seem to match this exact case.

Steps to reproduce:

[cfcrash.zip](https://github.com/typetools/checker-framework/files/275524/cfcrash.zip)

1. The attached file contains a simple Maven project, with a single source file to demonstrate the issue.
- (Maven) In the `cfcrash` directory execute the command `mvn install`  
Expected: Successful build.  
Actual:  
`Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.5.1:compile (default-compile) on project cfcrash: Compilation failure  
error: SourceChecker.typeProcess: unexpected Throwable (ClassCastException) while processing P:\java\workspaces\test\cfcrash\src\main\java\test\cfcrash\Test.java; message: org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedIntersectionType cannot be cast to org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedArrayType`
- (Command line only) In the `cfcrash` directory execute the command (artifacts are reused from the repository):  
`javac -version -verbose -d .\target\classes -classpath ".\target\classes;p:\.m2\repository\org\checkerframework\checker\2.0.0\checker-2.0.0.jar;p:\.m2\repository\org\checkerframework\checker-qual\2.0.0\checker-qual-2.0.0.jar;p:\.m2\repository\org\checkerframework\jdk8\2.0.0\jdk8-2.0.0.jar;" -sourcepath ".\src\main\java;.\target\generated-sources\annotations;" .\src\main\java\test\cfcrash\Test.java -s .\target\generated-sources\annotations -processor org.checkerframework.checker.nullness.NullnessChecker -g -nowarn -target 1.8 -source 1.8 -encoding UTF-8 -AcheckPurityAnnotations -Alint=all -AprintAllQualifiers -AprintErrorStack -AshowSuppressWarningKeys -Awarns -Xbootclasspath/p:p:\.m2\repository\org\checkerframework\jdk8\2.0.0\jdk8-2.0.0.jar`  
Expected: No compilation error. (As happens without the processor.)  
Actual: See `cfcrash/error.txt` within the attachment for the complete output.

smillst:
[...]

Issue #701, opened 2016-05-04
Missing combination for isSubtype.


smillst:
`DefaultTypeHierarchy#isSubtype(ATM, ATM)`, (which is really isSubtypeOrConvertible) is not implemented for the following combinations (subtype_supertype):
- [x] Intersection_Primitive (`long l = t`, where the type of `t` is `T extends Long`) PR #710
- [x] Typevar_Primitive (`long l = t`, where the type of `t` is `T extends Long & Cloneable`) PR #710
- [x] Union_Intersection (When checking type argument is within bounds of type parameter.) PR ##727
- [x] Union_Union (When checking method arguments type argument substitutions) PR ##727
- [x] Array_Intersection (Testcase #738) Requires new asSuper #717 (PR #781)  

~~- [ ] Intersection_Typevar~~ (testcase in framework/tests/all-systems/Issue717.java. asSuper bug see #717)
~~- [ ] Typevar_Array (See #577)~~ Combo not valid. Error happens when a type variable is not substituted correctly

PR #745, opened 2016-05-26
Various fixes in QualifierPolymorphism

smillst:
Fixes the crash reported in #738, but the Issue738.java test case has another crash. (See #701)

### DIFF
--- a/framework/src/org/checkerframework/framework/util/QualifierPolymorphism.java
+++ b/framework/src/org/checkerframework/framework/util/QualifierPolymorphism.java
@@ -418,6 +418,10 @@ public class QualifierPolymorphism {
             assert actualType.getKind() == type.getKind();
             type = (AnnotatedDeclaredType) AnnotatedTypes.asSuper(types, atypeFactory, type, actualType);
 
+            if (type == null) {
+                return Collections.emptyMap();
+            }
+
             AnnotatedDeclaredType dcType = (AnnotatedDeclaredType)actualType;
 
             Map<AnnotationMirror, Set<? extends AnnotationMirror>> result =
--- a/framework/src/org/checkerframework/framework/util/QualifierPolymorphism.java
+++ b/framework/src/org/checkerframework/framework/util/QualifierPolymorphism.java
@@ -197,7 +197,7 @@ public class QualifierPolymorphism {
         // for super() and this() method calls, getReceiverType(tree) does not return the correct
         // type. So, just skip those.  This is consistent with skipping receivers of constructors
         // below.
-        if (type.getReceiverType() != null && !(TreeUtils.isSuperCall(tree))
+        if (type.getReceiverType() != null && !TreeUtils.isSuperCall(tree)
                 && !TreeUtils.isThisCall(tree)) {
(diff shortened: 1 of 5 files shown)


framework/src/org/checkerframework/framework/util/QualifierPolymorphism.java
```diff
@@ -192,7 +194,11 @@ public void annotate(MethodInvocationTree tree, AnnotatedExecutableType type) {
 
         Map<AnnotationMirror, Set<? extends AnnotationMirror>> matchingMapping = collector.visit(arguments, requiredArgs);
 
-        if (type.getReceiverType() != null) {
+        // for super() and this() method calls, getReceiverType(tree) does not return the correct
+        // type. So, just skip those.  This is consistent with skipping receivers of constructors
+        // below.
+        if (type.getReceiverType() != null && !(TreeUtils.isSuperCall(tree))
```
wmdietl: Why the parens around isSuperCall?

framework/src/org/checkerframework/framework/util/QualifierPolymorphism.java
```diff
@@ -231,6 +237,13 @@ public void annotate(AnnotatedExecutableType functionalInterface, AnnotatedExecu
 
         List<AnnotatedTypeMirror> args = functionalInterface.getParameterTypes();
         List<AnnotatedTypeMirror> requiredArgs = memberReference.getParameterTypes();
+        if (args.size() == requiredArgs.size() + 1) {
+            // If the member reference is a reference to an instance method of an arbitrary
+            // object, then first parameter of the functional interface corresponds to the
+            // receiver of the member reference.
+            requiredArgs = new ArrayList<>(requiredArgs);
```
wmdietl: Isn't this a rather inefficient way of doing this? This first copies requiredArgs into the new array, and then has to move everything over to add a first element.

framework/src/org/checkerframework/framework/util/QualifierPolymorphism.java
```diff
@@ -392,8 +417,7 @@ protected Void scan(AnnotatedTypeMirror type, Void p) {
 
             assert actualType.getKind() == type.getKind();
             type = (AnnotatedDeclaredType) AnnotatedTypes.asSuper(types, atypeFactory, type, actualType);
-            // TODO: type is null if type is an intersection type
-            // assert type != null;
+
             if (type == null) {
```
wmdietl: Now that intersection types are already explicitly handled at the beginning of this method, can this be strengthened to an errorAbort if the type is null?

wmdietl:
Just a few small comments.

framework/src/org/checkerframework/framework/util/QualifierPolymorphism.java
smillst: Not yet.  asSuper returns null erroneously in other cases.  In the new asSuper, it is an ErrorAbort if asSuper returns null and therefore this (and many other) null checks will be removed.  The new asSuper ErrorAborts is how I found this batch of bugs.

smillst:
Merged via 47470dc

Issue #700, opened 2016-05-04
Too many classes on the bootclasspath


wmdietl:
[This issue is motivated by a discussion here: https://github.com/typetools/annotation-tools/commit/6ded22f83487938efe4709d95c807e8288ed86d1]

Take this code:

```
import com.sun.tools.javac.Main;
import org.checkerframework.checker.nullness.NullnessChecker;
class XXX {}
```

It produces two errors with standard javac (with nothing on your classpath):

```
XXX.java:1: error: package com.sun.tools.javac does not exist
import com.sun.tools.javac.Main;
                          ^
XXX.java:2: error: package org.checkerframework.checker.nullness does not exist
import org.checkerframework.checker.nullness.NullnessChecker;
                                            ^
2 errors
```

That is, the compiler `com.sun.tools.javac.Main` doesn't put itself on the bootclasspath when compiling application code.

When using the Checker Framework javac, this example compiles without errors, because javac, the Checker Framework, and many other classes are put on the bootclasspath.

Using checker/bin/javac should give the same errors as normal javac.

This should not impact the use of `org.checkerframework.checker.nullness.NullnessChecker` as a `javac -processor` argument. That is, the processorpath should include the Checker Framework.

The only point for discussion is whether the qualifiers should be automatically included or not.
That is, when using the Checker Framework javac without extra flags, should

```
import org.checkerframework.checker.nullness.qual.Nullable;
class XXX {}
```

compile with or without errors?

Raising an error would remind the user that the resulting classfile references the annotation (though no harm occurs if the annotation is not available at run time: see http://stackoverflow.com/questions/3567413/why-doesnt-a-missing-annotation-cause-a-classnotfoundexception-at-runtime).
If they have to put the checker-qual.jar file on the classpath for compilation, that will be more obvious.

Not raising an error would make it easier to use the examples, but obscure the dependency.

It might be useful to split checker-qual.jar into two files: one that only contains the annotations and another that contains the runtime utility classes.
Depending only on the annotations will usually not cause the classfile to fail loading.
We could then implicitly include the qualifier-only jar file, but require explicit inclusion of the runtime utility classes.

wmdietl:
This is relevant to the discussion we had about annotations being a dependency:
http://stackoverflow.com/questions/3567413/why-doesnt-a-missing-annotation-cause-a-classnotfoundexception-at-runtime

wmdietl:
Now that this is merged, should we remove afuplume.jar from annotation-tools?

[...]

PR #747, opened 2016-05-26
Processorpath

mernst:
This pull request is intended to correct issue #700, "Too many classes on the bootclasspath".

### DIFF
--- a/framework/src/org/checkerframework/framework/util/CheckerMain.java
+++ b/framework/src/org/checkerframework/framework/util/CheckerMain.java
@@ -282,16 +282,13 @@ public class CheckerMain {
 
         String path = null;
 
-        int i = 0;
-        while (i < args.size()) {
-
-            if (args.get(i).equals("-cp") || args.get(i).equals("-classpath")) {
-                if (args.size() > i) {
-                    args.remove(i);
-                    path = args.remove(i);
-                } // else loop ends and we have a dangling -cp.  javac will issue a warning.
-            } else {
-                i++;
+        for (int i=0; i<args.size(); i++) {
+            if ((args.get(i).equals("-cp") || args.get(i).equals("-classpath"))
+                && (i + 1 < args.size())) {
+                args.remove(i);
+                path = args.remove(i);
+                // re-process whatever is currently at element i
+                i--;
             }
         }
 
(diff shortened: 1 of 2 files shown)


framework/src/org/checkerframework/framework/util/CheckerMain.java
```diff
@@ -287,6 +309,38 @@ protected static File extractFileArg(final String argumentName, final File alter
...
+    protected static List<String> extractPpOpts(final List<String> args) {
+        List<String> actualArgs = new ArrayList<String>();
+
+        String path = null;
+
+        int i = 0;
+        while (i < args.size()) {
+
+            if (args.get(i).equals("-processorpath")) {
+                if (args.size() > i) {
```
smillst: Should be `args.size() > i + 1`, right?  Would also be easier to spot the difference from the loop predicate  if it was `i + 1 < arg.size()`

framework/src/org/checkerframework/framework/util/CheckerMain.java
```diff
@@ -123,7 +135,17 @@ public void addToRuntimeBootclasspath(List<String> runtimeBootClasspathOpts) {
...
-        extractedOps.add(0, this.checkersJar.getAbsolutePath());
+        extractedOps.add(0, this.checkerQualJar.getAbsolutePath());
+        return extractedOps;
+    }
+
+    // Assumes that createCpOpts has already been run.
+    protected List<String> createPpOpts(final List<String> argsList) {
+        final List<String> extractedOps = extractPpOpts(argsList);
+        if (extractedOps.isEmpty()) {
+            extractedOps.addAll(this.cpOpts);
```
smillst: I don't understand why the classpath is added to the processor path if no explicit processor path is found.  (Or maybe that's not what's happening here?)

mernst: That's what is happening.  I added a comment explaining why.

framework/src/org/checkerframework/framework/util/CheckerMain.java
mernst: You are right.  I copied that from the extractCpOpts routine, which contained this bug.  I will fix both routines.  Thanks!

mernst:
Replaced by #761

Issue #391, opened 2015-07-03
@RequiresNonNull and field visibilty


GoogleCodeExporter:
```
Hi,

from what I gather @RequiresNonNull should be enforced by the nullness checker 
but I can't get it to work.

I have this (condensed) test case that produces no errors for these classes, 
only warnings about missing documentation:

class ClassA {
    @Nullable private String value = null;

    @RequiresNonNull("value") public String getValue() {
        return value;
    }
}

public class Tester {
    public static void main(String[] args) {
        ClassA a = new ClassA();
        a.getValue();
    }
}

No error is reported even with
@MonotonicNonNull private String value;
which is what I had in the beginning, but then I tried and noticed that it 
doesn't work even with the code above that explicitly sets the field to null.

I'm using version 1.8.9 with NetBeans 8.0.2 x64, Ant 1.9.4 and the javac 
included in JDK 1.8.0_u25 x64 on Windows 8.
I pass these options to the compiler:
-version -Xlint:all -Xdoclint:all

My ant target is attached.

The nullness checker seems to work fine otherwise, I can see various errors for 
other classes that seem sensible and I already have added some annotations to 
eliminate errors.
So I think that the issue, if there's any, is related to the particular 
annotations used in the sample above.
```

Original issue reported on code.google.com by `karlsand...@gmail.com` on 11 Jan 2015 at 9:22

Attachments:
- [Ant Task.xml](https://storage.googleapis.com/google-code-attachments/checker-framework/issue-391/comment-0/Ant Task.xml)

jthaine:
This issue has been fixed for private fields (but not yet for private methods) via commit 85e85ab1719ebfc0b3fa558ac7e08dd0f07f5b9e

mernst:
Because this is fixed for private fields, I am closing the issue.  If private methods become important, please file a new issue.

PR #750, opened 2016-05-26
Fix issue #391 for retrieval of fields only

jthaine:
This is ready for review, but please don't merge yet - I need to push a fix to Daikon first.

### DIFF
--- a/checker/tests/nullness/Issue391.java
+++ b/checker/tests/nullness/Issue391.java
@@ -7,11 +7,15 @@ import org.checkerframework.checker.nullness.qual.EnsuresNonNull;
 
 
 class ClassA {
-    @Nullable private String value = null;
+    private @Nullable String value = null;
 
-    @EnsuresNonNull("value") public void ensuresNonNull() {}
+    @EnsuresNonNull("value")
+    public void ensuresNonNull() {
+        value = "";
+    }
 
-    @RequiresNonNull("value") public String getValue() {
+    @RequiresNonNull("value")
+    public String getValue() {
         return value;
     }
 }
@@ -25,6 +29,7 @@ public class Issue391 {
 
     @EnsuresNonNull("field.value")
     void ensuresNonNull() {
+        field.ensuresNonNull();
     }
 
     void method2() {
@@ -37,14 +42,10 @@ public class Issue391 {
 
     void method3() {
         ensuresNonNull();
-        // TODO: the above method call should satisfy the condtion below
-        //:: error: (contracts.precondition.not.satisfied)
         method();
 
         ClassA a = new ClassA();
         a.ensuresNonNull();
-        // TODO: the above method call should satisfy the condtion below
-        //:: error: (contracts.precondition.not.satisfied)
         a.getValue();
     }
 }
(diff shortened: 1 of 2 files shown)


smillst:
Looks good to me. Merge when ready.

checker/tests/nullness/Issue391.java
```diff
@@ -0,0 +1,50 @@
...
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.checkerframework.checker.nullness.qual.RequiresNonNull;
+import org.checkerframework.checker.nullness.qual.EnsuresNonNull;
+
+
+class ClassA {
+    @Nullable private String value = null;
+
+    @EnsuresNonNull("value") public void ensuresNonNull() {}
```
wmdietl: I would put declaration annotations on the method on a separate line.

checker/tests/nullness/Issue391.java
```diff
@@ -0,0 +1,50 @@
+// Test case for Issue 391:
+// https://github.com/typetools/checker-framework/issues/391
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.checkerframework.checker.nullness.qual.RequiresNonNull;
+import org.checkerframework.checker.nullness.qual.EnsuresNonNull;
+
+
+class ClassA {
+    @Nullable private String value = null;
```
wmdietl: The type annotation on the return type should be directly in front of the return type.

checker/tests/nullness/Issue391.java
```diff
@@ -0,0 +1,50 @@
...
+        ClassA a = new ClassA();
+        //:: error: (contracts.precondition.not.satisfied)
+        a.getValue();
+        //:: error: (contracts.precondition.not.satisfied)
+        method();
+    }
+
+    void method3() {
+        ensuresNonNull();
+        // TODO: the above method call should satisfy the condtion below
```
wmdietl: For whom are the TODO comments? Is there a different issue for them? The test is not skipped, so should this be in a new test file that doesn't expect an unwanted error?

javacutil/src/org/checkerframework/javacutil/Resolver.java
```diff
@@ -81,6 +85,22 @@ public Resolver(ProcessingEnvironment env) {
             err.initCause(e);
             throw err;
         }
+
+        try {
+            ACCESSERROR = Class.forName("com.sun.tools.javac.comp.Resolve$AccessError");
+            ACCESSERROR_ACCESS = ACCESSERROR.getMethod("access", Name.class, TypeSymbol.class);
+            ACCESSERROR_ACCESS.setAccessible(true);
+        } catch (ClassNotFoundException e) {
+            Error err = new AssertionError(
```
wmdietl: Instead of re-throwing the error, the error message would be cleaner if you used `ErrorReporter.errorAbort`. This is an expected error, so you shouldn't use exceptions more than necessary.

javacutil/src/org/checkerframework/javacutil/Resolver.java
```diff
@@ -264,8 +287,12 @@ private Object getField(Object receiver, String fieldName) throws NoSuchFieldExc
     }
 
     private Symbol wrapInvocation(Method method, Object... args) {
```
wmdietl: The method name should be changed: this version wraps an invocation on resolve, which is different from the new `wrapInvocation`, which takes a target.

Issue #657, opened 2016-04-05
FlowExpressionParseException is caught and the error is not reported to the user (umbrella issue)


jthaine:
In several places in the framework and in checkers, FlowExpressionParseException are ignored, including in BaseTypeVisitor.checkPreconditions. In some cases this results in preconditions not being checked when they should be, for example. The assumption is that the error will be reported elsewhere, such as the declaration site of a variable, instead of a use of the variable. This assumption does not always hold.

Go over all instances in the framework and in checkers where a FlowExpressionParseException is caught, and for each case reconsider whether to report the error to the user.

jthaine:
Fix the immediate issue, suppress warnings for the tests that are failing, and file separate issues for those suppressed warnings.

jthaine:
Based on the warnings issued by Daikon and plume-lib type check (see the pull requests https://github.com/codespecs/daikon/pull/57/ and https://github.com/mernst/plume-lib/pull/16), the following issues need to be fixed before the fix for this issue is pushed:

- [ ] Issue #752 ( Flow Expression parser: add support for package names )
- [ ] Issue #764 ( Flow Expression parser: cannot parse expression referring to static field of outer class )
- [ ] Issue #767 ( Compilation order changes type checking involving preconditions )

Here are the numbers of warnings that were suppressed for these issues:

-18 warnings were suppressed in Daikon for errors due to one or both of #767 and #752.
-4 warnings were suppressed in Daikon, and 1 in plume-lib, for errors due to issue #767 alone.
-5 warnings were suppressed in Daikon for errors due to issue #752 alone.
-1 warning was suppressed in Daikon for an error due to issue #764 alone.

PR #754, opened 2016-05-27
(Blocked on other issues) Surface, rather than silently ignoring, errors in flow expression parsing.  Fixes issue #657

jthaine:
Surfacing errors fixes issue #657.

Please review but do not merge yet as there are numerous warnings in Daikon that have not been suppressed yet; surfacing them will break the plume-lib-typecheck and daikon-typecheck jobs.

Merge when all checkboxes in issue #657 are checked (that is, blocking issues are fixed) and when `@SuppressWarnings` have been removed from plume-lib and daikon pull requests that are related to those issues.


checker/src/org/checkerframework/checker/experimental/regex_qual/RegexQualifiedTransfer.java
```diff
@@ -82,6 +82,8 @@ public RegexQualifiedTransfer(QualAnalysis<Regex> analysis) {
                         thenStore.insertValue(firstParam, new RegexVal(0));
                     }
                 } catch (FlowExpressionParseException e) {
+                    // report errors here
+                    // checker.report(e.getResult(), n);
```
smillst: Should this be uncommented?

checker/src/org/checkerframework/checker/regex/RegexQualifiedTransfer.java
```diff
@@ -109,6 +109,8 @@ public RegexQualifiedTransfer(QualAnalysis<QualParams<Regex>> analysis) {
                 firstParam = FlowExpressionParseUtil.parse("#1", context,
                         analysis.getContext().getTypeFactory().getPath(n.getTree()));
             } catch (FlowExpressionParseException e) {
+                // report errors here
+                // checker.report(e.getResult(), n);
```
smillst: Should this be uncommented?

checker/src/org/checkerframework/checker/regex/classic/RegexTransfer.java
```diff
@@ -89,6 +90,8 @@ public RegexTransfer(RegexAnalysis analysis) {
                         thenStore.insertValue(firstParam, regexAnnotation);
                     }
                 } catch (FlowExpressionParseException e) {
+                    // report errors here
+                    ((RegexChecker) (analysis.getTypeFactory().getContext())).report(e.getResult(), n);
                     assert false;
```
smillst: Why still assert false?

checker/src/org/checkerframework/checker/i18nformatter/I18nFormatterTreeUtil.java
```diff
@@ -274,7 +274,8 @@ private void initialCheck(List<? extends ExpressionTree> theargs, ExecutableElem
                                 .parse(formatforArg, flowExprContext, atypeFactory.getPath(tree));
                             paramIndex = flowExprContext.arguments.indexOf(paramArg);
                         } catch (FlowExpressionParseException e) {
-                            // errors are reported at declaration site
+                            // report errors here
+                            checker.report(e.getResult(), tree);
```
smillst: Shouldn't this be an i18nformat.invalid.formatfor error instead of flowexpr.parse.error?

checker/src/org/checkerframework/checker/i18nformatter/I18nFormatterTransfer.java
```diff
@@ -63,7 +63,8 @@ public I18nFormatterTransfer(I18nFormatterAnalysis analysis, I18nFormatterChecke
                     thenStore.insertValue(firstParam, anno);
                 }
             } catch (FlowExpressionParseException e) {
-                // errors are reported at declaration site
+                // report errors here
+                checker.report(e.getResult(), node);
```
smillst: Shouldn't this be an i18nformat.invalid.formatfor error instead of flowexpr.parse.error?

checker/src/org/checkerframework/checker/i18nformatter/I18nFormatterTransfer.java
```diff
@@ -86,7 +87,8 @@ public I18nFormatterTransfer(I18nFormatterAnalysis analysis, I18nFormatterChecke
                 elseStore.insertValue(firstParam, builder.build());
 
             } catch (FlowExpressionParseException e) {
-                // errors are reported at declaration site
+                // report errors here
+                checker.report(e.getResult(), node);
```
smillst: Shouldn't this be an i18nformat.invalid.formatfor error instead of flowexpr.parse.error?

checker/src/org/checkerframework/checker/experimental/regex_qual/RegexQualifiedTransfer.java
jthaine: Ah, sorry I missed this. Do you know the proper way to report errors in the Qual Framework? I found a way to do it, but it clearly isn't the intended design:
`                analysis.getContext().getChecker().getCheckerAdapter().getVisitor().getTypeFactory().getContext().getChecker().report(e.getResult(), n);`

checker/src/org/checkerframework/checker/regex/classic/RegexTransfer.java
jthaine: To me the assert false says this should never happen, and I am just adding a second way for this error to be reported. I don't know enough about the design of the Regex Checker to determine that it's safe to remove it. At a glance, this method is parsing an expression (`"#1"`) that should always be valid for the given method (`isRegex`), so it does look like this should never fail unless there is a bug in the implementation, and that we should keep the assert.

checker/src/org/checkerframework/checker/experimental/regex_qual/RegexQualifiedTransfer.java
smillst: `analysis.getContext().getCheckerAdapter().report()` should work. But why not just ErrorAbort rather than issue a warning then asserting false?

checker/src/org/checkerframework/checker/regex/classic/RegexTransfer.java
smillst: I didn't notice it was just parsing `#1` -- thanks for pointing that out.  I would just ErrorAbort rather than issue a warning and asserting false.

checker/src/org/checkerframework/checker/i18nformatter/I18nFormatterTreeUtil.java
jthaine: Yes, fixed.

checker/src/org/checkerframework/checker/i18nformatter/I18nFormatterTransfer.java
jthaine: I think that would confuse the user. We are parsing a fixed string (`#1`) to retrieve an argument of a call to a method that was annotated with `@I18nChecksFormat` - there is no `@I18nFormatFor` annotation involved in this process.

checker/src/org/checkerframework/checker/i18nformatter/I18nFormatterTransfer.java
jthaine: I don't think it should be, for a similar reason as stated for the previous case.

checker/src/org/checkerframework/checker/i18nformatter/I18nFormatterTransfer.java
smillst: Ok I see now.  The user shouldn't see this warning ever.  I would ErrorAbort instead. (same for FormatTransfer)

smillst:
Looks good.  Merge when ready.

checker/tests/nullness/EnsuresNonNullIfTest2.java
```diff
@@ -8,6 +8,7 @@
 // Test case for a mysterious error.
 abstract class EnsuresNonNullIfTest2a {
     @EnsuresNonNullIf(result=true, expression="xxx")
+    //:: error: (flowexpr.parse.error)
```
wmdietl: Wouldn't it be nicer if the error were issued on the `@EnsuresNonNullIf`, not on the method declaration? There might be quite a distance between the two.

checker/src/org/checkerframework/checker/experimental/regex_qual/RegexQualifiedTransfer.java
```diff
@@ -82,7 +83,8 @@ public RegexQualifiedTransfer(QualAnalysis<Regex> analysis) {
                         thenStore.insertValue(firstParam, new RegexVal(0));
                     }
                 } catch (FlowExpressionParseException e) {
```
wmdietl: Why is the `try` block so large? Which statements other than `FlowExpressionParseUtil.parse` can throw this exception?

checker/src/org/checkerframework/checker/experimental/regex_qual/RegexQualifiedTransfer.java
```diff
@@ -82,7 +83,8 @@ public RegexQualifiedTransfer(QualAnalysis<Regex> analysis) {
...
-                    assert false;
+                    ErrorReporter.errorAbort("RegexQualifiedTransfer.visitMethodInvocation: could not parse " +
```
wmdietl: This exact same line is duplicated several times throughout these changes. Can you think of a way to reduce this duplication, e.g. by using a method in FlowExpressionParseException?

PR #769, opened 2016-06-02
Release: correct git command

smillst:
When pushing from the bare interm git repositories, the follow error is issued (though the release scripts thinks the push succeeds):
````
fatal: The current branch master has no upstream branch.
To push the current branch and set the remote as upstream, use

    git push --set-upstream origin master
````
The change in this PR specifies origin master in on of the pushes.  A less confusing alternative would be to set the upstream of master when the repository is cloned.

### DIFF
--- a/release/release_utils.py
+++ b/release/release_utils.py
@@ -407,7 +407,7 @@ def push_changes_prompt_if_fail(repo_root):
                 if not prompt_yn("Try again (responding 'n' will skip this push command but will not exit the script) ?"):
                     break
         if is_git(repo_root):
-            cmd = 'git -C %s push' % repo_root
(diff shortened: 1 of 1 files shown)


jthaine:
Looks good. However, I think it's now confusing that `push_changes_prompt_if_fail` specifies the remote and branch while `push_changes` doesn't. Please add a comment, or modify the docstrings, or rename methods to indicate that this is because `push_changes` (called by `commit_tag_and_push`) is used by `release_build` to push to the interm repos, while `push_changes_prompt_if_fail` is used by `release_push` to push to the live repos.

smillst:
Closed in favor of #780

PR #771, opened 2016-06-03
Issue warning if any supported qualifier @Target is incorrect.

smillst:
Allow annotations that are not qualifiers in the qual package to have any @Target.


framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -366,6 +368,17 @@ public AnnotatedTypeFactory(BaseTypeChecker checker) {
         }
     }
 
+    /**
+     * Issue an error if any of the support qualifiers has @Target that contains something besides
+     * TYPE_USE or TYPE_PARAMETER.
+     */
+    private void checkSupportedQuals() {
+        for (Class<? extends Annotation> annotationClass : supportedQuals) {
+            AnnotatedTypes.hasTypeQualifierElementTypes(
```
mernst: I think it's worth a comment to explain why a method with boolean return type is being called for side effect.

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
```diff
@@ -483,6 +482,9 @@ private final void printPaths() {
      * @return the set of loaded annotation classes
      */
     public final Set<Class<? extends Annotation>> getLoadedAnnotationClasses() {
+        if (loadedAnnotations == null) {
```
mernst: I would personally abstract this out into a getLoadedAnnotations method, because otherwise it could be easy to forget to do.  Then comment on line 96 that one should not use the field directly.

framework/tests/simple/README
```diff
@@ -0,0 +1,2 @@
+This directory holds one simple Java file used to start of checkers that test some initialization
```
mernst: Grammar error here; I'm not sure of the right fix.  Also missing period at end of second sentence.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -366,6 +368,17 @@ public AnnotatedTypeFactory(BaseTypeChecker checker) {
         }
     }
 
+    /**
+     * Issue an error if any of the support qualifiers has @Target that contains something besides
```
mernst: If the annotation doesn't contain TYPE_USE or TYPE_PARAMETER but does contain something else, I don't think this method throws an error, but the documentation suggests that it does.  (And it probably should.  You could just check the return value of hasTypeQualifierElementTypes.)
Also, replace "issue" by "throw", since issuing an error sounds like using the report mechanism.

smillst: ErrorReporter.errorAbort(...) does issue an error using the report mechanism and then aborts processing.  I updated the documentation to indicate that this method aborts.

mernst:
The continuous integration job is failing because LockHeld has `@Target({})`.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -366,6 +368,35 @@ public AnnotatedTypeFactory(BaseTypeChecker checker) {
...
     }
 
+    /**
+     * Issue an error and abort if any of the support qualifiers have @Target meta-annotations
+     * that contain something besides TYPE_USE or TYPE_PARAMETER. (@Target({}) is allowed)
+     */
+    private void checkSupportedQuals() {
+        for (Class<? extends Annotation> annotationClass : supportedQuals) {
+            ElementType[] elements = annotationClass.getAnnotation(Target.class).value();
+            String otherElementTypes = null;
```
mernst: I would find the code easier to understand if otherElementTypes was a list, and then printing used String.join, rather than mixing the string processing code in with adding values to otherElementTypes.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -366,6 +368,35 @@ public AnnotatedTypeFactory(BaseTypeChecker checker) {
...
+                    // than TYPE_USE or TYPE_PARAMETER then it isn't a valid qualifier
+                    if (otherElementTypes == null) {
+                        otherElementTypes = element.toString();
+                    } else {
+                        otherElementTypes += ", " + element.toString();
+                    }
+                }
+            }
+            if (otherElementTypes != null) {
+                StringBuffer buf = new StringBuffer("The @Target meta-annotation on ");
```
mernst: I suggest to add "type annotation" at the end of this string (before the annotation class name is added)

mernst:
After two tiny tweaks, this looks good to merge.  Thanks!

smillst:
Merged via dc9f949

PR #772, opened 2016-06-03
Can't call isSuptype on types whose Java types are not subtypes

smillst:


mernst:
This looks OK to me.

framework/src/org/checkerframework/framework/type/DefaultTypeHierarchy.java
```diff
@@ -541,7 +517,14 @@ public Boolean visitDeclared_Typevar(AnnotatedDeclaredType subtype, AnnotatedTyp
 
     @Override
     public Boolean visitDeclared_Union(AnnotatedDeclaredType subtype, AnnotatedUnionType supertype, VisitHistory visited) {
-        return visitUnionSupertype(subtype, supertype, visited);
+        Types types = checker.getTypeUtils();
+        for (AnnotatedDeclaredType supertypeAltern : supertype.getAlternatives()) {
+            if (TypesUtils.isErasedSubtype(types, subtype.getUnderlyingType(), supertypeAltern.getUnderlyingType())
```
wmdietl: Wouldn't it be more concise to add the check for subtyping of the underlying type to `isSubtype`?
Is there ever a reason for not including this check? It looks like it is always occurring together. The only exception I saw was in `visitIntersection_Intersection` which could keep the additional check.

framework/src/org/checkerframework/framework/type/DefaultTypeHierarchy.java
```diff
@@ -541,7 +517,14 @@ public Boolean visitDeclared_Typevar(AnnotatedDeclaredType subtype, AnnotatedTyp
...
+                    && isSubtype(subtype, supertypeAltern)) {
```
wmdietl: This is one of the few places that uses an `&&` to combine the two subtype tests. Below most of them are separate `if`s. Is there a reason for this?

wmdietl:
Is there a test case that fails without this change?

framework/src/org/checkerframework/framework/type/DefaultTypeHierarchy.java
smillst: >Is there ever a reason for not including this check?

Yes, for example isSubtype of Integer and int are not erased Java subtypes, so the check can't be added to isSubtype.

framework/src/org/checkerframework/framework/type/DefaultTypeHierarchy.java
smillst: I changed to code so all the checks use `&&`.

framework/src/org/checkerframework/framework/type/DefaultTypeHierarchy.java
wmdietl: It would seem much simpler to add that special casing into isSubtype.
An int is never a subtype of Integer, because boxing/unboxing would apply first.
Making many invocations of isSubtype simpler and documenting the boxing special case in isSubtype seems beneficial and clearer to me.

smillst: As [documented](http://types.cs.washington.edu/checker-framework/current/api/org/checkerframework/framework/type/TypeHierarchy.html#isSubtype-org.checkerframework.framework.type.AnnotatedTypeMirror-org.checkerframework.framework.type.AnnotatedTypeMirror-), the Java types of the arguments to isSubtype are not always in a subtype relationship.  

>It would seem much simpler to add that special casing into isSubtype.

It would not be correct to do so.  For example, `isSubtype(@C Cloneable & @A List<@B String, @A List<@B String>)` if the `isErasedSubtype` is moved to `isSubtype`, then `isSubtype` called recursively on `@C Cloneable` and `@A List<@B String>` would return false which would cause `isSubtype(@C Cloneable & @A List<@B String, @A List<@B String>)` to return false as well.

wmdietl: I don't understand your example. At the moment the code is

```
for (alternatives) {
  if (isErasedSubtype(A, B) && isSubtype(A, B) {
    return true;
  }
}
return false;
```

How is it not semantically equivalent to write:

```
for (alternatives) {
  if (isSubytpe(A, B)) {
    return true;
  }
}
return false;
```

and strengthen isSubtype:

```
boolean isSubtype(A, B) {
   if (!isErasedSubtype(A, B) {
     return false;
   }
}
```

smillst: For Intersection_Declared it would be equivalent.  It's not for  [Intersection_Intersection](https://github.com/typetools/checker-framework/pull/772/files/54a1731810bb71ca587bb9e25a87498e539e6421#diff-ace109c2136c5d7b905d5679ee384f43R567) nor [Intersection_Primitive](https://github.com/typetools/checker-framework/pull/772/files/54a1731810bb71ca587bb9e25a87498e539e6421#diff-ace109c2136c5d7b905d5679ee384f43R553).

wmdietl: I agree on I_I and noted that in my first comment.
Primitives need the boxing/unboxing handling and that would cover the Primitive cases.
Logically, subtyping on annotated types should be stronger than subtyping on the underlying type. Putting this logic together into isSubtype would make this clear and prevent forgetting about special cases in different invocations.
If we keep the two test separate, in a new location someone might forget about boxed/unboxed types needing special treatment.
If we put all this logic into isSubtype, there is no repetition and the purpose should be clearer.

smillst: The current behavior of isSubtype is undefined if the Java types are not in the correct relationship.  Are you proposing that isSubtype should return false if they are not in the correct relationship?

wmdietl: Yes. That's effectively what the invocations here enforce by first calling isErasedSubtype. Putting this directly into isSubtype would strengthen its specification and simplify the invocations.

smillst: It would probably be good to do this, but it is not as simple as adding isErasedSubtype to isSubtype.  Aside from primitives, isErasedSubtype fails for the following combinations to isSubtype (and there are probably more):
````
int[], Object[]
U extends Object & Iterable<Object>, Iterable<Object>
Throwable&Cloneable&Unions.MyInterface<String>,  Object&Cloneable&Unions.MyInterface<String>
````
So, if we would like to add this functionality, it should be done in a separate pull request.

wmdietl: Ok, if you think that is easier.
Did you have a test case for this pull request?

smillst: This is a refactoring, so there are already some test cases.  More complex ones require asSuper to work correctly.  See tests in all-systems with Union or Intersection in name.

wmdietl: Ok, so I'm not sure why splitting this up into two refactorings makes it easier, but if you prefer, let's do that. Please file an issue for the remaining work.
(The PR title made it sound like something doesn't work without this change.)

smillst: I have more test cases in asSuper that crash on master because of asSuper bugs that also crash on asSuper because of this issue.

PR #774, opened 2016-06-03
Rewrite AsSuper

### LINKED ISSUE #717: Rewrite AnnotatedTypes#asSuper(type, supertype)
In order to rewrite AnnotatedTypes#LUB (issue #643), AnnotatedTypes#asSuper() must also be rewritten.  It currently returns null when the "supertype" is an intersection or union type.

smillst:
The rewrite is in a new class [AsSuperVisitor](https://github.com/typetools/checker-framework/pull/774/files#diff-b41a83c7dc954a8830e7a07b1a6df424) that's probably the best place to start the review.

Daikon and plumb-lib both type check.

Fixes #717
(PR #781 should be merged after this.)
(framework/tests/all-systems/UnionCrash.java which I thought was an asSuper problem is still crashing, because of LUB. I opened #775 to fix this crash.)


framework/.classpath
```diff
@@ -11,7 +11,5 @@
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
 	<classpathentry combineaccessrules="false" exported="true" kind="src" path="/dataflow"/>
 	<classpathentry combineaccessrules="false" exported="true" kind="src" path="/javacutil"/>
-	<classpathentry combineaccessrules="false" kind="src" path="/annotation-file-utilities"/>
```
wmdietl: Why are these being removed?
`checker-framework/framework/src/org/checkerframework/common/wholeprograminference/AnnotationConverter.java` and other classes in `wholeprograminference` use those projects.

javacutil/src/org/checkerframework/javacutil/ErrorReporter.java
```diff
@@ -31,6 +31,21 @@ public static void errorAbort(String msg) {
         }
     }
 
+    /**
+     * Log an error message and abort processing.
+     * Call this method instead of raising an exception.
+     *
+     * @param msg the error message to log
```
wmdietl: It would be good to document that in contrast to the above method this is now a format string.

wmdietl:
The build fails on Java 8. I think on Friday we had discussed that you know how to fix that issue.
I made a few quick comments, but need to spend more time on this.

checker/src/org/checkerframework/checker/nullness/CollectionToArrayHeuristics.java
```diff
@@ -175,8 +175,7 @@ private boolean isHandledArrayCreation(Tree argument, String receiver) {
     private boolean isNonNullReceiver(MethodInvocationTree tree) {
         // check receiver
         AnnotatedTypeMirror receiver = atypeFactory.getReceiverType(tree);
-        AnnotatedDeclaredType collection = (AnnotatedDeclaredType) AnnotatedTypes.asSuper(processingEnv.getTypeUtils(), atypeFactory, receiver, collectionType);
-        assert collection != null;
+        AnnotatedDeclaredType collection = AnnotatedTypes.asSuper(atypeFactory, receiver, collectionType);
 
         if (collection.getTypeArguments().isEmpty()
```
wmdietl: Here and in a few places below there are checks whether type arguments are empty in places where that seems impossible. E.g. here the Collection always has a type argument. When are these checks triggered?
I didn't see in `AnnotatedTypes.asSuper` when it might return a raw type.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -2058,6 +2059,43 @@ public AnnotatedPrimitiveType getUnboxedType(AnnotatedDeclaredType type)
     }
 
     /**
+     * Returns AnnotatedDeclaredType with underlying type String and annotations copied from type.
```
wmdietl: The documentation doesn't really make clear what the use of this method is.
Should it be used when something is automatically converted to String? At the moment there is only a call in an error method.

framework/src/org/checkerframework/framework/type/AsSuperVisitor.java
```diff
@@ -0,0 +1,807 @@
...
+        return super.visit(type, superType, p);
+    }
+
+    /**
+     * The code in this class is assuming that the primary annotation of an {@link AnnotatedIntersectionType}
+     * is the greatest lower bound of the annotations on its direct super types and that the primary
+     * annotation of an {@link AnnotatedUnionType} is the least upper bound of its alternatives.
+     * This method makes this assumption true.
+     */
+    private void ensurePrimaryIsCorrectForUnionsAndIntersections(AnnotatedTypeMirror type) {
```
wmdietl: What is the use of primary annotations on Intersection and Union types?
If I remember correctly, Java 8 provides no syntax to provide an annotation for the overall type.
Would it prevent accidental additions by preventing primary annotations on Intersection/Union in the AnnotatedTypeMirror subclasses?
How can this property ever be violated if there is no syntax to specify something else?

framework/src/org/checkerframework/framework/type/AsSuperVisitor.java
```diff
@@ -0,0 +1,807 @@
...
+        if (type instanceof AnnotatedIntersectionType) {
```
wmdietl: There are lots of instanceof tests. Using getKind() would be cleaner.

framework/src/org/checkerframework/framework/type/AsSuperVisitor.java
smillst: In general, we need to actually decide what the primary annotation on a union or intersection should mean and then implement that.  So, I choose something sensible so that I could finish asSuper.
>What is the use of primary annotations on Intersection and Union types?

They are only useful so that getAnnotations returns a non-null value.

>If I remember correctly, Java 8 provides no syntax to provide an annotation for the overall type.
Would it prevent accidental additions by preventing primary annotations on Intersection/Union in the AnnotatedTypeMirror subclasses?
How can this property ever be violated if there is no syntax to specify something else?

A checker can set the primary annotation to what ever it likes.

checker/build.xml
```diff
@@ -1253,6 +1253,12 @@ So, use our own archived version.
       <ant dir="${basedir}/../../checker-framework-demos">
       </ant>
     </target>
+    <target name="check-demos-travis" depends="demos.check,-demos.pull,-demos.clone"
```
wmdietl: Empty line between targets.

framework/src/org/checkerframework/framework/type/AsSuperVisitor.java
```diff
@@ -0,0 +1,808 @@
...
+     * @return {@code superType} with annotations copied from {@code type} and type variables
+     * substituted from {@code type}.
+     */
+    @SuppressWarnings("unchecked")
+    public <T extends AnnotatedTypeMirror> T asSuper(AnnotatedTypeMirror type, T superType) {
+        if (type == null || superType == null) {
+            ErrorReporter.errorAbort("AsSuperVisitor type and supertype cannot be null.");
+            return null; // dead code
+        } else if (type == superType) {
+            return (T) type.deepCopy();
```
wmdietl: Can you add a comment in the javadoc about which parameters might be copies of each other? Are you guaranteed to get a fresh object back?

framework/src/org/checkerframework/framework/type/AsSuperVisitor.java
```diff
@@ -0,0 +1,808 @@
...
+        }
+        AnnotatedTypeMirror copySuperType = superType.deepCopy(false);
```
wmdietl: Can you comment why the visitor gets copies of the parameters? Does the visitor modify the parameters?
Why is one a shallow copy and the other not? Can you add comments?
Extract the type.deepCopy into another local variable.

framework/src/org/checkerframework/framework/type/AsSuperVisitor.java
smillst: The pre and post conditions of asSuper are documented in AnnotatedTypes#asSuper.  I improved the Javadoc to point to that method.

framework/src/org/checkerframework/framework/type/AsSuperVisitor.java
```diff
@@ -0,0 +1,813 @@
...
+            ErrorReporter.errorAbort("AsSuperVisitor type and supertype cannot be null.");
+            return null; // dead code
+        } else if (type == superType) {
+            return (T) type.deepCopy();
+        }
+
+        // This visitor modifies superType and may return type, so pass it copies so that it that
+        // parameters to asSuper are not changed and a copy is returned.
+        AnnotatedTypeMirror copyType = type.deepCopy();
+        AnnotatedTypeMirror copySuperType = superType.deepCopy(false);
```
smillst: I deleted the following comment for @wmdietl on this line:
>Can you comment why the visitor gets copies of the parameters? Does the visitor modify the parameters?
Why is one a shallow copy and the other not? Can you add comments?
Extract the type.deepCopy into another local variable.

smillst: Addressed a few lines above

smillst:
The Travis job now passes running all of the demos.  I output the resource stats when type checking Daikon with the Nullness Checker on [master](https://github.com/typetools/checker-framework/files/342316/check-nullness-master.txt) and on [asSuper](https://github.com/typetools/checker-framework/files/342317/check-nullness-asSuper.txt).  The memory usage was about the same.

smillst:
Merged via 754c114

Issue #756, opened 2016-05-27
Format String Checker, I18n Format String Checker: must implement LUB


jthaine:
The Format String Checker and I18n Format String Checker both have custom isSubtype implementations, which implies that they must also have custom LUB and GLB implementations. Currently GLB is implemented for both checkers but not LUB.

Related issues (GLB implementations): #691 and #723

PR #777, opened 2016-06-06
Fix issue #756 (Format String Checker, I18n Format String Checker: must implement LUB)

jthaine:
Questions:
-What would you like the GlbChecker classes renamed to?
-The GLB and LUB of two `@[I18n]InvalidFormat(...)` qualifiers is considered to be `@[I18n]InvalidFormat(null)`. Returning the same value for both LUB and GLB should be OK since `isSubtype` considers any two `@[I18n]InvalidFormat(...)` qualifiers to be siblings. I can't think of a scenario for which this could cause a problem, but please let me know if there is one.


smillst:
>-What would you like the GlbChecker classes renamed to?

LubGlbChecker would be fine. 

> -The GLB and LUB of two @[I18n]InvalidFormat(...) qualifiers is considered to be @[I18n]InvalidFormat(null). Returning the same value for both LUB and GLB should be OK since isSubtype considers any two @[I18n]InvalidFormat(...) qualifiers to be siblings.

If A is not a subtype of LUB(A,B), then LUB wasn't implemented correctly.  Similarly, if A is not a supertype of GLB(A,B), the GLB is not implemented correctly. Also, `@InvalidFormat(null)` is not a legal annotation mirror and might cause null pointer exceptions.

jthaine:
All `@InvalidFormat` annotations are siblings, and likewise for `@I18nInvalidFormat`. `isSubtype` will always treat them as siblings. Hence the requirements you stated about LUB and GLB are satisfied. In other words, with the current implementation where LUB and GLB return the annotation itself when it's computing the LUB/GLB of two identical `@InvalidFormat(...)` annotations, and where they return `@InvalidFormat(null)` otherwise, isSubtype won't be able to tell the difference. So I don't think this will be a problem.

As for `@InvalidFormat(null)` not being legal, I don't expect there to be a non-contrived scenario where two distinct `@InvalidFormat(...)` annotations (that are not themselves both `@InvalidFormat(null)`) will be LUBed or GLBed. So I think it's safe to return `@InvalidFormat(null)` in this case.

smillst:
>All @InvalidFormat annotations are siblings, and likewise for @I18nInvalidFormat. isSubtype will always treat them as siblings.

Siblings would mean that isSubtype(`@InvalidFormat(A), @InvalidFormat(B)`) is false.  Looking at the code, that isn't the case.  All `@InvalidFormat` annotations are equivalent qualifiers.

>As for @InvalidFormat(null) not being legal, I don't expect there to be a non-contrived scenario where two distinct @InvalidFormat(...) annotations (that are not themselves both @InvalidFormat(null)) will be LUBed or GLBed. So I think it's safe to return @InvalidFormat(null) in this case.

````
import static org.checkerframework.checker.formatter.qual.ConversionCategory.*;
import org.checkerframework.checker.formatter.qual.Format;

class Test {
    void m(boolean flag) {
        String.format(flag ? "%<f %d" :"%f %4", 0.8, 42);
    }
}
````
The above code cause the following crash:
````
error: No element with name 'value' in annotation @org.checkerframework.checker.formatter.qual.InvalidFormat; invoke the compiler with -AprintErrorStack to see the stack trace.
1 error
````
Instead of setting value to null, could the value Strings be concatenated if they aren't equal?

smillst:
Looks good to me.  Would you please squash and merge?

PR #585, opened 2016-02-05
Lock Checker implementation providing value protection semantics

jthaine:
The documentation for this feature is available in the Lock Checker manual chapter and in the javadocs for the lock annotations in this pull request.

PR #778, opened 2016-06-07
Improvements based on feedback from pull request #585

jthaine:


framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -2651,7 +2651,7 @@ protected void parseStubFiles() {
      * @return the annotation mirror for anno
      */
     @Override
-    public AnnotationMirror getDeclAnnotation(Element elt,
+    public final AnnotationMirror getDeclAnnotation(Element elt,
```
wmdietl: An `@see` to the three-argument getDeclAnnotation would be good and a note that that version should be overridden to adapt the behavior of this method and the `-NoAliases` version.

framework/src/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -49,12 +50,154 @@ public static ContractsUtils getInstance(
...
     /**
+     * Represents a pre- or postcondition that must be verified by
+     * {@code BaseTypeVisitor} or one of its subclasses. Automatically
+     * extracted from annotations with meta-annotations
+     * {@code @PreconditionAnnotation} or {@code @PostconditionAnnotation},
+     * such as {@code @RequiresNonNull} and {@code @EnsuresNonNull}.
+     * Can also be generated, such as in
+     * {@code LockVisitor.generatePreconditionsBasedOnGuards}.
+     */
+    public static class PreOrPostcondition {
```
wmdietl: This class is great! It made looking through the code so much easier compared to the previous Pair!

framework/src/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -49,12 +50,154 @@ public static ContractsUtils getInstance(
...
+
+        public PreOrPostcondition(String expression, String annotationString) {
+            this.expression = expression;
+            this.annotationString = annotationString;
+        }
+    }
+
+    /**
+     * An Iterable set of PreOrPostcondition entries.
+     * Implements the Iterable interface but not the Set interface.
```
wmdietl: I'm not so sure about the usefulness of this class and this comment. Can you document more why this is better than just using `Set<PreOrPostcondition>`?

framework/src/org/checkerframework/framework/util/ContractsUtils.java
wmdietl: Oh, the only thing I had wondered is if the correct camel-case is `PreOrPostCondition` to make the `Condition` part easier to notice.

framework/src/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -49,12 +50,154 @@ public static ContractsUtils getInstance(
...
+    }
+
+    /**
+     * Represents a conditional postcondition that must be verified by
+     * {@code BaseTypeVisitor} or one of its subclasses. Automatically
+     * extracted from annotations with meta-annotation
+     * {@code @ConditionalPostconditionAnnotation}, such as
+     * {@code EnsuresNonNullIf}.
+     */
+    public static class ConditionalPostcondition {
```
wmdietl: Should this be a subclass of `PreOrPostcondition`?
Maybe a note to keep the two classes in sync would be enough.

framework/src/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -49,12 +50,154 @@ public static ContractsUtils getInstance(
...
+     * Can also be generated, such as in
+     * {@code LockVisitor.generatePreconditionsBasedOnGuards}.
+     */
+    public static class PreOrPostcondition {
+        /**
+         * The expression for which the condition must hold, such as
+         * {@code "foo"} in {@code @RequiresNonNull("foo")}.
+         */
+        public final String expression;
+        /**
```
wmdietl: Add empty line between field declarations.

wmdietl:
Hi Javier, these are great improvements! I find the code much easier to read using the new utility classes.
Some small comments, but looks good to merge otherwise.

framework/src/org/checkerframework/framework/util/ContractsUtils.java
jthaine: I think of Precondition as one word. Throughout the code we follow this convention, such as in PreconditionAnnotation and ConditionalPostconditionAnnotation.

wmdietl: True, consistency is best.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
jthaine: The three-argument getDeclAnnotation is currently private. Should we make it protected?

wmdietl: There is a `getDeclAnnotations(Element elt)` version which is public, but non-final.
Maybe we should make public methods which are not intended to be overridden final?
Let's just keep this as it is and see whether we need to change this later.

wmdietl:
Final version looks really nice. Please merge.

smillst:
Merged via 74f97e2

Issue #752, opened 2016-05-27
Flow Expression parser: add support for package names


jthaine:
In order to refer to a static field in a class in a different package, a flow expression needs to include a package name. For example:

    package foo;
    public class A { public static @Nullable Object field; }

    package bar;
    public class B {
        @RequiresNonNull("foo.A.field")
        void method() { ... }
    }

The Flow Expression parser must support parsing of package names to enable this scenario.

PR #782, opened 2016-06-09
Fix issue #752 (Flow Expression parser: add support for package names)

jthaine:


checker/tests/nullness/Issue752.java
```diff
@@ -0,0 +1,42 @@
...
+public class Issue752 {
+
+    Issue752 field = new Issue752();
+    static Issue752 staticField = new Issue752();
+
+    @RequiresNonNull("java.lang")
+    //:: error: (flowexpr.parse.error)
+    void method1() { }
+
+    @RequiresNonNull("java.lang.String")
```
mernst: Shouldn't this issue an error?  The string might parse correctly as a type, but the Nullness Checker should not accept that as a legal argument to `@RequiresNonNull`.

checker/tests/nullness/Issue752.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    @RequiresNonNull("a.b.c.Issue752.staticField")
+    void method6() { }
+
+    @RequiresNonNull("a.b.c.Issue752.staticField.field")
+    void method7() { }
+
+    @RequiresNonNull("a.b.c.Issue752.field")
+    //:: error: (flowexpr.parse.error)
```
mernst: Why is this an error?

It would help to add a brief comment, for each string, indicating why it is an error.

mernst:
How does this pull request enable a method call to have a class (and optionally package) specified?  That is necessary if the method is static.  Could you add such a test case?

checker/tests/nullness/Issue752.java
jthaine: Agreed. Since fixing this is an item on my ongoing agenda, for the purposes of this pull request I added ".class" to these expressions so that they are valid. When I fix that work item, I'll have a test case for the expected error.

Issue #594, opened 2016-02-11
Confusing error message when checking generic type variable


atomicknight:
Environment: Checker Framework 1.9.11, Oracle JDK 1.8.0_65, Windows 7/8.1

The following class yields a confusing error message:
```java
package org.example;

import org.checkerframework.checker.nullness.qual.Nullable;

public class ResultWrapper<T> {
	private @Nullable T result = null;

	// Should return @Nullable T
	private T getResult() {
		return result;
	}
}
```
Error:
```
d:\workspace\cf-testcases\genericvariable1\src\main\java\org\example\ResultWrapper.java:9: error: [return.type.incompatible] incompatible types in return.
                return result;
                       ^
  found   : T extends @Initialized @Nullable Object
  required: T extends @Initialized @Nullable Object
```

mernst:
Confirmed.  Thanks for the bug report!

I added a test case as file
checker-framework/checker/tests/nullness/generics/GenericReturnField.java

smillst:
Using the `-AprintVerboseGenerics` option makes the found and require types distinct :

````
error: [return.type.incompatible] incompatible types in return.
        return result;
               ^
  found   : T[ extends @Initialized @Nullable Object super @Initialized @Nullable Void]
  required: T[ extends @Initialized @Nullable Object super @Initialized @NonNull Void]
````

This option is needs to be documented in the manual.

mernst:
It is good to document the command-line option.
More importantly, the Checker Framework mechanism for printing "found" and "required" needs to be adjusted so that it prints every component of the types that differ, regardless of the setting of `-AprintVerboseGenerics`.

wmdietl:
The mechanism to print types in error messages already has logic to print "invisible" annotations, if the error message is otherwise identical.
This mechanism should be easy to extend to also automatically use printVerboseGenerics if the error message would otherwise be identical.

mernst:
That mechanism should be used not just if the message would otherwise be identical, but anytime changes would not be visible.  For example, if we have:

 * found: `T[ extends @A super @B]`
 * required: `T[extends @C super @D]`

then it is not enough to print

 * found: `T[ extends @A]`
 * required: `T[extends @C]`

even though those (partial) printed representations are not identical.

wmdietl:
I agree. As soon as one of expected/found is a type variable or wildcard, the verbose generics option should be used. In general, the subtype test might have failed because of the lower bound, which is usually not shown.

mernst:
[...]

PR #785, opened 2016-06-09
AnnotatedTypeMirror.toString(true) uses all verbose settings

smillst:
Including printAllQualifiers and printVerboseGenerics.  This means that the
full generics information will be printed at assignment types incompatible when
the shorter toString does not differ between types. (This implements so of the
functionality discussed in #594, but not all.  In particular, https://github.com/typetools/checker-framework/issues/594#issuecomment-183120237 is not addressed.)

### DIFF
--- a/framework/src/org/checkerframework/framework/type/DefaultAnnotatedTypeFormatter.java
+++ b/framework/src/org/checkerframework/framework/type/DefaultAnnotatedTypeFormatter.java
@@ -77,7 +77,6 @@ public class DefaultAnnotatedTypeFormatter implements AnnotatedTypeFormatter {
     @Override
     public String format(final AnnotatedTypeMirror type, final boolean printVerbose) {
         formattingVisitor.setVerboseSettings(printVerbose);
-        formattingVisitor.setVerboseSettings(printVerbose);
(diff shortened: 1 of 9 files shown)


framework/src/org/checkerframework/framework/type/DefaultAnnotatedTypeFormatter.java
```diff
@@ -70,13 +70,14 @@ protected DefaultAnnotatedTypeFormatter(FormattingVisitor visitor) {
...
+        formattingVisitor.resetPrintVerboseSettings();
         return formattingVisitor.visit(type);
     }
 
     @Override
-    public String format(final AnnotatedTypeMirror type, final boolean printInvisibles) {
-        formattingVisitor.setCurrentPrintInvisibleSetting(printInvisibles);
+    public String format(final AnnotatedTypeMirror type, final boolean printVerbose) {
+        formattingVisitor.setVerboseSettings(printVerbose);
+        formattingVisitor.setVerboseSettings(printVerbose);
```
mernst: Duplicated line:  same as line above.

mernst:
This is ready to merge once you address my one comment.

PR #791, opened 2016-06-13
Improve FlowExpressionParseUtil documentation

mernst:


framework/src/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -151,6 +161,9 @@
     /**
      * Private implementation of {@link #parse} with a choice of which classes
      * of expressions should be parsed.
+     * Returns null only if {@code allowSelf} is true and the expression
```
jthaine: I am not seeing this in the code below.
Did you mean "Returns null only if {@code allowItself} is true, and 'itself' is passed in as (part of) the string to parse and no receiver named 'itself' could be found.", or am I missing something?

mernst: You are right.  I will fix this.

jthaine:
The changes look good, thanks.

PR #799, opened 2016-06-15
Rename the special lock expression "itself" to "<self>"

jthaine:


checker/tests/lock/FlowExpressions.java
```diff
@@ -24,17 +24,5 @@ public void method() {
         synchronized(getm()) {
             getm().field = new Object();
         }
-
```
mernst: Why is this block removed?

jthaine: It was testing a situation where "itself" could refer to either a variable (within the block) or "<self>" (outside of this block).

mernst:
Looks good.  I had one question.  Feel free to merge when that is resolved and the tests pass.

checker/tests/lock/FlowExpressions.java
mernst: OK, sounds good.  Thanks.

PR #800, opened 2016-06-15
Fields in methods whose receiver is not fully initialized

### LINKED ISSUE #779: Nullness Checker/Initialization Checker issue with UnderInitialization
Let us consider two files with the Nullness Checker. 
InitZero is in the repository as TODO
Init is in the repository as TODO

InitZero is the version without annotations. 
Java compiles it without a warning, but executing it will result in an NPE: 

```
at B.foo(InitZero.java:17) 
at A.<init>(InitZero.java:5) 
at B.<init>(InitZero.java:13) 
at InitZero.main(InitZero.java:23) 
```

Compiling it with the Checker Framework and `-processor nullness` gives: 

```
InitZero.java:5: error: [method.invocation.invalid] call to foo() not 
allowed on the given receiver. 
foo(); 
  ^ 
  found   : @UnderInitialization(A.class) @NonNull A 
  required: @Initialized @NonNull A 
1 error 
```

This is the expected behavior and it prevents the invalid invocation 
of a virtual method from the constructor. 

However, I then went ahead and annotated the foo method with 
`@UnderInitialization(A.class)`.
I expected the warning on the invocation of foo to go away and instead 
expected a warning on the implementation of B.foo. 

However, again running with `-processor nullness`, there was not a 
single warning and Init.java compiles successfully. When executing the 
resulting program, one gets the same NPE again. 

This is with the 2.0.1 release version of the Checker Framework. 

[This issue was first discussed here: https://groups.google.com/d/msg/checker-framework-dev/YhNZvCE2lmw/w3NxDIgHBgAJ ]

smillst:
Fixes #779 

(I'm currently testing this with daikon and plume-lib.  This change could cause new errors in those code bases.)

### DIFF
--- a/checker/tests/nullness/Initializer.java
+++ b/checker/tests/nullness/Initializer.java
@@ -63,8 +63,7 @@ class Initializer {
 
     String f = "";
     void t1(@UnknownInitialization @Raw Initializer this) {
-        // this is potentially uninitialized, but the static type of f, as well as
-        // the initializer guarantee that it is initialized.
+        //:: error: (dereference.of.nullable)
         this.f.toString();
     }
 
--- a/checker/tests/nullness/Initializer.java
+++ b/checker/tests/nullness/Initializer.java
@@ -66,5 +66,24 @@ class Initializer {
         //:: error: (dereference.of.nullable)
         this.f.toString();
     }
+    String fieldF = "";
 
+}
+class SubInitializer extends Initializer {
+
+    String f = "";
+    void subt1(@UnknownInitialization(Initializer.class) @Raw(Initializer.class) SubInitializer this) {
+        fieldF.toString();
+        super.f.toString();
+        //:: error: (dereference.of.nullable)
+        this.f.toString();
+    }
+    void subt2(@UnknownInitialization @Raw SubInitializer this) {
+        //:: error: (dereference.of.nullable)
+        fieldF.toString();
+        //:: error: (dereference.of.nullable)
+        super.f.toString();
+        //:: error: (dereference.of.nullable)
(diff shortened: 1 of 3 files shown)


mernst:
Looks good to me; squash and merge whenever you are ready.

smillst:
Results of daikon type-check with Nullness Checker: https://travis-ci.org/smillst/daikon-typecheck-nullness/jobs/137905342

wmdietl:
Both new errors in daikon are like the new error in checker/tests/nullness/Initializer.java
Can you expand Initializer and add a test with `(@UnknownInitialization(Initializer.class) @Raw(Initializer.class) Initializer this` and a similar field with initializer.
That would be a simple fix for Daikon.

In general, if a reference is `@UnknownInitialization(A.class)` I think one can assume that field initializers up to and including A.class have been executed. So instead of just checking for `isNotFullyInitializedReceiver(methodTree);` should it look at the frame of the receiver to decide?

smillst:
>Both new errors in daikon are like the new error in checker/tests/nullness/Initializer.java
Can you expand Initializer and add a test with `(@UnknownInitialization(Initializer.class) @Raw(Initializer.class) Initializer this` and a similar field with initializer.

Done

>In general, if a reference is `@UnknownInitialization(A.class)` I think one can assume that field initializers up to and including A.class have been executed. So instead of just checking for `isNotFullyInitializedReceiver(methodTree);` should it look at the frame of the receiver to decide?

This is implemented in InitializationAnnotatedTypeFactory#postAsMemberOf via the call to InitializationAnnotatedTypeFactory#computeFieldAccessType.

PR #814, opened 2016-06-22
add isAnonymous

Jianchu:
add method `boolean isAnonymous(TypeMirror type)` to check whether a TypeMirror is Anonymous type.

### DIFF
--- a/javacutil/src/org/checkerframework/javacutil/TypesUtils.java
+++ b/javacutil/src/org/checkerframework/javacutil/TypesUtils.java
@@ -138,13 +138,9 @@ public final class TypesUtils {
      * @return whether the argument is an anonymous type
      */
     public static boolean isAnonymous(TypeMirror type) {
-        if (type instanceof DeclaredType) {
-            if (((TypeElement) ((DeclaredType) type).asElement()).getNestingKind().equals(
-                    NestingKind.ANONYMOUS)) {
-                return true;
-            }
-        }
-        return false;
+        return (type instanceof DeclaredType)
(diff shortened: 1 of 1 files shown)


javacutil/src/org/checkerframework/javacutil/TypesUtils.java
```diff
@@ -132,9 +133,24 @@ public static boolean isThrowable(TypeMirror type) {
     }
 
     /**
+     * Returns true iff the argument is an anonymous type.
+     *
+     * @return whether the argument is an anonymous type
+     */
+    public static boolean isAnonymous(TypeMirror type) {
+        if (type instanceof DeclaredType) {
```
wmdietl: I would rewrite that as one if with && of the two conditions or, even simpler, a `return A && B`;

javacutil/src/org/checkerframework/javacutil/TypesUtils.java
```diff
@@ -132,9 +133,24 @@ public static boolean isThrowable(TypeMirror type) {
     }
 
     /**
+     * Returns true iff the argument is an anonymous type.
+     *
+     * @return whether the argument is an anonymous type
+     */
+    public static boolean isAnonymous(TypeMirror type) {
```
wmdietl: In a comment, `checker/src/org/checkerframework/checker/guieffect/GuiEffectTypeFactory.java` refers to `TypesUtils.isAnonymousType`, but such a method doesn't exist.
`InternalUtils.java` has `isAnonymousConstructor(final MethodTree method)` but a cross-reference doesn't seem necessary.
I didn't find any other method like this, so I think this is a good addition.

PR #815, opened 2016-06-23
using AppClassLoader to load Annotation classes of a checker.

CharlesZ-Chen:
Previously, the `AnnotationClassLoader` using itself's ClassLoader (which is null, usually means will using the bootstrap class loader) to load annotation Classes, which means the annotation class files need to be set in bootstrap classpath.

This will cause problems when loading an external annotation class outside of CheckerFramework. Even with correct `fullyQualifiedClassName`, the previously code will still throw ClassNotFoundException.

I changed the code to using the class loader returned by `AnnotationClassLoader#getAppClassLoader()`.
( This `getAppClassLoader()` method will return current checker's ClassLoader or return `SystemClassLoader` if checker's ClassLoader is null. )

This change will let the Checker Framework correctly loads external checker's qualifier classes, either by `BaseTypeChecker`'s ClassLoader, or could allow external checker simply put it's class files in environment variable `CLASSPATH` (or `-classpath`), and then let `SystemClassLoader` load those classes.

### DIFF
--- a/checker/manual/fenum-checker.tex
+++ b/checker/manual/fenum-checker.tex
@@ -123,23 +123,17 @@ If you define your own annotation(s), provide the name(s) of the annotation(s)
 through the \code{-Aquals} option, using a comma-no-space-separated notation:
 
 \begin{alltt}
-  javac -Xbootclasspath/p:\textit{/full/path/to/myProject/bin}:\textit{/full/path/to/myLibrary/bin} \ttbs
+  javac -classpath \textit{/full/path/to/myProject/bin}:\textit{/full/path/to/myLibrary/bin} \ttbs
         -processor org.checkerframework.checker.fenum.FenumChecker \ttbs
         -Aquals=\textit{myModule.qual.MyFenum} MyFile.java ...
 \end{alltt}
-
-The annotations listed in \code{-Aquals} must be accessible to
-the compiler during compilation in the classpath.  In other words, they must
-already be compiled (and, typically, be on the javac bootclasspath)
-before you run the Fenum Checker with \code{javac}.  It
-is not sufficient to supply their source files on the command line.
 
 You can also provide the fully-qualified paths to a set of directories
 that contain the annotations through the \code{-AqualDirs} option,
 using a colon-no-space-separated notation. For example:
 
 \begin{alltt}
-  javac -Xbootclasspath/p:\textit{/full/path/to/myProject/bin}:\textit{/full/path/to/myLibrary/bin} \ttbs
+  javac -classpath \textit{/full/path/to/myProject/bin}:\textit{/full/path/to/myLibrary/bin} \ttbs
         -processor org.checkerframework.checker.fenum.FenumChecker \ttbs
         -AqualDirs=\textit{/full/path/to/myProject/bin}:\textit{/full/path/to/myLibrary/bin} MyFile.java ...
 \end{alltt}
--- a/checker/manual/fenum-checker.tex
+++ b/checker/manual/fenum-checker.tex
@@ -128,7 +128,11 @@ through the \code{-Aquals} option, using a comma-no-space-separated notation:
         -Aquals=\textit{myModule.qual.MyFenum} MyFile.java ...
 \end{alltt}
 
-The annotations listed in \code{-Aquals} must be accessible to the compiler during compilation in the classpath.  In other words, they must already be compiled (and, typically, be on the javac classpath) before you run the Subtyping Checker with \code{javac}. It is not sufficient to supply their source files on the command line.
+The annotations listed in \code{-Aquals} must be accessible to
+the compiler during compilation in the classpath.  In other words, they must
+already be compiled (and, typically, be on the javac classpath)
+before you run the Subtyping Checker with \code{javac}.  It
+is not sufficient to supply their source files on the command line.
 
 You can also provide the fully-qualified paths to a set of directories
 that contain the annotations through the \code{-AqualDirs} option,
--- a/checker/manual/fenum-checker.tex
+++ b/checker/manual/fenum-checker.tex
@@ -131,7 +131,7 @@ through the \code{-Aquals} option, using a comma-no-space-separated notation:
 The annotations listed in \code{-Aquals} must be accessible to
 the compiler during compilation in the classpath.  In other words, they must
 already be compiled (and, typically, be on the javac classpath)
-before you run the Subtyping Checker with \code{javac}.  It
+before you run the Fenum Checker with \code{javac}.  It
 is not sufficient to supply their source files on the command line.
 
 You can also provide the fully-qualified paths to a set of directories
--- a/checker/manual/fenum-checker.tex
+++ b/checker/manual/fenum-checker.tex
@@ -128,6 +128,8 @@ through the \code{-Aquals} option, using a comma-no-space-separated notation:
         -Aquals=\textit{myModule.qual.MyFenum} MyFile.java ...
 \end{alltt}
 
+The annotations listed in \code{-Aquals} must be accessible to the compiler during compilation in the classpath.  In other words, they must already be compiled (and, typically, be on the javac classpath) before you run the Subtyping Checker with \code{javac}. It is not sufficient to supply their source files on the command line.
+
 You can also provide the fully-qualified paths to a set of directories
 that contain the annotations through the \code{-AqualDirs} option,
 using a colon-no-space-separated notation. For example:
--- a/framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
+++ b/framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
@@ -372,8 +372,8 @@ public class AnnotationClassLoader {
      * it has access to
      *
      * The classpaths will be obtained in the order of: 1) extension
-     * paths (from java.ext.dirs) 3) classpaths (set in CLASSPATH, or through
-     * -classpath and -cp) 4) paths accessible and examined by the classloader
+     * paths (from java.ext.dirs) 2) classpaths (set in CLASSPATH, or through
+     * -classpath and -cp) 3) paths accessible and examined by the classloader
      *
      * In each of these paths, the order of the paths as specified in the
      * command line options or environment variables will be the order returned
--- a/framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
+++ b/framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
@@ -371,8 +371,7 @@ public class AnnotationClassLoader {
      * environment variables, and by examining the classloader to see what paths
      * it has access to
      *
-     * The classpaths will be obtained in the order of: 1) boot classpaths (set
-     * using -bootclasspath -Xbootclasspath or -J-Xbootclasspath) 2) extension
+     * The classpaths will be obtained in the order of: 1) extension
      * paths (from java.ext.dirs) 3) classpaths (set in CLASSPATH, or through
      * -classpath and -cp) 4) paths accessible and examined by the classloader
      *
@@ -384,9 +383,6 @@ public class AnnotationClassLoader {
      */
     private final Set<String> getClasspaths() {
         Set<String> paths = new LinkedHashSet<String>();
-
-        // add all paths in Xbootclasspath
-        paths.addAll(Arrays.asList(System.getProperty("sun.boot.class.path").split(":")));
 
         // add all extension paths
         paths.addAll(Arrays.asList(System.getProperty("java.ext.dirs").split(":")));
--- a/framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
+++ b/framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
@@ -677,7 +677,7 @@ public class AnnotationClassLoader {
 
         try {
             // load the class
-            cls = Class.forName(fullyQualifiedClassName);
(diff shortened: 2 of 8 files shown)


jyluo:
> This will cause problems when loading an external annotation class outside of CheckerFramework. 
It is currently required for the external annotation classes' directory to be put on the bootstrap class path and the classpath, see https://github.com/typetools/checker-framework/blob/master/checker/examples/units-extension/Makefile

> external checker's ClassLoader
I believe that all checkers written on top of CheckerFramework is executed using a single ClassLoader, not multiple.
Also, in the first line in getAppClassLoader(), the reference variable `checker `is of type `BaseTypeChecker`. It would seems impossible for you to have a second class loader even if it is an externally defined checker (if it is a subtype of BaseTypeChecker):
`ClassLoader applicationClassLoader = checker.getClass().getClassLoader();`

If you're able to confirm that using getAppClassLoader() will shift this dependency onto -classpath then please also update the CF manual and all examples (UnitsExtension, SubtypingExtension, FenumExtension) in checker/examples which uses the externally loaded annotations mechanism.

CharlesZ-Chen:
> I believe that all checkers written on top of CheckerFramework is executed using a single ClassLoader, not multiple.

Yes, you're right. Thanks for pointing that out.

> It is currently required for the external annotation classes' directory to be put on the bootstrap class path

I tried one external Checker built directly on top of Checker Framework, and it seems put it's classes' dir (both checker class and qual classes) on the `-classpath` works: https://github.com/CharlesZ-Chen/ReadChecker/blob/master/read-check.sh

The problem I meet is when developing another external Infer Checker that build on top of Checker Framework Inference: When I put it's classes on the system env `CLASSPATH`, with the correct fully qualifier name, the `AnnotationClassLoader` still cannot load it's qualifier classes. After I made the change of this pull request, `AnnotationClassLoader` could load my qualifier classes.

I still don't clearly know the underlying reason that why qualifier classes of a type-checking external checker could be loaded correctly by `AnnotationClassLoader`, while qualifiers of an infer external checker cannot be loaded. However, I think the change in this pull request is reasonable, because it enables the `AnnotationClassLoader` load external qualifier classes through the same ClassLoader of their checker, or through system `classpath` when their checker is loaded by bootstrap ClassLoader.

jyluo:
> I tried one external Checker built directly on top of Checker Framework, and it seems put it's classes' dir (both checker class and qual classes) on the -classpath works: https://github.com/CharlesZ-Chen/ReadChecker/blob/master/read-check.sh

Are the qualifiers for the ReadChecker placed in qual/ ? If so it may have already been loaded as part of instantiating the ReadChecker.

The definition of "externally defined qualifiers" is actually any qualifier which 1) is not in qual/ and more often 2) defined by the users of your checker. All checkers, external to CF's collection or not, should have a qual/ directory, and those annotations would be automatically loaded without any additional code by the developer. UnitsExtension demo has an external qualifier called Hz to illustrate this usage. Only these kinds of qualifiers need to be visible (without accounting for your change) on the bootclass path.

Please verify the source of the problem with CF Inference implementation of the checker (but I don't see why it wouldn't load, as other Inference checkers have already switched successfully to use the Loader: Both https://github.com/typetools/checker-framework-inference/tree/master/src/trusted and https://github.com/typetools/checker-framework-inference/tree/master/src/interning directly extend from BaseAnnotatedTypeFactory, which does not override createSupportedTypeQualifiers(). In ATF's createSupportedTypeQualifiers() we call getBundledTypeQualifiersWithPolyAll() to actually perform the automatic loading. (See https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java).

Note that https://github.com/typetools/checker-framework-inference/blob/master/src/checkers/inference/InferenceAnnotatedTypeFactory.java, overrides createSupportedTypeQualifiers but does not call getBundledTypeQualifiersWithPolyAll. 

> change in this pull request is reasonable

I think it is a useful change but from a different point of view: It is most likely that the user would have their custom checkers or custom qualifiers already visible in -cp so the current requirement asks them to duplicate the effort once more on -Xbootclasspath. If we can eliminate the need for the user to verbosely put the path of externally defined qualifiers on -Xbootclasspath then we should remove this requirement from our documentation and verify it via the 3 demos I listed.

CharlesZ-Chen:
> Are the qualifiers for the ReadChecker placed in qual/ ?

Ah, yes, ReadChecker's qualifiers is following the CF convention. Thanks for clear the definition of "external qualifiers".

> Please verify the source of the problem with CF Inference implementation of the checker (but I don't see why it wouldn't load, as other Inference checkers have already switched successfully to use the Loader

I think the reason why `AnnotationClassLoader` works on TrustedChecker and InterningChecker is because these two checkers are located inside the CF Inference project, which means `InferenceLauncher` will put them in the -Xbootclasspath when invoking the `InferenceMain`: 
https://github.com/typetools/checker-framework-inference/blob/master/src/checkers/inference/InferenceLauncher.java#L158

However, the Inference Checker I currently works on is outside of CF Inference, which means the `InferenceLauncher#getInferenceRuntimeBootclassPath()` won't attach my checker's class files on the -Xbootclasspath for `InferenceMain`, and thus the `AnnotationClassLoader` cannot load my qualifier classes because they are not on the -Xbootclasspath.

Currently I can put my checker's class files on `--targetclasspath` option, in order to let `InferenceLauncher` put them on the -Xbootclasspath, but I don't think this is a nice way to do it.

Could we merge this pull request? This could enable an external Inference Checker outside of CF Inference simply declare it's classfiles of qualifiers on `-classpath` ( which is an intuitive way to do it), and the Loader will works fine in this situation.

jyluo:
> This could enable an external Inference Checker outside of CF Inference simply declare it's classfiles of qualifiers on -classpath ( which is an intuitive way to do it), and the Loader will works fine in this situation.

I support this. But I think as a follow up, we should completely remove the current dependency of putting things on -Xbootclasspath.

smillst:
> If you're able to confirm that using getAppClassLoader() will shift this dependency onto -classpath then please also update the CF manual and all examples (UnitsExtension, SubtypingExtension, FenumExtension) in checker/examples which uses the externally loaded annotations mechanism.

This should be done before this pull request is merged.  Thanks!

CharlesZ-Chen:
@jyluo @smillst Could you tell me where are the locations in CF manual and CF examples that I should change according to this pull request? 

The locations I found in CF manual that might need to be changed are:

1. [Running Unit Checker](https://github.com/typetools/checker-framework/blob/master/checker/manual/units-checker.tex#L239)

2. [Running Fenum Checker](https://github.com/typetools/checker-framework/blob/master/checker/manual/fenum-checker.tex#L116)

3. [Using the Subtyping Checker] (https://github.com/typetools/checker-framework/blob/master/checker/manual/subtyping-checker.tex#L33)

And for examples need to be changed are:

1. [Makefile of fenum-extension](https://github.com/typetools/checker-framework/blob/master/checker/examples/fenum-extension/Makefile)

2. [Makefile of Unit-extension](https://github.com/typetools/checker-framework/blob/master/checker/examples/units-extension/Makefile)

3. [Makefile of Subtyping-extension](https://github.com/typetools/checker-framework/blob/master/checker/examples/subtyping-extension/Makefile)

Am I missing other places that might need also to be changed? Thanks!

CharlesZ-Chen:
@smillst @jyluo  I've already change those examples and CF manual I found.

jyluo:
@CharlesZ-Chen

in CF Manual:
https://github.com/typetools/checker-framework/blob/master/checker/manual/subtyping-checker.tex#L215
And additional locations: do a search for -Xbootclasspath on http://types.cs.washington.edu/checker-framework/current/checker-framework-manual.html

in Javadoc and comments:
grep search Xbootclasspath in all source files. In particular https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java

code change:
Remove this line: https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java#L389  and ensure everything still works.

Make sure the 3 extension examples work, and make sure any additional changes to the CF manual is tested as well in code. Eg maven integration, gradle, android plugin, intelliJ, ...

CharlesZ-Chen:
> Make sure the 3 extension examples work

I've already tested the 3 extension examples, all of them works fine with the changed code.

>  and make sure any additional changes to the CF manual is tested as well in code. Eg maven integration, gradle, android plugin, intelliJ, ...

Do we have automatic test suites to ensure those properties still holds? It's seems impossible for a developer to ensure them by manually testing them one by one. Just takes too much time and effort.

CharlesZ-Chen:
> And additional locations: do a search for -Xbootclasspath on http://types.cs.washington.edu/checker-framework/current/checker-framework-manual.html

Already done the searching. Find one more related location and already changed it.

> in Javadoc and comments:
grep search Xbootclasspath in all source files. 

I found one more related Xbootclasspath in framework/tests/annotationclassloader/Makefile. I've already changed and tested it.

> code change:
Remove this line: https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java#L389 and ensure everything still works.

I've already remove it, tested it with 3 extension examples, test Makefile of annotationclassloader under framework/tests, and also test it with my own external Checkers  build on top of CF and CF Inference.

CharlesZ-Chen:
Note: I leave this Xbootclasspath related in AnnotationClassLoader unchanged:
https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java#L437

Because I think for debugging we always want more informations could be available.

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
```diff
@@ -371,8 +371,7 @@ private final boolean checkDirForPackage(final File currentDir, final Iterator<S
      * environment variables, and by examining the classloader to see what paths
      * it has access to
      *
-     * The classpaths will be obtained in the order of: 1) boot classpaths (set
-     * using -bootclasspath -Xbootclasspath or -J-Xbootclasspath) 2) extension
+     * The classpaths will be obtained in the order of: 1) extension
      * paths (from java.ext.dirs) 3) classpaths (set in CLASSPATH, or through
```
jyluo: fix itemization

jyluo:
Pls also update the number itemization here: https://github.com/typetools/checker-framework/pull/815/commits/747bc3a2afe0f82478b10e005e1489cbff936ccb#diff-7a35ec4ec69bf487ddc7b240565d38f4R374 

And, I think the -J-Xbootclasspath args in annotationclassloader in framework/tests may also be removable by shifting them to -classpath. Let me know if it works.

Otherwise, looks good to me!

CharlesZ-Chen:
@jyluo @smillst Already fix the itemization.

> And, I think the -J-Xbootclasspath args in annotationclassloader in framework/tests may also be removable by shifting them to -classpath. Let me know if it works.

Yes, I shifted them all to -classpath, and it works.

CharlesZ-Chen:
@jyluo 

> And, I think the -J-Xbootclasspath args in annotationclassloader in framework/tests may also be removable by shifting

The travis test failed under Java7. Thus I reversed my changes.  Currently I still shift only the $(PROJECT_DIR) to -classpath.

checker/manual/fenum-checker.tex
```diff
@@ -123,23 +123,17 @@ \section{Running the Fenum Checker\label{fenum-running}}
...
 \begin{alltt}
-  javac -Xbootclasspath/p:\textit{/full/path/to/myProject/bin}:\textit{/full/path/to/myLibrary/bin} \ttbs
+  javac -classpath \textit{/full/path/to/myProject/bin}:\textit{/full/path/to/myLibrary/bin} \ttbs
         -processor org.checkerframework.checker.fenum.FenumChecker \ttbs
         -Aquals=\textit{myModule.qual.MyFenum} MyFile.java ...
 \end{alltt}
 
-The annotations listed in \code{-Aquals} must be accessible to
-the compiler during compilation in the classpath.  In other words, they must
-already be compiled (and, typically, be on the javac bootclasspath)
```
smillst: This paragraph is still valid only "bootclasspath" on this line should just be classpath

checker/manual/fenum-checker.tex
```diff
@@ -123,23 +123,23 @@ \section{Running the Fenum Checker\label{fenum-running}}
...
         -processor org.checkerframework.checker.fenum.FenumChecker \ttbs
         -Aquals=\textit{myModule.qual.MyFenum} MyFile.java ...
 \end{alltt}
 
 The annotations listed in \code{-Aquals} must be accessible to
 the compiler during compilation in the classpath.  In other words, they must
-already be compiled (and, typically, be on the javac bootclasspath)
-before you run the Fenum Checker with \code{javac}.  It
+already be compiled (and, typically, be on the javac classpath)
+before you run the Subtyping Checker with \code{javac}.  It
```
smillst: This should be Fenum, not Subtyping.

CharlesZ-Chen: Already fix it. Sorry~

CharlesZ-Chen: Already fixed it. Sorry~

smillst:
Thanks for updating the test and manual!  This looks good to me.   I'll squash and merge once Travis passes.

CharlesZ-Chen:
Thanks a lot! I do a final check and fix some typo.

CharlesZ-Chen:
@jyluo  You're right, lazy always get it's price, haha:)

Thanks a lot for your help towards this pull request!

PR #821, opened 2016-06-29
Correct TypeFromMemberVisitor#visitVariable

smillst:
This is a fix for an issue reported by @mcoblenz in PR #819


framework/src/org/checkerframework/framework/type/TypeFromMemberVisitor.java
```diff
@@ -27,8 +27,7 @@
     @Override
     public AnnotatedTypeMirror visitVariable(VariableTree node,
                                              AnnotatedTypeFactory f) {
-        AnnotatedTypeMirror result = f.fromTypeTree(node.getType());
-        result.clearAnnotations();
+        AnnotatedTypeMirror result = f.type(node.getType());
```
wmdietl: I think this is one of the few uses of `fromTypeTree`. I think there is some complication why you need to know whether the tree is a type tree or not.

PR #825, opened 2016-07-01
Avoid infinite recursion (Fixes #810)

### LINKED ISSUE #810: StackOverflowError from TypeForExpressionVisitor loop?
Save as C.java:
```java
class C {
    java.util.Map m = new java.util.HashMap();
    java.util.Set n = m.keySet();
}
```
See [gist](https://gist.github.com/dbrosoft/62d09371c5046134d32bdf5d9350e6ac) for full output, which begins with
```
error: SourceChecker.typeProcess: unexpected Throwable (StackOverflowError) while processing C.java
  Compilation unit: C.java
  Exception: java.lang.StackOverflowError; Stack trace: java.util.HashMap$ValueIterator.<init>(HashMap.java:1456)
```
Note that there is a 38-node cycle starting at line 33 and continuing to the end.

smillst:
Fixes #810


mernst:
This looks good to me.  When the tests pass, I'm OK with merging it.  (I guess that will happen after the release.)

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -942,10 +942,18 @@ public AnnotatedTypeMirror getAnnotatedType(Tree tree) {
             return treeCache.get(tree).deepCopy();
         }
 
-        AnnotatedTypeMirror type;
         if (TreeUtils.isClassTree(tree)) {
-            type = fromClass((ClassTree)tree);
-        } else if (tree.getKind() == Tree.Kind.METHOD ||
+            ClassTree classTree = (ClassTree) tree;
+            AnnotatedTypeMirror type = getAnnotatedTypeOfClassTree(classTree);
```
wmdietl: This method also checks whether the tree is in the cache, which already happened on line 941.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -942,10 +942,18 @@ public AnnotatedTypeMirror getAnnotatedType(Tree tree) {
...
+            if (shouldCache) {
```
wmdietl: Why is this not part of `getAnnotatedTypeOfClassTree`? That method checks the cache, so shouldn't it also update it?

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1699,6 +1702,27 @@ public AnnotatedDeclaredType getSelfType(Tree tree) {
     }
 
     /**
+     * Get the AnnotatedDeclaredType for a classTree without kicking off dataflow or calling
```
wmdietl: Isn't this the exact same sequence of calls as before?
How is dataflow not kicked off?

smillst: No, postProcessClassTree is not called when this method is called directly. Dataflow is only kicked off the first time addComputedAnnotations is called.  So, dataflow still might be started by this method, so I removed the references to dataflow from these comments.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
smillst: Because if the type is cached when getAnnotatedTypeOfClassTree is called from a different method, then postProcessClassTree would not be called for that tree.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
smillst: Yes, because that getAnnotatedTypeOfClassTree is called in other locations where the cache is not checked before the call.

wmdietl: You could move the handling of ClassTrees before the existing check for the cache, move updating of the cache into `getAnnotatedTypeOfClassTree`, and still ensure that postprocess is called if needed.

smillst: How would I check if postProcessClassTree needs to be called?

smillst:
Closed in favor of #833

PR #830, opened 2016-07-05
Ternary expressions with null and primitive (Fixes #829)

### LINKED ISSUE #829: Ternary expressions with null fail
The following program is no longer considered valid under v2.1.0:

```java
import javax.annotation.Nullable;

public class Checker {
  @Nullable
  public static Double getDouble(boolean flag) {
    return flag ? null : 1.0;
  }
}
```

This works under v2.0.0. The error message is:

```
Checker.java:8: error: [type.invalid] [@Initialized, @Nullable] may not be applied to the type "@Initialized @Nullable double"
    return flag ? null : 1.0;
```

smillst:
Fixes #829

### DIFF
--- a/checker/tests/nullness/Issue829.java
+++ b/checker/tests/nullness/Issue829.java
@@ -1,6 +1,5 @@
 // Test case for Issue 829
 // https://github.com/typetools/checker-framework/issues/829
-// @skip-test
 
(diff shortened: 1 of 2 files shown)


mernst:
Looks good to me, thanks.

PR #831, opened 2016-07-06
Fix NullPointerException in `DOTCFGVisualizer#visualizeBlockNode()` when analysis is null

wmdietl:
See review here:
https://github.com/opprop/checker-framework/pull/2

### DIFF
--- a/dataflow/src/org/checkerframework/dataflow/cfg/DOTCFGVisualizer.java
+++ b/dataflow/src/org/checkerframework/dataflow/cfg/DOTCFGVisualizer.java
@@ -186,7 +186,8 @@ public class DOTCFGVisualizer<A extends AbstractValue<A>,
         return this.sbDigraph.toString();
     }
 
-    protected void generateDotNodes(Set<Block> visited, ControlFlowGraph cfg, Analysis<A, S, T> analysis) {
+    protected void generateDotNodes(Set<Block> visited, ControlFlowGraph cfg,
+            /*@Nullable*/ Analysis<A, S, T> analysis) {
         IdentityHashMap<Block, List<Integer>> processOrder = getProcessOrder(cfg);
         this.sbDigraph.append("    node [shape=rectangle];\n\n");
         // definition of all nodes including their labels
@@ -347,6 +348,8 @@ public class DOTCFGVisualizer<A extends AbstractValue<A>,
 
     @Override
     public void visualizeBlockTransferInput(Block bb, Analysis<A, S, T> analysis) {
+        assert analysis != null : "analysis should be non-null when visualize transfer input of a block";
+
         TransferInput<A, S> input = analysis.getInput(bb);
         this.sbStore.setLength(0);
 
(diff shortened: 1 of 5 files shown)


dataflow/src/org/checkerframework/dataflow/cfg/DOTCFGVisualizer.java
```diff
@@ -347,6 +348,8 @@ public void visualizeSpecialBlock(SpecialBlock sbb) {
 
     @Override
     public void visualizeBlockTransferInput(Block bb, Analysis<A, S, T> analysis) {
+        assert analysis != null : "analysis should be non-null when visualize transfer input of a block";
```
mernst: The text in the string is not grammatical

dataflow/src/org/checkerframework/dataflow/cfg/DOTCFGVisualizer.java
```diff
@@ -397,12 +400,14 @@ public void visualizeBlockTransferInput(Block bb, Analysis<A, S, T> analysis) {
 
     @Override
     public void visualizeBlockNode(Node t, /*@Nullable*/ Analysis<A, S, T> analysis) {
-        A value = analysis.getValue(t);
-        String valueInfo = "";
-        if (value != null) {
-            valueInfo = "    > " + prepareString(value.toString());
+        this.sbBlock.append(prepareString(t.toString()) + "   [ " + prepareNodeType(t) + " ]");
+        A value = null;
```
mernst: If the variable is only used within the if statement, then it should be declared within the if statement, too, in the smallest possible enclosing scope.

dataflow/src/org/checkerframework/dataflow/cfg/DOTCFGVisualizer.java
wmdietl: How about
`analysis should be non-null when visualizing the transfer input of a block`

dataflow/src/org/checkerframework/dataflow/cfg/DOTCFGVisualizer.java
wmdietl: Agreed.

CharlesZ-Chen:
Already change the code comment and the declaration scope of local variable `v`.

dataflow/src/org/checkerframework/dataflow/cfg/DOTCFGVisualizer.java
```diff
@@ -347,6 +348,9 @@ public void visualizeSpecialBlock(SpecialBlock sbb) {
 
     @Override
     public void visualizeBlockTransferInput(Block bb, Analysis<A, S, T> analysis) {
+        assert analysis != null :
+            "analysis should be non-null when visualizeBlockTransferInput() get called.";
```
wmdietl: `analysis should be non-null when visualizeBlockTransferInput() is called.`
`analysis should be non-null when visualizing the transfer input of a block`

CharlesZ-Chen:
Already fixed it.

dataflow/src/org/checkerframework/dataflow/cfg/DOTCFGVisualizer.java
```diff
@@ -397,12 +400,14 @@ public void visualizeBlockTransferInput(Block bb, Analysis<A, S, T> analysis) {
...
     @Override
     public void visualizeBlockNode(Node t, /*@Nullable*/ Analysis<A, S, T> analysis) {
-        A value = analysis.getValue(t);
-        String valueInfo = "";
-        if (value != null) {
-            valueInfo = "    > " + prepareString(value.toString());
+        this.sbBlock.append(prepareString(t.toString()) + "   [ " + prepareNodeType(t) + " ]");
+        if (analysis != null) {
+            A value = null;
+            value = analysis.getValue(t);
```
mernst: Combine this line with the previous one.  If you were going to put them on separate lines, there would be no point in initializing `value` in its declaration, because it will be immediately overwritten.  It's even clearer to assign `value` to its real value at its declaration.

wmdietl: Thanks. I had the same comment https://github.com/typetools/checker-framework/commit/2b100092d8cc2b339f7a93c6cedbd11267cae80e#commitcomment-18151499 but unfortunately that was buried within a commit.

CharlesZ-Chen:
Thanks, already merged to one line.

Issue #684, opened 2016-04-27
org.checkerframework.framework should be independent of org.checkerframework.common


wmdietl:
The framework package should be independent of the common package. However, there are currently imports of the latter into the former, breaking the intended separation between the two packages.
This issue can be resolved by either breaking the dependency (preferred) or moving the dependency into the framework package.

For an under-approximation, in checker-framework/framework/src/org/checkerframework/framework run:
`grep -R "import org.checkerframework.common" .`
to get something like:

```
./type/GeneralAnnotatedTypeFactory.java:import org.checkerframework.common.basetype.BaseTypeChecker;
./type/AnnotatedTypeFactory.java:import org.checkerframework.common.basetype.BaseTypeChecker;
./type/AnnotatedTypeFactory.java:import org.checkerframework.common.reflection.DefaultReflectionResolver;
./type/AnnotatedTypeFactory.java:import org.checkerframework.common.reflection.MethodValAnnotatedTypeFactory;
./type/AnnotatedTypeFactory.java:import org.checkerframework.common.reflection.MethodValChecker;
./type/AnnotatedTypeFactory.java:import org.checkerframework.common.reflection.ReflectionResolver;
./type/AnnotationClassLoader.java:import org.checkerframework.common.basetype.BaseTypeChecker;
./type/GenericAnnotatedTypeFactory.java:import org.checkerframework.common.basetype.BaseTypeChecker;
./type/DefaultTypeHierarchy.java:import org.checkerframework.common.basetype.BaseTypeChecker;
./util/DependentTypes.java:import org.checkerframework.common.basetype.BaseTypeChecker;
./util/PurityAnnotatedTypeFactory.java:import org.checkerframework.common.basetype.BaseAnnotatedTypeFactory;
./util/PurityAnnotatedTypeFactory.java:import org.checkerframework.common.basetype.BaseTypeChecker;
./util/PurityChecker.java:import org.checkerframework.common.basetype.BaseTypeChecker;
./source/SourceChecker.java:import org.checkerframework.common.basetype.BaseTypeChecker;
./flow/CFAbstractAnalysis.java:import org.checkerframework.common.basetype.BaseTypeChecker;
./flow/CFCFGBuilder.java:import org.checkerframework.common.basetype.BaseTypeChecker;
./flow/CFAbstractTransfer.java:import org.checkerframework.common.basetype.BaseTypeChecker;
./flow/CFAnalysis.java:import org.checkerframework.common.basetype.BaseTypeChecker;
```

PR #833, opened 2016-07-07
Move postProcessClassTree to BaseTypeVisitor (Fixes #810)

### LINKED ISSUE #810: StackOverflowError from TypeForExpressionVisitor loop?

smillst:
Move the call that kicks off dataflow from addCommputedTypeAnnotations to BaseTypeVisitor.visitClass and call postProcessClassTree to the end of visitClass.
Fixes #810

Also, includes a bug fix uncovered by this change.


wmdietl:
I like this very much, having the pre/postProcessClassTree methods is much nicer.
The one thing I was struggling with is that this adds yet another dependency on BaseTypeVisitor to a class that is supposedly stand-alone.
When I investigated this for a bit, I found that `org.checkerframework.common.basetype.BaseTypeVisitor.setRoot(CompilationUnitTree)` is the only place that calls `AnnotatedTypeFactory.setRoot`, which is another prerequisite to a successfull ATF execution.
So the additional dependency for preProcessClassTree doesn't make this dependency worse.
This is related to Issue #684 which we should eventually tackle by a proper redesign of where all these classes are.
So let's merge this and see when we get to #684 .

PR #836, opened 2016-07-12
Refactor AnnotatedTypeFactory: Use lazy initialization in getSupportedTypeQualifiers()

CharlesZ-Chen:
`AnnotatedTypeFactory#createSupportedTypeQualifiers()` is an overridable method, and should avoid using it in constructor.

Currently using it in constructor would cause NullPointer Exception in `InferenceAnnotatedTypeFactory#createSupportedTypeQualifiers()` if `InferenceAnnotatedTypeFactory` doesn't use the deprecated method `InferenceMain#getRealTypeFactory()`.

As a solution, I change code to create supported type qualifiers by using lazy initialization, i.e. creating an empty set for `supportedQuals` during constructor, and creating the real supported qualifiers when `AnnotatedTypeFactory#getSupportedTypeQualifiers()` get called.


framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -336,8 +336,7 @@ public AnnotatedTypeFactory(BaseTypeChecker checker) {
         this.visitorState = new VisitorState();
 
         this.loader = new AnnotationClassLoader(checker);
-        this.supportedQuals = createSupportedTypeQualifiers();
-        checkSupportedQuals();
+        this.supportedQuals = new HashSet<>();
```
jyluo: Shouldn't leave HashSet as a raw type.

Change all of these to HashSet<Class<? extends Annotation>>

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -894,7 +892,11 @@ public AnnotationFormatter getAnnotationFormatter() {
      *         empty set if no qualifiers are supported
      */
     public final Set<Class<? extends Annotation>> getSupportedTypeQualifiers() {
-        return supportedQuals;
+        if (this.supportedQuals.isEmpty()) {
+            supportedQuals.addAll(createSupportedTypeQualifiers());
+            checkSupportedQuals();
+        }
+        return Collections.unmodifiableSet(supportedQuals);
```
jyluo: suggest you do `supportedQuals = Collections.unmodifiableSet(supportedQuals)` then `return supportedQuals`. The set of qualifiers shouldn't change after initializing the ATF.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -894,7 +892,11 @@ public AnnotationFormatter getAnnotationFormatter() {
      *         empty set if no qualifiers are supported
      */
     public final Set<Class<? extends Annotation>> getSupportedTypeQualifiers() {
-        return supportedQuals;
+        if (this.supportedQuals.isEmpty()) {
+            supportedQuals.addAll(createSupportedTypeQualifiers());
```
jyluo: `addAll()` reads whatever is returned by `createSupportedTypeQualifiers()`, essentially copying all of the annotation class objects into `supportedQuals`. I don't see why you've eliminated the contract that `createSupportedTypeQualifiers` returns an **immutable** set of qualifiers. Keeping the contract ensures that even during the creation of this set, we eliminate any chances of unwanted modification of the set of supported qualifiers.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
CharlesZ-Chen: Thanks a lot, I will change this.

CharlesZ-Chen: Thanks a lot, I will change this.

jyluo: You'll also have to remove `final` from the declaration of `supportedQuals`...

The design before your changes was: 1) use `final` to ensure that `supportedQuals` references only 1 copy of a set, and 2) that set is immutable. Together it strongly ensures that the qualifiers supported by an ATF is unchangeable after initializing the ATF.

Now it is weaker; without `final`, the reference _could_ be updated to point to any other set. But since it is a private field this will probably be fine.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
CharlesZ-Chen: I modified the contract because supportedQuals` will immediately becomes immutable after calling `createSupportedQualifiers()`. This is ensured by the lazy initialization in `getSupportedQualifiers()`. Thus, required `createSupportedQualifiers()` also return an immutable set seems duplicate.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
CharlesZ-Chen: Yes, I've already change `supprotedQuals` a non-final field.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
jyluo: If you look through the source code of `Collections.unmodifieableSet(s)`, all it does is return an `UnmodifiableSet<E>` which contains an internal reference to `s`, and where it forwards getter calls while throwing exceptions for setter calls. The computation cost is an extra object that has 1 reference inside. The gain is a strong guarantee of no modification of any set that comes from `createSupportedTypeQualifiers`.

`createSupportedTypeQualifiers` is a protected method that subclasses can call directly, and making it go from immutable to mutable won't break type checkers written by the community. The decision is up to you :)

http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/7u40-b43/java/util/Collections.java#Collections.unmodifiableSet%28java.util.Set%29

CharlesZ-Chen: The origin contract requires sub-classes to return an immutable set in `javadoc`, but it cannot prevent the subclass actually return a mutable set, like this implementation in current Checker Framework Inference master branch:
https://github.com/typetools/checker-framework-inference/blob/master/src/nninf/NninfAnnotatedTypeFactory.java#L58

Thus, I think remove the requirement of return immutable set from subclasses, and let final method `getSupportedQualifiers()` always return an immutable view of `supportedQualifiers` would be safer.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
CharlesZ-Chen: > Yes, I've already change `supprotedQuals` a non-final field.

I changed my mind. To using lazy initialisation on `supportedQuals`, there is a tradeoff:

1) makes `supportedQuals` a non-final field, which means we lose the reference immutability of `supportedQuals`, then we let `getSupportedQualifiers()` assign `supportedQuals` to an immutable set at each time it gets called:

``` java
supportedQuals = Collections.unmodifiableSet(supportedQuals);
return supportedQuals;
```

2) keep `supportedQuals` a final field, but now `supportedQuals` is a mutable set, thus `getSupportedQualifiers()` should ensure that it always return an immutable view of `supportedQuals` to the outside of `AnnotatedTypeFactory`:
```java
return Collections.unmodifiableSet(supportedQuals);
```

I prefer the latter one, since although we loose the `supportedQuals` to a mutable set, we still make sure it always refer to one same set.

In former decision, we can reset `supportedQuals` refer to a mutable set, and do whatever we want to `supportedQuals`.

jyluo: My suggestion for 1) is put supportedQuals = Collections.unmodifiableSet(supportedQuals); inside your if statement.

either way, modifications to the reference or to the set can only happen in ATF's code. Future code review will have to catch it.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
jyluo: An `assert supportedQuals.class == UnmodifiableSet.class` should have been there immediately after assigning to supportedQuals, which would have enforced this contract. Oh well!

CharlesZ-Chen: Nice solution!

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
CharlesZ-Chen: Agreed, no mater which decision we choose, further use of `supportedQuals` have to be careful and code review have to catch abnormal use of `supportedQuals`. Using 1) and only convert `supportedQuals` to immutable set when creating it by `createSupportedQualifiers()` seems good.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
wmdietl: Jeff, this is not a raw type. It is better style to use diamonds and not give the explicit type arguments.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
wmdietl: We had discussed this issue and decided to stay with a final field.

PR #841, opened 2016-07-13
Compile test Java files in the same directory together

smillst:
In order to speed up the JUnit tests, compile all test files in the same directory together. 

Documented in [checker/tests/README]( https://github.com/typetools/checker-framework/pull/841/files#diff-0884e120f5091b7a82489ce731af8ac8)
And also in the [CheckerFrameworkPerDirectoryTest JavaDoc](https://github.com/typetools/checker-framework/pull/841/files#diff-ab1b76c1119533b1d9b636a67f3967ff)


smillst:
Checker Framework Inference will require a minor change:
typetools/checker-framework-inference#38

framework/tests/framework/README
```diff
@@ -1,9 +1 @@
-To add a new file to the test suite, see
-  ../README
-
-To run the tests, do
-  ant -e -find build.xml framework-tests
-
-To run a single test, do something like:
-  cd $CHECKERFRAMEWORK/framework/tests/framework
-  (cd $CHECKERFRAMEWORK && ant -e dist) && javac -processor tests.util.TestChecker -classpath $CHECKERFRAMEWORK/framework/tests/build GenericTest10.java
+Tests in this directory may issue java errors.
```
mernst: Why was the previous README overwritten?  Should the new information have been appended to the old rather than replacing it?

smillst: I corrected the instructions and restored them.

framework/src/org/checkerframework/framework/test/TestUtilities.java
```diff
@@ -57,6 +57,39 @@
         return getJavaFilesAsArgumentList(dirs);
     }
 
+    public static List<List<File>> findJavaFilesPerDirectory(File parent, String... dirNames) {
```
mernst: A line or two of Javadoc for this new method and the one immediately below would be helpful.

mernst:
This looks fine to me -- when my few comments are addressed, it should be morged.
(The changes look a lot bigger on GitHub than they actually are.)

PR #843, opened 2016-07-15
Fixes crash when wildcard extend interfaces

smillst:
Given the following declaration:
````
class MyClass<T extends Number> { }
```
This type use is legal:
````
MyClass<? extends Serializable>
````
Because the wildcard is capture converted to `CAP#1 extends Number & Serializable`. (See [JSL 5.1.10](https://docs.oracle.com/javase/specs/jls/se8/html/jls-5.html#jls-5.1.10).)  The Checker Framework was crashing on these wildcards; this pull request fixes the crash


mernst:
Looks OK to me.

wmdietl:
Looks good.

PR #848, opened 2016-07-19
Refactor AnnotatedTypeFactory: Use lazy initialization in getSupportedTypeQualifiers()

CharlesZ-Chen:
`AnnotatedTypeFactory#createSupportedTypeQualifiers()` is an overridable method, and should avoid using it in constructor.

Currently using it in constructor would cause NullPointer Exception in `InferenceAnnotatedTypeFactory#createSupportedTypeQualifiers()` if `InferenceAnnotatedTypeFactory` doesn't use the deprecated method `InferenceMain#getRealTypeFactory()`.

As a solution, I change code to create supported type qualifiers by using lazy initialization, i.e. creating an empty set for `supportedQuals` during constructor, and creating the real supported qualifiers when `AnnotatedTypeFactory#getSupportedTypeQualifiers()` get called.

This PR is created after another round of code check.


framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -755,22 +755,21 @@ public TypeArgumentInference getTypeArgumentInference() {
...
      * </pre>
      *
      * The set of qualifiers returned by
-     * {@link #createSupportedTypeQualifiers()} must be an immutable
+     * {@link #createSupportedTypeQualifiers()} must be a fresh, mutable
      * set. The methods
      * {@link #getBundledTypeQualifiersWithoutPolyAll(Class...)} and
      * {@link #getBundledTypeQualifiersWithPolyAll(Class...)} each
-     * return an immutable set.
+     * must return a fresh, mutable set
```
smillst: Why does it have to be fresh and mutable?  The call to this method in getSupportedTypeQualifiers does not modify the returned set.  (I think the changes removing calls to Collections.unmodifiableSet(...) are good, I would just not require createSupportedTypeQualifiers to return a mutable set.)

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -894,7 +892,11 @@ public AnnotationFormatter getAnnotationFormatter() {
      *         empty set if no qualifiers are supported
      */
     public final Set<Class<? extends Annotation>> getSupportedTypeQualifiers() {
```
smillst: Please add a comment to the declaration of `supportedQuals` stating that getSupportedTypeQualifiers should be used instead of use supportedQuals directly.

smillst:
Would you please merge master into this branch so that Travis will run the tests on this branch? (Commit e67713c71bcec299949c65c58b1da19229b6057a fixed a bug in the Travis configuration.)

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
wmdietl: We wanted to allow that someone does

```
fromsuper = super.createSupportedTypeQualifiers();
fromsuper.add(MyAddition.class);
```

For that we need the guarantee that we can modify the object and that the modifications don't impact others.

smillst:
I don't quite understand the reason for this change. 
>AnnotatedTypeFactory#createSupportedTypeQualifiers() is an overridable method, and should avoid using it in constructor.

 `createSupportedTypeQualifiers` is not the only overridable method called in the AnnotatedTypeFactory constructor, for example, `createAnnotatedTypeFormatter` is called, too.

>Currently using it in constructor would cause NullPointer Exception in InferenceAnnotatedTypeFactory#createSupportedTypeQualifiers() if InferenceAnnotatedTypeFactory doesn't use the deprecated method InferenceMain#getRealTypeFactory().

The comment on InferenceMain#getRealTypeFactory() says that is is not really deprecated, but marked deprecated to discourage its use.

smillst:
Commits 	3140f33 and 046d753 are unrelated to this pull request and should be removed.

wmdietl:
I've pushed two changesets to opprop/master, which unfortunately now also shows up in this PR.
One fixes the bin-devel/javac script: https://github.com/typetools/checker-framework/pull/848/commits/3140f3390620719d9f9dc8555d4a1cc9ebe6118c
The other one is just a cleanup: https://github.com/typetools/checker-framework/pull/848/commits/046d7536cccf78972db28b0636705ff8b0ad532f

wmdietl:
This change is to remove the discouraged uses and instead implement the functionality cleanly.
Similar changes would be good for other methods - calling instance methods in constructors can lead to problems. We fixed the one problem we're running into.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
smillst: That makes sense.  Thanks for the explanation.

smillst:
Is this pull request is fixing the problem Jonathan described in the commit:
56c7a551d9eb0875cda2a4d8a6e6dfb747b625cf

wmdietl:
Yes, this allows us to use the realATF: https://github.com/opprop/checker-framework-inference/commit/684393abcf92015b884d772b26110e60c19cb198

smillst:
That makes sense.  Thanks for showing me the use.  Once 
https://github.com/typetools/checker-framework/pull/848#discussion_r71437053 and https://github.com/typetools/checker-framework/pull/848#issuecomment-233790513 are addressed, and the tests pass; I think this can be merged.

PR #859, opened 2016-07-23
Run some downstream tests in Travis, so they are tested for pull requests

mernst:


mernst:
I think this can be squashed and merged.
The sorting of filenames is a bit random to include here, sorry (but was needed during debugging and is still useful and is only in a couple of files at the end).

wmdietl:
Looks good.

PR #861, opened 2016-07-25
Run some downstream tests in Travis, so they are tested for pull requests

smillst:
The original pull request (#859) was merged, but then it was reverted (#860) because the test don't seem to be running.  For example the JDK7 JUnit test only took 1 min 4 sec.  Also, the tests in pull request #754 [fail](https://travis-ci.org/typetools/checker-framework/builds/147208752) on the branch, but [pass](https://travis-ci.org/typetools/checker-framework/builds/147208769) when merged to master.  Revert this change until we can fix this.


.travis-build.sh
```diff
@@ -23,42 +34,56 @@ export SHELLOPTS
...
   (cd checker && ant junit-tests-nojtreg-nobuild)
 fi
 
-if [[ "$1" != "junit" && "$1" != "misc" && "$1" != "downstream" ]]; then
+if [[ "${GROUP}" == "nonjunit" || "${GROUP}" == "all" ]]; then
   (cd checker && ant nonjunit-tests-nojtreg-nobuild jtreg-tests)
 
   # It's cheaper to run the demos test here than to trigger the
   # checker-framework-demos job, which has to build the whole Checker Framework.
   (cd checker && ant check-demos)
```
smillst: Demo's are also being run in the downstream group

wmdietl:
I'm for merging this.
The last execution in: https://travis-ci.org/typetools/checker-framework/builds/165709153 didn't seem to do anything extra for the downstream tests.
So after merging we should make sure that pull requests actually execute all the right downstream tests.

mernst:
The build you linked to didn't do anything downstream because it was a pull request, so trigger-travis.sh wasn't invoked.

wmdietl:
Isn't trigger-travis only used for the external downstream builds? It makes sense to me that we don't trigger external builds, as they wouldn't observe the merged repository from the pull request.
However, for the downstream targets that we build within the current build, running them for a pull request would make sense and give useful information. We would see whether something downstream breaks before deciding to merge the PR.

mernst:
Right, that is what is supposed to happen, and what I think is happening now.

Issue #590, opened 2016-02-10
Annotations on type parameters in stub files are sometimes ignored


atomicknight:
Environment: Checker Framework 1.9.11, Oracle JDK 1.8.0_65, Windows 7

When working with generic methods, type annotations on type parameters in stub files seem to be ignored under certain circumstances.

Here is a test case: https://github.com/atomicknight/cf-testcases/tree/master/methodgenerics1. The test case executes two compilations, one without stub files and one with a stub file for `java.util.concurrent.ExecutorService`. The first compilation returns these errors:
```
d:\workspace\cf-testcases\methodgenerics1\src\main\java\org\example\ExecutorServiceDecorator.java:[53,43] error: [override.return.invalid] <T/*DECL*/  extends @Initialized @Nullable Object> @Initialized @NonNull Future<T extends @Initialized @Nullable Object> submit(@Initialized @NonNull ExecutorServiceDecorator this, @Initialized @NonNull Callable<T extends @Initialized @Nullable Object> p0) in org.example.ExecutorServiceDecorator cannot override <T/*DECL*/  extends @Initialized @NonNull Object> @Initialized @NonNull Future<T extends @Initialized @NonNull Object> submit(@Initialized @NonNull ExecutorService this, @Initialized @NonNull Callable<T extends @Initialized @NonNull Object> p0) in java.util.concurrent.ExecutorService; attempting to use an incompatible return type
  found   : @Initialized @NonNull Future<T extends @Initialized @Nullable Object>
  required: @Initialized @NonNull Future<T extends @Initialized @NonNull Object>

d:\workspace\cf-testcases\methodgenerics1\src\main\java\org\example\ExecutorServiceDecorator.java:[53,66] error: [override.param.invalid] <T/*DECL*/  extends @Initialized @Nullable Object> @Initialized @NonNull Future<T extends @Initialized @Nullable Object> submit(@Initialized @NonNull ExecutorServiceDecorator this, @Initialized @NonNull Callable<T extends @Initialized @Nullable Object> p0) in org.example.ExecutorServiceDecorator cannot override <T/*DECL*/  extends @Initialized @NonNull Object> @Initialized @NonNull Future<T extends @Initialized @NonNull Object> submit(@Initialized @NonNull ExecutorService this, @Initialized @NonNull Callable<T extends @Initialized @NonNull Object> p0) in java.util.concurrent.ExecutorService; attempting to use an incompatible parameter type
[...]

PR #865, opened 2016-07-28
Correctly apply annotations from stub files to type parameters and wildcards

### LINKED ISSUE #824: "? super V" leads to unsatisfiable bounds
The interpretation of "? super V" is "? super V extends @NonNull Object". In the Guava Futures library, there are functions such as:

```java
static <I extends @Nullable Object, O extends @Nullable Object> ListenableFuture<O> transform(ListenableFuture<I> input, Function<? super I,? extends O> function);
```

This leads to the following code failing to type check:

```java
public static ListenableFuture</*@NullableType*/ String> nullableStringTransform(
    ListenableFuture</*@NullableType*/ String> input,
    Function</*@NullableType*/ String, /*@NullableType*/ String> f) {
  return Futures.transform(input, f);
}
```

This appears to be related to us defining annotations in astub files rather than main source code, it doesn't seem to be reproducible if I define equivalent interfaces inline. The error reported is:

```
  found   : @Initialized @NonNull Function<@Initialized @Nullable String, @Initialized @Nullable String>
  required: @Initialized @NonNull Function<? extends @Initialized @NonNull Object super @Initialized @Nullable String, ? extends @Initialized @Nullable String>
```

There is also a second problem. If I change the astub file by adding `@Nullable` to the `Function` parameter like so:

```java
static <I extends @Nullable Object, O extends @Nullable Object> ListenableFuture<O> transform(ListenableFuture<I> input, Function<@Nullable ? super I,? extends O> function);
```

I then cannot compile a non-null version of the method above:

```java
public static ListenableFuture<String> stringTransform(
    ListenableFuture<String> input,
    Function<String, String> f) {
  return Futures.transform(input, f);
}
```

That fails with the error:

```
  found   : @Initialized @NonNull Function<@Initialized @NonNull String, @Initialized @NonNull String>
  required: @Initialized @NonNull Function<@Nullable ? extends @Initialized @Nullable Object super @Initialized @Nullable String, ? extends @Initialized @NonNull String>
```

This seems to contradict the checker framework documentation which states that in `@Nullable X super T` `@Nullable` is an upper rather than a lower bound, and therefore should be substitutable with a non-null type.

smillst:
Fixes #824 and #590.


checker/jtreg/nullness/issue824/Class1.java
```diff
@@ -0,0 +1,20 @@
+/*
+ * @test
+ * @summary Test case for Issue 824 https://github.com/typetools/checker-framework/issues/824
+ * @compile -XDrawDiagnostics -Xlint:unchecked Class1.java
+ * @compile -XDrawDiagnostics -Xlint:unchecked -processor org.checkerframework.checker.nullness.NullnessChecker -AprintErrorStack -Anomsgtext Class2.java -Astubs=Class1.astub
```
wmdietl: It would be nice to ensure that the correct errors are produced when the stub file is not given. I.e. have a copy of line 5 without the stub option and expected errors.

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -131,6 +131,7 @@
      * Annotation to added to every method and constructor in the stub file.
      */
     private final AnnotationMirror fromStubFile;
+    private List<AnnotatedTypeMirror> typeParameters = new ArrayList<>();
```
wmdietl: How about making the field final and calling clear() instead of allocating a new object on line 440?

wmdietl: Some javadoc for the purpose of the field would be great.

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -435,9 +437,10 @@ private void parse(
                 stubWarnIfNotFound("StubParser ignoring: " + elt);
             }
         }
+        typeParameters = new ArrayList<>();
     }
 
-    private void parseType(
+    private List<AnnotatedTypeMirror> parseType(
```
wmdietl: It would be good to document what the meaning of the returned ATM is.

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -845,8 +868,11 @@ private void annotateParameters(
             TypeParameter param = typeParameters.get(i);
             AnnotatedTypeVariable paramType = (AnnotatedTypeVariable) typeArguments.get(i);
 
-            annotate(paramType, param.getAnnotations());
-            if (param.getTypeBound() != null && param.getTypeBound().size() == 1) {
+            if (param.getTypeBound() == null || param.getTypeBound().size() != 1) {
```
wmdietl: Is the `!= 1` case covering intersection and union types? Should it be `> 1`? Maybe add a comment why these don't need to be handled.

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -743,10 +746,30 @@ private void annotate(AnnotatedTypeMirror atype, Type typeDef) {
...
                 annotate(wildcardType.getSuperBound(), wildcardDef.getSuper());
+                annotate(wildcardType.getExtendsBound(), typeDef.getAnnotations());
+            } else {
+                annotate(atype, typeDef.getAnnotations());
+            }
+        } else if (atype.getKind() == TypeKind.TYPEVAR) {
+            //Add annotations from the declaration of the TypeVariable
+            AnnotatedTypeVariable typeVarUse = (AnnotatedTypeVariable) atype;
+            for (AnnotatedTypeMirror annotatedTypeMirror : typeParameters) {
+                if (annotatedTypeMirror.getKind() != TypeKind.TYPEVAR) {
```
wmdietl: Why do we need to allow this case? Can't we make the type of `typeParameters` more specific and only allow AnnotatedTypeVariables?

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -133,6 +133,12 @@
     private final AnnotationMirror fromStubFile;
 
     /**
+     * List of AnnotatedTypeMirrors for class or method type parameters that are in scope of the
+     * elements currently parsed.
+     */
+    private List<AnnotatedTypeVariable> typeParameters = new ArrayList<>();
```
wmdietl: Making this final and using `clear` would reduce memory pressure.

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -435,9 +442,13 @@ private void parse(
                 stubWarnIfNotFound("StubParser ignoring: " + elt);
             }
         }
+        typeParameters = new ArrayList<>();
     }
 
-    private void parseType(
+    /**
+     * @return List of AnnotatedTypeMirrors of the type's type parameters
```
wmdietl: ATM -> ATV.
I assume all returned ATV will be declaration type variables, not use.

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -506,9 +517,18 @@ private void parseType(
...
         }
 
-        annotateParameters(typeArguments, typeParameters);
+        annotateTypeParameters(typeArguments, typeParameters);
         annotateSupertypes(decl, type);
         putNew(atypes, elt, type);
+        List<AnnotatedTypeVariable> typeVariables = new ArrayList<>();
+        for (AnnotatedTypeMirror typeV : type.getTypeArguments()) {
+            if (typeV.getKind() != TypeKind.TYPEVAR) {
+                stubAlwaysWarn("Expected an AnnotatedTypeVariable but found " + typeV.getKind());
```
wmdietl: Add typeV to error message to make easier to find?

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -845,8 +880,11 @@ private void annotateParameters(
...
 
-            annotate(paramType, param.getAnnotations());
-            if (param.getTypeBound() != null && param.getTypeBound().size() == 1) {
+            if (param.getTypeBound() == null || param.getTypeBound().isEmpty()) {
+                // No bound so annotations are both lower and upper bounds
+                annotate(paramType, param.getAnnotations());
+            } else if (param.getTypeBound() != null && param.getTypeBound().size() == 1) {
+                annotate(paramType.getLowerBound(), param.getAnnotations());
                 annotate(paramType.getUpperBound(), param.getTypeBound().get(0));
             }
```
wmdietl: These cover cases of no bounds and exactly one bound. What about multiple bounds?

smillst: Type variable declarations can only have extends bounds.

smillst: Looking at the code, I'm not sure why TypeParameter#getTypeBounds returns a list.  Maybe for intersection types?

wmdietl: Yes, I guess it's for code like this:

````
class A<T extends @TA Iface1 & @TB Iface2>
````

It would be good to at least issue a warning if we see > 1 bounds.

smillst: I confirmed that there are > 1 bounds when the upperbound is an intersection.  I added a warning and I'm working on implementing intersection support.  (But not for this release.)

PR #854, opened 2016-07-20
Refactor AnnotatedTypeFactory: Use lazy initialization in getSupportedTypeQualifiers()

CharlesZ-Chen:
Sorry, I made a mistake when cherry-pick the other two unrelated changes.

I have to create this new PR, because I do a force-update on the opprop/master.

Original PR please refer #848

PR #866, opened 2016-07-29
Don't require checkers to be in a package (Fixes #854)

smillst:
(Fixes #854)

### DIFF
--- a/framework/jtreg/issue845/checker/NotInPackageChecker.java
+++ b/framework/jtreg/issue845/checker/NotInPackageChecker.java
@@ -1,0 +1,7 @@
+import org.checkerframework.common.basetype.BaseTypeChecker;
+
+/**
+ * Checker that is in the default package.  This tests for Issue 845
+ * https://github.com/typetools/checker-framework/issues/845
+ */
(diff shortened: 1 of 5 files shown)


framework/jtreg/issue845/checker/NotInPackageChecker.java
```diff
@@ -0,0 +1,7 @@
+import org.checkerframework.common.basetype.BaseTypeChecker;
+
+/**
+ * Checker that is in the default package.  This tests for Issue 845
+ * https://github.com/typetools/checker-framework/issues/845
```
wmdietl: Small comment: we usually have this comment about which issue it is in the first lines of the file.
For JTReg tests, I would suggest having the link with the test driver, Test.java in this case.

wmdietl:
Looks good!

framework/jtreg/issue845/checker/NotInPackageChecker.java
smillst: Test.java has the link in the `@summary` tag.  Would you prefer it else where?

smillst:
@wmdietl Thanks for the quick review.

PR #869, opened 2016-08-15
Fixes bug in getAnnotationInHierarchy involving @PolyAll

smillst:
QualifierHierarchy#getAnnotationInHierarchy used to return `@PolyAll` even if another annotation in that hierarchy is in the set.  

Also, renamed find findCorrespondingAnnotation to getCorrespondingAnnotation.  Though maybe getAnnotationInSameHierarchy would be better?

getCorrespondingAnnotation now calls getAnnotationInHierarchy rather than reimplementing the same functionality.  This uncovered a bug in QualifierHierarchyAdapter#getAnnotationInHierarchy which I fixed.

### DIFF
--- a/framework/src/org/checkerframework/framework/type/QualifierHierarchy.java
+++ b/framework/src/org/checkerframework/framework/type/QualifierHierarchy.java
@@ -9,6 +9,7 @@ import java.util.Map;
 import java.util.Set;
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.type.TypeKind;
+import org.checkerframework.framework.qual.PolyAll;
 import org.checkerframework.javacutil.AnnotationUtils;
 import org.checkerframework.javacutil.ErrorReporter;
 
@@ -564,15 +565,28 @@ public abstract class QualifierHierarchy {
         }
     }
 
+    /**
+     *
+     * @deprecated use {@link #getCorrespondingAnnotation(Collection, AnnotationMirror)} instead
+     */
+    @Deprecated
     public AnnotationMirror findCorrespondingAnnotation(
             AnnotationMirror aliased, Collection<? extends AnnotationMirror> a) {
-        AnnotationMirror top = this.getTopAnnotation(aliased);
-        for (AnnotationMirror anno : a) {
-            if (this.isSubtype(anno, top)) {
-                return anno;
-            }
-        }
-        return null;
+        return getCorrespondingAnnotation(a, aliased);
+    }
+
+    /**
+     * Returns the annotation from the hierarchy that corresponds to {@code annotationMirror} from
+     * a set of annotations.
+     *
+     * @param annos The set of annotations.
+     * @param annotationMirror annotation that is in same hierarchy as the returned annotation
+     * @return AnnotationMirror in annos in the same hierarchy as annotationMirror
+     */
+    public AnnotationMirror getCorrespondingAnnotation(
+            Collection<? extends AnnotationMirror> annos, AnnotationMirror annotationMirror) {
+        AnnotationMirror top = this.getTopAnnotation(annotationMirror);
+        return getAnnotationInHierarchy(annos, top);
     }
 
     /**
@@ -587,10 +601,17 @@ public abstract class QualifierHierarchy {
     public AnnotationMirror getAnnotationInHierarchy(
             Collection<? extends AnnotationMirror> annos, AnnotationMirror top) {
         AnnotationMirror annoInHierarchy = null;
-        for (AnnotationMirror rhsAnno : annos) {
-            if (isSubtype(rhsAnno, top)) {
-                annoInHierarchy = rhsAnno;
+        AnnotationMirror polyAll = null;
+        for (AnnotationMirror anno : annos) {
+            boolean isSubtype = isSubtype(anno, top);
+            if (isSubtype && AnnotationUtils.areSameByClass(anno, PolyAll.class)) {
+                polyAll = anno;
+            } else if (isSubtype) {
+                annoInHierarchy = anno;
             }
+        }
+        if (annoInHierarchy == null) {
+            return polyAll;
         }
         return annoInHierarchy;
     }
--- a/framework/src/org/checkerframework/framework/type/QualifierHierarchy.java
+++ b/framework/src/org/checkerframework/framework/type/QualifierHierarchy.java
@@ -567,53 +567,57 @@ public abstract class QualifierHierarchy {
 
     /**
      *
-     * @deprecated use {@link #getCorrespondingAnnotation(Collection, AnnotationMirror)} instead
+     * @deprecated use {@link #findAnnotationInSameHierarchy(Collection, AnnotationMirror)} instead
      */
     @Deprecated
     public AnnotationMirror findCorrespondingAnnotation(
             AnnotationMirror aliased, Collection<? extends AnnotationMirror> a) {
-        return getCorrespondingAnnotation(a, aliased);
+        return findAnnotationInSameHierarchy(a, aliased);
     }
 
     /**
-     * Returns the annotation from the hierarchy that corresponds to {@code annotationMirror} from
-     * a set of annotations.
+     * Returns the annotation in annos that is in the same hierarchy as annotationMirror
      *
-     * @param annos The set of annotations.
+     * If the annotation in the hierarchy is PolyAll, then the polymorphic qualifier in the
+     * hierarchy is returned instead of PolyAll.
+     *
+     * @param annos set of annotations to search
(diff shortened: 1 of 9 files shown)


framework/src/org/checkerframework/framework/type/QualifierHierarchy.java
```diff
@@ -564,15 +565,28 @@ public AnnotationMirror greatestLowerBound(
...
-        }
-        return null;
+        return getCorrespondingAnnotation(a, aliased);
+    }
+
+    /**
+     * Returns the annotation from the hierarchy that corresponds to {@code annotationMirror} from
+     * a set of annotations.
+     *
+     * @param annos The set of annotations.
```
mernst: Nit: In a Javadoc @param clause (and related clauses), the initial text is a
sentence fragment that starts with a lowercase (not capital) letter and
does not end with a period unless followed by another sentence.

framework/src/org/checkerframework/framework/type/QualifierHierarchy.java
```diff
@@ -564,15 +565,28 @@ public AnnotationMirror greatestLowerBound(
...
-            if (this.isSubtype(anno, top)) {
-                return anno;
-            }
-        }
-        return null;
+        return getCorrespondingAnnotation(a, aliased);
+    }
+
+    /**
+     * Returns the annotation from the hierarchy that corresponds to {@code annotationMirror} from
```
mernst: Hard to parse.   I would reorder to something like:

Returns the annotation that corresponds to the set annos.
The returned annotations will be in the same hierarchy as annotationMirror.

I presume it's an invariant that all the annotations in the set, and the second argument, must be from the same hierarchy, and the second argument is only necessary because the set might be empty.  (Actually, can the set be empty?  If not, what is the point of the second argument?  Are the elements in the set guaranteed to all be in different hierarchies?  Is the set required to be non-empty?

framework/src/org/checkerframework/framework/type/QualifierHierarchy.java
```diff
@@ -564,15 +565,28 @@ public AnnotationMirror greatestLowerBound(
...
+     * a set of annotations.
+     *
+     * @param annos The set of annotations.
+     * @param annotationMirror annotation that is in same hierarchy as the returned annotation
+     * @return AnnotationMirror in annos in the same hierarchy as annotationMirror
```
mernst: Does it always return one of the elements of the set?  This `@return` clause suggests that.

framework/src/org/checkerframework/framework/type/QualifierHierarchy.java
```diff
@@ -587,11 +601,18 @@ public AnnotationMirror findCorrespondingAnnotation(
     public AnnotationMirror getAnnotationInHierarchy(
             Collection<? extends AnnotationMirror> annos, AnnotationMirror top) {
         AnnotationMirror annoInHierarchy = null;
-        for (AnnotationMirror rhsAnno : annos) {
-            if (isSubtype(rhsAnno, top)) {
-                annoInHierarchy = rhsAnno;
+        AnnotationMirror polyAll = null;
```
mernst: Can both annoInHierarchy and polyAll be non-null?  (That would mean that there can be multiple annotations from the same hierarchy in the set.)

wmdietl:
If I have a specific type system with `@A`, `@B`, and `@PolyAB`, have a type that is annotated `@PolyAll` and ask for `getAnnotationInHierarchy(@A)`, will I get `@PolyAB` or `@PolyAll`?
As we require a specific polymorphic qualifier for the particular type system, it would make sense to me to return that qualifier instead of the general `@PolyAll`.

framework/src/org/checkerframework/framework/type/QualifierHierarchy.java
```diff
@@ -587,11 +601,18 @@ public AnnotationMirror findCorrespondingAnnotation(
...
+        for (AnnotationMirror anno : annos) {
+            boolean isSubtype = isSubtype(anno, top);
+            if (isSubtype && AnnotationUtils.areSameByClass(anno, PolyAll.class)) {
```
wmdietl: I find this logic harder to parse than necessary. Would it be easier to have one test for `isSubtype` and then another for whether it is PolyAll?
The original code terminated the loop as soon as the first subtype was found. This always iterates over the whole list. Is this necessary?

wmdietl:
Regarding the renaming from `findCorrespondingAnnotation` to `getCorrespondingAnnotation` or as alternative `getAnnotationInSameHierarchy`.
I like `find` better, because the method isn't a getter for some state of the qualifier hierarchy. The method finds the corresponding annotation in the set of annotations.
`findAnnotationInSameHierarchy` would be an alternative.

smillst:
@wmdietl  and @mernst thanks for the feedback!

`@PolyAll` violates our one annotation per hierarchy rule, which is the root cause of the bug this pull requests fix (and also the cause of the failures in the first commit).  I spent sometime yesterday trying to change the implementation of `@PolyAll` so that it does not violate this rule, but I wasn't successful.  
 
There are two problems:
 1. At no place in the code are the annotations in an AnnotatedTypeMirror guaranteed to only be explicit annotations. So, there is no where to translate `@PolyAll` to the correct polymorphic qualifiers.
 2. There is no easy way to indicate that a type is annotated with `@PolyAll` other than adding it to the set of annotation mirrors.  (Though, we could add a field to AnnotatedTypeMirror.)  

Neither problem is impossible to overcome, but it didn't seem worth the time to implement.

It turned out that using @wmdietl suggestion (https://github.com/typetools/checker-framework/pull/869#issuecomment-239956241) to return the polymorphic qualifier in the hierarchy rather than `@PolyAll`, fixed the failures in the first commit.  It should also help avoid future problems.

framework/src/org/checkerframework/framework/type/QualifierHierarchy.java
```diff
@@ -564,35 +565,59 @@ public AnnotationMirror greatestLowerBound(
...
-            }
-        }
-        return null;
+        return findAnnotationInSameHierarchy(a, aliased);
     }
 
     /**
-     * Returns the annotation from the hierarchy identified by its 'top' annotation
-     * from a set of annotations, using this QualifierHierarchy for subtype tests.
+     * Returns the annotation in annos that is in the same hierarchy as annotationMirror
```
mernst: Nit:  end sentences with a period.

framework/src/org/checkerframework/framework/type/QualifierHierarchy.java
```diff
@@ -564,35 +565,59 @@ public AnnotationMirror greatestLowerBound(
...
+    }
+
+    /**
+     * Returns the annotation in annos that is in the hierarchy of which annotationMirror is top
+     *
+     * If the annotation in the hierarchy is PolyAll, then the polymorphic qualifier in the
+     * hierarchy is returned instead of PolyAll.
+     *
+     * @param annos set of annotations to search
+     * @param top annotation that is the top of hierarchy of which the returned annotation belongs
```
mernst: of which => to which

mernst:
Looks good to me; feel free to merge.

framework/src/org/checkerframework/framework/type/QualifierHierarchy.java
```diff
@@ -564,35 +565,59 @@ public AnnotationMirror greatestLowerBound(
...
             Collection<? extends AnnotationMirror> annos, AnnotationMirror top) {
-        AnnotationMirror annoInHierarchy = null;
-        for (AnnotationMirror rhsAnno : annos) {
-            if (isSubtype(rhsAnno, top)) {
-                annoInHierarchy = rhsAnno;
+        boolean hasPolyAll = false;
+        for (AnnotationMirror anno : annos) {
+            if (!isSubtype(anno, top)) {
+                continue;
+            } else if (!AnnotationUtils.areSameByClass(anno, PolyAll.class)) {
```
mernst: As a minor issue, positive tests are easier to read than negative ones (that `!` is easy to overlook, and I did so on first reading).  This is acceptable, though.

wmdietl: The logic of this loop is rather hard to follow. There are two negated conditions, a continue, and a break. It is not direct to see when hasPolyAll is true. Is the following loop body equivalent?

```
if (isSubtype(anno, top)) {
  if (AnnotationUtils.areSameByClass(anno, PolyAll.class)) {
    // found PolyAll; continue to see whether something specific follows
    hasPolyAll = true;
  } else {
    // found a concrete subtype, return it
    return anno;
  }
}
```

wmdietl:
One small code comment that might make some logic more direct.
Otherwise looks good and can be merged.
Thanks!

Issue #867, opened 2016-07-29
Dataflow: postfix increment and decrements are treated like prefix postfix increment and decrements


smillst:
Postfix increments and decrement trees are converted to control flow graphs as though they were prefix increments and decrement.  (See [todo](https://github.com/typetools/checker-framework/blob/180a57a8e308172e3efc0d3cbeac1fbe482bd286//dataflow/src/org/checkerframework/dataflow/cfg/CFGBuilder.java#L4260)) The Value Checker has a [hack](https://github.com/typetools/checker-framework/blob/180a57a8e308172e3efc0d3cbeac1fbe482bd286/framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java#L134) to work around this issue.  The Array Index Upper and Lower Bound Checkers that Martin Kellogg (@kelloggm) is working on will also encounter this issue.

smillst:
A work around for checkers was add with this commit: 1d0f8c4

smillst:
This is also causing a problem in type-checking post-fix expressions.  See comment in BaseTypeVisitor#visitUnary.

PR #873, opened 2016-08-18
Work around for #867 (Postfixed increments and decrements)

smillst:
Removed Value Checker specific work around for #867 and add a more general one.


mernst:
Looks OK to me.

wmdietl:
Much nicer solution! Ok for me.

PR #882, opened 2016-08-25
Dataflow: Change CFAbstractValue to store a set of annotations rather than an ATM

### LINKED ISSUE #580: Wildcards: assertion failure in initialization checker
Request: `javac -processor nullness -Awarns -AprintErrorStack `[InitCheckAssertionFailure.java](https://gist.github.com/dbrosoft/f25faeff48947ada1dae)

Response:
```
error: SourceChecker.typeProcess: unexpected Throwable (AssertionError) while processing InitCheckAssertionFailure.java
  Compilation unit: InitCheckAssertionFailure.java
  Exception: java.lang.AssertionError; Stack trace: org.checkerframework.checker.initialization.InitializationAnnotatedTypeFactory.isInitializationAnnotation(InitializationAnnotatedTypeFactory.java:133)
[...]
```

smillst:
Because dataflow sometimes only has information about the primary annotation and only the refined primary annotation is ever used, dataflow should operate on sets of annotations rather than a possibly incomplete AnnotatedTypeMirror.

(Tested with daikon-typecheck)

Also, Fixes #580


checker/src/org/checkerframework/checker/nullness/NullnessAnalysis.java
```diff
@@ -41,6 +46,17 @@ public NullnessValue createAbstractValue(AnnotatedTypeMirror type) {
             // 'no information'.
             return null;
         }
-        return new NullnessValue(this, type);
+        if (type.getKind() == TypeKind.WILDCARD) {
+            AnnotatedWildcardType wildcard = (AnnotatedWildcardType) type;
+            Set<AnnotationMirror> annos = wildcard.getExtendsBound().getAnnotations();
+            return new NullnessValue(this, annos, type.getUnderlyingType());
+        }
+        return new NullnessValue(this, type.getAnnotations(), type.getUnderlyingType());
```
wmdietl: Would using `getEffectiveAnnotations` cover both wildcards and type variables and not require the special case above?

checker/src/org/checkerframework/checker/nullness/NullnessTransfer.java
```diff
@@ -61,6 +64,7 @@
 
     /** Annotations of the non-null type system. */
     protected final AnnotationMirror NONNULL, NULLABLE;
+    protected final AnnotationMirror NONRAW, INITIALIZED;
```
wmdietl: These two new fields seem to be set, but never used.

framework/src/org/checkerframework/framework/flow/CFAbstractAnalysis.java
```diff
@@ -115,6 +115,13 @@ public T createTransferFunction() {
     public abstract /*@Nullable*/ V createAbstractValue(AnnotatedTypeMirror type);
```
wmdietl: Should this method be removed?

framework/src/org/checkerframework/framework/flow/CFAbstractAnalysis.java
```diff
@@ -127,7 +134,22 @@ public CFValue defaultCreateAbstractValue(
             // 'no information'.
             return null;
         }
-        return new CFValue(analysis, type);
+        if (type.getKind() == TypeKind.WILDCARD) {
+            AnnotatedWildcardType wildcard = (AnnotatedWildcardType) type;
+            Set<AnnotationMirror> annos = wildcard.getExtendsBound().getAnnotations();
+            return new CFValue(analysis, annos, type.getUnderlyingType());
+        }
+        return new CFValue(analysis, type.getAnnotations(), type.getUnderlyingType());
```
wmdietl: Using `getEffectiveAnnotations` should avoid having to use a special case for wildcards.

framework/src/org/checkerframework/framework/flow/CFAbstractValue.java
```diff
@@ -4,550 +4,454 @@
...
-        if (sourceComp instanceof AnnotatedArrayType) {
-            assert dest instanceof AnnotatedArrayType;
-            copyArrayComponentAnnotations(
-                    (AnnotatedArrayType) sourceComp, (AnnotatedArrayType) destComp);
+    private boolean canBeMissingAnnotations(TypeMirror typeMirror) {
+        if (typeMirror == null) {
+            return false;
+        }
+        if (typeMirror.getKind() == TypeKind.WILDCARD) {
+            return canBeMissingAnnotations(((WildcardType) typeMirror).getExtendsBound());
```
wmdietl: Why different logic for wildcards and type variables? Why not the same recursive call in both cases?

framework/src/org/checkerframework/framework/flow/CFAbstractValue.java
```diff
@@ -4,550 +4,454 @@
...
-     * treated as 'top'.
+     * Returns whether this value is a subtype of the argument {@code other}. The annotations are
+     * compared per hierarchy, and missing annotations are treated as 'top'.
      */
+    @Deprecated
     public boolean isSubtypeOf(CFAbstractValue<V> other) {
-        if (other == null) {
-            // 'null' is 'top'
-            return true;
+        //        if (other == null) {
```
wmdietl: Remove the old code from comments?
Why does the method always return true?

framework/src/org/checkerframework/framework/flow/CFAbstractValue.java
```diff
@@ -4,550 +4,454 @@
...
-
-    /**
-     * The type (with annotations) corresponding to this abstract value.
-     */
-    protected final AnnotatedTypeMirror type;
-
-    public CFAbstractValue(CFAbstractAnalysis<V, ?, ?> analysis, AnnotatedTypeMirror type) {
-        assert AnnotatedTypes.isValidType(analysis.qualifierHierarchy, type)
-                : "Encountered invalid type: " + type.toString(true);
+    protected final GenericAnnotatedTypeFactory<V, ?, ?, ?> factory;
```
wmdietl: I'm a bit concerned about the memory consumption of these three additional fields. We'll have many abstract values and all of them will have the same references to factory, processingEnv, and hierarchy.

framework/src/org/checkerframework/framework/flow/CFAbstractValue.java
```diff
@@ -4,550 +4,454 @@
...
+    protected final ProcessingEnvironment processingEnv;
+    protected final QualifierHierarchy hierarchy;
+    private final TypeMirror underlyingType;
```
wmdietl: Previously `type` was protected. The replacements are private. Why?

framework/tests/all-systems/Issue263.java
```diff
@@ -1,6 +1,6 @@
 // Test case for Issue 263:
 // https://github.com/typetools/checker-framework/issues/263
-
+@SuppressWarnings({"regex", "tainting"})
```
wmdietl: Why do these become necessary?

smillst:
@wmdietl  Thanks for the feedback.  I added a comment explaining the wildcard special handling to CFAbstractValue.

framework/src/org/checkerframework/framework/flow/CFAbstractValue.java
smillst: The values used to store a reference to an ATM that has many more references than the few that  are here.  So, the memory consumption should be less than before.

framework/src/org/checkerframework/framework/flow/CFAbstractValue.java
```diff
@@ -4,550 +4,481 @@
...
  */
 public abstract class CFAbstractValue<V extends CFAbstractValue<V>> implements AbstractValue<V> {
 
     /**
-     * The analysis class this store belongs to.
+     * The analysis class this value belongs to.
      */
     protected final CFAbstractAnalysis<V, ?, ?> analysis;
 
     protected final TypeHierarchy typeHierarchy;
```
wmdietl: I think we could safe a lot of memory by not having all of these fields. There will be a lot CFAbstractValue instances and most of these fields will be the same.
Most of them are reachable through `analysis`. Calling the getter methods would safe storage.

framework/src/org/checkerframework/framework/flow/CFAbstractValue.java
```diff
@@ -4,550 +4,481 @@
...
+ * An implementation of an abstract value used by the Checker Framework
+ * org.checkerframework.dataflow analysis.
  *
- * @author Stefan Heule
+ * A value holds a set of annotations and a type mirror.  The set of annotations represents the
+ * primary annotation on a type; therefore, the set of annotations must have an annotation for
+ * each hierarchy unless the type mirror is a type variable or a wildcard that extends a type
+ * variable. Both type variables and wildcards may be missing a primary annotation.  For this set
+ * of annotations, there is an additional constraint that only wildcards that extend type
+ * variables can be missing annotations because
```
wmdietl: Strange end of sentence.

framework/src/org/checkerframework/framework/flow/CFAbstractValue.java
```diff
@@ -4,550 +4,481 @@
...
+ *
+ * In order to compute {@link #leastUpperBound(CFAbstractValue)} and
+ * {@link #mostSpecific(CFAbstractValue, CFAbstractValue)}, the case where one value has an
+ * annotation in a hierarchy and the other does not must be handled.  For type variables, the
+ * {@link AnnotatedTypeVariable} for the declaration of the type variable is used.  The
+ * {@link AnnotatedTypeVariable} is computed using the type mirror.  For wildcards, it is not
+ * always possible to get the {@link AnnotatedWildcardType} for the type mirror.  However, a
+ * CFAbstractValue's type mirror is only a wildcard if the type of some expression is a wildcard,
+ * so it is always the use of a wildcard rather than the declaration. For uses of wildcards, only
```
wmdietl: What is the declaration of a wildcard? Is this about the lack of captured types?

framework/src/org/checkerframework/framework/flow/CFAbstractValue.java
```diff
@@ -4,550 +4,481 @@
...
+ * the primary annotation on the upper bound is ever used.  So, the set of annotations represents
+ * the primary annotation on the wildcard's upper bound.  If that upper bound is a type variable,
```
wmdietl: If the upper bound of a wildcard is a type variable, why not use the annotations on the upper bound of that type variable?

framework/src/org/checkerframework/framework/flow/CFAbstractValue.java
```diff
@@ -4,550 +4,481 @@
...
-     * The type (with annotations) corresponding to this abstract value.
-     */
-    protected final AnnotatedTypeMirror type;
+        validateSet(this.getAnnotations(), this.getUnderlyingType());
+    }
 
     public CFAbstractValue(CFAbstractAnalysis<V, ?, ?> analysis, AnnotatedTypeMirror type) {
-        assert AnnotatedTypes.isValidType(analysis.qualifierHierarchy, type)
-                : "Encountered invalid type: " + type.toString(true);
+        if (type.getKind() == TypeKind.WILDCARD) {
```
wmdietl: I still don't understand why this isn't simply `this.annotations = type.getEffectiveAnnotations();`.

framework/src/org/checkerframework/framework/flow/CFAbstractValue.java
```diff
@@ -4,550 +4,481 @@
...
+            this.annotations = wildcard.getExtendsBound().getAnnotations();
+        } else {
+            this.annotations = type.getAnnotations();
+        }
+
         this.analysis = analysis;
-        this.type = type;
         this.typeHierarchy = analysis.getTypeHierarchy();
-    }
+        this.underlyingType = type.getUnderlyingType();
```
wmdietl: This seems to be an auxiliary constructor. You could avoid code duplication by calling the primary constructor.

framework/src/org/checkerframework/framework/flow/CFAbstractValue.java
```diff
@@ -4,550 +4,481 @@
...
-            AnnotatedTypeMirror extendsBound = ((AnnotatedWildcardType) t).getExtendsBound();
-            if (extendsBound != null) {
-                return extendsBound.getEffectiveAnnotations();
-            }
-        } else if (t.getKind() == TypeKind.TYPEVAR) {
-            AnnotatedTypeMirror upperBound = ((AnnotatedTypeVariable) t).getUpperBound();
-            if (upperBound != null) {
-                return upperBound.getEffectiveAnnotations();
-            }
+    @Pure // because the annotations field is final
```
wmdietl: The `because` comment doesn't make sense. The method only returns the field. Even if the field is not final, this method is pure. Same for `getUnderlyingType` below.

framework/src/org/checkerframework/framework/flow/CFAbstractValue.java
```diff
@@ -4,550 +4,481 @@
...
     }
 
-    /**
-     * @return the string representation as a comma-separated list
-     */
-    @SideEffectFree
-    @Override
-    public String toString() {
-        return getType().toString();
+    private AnnotatedTypeVariable getEffectTypeVar(TypeMirror typeMirror) {
```
wmdietl: A little documentation would be nice for many of these methods. What is an `EffectTypeVar`?

framework/src/org/checkerframework/framework/type/DefaultInferredTypesApplier.java
```diff
@@ -15,179 +19,111 @@
...
-                    type.addAnnotation(inferredAnnotation);
-                }
+    private void apply(
+            AnnotatedTypeMirror type,
+            AnnotationMirror inferred,
+            TypeMirror inferredTypeMirror,
+            AnnotationMirror top) {
+
+        AnnotationMirror primary = type.getAnnotationInHierarchy(top);
+        if (inferred == null && primary == null) {
```
wmdietl: I would find these much easier to read if the shared conditions are checked once, instead of repeatedly with additional conditions.

framework/src/org/checkerframework/framework/type/DefaultInferredTypesApplier.java
```diff
@@ -15,179 +19,111 @@
...
+        } else if (inferred == null) {
+            removePrimaryTypeVarApplyUpperBound(type, inferredTypeMirror, top, primary);
+        } else if (primary != null
+                && (omitSubtypingCheck || hierarchy.isSubtype(inferred, primary))) {
+            type.replaceAnnotation(inferred);
+        } else if (primary == null) {
+            Set<AnnotationMirror> lowerbounds =
+                    AnnotatedTypes.findEffectiveLowerBoundAnnotations(hierarchy, type);
+            AnnotationMirror lowerbound = hierarchy.findAnnotationInHierarchy(lowerbounds, top);
+            if (omitSubtypingCheck | hierarchy.isSubtype(inferred, lowerbound)) {
```
wmdietl: This should be logical or, not binary.

framework/src/org/checkerframework/framework/type/DefaultInferredTypesApplier.java
```diff
@@ -15,179 +19,111 @@
...
-                visit(type.getLowerBound(), inferred.getLowerBound(), top);
-            }
-            return null;
+    private void removePrimaryTypeVarApplyUpperBound(
+            AnnotatedTypeMirror type,
+            TypeMirror inferredTypeMirror,
+            AnnotationMirror top,
+            AnnotationMirror notInferred) {
+        if (inferredTypeMirror.getKind() != TypeKind.TYPEVAR) {
+            ErrorReporter.errorAbort("Missing annos");
```
wmdietl: The error message seems confusing. Shouldn't it be incorrect type kind? Add the type to the error message.

framework/tests/all-systems/Issue263.java
```diff
@@ -1,6 +1,9 @@
 // Test case for Issue 263:
 // https://github.com/typetools/checker-framework/issues/263
 
+// Suppression required because of Issue 742
```
wmdietl: 742 is a merged pull request titled "Make variables static, set in static initializer block". Should this be some other number?

javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -331,6 +331,22 @@ public static boolean containsSameByClass(
     }
 
     /**
+     * Returns the AnnotationMirror in {@code c} that has the same class as {@code anno}.
+     *
+     * @return AnnotationMirror with the same class as {@code anno} iff c contains anno, according
+     * to areSameByClass; otherwise, {@code null}
+     */
+    public static AnnotationMirror getAnnotationByClass(
```
wmdietl: To reduce duplication, do you want to change `containsSameByClass` to simply `return getAnnotationByClass(c, anno) != null;"?

framework/src/org/checkerframework/framework/flow/CFAbstractValue.java
smillst: `@A T extends @B Object`
`? extends T'

Calling type.getEffectiveAnnotations() on the wildcard would return `@B`, this is not precise enough for this case:
````
<T> T getValue(List<? extends T> list) {
  T t = list.get(0)
  return t;
}
````

Instead the CFValue should store an empty set in case the wildcard extends a type variable.

File: framework/src/org/checkerframework/framework/flow/CFAbstractValue.java (hunk 13 above)
Comment (smillst) [inline framework/src/org/checkerframework/framework/flow/CFAbstractValue.java]: For example
````
<@A T extends @B Object> T getValue(List<? extends T> list) {
  T t = list.get(0)
  return t;
}
````

If we store the effective annotation of `list.get(0)`, then the refined type of `t` is `@B T`` and an error will be issued at the return statement.

File: javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java (hunk 22 above)
Comment (smillst) [inline javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java]: done

File: framework/src/org/checkerframework/framework/flow/CFAbstractAnalysis.java (hunk 3 above)
Comment (smillst) [inline framework/src/org/checkerframework/framework/flow/CFAbstractAnalysis.java]: No, I'll document it.

File: checker/src/org/checkerframework/checker/lock/LockAnalysis.java [hunk 23]
```diff
@@ -40,4 +43,10 @@ public LockStore createCopiedStore(LockStore s) {
     public CFValue createAbstractValue(AnnotatedTypeMirror type) {
```
wmdietl: Why is the superclass of LockAnalysis `CFAbstractAnalysis` instead of `CFAnalysis`.
If this could be changed, these two `createAbstractValue` overrides would no longer be necessary.
If this can't be changed, add some documentation why.

framework/src/org/checkerframework/framework/flow/CFAbstractAnalysis.java
wmdietl: I don't like this design. We now have two abstract methods that each analysis class needs to implement.
I don't see any documentation in this class that makes clear how the two are different and when to use which.
See e.g. NullnessValue and NullnessAnalysis for these new methods that are suddenly necessary.

framework/src/org/checkerframework/framework/flow/CFAbstractValue.java
smillst: this is done.

wmdietl:
Building the annotated JDK seems to fail for the Nullness Checker. I restarted the travis build to make sure, but it still seems to fail.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -618,10 +618,12 @@ protected void checkPostconditions(MethodTree node, ExecutableElement methodElem
                     // check anything
                 } else {
                     CFAbstractValue<?> value = exitStore.getValue(expr);
-                    AnnotationMirror inferredAnno =
-                            value == null
-                                    ? null
-                                    : value.getType().getAnnotationInHierarchy(annotation);
+                    AnnotationMirror inferredAnno = null;
```
wmdietl: This code for going from a non-null value to the inferred AnnotationMirror is duplicated three times in this class.
If this is local to the class, put it in a private method. If similar code is outside, should it be a public method here or elsewhere?

framework/src/org/checkerframework/framework/flow/CFAbstractAnalysis.java
```diff
@@ -110,24 +108,40 @@ public T createTransferFunction() {
     public abstract S createCopiedStore(S s);
 
     /**
+     * Creates an abstract value from the annotated type mirror.  The value contains the set of
+     * primary annotations on the type; unless, the type is an AnnotatedWildcardType.  In that
+     * case, the annotations are a in the created value are the primary annotations on the
```
wmdietl: Strange wording `are a in the created value are`.

framework/src/org/checkerframework/framework/flow/CFAbstractValue.java
```diff
@@ -4,550 +4,478 @@
...
-                                    + result
-                                    + "toSearch="
-                                    + toSearch
-                                    + "top="
-                                    + top
-                                    + "source="
-                                    + source);
-            }
+    abstract class AnnotationSetAndTypeMirrorVisitor {
+        TypeMirror result;
```
wmdietl: Protected, in case a subclass ever needs to access it?

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
smillst: All three uses are in very similar methods for checking preconditions, postconditions, or conditional postconditions.  The right way to fix this would be to refactor all three methods, but this is unrelated to the changes in this pull request.

PR #884, opened 2016-08-25
Only support PolyAll by default if a polymorphic qualifier exists

### LINKED ISSUE #883: inconsistent polymorphism defaulting
In `AnnotatedTypeFactory#createSupportQualifiers()`, defaulting is set to support @PolyAll, and will add @PolyAll to the `supportedQuals`.

https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java#L807

However, in `AnnotatedTypeFactory#checkSupportedQuals()`, it will enforce a checker must have polymorphic qualifiers if @PolyAll is in the list of it's `supportedQuals`.

https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java#L420

This will cause an unexpected exception for a very simple type system that doesn't even need a polymorphic qualifier:

```java
java.lang.RuntimeException: Checker added @PolyAll to list of supported qualifiers, but the checker does not have a polymorphic qualifier.  Either remove @PolyAll from the list of supported qualifiers or add a polymorphic qualifier.
```

This is because by defaulting `AnnotatedTypeFactory` will add @PolyAll to `supportedQuals`, but because the type system is very simple and doesn't have a polymorphic qualifier, it would failed the PolyAll check in `AnnotatedTypeFactory #checkSupportedQuals()`.

It seems we need a consistent polymorphism defaulting in `AnnotatedTypeFactory` to avoid this problem. Or even better if we could enable/disable the @PolyAll depends on whether the checker has polymorphic qualifiers in it's supported qualifiers.

smillst:
This is consistent with document of `@PolyAll`
Fixes #883

### DIFF
--- a/framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
+++ b/framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
@@ -805,13 +805,13 @@ public class AnnotatedTypeFactory implements AnnotationProvider {
      *         none
      */
     protected Set<Class<? extends Annotation>> createSupportedTypeQualifiers() {
-        return getBundledTypeQualifiersWithoutPolyAll();
+        return getBundledTypeQualifiersWithPolyAll();
     }
 
     /**
      * Loads all annotations contained in the qual directory of a checker via
-     * reflection, and adds {@link PolyAll} and an explicit array of annotations
-     * to the set of annotation classes.
+     * reflection, and adds {@link PolyAll}, if a polymorphic type qualifier exists,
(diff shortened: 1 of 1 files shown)


wmdietl:
I think we wanted PolyAll to be supported with minimal effort by all type systems. So the PolyAll documentation might be too restrictive.
How about checking whether there is a type-system-specific polymorphic qualifier and if there is, automatically support PolyAll. If there is not, instead of raising an error, simply do not add PolyAll.

wmdietl:
Looks good!

PR #896, opened 2016-08-30
Correct subtyping for captured wildcard and uncaptured wildcard (Issue 870)

### LINKED ISSUE #870: Bogus error "lambda.param.type.incompatible"
The following program can be compiled by javac. If, however, the NullnessChecker is used, one gets 

_[9,33] error: [lambda.param.type.incompatible] incompatible parameter types in lambda expression._
```
package bar;

import java.util.stream.Stream;
import java.util.zip.ZipEntry;
import java.util.zip.ZipFile;

public class Bar {
	public static Stream<? extends ZipEntry> entries(ZipFile zipFile) {
		return zipFile.stream().filter(entry -> !entry.isDirectory() && entry.getName().endsWith(".xml"));
	}
}
```

No error is issued if you write
`(ZipEntry entry) -> !entry.isDirectory() && ...`

smillst:
Fixes #870

### DIFF
--- a/javacutil/src/org/checkerframework/javacutil/InternalUtils.java
+++ b/javacutil/src/org/checkerframework/javacutil/InternalUtils.java
@@ -16,6 +16,7 @@ import com.sun.tools.javac.code.Symbol;
 import com.sun.tools.javac.code.Symbol.TypeSymbol;
 import com.sun.tools.javac.code.Type;
 import com.sun.tools.javac.code.Type.AnnotatedType;
+import com.sun.tools.javac.code.Type.CapturedType;
 import com.sun.tools.javac.code.Types;
 import com.sun.tools.javac.processing.JavacProcessingEnvironment;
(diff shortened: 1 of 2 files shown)


javacutil/src/org/checkerframework/javacutil/InternalUtils.java
```diff
@@ -247,6 +248,19 @@ public static boolean isCaptured(TypeVariable typeVar) {
     }
 
     /**
+     * If typeVar is a captured wildcard, returns that wildcard; otherwise returns null.
+     */
+    public static WildcardType getCapturedWildcard(TypeVariable typeVar) {
+        if (isCaptured(typeVar)) {
+            if (typeVar instanceof AnnotatedType) {
```
wmdietl: I think you can simplify this and always call `unannotatedType`, it does no harm.
If you want to special case, the `instanceof` should be a Kind check.

PR #897, opened 2016-08-30
Add a declarative mechanism for a checker to indicate the relevant types. (Issue 872)

### LINKED ISSUE #872: Making checkers focus on certain types
Some type-checkers are mostly interested in one type or a few types.
 * Some only care about strings and related types such as CharSequences and StringBuilders:  the Regex, Format String, Internationalization Format String, Property File, Internationalization, and Signature String Checkers.
 * Some only care about ints:  Signedness, Lower Bound, and Upper Bound Checkers.  
 * Some care about strings and ints: Constant Value Checker.

A type-checker sets a default annotation for unannotated code.  This should sometimes be different for types that the type-system cares about and those it does not care about.  An example is unchecked code that will later be read as a library.  The Checker Framework defaults every return value to Top and every formal parameter to Bottom, to indicate where annotations are missing.  However, this should not be done for types that the type system doesn't care about.

As an example, consider the formatter type system, which currently produces the following error when type-checking code against the annotated JDK:

```
        f.format("%H", new Integer(4));
                                   ^
  found   : int
  required: @FormatBottom int
```

Integer is not annotated for the formatter type system (and it shouldn't need to be), so the formal parameter to the constructor is treated as Bottom.  However, the formatter type system defaults the literal 4 to `@UnknownFormat`, so the invocation of the constructor fails.

It doesn't really matter what the formatter uses for non-relevant types such as int, but it should be consistent in checked and unchecked code.  Top seems like a reasonable choice.  In this case, all non-String types should default to `@UnknownFormat`, in all contexts including unchecked code.

Rather than fixing this individually for each of the 10 type systems above, the Checker Framework should provide a declarative mechanism for a checker to indicate the relevant types.  All non-relevant types would be ignored by the type system.  This might improve efficiency, too.  (I'm not sure whether we would still want to change the defaulting of the non-relevant types in unchecked code.)  Then, the 10 type systems should be rewritten to use the new mechanism.

smillst:
Add a declarative mechanism for a checker to indicate the relevant types.

All irrelevant types are defaulted to top. Fixes #872


checker/src/org/checkerframework/checker/i18n/I18nChecker.java
```diff
@@ -2,6 +2,7 @@
 
 import java.util.ArrayList;
 import java.util.Collection;
+import org.checkerframework.framework.qual.RelevantJavaTypes;
```
wmdietl: The import is the only change in this file. Is this spurious?

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -277,9 +279,20 @@ protected TreeAnnotator createTreeAnnotator() {
...
+            Class<?>[] classes = relevantJavaTypes.value();
+            typeAnnotators.add(
+                    new IrrelevantTypeAnnotator(
+                            this, getQualifierHierarchy().getTopAnnotations(), classes));
+        }
 
-        return new ListTypeAnnotator(new PropagationTypeAnnotator(this), implicitsTypeAnnotator);
+        implicitsTypeAnnotator = new ImplicitsTypeAnnotator(this);
+        typeAnnotators.add(implicitsTypeAnnotator);
+        typeAnnotators.add(new PropagationTypeAnnotator(this));
```
wmdietl: Isn't the PropagationTypeAnnotator at the end now, instead of previously at the beginning?
Can you add some comments what the intuition behind the ordering is?

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -277,9 +279,20 @@ protected TreeAnnotator createTreeAnnotator() {
...
+        return new ListTypeAnnotator(typeAnnotators.toArray(new TypeAnnotator[0]));
```
wmdietl: Here we convert a list to an array and then the ListTreeAnnotator goes from the array back to a list.
Would it be easier to have a second constructor that takes a List? The varargs constructor can then create the list and then call this new constructor.

framework/src/org/checkerframework/framework/type/typeannotator/IrrelevantTypeAnnotator.java
```diff
@@ -0,0 +1,113 @@
...
+import javax.lang.model.type.TypeKind;
+import javax.lang.model.type.TypeMirror;
+import javax.lang.model.util.Types;
+import org.checkerframework.framework.type.AnnotatedTypeFactory;
+import org.checkerframework.framework.type.AnnotatedTypeMirror;
+import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedExecutableType;
+import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedWildcardType;
+import org.checkerframework.javacutil.TypesUtils;
+
+public class IrrelevantTypeAnnotator extends TypeAnnotator {
```
wmdietl: Some javadoc would be good, in particular that the set of classes that the parameter takes are the relevant classes and that this Annotator does not annotate those types.

framework/src/org/checkerframework/framework/type/typeannotator/IrrelevantTypeAnnotator.java
```diff
@@ -0,0 +1,113 @@
...
+    List<TypeMirror> supportedTypes;
```
wmdietl: Some visibility modifiers for the field. If they should be default visible, let's add some comment.

framework/src/org/checkerframework/framework/type/typeannotator/IrrelevantTypeAnnotator.java
```diff
@@ -0,0 +1,113 @@
...
+    Set<TypeMirror> allFoundSupportedTypes;
+
+    boolean excludeArrays = false;
```
wmdietl: This is the only field which is not initialized from the constructor. Some documentation wold be good.

framework/src/org/checkerframework/framework/type/typeannotator/IrrelevantTypeAnnotator.java
```diff
@@ -0,0 +1,113 @@
...
+import javax.lang.model.util.Types;
+import org.checkerframework.framework.type.AnnotatedTypeFactory;
+import org.checkerframework.framework.type.AnnotatedTypeMirror;
+import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedExecutableType;
+import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedWildcardType;
+import org.checkerframework.javacutil.TypesUtils;
+
+public class IrrelevantTypeAnnotator extends TypeAnnotator {
+    List<TypeMirror> supportedTypes;
+    Set<TypeMirror> allFoundSupportedTypes;
```
wmdietl: What is the difference between the Set and the List version?

framework/src/org/checkerframework/framework/type/typeannotator/IrrelevantTypeAnnotator.java
```diff
@@ -0,0 +1,113 @@
...
+            AnnotatedTypeFactory typeFactory,
+            Set<? extends AnnotationMirror> annotations,
+            Class<?>[] classes) {
+        super(typeFactory);
+        this.annotations = annotations;
+
+        this.supportedTypes = new ArrayList<>(classes.length);
+        for (Class<?> clazz : classes) {
+            if (clazz.equals(Object[].class)) {
+                excludeArrays = true;
```
wmdietl: What happens if I have `String[].class`?

framework/src/org/checkerframework/framework/type/typeannotator/IrrelevantTypeAnnotator.java
```diff
@@ -0,0 +1,113 @@
...
+        } else if (typeMirror.getKind() == TypeKind.DECLARED) {
+            for (TypeMirror supportedType : supportedTypes) {
+                if (types.isSubtype(typeMirror, supportedType)) {
+                    shouldAnnotate = false;
+                    allFoundSupportedTypes.add(typeMirror);
+                    break;
+                }
+            }
+        } else if (typeMirror.getKind() == TypeKind.ARRAY) {
+            allFoundSupportedTypes.add(typeMirror);
```
wmdietl: Shouldn't the array type only be added if `excludeArrays` is true?

wmdietl:
Great new feature! Some questions, in particular about how arrays are handled.

smillst:
@wmdietl.  Thanks for the feedback.  I believe I have addressed it all in the two additional comments

framework/src/org/checkerframework/framework/type/typeannotator/IrrelevantTypeAnnotator.java
```diff
@@ -0,0 +1,125 @@
...
+                arraysAreRelevant = true;
+            } else {
+                relevantTypes.add(
+                        TypesUtils.typeFromClass(
+                                typeFactory.getContext().getTypeUtils(),
+                                typeFactory.getElementUtils(),
+                                clazz));
+            }
+        }
+        this.allFoundRelevantTypes = new HashSet<>(relevantTypes);
```
wmdietl: Should allFoundRelevantTypes be reset at some point? At the moment it looks like it can grow arbitrarily large.
Other caches are either limited in size or reset in ATF.setRoot, when the compilation unit changes.

framework/src/org/checkerframework/framework/type/treeannotator/ListTreeAnnotator.java
```diff
@@ -28,7 +31,15 @@
      */
     public ListTreeAnnotator(TreeAnnotator... annotators) {
         super(null);
-        this.annotators = Collections.unmodifiableList(Arrays.asList(annotators));
+        List<TreeAnnotator> annotatorList = new ArrayList<>();
+        for (TreeAnnotator annotator : annotators) {
```
wmdietl: Is this flattening of the sub-ListTreeAnnotators just for performance? Why is it necessary now?

framework/src/org/checkerframework/framework/type/treeannotator/ListTreeAnnotator.java
```diff
@@ -1,10 +1,13 @@
 package org.checkerframework.framework.type.treeannotator;
 
 import com.sun.source.tree.Tree;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import org.checkerframework.framework.type.AnnotatedTypeMirror;
+import org.checkerframework.framework.type.typeannotator.ListTypeAnnotator;
```
wmdietl: Why do you need *TypeAnnotator in the ListTreeAnnotator?

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -274,12 +282,34 @@ protected TreeAnnotator createTreeAnnotator() {
...
+            typeAnnotators.add(
+                    new IrrelevantTypeAnnotator(
+                            this, getQualifierHierarchy().getTopAnnotations(), classes));
+        }
+        typeAnnotators.add(new PropagationTypeAnnotator(this));
         implicitsTypeAnnotator = new ImplicitsTypeAnnotator(this);
-
-        return new ListTypeAnnotator(new PropagationTypeAnnotator(this), implicitsTypeAnnotator);
+        typeAnnotators.add(implicitsTypeAnnotator);
+        return new ListTypeAnnotator(typeAnnotators.toArray(new TypeAnnotator[0]));
```
wmdietl: Just for code readability, I would add that second constructor that takes a list directly and doesn't convert back and forth from arrays.

framework/src/org/checkerframework/framework/qual/RelevantJavaTypes.java
```diff
@@ -0,0 +1,30 @@
...
+ * type, {@code Integer}, and the unboxed, primitive type, {@code int}.
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+@Inherited
+public @interface RelevantJavaTypes {
+    /**
+     * Classes that are relevant to the checker. {@code Object[].class} implies that the checker
+     * is interested in all array types. A boxed type, such as {@code Integer.class}, implies
```
wmdietl: I would add a comment about what `Number[].class` means. At the moment the component type is ignored, if I read the code correctly.

framework/src/org/checkerframework/framework/type/treeannotator/ListTreeAnnotator.java
smillst: It just makes debugging a bit easier.

wmdietl:
I had another look through this and it all looks good. This can be merged as far as I'm concerned.
Thanks!

Issue #863, opened 2016-07-26
Layout of source code checker-qual.jar does not match layout of compiled checker-qual.jar


aw1621107:
The compiled .jar just has `org.checkerframework.*` and `plume.*` classes. The source jar does not contain anything relating to the `plume.*` classes, and has the `org.checkerframework.*` classes split across `checker/src/org/checkerframework/*` and `framework/src/org/checkerframework/`. In addition, the sources contain many classes not present in the compiled .jar, such as IDE-specific annotations.

Is there a reason behind this difference? From my limited knowledge of the Java ecosystem the compiled .jar and source .jar typically have the same layout, so I'm wondering why this bucks the trend. In addition, it confuses IDEs and requires one to manually specify the source locations.

smillst:
Thanks for bring this to our attention. The reason is that we reused the sources jar from the checker artifact so that there would be some source files for checker-qual.  (See https://github.com/typetools/checker-framework/issues/182#issuecomment-204114804 for more details.) However, we should change all the sources and javadoc jars shipped with our maven artifacts so that they match the layout in the classes jar.

mernst:
Thanks for raising the problem.  Can you clarify which compiled .jar you are talking about?

I didn't see the problem you are mentioning in either of these places:

 * I downloaded http://types.cs.washington.edu/checker-framework/current/checker-framework-2.1.1.zip, unpacked it, and examined file checker-framework-2.1.1/checker/dist/checker.jar.

 * I cloned the Checker Framework and ran "ant dist", then examined file checker-framework/checker/dist/checker.jar

smillst:
I believe the "source code checker-qual.jar" referred to by @aw1621107 is the one in Maven central: [checker-qual-2.1.1-sources.jar](http://central.maven.org/maven2/org/checkerframework/checker-qual/2.1.1/checker-qual-2.1.1-sources.jar).  This is just a copy of the jar file made via the "sources.jar" ant target in the checker directory. (See PR #642 and Issue #182 for more details.)

aw1621107:
Sorry, should have been more specific. @smillst got it.

PR #898, opened 2016-08-30
Make qual/UnitsRelations not depend on classes outside of checker-qual

smillst:
(This is a pre-step to fixing #863)


wmdietl:
See `checked` -> `check` comment from #900 
Otherwise looks good.

PR #900, opened 2016-08-30
Corrects source jars use mainly for Maven Central (Fixes 863)

### LINKED ISSUE #182: Maven artifact missing source-code and Javadoc
```
The Maven artifact currently provided is missing both source-code and Javadoc. 
This means that when we're using the IDE we cannot use code-complete to look up 
the meaning of an annotation (there is no Javadoc to look up) or walk through 
the source-code for debugging purposes.

The process at 
https://docs.sonatype.org/display/Repository/Sonatype+OSS+Maven+Repository+Usage
+Guide will generate both source-code and Javadoc for your releases 
automatically.
```

Original issue reported on code.google.com by `cow...@bbs.darktech.org` on 5 Dec 2012 at 4:03

### LINKED ISSUE #712: checker-qual.jar and checker-compat-qual.jar should be self-contained
At the moment, as raised here https://groups.google.com/d/msg/checker-framework-discuss/uCBCRVcpM_w/VuNEI-XwBQAJ, one currently has to sometimes include both `checker-qual.jar` and `checker-compat-qual.jar`.
Some classes are contained in both .jar files, causing issues.

Each `-qual.jar` file should be self-contained and contain all necessary classes.

### LINKED ISSUE #863: Layout of source code checker-qual.jar does not match layout of compiled checker-qual.jar

smillst:
checker-source.jar: layout now matches the layout of class files in checker.jar
checker-qual.jar: only contains org.checkerframework.* classes (used to contain plume classes)
    Also compiled such that classes cannot depend on other classes in the framework.
checker-qual-source.jar: new, used for Maven Central. Layout matches checker-qual.jar
checker-compat-qual.jar: only contains files in compatqual packages.
    Also compiled such that classes cannot depend on other classes in the framework.
checker-compat-qual-source.jar: new, used for Maven Central. Layout matches checker-compat-qual.jar
Fixes #863
Fixes #182
Fixes #712


wmdietl:
Had a quick look and the changes look good to me. Thanks for fixing these issues!

checker/src/org/checkerframework/checker/units/qual/UnitsRelations.java
```diff
@@ -16,5 +16,9 @@
     /**
      * @return the UnitsRelations subclass to use
      */
-    Class<? extends org.checkerframework.checker.units.UnitsRelations> value();
+    // The more precise type is Class<? extends org.checkerframework.checker.units.UnitsRelations>,
+    // but org.checkerframework.checker.units.UnitsRelations is not in checker-qual.jar, nor can
+    // it be since it uses AnnotatedTypeMirrors.  So use a less precise type and checked that it
```
wmdietl: checked -> check

framework/src/org/checkerframework/qualframework/poly/SimpleQualifierParameterAnnotationConverter.java
```diff
@@ -32,7 +33,7 @@
      * The default "Target" in an annotation is the primary qualifier.
      * We can't use null in the annotation, so we use this special value.
      */
-    public static final String PRIMARY_TARGET = "_primary";
+    public static final String PRIMARY_TARGET = DefaultValue.PRIMARY_TARGET;
```
wmdietl: I think it would be clearer to move the javadoc from this field into DefaultValue and then remove this field.

PR #908, opened 2016-09-08
Delete the Qualifier Framework

### LINKED ISSUE #122: Disallow @Regex qualifier on non-String types
```
The Regex checker allows qualifiers on any Java types, not just Strings.  The 
attached program shows how one can qualify basic types like boolean and object 
types like Boolean, which do not make sense as regular expressions.

I'll assign this to Eric, since he has the most recent experience with the 
Regex checker.

1. Compile attached program using checkers.regex.RegexChecker
2. Observe that no errors are reported.

It would be better to issue an error for any annotation on a non-String type.


```

Original issue reported on code.google.com by `charlie....@gmail.com` on 13 Jan 2012 at 10:55

Attachments:
* [NonStringRegex.java](https://storage.googleapis.com/google-code-attachments/checker-framework/issue-122/comment-0/NonStringRegex.java)

### LINKED ISSUE #488: Qualifier parameters: remove shim implementation
Here are two approaches to this:
1) Build on current system and remove the shims individually
2) Create a new project and start from scratch.

Here are the components involved:
	Visitor (Common assignment check, post-conditions)
	Dataflow (There is no shim implementation, and currently NO way to write an extended dataflow using qualifiers)
	TypeHierarchyAdapter (isSubtypeImpl, all the subtyping logic with types)
	QualifierHierarchyAdapter (Not much work)		
Replace TypeFromTree with QualifierFromTree

### LINKED ISSUE #495: Compare performance of qualifier parameter system to classic system
At the moment we have two implementations of the Regex Checker, at:

https://github.com/typetools/checker-framework/tree/master/checker/src/org/checkerframework/checker/regex
https://github.com/typetools/checker-framework/tree/master/checker/src/org/checkerframework/checker/regex/classic

and of the Tainting Checker, at:

https://github.com/typetools/checker-framework/tree/master/checker/src/org/checkerframework/checker/tainting
https://github.com/typetools/checker-framework/tree/master/checker/src/org/checkerframework/checker/tainting/classic

We should perform performance comparisons between the two implementations and use them to drive optimizations.
If/Once the performance is acceptable, we can deprecate and later remove the classic versions.

### LINKED ISSUE #618: TaintingChecker does not run under certain circumstances
~~The TaintingChecker seems to not run at all in a larger project.~~ (Edit: the larger project used classic.Untainted; switching to "normal" Untainted leaves only the ordering issue)
Trying to reproduce the issue in a toy project (https://github.com/rkraneis/checker-labs) I found that it does not run if it is listed/called after the RegexChecker (cf. [the linked toy projects pom.xml](https://github.com/rkraneis/checker-labs/blob/5919122b5b5c233c211c5fd663925f1769414d7c/pom.xml#L40L42) and [main class](https://github.com/rkraneis/checker-labs/blob/5919122b5b5c233c211c5fd663925f1769414d7c/src/main/java/io/github/rkraneis/checker/labs/NewMain1.java)). Tested with versions 1.9.11 and 1.9.12.

### LINKED ISSUE #647: Unexpected interference of RegexChecker and TaintingChecker
I'm fairly new in using the Checker Framework, so this as well may be a PEBKAC.

Wanted to create a checked version of [ParamConverterProvider](https://docs.oracle.com/javaee/7/api/javax/ws/rs/ext/ParamConverterProvider.html) with nested classes of [ParamConverter<T>](https://docs.oracle.com/javaee/7/api/javax/ws/rs/ext/ParamConverter.html), which passes the Nullness, Tainting and IGJ checkers. I ended up with the following astub:
```
public interface ParamConverter<@Nullable @Mutable @Untainted T> {
    public @Nullable @Mutable @Untainted T fromString(@Nullable @Tainted String value);
    public @Nullable @Untainted String toString(@Nullable @Mutable @Untainted T value);
}

public interface ParamConverterProvider {
    public <@Nullable @Mutable @Untainted T> @Nullable @Untainted ParamConverter<@Nullable @Mutable @Untainted T> getConverter(@Untainted Class<@Nullable @Mutable @Untainted T> rawType, @Nullable @Untainted Type genericType, Annotation @Nullable @Untainted [] annotations);
}
```
(Wonder why you need `@Nullable`? Check out [this link](http://stackoverflow.com/questions/24256095/jax-rs-paramconverter-paramconverterprovider-method-return-type-mismatch).)

Everything works perfectly with the Maven build, when invoking all the above mentioned checkers. However, as soon as I try to run the Tainting and the Regex checkers in the same build process, I get a lot of the following kind of errors:
```
[ERROR] found   : ParamConverter<T extends Object>
[ERROR]
[ERROR] required: ParamConverter<T extends Object>
```
and
```
[ERROR] found   : Type
[ERROR]
[ERROR] required: Type
```
and so on.

If I run only a bunch of checkers along with the Tainting BUT the Regex, everything is fine.
If I run only a bunch of checkers along with the Regex BUT the Tainting, everything is fine.
Any ideas, suggestions?

### LINKED ISSUE #713: Regex Checker documentation mentions incorrect annotation names
The Checker Framework manual documents the "classic" Regex Checker, contained in `org.checkerframework.checker.regex.classic'.  The  "modern" Regex Checker has different annotations; for instance, it doesn't contain `@RegexBottom`.

More confusingly, the "modern" Regex Checker uses `TOP` and `BOTTOM` from the "experimental" package at `org.checkerframework.checker.experimental.regex_qual.Regex`.

smillst:
The [commit](a04663038b93a65f455d9b97839d2380151ce3d5) that actually deletes the files and manual sections related to the Qual Framework is on the branch deleteQualFramework.  This is a pull request against that branch. These are the changes I had to make so that the tests passes.  (This should make it easier to review the changes that actually need to be reviewed.  I can post a pull request for the delete commit if desired.)

Note, the handling of Matcher.groupCount was only in the RegexQualTransfer, so I ported that code to the RegexTransfer.

Fixes #713, Fixes #618, Fixes #647, Fixes #495, Fixes #122, Fixes #488 

(Tested on daikon-typecheck)


checker/src/org/checkerframework/checker/tainting/TaintingChecker.java
```diff
@@ -13,5 +13,5 @@
  *
  * @checker_framework.manual #tainting-checker Tainting Checker
  */
-@SuppressWarningsKeys("untainted")
+@SuppressWarningsKeys({"untainted", "tainting"})
```
mernst: Do we need both of these keys?

checker/src/org/checkerframework/checker/tainting/qual/Tainted.java
```diff
@@ -13,7 +13,7 @@
  * This annotation is associated with the {@link  org.checkerframework.checker.tainting.TaintingChecker}.
  *
  * @see Untainted
- * @see  TaintingChecker
+ * @see org.checkerframework.checker.tainting.TaintingChecker
```
mernst: Is the full qualification needed?  (OK if it is.)

mernst:
This looks fine.  So does the commit that actually deletes the files and manual sections.  I say go for it and merge these.

checker/src/org/checkerframework/checker/tainting/qual/Tainted.java
smillst: Yes, qualifiers are compiled separately from the rest of the framework to ensure that checker-qual.jar is self contained.  So, if TaintingChecker is imported rather than fully-qualified, compiling checker-qual fails.

PR #912, opened 2016-09-13
Fixes #818 (Static fields in dataflow)

### LINKED ISSUE #818: Dataflow: Static field used in inner class
A static field of can be referred to with or without the classname: `field` or `MyClass.field`.  Dataflow does not recognize these as the same field in an inner class.  For example: 

````
class StaticField {
    public static @Nullable Object o = null;
    class StaticInnerClass {
        void method() {
           StaticField.o = new Object();
           o.toString(); // false positive dereference.of.nullable
        }
        void method2() {
           o = new Object();
           StaticField.o.toString();// false positive dereference.of.nullable
        }
    }
}
````
More test cases in checker/tests/nullness/flow/Issue818.java

Blocking #764

smillst:
Fixes #818

### DIFF
--- a/checker/tests/nullness/flow/Issue818.java
+++ b/checker/tests/nullness/flow/Issue818.java
@@ -1,8 +1,6 @@
 // Test case for Issue #818
 // https://github.com/typetools/checker-framework/issues/818
 
-// @skip-test
-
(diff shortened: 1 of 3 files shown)


mernst:
Looks good to me.

PR #914, opened 2016-09-13
Fixes #764 (Flow expression parsing of static fields)

### LINKED ISSUE #764: Flow Expression parser: cannot parse expression referring to static field of outer class
When type checking this code with the Nullness Checker:
````

public class Issue764 {
    public static @Nullable Object field = null;
    static class MyClass {
        @RequiresNonNull("field")
        public static void method() { }

        public void otherMethod() {
            field = new Object();
            method();
        }
        public void otherMethod2() {
            //:: error: (contracts.precondition.not.satisfied)
            method();
        }
    }
````
Expected: `contracts.precondition.not.satisfied` on the call to `method()`
Actual:  no error

 e717441 is a possible fix for this issue, ~~but it can't be fully tested because of issue general dataflow problems with referencing a static field in an inner class (#818).~~

~~Blocked by #818~~
Revert part of this comment that suppresses a warning related to this issue: https://github.com/codespecs/daikon/commit/437ca4a30c8fd8fa2489e26ae38253117640ad1c

### PR DESCRIPTION
**Pull request by smillst**
Fixes #764.

(Merge codespecs/daikon#75 after this is merged.)

### DIFF
--- a/framework/src/org/checkerframework/framework/util/FlowExpressionParseUtil.java
+++ b/framework/src/org/checkerframework/framework/util/FlowExpressionParseUtil.java
@@ -9,9 +9,11 @@ import com.sun.source.tree.MethodTree;
 import com.sun.source.tree.Tree;
 import com.sun.source.tree.VariableTree;
 import com.sun.source.util.TreePath;
+import com.sun.tools.javac.code.Symbol;
 import com.sun.tools.javac.code.Symbol.ClassSymbol;
 import com.sun.tools.javac.code.Symbol.MethodSymbol;
 import com.sun.tools.javac.code.Symbol.PackageSymbol;
+import com.sun.tools.javac.code.Type;
 import com.sun.tools.javac.code.Type.ArrayType;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: framework/src/org/checkerframework/framework/util/FlowExpressionParseUtil.java [hunk 1]
```diff
@@ -1076,6 +1078,35 @@ public FlowExpressionContext copyAndUseOuterReceiver() {
         }
     }
 
+    /**
+     * Returns the innermost enclosing type of the given type, or Type.noType if no such type was
+     * found.
```
wmdietl: Can you add some more explanation why this method is necessary?
Why is calling `DeclaredType.getEnclosingType` not the right thing in this case?
Why is this only an issue for static fields?
Why is this only an issue in `FlowExpressionParseUtil.java` and not in other uses of getEnclosingType?

smillst: Addressed here 3f692b9

PR #918, opened 2016-09-19
Remove bcel stub files

smillst:
checker/lib/bcel.jar or checker/lib/commons-bcel-6.0.jar should be used instead.


smillst:
I'm merging this to fix the plume-lib and daikon type-checking failures.

PR #921, opened 2016-09-20
Fix lost @GuardedBy refinements

smillst:
Fix Lock Checker bug where expressions lost there refinement in the @GuardedBy hierarchy if they became @LockPossiblyHeld.

### DIFF
--- a/checker/src/org/checkerframework/checker/lock/LockStore.java
+++ b/checker/src/org/checkerframework/checker/lock/LockStore.java
@@ -113,9 +113,10 @@ public class LockStore extends CFAbstractStore<CFValue, LockStore> {
      */
     private CFValue changeLockAnnoToTop(Receiver r, CFValue currentValue) {
         if (currentValue == null) {
-            currentValue =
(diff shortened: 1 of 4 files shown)


checker/src/org/checkerframework/checker/lock/LockStore.java
```diff
@@ -62,36 +70,66 @@ public LockStore leastUpperBound(LockStore other) {
...
     }
 
+    /**
+     * Makes a new CFValue with the same annotations as currentValue except that the annotation
+     * in the LockPossiblyHeld hierarchy is set to LockPossiblyHeld. If currentValue is null,
+     * then a new value is created where the annotation set is LockPossiblyHeld and GuardedByUnknown
+     */
+    private CFValue changeLockAnnoToTop(Receiver r, CFValue currentValue) {
+        if (currentValue == null) {
+            currentValue =
```
wmdietl: Instead of setting currentValue here, would it be nicer to set currentSet and currentType here, and then have shared code below? At the moment it looks a bit confusing when this new Value is used.

Issue #370, opened 2015-07-03
CLIMB-to-top and annotated libraries


GoogleCodeExporter:
```
This code

  List<@Nullable String> lns = Arrays.asList("foo", null, "bar");

should type-check because of the annotated JDK, which contains:

  class Arrays {
    public static <T> List<T> asList(T... a);
  }

That JDK annotation should be equivalent to

   public static <T extends @Nullable Object> List<T> asList(T... a);

because of the CLIMB-to-top defaulting rule.

A test case appears at 
checker-framework/checker/tests/nullness/AnnotatedJdkTest.java

```

Original issue reported on code.google.com by `michael.ernst@gmail.com` on 30 Sep 2014 at 7:16

wmdietl:
I think this issue can be closed: the reference test works, the annotated JDK is now built with the actual type systems enabled, and all annotations are stored in the bytecode.
A similar comment can be found in #924.
Is there anything left for the first task in https://github.com/typetools/checker-framework/issues/370#issuecomment-118319086 ?

mernst:
I agree.  Thanks for noticing this!

Issue #12, opened 2015-07-03
Nullness Checker makes wrong assumptions about unannotated libraries


GoogleCodeExporter:
```
The Nullness Checker issues no warnings for uses of unannotated libraries
where its methods returns null.

From Mike:


I speculate that maybe with NNEL as the default, that the Checker Framework
by default assumes all parameters and return values are non-null in
unannotated code.  This is an unfortunate default.

One idea for working around it would be to set the default to NNEL if there
is any evidence in the class of consideration of the nullness checker:
 * class was mentioned explicitly on the command line
 * import of any of its packages
 * use of any of its annotations or methods
 * use of @SuppressWarnings with any of its keys
If none of this evidence is present, the default would be "non-null except
return values are @Nullable and public fields are @PolyNull".

I'm not convinced this is a perfect solution, but it does seem to be an
improvement over the current situation.

```

Original issue reported on code.google.com by `msaeed43` on 24 Oct 2009 at 12:22

mernst:
This issue can be resolved by making
`-AuseDefaultsForUncheckedCode=-source,bytecode`
be the default behavior.  We should do experiments passing that command-line option to see how much work would remain, to make the Checker Framework usable with that configuration.

PR #924, opened 2016-09-22
Unannotated upper bounds in bytecode

smillst:
Assume that type variable upper bounds of type Object are implicit upper bounds -- that is they were not explicitly written in source code.  

Also, the default for implicit upper bounds in the Fenum Checker is now @FenumTop instead of @FenumUnqualified.

(Note this change was prompted by [this StackOverflow comment]( http://stackoverflow.com/questions/39621196/how-to-implement-micro-types-using-type-qualifiers-and-checker-framework?noredirect=1#comment66553179_39621196). But it was also proposed by @mernst  in https://github.com/typetools/checker-framework/issues/370#issuecomment-118319078 and then moved to https://github.com/typetools/checker-framework/issues/12#issuecomment-118317217, but I don't think this closes either issue.  Though maybe #370 is now fixed? )


checker/jtreg/nullness/issue824/Class2.java
```diff
@@ -1,9 +1,10 @@
 /*
  * @test
  * @summary Test case for Issue 824 https://github.com/typetools/checker-framework/issues/824
+ * The defaults for type variable upper bounds with type Object changed since the issue was filed.  So, this test case has been changed so that annotations on type variable bounds in stub files is still tested.
```
mernst: I would break this comment across lines (but keep the `@compile` commands on a single line even though they are long).

checker/tests/fenumswing/IdentityArrayList.java
```diff
@@ -1,10 +1,14 @@
 import java.util.Arrays;
+import org.checkerframework.checker.fenum.qual.FenumTop;
 
 /*
  * This test case violates an assertion in the compiler.
  * It does not depend on the Fenum Checker, it breaks for any checker.
  */
 public class IdentityArrayList {
+    // The type the third argument to Arrays.copyOf should be:
```
mernst: English typo.
Also, "should be" suggests a limitation/bug somewhere.  Can you be explicit about where that bug/limitation is?

framework/src/org/checkerframework/framework/util/defaults/QualifierDefaults.java
```diff
@@ -946,15 +947,17 @@ public Void scan(AnnotatedTypeMirror t, AnnotationMirror qual) {
                     case IMPLICIT_UPPER_BOUND:
```
mernst: This isn't your fault, but I found the lack of pre-existing documentation in the class made this code review difficult to perform.  For example, the location field that is the expression for this case statement isn't documented.  What is the relationship between apply and doApply?  (Neither has documentation.)  Could you improve the documentation, either in this pull request or in another one that we can apply before this one?

framework/src/org/checkerframework/framework/util/defaults/QualifierDefaults.java
```diff
@@ -1061,6 +1058,11 @@ protected void visitBounds(
         }
     }
 
+    /**
+     * Specifies whether the type variable or wildcard has an explicit upper bound (UPPER), an
+     * explicit lower bound (LOWER), or no explicit bounds (UNBOUNDED).
+     * lower bound
```
mernst: What is this line?

framework/src/org/checkerframework/framework/util/defaults/QualifierDefaults.java
```diff
@@ -1074,15 +1076,11 @@ protected void visitBounds(
...
         /**
-         * Neither bound is specified, BOTH are implicit
-         */
-        UNBOUND,
-
-        /**
-         * For bytecode, or trees for which we no longer have the compilation unit.
-         * We treat UNKNOWN bounds as if they are an UPPER bound.
+         * Neither bound is specified, BOTH are implicit. (If a type variable is declared in
+         * byte code and the type of the upper bound is Object, then assume that the bound was not
```
mernst: "byte code" => "bytecode"

framework/src/org/checkerframework/framework/util/defaults/QualifierDefaults.java
```diff
@@ -1124,7 +1122,7 @@ private static BoundType getTypeVarBoundType(
     }
 
     /**
-     * @return the boundType (UPPER, UNBOUND, or UNKNOWN) of the declaration of typeParamElem
+     * @return the boundType (UPPER, UNBOUNDED, or UNKNOWN) of the declaration of typeParamElem
```
mernst: UNKNOWN doesn't exist any more, so all occurrences of it should be removed from the Javadoc.

checker/tests/fenumswing/IdentityArrayList.java
```diff
@@ -1,10 +1,15 @@
...
+import org.checkerframework.checker.fenum.qual.FenumTop;
 
 /*
  * This test case violates an assertion in the compiler.
  * It does not depend on the Fenum Checker, it breaks for any checker.
  */
 public class IdentityArrayList {
+    // The type of the third argument to Arrays.copyOf should be:
+    // Class<? extends T @FenumTop []>
+    // But the annotated JDK does not have annotations for the Fenum Checker
```
mernst: Text is easier to read if you end a sentence with a period.  Thanks!

framework/src/org/checkerframework/framework/util/defaults/QualifierDefaults.java
smillst: This is done in #925.

checker/jtreg/nullness/issue824/Class2.java
mernst: In the new GitHub code review process, I'm not sure how to mark this as resolved so it no longer shows up in the conversation.  Do you know?

framework/src/org/checkerframework/framework/util/defaults/QualifierDefaults.java
mernst: Looks good, thanks.

checker/jtreg/nullness/issue824/Class2.java
smillst: As far as I can tell, you have to submit a new review.  (See https://help.github.com/articles/approving-a-pull-request-with-required-reviews/)

mernst:
Looks good.  (Though with the new code review process, I'm not sure which comments have been addressed; maybe we shouldn't use the consolidated review feature?  Thanks!

checker/jtreg/nullness/issue824/Class2.java
mernst: I found the button; it was on the red X for "required code review" at the bottom of the page where the CI test results also appear.

changelog.txt
```diff
@@ -1,5 +1,15 @@
 ---------------------------------------------------------------------------
 
+Version 2.1.4, 3 October 2016
+
+In byte code, type variable upper bounds of type Object may or may not have
```
wmdietl: "byte code" => "bytecode"

checker/tests/fenumswing/IdentityArrayList.java
```diff
@@ -1,10 +1,15 @@
 import java.util.Arrays;
+import org.checkerframework.checker.fenum.qual.FenumTop;
 
 /*
  * This test case violates an assertion in the compiler.
  * It does not depend on the Fenum Checker, it breaks for any checker.
  */
 public class IdentityArrayList {
+    // The type of the third argument to Arrays.copyOf should be:
```
wmdietl: Shouldn't this be put into the `https://github.com/typetools/checker-framework/blob/master/checker/src/org/checkerframework/checker/fenum/jdk.astub` file?

framework/src/org/checkerframework/framework/util/defaults/QualifierDefaults.java
```diff
@@ -1085,15 +1086,11 @@ protected void visitBounds(
...
         /**
-         * Neither bound is specified, BOTH are implicit
-         */
-        UNBOUND,
-
-        /**
-         * For bytecode, or trees for which we no longer have the compilation unit.
-         * We treat UNKNOWN bounds as if they are an UPPER bound.
+         * Neither bound is specified, BOTH are implicit. (If a type variable is declared in
+         * bytecode and the type of the upper bound is Object, then assume that the bound was not
```
wmdietl: The comment in parenthesis could be more explicit. Who is making an assumption? Is it trying to say that, even though there is an explicit upper bound, instead of UPPER we use UNBOUNDED for type variables from bytecode that have Object as bound?

framework/tests/all-systems/GetClassTest.java
```diff
@@ -6,7 +6,9 @@ void context() {
         Integer i = 4;
         i.getClass();
         Class<?> a = i.getClass();
+        @SuppressWarnings("fenum:assignment.type.incompatible")
```
wmdietl: A brief comment for why the supressions are ok would be useful. Or alternatively, add the `@FenumTop` to the upper bound?

Issue #643, opened 2016-04-01
Rewrite AnnotatedTypes#leastUpperBound


smillst:
AnnotatedTypes#leastUpperBound creates the LUB of the underlying types and then creates an empty AnnotatedTypeMirror using the LUB.  Then it attempts to add annotations to the empty AnnotatedTypeMirror.  During this process, the empty AnnotatedTypeMirror is treated as though it is a complete type.  This causes assertions to fail in other places in the code (#580 is one example).

AnnotatedTypes#leastUpperBound is used for three purposes and each purpose will be fixed differently:

- [ ] To compute the type of a conditional expression:  They type of the conditional expression will be the type of its assignment context.  (With some exceptions if the conditional expression is a numeric or boolean expression.) 

- [ ] To solve constraints during type argument inference: Use a corrected version of AnnotatedTypes#leastUpperBound (Fix should also fix  #775 and #906)

- [x] To merge types in dataflow: (Issues  #741 and #580): Because dataflow sometimes only has information about the primary annotation and only the refined primary annotation is ever used, dataflow operate on sets of annotations and will implement its own LUB.

smillst:
In 2013, we attempted to rewrite the LUB code see PR  #655.  Note from Stefan Heule:
>At some point we had problems that taking the LUB of two annotated types
 was not computed correctly, because we initially took the somewhat naive
 approach of taking the LUB of the underlying Java type and just adding the
 LUB of the annotations on both types.
>
 >To fix this problem, we tried to implement the same LUB strategy as javac
 does for Java types in Types and have the same computation that also
 considers annotations in AnnotatedTypes.  Unfortunately, this turned out to
 be rather challenging because the computation is not straight-forward, and
 the existing code hardly documented.  This was done in the lub branch.
>
 >In the end we abandoned this approach (we never got it working) and decided
 to fix the our existing LUB computation.  Now, we consider all necessary
 cases and also add the right annotations to element types, bounds, etc.
 This is important for taking the LUB of array types, type variables,
 wildcards and so on.  This required some special casing and trial-and-error
 to get all the behavior we want, but in the end it seemed easier than
 trying to understand and reimplement the javac LUB method.  It also seems
 to work reliably now.

smillst:
[...]

PR #928, opened 2016-09-29
The primary annotation on an intersection type is the GLB of the prim…

### LINKED ISSUE #868: AnnotatedIntersectionTypes are not defaulted correctly
given test case:

``` java
import java.util.List;
public class Test {
    <E extends @Nullable Object>
    void test (E e) {
        e.toString();
    }    
}
```

The nullness checker will give a "dereference.of.nullable" error on `e.toString()`, because the erased type of type variable `E` would have an annotation @Nullable from it's upper bound `@Nullable Object` as written in the source code.

However, if the upper bound of type variable `E` becomes an intersection type with @Nullable, as below test case showed:

``` java
import java.util.List;
public class Test {
    <E extends @Nullable Object & @Nullable List>
    void test2 (E e) {
        e.toString();
    } 
}
```

At this time nullness checker won't give any errors and pass this test case.

It seems nullness checker should be expected to give a consistent response for above two test cases, while currently it failed the first one but passed the second one.

Moreover, if add a print statement to print the `annotations` field of erased atm in `AnnotatedTypeVariable#getErased()`:

``` java
@Override
        public AnnotatedTypeMirror getErased() {
            // |T extends A&B| = |A|
            AnnotatedTypeMirror atm = this.getUpperBound().getErased();
            System.out.println("===== typeVar: " + this + "\terased: " + atm + "\terasedAnnos: " + atm.getAnnotations());
            return atm;
        }
```

The result would be:

```
first test case:
typeVar: E extends @Initialized @Nullable Object
erased: @Initialized @Nullable Object
erasedAnnos: [@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.Nullable]

second test case:
typeVar: E extends @Initialized @Nullable Object & @Initialized @Nullable List
erased: @Initialized @Nullable Object & @Initialized @Nullable List
erasedAnnos: [@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull]
```

It seems in the second test case, the erased annotations is @NonNull, but the corresponding erased type is `@Initialized @Nullable Object & @Initialized @Nullable List`.

I tried to debug it, but it turns out the logic behind `getErased()` is quite complex. Does anyone have some idea about how to fix it?

Thanks!

smillst:
The primary annotation on an intersection type is the GLB of the primary annotations of the bounds (a.k.a direct super types).  Calling addAnnotation on an annotated intersection type adds that annotation to any bound that does not have an annotation in that hierarchy. (Tested on daikon-typecheck)  

(More clean up of intersections is required in the rest of the code, but this must wait until LUB(#643) is corrected.)  

Fixes #868

### DIFF
--- a/framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java
+++ b/framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java
@@ -261,7 +261,8 @@ public abstract class AnnotatedTypeMirror {
         }
         if (atypeFactory.isSupportedQualifier(aliased)) {
             QualifierHierarchy qualHier = this.atypeFactory.getQualifierHierarchy();
-            AnnotationMirror anno = qualHier.findAnnotationInSameHierarchy(annotations, aliased);
+            AnnotationMirror anno =
+                    qualHier.findAnnotationInSameHierarchy(getAnnotations(), aliased);
             if (anno != null) {
                 return anno;
             }
@@ -304,7 +305,7 @@ public abstract class AnnotatedTypeMirror {
      *
      * @return  a unmodifiable set of the annotations on this
      */
-    public final Set<AnnotationMirror> getAnnotations() {
+    public Set<AnnotationMirror> getAnnotations() {
         return Collections.unmodifiableSet(annotations);
     }
 
@@ -2081,6 +2082,14 @@ public abstract class AnnotatedTypeMirror {
         }
     }
 
+    /**
+     * The primary annotation on an intersection type is the GLB of the primary annotations
+     * of the bounds.
+     * <p>
+     * Example (java) subtyping relationship:<br>
+     * {@code C <: A & B <: A <:Object, where class C extends A implements B {...}}
+     * <p>
+     */
     public static class AnnotatedIntersectionType extends AnnotatedTypeMirror {
 
         /**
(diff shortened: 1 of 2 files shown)


mernst:
Looks good to me.

framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java
```diff
@@ -2081,6 +2082,14 @@ public AnnotatedTypeMirror getErased() {
         }
     }
 
+    /**
+     * The primary annotation on an intersection type is the GLB of the primary annotations
+     * of the bounds.
+     * <p>
+     * Example (java) subtyping relationship:<br>
```
wmdietl: What is the `(java)` comment for?

framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java
```diff
@@ -2081,6 +2082,14 @@ public AnnotatedTypeMirror getErased() {
         }
     }
 
+    /**
+     * The primary annotation on an intersection type is the GLB of the primary annotations
+     * of the bounds.
```
wmdietl: Also say that there is no syntax to explicitly write the primary annotations in source code.

framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java
```diff
@@ -2101,6 +2110,36 @@ private AnnotatedIntersectionType(
         }
 
         @Override
+        public Set<AnnotationMirror> getAnnotations() {
+            Set<AnnotationMirror> glbs = null;
```
wmdietl: What is the relationship between the value of the `annotations` field and the result of this method?
So far, the two were interchangeable. I'm a bit fearful of removing the `final` on the parent method and allowing such overrides. You changed one use of `annotations` to the method call. Are there others that should be changed?

smillst:
Status update: I'm still considering how to address Werner's feedback. (Specifically https://github.com/typetools/checker-framework/pull/928#discussion_r81566994)

PR #938, opened 2016-10-06
Add util method for obtaining class loader

CharlesZ-Chen:
Add an util method for obtaining class loader for a class in `javacutil.InternalUtils`.

The logic is borrowed from `AnnotationClassLoader#getAppClassLoader()`:

"it will first try to obtain the class loader of the target class, if that is not available then it will try to obtain the system class loader."

Move this logic to `javacutil` could benefit checker-framework to have a unify mechanism of obtaining the class loader --- all place that might need support loading external classes should first get the `classloader` by calling this util method and then use `Class.forname(name, ..., classloader)` to load class.

We currently put this util method in `javacutil.InternalUtils` as it seems other place would be even more not suitable for this util method.


wmdietl:
Looks good to me.

PR #940, opened 2016-10-07
Migrate off deprecated javac option 'attrParseOnly'

cushon:
The option was removed as part of JDK-8148808 [1][2]. Setting
`-XDcompilePolicy=ATTR_ONLY` provides equivalent behaviour.

[1] https://bugs.openjdk.java.net/browse/JDK-8148808
[2] http://hg.openjdk.java.net/jdk9/dev/langtools/rev/645b5debcb07


mernst:
Thanks for the patch, Liam!

PR #950, opened 2016-10-13
Add new Opt utility class to provide features similar to Optional.	

wmdietl:
First version of new Opt utility class.

This currently only builds under Java 8.
Our releases currently build only for Java 7. We need to discuss how to best integrate this new feature.

I also need to add a few more tests and mention it in the manual. Let me know if you want to help with either one.


mernst:
I added Javadoc API documentation and also added references in the manual.

wmdietl:
I'm wondering whether as a first solution, it would be enough to change
https://github.com/typetools/checker-framework/blob/master/release/release_vars.py#L228
to use Java 8 for the release build.
@smillst do you think that would work?

At the moment Opt.java still uses annotations-in-comments and all our targets pass `-source 7 -target 7`, so the end result should still be Java 7 files. However, we would compile Opt.java and it would be included in the release.

In the long run, it might be nice to add a new target that compiles the Java >=8 files separately, with a Java >=8 compiler. But that doesn't seem urgent.

wmdietl:
I added a few more tests and fixed some wildcard bounds.
I think we can integrate this. The only open issue is making sure that the release includes the Opt class.

checker/build.xml
```diff
@@ -203,6 +192,8 @@
             <path>
                 <fileset dir="${src}">
                     <include name="**/*.java"/>
+                    <!-- TODO: we actually want to build these sources with a Java 8 compiler. -->
```
mernst: Can we resolve this TODO before merging?

checker/build.xml
```diff
@@ -297,6 +288,11 @@
             <package name="org.checkerframework.common.*"/>
             <package name="org.checkerframework.checker.*"/>
 
+            <fileset dir="${checker.loc}/${src}">
+                <include name="**/*.java"/>
+                <exclude name="${checker.jdk8orhigher.sources}" if="${jdk7orlower}"/>
```
mernst: How can we make sure that Opt appears in the Javadoc?  It would be better to always build the Javadoc using Java 8 if it is available.  That seems fine even if javac uses Java 7.  Furthermore, when making a release it should be required that Java 8 is available.

checker/build.xml
```diff
@@ -383,6 +379,7 @@
                 <include name="**/UnitsTools.java"/>
                 <include name="**/SignednessUtil.java"/>
                 <include name="**/I18nFormatUtil.java"/>
+                <include name="${checker.jdk8orhigher.sources}" if="${jdk8orhigher}"/>
```
mernst: I think these source files should be copied to checker-qual-source.jar regardless of the JDK version, and the file should be copied to checker-qual.jar if it can be made, which is if JDK 8 is available regardless of what the default version of the JDK is.

wmdietl: We still want to be able to build with only JDK 7 installed. It would be a larger refactoring of the build to check for all installed JDKs and build differently depending on that.
If the release is done with JDK 7 & 8 installed, we should use JDK 8 as the default, to ensure Opt is included.

checker/build.xml
wmdietl: That's what I'm suggesting https://github.com/typetools/checker-framework/pull/950#issuecomment-253926070
We can't include Opt in a JDK 7 only build, as javadoc would fail.

checker/build.xml
wmdietl: That would be a larger refactoring that shouldn't block availability of this feature.

mernst:
I think a better overall strategy would be to permit the caller to supply information about an extra JDK that is installed, and if JDK 7 and JDK 8 are both available, then use both for the appropriate files.  To simplify the logic, I think it would even be acceptable to require that JDK 8 is available to compile the Checker Framework.  (Though we shouldn't require that JDK 8 be available in order to run the Checker Framework.)

As another comment, it would ease reviewing to make independent changes in different code reviews.  For example, the change from NullnessUtils to NullnessUtil should be in a different pull request.

wmdietl:
I agree that such a restructuring of the build system would be good.
However, it is another independent improvement. You proposed the current build changes in https://github.com/typetools/checker-framework/pull/950/commits/0cec90fff14024bb79068b4516e68f0c99c5569d and I fixed them up to work for JDK 7 and 8.
We can file an enhancement request that the build system needs an improvement. But all of this is independent of the Opt changes.

mernst:
The build system changes seem necessary in order to make a release that incorporates Opt.
I think it makes sense to do the changes now, while we have the code in mind and aren't facing a release deadline, rather than doing part of the changes.  I think we will end up spending more time overall if we  delay tasks that are definitely needed until later when there is a deadline and we have forgotten about the changes.

wmdietl:
What about https://github.com/typetools/checker-framework/pull/950/commits/0cec90fff14024bb79068b4516e68f0c99c5569d will not work for the release? It will result in a checker.jar that contains Opt, documentation for Opt, and is still runnable on JDK 7.

wmdietl:
That link should have been https://github.com/typetools/checker-framework/pull/950#issuecomment-253926070

smillst:
>I'm wondering whether as a first solution, it would be enough to change
https://github.com/typetools/checker-framework/blob/master/release/release_vars.py#L228
to use Java 8 for the release build.
@smillst do you think that would work?

Doing this makes it too easy for us to screw something up and end up with JDK 8 classfiles or classfiles that use JDK 8 libraries.   

> I think a better overall strategy would be to permit the caller to supply information about an extra JDK that is installed, and if JDK 7 and JDK 8 are both available, then use both for the appropriate files

I agree with this.  We already use both JDK 7 and JDK 8 during the release.  I'll change the build files to use JDK 8 when appropriate and available.

In the meantime, I'm going to cherry-pick the irrelevant commits to master so that the squash of this pull request only include adding Opt.java.

checker/build.xml
```diff
@@ -256,6 +254,20 @@
         <delete file="${build}/.timestamp"/>
     </target>
 
+    <target name="-compile-java8" depends="-force-compile-java8"
+            description="Compiles java 8 sources if JAVA_8_HOME is set or javac is already java8">
+        <antcall target="compile-java8-sources">
```
wmdietl: Isn't this ant call redundant, as `-force-compile-java8" already compiled the same code?
For the javadoc targets, this looks nicely separated into targets that have different conditions.

checker/build.xml
```diff
@@ -22,25 +22,20 @@
 
     <property environment="env"/>
 
+    <condition property="should-force-compile-java8">
```
wmdietl: Could you add a comment for what the intuition for this variable is? It is confusing that it is checking for `jdk7orlower`, so it must mean something other than I would have expected.

checker/build.xml
```diff
@@ -438,6 +477,43 @@
         <delete dir="${checker-qual-classes-tmp}" failonerror="false"/>
     </target>
 
+    <target name="-checker-qual-java8" depends="-force-checker-qual-java8">
+        <antcall target="compile-java8-sources">
```
wmdietl: Same comment as for `-compile-java8` above.

framework/build.xml
```diff
@@ -226,6 +226,11 @@
             <package name="org.checkerframework.common.*"/>
             <package name="org.checkerframework.checker.*"/>
 
+            <fileset dir="${checker.loc}/${src}">
```
wmdietl: Does the javadoc target here not need the same modifications as in checker/build.xml?

checker/build.xml
smillst: No, -force-compile-java8 only runs if jdk7orlower is true and compile-java8-sources only runs if jdk8orhigher.  So at most only on of the targets will run.

checker/build.xml
smillst: No, same reason as above.

framework/build.xml
smillst: The javadoc target in checker/build.xml has the same modifications as the javadoc target in framework/build.xml.  Only javadoc.jar target was modified to build jdk8 sources when jdk7orlower is true.  framework/build.xml does not have a javadoc.jar target.

Issue #850, opened 2016-07-19
Improper formatting when doing git commit


CharlesZ-Chen:
I tried commit some changes on my local clone of Checker Framework.

But I get `Improper formatting: framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java` error when try to commit my local changes.

I'm sure I don't mis-introduce tab indentations, and also other formatting looks good:

```java
-            AnnotationMirror anno = qualHier.findCorrespondingAnnotation(aliased, annotations);
+
+            AnnotationMirror anno = null;
+            if (this.getKind() == TypeKind.INTERSECTION) {
+                anno = qualHier.findCorrespondingAnnotation(aliased, this.directSuperTypes().get(0).getAnnotations());
+            } else {
+                anno = qualHier.findCorrespondingAnnotation(aliased, annotations);
+            }
+
```

Does anyone know how to solve this problem? What format exactly checker framework would check before committing changes?

Thanks a lot!

smillst:
The Checker Framework follows the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html) with the exception of indenting with 4 spaces instead of 2.  To reformat the code so that you can commit changes, run `ant reformat`.

CharlesZ-Chen:
Hi smillst, I tried running `ant reformat`, but I got error:

```
/Users/charleszhuochen/Programming/UWaterloo/jsr308/checker-framework/build.xml:238: Execute failed: java.io.IOException: Cannot run program "python": error=7, Argument list too long
	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
	at java.lang.Runtime.exec(Runtime.java:620)
	at org.apache.tools.ant.taskdefs.launcher.Java13CommandLauncher.exec(Java13CommandLauncher.java:58)
	at org.apache.tools.ant.taskdefs.Execute.launch(Execute.java:428)
	at org.apache.tools.ant.taskdefs.Execute.execute(Execute.java:442)
	at org.apache.tools.ant.taskdefs.ExecTask.runExecute(ExecTask.java:629)
	at org.apache.tools.ant.taskdefs.ExecuteOn.runParallel(ExecuteOn.java:715)
	at org.apache.tools.ant.taskdefs.ExecuteOn.runExec(ExecuteOn.java:478)
	at org.apache.tools.ant.taskdefs.ExecTask.execute(ExecTask.java:496)
[...]

PR #952, opened 2016-10-20
Whole-program-inference: Increases precision of annosToIgnore map.

pbsf:
Annotations are inserted into Scenes during the whole-program-inference. Before writing those Scenes to .jaif files, annotations that should be ignored (e.g. implicit annotations) are removed from the Scenes.

This change increases the precision of annotations that should be ignored.

Previously, a map of `TypeUseLocation - > Annotation` contained all annotations that should be ignored. This means that, for example, once an `{RETURN: @Interned}` entry is inserted on this map, `@Interned` would be removed from all method return types on all Scenes.

The map is now `Pair(ATypeElement.toString(), TypeUseLocation) -> Annotation`, which has a higher precision since `ATypeElement.toString()` contains text such as the name of the method, parameters, return type and parameter types.


mernst:
This indicates incorrect formatting in one of your files:
https://travis-ci.org/typetools/checker-framework/jobs/169114392#L6200
Could you please fix the formatting?  Then we'll see the tests passing and can proceed with the review.

I'm also curious how you are able to commit to the repository with incorrect formatting, which is supposed to be impossible.  Maybe you are running on Windows?

pbsf:
I'm running on a mac. A while ago I removed the pre-commit hook in the .git/hooks folder, because I was having some problems locally with the formatter and wanted to commit my work. I remembered that now and cloned the repository from scratch, ran "ant prep" on both `checker` and `framework` directories to attempt to fix this. 

Even after that I couldn't commit the fix due to the same error as in https://github.com/typetools/checker-framework/issues/850. Suzanne's suggestion on that issue worked for me.

framework/src/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenesHelper.java
```diff
@@ -58,10 +59,10 @@
 public class WholeProgramInferenceScenesHelper {
 
     /**
-     * Maps a TypeUseLocation to a set of names of annotations that should
-     * not be added to .jaif files for that location.
+     * Maps the content of an ATypeElement and its TypeUseLocation to a set of
```
mernst: I think "content" would be clearer as "string representation".

framework/src/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenesHelper.java
```diff
@@ -248,13 +249,13 @@ private void removeIgnoredAnnosFromScene(AScene scene) {
...
         Set<Annotation> annosToRemove = new HashSet<>();
-        Set<String> annosToIgnoreForLocation = annosToIgnore.get(loc);
+        Set<String> annosToIgnoreForLocation = annosToIgnore.get(Pair.of(typeEl.toString(), loc));
         if (annosToIgnoreForLocation == null) {
             // No annotations to ignore for that position.
             return;
         }
         for (Annotation anno : typeEl.tlAnnotationsHere) {
-            if (annosToIgnoreForLocation.contains(anno.def().name)) {
+            if (annosToIgnoreForLocation.contains(anno.def().toString())) {
```
mernst: This change seems independent of the others in the pull request.  I'm not saying it's wrong, but I'm wondering whether it is an independent bug fix or is something that is related to the rest of the pull request.

mernst:
This looks good to me and can be merged when my comments are addressed.
I see a new TODO in the code; what is its priority?

framework/src/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenesHelper.java
```diff
@@ -383,11 +384,11 @@ private boolean shouldIgnore(
...
         if (implicitFor != null) {
             TypeKind[] types = implicitFor.types();
+            TypeKind atmKind = atm.getUnderlyingType().getKind();
             for (TypeKind tk : types) {
-                if (tk == atm.getKind()) {
-                    return true;
-                }
+                if (tk == atmKind) return true;
             }
+            // TODO: Handle implicitFor.literals()
```
mernst: I presume this means that if an annotation is implicit for literals, and whole program inference has inferred it for a literal, then it shouldn't be output into the source code because doing so would only produce clutter.  Is that right?
Does whole program inference ever annotate literals?  I expect it to mostly annotate method signatures rather than method calls.

mernst:
Could you please answer the one question about the TODO so the pull request can be merged?
Also, you will need to pull from upstream, since conflicts have appeared during the wait.

framework/src/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenesHelper.java
pbsf: Yes, you are right. Whole program inference never annotates literals, only variables. I confused the whole program concepts a bit.

mernst: OK.  Can you remove that line if it is not needed?  Thanks!

PR #954, opened 2016-10-26
Refactor GenericAnnotatedTypeFactory methods to be more finegrained.

luqmana:
Also move check for qualifier defaults into separate method. The motivation for this is so that we can override `ensureDefaultsExist` in checker-framework-inference. We run into the error after removing Unqualified as a supported Type Qualifier but when dealing with inference we don't need a default.

### DIFF
--- a/framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
+++ b/framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
@@ -425,7 +425,11 @@ public abstract class GenericAnnotatedTypeFactory<
     protected final QualifierDefaults createQualifierDefaults() {
         QualifierDefaults defs = new QualifierDefaults(elements, this);
         addCheckedCodeDefaults(defs);
+        addCheckedStandardDefaults(defs);
         addUncheckedCodeDefaults(defs);
+        addUncheckedStandardDefaults(defs);
+        checkForDefaultQualifierInHierarchy(defs);
+
         return defs;
     }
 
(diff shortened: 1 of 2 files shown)


wmdietl:
Looks good to me. The downstream tests failed with `The job exceeded the maximum time limit for jobs, and has been terminated.` I've restarted the build.

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -564,9 +566,32 @@ protected void addUncheckedCodeDefaults(QualifierDefaults defs) {
...
         Set<? extends AnnotationMirror> bottoms = this.qualHierarchy.getBottomAnnotations();
         defs.addUncheckedStandardDefaults(tops, bottoms);
+    }
+
+    /**
+     * Ensure that we have defaults specified.
+     *
+     * @param defs {@link QualifierDefaults} object to which defaults are added
+     */
+    protected void ensureDefaultsExist(QualifierDefaults defs) {
```
smillst: The name of the method is "ensureDefaultsExist" but the method is actually checking for a default qualifier in hierarchy.

Also, is there a reason that this is in it's own method rather than in createQualifierDefaults?  Perhaps it is expected to be overridden?

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -564,9 +566,32 @@ protected void addUncheckedCodeDefaults(QualifierDefaults defs) {
...
+     * @param defs {@link QualifierDefaults} object to which defaults are added
+     */
+    protected void addUncheckedStandardDefaults(QualifierDefaults defs) {
         Set<? extends AnnotationMirror> tops = this.qualHierarchy.getTopAnnotations();
         Set<? extends AnnotationMirror> bottoms = this.qualHierarchy.getBottomAnnotations();
         defs.addUncheckedStandardDefaults(tops, bottoms);
+    }
+
+    /**
+     * Ensure that we have defaults specified.
```
smillst: Do use "we" in comments.  Instead, you could just say "Ensure defaults are specified".  But that isn't what the method is actually doing.  It's checking that a default qualifier in (at least one) hierarchy has been set and issue an error if it has not.

framework/src/org/checkerframework/framework/util/defaults/QualifierDefaults.java
```diff
@@ -199,6 +199,20 @@ public String toString() {
     }
 
     /**
+     * Check that we have defaults specified.
```
smillst: Don't use "we".  See comment above.

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
luqmana: Yes, I made this a separate method so I could override it in checker-framework-inference where we might not necessarily have a default qualifier which is fine. Before we were adding Unqualified to our supported type qualifiers and thus sidestepped that error.

smillst:
@luqmana  Thanks for the changes; I'll merge it shortly.  

In the future, please don't rebase branches in a pull request (or force any git command).  It destroys the history and makes it impossible to tell what changes were made in response to comments.  (Also, I wasn't notified that you had made any changes to this pull request, so I didn't know to review them.)  We always [squash and merge](https://help.github.com/articles/about-pull-request-merges/#squash-and-merge-your-pull-request-commits) pull request so that master contains one commit for the pull request, so there is no need to rebase the branch.

PR #955, opened 2016-10-27
Document how to create a type alias or typedef, and tradeoffs

mernst:


checker/manual/subtyping-checker.tex
```diff
@@ -234,8 +234,110 @@ \section{Subtyping Checker example\label{subtyping-example}\label{encrypted-exam
...
+
+If you want to create a type alias or typedef, you have two options.
+Which one you choose depends on which operations you wish your new type to support.
+
+\begin{itemize}
+\item
+  Use the Fake Enum Checker (\chapterpageref{fenum-checker}) if you wish to
+  prevent all operations on the elements of the new type.  Suppose you
+  create a new fenum type \<CompassDirection> with \<int> values \<NORTH>,
+  \<SOUTH>, \<EAST>, and \<WEST>.  The operation \<NORTH + SOUTH> should be
```
smillst: This contradicts the Fenum Checker documentation: "The checker ensures that only compatible fenum types are used in comparisons and arithmetic operations (if applicable to the annotated type)."  The implementation also allows binary operations so long as both operands have the same primary annotation and it permits all string concatenations.

checker/manual/subtyping-checker.tex
```diff
@@ -234,8 +234,110 @@ \section{Subtyping Checker example\label{subtyping-example}\label{encrypted-exam
...
+
+\begin{Verbatim}
+void method(@NotMyType Object o) { ... }
+<@U extends @MyTypeUnknown Object> void use(List<U> list) {
+  // type error: list.get(0) has type @MyTypeUnknown, method expects @NotMyType
+  method(list.get(0));
+}
+\end{Verbatim}
+
+To make the code type-chekc, it is necessary to write an explicit
```
smillst: `chekc` should be `check`

checker/manual/subtyping-checker.tex
mernst: Thanks for pointing that out.  In that case, what is the use case for the Fenum Checker?  That it permits easier creation of type checkers (less code for the checker and for the annotations), albeit with less flexibility since it enforces one choice about the default?  Are there other benefits?  What guidance should we give to users?

smillst: Fenum Checker differences from the Subtyping Checker:
1. Issues warnings when fenum of different types are used in binary operations. (Excluding string concatenations) 
2. Issues warnings when the type of the expression in the case statement differs from the type of the expression in the enclosing switch statement. 
3. Unannotated types cannot be mixed with Fenum types
4. Easier to create new types

So, the use case would be to prevent operations on operands of differing types.  So both NORTH + 1 and NORTH + MONDAY should be forbidden. Though, perhaps a comparison is a better example:  NORTH == MONDAY should be forbidden.

mernst: That makes sense.  I had interpreted your previous comment differently.

PR #963, opened 2016-11-08
Infer type arguments for interdependent method invocations (fixes #953)

### LINKED ISSUE #953: Type argument inference crash on interdependent methods
Here is a fairly simple program (cut down from my real code) with a line of stream-processing code:

```
import java.util.*;
import java.util.stream.Collectors;

class CheckerTest
{
    public static void test()
    {
        List<String> initial = new ArrayList<>();
        List<Integer> counts = initial.stream().skip(1).map(l -> count("", l)).collect(Collectors.toList());
    }
    private static int count(String s, String l)
    {
        return 0;
    }
}
```

When I try to compile that using `javac -processor NullnessChecker CheckerTest.java` (version 2.1.4) I get these errors:

```
  type: Object
  superType: Integer; invoke the compiler with -AprintErrorStack to see the stack trace.
  type: @Initialized @Nullable Object
  superType: Integer; invoke the compiler with -AprintErrorStack to see the stack trace.
2 errors
```

I don't think there should be any errors: it's perfectly valid Java code which javac compiles fine when not running the annotation processor.

zenixan:


wmdietl:
Thanks for the pull request!
I've run `ant reformat` to make the code comply with style guide.
I also enabled the corresponding tests, to make sure the fix works.
This should make the Travis output more useful.
I guess Suzanne will review the changes.

mernst:
Please change the title of the pull request to be informative.  You can put "(fixes #953)" after the main title, but that should not be the entire description of the pull request.
The same advice goes for commit messages.

mernst:
Thanks for the pull request.
Right now, the [tests](https://travis-ci.org/typetools/checker-framework/builds/174183847) are failing.  I haven't double-checked, but my guess is that it's one of the two tests that were enabled in the pull request.
Can you reproduce this failure locally?  Do you know how to proceed in fixing it, or do you need help from us?
Once the tests are passing, we'll proceed with reviewing the pull request.

zenixan:
> Do you know how to proceed in fixing it, or do you need help from us?

Sorry, I have fixed the crash (`Issue953.java`) during build of my project,
but I still cannot fix issue (`Issue953b.java`) for nullness checker.

I found that a set of top annotations in the qualifier hierarchy is defined as
`[@org.checkerframework.checker.initialization.qual.UnknownInitialization, @org.checkerframework.checker.nullness.qual.Nullable]`.

Then I have checked `ConstraintMap` and found that for the any type variable
the target constraint is defined as `TargetConstraints.Equalities` with a set of top annotations:
`[@org.checkerframework.checker.initialization.qual.UnknownInitialization]`.
That's explain why `EqualitiesSolver.mergeTypesAndPrimaries()` cannot infer a type.

Unfortunately I don't have idea how to fix `EqualitiesSolver` at current moment.

zenixan:
@smillst Thank you for reviewing and moving forward issues with type inference!
Do I need try to fix this pull request or currently you are working on the proper solution?

mernst:
It would be great if you could correct the issues with this pull request.  With corrections, it makes a positive impact and will improve the Checker Framework.  It would take a lot more work to implement a different, more general solution; we plan to do that, but it won't happen soon.  We would love to have your improvements in the meanwhile.

Thanks!

mernst:
I wanted to ping you about this pull request -- not to nag, but because of our eagerness for it.
And also because other commits have conflicted with your changes in the meanwhile.  If you can merge your changes and address the code review feedback, we can get your changes into the Checker Framework.  Thanks!

zenixan:
Sorry, comments helped me to understand that chosen approach is wrong.
I have tested this pull request on the latest code from master branch and found that it gives a zero effect.
So now I'm trying another approach by copying type parameters from the parent receiver, 
i. e.`typeFactory.getPath(expressionTree).getParentPath().getLeaf()`
I think it will better to close this pull request because it does not achieve a declared goal.

mernst:
OK, that sounds fine.  If there are still problems with the Checker Framework's output (either false warnings or missed alarms), please report them.  Also, we are always happy to receive pull requests.  Looking at yours led to fixing several bugs, so even if it didn't get merged, it was extremely valuable.  Thanks again!

PR #966, opened 2016-11-14
Prevent needlessly recomputing left and right operands of binary trees

smillst:
Also, add an all-systems test for big binary trees.

### DIFF
--- a/framework/src/org/checkerframework/framework/type/treeannotator/PropagationTreeAnnotator.java
+++ b/framework/src/org/checkerframework/framework/type/treeannotator/PropagationTreeAnnotator.java
@@ -113,7 +113,7 @@ public class PropagationTreeAnnotator extends TreeAnnotator {
     @Override
     public Void visitCompoundAssignment(CompoundAssignmentTree node, AnnotatedTypeMirror type) {
         if (hasPrimaryAnnotationInAllHierarchies(type)) {
-            // If the type is already has a primary annotation in all hierarchies, then the
+            // If the type already has a primary annotation in all hierarchies, then the
             // propagated annotations won't be applied.  So don't compute them.
             return null;
         }
@@ -128,7 +128,7 @@ public class PropagationTreeAnnotator extends TreeAnnotator {
     @Override
     public Void visitBinary(BinaryTree node, AnnotatedTypeMirror type) {
         if (hasPrimaryAnnotationInAllHierarchies(type)) {
-            // If the type is already has a primary annotation in all hierarchies, then the
+            // If the type already has a primary annotation in all hierarchies, then the
             // propagated annotations won't be applied.  So don't compute them.
             // Also, calling getAnnotatedType on the left and right operands is potentially
             // expensive.
@@ -147,7 +147,7 @@ public class PropagationTreeAnnotator extends TreeAnnotator {
     @Override
     public Void visitUnary(UnaryTree node, AnnotatedTypeMirror type) {
         if (hasPrimaryAnnotationInAllHierarchies(type)) {
-            // If the type is already has a primary annotation in all hierarchies, then the
+            // If the type already has a primary annotation in all hierarchies, then the
             // propagated annotations won't be applied.  So don't compute them.
             return null;
         }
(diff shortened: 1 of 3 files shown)


framework/src/org/checkerframework/framework/type/treeannotator/PropagationTreeAnnotator.java
```diff
@@ -170,4 +190,14 @@ public Void visitTypeCast(TypeCastTree node, AnnotatedTypeMirror type) {
         }
         return null;
     }
+
+    private boolean isFullyAnnotated(AnnotatedTypeMirror type) {
```
wmdietl: I'm wondering why it is enough to only check the main modifier. For a generic type, why don't you also need to check whether the type arguments are fully annotated? Similar for other types.
Adding a comment explaining the reasoning for this would be helpful.

framework/src/org/checkerframework/framework/type/treeannotator/TreeAnnotator.java
```diff
@@ -32,4 +33,20 @@ public TreeAnnotator(AnnotatedTypeFactory atypeFactory) {
...
     }
+
+    /**
+     * When overriding this method, getAnnotatedType on the left and right operands should only be
+     * called when absolutely necessary. Otherwise, the checker will be very slow on heavily nested
+     * binary trees. (For example, a + b + c + d + e + f + g + h.)
+     *
+     * <p>If a checker's performance is still too slow, the types of binary trees could be computed
+     * in a subclass of {@link org.checkerframework.framework.flow.CFTransfer}. When computing the
+     * types in a transfer, look up in the value in the store rather than the AnnotatedTypeFactory.
```
wmdietl: `in the value` -> `the value`

framework/src/org/checkerframework/framework/type/treeannotator/TreeAnnotator.java
```diff
@@ -32,4 +33,20 @@ public TreeAnnotator(AnnotatedTypeFactory atypeFactory) {
...
+     * Then this method should annotated binary trees with top so that the type applied in the
```
wmdietl: `annotated` -> `annotate` ?

framework/src/org/checkerframework/framework/type/treeannotator/TreeAnnotator.java
```diff
@@ -32,4 +33,20 @@ public TreeAnnotator(AnnotatedTypeFactory atypeFactory) {
...
+     * transfer is always a subtype of the type the AnnotatedTypeFactory computes.
+     */
+    @Override
+    public Void visitBinary(BinaryTree node, AnnotatedTypeMirror mirror) {
```
wmdietl: In the PropagationTreeAnnotator above you also override other methods and check whether the types are fully annotated. Should there be similar comments like this one for these other methods?

framework/tests/all-systems/BigBinaryTrees.java
```diff
@@ -0,0 +1,271 @@
+// Test to check for common problem with BinaryTrees.
+// See comment on TreeAnnotator#visitBinary.
```
wmdietl: Isn't the problem mostly a timing issue, i.e. the checker will run very slowly on this example?
Or might checkers fail with a stackoverflow or some other exception?
A note would help to see whether any manual action/check is necessary.

framework/src/org/checkerframework/framework/type/treeannotator/PropagationTreeAnnotator.java
```diff
@@ -112,6 +112,11 @@ public Void visitNewArray(NewArrayTree tree, AnnotatedTypeMirror type) {
 
     @Override
     public Void visitCompoundAssignment(CompoundAssignmentTree node, AnnotatedTypeMirror type) {
+        if (hasPrimaryAnnotationInAllHierarchies(type)) {
+            // If the type is already has a primary annotation in all hierarchies, then the
```
wmdietl: Remove first `is`.

framework/tests/all-systems/BigBinaryTrees.java
```diff
@@ -0,0 +1,271 @@
+// Test to ensure that checkers can type-check big binary trees in an
+// exceptable amount of time.  See comment on TreeAnnotator#visitBinary.
```
wmdietl: exceptable -> acceptable

PR #970, opened 2016-11-20
Split formatting to allow re-use from down-stream projects

wmdietl:
See https://github.com/typetools/checker-framework-inference/pull/44 for example use.


mernst:
Looks fine to me.  Thanks!

PR #971, opened 2016-11-20
Use REPO_SLUG to ensure dependencies stay within repository organization

wmdietl:


.travis-build-without-test.sh
```diff
@@ -5,12 +5,14 @@ set -e
...
 
+SLUGOWNER=${TRAVIS_REPO_SLUG%/*}
+
 ## Build annotation-tools (Annotation File Utilities)
 if [ -d ../annotation-tools ] ; then
     # Older versions of git don't support the -C command-line option
     (cd ../annotation-tools && git pull)
 else
-    (cd .. && git clone --depth 1 https://github.com/typetools/annotation-tools.git)
+    (cd .. && git clone --depth 1 https://github.com/${SLUGOWNER}/annotation-tools.git)
```
smillst: This forces people with forks of checker-framework to have a fork of annotation-tools and to make sure it is kept up-to-date.  Could you add a check to see if  https://github.com/${SLUGOWNER}/annotation-tools.git exists and use typetools if it does not?

PR #987, opened 2016-12-12
Improve check for whether the found and required types should be printed verbosely.

### LINKED ISSUE #974: Don't print two different types identically in an error message
Issue #973 shows some Checker Framework output:

```
SuperConsumer.java:10: error: [argument.type.incompatible] incompatible types in argument.
  found   : List<Object>
  required: ? extends Object super List<Object>
```

Note that there are two occurrences of `List<Object>` in the output.

Running the test case with the `-AprintAllQualifiers` command-line option gives a less confusing error messages, because it becomes clear that the two different occurrences of `List<Object>` in the error message are actually two different types:

```
SuperConsumer.java:10: error: [argument.type.incompatible] incompatible types in argument.
        consumer.accept(list);
                        ^
  found   : @UnknownKeyFor List<@UnknownKeyFor Object>
  required: ? extends @UnknownKeyFor Object super @KeyForBottom List<@KeyForBottom Object>
```

The Checker Framework should never print an error message in which two different types have the same printed representation.
 * One way to do this would be to always print types under the assumption of standard defaulting, even though the Checker Framework uses slightly different defaulting rules in different locations.
 * Another way to do this would be to scan all the types in the error message and turn on `-AprintAllQualifiers` or `-AprintVerboseGenerics` if there are any two different types with the same printed representation.
 * There may be other possible solutions as well.

smillst:
Fixes #974

### DIFF
--- a/framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
+++ b/framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
@@ -1840,14 +1840,10 @@ public class BaseTypeVisitor<Factory extends GenericAnnotatedTypeFactory<?, ?, ?
      * Return whether or not the verbose toString should be used when printing the two annotated
      * types.
      *
-     * <p>If any simple toString of any annotated type in the first atm or second atm is equal to
-     * any toString of any annotated type in the first or second atm and if the verbose toStrings
-     * differ, then the verbose toString should be used.
-     *
-     * @param atm1 the first atm
-     * @param atm2 the second atm
-     * @return whether or not the verbose toString should be used when printing the two annotated
-     *     types.
+     * @param atm1 the first AnnotatedTypeMirror
+     * @param atm2 the second AnnotatedTypeMirror
+     * @return true iff there are two annotated types (in either ATM) such that their toStrings are
+     *     the same but their verbose toStrings differ
      */
     private boolean shouldPrintVerbose(AnnotatedTypeMirror atm1, AnnotatedTypeMirror atm2) {
         String atm1ToString = atm1.toString();
@@ -1861,6 +1857,7 @@ public class BaseTypeVisitor<Factory extends GenericAnnotatedTypeFactory<?, ?, ?
 
         SimpleAnnotatedTypeScanner<Boolean, Void> checkForMismatchedToStrings =
                 new SimpleAnnotatedTypeScanner<Boolean, Void>() {
+                    /** Maps from a type's toString to its verbose toString */
                     Map<String, List<AnnotatedTypeMirror>> map = new HashMap<>();
 
                     @Override
@@ -1906,7 +1903,7 @@ public class BaseTypeVisitor<Factory extends GenericAnnotatedTypeFactory<?, ?, ?
         if (result1 != null && result1) {
             return true;
         }
-        // Call reset to clear the visitor history, but not the map for Strings to types.
+        // Call reset to clear the visitor history, but not the map
         checkForMismatchedToStrings.reset();
         Boolean result2 = checkForMismatchedToStrings.visit(atm2);
         return result2 == null ? false : result2;
--- a/framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
+++ b/framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
@@ -1862,17 +1862,9 @@ public class BaseTypeVisitor<Factory extends GenericAnnotatedTypeFactory<?, ?, ?
 
                     @Override
                     protected Boolean reduce(Boolean r1, Boolean r2) {
-                        if (r1 == null) {
-                            if (r2 == null) {
-                                return false;
-                            } else {
-                                return r2;
-                            }
-                        } else if (r2 == null) {
-                            return r1;
-                        } else {
-                            return r1 || r2;
-                        }
+                        r1 = r1 == null ? false : r1;
+                        r2 = r2 == null ? false : r2;
+                        return r1 || r2;
                     }
 
                     @Override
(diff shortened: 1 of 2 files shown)


framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1822,19 +1823,100 @@ protected void commonAssignmentCheck(
...
+                                    String otherVerbose = other.toString(true);
+                                    if (!verbose.equals(otherVerbose)) {
+                                        return true;
+                                    }
+                                }
+                                return false;
+                            }
+                        };
+        Boolean result = checkToStringOfOther.visit(atm1, map);
+        return result == null ? false : result;
```
mernst: When would the result be null?

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1822,19 +1823,100 @@ protected void commonAssignmentCheck(
...
+                    Map<String, List<AnnotatedTypeMirror>> map = new HashMap<>();
+
+                    @Override
+                    protected Map<String, List<AnnotatedTypeMirror>> defaultAction(
+                            AnnotatedTypeMirror type, Void aVoid) {
+                        if (type == null) {
+                            return map;
+                        }
+                        String simple = type.toString();
+                        List<AnnotatedTypeMirror> list;
```
mernst: Is there a library that would abstract out the process of adding to a multimap?  (Maybe not one that we wish the CF to depend upon?)

I think we can use the same visitor for both ATMs, and that doing so will not only shorten the code but will solve a problem with the current design.

It is a problem if any two types anywhere in the message are printed the same but different under the hood.  The current pull request uses verbose printing if two types in different ATMs have that property, but not if two types in the same ATM have that property.  I think it would be simpler and safer to use the same visitor for both ATMs.  It would always check for duplicates (causing `shouldPrintVerbose` to return true if there is any duplicate) and always add to the map if there isn't a duplicate.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1822,19 +1823,95 @@ protected void commonAssignmentCheck(
...
+                                list.add(type);
+                                return true;
+                            }
+                        }
+                        list.add(type);
+                        return false;
+                    }
+                };
+        Boolean result1 = checkForMismatchedToStrings.visit(atm1);
+        // SimpleAnnotatedTypeScanner#scan returns null is it encounters a null AnnotatedTypeMirror
```
mernst: is -> if

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1822,19 +1823,95 @@ protected void commonAssignmentCheck(
...
+        if (result1 != null && result1) {
+            return true;
+        }
+        // Call reset to clear the visitor history, but not the map for Strings to types.
```
mernst: for -> from

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1822,19 +1823,95 @@ protected void commonAssignmentCheck(
...
+        checkForMismatchedToStrings.reset();
+        Boolean result2 = checkForMismatchedToStrings.visit(atm2);
+        return result2 == null ? false : result2;
```
mernst: Can you add a comment, somewhere, indicating under what condition visit returns null?

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1822,19 +1823,95 @@ protected void commonAssignmentCheck(
...
+            return true;
+        }
+
+        SimpleAnnotatedTypeScanner<Boolean, Void> checkForMismatchedToStrings =
+                new SimpleAnnotatedTypeScanner<Boolean, Void>() {
+                    Map<String, List<AnnotatedTypeMirror>> map = new HashMap<>();
+
+                    @Override
+                    protected Boolean reduce(Boolean r1, Boolean r2) {
+                        if (r1 == null) {
```
mernst: I think it's clearer and shorter to do:
```
if (r1==null) r1=false;
if (r2==null) r2=false;
return r1 || r2;
```

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
mernst: Should null AnnotatedTypeMirrors happen?

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1822,19 +1823,95 @@ protected void commonAssignmentCheck(
...
+                        if (type == null) {
+                            return false;
+                        }
+                        String simple = type.toString();
+                        List<AnnotatedTypeMirror> list = map.get(simple);
+                        if (list == null) {
+                            list = new ArrayList<>();
+                            map.put(simple, list);
+                        }
+                        String verbose = type.toString(true);
```
mernst: Please comment that this isn't List.contains because it's testing toString().equals() instead of equals().

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1822,19 +1823,85 @@ protected void commonAssignmentCheck(
...
+        }
+
+        SimpleAnnotatedTypeScanner<Boolean, Void> checkForMismatchedToStrings =
+                new SimpleAnnotatedTypeScanner<Boolean, Void>() {
+                    /** Maps from a type's toString to its verbose toString */
+                    Map<String, String> map = new HashMap<>();
+
+                    @Override
+                    protected Boolean reduce(Boolean r1, Boolean r2) {
+                        if (r1 == null) {
```
mernst: Wouldn't it be easier to write this code as:
```
if (r1 == null) r1 = false;
if (r2 == null) r2 = false;
return r1 || r2;
```
?

(You did a similar change, but in a place where it undoes short-circuiting.  Did you mean to do it here?)

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1822,19 +1823,85 @@ protected void commonAssignmentCheck(
...
+                        }
+                    }
+                };
+        Boolean r1 = checkForMismatchedToStrings.visit(atm1);
+        // Call reset to clear the visitor history, but not the map from Strings to types.
+        checkForMismatchedToStrings.reset();
+        Boolean r2 = checkForMismatchedToStrings.visit(atm2);
+
+        // SimpleAnnotatedTypeScanner#scan returns null if it encounters a null AnnotatedTypeMirror.
+        // This shouldn't happen if the atm1 and atm2 are well-formed.
```
mernst: If they are ill-formed, I feel it is better for the Checker Framework to abort with an error (so that we get a bug report) than to continue silently.  Is there a reason it's better to ignore the error, such as that it will certainly be detected elsewhere and the CF will halt there?

smillst: If the ATM wasn't wellformed it would likely cause a crash elsewhere. I just want to prevent a crash report from a user here, because it would be very hard to debug.  But I really don't think the result will ever actually be null.

PR #991, opened 2016-12-14
Properly format the Eclipse plug-in sources.

wmdietl:
Let's also properly format the Eclipse plug-in sources.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -137,7 +137,7 @@
 
   <fileset id="formatted.java.files" dir="."
            includes="**/*.java"
-           excludes="**/checker/jdk/,**/stubparser/,**/eclipse/,**/nullness-javac-errors/,**/annotated/"/>
+           excludes="**/checker/jdk/,**/stubparser/,**/nullness-javac-errors/,**/annotated/"/>
 
   <import file="./build-styleguide.xml"/>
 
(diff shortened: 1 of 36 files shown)


mernst:
This looks reasonable to me.  Thanks!

I have two questions/concerns, which is why I didn't format the Eclipse code in the first place.
 * Was the original version of the Eclipse code copied from somewhere else, such as generated by a tool?  In that case, it would be best to retain that formatting, to make it easier to compare and to see the changes we have made.
 * What is the code style of the Eclipse project?  If the argument for formatting the Checker Framework proper was to maintain consistency with javac, then it would make sense to format the Eclipse plugin similarly to how Eclipse is formatted.

wmdietl:
Thanks for the comments!

- I didn't find any comments that would indicate that the .java files were auto-generated and I didn't find any source files from which they would be generated. The developer readme also contains no indication that any of the .java files are auto-generated.

- The code style for Eclipse seems to be the same as the Oracle style: https://wiki.eclipse.org/Coding_Conventions
However, most of the code is in the style that Jonathan liked.
I think for consistency we can reformat the Eclipse files.

Regarding the other excludes:
-  `**/checker/jdk/` I think it's fine to not reformat the annotated JDK files. This directory will be removed eventually, anyways.

- `**/stubparser/` This contains mostly auto-generated files, so we should not reformat. (Should the pattern be more specific, to not include all stubparser directories?)

- `**/nullness-javac-errors/` Couldn't we reformat this test case? If not, it would be good to add to the README why.

- `**/annotated/` Why do we need this? I assume it is supposed to match `checker/jdk/annotated`, but that directory only contains .class files.

mernst:
Thanks for the info.  Let's reformat, then.  I agree with your other suggestions, too.  Thanks!

PR #996, opened 2016-12-20
Update the Lock Checker to use the expression annotation support.

smillst:
The Lock Checker hacked BaseTypeVisitor's precondition code to issue a contracts.precondition.not.satisfied.field error when a lock was not held. This commit removes that hack from BaseTypeVisitor and a "lock.not.held" error is issued instead.

(Daikon type-checks with this change and takes about the same amount of time.)


smillst:
@wmdietl I add some more documentation.  Do you want to review again or should I just merge?

PR #997, opened 2016-12-21
Unsignedness checker

mernst:


mernst:
I found the method names confusing.  For example, the method that was named `isMaskedShift` took two literals as arguments, but no masking or shifting operations.  I have attempted to rename the methods and document them, but I may have gotten this wrong.  Please review the documentation and naming and correct them.  Also, please give concrete code examples in the Javadoc for the two major methods that you added.  Without a clear specification, it's hard for me to determine whether the code is correct.

The `MaskedShift.java` tests contain only positive tests -- places where the test is supposed to succeed.  Could you please add some negative tests as well?  These are places where you expect the Signedness Checker to issue a warning.

Also, please make the tests include at least a couple of each of the following varieties of expressions:
 * both parenthesized and unparenthesized expressions
 * both AND and OR masking operations
 * masking operations with the shift on the right-hand side and the left-hand side
 * left-shift operations

The OR logic looks wrong to me.  An OR mask would need to involve negation in order to be ignorable, right?  As stated, the test is for the OR mask to contain 1, which has no effect.

checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -23,6 +22,124 @@ public SignednessVisitor(BaseTypeChecker checker) {
...
+    }
+
+    /**
+     * Given a masking operation of the form {@code x & maskLit} or {@code x | maskLit}, return true
+     * iff the masking operation results in the same output regardless of the value of the numBits
+     * most significant bits of x. This is if the numBits most significant bits of mask are 0 for
+     * AND, and 1 for OR. For example, assuming that numBits is 4, the following is true about AND
+     * and OR masks:
+     *
+     * <p>{@code x & 0x0... == 0x0...;} {@code x | 0xF... == 0xF...;}
```
mernst: It's slightly confusing to use `x` as the variable when you use the letter x in the constant `0x0...`.
Also, x can be an arbitrary expression, but `x` suggests it's a variable; "expr" might be better.
Also, the ellipsis is different on the left and right side of the `==`, but I'm not sure exactly how to express that fact which took me a moment to realize.

checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -23,6 +22,124 @@ public SignednessVisitor(BaseTypeChecker checker) {
...
+     */
+    private boolean maskIgnoresMSB(Kind maskKind, LiteralTree numBitsLit, LiteralTree maskLit) {
+        long numBits = getLong(numBitsLit.getValue());
+        long mask = getLong(maskLit.getValue());
+
+        // If maskLit was an int, then shift mask so that the numBits most significant bits are in the right position
+        if (maskLit.getKind() != Kind.LONG_LITERAL) {
+            mask <<= 32;
+        }
+
```
mernst: I would remove this line, but add a blank line between lines 74 and 75.  The block from 68 to 74 has the goal of putting the numBits most significant bits as the least significant bits, zeroing out the rest.  Say that from the beginning.  I think it would be better to use one operation for each, one on each arm of the if on line 69 -- I think that would be simpler than the current flow.

checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -23,6 +22,124 @@ public SignednessVisitor(BaseTypeChecker checker) {
...
+        // Shift the numBits most significant bits to become the numBits least significant bits, zeroing out the rest
+        mask >>>= (64 - numBits);
+        if (maskKind == Kind.AND) {
+            // Check that the numBits most significant bits of the mask were 0
+            return mask == 0;
+        } else if (maskKind == Kind.OR) {
+            // Check that the numBits most significant bits of the mask were 1
+            return mask == (1 << numBits) - 1;
+        } else {
+            return false;
```
mernst: Is this an expected condition or an error?  If expected, you could put the test early in the method to avoid unnecessary computation.  But I think it's impossible, in which case rather than returning false it's better to terminate the program with an error or at least print some kind of diagnostic to help Checker Framework developers discover and fix the problem.  If you write a condition in the code, that implies to a reader that it is possible -- make clear if it isn't, and do something appropriate.

checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -23,6 +22,124 @@ public SignednessVisitor(BaseTypeChecker checker) {
...
+        }
+    }
+
+    /**
+     * Determines if a right shift operation, {@code >>} or {@code >>>}, is masked with a masking
+     * operation of the form {@code shiftOp & maskLit} or {@code shiftOp | maskLit} such that the
+     * mask renders the shift signedness irrelevent by destroying the bits introduced by the shift.
```
mernst: Clarify "introduced by": you mean sign bits duplicated into the shift result.
Also, clarify "shift signedness" with a parenthesis: "(>> vs >>>)".

checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -23,6 +22,124 @@ public SignednessVisitor(BaseTypeChecker checker) {
...
+     * For example, the following pairs of right shifts on {@code byte b} both produce the same
+     * results under any input, because of their masks:
+     *
+     * <p>{@code (b >> 4) & 0x0F == (b >>> 4) & 0x0F;} {@code (b >> 4) | 0xF0 == (b >>> 4) | 0xF0;}
```
mernst: Put these examples on separate lines or in separate paragraphs, for readability.

checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -23,6 +22,124 @@ public SignednessVisitor(BaseTypeChecker checker) {
...
+     *
+     * @param shiftOp a right shift operation: {@code >>} or {@code >>>}
```
mernst: This comment isn't right:  shiftOp is something like `expr1 >> expr2`, not just `>>`, right?

Separately, "shiftOp" sounds like an operator rather than an expression or tree; I would rename the variable.  The same goes for maskOp, which sounds to me like the operator itself.

checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -23,6 +22,124 @@ public SignednessVisitor(BaseTypeChecker checker) {
...
+     * @return true iff the right shift is masked such that a signed or unsigned right shift has the
+     *     same effect
+     */
+    private boolean isMaskedShift(BinaryTree shiftOp) {
+        // parent is the operation or statement that immediately contains shiftOp
```
mernst: "immediately" is confusing.  Maybe it's enough to add "not counting parentheses." at the end.

checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -23,6 +22,124 @@ public SignednessVisitor(BaseTypeChecker checker) {
...
+        Tree parent;
+        // topChild is the top node in the chain of nodes from shiftOp to parent
```
mernst: Is this either shiftOp or the outermost parenthesis -- than is, a direct child of `parent`?

checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -23,6 +22,124 @@ public SignednessVisitor(BaseTypeChecker checker) {
...
+        Tree topChild;
```
mernst: If this is a child of parent, maybe it's better for its name to refer to that.
Or, you could rename "parent" to "enclosing" and then this would be "enclosingChild".

checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -23,6 +22,124 @@ public SignednessVisitor(BaseTypeChecker checker) {
...
+        }
+
+        if (!isMask(parent)) {
+            return false;
+        }
+
+        BinaryTree maskOp = (BinaryTree) parent;
+        ExpressionTree shiftExpr = shiftOp.getRightOperand();
+
+        // Determine which child of maskOp leads to shiftOp. The other one is the mask expression
```
mernst: Throughout, whenever you have a full sentence, terminate it with punctuation (here, a period).  Otherwise, it's too confusing to read -- I was taken aback and had to slow down and reread to realize there was nothing missing here.

mernst:
Looks great!  Thanks.

Issue #30, opened 2015-07-03
@Raw errors in places in which everything is @NonRaw


GoogleCodeExporter:
```
What steps will reproduce the problem?

interface Consumer<A extends /*@Nullable*/ Object> {
  public void consume(A object);
}

class Utils {
  public static <A extends /*@Nullable*/ Object>
      Consumer<A> cast(final Consumer<? super A> consumer) {
    return new Consumer<A>() {
      @Override public void consume(A object) {
        consumer.consume(object);   // error 1: see issue 28 for details
      }
    };
  }
  public static <A extends /*@Nullable*/ Object> Consumer<A> 
      getConsumer() {
    // null for simplicity, but could be anything
    Consumer</*@Nullable*/ Object> nullConsumer = null;  
    return cast(nullConsumer);  // error2: this issue
  }
}

What is the expected output? What do you see instead?

Ignore error 1 (see issue 28 instead).
Error 2: I get

 incompatible types.
   return cast(nullConsumer);
              ^
 found   : @NonRaw Consumer<@Raw A extends @Raw Object>
 required: @NonRaw Consumer<@NonRaw A extends @NonRaw Object>

I don't understand what @Raw/@NonRaw has to do with anything here.
I'd expect this code to compile cleanly.

What version of the product are you using? On what operating system?

Linux, javac 1.7.0-jsr308-1.0.1


Please provide any additional information below.

```

Original issue reported on code.google.com by `piotr.zi...@gmail.com` on 3 Dec 2009 at 5:08

PR #999, opened 2016-12-21
Merge in all index checker changes.

kelloggm:
I'm creating this pull request that asks for the current state of the index checker (on my fork) to be merged into a branch on typetools. The goal is to give Suzanne something to review.

This pull request includes all three Index Checker subcheckers: minlen, lowerbound, and upperbound.

Although the manual is included here, it's definitely out-of-date and needs to be updated. That's on my todo list, so it will happen _eventually_, but I wanted to let Suzanne start looking at some of the code instead of making her wait for me to finish things.


checker/src/org/checkerframework/checker/upperbound/UpperBoundStore.java
```diff
@@ -0,0 +1,157 @@
...
+            AnnotatedTypeFactory atypeFactory,
+            Receiver caller) {
+
+        UpperBoundAnnotatedTypeFactory factory = (UpperBoundAnnotatedTypeFactory) atypeFactory;
+        CFValue value = this.getValue(rec);
+        Set<AnnotationMirror> atm = value.getAnnotations();
+        if (AnnotationUtils.containsSameByClass(atm, LTEqLengthOf.class)) {
+            AnnotationMirror anno = AnnotationUtils.getAnnotationByClass(atm, LTEqLengthOf.class);
+            String[] vals = UpperBoundUtils.getValue(anno);
+            if (vals.length != 1 || !vals[0].equals(caller.toString())) {
```
smillst: `vals[0].equals(caller.toString())` is important to the correctness of this method.  Please move it to an if block surrounding the code that is executed if true.  In it's current location it was diffacult for me to understand what was going on.

checker/src/org/checkerframework/checker/upperbound/UpperBoundStore.java
```diff
@@ -0,0 +1,157 @@
...
+        super(other);
+    }
+
+    public UpperBoundStore(
+            CFAbstractAnalysis<CFValue, UpperBoundStore, ?> analysis, boolean sequentialSemantics) {
+        super(analysis, sequentialSemantics);
+    }
+
+    // If something is removed from a list it reduces the minlen of anything that could be an alias of the list by 1.
+    // If a list is cleared, anything that could be an alias of this list goes to UpperBoundUnknown.
```
smillst: These comments are out-of-date.  Instead of a comment here.  I would add a comment for each if block in the method itself.

checker/src/org/checkerframework/checker/upperbound/UpperBoundStore.java
```diff
@@ -0,0 +1,157 @@
...
+        if (add) {
+            for (FlowExpressions.LocalVariable rec : localVariableValues.keySet()) {
+                applyAdd(rec, replace, atypeFactory, caller);
+            }
+            for (FieldAccess rec : fieldValues.keySet()) {
+                applyAdd(rec, replace, atypeFactory, caller);
+            }
+        }
+        for (Receiver rec : replace.keySet()) {
+            replaceValue(rec, replace.get(rec));
```
smillst: Is there reason why each method doesn't call replaceValue directly?  (If so, please add a comment a loop through the  entry set so you don't have to call Map.get.)

checker/src/org/checkerframework/checker/upperbound/UpperBoundStore.java
```diff
@@ -0,0 +1,157 @@
...
+        } else if (AnnotationUtils.containsSameByClass(atm, LTLengthOf.class)) {
+            AnnotationMirror anno = AnnotationUtils.getAnnotationByClass(atm, LTLengthOf.class);
+            String[] vals = UpperBoundUtils.getValue(anno);
+            if (vals.length != 1 || !vals[0].equals(caller.toString())) {
+                return;
+            }
+            AnnotationMirror newAnno = factory.createLTOMLengthOfAnnotation(caller.toString());
+            CFValue val =
+                    analysis.createSingleAnnotationValue(
+                            factory.getQualifierHierarchy().greatestLowerBound(newAnno, anno),
```
smillst: Why glb?

checker/src/org/checkerframework/checker/upperbound/UpperBoundTransfer.java
```diff
@@ -0,0 +1,440 @@
...
+    // Refine the type of expressions used as an array dimension to be
+    // less than length of the array to which the new array is
+    // assigned.
+    @Override
+    public TransferResult<CFValue, UpperBoundStore> visitAssignment(
+            AssignmentNode node, TransferInput<CFValue, UpperBoundStore> in) {
+        AnnotationMirror UNKNOWN = atypeFactory.UNKNOWN;
+        TransferResult<CFValue, UpperBoundStore> result = super.visitAssignment(node, in);
+
+        // When an existing array is assigned into, the store needs to be cleared -
```
smillst: Is this even working? Issue kelloggm/checker-framework#30 is still open and the [test case](https://github.com/kelloggm/checker-framework/blob/master/checker/tests/upperbound/Reassignment.java) isn't issuing an error.  Should this code just be removed until something sensible can be implemented?

checker/src/org/checkerframework/checker/upperbound/UpperBoundTransfer.java
```diff
@@ -0,0 +1,440 @@
...
+            // This means that the existing store needs to be invalidated.
+            // As far as I can tell the easiest way to do this is to just
+            // create a new TransferResult.
+            TransferResult<CFValue, UpperBoundStore> newResult =
+                    new RegularTransferResult<CFValue, UpperBoundStore>(
+                            result.getResultValue(), new UpperBoundStore(analysis, true));
+            result = newResult;
+        }
+
+        // This handles when a new array is created.
```
smillst: This comment isn't helpful.  How are new arrays handled?  Ok, now I see that comment on the method should be moved here.

checker/src/org/checkerframework/checker/upperbound/UpperBoundVisitor.java
```diff
@@ -0,0 +1,131 @@
...
+    public UpperBoundVisitor(BaseTypeChecker checker) {
+        super(checker);
+        this.IndexFirstArgListMethods = new ArrayList<ExecutableElement>();
+        this.IndexFirstArgListMethods.add(
+                TreeUtils.getMethod(
+                        "java.util.List", "get", 1, checker.getProcessingEnvironment()));
+        this.IndexFirstArgListMethods.add(
+                TreeUtils.getMethod(
+                        "java.util.List", "set", 2, checker.getProcessingEnvironment()));
+        //// can't handle until TreeUtils.getMethod has a way to precisly handle method overloading
```
smillst: Is there a typetools/checker-framework issue for this?  This seem pretty easy to implement.

checker/src/org/checkerframework/checker/upperbound/UpperBoundVisitor.java
```diff
@@ -0,0 +1,131 @@
...
+        // less than the minimum length of the array. If it could be any
+        // of several values, only the max is of interest.
+        Integer valMax = atypeFactory.valMaxFromExpressionTree(indexTree);
+        Integer minLen = atypeFactory.minLenFromExpressionTree(arrTree);
+
+        // Is indexType LTL/LTOM of a set containing arrName?
+        if ((indexType.hasAnnotation(LTLengthOf.class)
+                        || indexType.hasAnnotation(LTOMLengthOf.class))
+                && (UpperBoundUtils.hasValue(indexType, arrName)
+                        || UpperBoundUtils.hasValue(indexType, arrTree.toString()))) {
```
smillst: Why the comparison against arrTree.toString()?  Is there something wrong with the ExpressionAnnotationSupport?

checker/src/org/checkerframework/checker/upperbound/UpperBoundVisitor.java
```diff
@@ -0,0 +1,131 @@
...
+import org.checkerframework.common.basetype.BaseTypeVisitor;
+import org.checkerframework.dataflow.analysis.FlowExpressions;
+import org.checkerframework.framework.source.Result;
+import org.checkerframework.framework.type.AnnotatedTypeMirror;
+import org.checkerframework.javacutil.TreeUtils;
+
+/** Warns about array accesses that could be too high. */
+public class UpperBoundVisitor extends BaseTypeVisitor<UpperBoundAnnotatedTypeFactory> {
+
+    protected final List<ExecutableElement> IndexFirstArgListMethods;
```
smillst: It seems like these methods should just be annotated like `add(@IndexFor("this") int i)`.  The code that actually uses this list looks similar to the fix for kelloggm/checker-framework#14.  Should this special handling be removed once that issue is fixed?  If so, please add a comment to that issue.
`IndexFirstArgListMethods` should be `indexFirstArgListMethods`.  Also, please add a comment.

checker/src/org/checkerframework/checker/upperbound/UpperBoundVisitor.java
```diff
@@ -0,0 +1,131 @@
...
+    @Override
+    public Void visitMethodInvocation(MethodInvocationTree tree, Void type) {
+        if (isFirstArgListMethod(tree)) {
+            ExpressionTree indexTree = tree.getArguments().get(0);
+            ExpressionTree lstTree = tree.getMethodSelect();
+            String lstName = FlowExpressions.internalReprOf(this.atypeFactory, lstTree).toString();
+            AnnotatedTypeMirror indexType = atypeFactory.getAnnotatedType(indexTree);
+            String[] parts = lstTree.toString().split("\\.");
+            // the last part of the methodName is the method (get, remove, etc) so get second to last part
+            String localName = parts[parts.length - 2];
```
smillst: `localName` should not be needed.  `lstName` should be sufficient.  If it's not, then there is a problem with expression annotation support that needs to be fixed.

checker/src/org/checkerframework/checker/upperbound/UpperBoundVisitor.java
```diff
@@ -0,0 +1,131 @@
...
+            // Need to be able to check these as part of the conditional below.
+            // Find max because it's important to determine whether the index is
+            // less than the minimum length of the array. If it could be any
+            // of several values, only the max is of interest.
+            Integer valMax = atypeFactory.valMaxFromExpressionTree(indexTree);
```
smillst: This code appears to be copied from the array index check above.  Please put the duplicated code in a method and call it in both places.

checker/src/org/checkerframework/checker/lowerbound/LowerBoundAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,795 @@
...
+
+        /**
+         * If the argument valueType indicates that the Constant Value Checker knows the exact value
+         * of the annotated expression, returns that integer. Otherwise returns null. This method
+         * should only be used by clients who need exactly one value - such as the binary operator
+         * rules - and not by those that need to know whether a valueType belongs to an LBC
+         * qualifier. Clients needing a qualifier should use lowerBoundAnmFromValueType instead of
+         * this method.
+         */
+        private Integer maybeValFromValueType(AnnotatedTypeMirror valueType) {
```
kelloggm: I used the "maybe" name in reference to the [Maybe](https://wiki.haskell.org/Maybe) type in haskell. I understand that not being a good name in Java, though, and I'll make the suggested changes.

checker/src/org/checkerframework/checker/minlen/MinLenTransfer.java
```diff
@@ -0,0 +1,584 @@
...
+        this.listToArray1 = TreeUtils.getMethod("java.util.List", "toArray", 1, env);
+        this.arrayAsList = TreeUtils.getMethod("java.util.Arrays", "asList", 1, env);
+    }
+
+    @Override
+    public TransferResult<CFValue, MinLenStore> visitAssignment(
+            AssignmentNode node, TransferInput<CFValue, MinLenStore> in) {
+        TransferResult<CFValue, MinLenStore> result = super.visitAssignment(node, in);
+
+        // When an array is created using another array's length as the dimension, transfer
```
kelloggm: I see why what I did is wrong.

My intention was to transfer the information to the right side of the assignment; i.e. when the programmer writes `x = new int[y.length]` then x would have the same type as y. But the framework should handle that, and the code should make `new int[y.length]` have the same type as y, instead.

Is that right?

kelloggm: Despite that, when I changed the code, a test started failing: ArrayCopy.java in tests/minlen. I would expect this to be handled by the framework, but it's not; do you know why? The change I made was to keep the roughly the same code, but move it to visitArrayCreation instead of visitAssignment, and remove the first two if statements (which checked whether an array was being assigned into). I also made the obvious changes based on what node now is (i.e. node instead of (ArrayCreationNode)(node.getExpression()).

checker/src/org/checkerframework/checker/minlen/MinLenStore.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    // If we remove from a list it reduces the minlen of anything that could be an alias of the list by 1.
+    // if we clear a list anything that could be an alias of this list goes to MinLen(0).
+    @Override
+    public void updateForMethodCall(
+            MethodInvocationNode n, AnnotatedTypeFactory atypeFactory, CFValue val) {
+        Receiver caller = FlowExpressions.internalReprOf(atypeFactory, n.getTarget().getReceiver());
+        String methodName = n.getTarget().getMethod().toString();
+        boolean remove = methodName.startsWith("remove(");
```
kelloggm: I wrote code that did so and as a result we get several test failures in ListSupport.java in tests/minlen. All stem from places where there should have been refinements (because remove or clear was called), but TreeUtils.getMethod doesn't find them. I think this is due to the problem we've talked about before where TreeUtils.getMethod doesn't find all overloaded versions of methods. Before, we just hoped we found the right one and if not then it wasn't a big deal: we just missed a false positive we could have eliminated. Here, though, at issue is soundness, which is why I've left it the way it is: we conservatively assume that any method named "remove" is List.remove, and the same with clear. This catches all the ones we care about. If there were a way to get all methods in a class with a given parameter set, I could use that here instead.

checker/src/org/checkerframework/checker/minlen/MinLenTransfer.java
```diff
@@ -0,0 +1,584 @@
...
+        } else if (TreeUtils.isMethodInvocation(node.getTree(), listToArray, env)
+                || TreeUtils.isMethodInvocation(node.getTree(), listToArray1, env)) {
+            if (node.getTarget().getReceiver().getTree() == null) {
+                return result;
+            }
+            AnnotatedTypeMirror ATM =
+                    atypeFactory.getAnnotatedType(node.getTarget().getReceiver().getTree());
+            AnnotationMirror anno = ATM.getAnnotation(MinLen.class);
+            int value = MinLenAnnotatedTypeFactory.getMinLenValue(anno);
+            AnnotationMirror AM = atypeFactory.createMinLen(value + 1);
```
kelloggm: Because this transfer is for >. If something is greater than a @MinLen(3), it should be @MinLen(4), for instance.

smillst:
@kelloggm when you have addressed my comments, please reassign this pull request to me and I'll take another look.  Thanks!

checker/src/org/checkerframework/checker/minlen/MinLenTransfer.java
smillst: The problem with moving this code to visitArrayCreation is that values of new arrays are not actually stored in the store.  I moved this code to the TreeAnnotator and the minlen test pass.

checker/src/org/checkerframework/checker/minlen/MinLenStore.java
smillst: I'll fix this.

checker/src/org/checkerframework/checker/minlen/MinLenTransfer.java
```diff
@@ -0,0 +1,584 @@
...
+        }
+
+        return result;
+    }
+
+    /**
+     * This struct contains all of the information that the refinement functions need. It's called
+     * by each node function (i.e. greater than node, less than node, etc.) and then the results are
+     * passed to the refinement function in whatever order is appropriate for that node. Its
+     * constructor contains all of its logic. I originally wrote this for LowerBoundTransfer but I'm
```
kelloggm: I spent an uncomfortable amount of time this afternoon trying to make this happen. I've cleaned up the code a bit and made it somewhat better, but it's still not really there---there's an IndexAbstractTransfer, but only the LBC is able to use it because the others need to be using their own stores and I wasn't able to figure out a good way around that. Similar problem with trying to abstract out just RefinementInfo.

checker/src/org/checkerframework/checker/minlen/MinLenAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,299 @@
...
+            }
+            return super.visitIdentifier(tree, type);
+        }
+    }
+
+    public ValueAnnotatedTypeFactory getValueAnnotatedTypeFactory() {
+        return valueAnnotatedTypeFactory;
+    }
+
+    protected static int getMinLenValue(AnnotationMirror annotation) {
```
kelloggm: The use in MinLenStore#updateForMethodCall doesn't fit this pattern. So, I'm going to leave the existing method, and add another method that abstracts the other three uses that do fit the pattern you identified.

checker/src/org/checkerframework/checker/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,770 @@
...
+            return createLTEqLengthOfAnnotation(stringList.toArray(new String[0]));
+        }
+        return super.aliasedAnnotation(a);
+    }
+
+    /**
+     * Queries the MinLen Checker to determine if there is a known minimum length for the array. If
+     * not, returns null.
+     */
+    public Integer minLenFromExpressionTree(ExpressionTree tree) {
```
kelloggm: Since 0 is a valid minlen, making this return 0 when there is no minlen erases that distinction - which its callers make use of. However, -1 isn't a valid minlen, so I'm going to do as you suggest, but with -1 instead of 0.

checker/src/org/checkerframework/checker/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,770 @@
...
+        /**
+         * Determines the least upper bound of a1 and a2. If a1 and a2 are both the same type of
+         * Value annotation, then the LUB is the result of taking the intersection of values from
+         * both a1 and a2.
+         *
+         * @return the least upper bound of a1 and a2
+         */
+        @Override
+        public AnnotationMirror leastUpperBound(AnnotationMirror a1, AnnotationMirror a2) {
+            if (!AnnotationUtils.areSameIgnoringValues(
```
kelloggm: I don't know. I removed it.

checker/src/org/checkerframework/checker/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,770 @@
...
+                List<Object> lhsValues = getElementValueArray(lhs, "value", Object.class, true);
+                List<Object> rhsValues = getElementValueArray(rhs, "value", Object.class, true);
+                return rhsValues.containsAll(lhsValues);
+            } else if (isSubtypeRelaxed(rhs, lhs)) {
+                /* Different types that are subtypes of each other ->
+                 * rhs is a subtype of lhs iff lhs.value contains rhs.value.
+                 */
+                List<Object> lhsValues = getElementValueArray(lhs, "value", Object.class, true);
+                List<Object> rhsValues = getElementValueArray(rhs, "value", Object.class, true);
+                return rhsValues.containsAll(lhsValues);
```
kelloggm: I actually think it's the documentation that's wrong here; I'm going to fix that, but the intuition for why it's wrong is that you should always be able to assign a type with more information into a type with less. For instance, consider this:

```java
void test (@LTLengthOf({"a", "b"}) int iab, int[] a, int[] b) {
    @LTLengthOf("a") int ia = iab;
}
```

That assignment is definitely safe, and should be allowed by the type rules. The current rules do allow it, but switching lhs and rhs in isSubtype would cause it to fail.

checker/src/org/checkerframework/checker/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,770 @@
...
+
+        @Override
+        public Void visitMemberSelect(MemberSelectTree tree, AnnotatedTypeMirror type) {
+            if (tree.getIdentifier().contentEquals("length")
+                    && InternalUtils.typeOf(tree.getExpression()).getKind() == TypeKind.ARRAY) {
+                String arrName =
+                        FlowExpressions.internalReprOf(this.atypeFactory, tree.getExpression())
+                                .toString();
+                type.replaceAnnotation(
+                        qualHierarchy.greatestLowerBound(
```
kelloggm: Because we're combining the new information (that we have an array's length) with any other (possibly more specific) information that we might already have about the array length. For instance, if we already know that `a.length` is LTEL("a", "b"), we don't want to overwrite that with just LTEL("a")

checker/src/org/checkerframework/checker/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,770 @@
...
+
+        /**
+         * This exists specifically for Math.min. We need to special case Math.min because it has
+         * unusual semantics: it can be used to combine annotations for the UBC, so it needs to be
+         * special cased. Do not special case other methods here unless you have a compelling reason
+         * to do so.
+         */
+        @Override
+        public Void visitMethodInvocation(MethodInvocationTree tree, AnnotatedTypeMirror type) {
+            ExecutableElement fcnMin =
```
kelloggm: Nevermind; this functionality exists and I'll use it.

checker/src/org/checkerframework/checker/upperbound/UpperBoundAnnotatedTypeFactory.java
smillst: If `type` has an annotation at this point, it's `@LowerBoundUnknown`, so the glb isn't needed here.  Also, if `type` might have some other annotation, glb might produce bottom and the code would need to guard against that.

kelloggm: Why couldn't type have another annotation besides `@UpperBoundUnknown` (which is what I assume you mean)? Where would that guarantee come from?

checker/src/org/checkerframework/checker/upperbound/UpperBoundStore.java
kelloggm: In this case it's definitely wrong - we know that the type's value is only a single name of an array, so there's no need. I'll remove it from this method.

checker/src/org/checkerframework/checker/upperbound/UpperBoundTransfer.java
kelloggm: It only works in a subset of cases. I'll remove it, though - when I implemented it, I thought it would solve the problem, but we've since determined that it doesn't.

checker/src/org/checkerframework/checker/upperbound/UpperBoundAnnotatedTypeFactory.java
smillst: The annotated type of a field access is computed as follows: 
1. The type of the declaration of the field is computed. 
    A. Annotations explicitly written on the element are applied. (Array.length can't be explicitly annotated, so nothing is applied.)
    B. Annotations on explicitly written on declaration of the type of the field are applied. (The type of `length` is int, so there are no annotations.)
    C. The TypeAnnotator is applied on the type. (The TypeAnnotator doesn't do anything with ints, so no annotations are applied.)
    D. Defaults are applied.  `length` is a field so, it is defaulted to top.
2. The type is viewpointed adapted based on the type of the expression that access the field.  (There is not viewpoint adaption for Arrays, so this doesn't effect the type, so it is still annotated with top.)
3. The TreeAnnotator is applied.  (The annotation on type is top and can only be top.)
4. The TypeAnnotator is applied.
5. Defaults are applied.
6. Annotations from data flow are applied. (If the annotation from dataflow is a subtype of the annotation on type, then the dataflow annotation is applied.  Otherwise, type is unchanged.)

So, unless the default annotation for fields is changed, then the annotation on `type` at this point can only be `@UpperBoundUnknown`.

kelloggm: What if this isn't the first time that this array's length is annotated, though? I think at this point it's likely that I'm misunderstanding something, but I'm thinking of something along these lines:
```java
void test(@LTLengthOf("#2") int x, int[] a, int[] b) {
     if (x == b.length) { 
         // At this point, b.length is @LTEqLengthOf("a", "b") via dataflow refinement
         int y = b.length; // I'm concerned with b.length's type at this point. b.length has 
                                    // already be refined, so if I don't call glb, y will have type 
                                    // @LTEqLengthOf("b") instead of {"a","b"}.


     }
}
```

What am I misunderstanding?

checker/src/org/checkerframework/checker/upperbound/UpperBoundTransfer.java
kelloggm: Reassignment.java (the test case you linked to) does issue an error because of this code. I'm going to set it to skip, though; I think no solution is better than an incomplete one.

checker/src/org/checkerframework/checker/upperbound/UpperBoundAnnotatedTypeFactory.java
smillst: They type of `b.length` is computed following the steps outlined above each time it appears in code.  So the type of `b.length` passed to the TreeAnnotator is always `@UpperBoundUnknown int`.  If this is the second appearance  of `b.length` then the dataflow refinement is not applied until after the TreeAnnotator is applied.

kelloggm: Thanks for explaining. I'll make the change you requested; I'm glad I now understand what's going on. I'll also look through the code and see if I did this anywhere else (I probably did...).

smillst: If you did it in a dataflow class, then it is probably right. (And probably why you did it here.) Though you need to make sure bottom won't be returned.

kelloggm: Yeah. I'll add some checks to ensure that bottom isn't ever returned in a place we don't want it.

By "look through the code" I really meant "look through the type factories" - glb is probably wrong when it's used there unless there's a good reason.

kelloggm: Actually I don't think the UBC's glb can return bottom unless one of the inputs is bottom, so I don't think the checks are needed.

smillst:
Ok sounds good.

checker/src/org/checkerframework/checker/upperbound/UpperBoundAnnotatedTypeFactory.java
smillst: Right, but there is still this case: `glb(@LTL("a"), @LTEqL("b")) = @LTL("a","b")`.  But don't bother looking through the code.  I'll check for this during my next review.

checker/src/org/checkerframework/checker/upperbound/UpperBoundVisitor.java
kelloggm: I agree that this code can be removed once the JDK is completely annotated for the whole index checker.
However, I'm unsure of how this code relates to [kelloggm/checker-framework#14](https://github.com/kelloggm/checker-framework/issues/14). My planned solution to that issue is to add a refinement rule to UpperBoundTransfer that, whenever it encounters a compile time constant, it checks the MinLenStore at that program point and looks to see if the value is less than the minlen of that array; if it is, it would add an LTL of that array. The only reason that's not implemented yet is that I don't know how to get a MinLenStore in UpperBoundTransfer.

checker/src/org/checkerframework/checker/upperbound/UpperBoundVisitor.java
kelloggm: I don't know. This would have to have come from Joe, though; he wrote this code. I think what he needs is a way to get only the version of list.remove that takes an index. I think the solution to this, though, isn't to implement that but rather to completely annotate the JDK, which solves this problem entirely.

checker/src/org/checkerframework/checker/upperbound/UpperBoundVisitor.java
kelloggm: Apparently yes. I tried removing this code and it results in a large number of test failures from the list support tests. I don't feel qualified to figure out what's going on, but if you comment out the two places this line appears in UpperBoundVisitor and run upperbound-tests, and you'll see all the failures (there are apparently 30 cases where arrTree.toString is correct and arrName is not).

checker/src/org/checkerframework/checker/upperbound/UpperBoundVisitor.java
smillst: Oh, I've never communicated my idea for how to solve kelloggm/checker-framework#14.  Which we should discuss before you implement anything.   (Basically, my idea is to allow assignment where the left hand side is `@LTL("a")`, the minlen of `a` is `n`, and the value of right hand side is a less than `n`. )  

For the purpose of this code review, you can ignore this comment.

checker/src/org/checkerframework/checker/upperbound/UpperBoundVisitor.java
smillst: Ok, I deal with this.

checker/src/org/checkerframework/checker/upperbound/UpperBoundVisitor.java
kelloggm: That is something we need to talk about. If I recall correctly, I even scheduled a meeting with you to talk about it ages ago but we discovered the soundness bug [#30](https://github.com/kelloggm/checker-framework/issues/30) and ended up talking about that instead.

checker/src/org/checkerframework/checker/upperbound/UpperBoundVisitor.java
smillst: I'll figure this out.

checker/src/org/checkerframework/checker/upperbound/UpperBoundVisitor.java
kelloggm: I'm wrong, you don't need to. I commented out the wrong line.

checker/src/org/checkerframework/checker/upperbound/UpperBoundVisitor.java
kelloggm: You don't need to; there's not a problem. I commented out the wrong line.

kelloggm: I thought I had commented out localName (mentioned in your comment below) but I had actually commented out lstName. If I comment out both localName and the arrTree.toString(), all tests still pass. So I don't think there's a real problem, and I'm not sure why this code is here. I'll delete it.

smillst: Great. Thanks!

kelloggm:
Reassigned @smillst because I believe I've addressed all concerns. SameLen should also be reviewable now.

checker/src/org/checkerframework/checker/samelen/SameLenAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,268 @@
...
+import org.checkerframework.javacutil.AnnotationUtils;
+
+/**
+ * The SameLen Checker is used to determine whether there are multiple arrays in a program that
+ * share the same length. It is part of the Index Checker, and is used as a subchecker by the Index
+ * Checker's components.
+ */
+public class SameLenAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
+
+    public static AnnotationMirror UNKNOWN;
```
smillst: As discussed previously, don't use static fields.

checker/src/org/checkerframework/checker/samelen/SameLenAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,268 @@
...
+
+    /**
+     * Provides a way to query the Constant Value Checker, which computes the values of expressions
+     * known at compile time (constant prop + folding).
+     */
+    private final ValueAnnotatedTypeFactory valueAnnotatedTypeFactory;
```
smillst: As discussed previously, Don't save a sub AnnotatedTypeFactory in a field.

checker/src/org/checkerframework/checker/samelen/SameLenAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,268 @@
...
+     * @param a1 an annotation with a value field
+     * @param a2 an annotation with a value field
+     * @return the set union of the two value fields
+     */
+    private AnnotationMirror getCombinedSameLen(String[] a1, String[] a2) {
+
+        List<String> a1Names = new ArrayList<>();
+        List<String> a2Names = new ArrayList<>();
+
+        for (String s : a1) {
```
smillst: Use `Collections.addAll(a1Names, a1);` instead

checker/src/org/checkerframework/checker/samelen/SameLenAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,268 @@
...
+        }
+
+        HashSet<String> newValues = new HashSet<String>(a1Names.size() + a2Names.size());
+
+        newValues.addAll(a1Names);
+        newValues.addAll(a2Names);
+        Object[] values = newValues.toArray();
+        String[] names = new String[values.length];
+        for (int i = 0; i < names.length; i++) {
+            names[i] = values[i].toString();
```
smillst: Use the https://docs.oracle.com/javase/7/docs/api/java/util/List.html#toArray(T[]) rather than this loop.

checker/src/org/checkerframework/checker/samelen/SameLenAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,268 @@
...
+     * @param sl1 the current annotation of the first array
+     * @param sl2 the current annotation of the second array
+     * @return a combined SameLen annotation
+     */
+    public AnnotationMirror createCombinedSameLen(
+            String a, String b, AnnotationMirror sl1, AnnotationMirror sl2) {
+        String[] a1 = {a, b};
+        ArrayList<String> slStrings = new ArrayList<>();
+        if (AnnotationUtils.areSameByClass(sl1, SameLen.class)) {
+            for (String s : SameLenUtils.getValue(sl1)) {
```
smillst: Collections.addAll

checker/src/org/checkerframework/checker/samelen/SameLenAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,268 @@
...
+            super(factory);
+        }
+    }
+
+    public AnnotationMirror createSameLen(String val) {
+        String[] vals = {val};
+        return createSameLen(vals);
+    }
+
+    public AnnotationMirror createSameLen(String[] val) {
```
smillst: Use a varargs so that you don't need both createSameLen methods. 

`createSameLen(String... val)`

checker/src/org/checkerframework/checker/samelen/SameLenAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,268 @@
...
+        AnnotationBuilder builder = new AnnotationBuilder(processingEnv, SameLen.class);
+        builder.setValue("value", val);
+        return builder.build();
+    }
+
+    public AnnotationMirror createSameLenBottom() {
+        AnnotationBuilder builder = new AnnotationBuilder(processingEnv, SameLenBottom.class);
```
smillst: Create a BOTTOM field and save this value rather than recomputing over and over.

checker/src/org/checkerframework/checker/samelen/SameLenAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,268 @@
...
+        return builder.build();
+    }
+
+    public AnnotationMirror createSameLenUnknown() {
+        AnnotationBuilder builder = new AnnotationBuilder(processingEnv, SameLenUnknown.class);
```
smillst: Use the UNKNOWN field rather than recomputing.

checker/src/org/checkerframework/checker/samelen/SameLenTransfer.java
```diff
@@ -0,0 +1,134 @@
...
+import org.checkerframework.dataflow.analysis.TransferInput;
+import org.checkerframework.dataflow.analysis.TransferResult;
+import org.checkerframework.dataflow.cfg.node.*;
+import org.checkerframework.framework.flow.CFAnalysis;
+import org.checkerframework.framework.flow.CFStore;
+import org.checkerframework.framework.flow.CFTransfer;
+import org.checkerframework.framework.flow.CFValue;
+
+/**
+ * The transfer function for the SameLen checker. Contains two interesting cases: * If an array is
```
smillst: I think the Javadoc reformatting jumbled this a bit.

checker/src/org/checkerframework/checker/samelen/SameLenTransfer.java
```diff
@@ -0,0 +1,134 @@
...
+ * created using another array's length for its length, then the arrays have the same length. * If
+ * the lengths of two arrays are explicitly checked to be equal, they have the same length.
+ */
+public class SameLenTransfer extends CFTransfer {
+
+    // The ATF (Annotated Type Factory).
+    private SameLenAnnotatedTypeFactory aTypeFactory;
+
+    /** Easy shorthand for SameLenUnknown.class, basically. */
+    public static AnnotationMirror UNKNOWN;
```
smillst: No static fields.

checker/src/org/checkerframework/checker/samelen/SameLenTransfer.java
```diff
@@ -0,0 +1,134 @@
...
+        TransferResult<CFValue, CFStore> result = super.visitAssignment(node, in);
+
+        // Check if an array is being created.
+
+        if (node.getExpression() instanceof ArrayCreationNode) {
+            ArrayCreationNode acNode = (ArrayCreationNode) node.getExpression();
+            if (acNode.getDimensions().size() == 1 && isArrayLengthAccess(acNode.getDimension(0))) {
+                // This array that's being created is the same size as the other one.
+                AnnotationMirror combinedSameLen =
+                        aTypeFactory.createCombinedSameLen(
```
smillst: This is impossible to read.  Please extract some of the arguments to local variables.

checker/src/org/checkerframework/checker/samelen/SameLenTransfer.java
```diff
@@ -0,0 +1,134 @@
...
+        return (node instanceof FieldAccessNode
+                && ((FieldAccessNode) node).getFieldName().equals("length")
+                && ((FieldAccessNode) node).getReceiver().getType().getKind() == TypeKind.ARRAY);
+    }
+    /**
+     * Handles refinement of equality comparisons. Looks for a.length and b.length, then annotates
+     * both a and b to have SameLen of each other in the store.
+     */
+    private void refineEq(Node left, Node right, CFStore store) {
+        if (isArrayLengthAccess(left)) { // FIXME add list support here
```
smillst: Remove comment

checker/src/org/checkerframework/checker/samelen/SameLenTransfer.java
```diff
@@ -0,0 +1,134 @@
...
+            if (isArrayLengthAccess(right)) {
+                Receiver leftRec =
+                        FlowExpressions.internalReprOf(
+                                analysis.getTypeFactory(), ((FieldAccessNode) left).getReceiver());
+                Receiver rightRec =
+                        FlowExpressions.internalReprOf(
+                                analysis.getTypeFactory(), ((FieldAccessNode) right).getReceiver());
+                AnnotationMirror combinedSameLen =
+                        aTypeFactory.createCombinedSameLen(
```
smillst: This is impossible to read. Please extract some of the arguments to local variables.

checker/src/org/checkerframework/checker/samelen/SameLenUtils.java
```diff
@@ -0,0 +1,20 @@
...
+import org.checkerframework.checker.upperbound.qual.*;
+import org.checkerframework.javacutil.AnnotationUtils;
+
+public class SameLenUtils {
+
+    /**
+     * Used to get the list of array names that an annotation applies to. Can return null if the
+     * list would be empty.
+     */
+    public static String[] getValue(AnnotationMirror anno) {
```
smillst: Is there any reason this returns an array rather than a list?  The ATF has a lot of code converting arrays to/from lists because of this return type.

smillst:
The 4 checkers are ready to be merged into master, but I'm going to do this from the command line rather than merging this pull request.  So, I'm going to close this pull request.

PR #1001, opened 2016-12-28
Annotate Optional.orElse with @PolyNull (fixes #1000)

### LINKED ISSUE #1000: Could `Optional.orElse(T)` be annotated with `PolyNull`
Hello,
The method [`Optional.orElse(T)`][1] is currently tagged as follow.

     public @Nullable T orElse(@Nullable T other)

Which means the following code does not type check:

```Java
import java.util.Optional;

class Test {
  String foo() {
    Optional<String> opt = Optional.of("abc");
    return opt.orElse("");
  }
}
```
And instead, give the following error:
```
$ javacheck -version
javac 1.8.0-jsr308-2.1.6
$ javacheck -processor NullnessChecker Test.java
Test.java:6: error: [return.type.incompatible] incompatible types in return.
    return opt.orElse("");
                     ^
1 error
```

Would not it make sense to change the annotation to use [`PolyNull`][2] instead?

[1]: https://github.com/typetools/checker-framework/blob/checker-framework-2.1.6/checker/jdk/nullness/src/java/util/Optional.java#L253
[2]: http://types.cs.washington.edu/checker-framework/current/api/org/checkerframework/checker/nullness/qual/PolyNull.html

mernst:
This makes the annotated JDK more precise.
The changeset also restricts the type argument to `Optional` to be `@NonNull`, and it removes redundant, conflicting nullness annotations from the lock-annotated JDK.

### DIFF
--- a/checker/jdk/nullness/src/java/util/Optional.java
+++ b/checker/jdk/nullness/src/java/util/Optional.java
@@ -49,7 +49,7 @@ import org.checkerframework.checker.nullness.qual.PolyNull;
  *
  * @since 1.8
  */
-public final class Optional<@NonNull T extends @NonNull Object> {
+public final class Optional<T extends @NonNull Object> {
     /**
      * Common instance for {@code empty()}.
      */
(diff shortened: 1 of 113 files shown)


checker/jdk/nullness/src/java/util/Optional.java
```diff
@@ -48,7 +49,7 @@
  *
  * @since 1.8
  */
-public final class Optional<T> {
+public final class Optional<@NonNull T extends @NonNull Object> {
```
wmdietl: The @NonNull on T is not necessary.

PR #1002, opened 2017-01-03
Add more @Nullable annotations to PreparedStatement

corcorb:
Updated annotated `java.sql.PreparedStatement` to allow most `#setXXX` calls to allow a `null` parameter.
`PreparedStatement` had previously been updated to allow `#setString` and `#setObject` to take `null` parameters.

This has been verified as allowed using the MySQL driver as a reference implementation.

Note that this commit does not update the following methods, because I could not find an actual implementation of them; I believe they may also accept `null`, but could not confirm:
- `#setArray`
- `#setRef`
- `#setRowId`
- `#setSQLXML`

### DIFF
--- a/checker/jdk/nullness/src/java/sql/PreparedStatement.java
+++ b/checker/jdk/nullness/src/java/sql/PreparedStatement.java
@@ -223,7 +223,7 @@ public interface PreparedStatement extends Statement {
      * marker in the SQL statement; if a database access error occurs or
      * this method is called on a closed <code>PreparedStatement</code>
      */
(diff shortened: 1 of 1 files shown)


mernst:
I'm accepting this not because of the test (MySQL might behave differently than other drivers), but because of the JBDC documentation [Sending JDBC NULL as an IN parameter](https://docs.oracle.com/javase/6/docs/technotes/guides/jdbc/getstart/preparedstatement.html#1003559).  That suggests that the other routines you mentioned may also validly take a nullable parameter.

Thanks for the patch!

PR #1007, opened 2017-01-06
Fix links to manual in README.txt

msridhar:
The previous links were broken.


mernst:
Manu, thanks for the correction.
You are totally right.
The reason is that we are in the middle of a restructuring of the web page, and the repository is inconsistent during that (hopefully brief?) period.
So I'm going to close this issue for now.
Thanks again.

msridhar:
Ok, no worries :)  Thanks!

PR #1016, opened 2017-01-30
Rename expression annotations to dependent type annotations.

smillst:


mernst:
Looks good.  Please fix the spelling errors where "dependant" is used where "dependent" should be.

PR #1017, opened 2017-01-31
Makes LBC a subchecker of UBC

kelloggm:
To allow the following to be type checked:

```java
int[] newArray = new int[firstArray.length + secondArray.length];
        for (int i = 0; i < firstArray.length; i++) {
            newArray[i] = firstArray[i]; //or newArray[i] = secondArray[i];
        }
```

by checking for a sum of NonNegative integers used in the creation of an array; if one is found, those integers must be LTEL of the new array.


smillst:
I haven't reviewed this because I completely rewrote the method this code modifies when adding offsets.  After I merge the offset changes, then we can look at re-adding this feature.

PR #1018, opened 2017-01-31
Fix SameLen bug.

smillst:
Some of the tests in the new test case were failing.


checker/src/org/checkerframework/checker/index/samelen/SameLenTransfer.java
```diff
@@ -42,32 +46,33 @@ public SameLenTransfer(CFAnalysis analysis) {
...
-                AnnotationMirror arrayLengthAnnotation =
+                AnnotationMirror arrayLengthNodeAnnotation =
                         aTypeFactory
                                 .getAnnotatedType(arrayLengthNodeReceiver.getTree())
                                 .getAnnotationInHierarchy(UNKNOWN);
 
                 AnnotationMirror combinedSameLen =
                         aTypeFactory.createCombinedSameLen(
-                                otherRec.toString(),
                                 targetRec.toString(),
```
mernst: Target is "new T[a.length]".  What is targetRec.toString?  Does it ever have a name?

mernst:
Looks good other than my question about the purpose of one expression.

checker/src/org/checkerframework/checker/index/samelen/SameLenTransfer.java
smillst: Target is the lefthand side of the assignment, so if the assignment is `a = new T[b.length]` then ` targetRec.toString()` is `a`.  I updated the comments to make this clear.

PR #1019, opened 2017-01-31
Call the common assignment check for pre/post increments/decrements.

smillst:


framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1216,8 +1216,14 @@ public Void visitUnary(UnaryTree node, Void p) {
                 || (node.getKind() == Tree.Kind.PREFIX_INCREMENT)
                 || (node.getKind() == Tree.Kind.POSTFIX_DECREMENT)
                 || (node.getKind() == Tree.Kind.POSTFIX_INCREMENT)) {
-            AnnotatedTypeMirror type = atypeFactory.getAnnotatedType(node.getExpression());
-            checkAssignability(type, node.getExpression());
+            AnnotatedTypeMirror varType = atypeFactory.getAnnotatedTypeLhs(node.getExpression());
+            // For postfix increments/decrements, the value type is incorrect due to the workaround
```
mernst: Can you expand on this comment?  I think you mean that it's correct to issue the error message in all cases, but that in half of cases (the postfix ones), the error message will be slightly misleading in that the type in the error message will be incorrect.
Is that interpretation correct?  (Probably not, based on the fact that the test case indicates that a needed error isn't issued at all!)
Other than that, this looks good to me.

PR #1022, opened 2017-02-03
Changed upperbound manual picture to give an example

kelloggm:
in response to Mike's request. The figure in the upperbound section of the manual no longer says "java.lang.String" and instead says "myArray". I also made what "myArray" means explicit in the caption.


mernst:
Thanks!

PR #1023, opened 2017-02-04
Add random special case to UBC

kelloggm:
This handles [this issue](https://github.com/kelloggm/checker-framework/issues/46) in my fork. See the conversation about the issue for why this needed to be solved with a special case in the checker instead of a JDK annotation, as was originally proposed.


mernst:
Can you please add a test case as well?

kelloggm:
The test for this is no longer set to skip.

PR #1024, opened 2017-02-05
Index annotated JDK

mernst:


checker/jdk/index/src/java/io/Bits.java
```diff
@@ -0,0 +1,124 @@
...
+ * using big-endian byte ordering.
+ */
+class Bits {
+
+    /*
+     * Methods for unpacking primitive values from byte arrays starting at
+     * given offsets.
+     */
+
+    static boolean getBoolean(byte[] b, int off) {
```
mernst: This file, `Bits.java`, isn't annotated.  Why is it in the annotated JDK, and why does it have an

```java
```

statement?

kelloggm: I've removed all cases like this. Every file with an import statement now has annotations. I didn't remove any files that had no annotations, though, to avoid breaking the JDK build (and because all those files were examined and no annotations were found to be needed).

mernst: Why isn't an annotation needed on the given code?  On the line where this comment is anchored, I think the formal parameter `off` must have type `@IndexFor("#1")`.  Am I confused about that?

kelloggm: It's not a public method, so it doesn't need an annotation. I only annotated the interface between the JDK and outside code, so a package-private method like this one doesn't need an annotation (even though, if I were annotating the interior of the JDK, it would).

mernst:
This `@MinLen` annotation in `ObjectStreamClass` looks wrong, since the length is zero:
```
    public static final ObjectStreamField @MinLen(1) [] NO_FIELDS =
        new ObjectStreamField[0];
```

In DualPivotQuicksort, why is right annotated but left is not?
```
    /**
     * Sorts the specified range of the array by Dual-Pivot Quicksort.
     *
     * @param a the array to be sorted
     * @param left the index of the first element, inclusive, to be sorted
     * @param right the index of the last element, inclusive, to be sorted
     * @param leftmost indicates if this part is the leftmost in the range
     */
    private static void sort(int[] a, int left, @IndexFor("#1") int right, boolean leftmost) {
```

I have a similar question in Arrays.java, for many overloads of copyOfRange, including this one:
```
    public static <T> T[] copyOfRange(T[] original, @IndexFor("#1") int from, int to) {
```

This looks like an opportunity for a special case in the Index Checker, since the result has no more elements than limit (if limit is non-zero):
```
    public String[] split(String regex, int limit) {
```

In StringBuffer, dstBegin is an index for dst:
```
    public synchronized void getChars(int srcBegin, int srcEnd, char[] dst,
                                      int dstBegin)
```

Some annotations are inconsistent between classes.
For example, `Collections.lastIndexOf` and `ArrayList.lastIndexOf` are annotated to return `@GTENegativeOne`, but `List.lastIndexOf` is not.  
`StringBuilder`'s `lastIndexOf(string, int)` does not have its `int fromIndex` annotated with an annotation in comments, but `String`'s `lastIndexOf(string, int)` does.
Can you do a consistency check to find more such instances?

kelloggm:
To respond to a few of your comments:

    public static <T> T[] copyOfRange(T[] original, @IndexFor("#1")
int from, int to) {
>
> CopyOfRange has unusual behavior; the documentation specifically calls out
that "to" does not need to be in the array. It allocates additional empty
space in the new array if to is outside the range. Therefore, it doesn't
need to be annotated @IndexFor. If it is negative (or, rather, if "to -
from" is negative), then the method always throws an
IllegalArgumentException. We could annotate "to" to require it to be
@NonNegative, but that would prevent only that IllegalArgumentException
(and only in some cases).

This looks like an opportunity for a special case in the Index Checker,
> since the result has no more elements than limit (if limit is non-zero):
>
>     public String[] split(String regex, int limit) {
>
> Should I create an issue for this? I agree that it could be an Index
Checker special case, but I'm not sure it's needed.

 Can you do a consistency check to find more such instances?


I'm not sure how I could do this in general. For any given method, I can
check that I'm being consistent throughout using grep (and I'll do that for
some of the more common methods now that it's clear there is a problem),
but I'm curious if there is a better, more general way.

Everything else you pointed out is a mistake, I think. I'll correct them
and update the pull request.

mernst:
You are right about `copyOfRange`; I didn't read the documentation closely enough.

Let's leave out the special case for `split`, for now.  We can add it when needed.

I'll write a program to look for inconsistencies, after you are done and the pull request is assigned to me again.

checker/jdk/index/src/java/io/Bits.java
mernst: This is acceptable, though it does make automated consistency checking much more annoying.

mernst:
Here are some more missing annotations and inconsistencies.

GregorianCalendar has a number of values (such as int month) that could be annotated as `@Positive`.
The same goes for Date.

public ArrayDeque(int numElements) {

Vector's addAll is annotated but those of other collections (ArrayList, etc.) are not.
Same for other methods, such as get().  (BitSet.get is annotated, though.)
and listIterator, remove(), set(), subList(), etc.

Character.charCount, Character.digit, forDigit

StringBuilder.delete, deleteCharAt, insert, replace, and other methods in the same class.
(StringBuffer has annotations for at least some of these, maybe all.)

In LineNumber*,  setLineNumber and getLineNumber can be annotated with `@NonNegative`.

Variables named `int radix` throughout.

Integer.highestOneBit, lowestOneBit

mark(int readlimit)  in:
  BufferedInputStream
  InflaterInputStream
  DeflaterInputStream
  CharArrayReader
  PushbackReader
  Reader
  ByteArrayInputStream

Collections.nCopies

In function offsetByCodePoints, int codePointOffset is annotated in StringBuffer but not in Character or String.
There are other occurrences of offsetByCodePoints, too.

imer.purge

FileChannel.read (two overloaded versions)

FileInputStream.read() is annotated but RandomAccessFile.read is not.  Please check this method throughout the codebase.  (For example, StringBufferInputStream is one other example.)

readUnsignedByte, readUnsignedShort (multiple occurrences)

ZipFile.size

ObjectInputStream.skipBytes

String.split

Collections.swap (or are you saving that for when you annotate lists?)

Object.wait

StreamTokenizer.whitespaceChars, StreamTokenizer.wordChars

ObjectOutputStream.write(byte[] b, int off, int len)


For future reference, here is how I looked for inconsistencies:
```
(ag "@[A-Z]|public" "~/research/types/checker-framework-fork-kelloggm-branch-index-jdk/checker/jdk/index/src/")
(progn
  (read-only-mode 0)
  (set-tab-width 80)
  ;; delete comments
  (delete-matching-lines ": *\\(\\*\\|//\\)" nil nil t)
  (delete-matching-lines "@\\(Override\\|Deprecated\\|SuppressWarnings\\|SafeVarArgs\\)" nil nil t)
  (delete-non-matching-lines "[,()]" nil nil t)
  (save-excursion
    (replace-string "/*@ " "/*@"))
  (save-excursion
    (while (re-search-forward "^\\(.*?\\)	" nil t)
      (replace-match (concat "\\1" (make-string (- 30 (length (match-string 1))) ? )))))
  (query-replace-regexp "\\(.*: +\\(.*? \\([A-Za-z_0-9]+\\)(.*\\)\\)" "\\3	\\2	\\1")
)
(sort-lines)
```
and then visually examine.

kelloggm:
I have addressed your comments. What follows is specific reasoning about my decisions about each comment:

I don’t think it’s appropriate to annotate variables that have nothing to do with arrays, even if they might need to be @NonNegative. So, I did not annotate the month, year, etc. variables in Date or GregorianCalendar.

ArrayDeque was missing an annotation.

addAll was wrong in the non-Vector versions. Fixed.

Character.forDigit: again, I don’t think annotating non-array related variables is appropriate. This method has specific error behavior that we shouldn’t presume to override.

Character#charCount and Character#digit were missing annotations.

I’m not sure why I missed radix. Fixed all instances.

StringBuilder was missing annotations. They’re there now.

LineNumberReader and LineNumberInputStream were both missing annotations on getLineNumber and setLineNumber. Fixed.

No annotations on Integer.highestBit and Integer.lowestBit are correct. Both can return Integer.MIN

I think the mix of annotations and not-annotations for mark(int readlimit) in various classes is right. When it’s supported, it’s required to be non-negative, but if it isn’t supported, there’s no reason to require that the value be non-negative; it will be thrown away.

Collections.nCopies was a missing annotation.

offsetByCodePoints was inconsistent and has been fixed. There are not other instances.

Timer.purge was missing an annotation.

There were several inconsistencies with read. I think I’ve resolved them all.

readUnsigned* was missing annotations.

ZipFile.size() was missing an annotation.

ObjectInputStream.skipBytes is annotated correctly. An non-externally visible method is the one without an annotation.

Wrt String.split: I thought we had decided not to touch this, and special case it if we found it was problematic.

Collections.swap isn’t going to be touched until we have list support.

Object.wait was missing an annotation.

The StreamTokenizer methods you mentioned accept all arguments, and modify them if they’re too high or too low. So they don’t need annotations.

ObjectOutputStream.write(byte[] b, int off, int len) is annotated correctly (with appropriate @IF/@IOH annotations). There’s another similar looking method in the class with no annotations that’s inside a private inner class.

I also wasn't able to get your script for looking for inconsistencies working. I translated some of it to bash, though, and was able to look through for additional inconsistencies. I fixed one or two but it mostly looks good to me at this point.

mernst:
Thanks for the improvements.  A couple reactions:

> I don’t think it’s appropriate to annotate variables that have nothing to do with
> arrays, even if they might need to be `@NonNegative`. So, I did not annotate the
> month, year, etc. variables in Date or GregorianCalendar.

OK, for now.  Eventually, this will need to be done, and it may be harder to find all the cases then.  That's why I was thinking of doing it now.

> I think the mix of annotations and not-annotations for mark(int readlimit) in
> various classes is right. When it’s supported, it’s required to be non-negative,
> but if it isn’t supported, there’s no reason to require that the value be
> non-negative; it will be thrown away.

I feel that all the parameters for `mark(int readAheadLimit)` should be annotated as `@NonNegative` for two reasons.
 * It's probably a mistake if a user passes a nonsensical value, and the user would appreciate knowing that.
 * It's required for behavioral subtyping.  A subtype can only strengthen, not weaken, the specification.  Currently, the Reader specification is:

        void mark(int readAheadLimit)

   and the FilterReader (which extends Reader) specification is:

        void mark(@NonNegative int readAheadLimit)

   This says that Reader.mark can take any argument.  A subtype implementation must be able to take any argument that a supertype implementation can, but that is not the case with the current annotations.

> Wrt String.split: I thought we had decided not to touch this, and special case it if we found it was problematic.

The returned array has length at least 1.  This is a `@MinLen(1)` annotation.  A special case could be more precise, but even this gives some information.

> The StreamTokenizer methods you mentioned accept all arguments, and modify them if they’re too high or too low. So they don’t need annotations.

I agree with this implementation detail.  However, such a call is undesirable and most likely indicates a client error.  I think it's best to issue a warning, which will help the user improve the code.  In the unlikely case that the user meant to pass out-of-range values, the user can suppress this warning.  We have made similar annotations elsewhere in the JDK.

kelloggm:
I have changed all instances of mark(int readAheadLimit) so that the parameter needs to be @NonNegative. Your argument makes sense; the second part (about behavioral sub typing) makes the change absolutely imperative.

String.split did require a MinLen(1) annotation; I thought you were still talking about the special case and overlooked the more general case.

I also made the changes to StreamTokenizer (annotating every int hi and int low as @NonNegative). I agree that it’s likely undesirable to be passing out-of-range arguments, but I wanted to err on the side of not warning users about potentially correct behavior when it was technically safe. However, I defer to your judgement in this case, especially because you’re right that we have similar annotations elsewhere in the JDK.

mernst:
This is ready to merge.
I'm worried that I might become the author of the squashed commit if I do it.  I don't know that would happen, but if so, that wouldn't give you credit for your work.)  Do you want me to do that, or do you want to close this, create a new pull request from the same branch, and then squash and merge that?

kelloggm:
I can do that if you think it's worthwhile. Personally, I'm not particularly concerned about attribution for this particular piece of work; we both know who did what, and I'm not sure who else would be paying attention.

smillst:
Closed in favor of #1034, which was merged in https://github.com/typetools/checker-framework/commit/17e90d755f47af08f380477f88bdd38bce17a8d8.

PR #1025, opened 2017-02-06
Samelen fix

kelloggm:
Fixes https://github.com/kelloggm/checker-framework/issues/87.

The issue was some blatantly wrong code in UpperBoundAnnotatedTypeFactory, and a missing method is SameLenTypeFactory. 

Also moves the tests Mike created earlier to the correct test folder, and comments out the one that fails currently (it's for https://github.com/kelloggm/checker-framework/issues/88).


checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
```diff
@@ -214,4 +214,8 @@ public AnnotationMirror createSameLen(String... val) {
         builder.setValue("value", val);
         return builder.build();
     }
+
+    public AnnotationMirror createSameLenUnknown() {
```
smillst: This method isn't creating anything, so the method name is confusing.  Either change the name or delete the method and make the field public.

kelloggm: I gave it this name to be consistent with the pattern of "createX" methods, like createSameLen right above it. From the view of the user, they're receiving a fresh @SameLenUnknown annotation (but there's only one, so there's no need to create another, which I view as an implementation detail).

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -20,15 +20,7 @@
...
-import org.checkerframework.checker.index.qual.IndexFor;
-import org.checkerframework.checker.index.qual.IndexOrHigh;
-import org.checkerframework.checker.index.qual.IndexOrLow;
-import org.checkerframework.checker.index.qual.LTEqLengthOf;
-import org.checkerframework.checker.index.qual.LTLengthOf;
-import org.checkerframework.checker.index.qual.LTOMLengthOf;
-import org.checkerframework.checker.index.qual.MinLen;
-import org.checkerframework.checker.index.qual.UpperBoundBottom;
-import org.checkerframework.checker.index.qual.UpperBoundUnknown;
+import org.checkerframework.checker.index.qual.*;
```
mernst: The Checker Framework standard (love it or hate it) is to not use wildcard imports.  Please revert the change and add any more needed imports.

checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
smillst: The `createSameLen(String...)` method creates a new object so this name makes sense.  This method does not create anything.   

> From the view of the user, they're receiving a fresh @SameLenUnknown annotation (but there's only one, so there's no need to create another, which I view as an implementation detail).

The method name leads the user to believe a fresh object is created each time, but it does not create a fresh object.  If you had named the methods getX, then it might be an implementation detail whether or not a new object is created each time.

We normally use a public final field for the top and bottom annotations in a type system. So, please just make the field public and final.

kelloggm: I made this change. Thanks for explaining the rationale.

PR #1026, opened 2017-02-06
Binary case of averaging IndexFor annotations

kelloggm:
Fixes two concrete problems that caused the following code to not be validated by the index checker:

```java

public class IndexForBug {
	
	public static void bug(int[] a, @IndexFor("#1") int i, @IndexFor("#1") int j) {
		@IndexFor("a") int k = (i + j)/2; //upperbound unknown
	}
} 
```

First, the LBC was missing a case for dividing by positive constants larger than one when the other type is @NonNegative. That case is added here.

Second, a special case was added to the UBC so that it knows that adding two LTL ints and then dividing them by two results in another LTL int. This only handles the case when two LTLs are added.


checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -2,12 +2,7 @@
 
 import static org.checkerframework.javacutil.AnnotationUtils.getElementValueArray;
 
-import com.sun.source.tree.BinaryTree;
-import com.sun.source.tree.ExpressionTree;
-import com.sun.source.tree.MemberSelectTree;
-import com.sun.source.tree.MethodInvocationTree;
-import com.sun.source.tree.Tree;
-import com.sun.source.tree.UnaryTree;
+import com.sun.source.tree.*;
```
mernst: For consistency with the rest of the Checker Framework, please don't use wildcard imports.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -354,6 +349,16 @@ public QualifierHierarchy createQualifierHierarchy(MultiGraphFactory factory) {
     }
 
     /**
+     * Check if type is either LTL or LTOM. Prevents duplication and makes code more readable.
+     *
+     * @param type
```
mernst: Fill in, or remove, empty Javadoc tags.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -354,6 +349,16 @@ public QualifierHierarchy createQualifierHierarchy(MultiGraphFactory factory) {
     }
 
     /**
+     * Check if type is either LTL or LTOM. Prevents duplication and makes code more readable.
```
mernst: Rather than "check", please use more specific language, such as "Return true if".

I suggest that you delete the second sentence.  A reader will assume that you didn't add this to make the code less readable...

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -648,24 +653,56 @@ private void addAnnotationForDivide(
...
         }
 
         private void addAnnotationForLiteralDivide(
-                int val, AnnotatedTypeMirror nonLiteralType, AnnotatedTypeMirror type) {
-            if (nonLiteralType.hasAnnotation(LTLengthOf.class)
-                    || nonLiteralType.hasAnnotation(LTOMLengthOf.class)) {
+                int val,
+                ExpressionTree leftTree,
+                AnnotatedTypeMirror leftType,
+                AnnotatedTypeMirror type) {
```
mernst: What is `type`?  Please write a Javadoc comment for every method, including private ones.  (Do this throughout the Index Checker.)  That will help readers to understand the code, and IDEs can show private as well as public Javadoc.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -648,24 +653,56 @@ private void addAnnotationForDivide(
...
                 } else if (val == 1) {
-                    type.addAnnotation(nonLiteralType.getAnnotationInHierarchy(UNKNOWN));
+                    type.addAnnotation(leftType.getAnnotationInHierarchy(UNKNOWN));
+                }
+            } else if (val == 2) {
+                // The average of two LTL/LTOMs is LTL.
+
+                // Necessary because otherwise an expression in paretheses doesn't make it past
+                // the next check...
+                if (leftTree.getKind() == Tree.Kind.PARENTHESIZED) {
```
mernst: I would make this a while loop, just in case.  It's no more complicated but handles the case of an extra set of parentheses.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -685,9 +722,7 @@ private void addAnnotationForLiteralPlus(
                 return;
             }
             if (val < -1) {
-                if (nonLiteralType.hasAnnotation(LTLengthOf.class)
-                        || nonLiteralType.hasAnnotation(LTOMLengthOf.class)
-                        || nonLiteralType.hasAnnotation(LTEqLengthOf.class)) {
+                if (isLTL(nonLiteralType) || nonLiteralType.hasAnnotation(LTEqLengthOf.class)) {
```
mernst: I found the use of isLTL helpful above (thanks!), but here I don't think it buys as much, and is surprising because LTEqLength looks like something isLTL would handle.  Another way to think of this is:  the previous code could be read on its own.  This code requires the reader to look elsewhere in order to understand it, and at the savings of just one clause (from 3 to 2 clauses in the or statement).

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -648,24 +653,56 @@ private void addAnnotationForDivide(
...
                 } else if (val == 1) {
-                    type.addAnnotation(nonLiteralType.getAnnotationInHierarchy(UNKNOWN));
+                    type.addAnnotation(leftType.getAnnotationInHierarchy(UNKNOWN));
+                }
+            } else if (val == 2) {
+                // The average of two LTL/LTOMs is LTL.
+
+                // Necessary because otherwise an expression in parentheses doesn't make it past
+                // the next check...
+                if (leftTree.getKind() == Tree.Kind.PARENTHESIZED) {
```
mernst: I would make this a while loop, just in case. It's no more complicated but handles the case of an extra set of parentheses.  It's what is (or should be) done elsewhere.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -662,24 +672,64 @@ private void addAnnotationForDivide(
...
                 AnnotatedTypeMirror leftType = getAnnotatedType(left);
-                addAnnotationForLiteralDivide(maybeValRight, leftType, type);
+                addAnnotationForLiteralDivide(maybeValRight, left, leftType, type);
             }
         }
 
+        /**
+         * Handles division when the right side is a compile-time constant.
+         *
+         * @param val the integer value of the right side.
```
mernst: Nit:  Javadoc style says that sentence fragments shouldn't end with a period unless they are followed by a sentence.  So the periods in the Javadoc for `type` are good, but not for the other three formal parameters.

checker/src/org/checkerframework/checker/index/lowerbound/LowerBoundAnnotatedTypeFactory.java
```diff
@@ -615,7 +615,7 @@ private void addAnnotationForLiteralDivideLeft(
         /**
          * When the value on the right is known at compile time. If the value is zero, then this is
```
mernst: I would move all but the last sentence to line 624 where the code is; that prevents some cross-referencing.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -662,24 +672,64 @@ private void addAnnotationForDivide(
...
+                    String[] names = getValue(leftType.getAnnotationInHierarchy(UNKNOWN));
                     type.replaceAnnotation(createLTLengthOfAnnotation(names));
                 } else if (val == 1) {
-                    type.addAnnotation(nonLiteralType.getAnnotationInHierarchy(UNKNOWN));
+                    type.addAnnotation(leftType.getAnnotationInHierarchy(UNKNOWN));
+                }
+            } else if (val == 2) {
+                // The average of two LTL/LTOMs is LTL.
+
+                // Necessary because otherwise an expression in parentheses doesn't make it past
```
mernst: `TreeUtils.skipParens` does this.  I suggest you use it instead.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -699,8 +749,8 @@ private void addAnnotationForLiteralPlus(
                 return;
             }
             if (val < -1) {
-                if (nonLiteralType.hasAnnotation(LTLengthOf.class)
```
mernst: To keep the diff focused, I suggest undoing this change and saving it for a separate pull request if it is needed.

mernst:
Looks great.  Feel free to merge after you address the current comments and the tests pass.

PR #1030, opened 2017-02-08
Ensure that LBC is actually taking advantage of value checker knowledge

kelloggm:
The LBC was not taking advantage of all value checker knowledge. Especially in this case:
```java
class ArrayIntro {
    void test() {
        int a = 9;
        a += 5;
        a -= 2;
        int [] arr5 = new int[a]; // a has type GTEN1, not type POS like it should.
    }
}
```
In that code, the LBC wasn't learning the value of a correctly, because of some mishandling somewhere of compound assignments. However, to make the solution more general, this PR queries the value checker each time an identifier is encountered, and annotates the identifier if the value checker is able to produce a more precise type than the LBC on its own. Also, changes several old tests that were incorrect once the new rule was implemented.


mernst:
This looks reasonable to me.

I am a bit troubled by widescale removal of tests.  I didn't look at them closely, but would it have been possible to update the expected errors rather than just deleting so much test code?

checker/src/org/checkerframework/checker/index/lowerbound/LowerBoundAnnotatedTypeFactory.java
```diff
@@ -300,6 +301,25 @@ public Void visitMemberSelect(MemberSelectTree tree, AnnotatedTypeMirror type) {
         }
 
         /**
+         * Compound assignments aren't desugared correctly in some way, which prevents x from
+         * being @Positive after, e.g., this code is executed: "x = 5; x -= 2;". This code is more
+         * general and ensures that any identifier with a compile-time known value is represented
+         * correctly.
+         */
+        @Override
+        public Void visitIdentifier(IdentifierTree tree, AnnotatedTypeMirror type) {
```
smillst: This won't work if the type of the identifier is computed using an element instead of a tree. 

Instead do something like this:
https://github.com/typetools/checker-framework/blob/master/checker/src/org/checkerframework/checker/index/minlen/MinLenAnnotatedTypeFactory.java#L239-L259

smillst: Also, the LBC should allow the following code.
````
@IntVal({1, 2}) int getOneOrTwo() {...}
void test() {
  int[] a = new int[getOneOrTwo()];
}
````
Which should work once my above suggestion is implemented.

**Comment by kelloggm** [comment]
@mernst The tests I removed are definitely redundant; I kept one example of each "kind" of test from these test files. I wrote these tests at the very beginning of the project when I didn't really understand how the CF works, so much of the "testing" was actually not doing anything. I just removed the fluff.

File: checker/src/org/checkerframework/checker/index/lowerbound/LowerBoundAnnotatedTypeFactory.java (hunk 1 above)
Comment (smillst) [inline checker/src/org/checkerframework/checker/index/lowerbound/LowerBoundAnnotatedTypeFactory.java]: Also, I don't see the ArrayIntro.java test case in the diffs.  Did you forget to add it or unskip it?

File: checker/src/org/checkerframework/checker/index/lowerbound/LowerBoundAnnotatedTypeFactory.java [hunk 2]
```diff
@@ -95,6 +95,42 @@ public LowerBoundAnnotatedTypeFactory(BaseTypeChecker checker) {
...
+     * type. If the new lower bound type is more specific than type, convert type to that type.
+     *
+     * @param valueType the value checker's type
+     * @param type the current lower bound type of the expression being evaluated
+     */
+    private void addLowerBoundTypeFromValueType(
+            AnnotatedTypeMirror valueType, AnnotatedTypeMirror type) {
+        AnnotationMirror anm = getLowerBoundAnnotationFromValueType(valueType);
+        if (!type.isAnnotatedInHierarchy(UNKNOWN)) {
+            type.addAnnotation(anm);
```
smillst: Type variables may or may not have a "primary" annotation -- that is AnnotatedTypeVariable#getAnnotations() may be empty, even if the type is well formed and completely annotated.  In this case, this branch is taken.  If 'anm' is `@LowerBoundUnknown`, it is added as a primary annotation, which sets the annotation on the upper and lower bound to `@LowerBoundUnknown`, too. (This cause the error you reported.)

So, to fix the error, don't add 'anm' if it is `@LowerBoundUnknown`.

kelloggm: Thanks. I don't really understand what you mean by the "primary" annotation, but it's likely an implementation detail of the framework I just haven't encountered before. The solution is quite concrete.

PR #1031, opened 2017-02-09
Add array offsets to the Upper Bound Checker

smillst:


checker/tests/upperbound/OffsetExample.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public static boolean isSubarray(Object[] a, Object[] sub, int a_offset) {
+        int a_len = a.length - a_offset;
+        int sub_len = sub.length;
+        if (a_len < sub_len) {
+            return false;
+        }
+        for (int i = 0; i < sub_len; i++) {
+            if (sub[i] == a[a_offset + i]) { // UBC warns here
```
kelloggm: Presumably that comment is no longer accurate. Ditto for the one on line 84.

checker/src/org/checkerframework/checker/index/upperbound/OffsetEquation.java
```diff
@@ -0,0 +1,469 @@
...
+            parseRecurMinus(node, factory, eq);
+        } else {
+            parseRecurPlus(node, factory, eq);
+        }
+        return eq;
+    }
+
+    private static void parseRecurPlus(
+            Node node, UpperBoundAnnotatedTypeFactory factory, OffsetEquation eq) {
+        if (node.getTree() != null && TreeUtils.isExpressionTree(node.getTree())) {
```
smillst: I don't understand this comment.  The smallest value of all possible values is required, so I guess the code could call the ValueATF directly, but I don't see the value in that.

The ValueLiteralNode code is duplicated by the Value Checker, so it could be removed.

kelloggm: I wasn't quite sure how much was duplicated in the ValueChecker. Looking at it again, I agree the comment isn't super clear. I think removing the ValueLiteralNode code is a good move, and I think the rest is reasonable.

checker/src/org/checkerframework/checker/index/upperbound/OffsetEquation.java
```diff
@@ -0,0 +1,469 @@
...
+    /**
+     * Adds or subtracts the other equation to a copy of this one.
+     *
+     * <p>If subtraction is specified, then every term in other is subtracted.
+     *
+     * @param op '-' for subtraction or '+' for addition
+     * @param other equation to add or subtract
+     * @return a copy of this equation +/- other
+     */
+    public OffsetEquation copyAdd(char op, OffsetEquation other) {
```
smillst: This is addressed below.

checker/src/org/checkerframework/checker/index/upperbound/OffsetEquation.java
```diff
@@ -0,0 +1,469 @@
...
+     * <p>Otherwise the return equation is created by converting the node to a {@link
+     * FlowExpressions.Receiver} and then then added as a term to the returned equation. If op is
+     * '-' then it is a subtracted term.
+     *
+     * @param node Node from which to create offset equation
+     * @param factory AnnotationTypeFactory
+     * @param op '+' or '-'
+     * @return an offset equation from the Node
+     */
+    public static OffsetEquation createOffsetFromNode(
```
smillst: This is addressed below.

kelloggm:
Suzanne, why did you add the two additional checks to checker/tests/index/IndexForAverage.java? I would not expect that test to pass - the LBC cannot determine that ((i - 1) + j)/2 is NonNegative when i and j are NonNegative right now. It could be amended to determine that fact (as it probably should), but I don't think the test belongs in this commit.

smillst:
>Suzanne, why did you add the two additional checks to checker/tests/index/IndexForAverage.java? I would not expect that test to pass - the LBC cannot determine that ((i - 1) + j)/2 is NonNegative when i and j are NonNegative right now. It could be amended to determine that fact (as it probably should), but I don't think the test belongs in this commit.

You're talking about this testcase:
````
public static void bug2(int[] a, @IndexFor("#1") int i, @IndexFor("#1") int j) {
        @IndexFor("a") int k = ((i - 1) + j) / 2;
        @IndexFor("a") int h = ((i + 1) + j) / 2;
    }
````

Which will fail for the LBC.  When I wrote this test, it was run with just the UBC.  I'll re-write it so that the @IndexFor annotations are @LTLengthOf annotations.

kelloggm:
Thanks. Sorry I updated the testing infrastructure under you.

PR #1033, opened 2017-02-09
Index test reorganization

kelloggm:
Compresses the test directories for the index checker's components into a single, unified test directory instead of four separate pieces. This allows for broader testing (i.e. all checkers run on all tests) and creates less confusion and duplication when creating/maintaining tests. The pull request eliminates the lowerbound-tests, upperbound-tests, minlen-tests, and samelen-tests ant targets and replaces them with a single index-tests target. This target already existed in the main repository, but was unused.

The pull request also appears to remove some tests; they aren't actually being removed, but those version are being removed because an identical duplicate existed in one of the other index test suites.


mernst:
As far as I am concerned, this can be squashed and merged as soon as the tests pass.

PR #1038, opened 2017-02-13
Test case and fix for Issue 1027. Add Google Inc. as contributor.

wmdietlGC:


docs/manual/troubleshooting.tex
```diff
@@ -1124,6 +1124,7 @@ \section{Credits and changelog\label{credits}}
 David Lazar,
 David McArthur,
 Eric Spishak,
+Google Inc. (via @wmdietlGC),
```
mernst: Nit: please put `\` after `.` to signal to LaTeX that this period doesn't end a sentence.

framework/src/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -263,24 +265,29 @@ public void standardizeVariable(Tree node, AnnotatedTypeMirror type, Element ele
...
+                if (enclTree.getKind() == Kind.METHOD) {
+                    MethodTree methodTree = (MethodTree) enclTree;
                     TypeMirror enclosingType = ElementUtils.enclosingClass(ele).asType();
                     FlowExpressionContext parameterContext =
                             FlowExpressionContext.buildContextForMethodDeclaration(
                                     methodTree, enclosingType, factory.getContext());
                     standardizeDoNotUseLocals(parameterContext, path, type);
-                    break;
+                } else {
+                    // If there is no enclosing method, then the parameter is a parameter to a lambda
```
mernst: It's not that there is no enclosing method, but that the nearest enclosing thing was a lambda rather than a method.  Am I right about that?
(I realize that you copied the comment from the existing code, but I think it can still be improved.)

mernst:
Looks great, thanks a lot!
I have just two minor nits, then this can be merged.

framework/src/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
wmdietlGC: Yes, there might still be a method around the lambda, but the most enclosing tree was a lambda.
I've added short comments to both branches.
I think enclTree will always be non-null; I can't think of a way to get a parameter without either of these two cases being true. We'll see if we ever get a NPE on the if condition.

PR #1048, opened 2017-02-14
Uncomments a test that passes in the Index Checker.

kelloggm:


checker/tests/index/LiteralArray.java
```diff
@@ -2,7 +2,7 @@
 // test case for issue #67:
 // https://github.com/kelloggm/checker-framework/issues/67
 
-// @skip-test until issue is fixed
+import org.checkerframework.checker.index.qual.*;
 
 public class LiteralArray {
```
wmdietl: Test cases shouldn't use annotations in comments.

PR #1056, opened 2017-02-17
Makes samelen symmetric in most cases

kelloggm:
The fundamental problem is that when a programmer writes annotations, they aren't automatically symmetric. Here's a smaller test case that showcases the problem that this PR fixes:

```java
    void test(int @SameLen("#2") [] a, int[] b) {
	int[] c = new int[a.length];
	for(int i = 0; i < c.length; i++) { // i's type is @LTL("c")
	    b[i] = 1; // b's type is @SameLenUnknown, so fails
	    int @SameLen({"a", "b", "c"}) [] d = c; // succeeds
	}
    }
```

SameLen types are symmetric in the sense that all the ones that are inferred are applied to both of the arrays involved. So for instance in the example above, `int [] c = new int[a.length]` results in both a and c being updated. But it doesn't update the other SameLen types that are associated with a and c - which is to say, while both a and c become `@SameLen({"a","b","c"})`, the type of b never changes. This PR changes that so that all arrays listed in a SameLen type are updated whenever one is.

### DIFF
--- a/checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
+++ b/checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
@@ -94,14 +94,20 @@ public class SameLenAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
      * For the use of the transfer function; generates a SameLen that includes a and b, as well as
      * everything in sl1 and sl2, if they are SameLen annotations.
      *
-     * @param a the name of the first array
-     * @param b the name of the second array
+     * @param aRec receiver representing the first array
+     * @param bRec receiver representing the second array
      * @param sl1 the current annotation of the first array
      * @param sl2 the current annotation of the second array
      * @return a combined SameLen annotation
      */
     public AnnotationMirror createCombinedSameLen(
-            String a, String b, AnnotationMirror sl1, AnnotationMirror sl2) {
+            FlowExpressions.Receiver aRec,
+            FlowExpressions.Receiver bRec,
+            AnnotationMirror sl1,
+            AnnotationMirror sl2) {
+
+        String a = aRec.toString();
+        String b = bRec.toString();
 
         List<String> aValues = new ArrayList<String>();
         aValues.add(a);
--- a/checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
+++ b/checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
@@ -1,6 +1,5 @@
 package org.checkerframework.checker.index.samelen;
 
-import com.sun.source.util.TreePath;
 import java.lang.annotation.Annotation;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -9,7 +8,6 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 import javax.lang.model.element.AnnotationMirror;
-import javax.lang.model.type.TypeMirror;
 import org.checkerframework.checker.index.qual.SameLen;
 import org.checkerframework.checker.index.qual.SameLenBottom;
 import org.checkerframework.checker.index.qual.SameLenUnknown;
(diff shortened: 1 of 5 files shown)


checker/src/org/checkerframework/checker/index/samelen/SameLenTransfer.java
```diff
@@ -73,16 +75,46 @@ public SameLenTransfer(CFAnalysis analysis) {
...
+        AnnotationMirror leftAnno =
+                aTypeFactory
+                        .getAnnotatedType(node.getExpression().getTree())
+                        .getAnnotationInHierarchy(UNKNOWN);
+
+        if (leftAnno != null && AnnotationUtils.areSameByClass(leftAnno, SameLen.class)) {
+
+            AnnotationMirror combinedSameLen =
+                    aTypeFactory.createCombinedSameLen(
+                            node.getTarget().toString(),
```
smillst: Instead of the Tree.toString(), the Receiver.toString should be used, because the Tree toString isn't standardized.  For example if the Tree is (this).field, the toString returns (this).field instead of this.field. Same goes for the next line.  Can you change the type of the parameters of createCombinedSameLen to Receiver to catch this problem sooner?

checker/src/org/checkerframework/checker/index/samelen/SameLenTransfer.java
```diff
@@ -73,16 +75,46 @@ public SameLenTransfer(CFAnalysis analysis) {
...
+
+            propagateCombinedSameLen(combinedSameLen, node, result.getRegularStore());
+        }
+
         return result;
     }
 
+    private void propagateCombinedSameLen(
+            AnnotationMirror combinedSameLen, Node node, CFStore store) {
+        // Get the value of the combined samelen type, and then
```
smillst: Please move this comment to the method Javadoc.

checker/src/org/checkerframework/checker/index/samelen/SameLenTransfer.java
```diff
@@ -73,16 +75,46 @@ public SameLenTransfer(CFAnalysis analysis) {
...
+        // insert that combined SameLen into the receiver for each.
```
smillst: I think you mean"insert the combined SameLen into the store for each array mentioned in the SameLen".

checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
```diff
@@ -108,7 +114,30 @@ public AnnotationMirror createCombinedSameLen(
             bValues.addAll(SameLenUtils.getValue(sl2));
         }
 
-        return getCombinedSameLen(aValues, bValues);
+        AnnotationMirror res = getCombinedSameLen(aValues, bValues);
+        return res;
+    }
+
+    public FlowExpressions.Receiver arrayReceiverFromString(
```
smillst: Is there anything particular to arrays in the method or is it generally useful?  I'm thinking we might want to move it to the super class. (Or perhaps into the CFAbstractTransfer class) Thoughts?

kelloggm: I had that thought while writing it. I wasn't sure of the answer and I figured you would comment on it, since I named it as if it needed an array and then didn't use the property. If we moved it into a superclass, that would allow MinLenAnnotatedTypeFactory#getMinLenFromString to call it as well (which is especially good, since the code in that method is all duplicated in getMinLenFromString, as well (or more accurately I stole the code from there...)).

smillst: In the offset branch, which I actually just merged, I moved that code to the GenericATF. So you could move this method there, rename it to some more generic, add documentation, and have getAnnotationFromJavaExpressionString call the new method.  Or can do it after this PR is merged.

kelloggm: I'll do it for this PR. I think it's relevant enough here that we should do it.

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -736,6 +726,35 @@ public AnnotationMirror getAnnotationFromJavaExpressionString(
     }
 
     /**
+     * Produces the receiver associated with expression on currentPath.
+     *
+     * @param expression Java expression
+     * @param currentPath location at which expression is evaluated
+     * @return
```
smillst: Travis is going to complain about no text after `@return`

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -736,6 +726,35 @@ public AnnotationMirror getAnnotationFromJavaExpressionString(
...
+                        r,
+                        FlowExpressions.getParametersOfEnclosingMethod(this, currentPath),
+                        this.getContext());
+
+        FlowExpressions.Receiver receiver;
+        try {
+            receiver = FlowExpressionParseUtil.parse(expression, context, currentPath, true);
+        } catch (FlowExpressionParseUtil.FlowExpressionParseException ex) {
+            // ignore parse errors.
+            return null;
```
smillst: I think that that the exception should be handled by the caller, which is why the signature of `getAnnotationFromJavaExpressionString` says it throws the error.  FlowExpressoinParseExceptions contain a message about why the parsing failed.  Calls of this method might want to surface that information to the user.

PR #1058, opened 2017-02-17
Allow type hierarchies to implement a widen upper bound.

smillst:
This prevents dataflow from an infinite loop if a type hierarchy has an infinite ascending chain.


checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
```diff
@@ -153,6 +153,10 @@ public final boolean isUnknownOrBottom() {
 
     public abstract UBQualifier lub(UBQualifier other);
 
+    public UBQualifier widenUb(UBQualifier obj) {
```
wmdietlGC: It's `UB` in the class name, but `Ub` in the method name. That's inconsistent.

dataflow/src/org/checkerframework/dataflow/analysis/Analysis.java
```diff
@@ -74,6 +74,18 @@
     protected IdentityHashMap<Block, S> elseStores;
 
     /**
+     * Number of times every block has been analyzed since the last time widening was applied. Null,
+     * if maxCountBeforeWidening is -1 which implies widing isn't using for this analysis. .
```
wmdietlGC: `using` -> `used`.
Two dots at end of sentence.

dataflow/src/org/checkerframework/dataflow/analysis/Analysis.java
```diff
@@ -74,6 +74,18 @@
...
+     */
+    protected IdentityHashMap<Block, Integer> blockCount;
+
+    /**
+     * Number of times a block can be analyzed before the widening. -1 implies that widening
```
wmdietlGC: `the widening` -> `widening`.

dataflow/src/org/checkerframework/dataflow/analysis/Analysis.java
```diff
@@ -462,12 +475,24 @@ protected void addStoreBefore(
...
+            count = count == null ? 0 : count;
+            shouldWiden = count >= maxCountBeforeWidening;
+        }
 
         switch (kind) {
             case THEN:
                 {
                     // Update the then store
-                    S newThenStore = (thenStore != null) ? thenStore.leastUpperBound(s) : s;
+                    S newThenStore;
```
wmdietlGC: This logic is duplicated 5 times. Extract in helper method.
I would check for `(thenStore == null)` just once first, as that is common to the two branches.

dataflow/src/org/checkerframework/dataflow/analysis/Store.java
```diff
@@ -54,6 +54,25 @@
     S leastUpperBound(S other);
 
     /**
+     * Compute an upper bound of two stores that is wider than the least upper bound of the two
+     * stores.
```
wmdietlGC: Any advice on how much wider? Is it good enough to just return TOP?
When is the method used?

dataflow/src/org/checkerframework/dataflow/analysis/Store.java
```diff
@@ -54,6 +54,25 @@
...
+     *
+     * <p>(This method is optional, if not required, implementations should call leastUpperBound.)
```
wmdietlGC: It's not optional if I have to implement it. Clarify the description.

dataflow/src/org/checkerframework/dataflow/analysis/Store.java
```diff
@@ -54,6 +54,25 @@
...
+     *
+     * <p><em>Important</em>: This method must fulfill the following contract:
+     *
+     * <ul>
+     *   <li>Does not change {@code this}.
+     *   <li>Does not change {@code other}.
+     *   <li>Returns a fresh object which is not aliased yet.
+     *   <li>Returns an object of the same (dynamic) type as {@code this}, even if the signature is
```
wmdietlGC: What do you mean with the `signature` here?

framework/src/org/checkerframework/framework/flow/CFAbstractStore.java
```diff
@@ -788,6 +788,15 @@ public S copy() {
...
     public S leastUpperBound(S other) {
+        return upperBound(other, true);
+    }
+
+    @Override
+    public S widenUpperBound(S other) {
+        return upperBound(other, false);
+    }
+
+    public S upperBound(S other, boolean lub) {
```
wmdietlGC: New method should be documented.

framework/src/org/checkerframework/framework/flow/CFAbstractStore.java
```diff
@@ -809,7 +820,13 @@ public S leastUpperBound(S other) {
         {
             V otherVal = other.thisValue;
             V myVal = thisValue;
-            V mergedVal = myVal == null ? null : myVal.leastUpperBound(otherVal);
+            V mergedVal =
```
wmdietlGC: This logic is again duplicated multiple times in this file.
Why is this the only time that the value can be null?

framework/src/org/checkerframework/framework/flow/CFAbstractStore.java
wmdietlGC: A similar parameter below is called `shouldWiden`. It would be nice to use the same name and logic for the boolean parameter.

framework/src/org/checkerframework/framework/flow/CFAbstractValue.java
```diff
@@ -336,43 +344,53 @@ public V leastUpperBound(/*@Nullable*/ V other) {
...
 
         public LubVisitor(
                 TypeMirror result,
                 TypeMirror aTypeMirror,
                 TypeMirror bTypeMirror,
                 Set<AnnotationMirror> aSet,
                 Set<AnnotationMirror> bSet,
-                Set<AnnotationMirror> lubSet) {
+                Set<AnnotationMirror> lubSet,
+                boolean widen) {
```
wmdietlGC: `widen` -> `shouldWiden` for consistency. Or other way around.

framework/src/org/checkerframework/framework/flow/CFAbstractValue.java
```diff
@@ -336,43 +344,53 @@ public V leastUpperBound(/*@Nullable*/ V other) {
...
             super(result, aTypeMirror, bTypeMirror, aSet, bSet);
             this.lubSet = lubSet;
+            this.widen = widen;
+        }
+
+        private AnnotationMirror lub(AnnotationMirror a, AnnotationMirror b) {
```
wmdietlGC: Documentation.

wmdietlGC: In this class you introduced a helper method. Above I suggest that you do the same. The name `lub` isn't good, as it doesn't always perform `lub`.

dataflow/src/org/checkerframework/dataflow/analysis/Store.java
smillst: This is copied from the leastUpperBound Javadoc above. I think it is referring to the signature of a method that implements this method.

framework/src/org/checkerframework/framework/flow/CFAbstractStore.java
smillst: I don't really know.  I would guess it's because the other the myVal equivalent is from a map, so the map must have the property that values are never null.

dataflow/src/org/checkerframework/dataflow/analysis/Store.java
smillst: This is in the dataflow package, so that kinds of Values are in the store isn't known.  So, it's impossible to give that advice here.  But as far as how to implement QualifierHierarchy#widdenUpperBound, I have not advice as it depends on the type system.

dataflow/src/org/checkerframework/dataflow/analysis/Store.java
wmdietlGC: I guessed method signature, too, but that doesn't really make sense: it first says that it has to have the same dynamic type. If it's the same type, it has exactly the same method signatures.

dataflow/src/org/checkerframework/dataflow/analysis/Store.java
wmdietlGC: But there must be some advice on when the method is used. Usually it is to jump to a higher abstraction to allow faster termination of a fixed point computation.

dataflow/src/org/checkerframework/dataflow/analysis/Analysis.java
```diff
@@ -521,12 +541,31 @@ protected void addStoreBefore(
...
+                blockCount.put(b, count + 1);
+            }
+        }
 
         if (addBlockToWorklist) {
             addToWorklist(b);
         }
     }
 
+    private S mergeStores(S previousStore, S newStore, boolean shouldWiden) {
```
wmdietl: I think this code would be easier to read as:

```
if (newstore == null) {
  newThenStore = previousStore;
} else if (shouldWiden) {
  newThenStore = newStore.widenUpperBound(previousStore);
} else {
  newThenStore = newStore.leastUpperBound(previousStore);
}
```

framework/src/org/checkerframework/framework/flow/CFAbstractStore.java
```diff
@@ -871,6 +881,10 @@ public S leastUpperBound(S other) {
         return newStore;
     }
 
+    private V upperBound(boolean shouldWiden, V otherVal, V thisVal) {
```
wmdietl: For consistency with the `mergeStores` method above, I would use `(V, V, boolean)`. Some javadoc would still be useful.

framework/src/org/checkerframework/framework/flow/CFAbstractStore.java
```diff
@@ -871,6 +881,10 @@ public S leastUpperBound(S other) {
...
+        return shouldWiden ? thisVal.leastUpperBound(otherVal) : thisVal.widenUpperBound(otherVal);
```
wmdietl: Aren't the two branches in the wrong order?

dataflow/src/org/checkerframework/dataflow/analysis/Store.java
```diff
@@ -54,6 +54,26 @@
     S leastUpperBound(S other);
 
     /**
+     * Compute an upper bound of two stores that is wider than the least upper bound of the two
+     * stores.
+     *
+     * <p>(This might not be require for an analysis, if not required, implementations should call
```
wmdietl: Would this sentence be clearer as 'A particular analysis might not require widening and will instead use leastUpperBound.`
Currently the sentence seems to suggest that an analysis has to implement something specific, where it actually should say that if you don't need this feature, don't call it. But maybe I'm misunderstanding what it is trying to say...

dataflow/src/org/checkerframework/dataflow/analysis/Store.java
smillst:
````
class MyStore implements A {
  MyStore lub(MyStore s) {...}
}

class MySubStore extends MyStore{
  // Doesn't implement lub
}
````

If lub is called on an object with type MySubStore, the result should have the dynamic type of MySubStore.

File: checker/tests/index/Loops.java [hunk 16]
```diff
@@ -49,4 +49,21 @@ public void test4(int[] src) {
         int[] optoSft = new int[patternLength];
         for (int i = patternLength; i > 0; i--) {}
     }
+
+    public void test5(
+            boolean flag,
+            @LTLengthOf(value = "#1", offset = "-1000") int offset,
```
wmdietl: Shouldn't it be illegal to use the boolean first parameter as value of `LTLengthOf`?

checker/tests/index/Loops.java
```diff
@@ -49,4 +49,21 @@ public void test4(int[] src) {
...
+            @LTLengthOf("#1") int offset2) {
+        int otherOffset = offset;
+        while (flag) {
+            otherOffset += 1;
+            offset++;
+            //:: error: (compound.assignment.type.incompatible)
+            offset += 1;
```
wmdietl: Why is this forbidden, but the equivalent `offset++` is allowed?

dataflow/src/org/checkerframework/dataflow/analysis/Analysis.java
```diff
@@ -74,6 +74,18 @@
     protected IdentityHashMap<Block, S> elseStores;
 
     /**
+     * Number of times every block has been analyzed since the last time widening was applied. Null,
+     * if maxCountBeforeWidening is -1 which implies widening isn't using for this analysis.
```
wmdietl: using -> used

dataflow/src/org/checkerframework/dataflow/analysis/Analysis.java
```diff
@@ -462,12 +475,19 @@ protected void addStoreBefore(
             Block b, Node node, S s, Store.Kind kind, boolean addBlockToWorklist) {
         S thenStore = getStoreBefore(b, Store.Kind.THEN);
         S elseStore = getStoreBefore(b, Store.Kind.ELSE);
+        boolean shouldWiden = false;
+        Integer count = null;
+        if (blockCount != null) {
+            count = blockCount.get(b);
+            count = count == null ? 0 : count;
```
wmdietl: The else branch is a self-assignment. To make this clearer, I would write:
```
if (count == null) {
  count = 0;
}
```

framework/src/org/checkerframework/framework/type/QualifierHierarchy.java
```diff
@@ -114,6 +114,33 @@ public abstract boolean isSubtype(
...
+     * If the type hierarchy has an infinite ascending chain, then the dataflow analysis might never
+     * reach a fixed point. To prevent this, implement this method such that it returns an upper
+     * bound for the two qualifiers that is wider than the least upper bound. If this method is
+     * implemented, also override {@link #implementsWidening()} and return true.
+     *
+     * <p>Otherwise, returns the least upper bound of the two annotations.
+     *
+     * @param a1 annotation
+     * @param a2 annotation
+     * @return an upper bound that is wider than the least upper bound of a1 and a2
```
wmdietl: The `is` should be `might be` ?

dataflow/src/org/checkerframework/dataflow/analysis/Analysis.java
```diff
@@ -74,6 +74,18 @@
...
+     * Number of times every block has been analyzed since the last time widening was applied. Null,
+     * if maxCountBeforeWidening is -1 which implies widening isn't using for this analysis.
+     */
+    protected IdentityHashMap<Block, Integer> blockCount;
+
+    /**
+     * Number of times a block can be analyzed before widening. -1 implies that widening shouldn't
+     * be used.
+     */
+    protected int maxCountBeforeWidening = -1;
```
wmdietl: I found how this field gets set confusing. It first is -1 here and then gets over-written in the subclass.
How about making the field final, adding a new constructor that takes a parameter for the field, and have the current constructors call the new constructor with -1?

dataflow/src/org/checkerframework/dataflow/analysis/Store.java
wmdietl: Ah! We should see how to express this more clearly in both cases...

dataflow/src/org/checkerframework/dataflow/analysis/Analysis.java
```diff
@@ -74,6 +74,18 @@
     protected IdentityHashMap<Block, S> elseStores;
 
     /**
+     * Number of times every block has been analyzed since the last time widening was applied. Null,
+     * if maxCountBeforeWidening is -1 which implies widening isn't using for this analysis.
+     */
+    protected IdentityHashMap<Block, Integer> blockCount;
```
wmdietlGC: Building on the comment about making maxCountBeforeWidening final: we could also make this field final, initialize it to something non-null if needed, and then reset the map instead of creating a new one on every `init`.

smillst: All of the other hash maps fields in this class are re-created on every init.  Why is clearing a map faster than creating a new one?

dataflow/src/org/checkerframework/dataflow/analysis/Store.java
smillst: I find the original text clear.  Could you suggest how to change the comment so that it is easy to understand?

PR #1060, opened 2017-02-17
Make LBC a subchecker of UBC; use its power in one case

kelloggm:
The case of interest is described in kelloggm#44: when an array is created with a length equal to a sum, and one of the expressions being summed is `@NonNegative`, the other becomes LTEL of the array. Likewise, if one expression is `@Positive`, the other become LTL.


checker/src/org/checkerframework/checker/index/IndexChecker.java
```diff
@@ -1,22 +1,20 @@
...
 
 /**
  * A type checker for preventing out-of-bounds accesses on arrays. Contains two subcheckers that do
  * all of the actual work: the Lower Bound Checker and the Upper Bound Checker.
  *
  * @checker_framework.manual #index-checker Index Checker
  */
-public class IndexChecker extends LowerBoundChecker {
+public class IndexChecker extends UpperBoundChecker {
     // Ideally, the Index Checker would be an AggregateChecker that ran the Lower and Upper Bound
```
smillst: Just delete this whole comment.  Instead, you can just say something about how the Upper Bound Checker runs the other checkers and then reference https://github.com/typetools/checker-framework/blob/master/checker/src/org/checkerframework/checker/index/Checker-dependencies.txt.  (And that file needs to be updated now.)

smillst: And I would move the contents of that file to a comment in this class.  Otherwise it won't be found.

checker/src/org/checkerframework/checker/index/IndexChecker.java
```diff
@@ -1,22 +1,20 @@
...
-    // is a subclass of the LowerBoundChecker.  If isIndexChecker() returns true, then the
-    // LowerBoundChecker runs the UpperBoundChecker as a subchecker.
+    // is a subclass of the UpperBoundChecker, which runs the LowerBoundChecker as a subchecker
+    // anyway.
 
-    @Override
-    protected boolean isIndexChecker() {
-        return true;
-    }
+    // TODO: we should discuss what's actually going on here, because at this point the Index Checker
```
smillst: Yes, it's just for the name now, which seems right to me.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
```diff
@@ -54,6 +56,7 @@ public UpperBoundTransfer(CFAnalysis analysis) {
             Receiver arrayRec =
                     FlowExpressions.internalReprOf(analysis.getTypeFactory(), node.getTarget());
             String arrayString = arrayRec.toString();
+
```
smillst: Delete the new line to minimize the diff.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
```diff
@@ -124,14 +131,30 @@ private void knownToBeArrayLength(
...
         store.insertValue(leftRec, atypeFactory.convertUBQualifierToAnnotation(newLeft));
 
         UBQualifier newRight = right.glb(newInfo.plusOffset(node.getLeftOperand(), atypeFactory));
+        newRight = accountForLowerBoundAnnos(node.getLeftOperand(), newRight, arrayExp);
+
         Receiver rightRec = FlowExpressions.internalReprOf(atypeFactory, node.getRightOperand());
         store.insertValue(rightRec, atypeFactory.convertUBQualifierToAnnotation(newRight));
     }
 
+    /** If the node is NN but not POS, add an LTEL to the qual. If POS, an LTL. */
```
smillst: 'is NN but not POS` is just zero?

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
```diff
@@ -124,14 +131,30 @@ private void knownToBeArrayLength(
...
+    private UBQualifier accountForLowerBoundAnnos(Node node, UBQualifier qual, String arrayExp) {
+        if (atypeFactory.hasLowerBoundTypeByClass(node, NonNegative.class)) {
+            if (atypeFactory.hasLowerBoundTypeByClass(node, Positive.class)) {
+                qual = qual.glb(UBQualifier.createUBQualifier(arrayExp, "0"));
```
smillst: I don't see how this branch will be reached since the node will either be NN or POS, but not both.

checker/tests/index/Pilot3ArrayCreation.java
```diff
@@ -1,7 +1,5 @@
 // This test case is for issue 44: https://github.com/kelloggm/checker-framework/issues/44
 
-// @skip-test until the test is fixed
-
 class Pilot3ArrayCreation {
     void test(int[] firstArray, int[] secondArray[]) {
         int[] newArray = new int[firstArray.length + secondArray.length];
```
smillst: Would you please add more test cases?  Such as where one of the operands is positive and greater than negative one.

checker/src/org/checkerframework/checker/index/IndexChecker.java
kelloggm: The comments in that file were up-to-date wrt the changes I'm making in this pull request already; an oversight on my part, I suppose.

kelloggm: I moved it all to IndexChecker.java, though.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
kelloggm: Sort of. But a NonNegative could be Positive _most_ of the time, but not in every conceivable situation. Or the checker might just not be able to prove it.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
kelloggm: I'll refactor this to be better. I agree with you that it's wrong. These should just be separate cases at the top level.

kelloggm:
I've addressed all your comments and added more comprehensive tests.

checker/tests/index/Pilot3ArrayCreation.java
kelloggm: The new test ArrayCreationChecks.java covers all these cases I think

Issue #1006, opened 2017-01-05
Exception during nullness checking of stream items


twistedsquare:
Here's a minimal version of the problematic code:

```
import java.util.function.Function;
import java.util.stream.*;
import java.util.*;

public class StreamProblem
{
    public static Map<String, Long> getFreq(String src)
    {
        return Stream.of(src).collect(Collectors.groupingBy(Function.identity(), Collectors.counting()));
    }	
}
```

Compiling with the nullness processor (version 2.1.7) gives an exception in the checker:
```
javac -cp checker-2.1.7.jar -processor org.checkerframework.checker.nullness.NullnessChecker -AprintErrorStack StreamProblem.java
error: AsSuperVisitor: Unexpected combination: type: NULL superType: NULL.
  type:  null
  superType:  null
  Compilation unit: StreamProblem.java
  Exception: java.lang.Throwable; Stack trace: org.checkerframework.framework.source.SourceChecker.errorAbort(SourceChecker.java:721)
  org.checkerframework.javacutil.ErrorReporter.errorAbort(ErrorReporter.java:25)
  org.checkerframework.framework.type.visitor.AbstractAtmComboVisitor.defaultAction(AbstractAtmComboVisitor.java:44)
  org.checkerframework.framework.type.visitor.AbstractAtmComboVisitor.visitNull_Null(AbstractAtmComboVisitor.java:377)
  org.checkerframework.framework.util.AtmCombo.accept(AtmCombo.java:514)
  org.checkerframework.framework.type.visitor.AbstractAtmComboVisitor.visit(AbstractAtmComboVisitor.java:50)
  org.checkerframework.framework.type.AsSuperVisitor.visit(AsSuperVisitor.java:86)
  org.checkerframework.framework.type.AsSuperVisitor.asSuper(AsSuperVisitor.java:68)
  org.checkerframework.framework.util.AnnotatedTypes.asSuper(AnnotatedTypes.java:131)
  org.checkerframework.framework.util.AtmLubVisitor.lubWithNull(AtmLubVisitor.java:70)
  org.checkerframework.framework.util.AtmLubVisitor.lub(AtmLubVisitor.java:53)
  org.checkerframework.framework.util.AnnotatedTypes.leastUpperBound(AnnotatedTypes.java:638)
  org.checkerframework.framework.util.AnnotatedTypes.leastUpperBound(AnnotatedTypes.java:619)
[...]

Issue #979, opened 2016-12-06
Implement Java 8 type argument inference


smillst:
The Checker Framework should implement [Java 8 type inference](https://docs.oracle.com/javase/specs/jls/se8/html/jls-18.html) for method type arguments, new class trees that use diamonds, and raw types.  (Currently, the Checker Framework implements Java 7 type inference.) 
 This involves completely replacing/rewriting the implementation in the `framework.utils.typeinference` package.  Also, the code that adds type arguments to raw types (`AnnotatedDeclaredType#getTypeArguments`) and the code that infers type arguments for diamonds (`AnnotatedTypeFactory#fromNewClass`) should be changed to use type argument inference.

The Java 8 algorithm specified in the JLS cannot be used as is.  It must be adapted to use with annotated types.  In particular, annotated types complicate the JLS  algorithm in following ways. 

1.  In the JLS, if there is a subtyping constraint against a null type, the constraint is reduced to "true".[[JLS  #18.2.3]](https://docs.oracle.com/javase/specs/jls/se8/html/jls-18.html#jls-18.2.3)  However, if null isn't annotation with the bottom qualifier, then the constraint should be reduced to new constraint where the primary annotation of the type must be a subtype of the annotation on null.

2.  If a use of a type variable has a primary annotation, then constraints against that use should not include that annotation.  In other words, that constraint should be against the Java type rather than the annotated type.  But if the type system has a multigraph qualifier hierarchy, then there should still be a constraint against an annotated type for some of the hierarchies.   (See examples in checker/tests/nullness/generics/MethodTypeVars7.java.) 

3. A class can be marked with `@Covariant` to signify that one or more of its type parameters can be treated covariantly.  Covariant type parameters should generate subtyping constraints rather than equality constraints.

4. Method postconditions. See #1345. 

Also, once Java 8 inference is implemented, the field `AnnotatedWildcardType#uninferredTypeArgument` should be removed and the `-AconservativeUninferredTypeArguments` option should also be removed.
In the meanwhile, if you are concerned that the Checker Framework is suffering false negatives due to this issue, use `-AconservativeUninferredTypeArguments`.  A user [reported](https://github.com/typetools/checker-framework/issues/2235#issuecomment-452862815) that this required editing 0.5% of the lines of code in his project.

[...]

PR #1065, opened 2017-02-21
Work-around for Issue #979.

wmdietlGC:
Includes an option to get the conservative behavior we currently have.
This required propagating information about uninferred wildcards in asSuper and LUB computations.
Added new test cases. Existing test cases for #979 now pass, except for one that now fails because of #1006.


framework/src/org/checkerframework/framework/type/DefaultTypeHierarchy.java
```diff
@@ -116,6 +116,9 @@
     // passing annotations to qualifierHierarchy.
     protected AnnotationMirror currentTop;
 
+    // Whether to work around Issue 979
+    protected final boolean shouldWorkAroundIssue979;
```
smillst: I would name this something like `ignoreUninferredTypeArgs` so that we don't have to remember or look up what issue 979 is.

framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java
```diff
@@ -2027,7 +2027,8 @@ public AnnotatedTypeMirror getErased() {
         // argument inference and raw type handling is improved.
         private boolean uninferredTypeArgument = false;
 
-        /* package-scope */ void setUninferredTypeArgument() {
+        // This method should only be used within the framework
```
smillst: This should be a Javadoc comment.

wmdietlGC:
Yes, this is the patch I had previously mentioned. I had to expand it to also account for situations where the uninferred TA wildcard was LUBed with something else, so there are a few more changes now.
I've improved the comments a bit.
Do note that this is just a hacky workaround for Issue 979 and that we shouldn't close that issue. Correctly implementing a fix for Issue 979 would still be great.

smillst:
Could you please add the new command line option to the change log?

wmdietlGC:
Added something, including the other issues I have recently closed.

smillst:
Thanks!  You don't have to worry about adding closed issue.  I query the issue tracker and collect all the issues that have been closed since the last release.

(And this is fine to merge once the tests pass.)

PR #1067, opened 2017-02-22
Make subclasses of BaseTypeVisitor obey skipDefs.

### LINKED ISSUE #1036: Certain checkers partially ignore skipDefs/onlyDefs
Some checkers, such as the Initialization checker and the Interning checker, override `BaseTypeVisitor.visitClass` and execute code before (and after) calling `super.visitClass`. So some code is executed even for classes which should be ignored.

This is a problem for me, since I have one (generated) class with almost 4000 static final variables which I want all checkers to ignore, but `InitializationVisitor.visitClass` calls `atypeFactory.preProcessClassTree` on it which takes a LONG time (many minutes) before eventually `super.visitClass` is called.

Copy'n'pasting the beginning of `BaseTypeVisitor.visitClass` (which calls `checker.shouldSkipDefs(node)` to determine whether to skip this class) fixes this problem.

smillst:
Fixes #1036

### DIFF
--- a/checker/src/org/checkerframework/checker/initialization/InitializationVisitor.java
+++ b/checker/src/org/checkerframework/checker/initialization/InitializationVisitor.java
@@ -268,7 +268,7 @@ public class InitializationVisitor<
     protected final List<VariableTree> initializedFields;
 
     @Override
-    public Void visitClassOverride(ClassTree node, Void p) {
+    public void visitClassOverride(ClassTree node) {
         // call the ATF with any node from this class to trigger the org.checkerframework.dataflow
         // analysis.
         atypeFactory.preProcessClassTree(node);
@@ -292,7 +292,7 @@ public class InitializationVisitor<
             }
         }
 
-        Void result = super.visitClass(node, p);
+        super.visitClassOverride(node);
 
         // Is there a static initializer block?
         boolean hasStaticInitializer = false;
@@ -323,8 +323,6 @@ public class InitializationVisitor<
             List<AnnotationMirror> receiverAnnotations = Collections.emptyList();
             checkFieldsInitialized(node, isStatic, store, receiverAnnotations);
         }
-
-        return result;
     }
 
     @Override
--- a/checker/src/org/checkerframework/checker/initialization/InitializationVisitor.java
+++ b/checker/src/org/checkerframework/checker/initialization/InitializationVisitor.java
@@ -269,10 +269,6 @@ public class InitializationVisitor<
 
     @Override
     public void processClassTree(ClassTree node) {
-        // call the ATF with any node from this class to trigger the org.checkerframework.dataflow
-        // analysis.
-        atypeFactory.preProcessClassTree(node);
-
         // go through all members and look for initializers.
         // save all fields that are initialized and do not report errors about
         // them later when checking constructors.
--- a/checker/src/org/checkerframework/checker/initialization/InitializationVisitor.java
+++ b/checker/src/org/checkerframework/checker/initialization/InitializationVisitor.java
@@ -268,13 +268,7 @@ public class InitializationVisitor<
     protected final List<VariableTree> initializedFields;
 
     @Override
-    public Void visitClass(ClassTree node, Void p) {
-        if (checker.shouldSkipDefs(node)) {
-            // Not "return super.visitClass(node, p);" because that would
-            // recursively call visitors on subtrees; we want to skip the
-            // class entirely.
-            return null;
-        }
+    public Void visitClassOverride(ClassTree node, Void p) {
         // call the ATF with any node from this class to trigger the org.checkerframework.dataflow
         // analysis.
         atypeFactory.preProcessClassTree(node);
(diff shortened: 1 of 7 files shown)


framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -260,50 +260,58 @@ public Void scan(Tree tree, Void p) {
...
+        visitorState.setClassType(atypeFactory.getAnnotatedType(classTree));
+        visitorState.setClassTree(classTree);
         visitorState.setMethodReceiver(null);
         visitorState.setMethodTree(null);
         visitorState.setAssignmentContext(null);
 
         try {
-            if (!TreeUtils.hasExplicitConstructor(node)) {
-                checkDefaultConstructor(node);
+            if (!TreeUtils.hasExplicitConstructor(classTree)) {
```
wmdietlGC: An alternative solution to this issue might be to extract this block (maybe a bit more or less) into a new method and then have that method as the common extension point that subclasses should override if they want to change behavior.
That way, most subclasses don't need to remember to also skip classes in certain situation, which this PR has to copy in 5 places.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -261,21 +261,33 @@ public Void scan(Tree tree, Void p) {
...
+
+    /**
+     * Type-check classTree. Subclasses should override this method instead of {@link
+     * #visitClass(ClassTree, Void)}.
+     *
+     * @param classTree class to check
+     * @param p null
+     * @return null
+     */
+    public Void visitClassOverride(ClassTree classTree, Void p) {
```
felixfontein: Maybe this would be a good moment to get rid of the `Void p` argument for `visitClassOverride()`. Or is the value of `p` actually needed somewhere?

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -260,50 +260,70 @@ public Void scan(Tree tree, Void p) {
...
-        if (checker.shouldSkipDefs(node)) {
-            // Not "return super.visitClass(node, p);" because that would
+    public final Void visitClass(ClassTree classTree, Void p) {
+        if (checker.shouldSkipDefs(classTree)) {
+            // Not "return super.visitClass(classTree, p);" because that would
             // recursively call visitors on subtrees; we want to skip the
             // class entirely.
             return null;
         }
+        return visitClass(classTree, p);
```
wmdietlGC: Shouldn't this be `visitClassOverride`?

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
wmdietlGC: I agree with Felix. We can't get rid of the `Void` in the visitor method, but we can have a nicer signature for this method.

Also, let's brainstorm a nicer method name. How about `processClassTree`, similar to the `pre/postProcessClassTree` we have in the ATF.

framework/src/org/checkerframework/common/util/report/ReportVisitor.java
```diff
@@ -140,7 +140,13 @@ public Void visitImport(ImportTree node, Void p) {
     */
 
     @Override
-    public Void visitClass(ClassTree node, Void p) {
+    public Void visitClassOverride(ClassTree node, Void p) {
+        if (checker.shouldSkipDefs(node)) {
```
wmdietlGC: This part can be removed?

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
smillst: I changed the signature to be void and only take the ClassTree as a parameter.

If we call it `processClassTree`, then visitClass should be changed to this:

```
if (shouldSkip(classTree)) {
   return null;
}
atypeFactory.preProcessClassTree(classTree);
try {
  processClassTree(classTree);
  atypeFactory.postProcessClassTree(classTree);
}  finally {
  // reset the visitor state
}
return null;
```

wmdietlGC: It would definitely be nice if updating the visitor state is done outside of the method that subclasses override. So if you add "initialize visitor state" somewhere in your code, that would be nice.
I don't quite see why that is connected with picking a meaningful name, but that's a good change.

smillst:
@wmdietlGC This is ready to review again.  Note the "downstream" tests failed because I will need to adapt inference to this change.

checker/src/org/checkerframework/checker/initialization/InitializationVisitor.java
```diff
@@ -268,7 +268,7 @@ public Void visitBlock(BlockTree node, Void p) {
     protected final List<VariableTree> initializedFields;
 
     @Override
-    public Void visitClass(ClassTree node, Void p) {
+    public void processClassTree(ClassTree node) {
         // call the ATF with any node from this class to trigger the org.checkerframework.dataflow
         // analysis.
         atypeFactory.preProcessClassTree(node);
```
wmdietlGC: This is already part of visitClass, so I don't think we still need to call it here.

PR #1068, opened 2017-02-22
Adds support for subtracting NonNegatives and Positives from UBC annotations

kelloggm:


checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -460,10 +467,15 @@ public Void visitBinary(BinaryTree tree, AnnotatedTypeMirror type) {
...
                 case PLUS:
-                case MINUS:
                     // Dataflow refines this type if possible
                     type.addAnnotation(UNKNOWN);
                     break;
+                case MINUS:
+                    // Dataflow refines this type most of the time.
+                    // But when a NonNegative is subtracted from a LTL,
+                    // the result is still LTL without an offset.
+                    addAnnotationForMinus(left, right, type);
```
smillst: This needs to be done in dataflow or else information about offsets will be lost.  See test cases I added: https://github.com/typetools/checker-framework/pull/1068/commits/aaa2cabd04d9e47a0b867e369cb9cb652d28dfca

kelloggm:
I've made the changes you requested, especially moving the code into Transfer. It fixes the new tests you provided, and also fixes at least one other test I discovered. I think it should be ready to merge modulo the tests passing, and the zero-expressions you merged in; not sure why you did that, though.

smillst:
> the zero-expressions you merged in; not sure why you did that, though.

Because I wanted to use "0" as an offset in my example, but "0" wasn't recognized as an int literal.  That bug is fixed is PR #1071.  

I'll review these changes. Thanks!

smillst:
This change be merged once all the tests pass and PR #1071 has been merged.

framework/src/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -100,9 +100,9 @@ private static Pattern anchored(String regex) {
     /** Matches a field access. Capturing groups 1 and 2 are the object and field. */
     protected static final Pattern memberselect = anchored("([^.]+)\\.(.+)");
     /** Matches integer literals */
-    protected static final Pattern intPattern = anchored("[-+]?[1-9][0-9]*");
+    protected static final Pattern intPattern = anchored("0|[-+]?[1-9][0-9]*");
```
mernst: Please undo the changes to this line and `longPattern`.

framework/tests/flowexpression/ValueLiterals.java
```diff
@@ -0,0 +1,11 @@
+import testlib.flowexpression.qual.FlowExp;
```
mernst: This looks like it shouldn't be showing up as a diff in your pull request, either.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
```diff
@@ -303,14 +303,23 @@ private void refineNeqArrayLength(
     /**
      * If some Node a is known to be less than the length of some array, x, then, the type of a - b,
```
mernst: Delete commas after "array", "then", and "b".

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
```diff
@@ -303,14 +303,23 @@ private void refineNeqArrayLength(
...
      * is @LTLengthOf(value="x", offset="b"). If b is known to be less than the length of some other
```
mernst: Should the offset be `-b`?  I would think that "a+b" is `@LtLengthOf("x", "b")`.

mernst:
There are conflicts that need to be resolved.
Also, something about this pull requests seems to be making all the Travis tests time out.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
smillst: No, if `a < x.length`, then `a - b` is less than x.length if `b` is added to `a - b` and thus the offset is `+b`.

mernst: What is the semantics of `@LtLengthOf` with an offset?  It's not documented, so I don't have a way of interpreting whether a given expression has that type.

smillst: I added documentation to LTLengthOf in PR #1073.

PR #1070, opened 2017-02-22
Index Checker: Refine the operands of multiplication expressions.

smillst:
If the expression is known to be the length of some array and the other operand is nonnegative or positive.

Fixes kelloggm/issue#88

### DIFF
--- a/checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
+++ b/checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
@@ -82,25 +82,25 @@ public class UpperBoundTransfer extends IndexAbstractTransfer {
         } else if (node instanceof NumericalMultiplicationNode) {
             Node right = ((NumericalMultiplicationNode) node).getRightOperand();
             Node left = ((NumericalMultiplicationNode) node).getLeftOperand();
-            knownToBeArrayLengthMultiplication(right, left, array, in, store);
             knownToBeArrayLengthMultiplication(left, right, array, in, store);
+            knownToBeArrayLengthMultiplication(right, left, array, in, store);
         }
     }
 
     /**
-     * The positive node times node is known to be exactly the length of arrayExp. If the positive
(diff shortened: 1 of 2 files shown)


checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
```diff
@@ -77,6 +79,37 @@ private void knownToBeLessThanLengthOf(
...
+            Node left = ((NumericalMultiplicationNode) node).getLeftOperand();
+            knownToBeArrayLengthMultiplication(right, left, array, in, store);
+            knownToBeArrayLengthMultiplication(left, right, array, in, store);
+        }
+    }
+
+    /**
+     * The positive node times node is known to be exactly the length of arrayExp. If the positive
+     * node is really positive, then node is less than or equal to the length of arrayExp. If
+     * positive is greater than 1, then node is less than the length of arrayExp.
```
mernst: This was hard for me to interpret on first reading.  Can you put the identifiers within `{@code ...}` to help readers parse the text?

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
```diff
@@ -77,6 +79,37 @@ private void knownToBeLessThanLengthOf(
...
+            Node right = ((NumericalMultiplicationNode) node).getRightOperand();
+            Node left = ((NumericalMultiplicationNode) node).getLeftOperand();
+            knownToBeArrayLengthMultiplication(right, left, array, in, store);
+            knownToBeArrayLengthMultiplication(left, right, array, in, store);
+        }
+    }
+
+    /**
+     * The positive node times node is known to be exactly the length of arrayExp. If the positive
+     * node is really positive, then node is less than or equal to the length of arrayExp. If
```
mernst: Confusing:  `positive` might not really be positive?  Should it have a different name in that case?

PR #1071, opened 2017-02-22
0 and 0L are valid Java expression

smillst:


framework/src/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -100,9 +100,9 @@ private static Pattern anchored(String regex) {
     /** Matches a field access. Capturing groups 1 and 2 are the object and field. */
     protected static final Pattern memberselect = anchored("([^.]+)\\.(.+)");
     /** Matches integer literals */
-    protected static final Pattern intPattern = anchored("[-+]?[1-9][0-9]*");
+    protected static final Pattern intPattern = anchored("[-+]?[1-9]*[0-9]");
```
mernst: Looks wrong.  Does this still match 100?

framework/tests/flowexpression/ValueLiterals.java
```diff
@@ -2,4 +2,10 @@
 
 public class ValueLiterals {
     void test(@FlowExp("0") Object a, @FlowExp("0L") Object b) {}
+
+    void test2(@FlowExp("1000") Object a, @FlowExp("100L") Object b) {}
+    //:: error: (expression.unparsable.type.invalid)
+    void test3(@FlowExp("01000") Object a) {}
+    //:: error: (expression.unparsable.type.invalid)
+    void test4(@FlowExp("0100L") Object b) {}
```
mernst: This is legal -- it represents the value 64.  What is the reason to forbid it?

Also, `-0L` is legal but is forbidden by the current regex.

PR #1081, opened 2017-02-25
FlowExpressionParse should parse array lengths and clone()

smillst:
Is there a better way to get the element for `length` or `clone` than using the resolver?


mernst:
Looks OK to me.  I don't have an answer to your question.  Maybe @wmdietlGC does.

mernst:
I see that this pull request handles `length`, but I don't see that it handles `clone`.  Is that missing, or did it already work, or is the commit message wrong?

wmdietl:
Not sure whether there would be a better way; as this isn't performance critical, I think we can just do this.
Regarding not handling clone explicitly: there is a check for method calls on arrays. That should include clone, hashCode, and toString, and any other method callable on an array.

PR #1082, opened 2017-02-25
Allows UB transfer to handle arrays that are cast when they're created

kelloggm:
Consider the following testcase:

```java

class ArrayIntroWithCast<T> {

   void test(String[] a, String[] b) {
        Object result = new Object[a.length + b.length];
	System.arraycopy(a, 0, result, 0, a.length);
    }
    
    void test2(String[] a, String[] b) {
	@SuppressWarnings("unchecked")
        T[] result = (T[]) new Object[a.length + b.length];
	System.arraycopy(a, 0, result, 0, a.length);
    }
}
```

The first function (`test`) compiles with no warnings, but the second function (`test2`) does not in master. The only difference is the addition of the cast to a generic argument. This PR fixes this problem, allowing both versions to compile. See kelloggm#106

### DIFF
--- a/checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
+++ b/checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
@@ -40,8 +40,8 @@ public class UpperBoundTransfer extends IndexAbstractTransfer {
     }
 
     // Refine the type of expressions used as an array dimension to be
-    // less than length of the array to which the new array is
-    // assigned.  For example int[] array = new int[expr]; the type of expr is @LTEqLength("array")
+    // less than length of the array to which the new array is assigned.
+    // For example, in "int[] array = new int[expr];", the type of expr is @LTEqLength("array").
     @Override
     public TransferResult<CFValue, CFStore> visitAssignment(
             AssignmentNode node, TransferInput<CFValue, CFStore> in) {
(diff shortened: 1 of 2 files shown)


checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
```diff
@@ -46,8 +47,20 @@ public UpperBoundTransfer(CFAnalysis analysis) {
...
         TransferResult<CFValue, CFStore> result = super.visitAssignment(node, in);
 
-        if (node.getExpression() instanceof ArrayCreationNode) {
-            ArrayCreationNode acNode = (ArrayCreationNode) node.getExpression();
+        Node expNode = node.getExpression();
+
+        ArrayCreationNode acNode =
+                expNode instanceof ArrayCreationNode ? (ArrayCreationNode) expNode : null;
+
+        if (expNode instanceof TypeCastNode) {
```
mernst: Shouldn't this be a `while` loop rather than an `if` statement?  And shouldn't it also call `TreeUtils.skipParens`?  The `while`  loop could be patterned after that one.  Or there could be a new routine that skips over parens and typecasts.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
```diff
@@ -46,8 +47,20 @@ public UpperBoundTransfer(CFAnalysis analysis) {
...
+            Node innerExpNode = ((TypeCastNode) expNode).getOperand();
+            acNode =
+                    innerExpNode instanceof ArrayCreationNode
+                            ? (ArrayCreationNode) innerExpNode
+                            : null;
```
mernst: Under what circumstances can the result be `null`?  I would think you would get to an `ArrayCreationNode` eventually.  Please remove this possibility or document under what circumstances you cannot get to an `ArrayCreationNode`.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
kelloggm: It should call skipParens, but I don't see how/why there would be more than one type cast over an array creation node. I can add such a loop if you think there's a compelling case for it, but I don't know why a programmer would write that.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
kelloggm: It might be the case that there is no array creation node. This is visiting all assignment statements, and then only actually doing something when there is actually an array creation node on the right side. So, I structured the code to look for an array creation node, and then if there isn't one to set the variable to null. Since there is more than one way to find an array creation node, it might be set to non-null in more than one place. Then, there is a check to see if there was actually an array creation node.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
kelloggm: Actually, it can't call `TreeUtils.skipParens`, because it's a `Node` and we want it to stay at that level. Further, I don't see a ParenNode or anything similar; does such a thing even exist at the Node level?

smillst: Parenthesis are skipped when the control flow graph is created, so there is no node for them.

PR #1090, opened 2017-02-28
Index Checker: value offsets

smillst:
If an expression has a know value, then both the know value and the expression should be used as offsets.  This fixes the first part of https://github.com/kelloggm/checker-framework/issues/108.

### DIFF
--- a/checker/src/org/checkerframework/checker/index/upperbound/OffsetEquation.java
+++ b/checker/src/org/checkerframework/checker/index/upperbound/OffsetEquation.java
@@ -383,10 +383,37 @@ public class OffsetEquation {
     }
 
     /**
-     * Creates an offset equation from the Node.
-     *
-     * <p>If node is an int value known at compile time, then the return equation is just the int
+     * If node is an int value known at compile time, then the returned equation is just the int
      * value or if op is '-', the return equation is the negation of the int value.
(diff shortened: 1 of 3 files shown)


checker/src/org/checkerframework/checker/index/upperbound/OffsetEquation.java
```diff
@@ -383,11 +383,38 @@ private static int indexOf(String string, char a, char b, int index) {
...
+     * @param op '+' or '-'
+     * @return an offset equation from value of known or null if the value isn't known
+     */
+    public static OffsetEquation createOffsetFromNodesValue(
+            Node node, UpperBoundAnnotatedTypeFactory factory, char op) {
+        if (node.getTree() != null && TreeUtils.isExpressionTree(node.getTree())) {
+            Integer i;
+            if (op == '+') {
+                i = factory.valMinFromExpressionTree((ExpressionTree) node.getTree());
+            } else {
```
kelloggm: Shouldn't this be an `else if` that checks for `'-'` instead of just an else? Then you can add an else block that also returns null.

smillst: I added an assert.

PR #1091, opened 2017-02-28
Allow SameLen to correctly handle assignments.

kelloggm:
Consider the following test case, from kelloggm#110:

```java

class SimpleCollection {
    private int[] values;

    @IndexOrHigh("values") int size() {
	return values.length;
    }

    void interact_with_other(SimpleCollection other) {
	for(int j = 0; j < other.size(); j++) {
	    int k = other.values[j]; // No warning from index checker
	}
	int[] othervalues = other.values;
	for(int i = 0; i < other.size(); i++) {
	    int k = othervalues[i]; // Index checker issues warning
	}
    }
}
```

This patch fixes that extraneous warning by adding a rule so that when something is assigned into an array, that array and the thing being assigned become part of the same SameLen type. It also corrects a bug in propagateCombinedSameLen that would sometimes pass a null TreePath as the currentPath. I'm not certain of the cause of that, but this fixes it by doing nothing when the TreePath is null. Also fixes a typo in some of the names in SameLenTransfer#visitAssignment, which referred to the right side of an assignment (i.e. the source) as the "left side", which doesn't make sense.

### DIFF
--- a/checker/src/org/checkerframework/checker/index/samelen/SameLenTransfer.java
+++ b/checker/src/org/checkerframework/checker/index/samelen/SameLenTransfer.java
@@ -119,22 +119,19 @@ public class SameLenTransfer extends CFTransfer {
      */
     private void propagateCombinedSameLen(
             AnnotationMirror combinedSameLen, Node node, CFStore store) {
+        TreePath currentPath = aTypeFactory.getPath(node.getTree());
+        if (currentPath == null) {
+            return;
+        }
         for (String s : SameLenUtils.getValue(combinedSameLen)) {
             Receiver recS;
             try {
-                TreePath currentPath = aTypeFactory.getPath(node.getTree());
-                if (currentPath != null) {
-                    recS = aTypeFactory.getReceiverFromJavaExpressionString(s, currentPath);
-                } else {
-                    return;
-                }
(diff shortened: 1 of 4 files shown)


checker/src/org/checkerframework/checker/index/samelen/SameLenTransfer.java
```diff
@@ -77,14 +78,30 @@ public SameLenTransfer(CFAnalysis analysis) {
...
                 aTypeFactory
                         .getAnnotatedType(node.getExpression().getTree())
                         .getAnnotationInHierarchy(UNKNOWN);
 
-        if (leftAnno != null && AnnotationUtils.areSameByClass(leftAnno, SameLen.class)) {
+        // If the left side of the assignment is an array, then have both the right and left side be SameLen
+        // of each other.
+
+        if (node.getTarget().getType().getKind() == TypeKind.ARRAY) {
+            AnnotationMirror combinedSameLen =
```
smillst: Would you please break this up into multiple lines so it is easier to read?

kelloggm: I've done this.

PR #1097, opened 2017-03-02
SameLen: refine operands of == and != correctly.

smillst:


checker/src/org/checkerframework/checker/index/samelen/SameLenTransfer.java
```diff
@@ -188,14 +179,18 @@ AnnotationMirror getAnno(Node n) {
...
             CFValue secondValue,
             boolean notEqualTo) {
-
-        if (notEqualTo) {
-            // Refinement in the else store if this is a.length != b.length.
-            refineEq(firstNode, secondNode, result.getElseStore());
-        } else {
-            // Refinement in the then store if this is a.length == b.length.
-            refineEq(firstNode, secondNode, result.getThenStore());
+        CFStore equal = notEqualTo ? result.getElseStore() : result.getThenStore();
```
kelloggm: Can we rename this to `equalStore` is something else that indicates its a store? The code below is difficult to read with it named `equal`

checker/tests/index/SameLenEqualsRefinement.java
```diff
@@ -0,0 +1,42 @@
+import org.checkerframework.checker.index.qual.*;
+import org.checkerframework.dataflow.qual.Pure;
+
+class SameLenEqualsRefinementa {
```
kelloggm: Name is wrong: should be `SameLenEqualsRefinement` (extra a at the end)

Issue #437, opened 2015-07-03
failure to recognize generic type in non-static inner class


GoogleCodeExporter:
```
What steps will reproduce the problem?
1. Run the checker on the attached files

What is the expected output? What do you see instead?
I expect the checker to pass, or at least not fail in this way.
Foo.java:9

found   : T extends @Initialized @Nullable Object
required: @UnknownInitialization @Nullable Integer

Even though T in this case is Integer



What version of the product are you using? On what operating system?
1.8.11 Linux

Please provide any additional information below.

```

Original issue reported on code.google.com by `long...@google.com` on 22 Jun 2015 at 9:04

Attachments:
- [Bar.java](https://storage.googleapis.com/google-code-attachments/checker-framework/issue-437/comment-0/Bar.java)
- [Foo.java](https://storage.googleapis.com/google-code-attachments/checker-framework/issue-437/comment-0/Foo.java)

JonathanBurke:
I have looked into this more closely.  Unfortunately, the required fix is a bit more complicated and may take some time to implement.

smillst:
Remove warning suppression on framework/tests/all-systems/Issue263.java for regex and tainting when this issue is fixed.

wmdietlGC:
I've added the test from this issue as `framework/tests/all-systems/Issue437.java` and removed the warning suppressions on Issue263.

Issue #134, opened 2015-07-03
Type checking does not respect bound type variables in enclosing classes


GoogleCodeExporter:
```
Another test case I've discovered: InnerImpl.get()'s return type should be Foo, 
not a V extends Object.


$ java -Xbootclasspath/p:$CHECKERS/binary/jsr308-all.jar -jar 
$CHECKERS/binary/jsr308-all.jar -processor checkers.regex.RegexChecker
 Test7.java
Test7.java:13: error: incompatible types.
        Foo foo = get();
                     ^
  found   : V extends Object
  required: Foo
1 error


$ cat Test7.java
class Test7 {
  public interface Foo {}

  class Outer<V> {
    class Inner {
      V get() { return null; }
    }
  }

  class OuterImpl extends Outer<Foo> {
    class InnerImpl extends Inner {
      void test() {
        Foo foo = get();
      }
    }
  }
}
```

Original issue reported on code.google.com by `mdemp...@google.com` on 8 Mar 2012 at 5:44

wmdietlGC:
Fixed by https://github.com/typetools/checker-framework/pull/1099

PR #1099, opened 2017-03-03
Improvements to handling of nested classes: correctly populate type

wmdietlGC:
arguments of enclosing classes, output enclosing classes in error
messages, and enable and add related tests.

### DIFF
--- a/framework/tests/all-systems/Issue263.java
+++ b/framework/tests/all-systems/Issue263.java
@@ -1,7 +1,6 @@
 // Test case for Issue 263:
 // https://github.com/typetools/checker-framework/issues/263
 
-@SuppressWarnings({"regex", "tainting"})
 abstract class Outer<T> {
 
     public class Inner {
--- a/framework/tests/all-systems/Issue263.java
+++ b/framework/tests/all-systems/Issue263.java
@@ -1,8 +1,6 @@
 // Test case for Issue 263:
 // https://github.com/typetools/checker-framework/issues/263
 
-// Suppression required because of Issue 724
-// https://github.com/typetools/checker-framework/issues/724
 @SuppressWarnings({"regex", "tainting"})
 abstract class Outer<T> {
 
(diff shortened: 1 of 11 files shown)


framework/tests/all-systems/Issue263.java
```diff
@@ -1,8 +1,6 @@
 // Test case for Issue 263:
 // https://github.com/typetools/checker-framework/issues/263
 
-// Suppression required because of Issue 724
-// https://github.com/typetools/checker-framework/issues/724
 @SuppressWarnings({"regex", "tainting"})
```
smillst: Should the warning suppress be removed now?

wmdietlGC: Yes, I don't know why I only removed the comments.

wmdietlGC:
Which comment do you mean with 'Does this pull request fix 724? If so, this comment in AnnotatedTypes#outerAsSuper needs to be addressed:'
The link seems to point to all changes.

wmdietlGC:
I found this one: https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/util/AnnotatedTypes.java#L185
Let me try whether that can also be removed.

framework/tests/all-systems/Issue263.java
wmdietlGC: Removing the SuppressWarnings works.

wmdietlGC:
Removing https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/util/AnnotatedTypes.java#L185 doesn't work unfortunately.
So I guess some of the cases of Issue #724 are fixed, but not all.
This PR still fixes a few actual crashes, so I would argue for merging it.

smillst:
These changes look correct to me and should be merged once the test pass.  Is #134 fixed by these changes?

wmdietlGC:
I added a test for #134 and for #437, uncommented some other tests, and added my own test that made me interested in this issue.
Some parts of #724 also work now, but https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/util/AnnotatedTypes.java#L185 can't be removed yet. I'll add a comment to that effect to #724, but leave it open.
I'm still debugging a few other crashes that are also related to inner classes, so more will come.

PR #1100, opened 2017-03-03
Add refinement to Upper Bound Checker:

smillst:
If expression i has type @LTLengthOf(value = "f2", offset = "f1.length") int and expression j is less than or equal to the length of f1, then the type of i + j is @LTLengthOf("f2").

Fixes kelloggm/checker-framework#108 and fixes kelloggm/checker-framework#113

### DIFF
--- a/checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
+++ b/checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
@@ -671,7 +671,8 @@ public abstract class UBQualifier {
         /**
          * Returns a copy of this qualifier with array-offset pairs where in the original the offset
          * contains an access of an array length in arrays. The array length access has been removed
-         * from the offset.
+         * from the offset. If the original qualifier has no array length offsets, then UNKNOWN is
+         * returned.
          *
          * @param arrays access of the length of these arrays are removed
          * @return Returns a copy of this qualifier with some offsets removed
@@ -795,8 +796,8 @@ public abstract class UBQualifier {
          * applied to each offset.
          *
          * <p>If the {@link OffsetEquationFunction} returns null, it's not added as an offset. If
-         * after all functions have been applied, an array as no offsets, then that array is not
-         * added to the returned qualifier. If no arrays are added to the returned qualifier, the
+         * after all functions have been applied, an array has no offsets, then that array is not
+         * added to the returned qualifier. If no arrays are added to the returned qualifier, then
          * UNKNOWN is returned.
          *
          * @param f function to apply
--- a/checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
+++ b/checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
@@ -681,13 +681,14 @@ public abstract class UBQualifier {
             if (arrays.isEmpty()) {
                 return UpperBoundUnknownQualifier.UNKNOWN;
             }
-            return computeNewOffsets(
+            OffsetEquationFunction removeArrayLengthsFunc =
                     new OffsetEquationFunction() {
                         @Override
                         public OffsetEquation compute(OffsetEquation eq) {
                             return eq.removeArrayLengths(arrays);
                         }
-                    });
+                    };
+            return computeNewOffsets(removeArrayLengthsFunc);
         }
         /**
          * Returns a copy of this qualifier with array-offset pairs where in the original the offset
@@ -701,7 +702,7 @@ public abstract class UBQualifier {
             if (arrays.isEmpty()) {
                 return UpperBoundUnknownQualifier.UNKNOWN;
             }
-            return computeNewOffsets(
+            OffsetEquationFunction removeArrayLenFunc =
                     new OffsetEquationFunction() {
                         @Override
                         public OffsetEquation compute(OffsetEquation eq) {
@@ -714,17 +715,19 @@ public abstract class UBQualifier {
                             }
                             return newEq;
                         }
-                    });
+                    };
+            return computeNewOffsets(removeArrayLenFunc);
         }
 
         private UBQualifier addOffset(final OffsetEquation newOffset) {
-            return computeNewOffsets(
+            OffsetEquationFunction addOffsetFunc =
                     new OffsetEquationFunction() {
                         @Override
                         public OffsetEquation compute(OffsetEquation eq) {
                             return eq.copyAdd('+', newOffset);
                         }
-                    });
+                    };
+            return computeNewOffsets(addOffsetFunc);
         }
 
         /**
@@ -742,7 +745,7 @@ public abstract class UBQualifier {
             if (divisor == 1) {
                 return this;
             } else if (divisor > 1) {
-                return computeNewOffsets(
+                OffsetEquationFunction divideFunc =
                         new OffsetEquationFunction() {
                             @Override
                             public OffsetEquation compute(OffsetEquation eq) {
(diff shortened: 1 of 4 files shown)


checker/tests/index/UpperBoundRefinement.java
```diff
@@ -0,0 +1,41 @@
...
+        double[] f1 = new double[10];
+        double[] f2 = new double[20];
+
+        for (int j = 0; j < f2.length; j++) {
+            f2[j] = j;
+        }
+        for (int i = 0; i < f2.length - f1.length; i++) {
+            //fill up f1 with elements of f2
+            for (int j = 0; j < f1.length; j++) {
+                f1[j] = f2[i + j]; // Index checker warns here
```
kelloggm: We should remove this comment since this change is all about removing this warning!

checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
```diff
@@ -652,16 +668,62 @@ public UBQualifier minusOffset(int value) {
...
-                }
-                plusMap.put(entry.getKey(), plus);
+        /**
+         * Returns a copy of this qualifier with array-offset pairs where in the original the offset
+         * contains an access of an array length in arrays. The array length access has been removed
+         * from the offset.
+         *
+         * @param arrays access of the length of these arrays are removed
+         * @return Returns a copy of this qualifier with some offsets removed
+         */
```
kelloggm: The explanation here isn't very clear. I'm not sure if having at least one array-offset pair is a precondition of the function or not; it seems to me that if there aren't any such pairs in the input this will return null (based on the call to OffsetEquation#removeArrayLengths). The documentation should be explicit about this precondition. There also seems to be an inconsistency between what happens when `arrays` is empty (return UpperBoundUnknown) and what happens when there are no array-offset pairs in the qualifier (return null).

checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
```diff
@@ -652,16 +668,62 @@ public UBQualifier minusOffset(int value) {
...
+                    });
+        }
+        /**
+         * Returns a copy of this qualifier with array-offset pairs where in the original the offset
+         * contains an access of an array length in arrays. The array length access has been removed
+         * from the offset. If the offset also has -1 then -1 is also removed.
+         *
+         * @param arrays access of the length of these arrays are removed
+         * @return Returns a copy of this qualifier with some offsets removed
+         */
```
kelloggm: Same comment as above on this explanation; in addition, you explicitly return null below in this version. An explanation here of what the difference between returning null and returning UBU would be helpful.

checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
```diff
@@ -721,6 +776,52 @@ public String toString() {
...
+             */
+            OffsetEquation compute(OffsetEquation eq);
+        }
+
+        /**
+         * Returns a new qualifier that is a copy of this qualifier with the OffsetEquationFunction
+         * applied to each offset.
+         *
+         * <p>If the {@link OffsetEquationFunction} returns null, it's not added as an offset. If
+         * after all functions have been applied, an array as no offsets, then that array is not
```
kelloggm: "as" should be "has"

checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
```diff
@@ -721,6 +776,52 @@ public String toString() {
...
+         * added to the returned qualifier. If no arrays are added to the returned qualifier, the
```
kelloggm: "then" not "the"

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
```diff
@@ -346,11 +352,49 @@ private void refineNeqArrayLength(
...
     }
 
     /**
+     * Return the result of adding i to j,when expression i has type @LTLengthOf(value = "f2",
+     * offset = "f1.length") int and expression j is less than or equal to the length of f1, then
+     * the type of i + j is @LTLengthOf("f2").
+     *
+     * <p>Return the result of adding i to j,when expression i has type @LTLengthOf(value = "f2",
+     * offset = "f1.length - 1") int and expression j is less than the length of f1, then the type
+     * of i + j is @LTLengthOf("f2").
```
kelloggm: Duplicate comment

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
```diff
@@ -346,11 +352,49 @@ private void refineNeqArrayLength(
...
+     *
+     * @param i expression added to j
```
kelloggm: These are both "the type of the expression added to j"/"the type of the expression added to i"

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
```diff
@@ -346,11 +352,49 @@ private void refineNeqArrayLength(
...
+            if (i.isLessThanLengthOf(array)) {
+                lessThan.add(array);
+            } else if (i.hasArrayWithOffsetNeg1(array)) {
+                lessThanOrEqaul.add(array);
+            }
+        }
+        UBQualifier lessThanEqQ = j.removeArrayLengthAccess(lessThanOrEqaul);
+        UBQualifier lessThanQ = j.removeArrayLengthAccessAndNeg1(lessThan);
+
+        return lessThanEqQ.glb(lessThanQ);
```
kelloggm: How this works is non-obvious when reading the code. I think a comment in this method somewhere explaining how it's accomplished would be helpful (the javadoc comment does a good job explaining what is done, but the how also deserves an explanation).

checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
smillst: 1. an array-offset pair is not required.  If one isn't found than UNKNOWN is returned.  (I'll add this to the Javadoc.)

2.  This method does not return null. OffsetEquation#removeArrayLengths returns null, but computeNewOffsets never returns null.  Would the following be clearer:

````java
     public UBQualifier removeArrayLengthAccess(final List<String> arrays) {
            if (arrays.isEmpty()) {
                return UpperBoundUnknownQualifier.UNKNOWN;
            }
            OffsetEquationFunction removeArrayLen =  new OffsetEquationFunction() {
                        @Override
                        public OffsetEquation compute(OffsetEquation eq) {
                            return eq.removeArrayLengths(arrays);
                        }
                    };
            return computeNewOffsets(removeArrayLen);
        }
````

checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
smillst: Same explanation as above.  I'll rewrite this method like the one above if that seems clearer to you.

smillst:
@kelloggm Could you please take another look?  In particular at this reply (https://github.com/typetools/checker-framework/pull/1100#discussion_r104745517).  (I stopped the travis jobs as I'm going to change the code again depending on your reply to that comment.)

checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
kelloggm: Yes, I think that (2) would be clearer. I agree with the addition of (1) to the javadoc.

kelloggm:
I've replied; sorry I've been slow to respond today!

smillst:
Thanks! I just wanted to make sure the comment didn't get lost!

smillst:
@kelloggm This is ready for you to review again if you want.

kelloggm:
I've looked over it again and I think it's good to go.

PR #1101, opened 2017-03-03
Unsignedness checker

cmackie:
Mainly introduces checking for casts which render the signedness of right shifts irrelevant, similar to earlier work to check masks.


checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -91,6 +106,34 @@ private boolean maskIgnoresMSB(Kind maskKind, LiteralTree numBitsLit, LiteralTre
...
+     * <pre>{@code (short) (myInt >> 16)}</pre>
+     *
+     * and for
+     *
+     * <pre>{@code (short) (myInt >>> 16)}</pre>
+     *
+     * because these two expressions are guaranteed to have the same result.
+     *
+     * @param numBitsLit the LiteralTree whose value is numBits
+     * @param bitDiff the difference between the widths of the type of the right shift (which is int
```
mernst: I find this definition of the formal parameter confusing.  I think the interface would be easier to understand if this routine took the two bit widths and did the subtraction itself.  There doesn't seem to be much benefit to the client doing the subtraction instead of moving it into this routine.

Or, you could pass the types into this routine.

Or just inline the routine into the one place it is called, since it is only two lines long.

In any event, it seems odd to have such a confusing interface and so much repeated documentation to support a two-line method.

checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -149,6 +192,99 @@ private boolean isMaskedShift(BinaryTree shiftExpr) {
     }
 
     /**
+     * Determines if a right shift operation, {@code >>} or {@code >>>}, is type casted such that
+     * the cast renders the shift signedness ({@code >>} vs {@code >>>}) irrelevent by destroying
```
mernst: I would use "discarding" rather than "destroying".

checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -91,6 +106,73 @@ private boolean maskIgnoresMSB(Kind maskKind, LiteralTree numBitsLit, LiteralTre
...
+     * because these two expressions are guaranteed to have the same result.
+     *
+     * @param shiftTypeKind the kind of the type of the shift literal (BYTE, CHAR, SHORT, INT, or
+     *     LONG)
+     * @param castTypeKind the kind of the cast target type (BYTE, CHAR, SHORT, INT, or LONG)
+     * @param numBitsLit the LiteralTree whose value is numBits
+     * @return true iff introduced bits are discarded
+     */
+    private boolean castIgnoresMSB(
+            TypeKind shiftTypeKind, TypeKind castTypeKind, LiteralTree numBitsLit) {
```
mernst: `numBits` is a confusing name, because in the body of the routine there are a lot of variables that count a number of bits.  When you name a variable, name it according to its *semantics*, not its *type*.  (The type here is the number of bits, which is more informative than naming it `i` for `int` but still doesn't get at the semantics.)  I would call it `shiftAmount` or something like that.

checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -91,6 +106,73 @@ private boolean maskIgnoresMSB(Kind maskKind, LiteralTree numBitsLit, LiteralTre
...
+        // Determine number of bits in the shift type, note shifts upcast to int
+        long shiftBits = 0;
```
mernst: Never initialize a variable with a value that will never be used.  That's confusing, because programmers will assume that a value is assigned only because it is important and may be used.  In this case, you can just leave off the initializer.

checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -91,6 +106,73 @@ private boolean maskIgnoresMSB(Kind maskKind, LiteralTree numBitsLit, LiteralTre
...
+        switch (shiftTypeKind) {
+            case INT:
+                shiftBits = 32;
+                break;
+            case LONG:
+                shiftBits = 64;
+                break;
+            default:
```
mernst: What are the other possibilities?  If there are no other possibilities, then make the default case throw an exception to indicate that there is a problem, rather than masking the error by silently proceeding with a possibly-confusing result.

checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -91,6 +106,73 @@ private boolean maskIgnoresMSB(Kind maskKind, LiteralTree numBitsLit, LiteralTre
...
+                return false;
+        }
+
+        // Determine number of bits in the cast type
+        long castBits = 0;
```
mernst: Remove initializer.

checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -91,6 +106,73 @@ private boolean maskIgnoresMSB(Kind maskKind, LiteralTree numBitsLit, LiteralTre
...
+                castBits = 32;
+                break;
+            case LONG:
+                castBits = 64;
+                break;
+            default:
+                return false;
+        }
+
+        // Determine the bit difference between the shift and cast and the number of bits shifted
```
mernst: This comment is not very helpful.  It just repeats what occurs below.  And, it is confusing because it describes two lines of code but the two occurrences of "and" are confusing to parse.

checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -91,6 +106,73 @@ private boolean maskIgnoresMSB(Kind maskKind, LiteralTree numBitsLit, LiteralTre
...
+        long bitDiff = shiftBits - castBits;
```
mernst: This variable name describes the type but not the semantics.  I would call it `bitsDiscarded`, which is more informative and makes the code easier to read.

checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -91,6 +106,73 @@ private boolean maskIgnoresMSB(Kind maskKind, LiteralTree numBitsLit, LiteralTre
...
+        long numBits = getLong(numBitsLit.getValue());
```
mernst: This is incorrect.  Section 15.19 of the Java Language Specification says that only the 5 or 6 least significant bits are used, so you need to do the same.  Note that using only the least significant bits yields a non-negative shift amount.
You could add some tests with cast amounts that are negative or ar bigger than 63.

checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -91,6 +106,73 @@ private boolean maskIgnoresMSB(Kind maskKind, LiteralTree numBitsLit, LiteralTre
...
+
+        return numBits <= bitDiff || numBits == 0;
```
mernst: Renaming variables will make the first part of this test easier to understand:  `shiftAmount < bitsDiscarded`.
Could the second half be done early in the routine rather than here?  It's about a separate case.

checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -149,6 +231,61 @@ private boolean isMaskedShift(BinaryTree shiftExpr) {
...
+            }
+        }
+
+        if (!isPrimitiveCast(enclosing)) {
+            return false;
+        }
+
+        TypeCastTree cast = (TypeCastTree) enclosing;
+
+        // Determine the type of the cast target
```
mernst: This code feels repetitive to me.  `isPrimitiveCast` just did much of this work, and now you are doing it again.
How about replacing `isPrimitiveCast` by `primitiveCastType` that returns null if the argument is not a primitive cast, and otherwise returns the type?  Then the client (this code) can just test against null.  It reduces code duplication, reduces run time, and makes this code shorter and easier to understand by abstracting out implementation details.

checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -149,6 +231,61 @@ private boolean isMaskedShift(BinaryTree shiftExpr) {
...
+        ExpressionTree underlyingType = annotatedType.getUnderlyingType();
+        PrimitiveTypeTree castPrimitiveType = (PrimitiveTypeTree) underlyingType;
+        TypeKind castTypeKind = castPrimitiveType.getPrimitiveTypeKind();
+
+        // Determine the type of the shift result
+        TypeKind shiftTypeKind =
+                atypeFactory.getAnnotatedType(shiftExpr).getUnderlyingType().getKind();
+
+        // Determine shift literal
+        ExpressionTree shift = shiftExpr.getRightOperand();
```
mernst: It's a bit confusing that you have three variables named `shift*`, and only this one has no suffix indicating its type.
I don't have a concrete suggestion, however, and this comment is more minor than other ones in this review.
(I suppose you could abstract it out into a routine that returns an Integer, or null if the shift amount is not known at compile time.)

checker/src/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -34,6 +31,24 @@ private boolean isMask(Tree node) {
...
 
+    /** @return true iff node is a primitive type cast */
+    private boolean isPrimitiveCast(Tree node) {
+        if (node.getKind() != Kind.TYPE_CAST) {
+            return false;
+        }
+
+        TypeCastTree cast = (TypeCastTree) node;
+        Tree castType = cast.getType();
+        if (castType.getKind() != Kind.ANNOTATED_TYPE) {
```
cmackie: I believe that by the time the visitor traverses the AST all types are wrapped by some annotated type, but it is possible that I'm wrong about this.

PR #1105, opened 2017-03-03
Issue unboxing.of.nullable in more places

### LINKED ISSUE #1087: Unboxing of @Nullable in assignments
Take

```

class Unboxing {
  void foo(@Nullable Integer in) {
    // I would expect unboxing.of.nullable, but instead get assignment.type.incompatible.
    int q = in;
  }

  int bar(@Nullable Integer in) {
    // Similarly here with return.type.incompatible.
    return in;
  }
}
```

I would expect `unboxing.of.nullable` as error for both the assignment and the return statement; that would be clearer than the current incompatible type messages.

There are tests in checker//tests/nullness/Boxing.java that contain other expressions where the expected error is produced.

smillst:
Fixes #1087

### DIFF
--- a/checker/src/org/checkerframework/checker/nullness/NullnessVisitor.java
+++ b/checker/src/org/checkerframework/checker/nullness/NullnessVisitor.java
@@ -203,6 +203,24 @@ public class NullnessVisitor
         atypeFactory.replacePolyQualifier(varType, valueExp);
         super.commonAssignmentCheck(varType, valueExp, errorKey);
     }
+
+    @Override
+    protected void commonAssignmentCheck(
+            AnnotatedTypeMirror varType,
+            AnnotatedTypeMirror valueType,
+            Tree valueTree,
+            String errorKey) {
+        if (TypesUtils.isPrimitive(varType.getUnderlyingType())
+                && !TypesUtils.isPrimitive(valueType.getUnderlyingType())) {
+            boolean succeed = checkForNullability(valueType, valueTree, UNBOXING_OF_NULLABLE);
+            if (!succeed) {
+                // Only issue the unboxing of nullable error.
+                return;
+            }
+        }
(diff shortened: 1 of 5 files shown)


checker/src/org/checkerframework/checker/nullness/NullnessVisitor.java
```diff
@@ -424,16 +442,33 @@ public Void visitTypeCast(TypeCastTree node, Void p) {
...
      */
-    private void checkForNullability(ExpressionTree tree, /*@CompilerMessageKey*/ String errMsg) {
+    private boolean checkForNullability(
+            ExpressionTree tree, /*@CompilerMessageKey*/ String errMsg) {
         AnnotatedTypeMirror type = atypeFactory.getAnnotatedType(tree);
+        return checkForNullability(type, tree, errMsg);
+    }
+
+    /**
+     * Issues a 'dereference.of.nullable' if the type is not of a {@link NonNull} type.
```
wmdietl: Shouldn't this first sentence be changed as the version above, i.e. replace the 'dereference.of.nullable' by 'the error message'.
Some indication of what the difference between the two methods is would also be good.

checker/src/org/checkerframework/checker/nullness/NullnessVisitor.java
```diff
@@ -424,16 +442,33 @@ public Void visitTypeCast(TypeCastTree node, Void p) {
     // ///////////// Utility methods //////////////////////////////
 
     /**
-     * Issues a 'dereference.of.nullable' if the type is not of a {@link NonNull} type.
+     * Issues the error message if the type is not of a {@link NonNull} type.
      *
      * @param tree the tree where the error is to reported
```
wmdietl: 'is to reported' -> 'is to be reported' or 'should be reported'? Same in the copy below.

PR #1106, opened 2017-03-03
Only report error in Pre/Post condition expression once.

### LINKED ISSUE #1051: Duplicate error for invalid RequiresNonNull
Take this simple example:

```
import org.checkerframework.checker.nullness.qual.RequiresNonNull;

class ReqNN {
  @RequiresNonNull("foo")
  void bar() {}
}
```

Running the Nullness Checker gives:

```
ReqNN.java:5: error: [flowexpr.parse.error] cannot parse the expression 'foo: identifier not found'
  void bar() {}
       ^
ReqNN.java:5: error: [flowexpr.parse.error] cannot parse the expression 'foo: identifier not found'
  void bar() {}
       ^
2 errors
```

One error would be better.

Running javac with `-doe` shows the two locations where the error is generated:

```
...
	at org.checkerframework.framework.source.SourceChecker.report(SourceChecker.java:1460)
	at org.checkerframework.framework.flow.CFAbstractTransfer.addInformationFromPreconditions(CFAbstractTransfer.java:533)
	at org.checkerframework.framework.flow.CFAbstractTransfer.initialStore(CFAbstractTransfer.java:255)
	at org.checkerframework.framework.flow.CFAbstractTransfer.initialStore(CFAbstractTransfer.java:94)
	at org.checkerframework.dataflow.analysis.Analysis.init(Analysis.java:440)
	at org.checkerframework.dataflow.analysis.Analysis.performAnalysis(Analysis.java:161)
...
```

and

```
...
	at org.checkerframework.framework.source.SourceChecker.report(SourceChecker.java:1460)
	at org.checkerframework.common.basetype.BaseTypeVisitor.checkContractsAtMethodDeclaration(BaseTypeVisitor.java:524)
	at org.checkerframework.common.basetype.BaseTypeVisitor.visitMethod(BaseTypeVisitor.java:458)
	at org.checkerframework.checker.initialization.InitializationVisitor.visitMethod(InitializationVisitor.java:355)
	at org.checkerframework.checker.initialization.InitializationVisitor.visitMethod(InitializationVisitor.java:57)
...
```

smillst:
Fixes #1051

### DIFF
--- a/checker/tests/nullness/RequiresNonNullTest.java
+++ b/checker/tests/nullness/RequiresNonNullTest.java
@@ -143,4 +143,10 @@ class RequiresNonNullTest {
     void test() {
         o = null;
     }
+
+    // Test case for Issue 1051
+    // https://github.com/typetools/checker-framework/issues/1051
+    //:: error: (flowexpr.parse.error)
+    @RequiresNonNull("thisShouldIssue1Error")
+    void testIssue1051() {}
 }
--- a/checker/tests/nullness/RequiresNonNullTest.java
+++ b/checker/tests/nullness/RequiresNonNullTest.java
@@ -144,9 +144,9 @@ class RequiresNonNullTest {
         o = null;
     }
 
+    @RequiresNonNull("thisShouldIssue1Error")
     // Test case for Issue 1051
     // https://github.com/typetools/checker-framework/issues/1051
     //:: error: (flowexpr.parse.error)
-    @RequiresNonNull("thisShouldIssue1Error")
     void testIssue1051() {}
 }
--- a/framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
+++ b/framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
@@ -529,8 +529,7 @@ public abstract class CFAbstractTransfer<
                                 false);
                 info.insertValue(expr, annotation);
             } catch (FlowExpressionParseException e) {
-                // report errors here
-                analysis.checker.report(e.getResult(), methodTree);
+                // Errors are reported by BaseTypeVisitor.checkContractsAtMethodDeclaration()
             }
         }
     }
--- a/checker/jtreg/nullness/preciseErrorMsg/Class1.java
+++ b/checker/jtreg/nullness/preciseErrorMsg/Class1.java
@@ -1,0 +1,15 @@
+/*
+ * @test
+ * @summary
+ * Test case for Issue 1051
+ * https://github.com/typetools/checker-framework/issues/1051
+ *
+ * @compile/fail/ref=Class1.out -XDrawDiagnostics -Xlint:unchecked -processor org.checkerframework.checker.nullness.NullnessChecker -AprintErrorStack -Anomsgtext Class1.java
+ *
+ */
+import org.checkerframework.checker.nullness.qual.RequiresNonNull;
+
+public class Class1 {
+    @RequiresNonNull("instanceField")
+    public static void foo() {}
+}
--- a/checker/jtreg/nullness/preciseErrorMsg/Class1.out
+++ b/checker/jtreg/nullness/preciseErrorMsg/Class1.out
@@ -1,0 +1,2 @@
+Class1.java:14:24: compiler.err.proc.messager: (flowexpr.parse.error)
+1 error


checker/tests/nullness/RequiresNonNullTest.java
```diff
@@ -143,4 +143,10 @@ void notHiddenTest() {
     void test() {
         o = null;
     }
+
+    // Test case for Issue 1051
+    // https://github.com/typetools/checker-framework/issues/1051
+    //:: error: (flowexpr.parse.error)
```
wmdietl: Does this test actually fail if there are two errors? Doesn't it need to be a jtreg test to look at all reported errors?

smillst: I thought so, but I'll double check.

framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
```diff
@@ -529,8 +529,7 @@ protected void addInformationFromPreconditions(
                                 false);
                 info.insertValue(expr, annotation);
             } catch (FlowExpressionParseException e) {
-                // report errors here
-                analysis.checker.report(e.getResult(), methodTree);
```
wmdietlGC: Similar code is further down in `processPostconditions` and `processConditionalPostconditions`.
Shouldn't those also be changed?

More generally, I think no error reporting should be done as part of flow analysis. Not all type systems use flow and checks should be implemented in the visitor. Flow refinement strictly improves the precision, but doesn't report additional bugs.

In `org.checkerframework.framework.flow.CFAbstractAnalysis#checker` the documentation states: `A checker used to do error reporting.`
Once we remove the three error reports with this PR, the only other uses of that field are to look up options and warning suppressions.
Can we update the documentation of that field as part of this PR, to document that we don't want any use of `checker.report` in the flow part.

checker/tests/nullness/RequiresNonNullTest.java
```diff
@@ -143,4 +143,10 @@ void notHiddenTest() {
     void test() {
         o = null;
     }
+
+    @RequiresNonNull("thisShouldIssue1Error")
+    // Test case for Issue 1051
+    // https://github.com/typetools/checker-framework/issues/1051
+    //:: error: (flowexpr.parse.error)
```
wmdietlGC: The error would be much better on the `@RequiresNonNull`. A method might have multiple annotations that use flow expressions and a single error on the method isn't very useful.
I couldn't find an issue to update this. Should I create one?

framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
smillst: >Similar code is further down in `processPostconditions` and `processConditionalPostconditions`.
Shouldn't those also be changed?

The BaseTypeVisitor parses preconditions at both the declaration of a method and at it uses.  In both cases, if parsing fails, an error is used.  The BaseTypeVisitor does not parse postconditions at uses of methods, so an error won't be issued if parsing fails.  If this check is added to the BaseTypeVisitor, then these other two error reports can be removed.  (In this case, only type systems that use flow can benefit from postcondition annotations.)

checker/tests/nullness/RequiresNonNullTest.java
smillst: It's fine if you want to create one.  Note, that the `flowexpr.parse.error` includes the expression string that failed to parse.  So the user can figure out which annotation failed, though it would be nicer if the error were on the correct annotation.

framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
wmdietlGC: I'm not sure what you mean with 'In this case...`.
If we move these checks to the BaseTypeVisitor, all type systems would benefit from them, not just those that use flow refinement. That would seem like a good thing. Do you agree with this or not?

smillst: I think it's a neutral thing to move the checks (though moving the checks means that the expressions will be parsed twice, which isn't a good thing). The only type systems that have postcondition annotations are those that use flow refinement, so it doesn't benefit more type systems to move these errors.  

I agree that flow refinement should not issue type-checking errors, but flow refinement might issue other errors/warnings related to flow refinement.  I would argue that these errors are related to flow refinement rather than type-checking errors. So, I don't think it is a problem to leave the checks as is.  However, if you feel strongly that they should be in the BaseTypeVisitor, then it is ok with me if you move them.

checker/tests/nullness/RequiresNonNullTest.java
wmdietl: Created https://github.com/typetools/checker-framework/issues/1129 to track this.

framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
wmdietl: Summary of discussion: without flow, one gets a parse error only at the method declaration. With flow, one gets an additional error at method invocations. Post-conditions are only used for flow refinement, so it would be unnecessary overhead to also check them at invocations.

PR #1107, opened 2017-03-03
Correctly check for the invariant annotation on type vars.

### LINKED ISSUE #1044: Uninitialized field check for type parameters with @Nullable bound
The following program:

```
public class TypeVarInit<V> {
  V f;

    new TypeVarInit<String>().f.toString();
  }
}
```

compiles without error with the Nullness Checker.
When executing it, it results in a NullPointerException:

```
	at TypeVarInit.main(TypeVarInit.java:5)
```

I would have expected:

```
TypeVarInit.java:1: error: [initialization.fields.uninitialized] the constructor does not initialize fields: f
public class TypeVarInit<V> {
       ^
1 error
```

Note that if the upper bound of `V` is changed from the implicit `@Nullable Object` to `@NonNull Object` we get the expected error:

```
TypeVarInit.java:1: error: [initialization.fields.uninitialized] the constructor does not initialize fields: f
public class TypeVarInit<V extends Object> {
       ^
1 error
```

smillst:
Fixes #1044


checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -159,6 +159,17 @@ protected boolean isInitializationAnnotation(AnnotationMirror anno) {
      */
     public abstract AnnotationMirror getFieldInvariantAnnotation();
 
+    /**
+     * Returns whether or not field has the invariant annotation.
```
wmdietlGC: Replacing `field` with `{@code field}` would make it clearer that the parameter is meant here.

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -159,6 +159,17 @@ protected boolean isInitializationAnnotation(AnnotationMirror anno) {
...
+     *
+     * <p>If the field is a type variable, this method return false if any possible instantiation of
```
wmdietlGC: return -> returns.

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -159,6 +159,17 @@ protected boolean isInitializationAnnotation(AnnotationMirror anno) {
...
+     * the type parameter could have the invariant annotation.
+     *
+     * @param field field that might have invariant annotation
+     * @return whether or not field has the invariant annotation
+     */
+    protected abstract boolean hasInvariantAnnotation(VariableTree field);
```
wmdietlGC: For symmetry with `getFieldInvariantAnnotation` how about naming this `hasFieldInvariantAnnotation`?

checker/src/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
```diff
@@ -507,8 +507,26 @@ public NullnessTypeAnnotator(InitializationAnnotatedTypeFactory<?, ?, ?, ?> atyp
 
     @Override
     public AnnotationMirror getFieldInvariantAnnotation() {
-        Elements elements = processingEnv.getElementUtils();
-        return AnnotationUtils.fromClass(elements, NonNull.class);
+        return NONNULL;
+    }
+
+    /**
+     * Returns whether or not field has the invariant annotation.
```
wmdietlGC: Same comments as on javadoc for the overridden method. Can the common parts be inherited to reduce duplication?

checker/src/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
```diff
@@ -507,8 +507,26 @@ public NullnessTypeAnnotator(InitializationAnnotatedTypeFactory<?, ?, ?, ?> atyp
...
+     *
+     * <p>If the field is a type variable, this method return false if any possible instantiation of
+     * the type parameter could have the invariant annotation. In other words, is the lower
+     * bound @NonNull?
+     *
+     * @param field field that might have invariant annotation
+     * @return whether or not field has the invariant annotation
+     */
+    @Override
+    protected boolean hasInvariantAnnotation(VariableTree field) {
```
wmdietlGC: Isn't this the implementation for all type systems? Instead of making the method abstract, how about using this implementation in the superclass? Then any future subclasses that need to change this behavior could still override the method.

checker/tests/nullness/init/Issue1044.java
```diff
@@ -0,0 +1,47 @@
...
+
+        Inner7() {}
+    }
+
+    static class Inner8<@Nullable T extends @Nullable Object> {
+        T f;
+
+        Inner8() {}
+    }
+}
```
wmdietlGC: For completeness, isn't `Inner9<V extends Object>` missing?

checker/src/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
smillst: This is only the implementation for systems where the invariant annotation is bottom.  If it is some other annotation, then a type is has the invariant annotation if the lower bound annotation is a subtype of (or equal to) the invariant annotation and the upper bound is a super type of (or equal to) the annotation.  

By making this method abstract, it forces future subclasses to implement this method.

wmdietl: Adding some explanation like this to the javadoc would be useful.

checker/src/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
wmdietl: Would using `{@inheritDoc}` remove the duplication?

smillst:
@wmdietl  I've lost track.  Is this ready to be merged or do you want to review it again?

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -159,6 +159,17 @@ protected boolean isInitializationAnnotation(AnnotationMirror anno) {
      */
     public abstract AnnotationMirror getFieldInvariantAnnotation();
 
+    /**
+     * Returns whether or not {@code field} has the invariant annotation.
+     *
+     * <p>If the {@code field} is a type variable, this method returns false if any possible
```
wmdietlGC: Doesn't it return true if any possible instantiation of the type parameter could have the invariant annotation?
Adding a reference to the NullnessATF or an explanation of the check against the lower bound (if the invariant annotation is bottom) would also be helpful.

PR #1109, opened 2017-03-04
Make caches in AnnotationUtils thread safe

### LINKED ISSUE #1108: make caches in AnnotationUtils thread safe
It'd be great if the [static caches in `AnnotationUtils`](https://github.com/typetools/checker-framework/blob/b7f67fc1c973439e8c2644490ca1d94eafc19a06/javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java#L60) could be made into instance fields instead (there are already `TODO` comments in the code to this effect).  Beyond the issue of memory leaks (motivating the `AnnotationUtils.clear()` method), these caches are not thread safe.  I am using the Checker dataflow framework from within [Buck](https://buckbuild.com/), and Buck runs multiple javac instances concurrently within the same JVM.  I got lucky and observed a crash due to a data race on these fields, but the issue could have lurked for a long time.

I have a patch to at least make the caches thread safe, while leaving them as static fields (via locking).  Changing them to be instance fields required more pervasive code changes than I felt capable of.  If doing the proper instance field fix isn't something doable in the short term, I am willing to contribute a fix to at least make them thread safe.

Also, FWIW, I observed that the attempt to bound [the `annotationMirrorNames` cache](https://github.com/typetools/checker-framework/blob/b7f67fc1c973439e8c2644490ca1d94eafc19a06/javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java#L69) using an LRU cache wasn't working.  The cache was still growing without bound, so I had to insert `clear()` calls anyway.  Not sure what exactly was going on; maybe things break when the cache persists across multiple javac instances?

msridhar:
Fixes #1108 

Mostly this is a matter of using Collections.synchronizedMap().  We also use an explicit synchronized block for when a containsKey() operation is followed by a get(), since AnnotationUtils.clear() might be called in between.

### DIFF
--- a/javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
+++ b/javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
@@ -98,10 +98,9 @@ public class AnnotationUtils {
      * @return an {@link AnnotationMirror} of type {@code} name
      */
     public static AnnotationMirror fromName(Elements elements, CharSequence name) {
-        synchronized (annotationsFromNames) {
-            if (annotationsFromNames.containsKey(name)) {
-                return annotationsFromNames.get(name);
-            }
+        AnnotationMirror res = annotationsFromNames.get(name);
+        if (res != null) {
+            return res;
         }
         final DeclaredType annoType = typeFromName(elements, name);
         if (annoType == null) {
@@ -171,10 +170,9 @@ public class AnnotationUtils {
 
     /** @return the fully-qualified name of an annotation as a String */
     public static final /*@Interned*/ String annotationName(AnnotationMirror annotation) {
-        synchronized (annotationMirrorNames) {
-            if (annotationMirrorNames.containsKey(annotation)) {
-                return annotationMirrorNames.get(annotation);
-            }
+        String res = annotationMirrorNames.get(annotation);
+        if (res != null) {
+            return res;
         }
         final DeclaredType annoType = annotation.getAnnotationType();
         final TypeElement elm = (TypeElement) annoType.asElement();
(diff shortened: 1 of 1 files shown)


msridhar:
I ran out of time to do full benchmarking, unfortunately (it seems to take a while).  I wanted to get the PR up in case someone else has time.

javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -67,21 +67,27 @@ public static void clear() {
      * so they can be compared with ==.
      */
     private static final Map<AnnotationMirror, /*@Interned*/ String> annotationMirrorNames =
-            CollectionUtils.createLRUCache(ANNOTATION_CACHE_SIZE);
+            Collections.synchronizedMap(
```
wmdietl: Could you create a new `CollectionUtils.createSynchronizedLRUCache` instead of duplicating this logic multiple times?

javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -92,8 +98,10 @@ public static void clear() {
      * @return an {@link AnnotationMirror} of type {@code} name
      */
     public static AnnotationMirror fromName(Elements elements, CharSequence name) {
-        if (annotationsFromNames.containsKey(name)) {
-            return annotationsFromNames.get(name);
+        synchronized (annotationsFromNames) {
```
wmdietl: Would it be shorter and still thread safe to do:

```
AnnotationMirror res = annotationsFromNames.get(name);
if (res != null) {
  return res;
}
...
```

Instead of checking and then getting the value, why not just get the value and see whether we got something?

javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
msridhar: Are you sure?  It's only duplicated twice (the other two cases wrap a `HashMap` rather than an LRU cache).

wmdietl: Oh, sorry about that. Yes, not worth adding a new method.

msridhar:
Addressed the second comment about just using `get()`; nice catch.

mernst:
Thanks for this contribution, Manu!

PR #1112, opened 2017-03-07
Adds polymorphic annotations to the Index Checkers

smillst:
Fixes kelloggm/issue#73.


kelloggm:
I think this code looks good, but this PR probably should uncomment all of the Poly qualifiers in the index JDK, as well. They all have a space between the `@` and the `Poly...`, which needs to be removed.

smillst:
I found bugs in the default implementations of lub and glb when polymorphic qualifiers are used.  I'm going to extract the fixes for those bugs and post a separate pull request with those changes. Then I'll open a new pull request with the Index Checker poly qualifiers. So, I'm going to close this pull request.

PR #1114, opened 2017-03-08
Tutorial improvements

smillst:
Reference Eclipse plugin manual that contains more information rather than duplicating less information in the tutorial.


mernst:
I don't think "here" should ever be used as anchor text.  I fixed that, and otherwise it looks good.

PR #1116, opened 2017-03-08
Comment out more annotations in String.

kelloggm:
I'm pretty sure they're all commented out now.

### DIFF
--- a/checker/jdk/index/src/java/lang/String.java
+++ b/checker/jdk/index/src/java/lang/String.java
@@ -2992,7 +2992,7 @@ public final class String
      * @return  a string of length <code>1</code> containing
      *          as its single character the argument <code>c</code>.
      */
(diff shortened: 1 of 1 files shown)


wmdietl:
What is the point of all these commented-out annotations? If this is work-in-progress, why don't you do it in a branch?

mernst:
The eventual goal is to support:
 * arrays
 * Strings
 * Lists
We had planned to do arrays and Strings, but have backed off on Strings temporarily.  They could be moved elsewhere, but we hope to get back to them soon and that would make merging harder because there are still some annotations on Strings that are required even for the array support.

I agree it will be better to have them all uncommented and Strings supported, but I hope it won't be too long until that happens.  If we had known we were going to defer Strings, we wouldn't have had them in the repository in the first place, of course.

PR #1118, opened 2017-03-09
Index Checker: update common assignment checks to properly descend into constant arrays

kelloggm:
Resolves kelloggm#109

### DIFF
--- a/checker/tests/index/ConstantArrays.java
+++ b/checker/tests/index/ConstantArrays.java
@@ -23,11 +23,11 @@ class ConstantArrays {
         )
         int[] a = {2, 3, 0};
 
-        //:: error: (array.initializer.type.incompatible)::error: (assignment.type.incompatible)
         @LTLengthOf(
             value = {"b", "b2"},
             offset = {"-2", "5"}
         )
+        //:: error: (array.initializer.type.incompatible)::error: (assignment.type.incompatible)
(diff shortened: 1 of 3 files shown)


kelloggm:
It should; that's embarrassing! It looks like I forgot to add it, but it's there now. My apologies.

checker/tests/index/ConstantArrays.java
```diff
@@ -0,0 +1,8 @@
+import org.checkerframework.checker.index.qual.*;
+
+class ConstantArrays {
+    void test() {
+        int[] b = new int[4];
+        @LTLengthOf("b") int[] a = {0, 1, 2, 3};
+    }
+}
```
smillst: Please add more test cases.  Some that fail, some where the array is less than or equal to some array, somewhere it is less than or equal given some offset.

kelloggm:
I've added more comprehensive tests. It's a good thing I did, too, since I also discovered a serious error in my solution (that was just a typo, but made it totally wrong): when checking if each element of the array was less than the minlen, I was blindly assigning them into the boolean instead of anding the result of each being less than the minlen with the existing boolean result. This means that I was, in effect, just checking if the last element was a legal index.

kelloggm:
Your changes look good and simplify this a lot; thanks! I'll merge when the tests have passed.

PR #1123, opened 2017-03-10
Remove incorrect annotations from binarySearch

mernst:
`binarySearch()` returns a value in the range -a.length-1 to a.length-1, inclusive.  The `@GTENegativeOne` annotations on its return type were incorrect.


kelloggm:
Should the annotation then be `@LTLengthOf("#1")`?

mernst:
Yes, that would be a correct annotation. I have another proposal that I
will write up later tonight.

Issue #1124, opened 2017-03-10
giving same error 2 times


shivanshsoni:
Why it is giving same error 2 times with same location. 

My code was:

![screenshot 11](https://cloud.githubusercontent.com/assets/24546371/23789671/df7b3c88-05a1-11e7-8c62-13b8c1d28b4a.png)

error on cmd:

![screenshot 10](https://cloud.githubusercontent.com/assets/24546371/23789714/0247eef0-05a2-11e7-855f-9a4bf31344c0.png)

mernst:
This issue is a duplicate of issue #1051.
There is already a pull request open, #1106.

PR #1125, opened 2017-03-10
this will show error one time (solved issue#1124)

shivanshsoni:
see issue #1124  created by me. I tried its solution. Please give your review


mernst:
This proposed fix fails tests.  Please revise it so that it passes all the tests.  (It's a good idea to make sure it passes all the tests before you submit a pull request.)

mernst:
There is already a more correct pull request (#1106) open to solve the underlying problem.

PR #1126, opened 2017-03-10
Adds polymorphic qualifiers for the Index Checkers

smillst:
This is the same code that was in (closed) PR #1112, except the new poly qualifiers have proper Javadocs and are mentioned in the manual.  I also add `@PolyIndex` that is syntactic sugar for `@PolyUpperBound @PolyLowerBound`.


mernst:
The documentation looks good to me now.

Issue #1127, opened 2017-03-10
Non-default checker-framework directory name


wmdietlGC:
It is sometimes convenient to have a directory that contains a single `annotation-tools` and `jsr308-langtools` directory, but multiple `checker-framework-x` directories.
Each CF build should stay within their directory and use the shared AFU and langtools.

This is currently prevented by the following line in the top-level `build.xml`:

```
  <property name="checkerframework" value="${basedir}/../checker-framework"/>
```

I would like to change this to:

```
  <property name="checkerframework" value="${basedir}"/>
```

Variable `basedir` is a relative variable, so I don't see the benefit in going up and then down into the wrong `checker-framework` directory.

I introduced this behavior in https://github.com/typetools/checker-framework/commit/4dd30cd024d8d52c7cad2cae3ce9281a82251f77

I think this might be a problem for the down-stream `checker-framework-inference`, but I would argue that it should set the correct `checker-framework` directory instead of making a naming assumption here.

Anyone see a reason not to change this?

wmdietlGC:
PR: https://github.com/typetools/checker-framework/pull/1128

PR #1128, opened 2017-03-10
Use non-specifc base directory for Checker Framework builds.

wmdietlGC:
Fixes https://github.com/typetools/checker-framework/issues/1127


mernst:
This change seems reasonable to me.  Thanks!

PR #1130, opened 2017-03-10
Fix #1120 by treating final classes specially during initialization.

### LINKED ISSUE #1120: Initialization of final classes
Take the following code:

```
class A {
  Object f = new Object();
}

final class B {
  Object g;

  B() {
    // this is @UnderInitialization(A.class)
    g = new Object();
    // this is @UnderInitialization(B.class)
    this.bar();
  }

  void bar() {
    g.toString();
  }
}
```

Running the Nullness Checker gives an error:

```
FinalInit.java:11: error: [method.invocation.invalid] call to bar() not allowed on the given receiver.
                this.bar();
                        ^
  found   : @UnderInitialization(B.class) @NonNull B
  required: @Initialized @NonNull B
```

I cannot think of a situation in which allowing this would be unsound.
The manual in 3.8.4 mentions: "A non-solution is to prevent subclasses from overriding process() by using final on the class or method. This doesn’t work because even if process() is not overridden, it might call other methods that are overridden."
I see the issue with final methods and how giving them special treatment would be unsound.
However, for final classes I do not see an issue. If class B is final, none of its methods can be overridden.
The FbC paper discusses final classes in the comparison with Masked Types, but doesn't discuss this situation.

The changed rule could be something like: For a final class B, `@UnknownInitialization(B.class) B` and `@UnderInitialization(B.class) B` are subtypes of `@Initialized B`.
Or, instead of changing the subtyping relationship, when a final class B initialized all its fields, instead of becoming `@UnderInitialization(B.class) B` it becomes `@Initialized B`. 
This would allow the invocation of `bar` above without making the receiver type of bar `@UnknownInitialization(B.class)`.

If a final class is changed to a non-final class, invocation errors would appear in the code. I would probably still mark methods that are used in constructors with `@UnderInitialization`, but this seems like a false positive to me.

@mernst @stefanheule can you think of a reason for why this change would be unsound? Is there something I'm missing? If so, it would be good to add to the manual.

wmdietlGC:


checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -391,7 +391,9 @@ protected void setSelfTypeInInitializationCode(
             if (store != null) {
                 List<AnnotationMirror> annos = Collections.emptyList();
                 if (getUninitializedInvariantFields(store, path, false, annos).size() == 0) {
-                    if (useFbc) {
+                    if (classType.isFinal()) {
```
smillst: The comment on line 388 is now out of date.  Could you please update it?

wmdietlGC:
Thanks for the comments. I've updated the documentation.

PR #1135, opened 2017-03-13
Adds SameLen check to UBC common assignment check

kelloggm:
Fixes kelloggm#124's simple form. I opened kelloggm#127 to describe the more complex version.

### DIFF
--- a/checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
+++ b/checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
@@ -789,6 +789,30 @@ public abstract class UBQualifier {
             return false;
         }
 
+        /**
+         * For checking whether a sameLen replacement is legitimate. Checks whether replacing array
+         * with replacementArray in this qualifier creates replacementArray's entry in other.
+         */
+        public boolean isValidReplacement(
+                String array, String replacementArray, LessThanLengthOf other) {
+            Set<OffsetEquation> offsets = map.get(array);
+            if (offsets == null) {
+                return false;
+            }
+            Set<OffsetEquation> otherOffsets = other.map.get(replacementArray);
+            if (otherOffsets == null) {
+                return false;
+            }
+            for (OffsetEquation offset : offsets) {
+                for (OffsetEquation otherOffset : otherOffsets) {
+                    if (offset.equals(otherOffset)) {
+                        return true;
+                    }
+                }
+            }
+            return false;
+        }
+
         @Override
         public String toString() {
             return "LessThanLengthOf{" + "map=" + map + '}';
--- a/checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
+++ b/checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
@@ -790,8 +790,8 @@ public abstract class UBQualifier {
         }
 
         /**
-         * For checking whether a sameLen replacement is legitimate. Checks whether replacing array
-         * with replacementArray in this qualifier creates replacementArray's entry in other.
+         * Checks whether replacing array with replacementArray in this qualifier creates
+         * replacementArray's entry in other.
          */
         public boolean isValidReplacement(
                 String array, String replacementArray, LessThanLengthOf other) {
@@ -803,14 +803,13 @@ public abstract class UBQualifier {
             if (otherOffsets == null) {
                 return false;
             }
+            boolean result = offsets.size() > 0 && otherOffsets.size() > 0;
             for (OffsetEquation offset : offsets) {
                 for (OffsetEquation otherOffset : otherOffsets) {
-                    if (offset.equals(otherOffset)) {
-                        return true;
-                    }
+                    result = result && offset.equals(otherOffset);
                 }
             }
-            return false;
+            return result;
         }
 
         @Override
--- a/checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
+++ b/checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
@@ -803,13 +803,7 @@ public abstract class UBQualifier {
             if (otherOffsets == null) {
                 return false;
             }
-            boolean result = offsets.size() > 0 && otherOffsets.size() > 0;
-            for (OffsetEquation offset : offsets) {
-                for (OffsetEquation otherOffset : otherOffsets) {
-                    result = result && offset.equals(otherOffset);
-                }
-            }
-            return result;
+            return containsSame(offsets, otherOffsets);
         }
 
         @Override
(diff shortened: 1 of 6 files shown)


checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -128,32 +128,80 @@ protected void commonAssignmentCheck(
...
-                            .getMinLenAnnotatedTypeFactory()
-                            .getMinLenFromString(arrayName, valueExp, getCurrentPath());
-            boolean minLenOk = true;
-            for (Integer value : rhsValues) {
-                minLenOk =
-                        minLenOk && ltl.isValuePlusOffsetLessThanMinLen(arrayName, value, minLen);
+                            .getSameLenAnnotatedTypeFactory()
+                            .getSameLensFromString(arrayName, valueExp, getCurrentPath());
+
+            if (sameLenArrays.size() == 0) {
```
smillst: This would be easier to read as ` if (sameLenArrays.size() == 0 && !allValuesConstant) {`

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -128,32 +128,80 @@ protected void commonAssignmentCheck(
...
+                    break;
+                }
             }
-            if (!minLenOk) {
-                super.commonAssignmentCheck(varType, valueExp, errorKey);
-                return;
+
+            // SameLen checks
+            for (ExpressionTree exp : expressions) {
+                AnnotatedTypeMirror currentType = atypeFactory.getAnnotatedType(exp);
```
smillst: This recomputes the Upper Bound Type type again for each arrayName, but the type doesn't vary.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -128,32 +128,80 @@ protected void commonAssignmentCheck(
         // Either a singleton list of the single integer on the right,
         // or a list containing all the values in a constant array.
         List<Integer> rhsValues = new ArrayList<>();
-        // All the values in the must be compile-time constants.
```
smillst: This whole method has gotten very hard to understand.  Can you please try to clean up the code somewhat?  Part of the problem is that the special case for array initializers is mixed in with normal assignments.  Maybe create a new method checks whether an expressionTree is less than the length of some array expression?

kelloggm: I'll clean it up and reassign you afterwards

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
kelloggm: Maybe it's not clear, but it's computing the type of `exp`, which changes on each loop iteration.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
kelloggm: I've cleaned this up as best I can. It's still pretty bad, but I don't see any reasonable alternatives beyond what I've already done.

checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
```diff
@@ -789,6 +789,30 @@ public boolean isValuePlusOffsetLessThanMinLen(String array, int value, int minl
             return false;
         }
 
+        /**
+         * For checking whether a sameLen replacement is legitimate. Checks whether replacing array
```
smillst: The first sentence is not descriptive.   I don't know what a "legitimate" replacement is.  I would just remove this sentance.

checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
```diff
@@ -789,6 +789,30 @@ public boolean isValuePlusOffsetLessThanMinLen(String array, int value, int minl
...
+            if (offsets == null) {
+                return false;
+            }
+            Set<OffsetEquation> otherOffsets = other.map.get(replacementArray);
+            if (otherOffsets == null) {
+                return false;
+            }
+            for (OffsetEquation offset : offsets) {
+                for (OffsetEquation otherOffset : otherOffsets) {
+                    if (offset.equals(otherOffset)) {
```
smillst: I think this is incorrect.  See test case that I added.

checker/tests/index/ArrayAssignmentSameLen.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    void test3(int[] a, @LTLengthOf("#1") int i, @NonNegative int x) {
+        int[] c = a;
+        // See useTest3 for an example of why this assignment should fail.
+        @LTLengthOf(
+            value = {"c", "c"},
+            offset = {"0", "x"}
+        )
+        //:: error: (assignment.type.incompatible)
```
smillst: Currently this assignment does not issue an error.

checker/tests/index/ArrayAssignmentSameLen.java
```diff
@@ -0,0 +1,44 @@
...
+        i_index = index;
+    }
+
+    void test1(int[] a, int[] b, @LTEqLengthOf("#1") int index) {
+        int[] array = a;
+        @LTLengthOf(
+            value = {"array", "b"},
+            offset = {"0", "-3"}
+        )
+        int i = index;
```
kelloggm: Shouldn't this assignment fail? We know nothing about index's relationship with b.

smillst: Yes, I forgot to add the expected error. (Which is currently issued.)

checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
kelloggm: I think I've fixed this problem, and I'd appreciate if you could take another look.

checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
```diff
@@ -789,6 +789,29 @@ public boolean isValuePlusOffsetLessThanMinLen(String array, int value, int minl
...
+                return false;
+            }
+            Set<OffsetEquation> otherOffsets = other.map.get(replacementArray);
+            if (otherOffsets == null) {
+                return false;
+            }
+            boolean result = offsets.size() > 0 && otherOffsets.size() > 0;
+            for (OffsetEquation offset : offsets) {
+                for (OffsetEquation otherOffset : otherOffsets) {
+                    result = result && offset.equals(otherOffset);
```
smillst: If offsets = {0, "x"} and otherOffsets = {0,"x"}, this fails.  This happens in a test case like this:

````
void test4(int[] a, @LTLengthOf(value = {"#1","#1"}, offset = {"0","#3"}) int i, @NonNegative int x) {
  int[] c1 = a;
  @LTLengthOf(value = {"c1", "c1"},  offset = {"0", "x"} ) int z = i;
}
````

You can call `UBQualifier.containsSame(offsets, otherOffsets)` instead of this loop.

smillst:
@kelloggm I refactored the code and improved the documentation.  Please take a look and make sure I didn't screw anything up.

kelloggm:
It looks correct to me. Thank you for doing this; the code is much more comprehensible now.

PR #1137, opened 2017-03-13
Correct semantics for &

kelloggm:
Fixes kelloggm#122.

### DIFF
--- a/checker/src/org/checkerframework/checker/index/lowerbound/LowerBoundAnnotatedTypeFactory.java
+++ b/checker/src/org/checkerframework/checker/index/lowerbound/LowerBoundAnnotatedTypeFactory.java
@@ -772,6 +772,12 @@ public class LowerBoundAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
             return;
         }
 
+        AnnotatedTypeMirror leftType = getAnnotatedType(leftExpr);
+        if (leftType.hasAnnotation(NN) || leftType.hasAnnotation(POS)) {
+            type.addAnnotation(NN);
+            return;
+        }
+
         type.addAnnotation(UNKNOWN);
     }
 }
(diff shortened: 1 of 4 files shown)


checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -493,12 +495,36 @@ public Void visitBinary(BinaryTree tree, AnnotatedTypeMirror type) {
...
+                    || leftLBType.hasAnnotation(Positive.class)) {
+                type.addAnnotation(leftType.getAnnotationInHierarchy(UNKNOWN));
+            }
+
+            AnnotatedTypeMirror rightType = getAnnotatedType(right);
+            AnnotatedTypeMirror rightLBType =
+                    getLowerBoundAnnotatedTypeFactory().getAnnotatedType(right);
+            if (rightLBType.hasAnnotation(NonNegative.class)
+                    || rightLBType.hasAnnotation(Positive.class)) {
+                type.addAnnotation(rightType.getAnnotationInHierarchy(UNKNOWN));
```
smillst: This code doesn't account for the case when the left and right expressions are indexes for two different arrays:
````
int[] a;
int[] b;
void test(@IndexFor("a") int i, @IndexFor("b") int j) {
  a[i & j] = 1;
  b[i & j] = 2;
}
````

Comment (kelloggm) [inline checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java]: I've fixed the problem and added a test case. Thanks for pointing it out!

File: checker/src/org/checkerframework/checker/index/lowerbound/LowerBoundAnnotatedTypeFactory.java [hunk 2]
```diff
@@ -772,6 +772,12 @@ private void addAnnotationForAnd(
             return;
         }
 
+        AnnotatedTypeMirror leftType = getAnnotatedType(leftExpr);
+        if (leftType.hasAnnotation(NN) || leftType.hasAnnotation(POS)) {
```
smillst: Isn't a pos & pos a positive?
 
````
void test(@Postivie int x, @Postive int y){
    @Postive int z = x & y;
}
````

mernst: No.  1 & 2 == 0.

smillst: Oh right.  Thanks!  @kelloggm, would you please add the code in my comment as a test case?

PR #1138, opened 2017-03-13
Updates to jdk so that common classes with size() methods are annotated

kelloggm:
Fixes kelloggm#114 (hopefully).


kelloggm:
@mernst Does the list of classes I annotated seem reasonable for what you needed in kelloggm#114?

mernst:
Looks OK.  There are classes missing in un-annotated packages such as java.util.reflect and java.util.concurrent, and LinkedHashMap.java looks like an old version, but what's there looks good.

PR #1140, opened 2017-03-14
Account for @PolyMinLen in more places

smillst:
Fixes kelloggm/checker-framework#129


kelloggm:
Looks good to me. I looked through the MLC and can't find anywhere else you should change, also. Should be merged as soon as the tests pass.

PR #1143, opened 2017-03-14
Download the jdk.jars rather than building.

smillst:
Rather than rebuild the annotated jdk for each Travis build, download it from https://checkerframework.org/dev-jdk/.  (Currently the jdk.jars there have to be manually updated.) 

Also, when building jdk.jar the hash of the Checker Framework commit used is stored in the jar's manifest file.  This way we will know which version of the jdk.jar was used in the tests.  (The tests now output the manifest.)

Now that the jars aren't built for each separate build, then junit and non-junit tests can be merged.  Since building the jar is itself a test, a new Travis build that builds just the jar.jdk (for each Java version) has been added.

All of this makes the Travis tests faster and should end the frequent timeout failures.

### DIFF
--- a/.travis-build-without-test.sh
+++ b/.travis-build-without-test.sh
@@ -34,8 +34,3 @@ echo "... done: (cd ../annotation-tools/ && ./.travis-build-without-test.sh)"
 ## Compile
 echo "running \"ant dist-downloadjdk\" for checker-framework"
 (cd checker && ant dist-downloadjdk)
-
-# The implementation version listed in the Manifest is the hash of the commit of
-# the Checker Framework that created the jdk.jar.  Show it here to help debugging.
-jar -xvf checker/jdk/jdk7.jar META-INF/MANIFEST.MF && cat META-INF/MANIFEST.MF
-
--- a/.travis-build-without-test.sh
+++ b/.travis-build-without-test.sh
@@ -7,6 +7,11 @@ export SHELLOPTS
 
 SLUGOWNER=${TRAVIS_REPO_SLUG%/*}
 
+# Optional argument $2 is one of:
+#   jdk7, jdk8
+# If omited, then either jdk may be used.
+JDKVER=$1
+
 ## Build annotation-tools (Annotation File Utilities)
 if [ -d ../annotation-tools ] ; then
     # Older versions of git don't support the -C command-line option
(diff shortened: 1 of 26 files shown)


.travis-build-without-test.sh
```diff
@@ -7,6 +7,11 @@ export SHELLOPTS
 
 SLUGOWNER=${TRAVIS_REPO_SLUG%/*}
 
+# Optional argument $2 is one of:
```
mernst: This comment refers to $2, but the commented code uses $1 (on line 13).

.travis-build-without-test.sh
```diff
@@ -32,5 +37,20 @@ echo "Running:  (cd ../annotation-tools/ && ./.travis-build-without-test.sh)"
 echo "... done: (cd ../annotation-tools/ && ./.travis-build-without-test.sh)"
 
 ## Compile
-echo "running \"ant dist\" for checker-framework"
-ant dist
+#TODO: The docker images should be updated to have wget installed.
```
mernst: This will be complete in about 1.5 hours, so the comment and the two lines of code can be removed.

.travis-build.sh
```diff
@@ -9,11 +9,14 @@ if [[ "${GROUP}" == "" ]]; then
   export GROUP=all
 fi
 
-if [[ "${GROUP}" != "all" && "${GROUP}" != "junit" && "${GROUP}" != "nonjunit" && "${GROUP}" != "downstream" && "${GROUP}" != "misc" ]]; then
+if [[ "${GROUP}" != "all" && "${GROUP}" != "all-tests" && "${GROUP}" != "jdk.jar" && "${GROUP}" != "junit" && "${GROUP}" != "nonjunit" && "${GROUP}" != "downstream" && "${GROUP}" != "misc" ]]; then
   echo "Bad argument '${GROUP}'; should be omitted or one of: all, junit, nonjunit, downstream, misc."
```
mernst: The line above was updated but this line was not.

.travis-build-without-test.sh
```diff
@@ -7,6 +7,11 @@ export SHELLOPTS
 
 SLUGOWNER=${TRAVIS_REPO_SLUG%/*}
 
+# Optional argument $2 is one of:
+#   jdk7, jdk8
+# If omited, then either jdk may be used.
```
mernst: I'm not sure what "may be used" means.  Is this a desirable amount of flexibility for the script, and is it useful for people?  Would it be better to set a default?

.travis-build.sh
```diff
@@ -30,12 +33,16 @@ set -o xtrace
...
+# The above command downloads the JDK, so there is no need for a subsequent
+# command to build it except to test building it.
 
 set -e
 
+if [[ "${GROUP}" == "all-tests" || "${GROUP}" == "all" ]]; then
+  (cd checker && ant all-tests-nobuildjdk)
+fi
+
 if [[ "${GROUP}" == "junit" || "${GROUP}" == "all" ]]; then
```
mernst: Is it still useful to have the junit and nonjunit tests?  Before this pull request, all the permitted targets were disjoint, but not any longer and it might be good to document that (or remove the subtasks to avoid confusion over why they aren't being called in the test matrix).

checker/jdk/Makefile
```diff
@@ -18,7 +18,9 @@ all: clean classes
 # jdk.jar only contains those classes with annotations, not every class from
 # the original JDK.
 jdk.jar:
-	cd annotated; jar cf ../jdk.jar .
+	echo Implementation-Version: `git rev-parse --short HEAD` "\n" > manifest.txt
```
mernst: I would make this "Checker-Framework-Version" or "Checker-Framework-Commit" to be more explicit.

mernst:
This is a change to the Travis configuration files.
I had anticipated a change to the Ant buildfile.  That would make building the Checker Framework faster for everyone.  It would also solve the problems with long build times for the daikon-typecheck jobs.

smillst:
@mernst 
>This is a change to the Travis configuration files.
I had anticipated a change to the Ant buildfile. That would make building the Checker Framework faster for everyone. It would also solve the problems with long build times for the daikon-typecheck jobs.

The other Travis builds use the `.travis-build-without-test.sh` script to build the Checker Framework, so they will benefit from this change as is.  Users can manually download the jars and then run "ant dist-nobuildjdk" in the checker directory.  But I agree a new "dist-downloadjdk" target would be nice to have.  I'll add that and change the Travis script to use it. (And of course address your comments.)

smillst:
I've added an `dist-downloadjdk` target.

Issue #1142, opened 2017-03-14
 Nullness checker handling of ConcurrentHashMap


kanha95:
This is a bug.The root cause is that ConcurrentHashMap does not appear in the annotated JDK for the Nullness Checker.

The solution is to annotate file ConcurrentHashMap.java from the Java 8 JDK, and add it to checker/jdk/nullness/src/java/util/concurrent/ in the Checker Framework repository.

mernst:
This bug report is fragmentary and incomplete.  Please add a test case and any other information that is relevant.  Thanks!

kanha95:
Example 1: 

void foo() {

Map<Integer,@Nullable Integer> myMap =new ConcurrentHashMap<>();

myMap.put(1, 1);

myMap.put(2, 2);

myMap.put(3, 3);

Iterator<Integer> it1 = myMap.keySet().iterator();

while(it1.hasNext()) {

Integer key = it1.next();

System.out.println(key+" Map Value:"+myMap.get(key));

if(key.equals(1)) {

myMap.put(3, null);

}

}

}

Output: Nullness Checker passed!

N.B:

As per java docs ,

https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentHashMap.html

The concurrent-hashmap raises NullpointerException if any of the key or value is null. So here the Nullness Checker fails in catching the NullPointerException generated at the runtime. If bymistake any programmer annotes the value parameter to be @Nullable then the program will crash at the runtime.

Example 2:

void foo() {

Map<@Nullable Integer,Integer> myMap =new ConcurrentHashMap<>();

myMap.put(1, 1);

myMap.put(2, 2);

myMap.put(3, 3);

for(Map.Entry<@Nullable Integer,Integer> i:myMap.entrySet()){

System.out.println(i.getKey()+" "+i.getValue());

if(i.getValue()==2){

myMap.put(null, 4);

}

}

}

Output: Nullness Checker Passed!

N.B.

Here also the same problem arises. It is not able to catch the NullPointerException generated during runtime because of the above mentioned reasons.

rajatjain1997:
Can I work on this bug if @kanha95 is not assigned to it? I'd love to try annotating the ConcurrentQueue class for a start.

mernst:
Sure.  This is a nice, easy task to start with.  Please check whether there are other classes that also prohibit null and should be annotated at the same time.

wmdietl:
[...]

Issue #1144, opened 2017-03-15
Nullness checker gives a false negative for NULL values in some implementations of Queues


rajatjain1997:
The current implementation of the checker verifies the code below to be error free, i.e. NullPointerExceptions should not exist at runtime.

        public class NullnessCheck {	
	        public static void main(String[] args) {
		        ConcurrentLinkedQueue<@Nullable Integer> clq = new ConcurrentLinkedQueue<>();
		        clq.add(null);
            }
        }

However, when compiled, the code produces following errors -
'
        Exception in thread "main" java.lang.NullPointerException
        at java.util.concurrent.ConcurrentLinkedQueue.checkNotNull(ConcurrentLinkedQueue.java:914)
	        at java.util.concurrent.ConcurrentLinkedQueue.offer(ConcurrentLinkedQueue.java:327)
	        at java.util.concurrent.ConcurrentLinkedQueue.add(ConcurrentLinkedQueue.java:297)
	        at Test.NullnessCheck.main(NullnessCheck.java:12)
'

The same also happens with ArrayDeque and all other queues that don't allow null objects.

This happens because the current implementation of the queue interface allows Nullable objects-
`public interface Queue<E extends @Nullable Object> extends Collection<E>`

However every implementation of the interface, like the one above does not allow @Nullable objects. This is an issue throughout the collections framework including maps and lists where the implementing classes can restrict @Nullable objects while the superclasses won't. These classes need to be implemented separately, to make the checker complete. The checker should give a warning when @Nullable annotations are applied to these types.

wmdietl:
Isn't this just a duplicate of https://github.com/typetools/checker-framework/issues/1142 and the solution is to add non-null bounds for these concurrent collections?

rajatjain1997:
@wmdietl I'm sorry, I read the topics of the issues and since I'm not familiar with ConcurrentHashMap, I though that the author is trying to state a missing annotated file (And I particularly did not pay attention because my bug was about collections in general and stemmed from queues), and thus did not realize how similar they were. Only after posting did I realize it. Can you please merge these? I'm sorry if I caused any inconvenience.

mernst:
Duplicate of #1142.  This does give another important case that #1142 should handle.  Thanks for the report.

PR #1145, opened 2017-03-16
added annotations to ConcurrentHashMap.java from java 8 JDK

kanha95:
I have modified ConcurrentHashMap.java with all necessary annotations which will solve the issue "Nullness checker handling of ConcurrentHashMap" #1142


mernst:
As part of this pull request, could you also commit a test case that would fail without the annotated library class but passes with the annotated library class?

mernst:
What you have added is not a test.  For example, there is no way to run it, so no way for it to fail or succeed.
You need to add a file somewhere under `checker/tests/`, probably in `checker/tests/nullness/`.

kanha95:
why i got travis CI build failed mike?

mernst:
Can you say what you have done so far to investigate the problem?

checker/jdk/nullness/src/java/util/concurrent/ConcurrentHashMap.java
```diff
@@ -0,0 +1,6313 @@
...
+
+    /**
+     * Key-value entry.  This class is never exported out as a
+     * user-mutable Map.Entry (i.e., one supporting setValue; see
+     * MapEntry below), but can be used for read-only traversals used
+     * in bulk tasks.  Subclasses of Node with a negative hash field
+     * are special, and contain null keys and values (but are never
+     * exported).  Otherwise, keys and vals are never null.
+     */
+    static class Node<K extends @Nullable Object,V extends @Nullable Object> implements Map.Entry<K,V> {
```
wmdietl: You should use `@NonNull` as the upper bound, to prevent instantiating with anything that is nullable. Same for the enclosing class.

checker/tests/nullness/ConcurrentHashMapTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+myMap.put(3, null);
+
+}
+}
+
+
+
+/*
+Output without annotated ConcurrentHashMap.java: Nullness Checker passed!
```
wmdietl: Properly format your example. Add expected error messages, instead of adding this as a comment.
Make sure the test fails without your change and then passes with your change.

checker/jdk/nullness/src/java/util/concurrent/ConcurrentHashMap.java
kanha95: ok no problem. which methods of java 8 are causing the error ? can you help with that ?

kanha95:
public class ConcurrentHashMap<K,V> extends AbstractMap<K,V>
    implements ConcurrentMap<K,V>, Serializable

This above line implements ConcurrentMap interface. I can't find the ConcurrentMap interface inside the checker-framework. I think this also needs to be added up.

checker/jdk/nullness/src/java/util/concurrent/ConcurrentHashMap.java
wmdietlGC: Do you see the error logs? https://travis-ci.org/typetools/checker-framework/jobs/211887563

One issue is simply that test class names need to be unique per directory. It looks like 'Test' is already used. So pick a different name for your test, ideally based on the issue number.

Java 7-specific errors can be seen here: https://travis-ci.org/typetools/checker-framework/jobs/211887565 Package java.util.function doesn't exist on Java 7.

If your test case uses Java 8 features, add `// @below-java8-jdk-skip-test` somewhere in the beginning, as e.g. seen here: https://github.com/typetools/checker-framework/blob/master/checker/tests/nullness/java8/Issue1000.java

wmdietlGC:
Yes, might need to add other classes as well. As pointed out in https://github.com/typetools/checker-framework/issues/1144 there are also concurrent collections other than maps that should be added.

mernst:
The commit message should include "fixes #[ISSUENUMBER]", to make a connection between the pull request and the issue tracker.

kanha95:
done with all tests except the last one. it points out some "The ' characters around the executable and arguments are not part of the command". I googled it out and found that it's just a warning message .it does not affect the normal flow of the progam. Any suggestion? @mernst @wmdietlGC @wmdietl

wmdietlGC:
`  [apply] Improper formatting: /checker-framework/checker/tests/nullness/ConcurrentHashMapTest.java`

You need to properly format the source code.
You should setup your build environment to prevent yourself from even committing code like this.
Try running `ant reformat`.

kanha95:
@wmdietlGC  i am formatting using netbeans default formatter. Still it's not getting accepted. I am new to this. so kindly suggest.

checker/tests/nullness/ConcurrentHashMapTest.java
```diff
@@ -0,0 +1,16 @@
+
+import java.util.concurrent.ConcurrentHashMap;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+class Issue1142 {
```
mernst: In Java, the class name and the file name should match.  Please change one of them.

checker/tests/nullness/ConcurrentHashMapTest.java
```diff
@@ -0,0 +1,16 @@
+
```
mernst: By convention, when a test is about an issue in the issue tracker, we write a comment at the top of the file.  See other test files in the test suite.

checker/jdk/nullness/src/java/util/ConcurrentMap.java
```diff
@@ -0,0 +1,173 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
+package java.util;
```
mernst: This is not the correct package for `ConcurrentMap`.  Why did you change it?
I have the same question about `ConcurrentHashMap`.

mernst:
Werner said,
> Try running `ant reformat`.

He gave a very concrete suggestion.  You disregarded his suggestion and said:

> i am formatting using netbeans default formatter. Still it's not getting accepted. I am new to this. so kindly suggest.

I don't know what else to suggest.  When I tried `ant reformat` on your branch, it worked.

kanha95:
@wmdietlGC  can you please help me in using the ant reformat command ? how to setup the build environment for this? i have never done this type of formatting before. so your help will be valuable to me. i read the manual and but got confused.
Thanks in advance.

mernst:
From the command line, type `ant reformat` and then press the ENTER key.

kanha95:
@mernst  got it...thanks...i messed up with few files. so i am closing this pull request. i will be up with a fresh new PR.

PR #1146, opened 2017-03-16
Index Checker: subterm offsets

kelloggm:
fixes kelloggm#123

### DIFF
--- a/checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
+++ b/checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
@@ -20,7 +20,6 @@ import org.checkerframework.dataflow.analysis.TransferInput;
 import org.checkerframework.dataflow.analysis.TransferResult;
 import org.checkerframework.dataflow.cfg.node.ArrayCreationNode;
 import org.checkerframework.dataflow.cfg.node.AssignmentNode;
-import org.checkerframework.dataflow.cfg.node.BinaryOperationNode;
 import org.checkerframework.dataflow.cfg.node.FieldAccessNode;
 import org.checkerframework.dataflow.cfg.node.Node;
 import org.checkerframework.dataflow.cfg.node.NumericalAdditionNode;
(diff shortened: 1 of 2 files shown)


checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
```diff
@@ -204,6 +205,35 @@ private UBQualifier accountForLowerBoundAnnos(Node node, UBQualifier qual, Strin
...
+    }
+
+    private void propagateToSubterms(
+            BinaryOperationNode node, UBQualifier qualifier, CFStore store) {
+        boolean plusNode = node instanceof NumericalAdditionNode;
+        // Otherwise this should be a numerical subtraction node.
+
+        propagateToSubterm(
+                node.getRightOperand(), node.getLeftOperand(), qualifier, store, plusNode);
+        propagateToSubterm(
```
smillst: This code isn't right for subtraction.  For example, ` 1 - i < a.length`.  This code types `i` as `@LTL("a", offset="1")`, but that's not correct.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
```diff
@@ -223,6 +253,10 @@ protected void refineGT(
             knownToBeLessThanLengthOf(array, right, store, in);
         }
 
+        if (right instanceof NumericalAdditionNode || right instanceof NumericalSubtractionNode) {
+            propagateToSubterms((BinaryOperationNode) right, refinedRight, store);
```
smillst: Instead of creating a new method, I think the `knownToBeLessThanLengthOf` methods could be generalized to take a qualifier.  (And perhaps renamed to `propagateToSubterms` or maybe `propagateToOperands`).

checker/tests/index/ParserOffsetTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public void addition3(String[] a, @IndexFor("#1") int i) {
+        if ((i + 5) < a.length) {
+            @IndexFor("a") int j = i + 5;
+        }
+    }
+
+    @SuppressWarnings("lowerbound")
+    public void general_test(String[] a, @NonNegative int k) {
+        if (k - 5 < a.length) {
```
smillst: Please add test cases where the literal is the left operand (`1 - k`).

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
kelloggm: I tried this out and that's not the case. Consider the following code:

```java
  public void subtraction4(String[] a, @IndexFor("#1") int i) {
        if (1 - i < a.length) {
	    @LTLengthOf(value="a", offset="1") int k = i;
        }
    }
```

The expected warning is issued:
```java
ParserOffsetTest.java:51: error: [assignment.type.incompatible] incompatible types in assignment.
	    @LTLengthOf(value="a", offset="1") int k = i;
	                                               ^
  found   : @LTLengthOf(value="a", offset={}) int
  required: @LTLengthOf(value="a", offset="1") int
```

kelloggm: I dug a little deeper and the result of this code:

```java
    public void subtraction5(String[] a, int i) {
	if (1 - i < a.length) {
	    @IndexFor("a") int j = i;
	}
    }
```
is this warning:

```
ParserOffsetTest.java:58: error: [assignment.type.incompatible] incompatible types in assignment.
	    @IndexFor("a") int j = i;
	                           ^
  found   : @LTEqLengthOf("a") int
  required: @LTLengthOf(value="a", offset={}) int
```
which is still wrong: `i` is definitely not `@LTEqLengthOf("a")`

kelloggm: I'll see if I can figure out a way to avoid these problems.

kelloggm: I've updated the code so that it doesn't incorrectly type `i` after `if (1 - i < a.length) {`. But, this comes at a cost; in that block `1 - i` now also doesn't have a type, so there's a false positive.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
kelloggm: I'm not sure that's the right thing to do. While the methods are similar, `knownToBeLessThanLengthOf` has some strong and different assumptions than `propagateToSubterms`. Maybe there is some way to abstract out some of the internal logic; I'll look into that.

kelloggm: I looked through the code with an eye for reducing duplicated code, and I think that the code as-is is fine. In particular, all the methods that `knownToBeLessThanLengthOf` calls rely on the property that it is, in fact, known to be less than length of. For instance, the version of `knownToBeArrayLength` that takes a `NumericalAdditionNode` is able to account for lowerbound annotations because of that property.

kelloggm: So I don't think it makes sense for them to share code, since it would require the code to become quite complicated with special cases.

smillst:
@kelloggm  is this ready for me to review?

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
kelloggm: @smillst I was hoping you would respond to this comment chain

smillst: Sorry for the delay!  I'll take another look at the methods.  I have an idea of how to combine them, but it will be faster for me to just try it then to explain it.

smillst:
@kelloggm I combined the methods. Could you please review my changes?

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
```diff
@@ -72,158 +72,149 @@ public UpperBoundTransfer(CFAnalysis analysis) {
...
+            propagateToMultiplicationOperand(left, right, in, store, typeOfNode);
+            propagateToMultiplicationOperand(right, left, in, store, typeOfNode);
         }
     }
 
     /**
-     * {@code other} times {@code node} is known to be exactly the length of {@code arrayExp}.
+     * {@code other} times {@code node} is known to be {@code typeOfMultiplication}.
      *
      * <p>This implies that if {@code other} is positive, then {@code node} is less than or equal to
```
kelloggm: This comment is now out of date

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
```diff
@@ -72,158 +72,149 @@ public UpperBoundTransfer(CFAnalysis analysis) {
...
     /**
-     * The subtraction node, {@code node}, is known to be exactly the length of the array referenced
-     * by the arrayExp.
+     * The subtraction node, {@code node}, is known to be {@code typeOfSubtraction}.
      *
      * <p>This means that the left node is less than or equal to the length of the array when the
      * right node is subtracted from the left node.
      *
+     * @param typeOfSubtraction type of node
      * @param node subtraction node that is known to be equal to the length of the array referenced
```
kelloggm: Also out-of-date, since it still references `arrayExp`

kelloggm:
@smillst I took a look and it looks good. There were a few comments that are out-of-date that I marked but otherwise I think it's good to go.

PR #1153, opened 2017-03-19
Use null to represent TOP/UNKNOWN, and lists to represent all other sets of values

mernst:
BOTTOM is represented naturally as an empty list.

### DIFF
--- a/framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
+++ b/framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
@@ -223,8 +223,8 @@ public class ValueAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
 
         /**
          * If any constant-value annotation has &gt; MAX_VALUES number of values provided, treats
-         * the value as UnknownVal. Works together with ValueVisitor.visitAnnotation, which issues a
-         * warning to the user in this case.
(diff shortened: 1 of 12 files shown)


framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -947,18 +965,47 @@ private AnnotationMirror createNumberAnnotationMirror(List<Number> values) {
                 "ValueAnnotatedTypeFactory: unexpected class: " + first.getClass());
     }
 
+    /**
+     * Returns the set of possible values. Returns the empty list if no values are possible (for
+     * dead code). Returns null if any value is possible -- that is, if no estimate can be made --
+     * and this includes when there is no constant-value annotation so the argument is null.
+     */
     public static List<Long> getIntValues(AnnotationMirror intAnno) {
```
wmdietl: Why is the method called `getIntValues` if it returns a list of Long? Can you add a sentence to the javadoc to motivate this?
It would also be nice to reference the set of AnnotationMirrors that are supported.

mernst: Because it supports the `@IntVal` annotation.
I've added documentation of this, and I have documented all the parameters.

PR #1154, opened 2017-03-19
Fixes Issue #1142

kanha95:
ConcurrentHashMap.java & ConcurrentMap.java added to java.util.concurrent folder.
Issue1142.java added to checker/tests/nullness to support the above changes.


kanha95:
@mernst  
Hello Professor,
Reviews please. Kindly suggest if any further modifications is essential.

Thanking you

checker/tests/nullness/Issue1142.java
```diff
@@ -0,0 +1,14 @@
+// Issue 1142 https://github.com/typetools/checker-framework/issues/67
```
wmdietl: Wrong issue number in link.

checker/tests/nullness/Issue1142.java
```diff
@@ -0,0 +1,14 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+class Issue1142 {
+
+    void foo() {
+
+        ConcurrentHashMap<Integer, @Nullable Integer> chm = new ConcurrentHashMap<>();
```
wmdietl: The point of Issue 1142 was that you don't want to allow nullable keys or values. So this line should give an error.

checker/tests/nullness/Issue1142.java
```diff
@@ -0,0 +1,14 @@
...
+        chm.put(1, 1);
+        chm.put(3, 3);
+        chm.put(4, null);
```
wmdietl: This line also should give an error, as adding null will always raise a NPE.

checker/jdk/nullness/src/java/util/concurrent/ConcurrentMap.java
```diff
@@ -0,0 +1,173 @@
+/*
```
wmdietl: This additional header is not necessary.

checker/jdk/nullness/src/java/util/concurrent/ConcurrentHashMap.java
```diff
@@ -0,0 +1,42 @@
+package java.util.concurrent;
```
wmdietl: Why do you use two very different styles for the two classes? For this first one you just have the stubs and for the other one the complete source code.

checker/tests/nullness/Issue1142.java
kanha95: Yes Professor,
Currently this line is throwing an error, stating:
found: @Initialized @Nullable
required: @Initialized @NonNull

checker/tests/nullness/Issue1142.java
kanha95: Yes I agree,
But it is unnecessary to throw error here because at line 9 there is already an error. The programmer needs to solve it first to reach to line 12.
If someone writes : 
ConcurrentHashMap<Integer, Integer> chm = new ConcurrentHashMap<>();
Then in this case line 12 will throw an error stating:
found: null
required: @Initialized @NonNull
This way it raises lesser number of warnings making easier for the programmer to debug.

checker/tests/nullness/Issue1142.java
mernst: The test should pass exactly if the error is thrown.  Right now, you have written a test that passes only if the error is not thrown.

Please read the instructions for writing a test case, which appear in file `checker/tests/README`.

checker/jdk/nullness/src/java/util/concurrent/ConcurrentHashMap.java
kanha95: concurrent/ConcurrentMap.java is nothing but just a marker interface. It has lesser significance here.

checker/tests/nullness/Issue1142.java
kanha95: @mernst  So, i need to insert the "//:: error:" line wherever an error is expected to occur. Am i right?

checker/tests/nullness/Issue1142.java
kanha95: will fix this.

checker/tests/nullness/Issue1142.java
mernst: Yes.  But, you don't have to ask -- just try it, according to the instructions.

mernst:
A comment about commit messages:
One commit -- usually the very first one -- should include "fixes #1142", not as the full commit message but as part of it.  After that, please use a descriptive commit message for each commit.

kanha95:
@mernst  sure i will. 
I need a bit help. I tried a lot of possibilities to include an expected error message in the comment line. But all fails in the travis build. If i don't include the error line in the comment then it gets accepted. I went through the manual , javadocs & finally looked for examples in the tests section. But still didn't find any valuable answers.

mernst:
You mentioned that the Travis build is failing.
Are the tests passing locally?

kanha95:
@mernst  Yes professor. I have checked with netbeans and results are as follows:

/******

C:\Users\Rutu\Documents\NetBeansProjects\projectChecker\src\testClass.java:9: error: [type.argument.type.incompatible] incompatible types in type argument.
        ConcurrentHashMap<Integer, @Nullable Integer> chm1 = new ConcurrentHashMap<>();
  found   : @Initialized @Nullable Integer
C:\Users\Rutu\Documents\NetBeansProjects\projectChecker\src\testClass.java:12: error: [argument.type.incompatible] incompatible types in argument.
        chm1.put(1, null);
  found   : null
  required: @Initialized @NonNull Integer
2 errors
1 warning

******/

I have specified the same expected error in the comment lines.  But the travis build is failing. Kindly suggest and help.

mernst:
When I asked if the tests are passing locally, I meant the automated tests.
When you run the Checker Framework tests locally, what is the result?

It's fine to load a class into an IDE for debugging, but that's not what I was asking about in this case.

mernst:
The tests are passing locally for me.  This means there is something different with the Travis build.  I think I have identified it and the JDK 8 tests are now passing.  I'll work on the JDK 7 tests tomorrow.

checker/jdk/nullness/src/java/util/concurrent/ConcurrentHashMap.java
wmdietl: It is not a marker interface. You should use the same style for both classes.

checker/jdk/nullness/src/java/util/concurrent/ConcurrentHashMap.java
```diff
@@ -0,0 +1,42 @@
+package java.util.concurrent;
+import org.checkerframework.dataflow.qual.Pure;
+import org.checkerframework.dataflow.qual.SideEffectFree;
+import java.util.*;
+import org.checkerframework.checker.nullness.qual.KeyFor;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public class ConcurrentHashMap<@NonNull K extends @NonNull Object, @NonNull V extends @NonNull Object> extends AbstractMap<K, V>
```
wmdietl: You don't need the `@NonNull` before the type parameter names; on the upper bounds is enough.

checker/jdk/nullness/src/java/util/concurrent/ConcurrentMap.java
```diff
@@ -0,0 +1,169 @@
...
+ * <p>This interface is a member of the
+ * <a href="{@docRoot}/../technotes/guides/collections/index.html">
+ * Java Collections Framework</a>.
+ *
+ * @since 1.5
+ * @author Doug Lea
+ * @param <K> the type of keys maintained by this map
+ * @param <V> the type of mapped values
+ */
+public interface ConcurrentMap<K, V> extends Map<K, V> {
```
wmdietl: Like for ConcurrentHashMap, you should use `@NonNull` as the upper bounds, to prevent instantiation will `@Nullable`.

kanha95:
@mernst  Thanks a lot professor. I was striving hard to find the bug in the test case. I was also trying the checks with both jdk7 & jdk8. I even tried the code with Checker Framework Online Live Demo tool. It seems to work fine there as per the required results without any changes.

mernst:
Your changes should not appear in the Live Demo tool; I'm surprised the desired warnings were issued in that case.  I'll get back to you when the Travis JDK 7 build is working.

kanha95:
@mernst  Yes i meant without the files being added (ConcurrentHashMap.java & ConcurrentMap.java) there were no warnings in the online demo tool. And locally i checked for the warnings using jdk7 & jdk8.

kanha95:
@mernst  Hello professor,
Is the travis build ready for jdk7 now?

kanha95:
@mernst  Do i need to re-push the changes ?

checker/jdk/nullness/src/java/util/concurrent/ConcurrentHashMap.java
```diff
@@ -0,0 +1,42 @@
+package java.util.concurrent;
+import org.checkerframework.dataflow.qual.Pure;
+import org.checkerframework.dataflow.qual.SideEffectFree;
+import java.util.*;
+import org.checkerframework.checker.nullness.qual.KeyFor;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public class ConcurrentHashMap<K extends @NonNull Object, V extends @NonNull Object> extends AbstractMap<K, V>
```
wmdietlGC: All the same specifications should also appear in the supertype, `ConcurrentMap`. At the moment there are many more annotations here than in `ConcurrentMap`.

mernst:
@kanha95 Once you push your changes, the pull request is automatically updated.

kanha95:
@mernst  There was a problem with jdk7 travis build test. So you asked me to wait. So i was asking if i need to push again so that the Travis Build runs again.

checker/jdk/nullness/src/java/util/concurrent/ConcurrentHashMap.java
```diff
@@ -0,0 +1,42 @@
...
+     public ConcurrentHashMap(int initialCapacity, float loadFactor) {throw new RuntimeException("skeleton method");}
+     public ConcurrentHashMap(int initialCapacity) {throw new RuntimeException("skeleton method");}
+     public ConcurrentHashMap() {throw new RuntimeException("skeleton method");}
+     @Pure public boolean isEmpty() {throw new RuntimeException("skeleton method");}
+     @Pure public int size() {throw new RuntimeException("skeleton method");}
+     @Pure public @Nullable V get(@NonNull Object key){throw new RuntimeException("skeleton method");}
+     @Pure public boolean containsKey(@NonNull Object key) {throw new RuntimeException("skeleton method");}
+     @Pure public boolean containsValue(@NonNull Object value) {throw new RuntimeException("skeleton method");}
+     @Pure public boolean contains(@NonNull Object value) {throw new RuntimeException("skeleton method");}
+     public @Nullable V put(@NonNull K key,@NonNull V value) {throw new RuntimeException("skeleton method");}
```
wmdietlGC: You don't need the `@NonNull` on the K and V parameters, as their upper bounds are already non-null. This applies to all the other uses on K or V as well.

checker/jdk/nullness/src/java/util/concurrent/ConcurrentMap.java
```diff
@@ -0,0 +1,24 @@
+
```
wmdietlGC: Why have so much whitespace in this file?

checker/jdk/nullness/src/java/util/concurrent/ConcurrentHashMap.java
```diff
@@ -0,0 +1,42 @@
...
+     public @Nullable V replace(@NonNull K key, @NonNull V value) {throw new RuntimeException("skeleton method");}
+     public void clear() {throw new RuntimeException("skeleton method");}
+     @SideEffectFree public Set<@KeyFor("this") K> keySet() {throw new RuntimeException("skeleton method");}
+     @SideEffectFree public Collection<V> values() {throw new RuntimeException("skeleton method");}
+     @SideEffectFree public Set<Map.Entry<@KeyFor("this") K,V>> entrySet() {throw new RuntimeException("skeleton method");}
+     @SideEffectFree public Enumeration<K> keys() {throw new RuntimeException("skeleton method");}
+     @SideEffectFree public Enumeration<V> elements() {throw new RuntimeException("skeleton method");}
+     @SideEffectFree public Object clone() { throw new RuntimeException("skeleton method"); }
+
+	 }
```
wmdietlGC: Properly format the file.

checker/jdk/nullness/src/java/util/concurrent/ConcurrentHashMap.java
```diff
@@ -0,0 +1,42 @@
...
+import org.checkerframework.checker.nullness.qual.KeyFor;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public class ConcurrentHashMap<K extends @NonNull Object, V extends @NonNull Object> extends AbstractMap<K, V>
+        implements ConcurrentMap<K, V>, java.io.Serializable {
+
+    private static final long serialVersionUID = 0;
+
+     public ConcurrentHashMap(int initialCapacity,float loadFactor, int concurrencyLevel) {throw new RuntimeException("skeleton method");}
```
wmdietlGC: Could you consistently use `...) { throw new RuntimeException("skeleton method"); }`?

checker/jdk/nullness/src/java/util/concurrent/ConcurrentHashMap.java
kanha95: ok sure. i will change it

checker/jdk/nullness/src/java/util/concurrent/ConcurrentHashMap.java
kanha95: Yes at first i kept so. But that way it was not able to produce the 2 errors simultaneously. It was just generating the 1st error that it encountered ignoring the below lines. Would that be ok?

wmdietlGC: Are you sure you had the upper bounds when you tried that? Adding `@NonNull` to K or V shouldn't have any effect. Separately from this pull request, can you try writing a small test case?

kanha95: Yes it didn't work that way.

Consider this example:

import java.util.concurrent.ConcurrentHashMap;

class Issue1142 {

    void foo() {
        //:: error: (type.argument.type.incompatible)
        ConcurrentHashMap<Integer, @Nullable Integer> chm1 = new ConcurrentHashMap<>();

        //:: error: (argument.type.incompatible)
        chm1.put(1, null);
    }
}

It just produced the 1st error ignoring the 2nd one completely. Then after specifying @NonNull annotation before the K & V key it produced both the errors simultaneously. So i opted to use it this way.

kanha95: I wrote another simpler example to demonstrate this.

class A <T extends @NonNull Object> {
    
    void check (T var){
     System.out.println(var);   
    }
    
    public static void main(String[] args){
        
        A<@Nullable String> a=new A<>();
        a.check(null);
        
    }
    
}

Error: [type.argument.type.incompatible] incompatible types in type argument.
  found : @Initialized @Nullable String
line 10

wmdietlGC: This is a self-contained example:

```

class AnnoKV<T extends Object> {
  void foo(T p) {}
  void bar(@NonNull T p) {}

  void baz(AnnoKV<@Nullable String> p) {
    p.foo(null);
    p.bar(null);
  }
}
```

You get a warning both for the parameter and for the invocation of bar, but not for the invocation of foo.

This is intended behavior: you are notified that the instantiation is invalid, but then the declared type argument is used for checks. To get the error we would have to change the type from the invalid instantiation to something that respects the bound - which might not be easy to do.

Having the additional `@NonNull` on the parameters gives the extra warnings, but I don't think they are useful. They clutter the annotated JDK.

kanha95: Yes i do agree. Moreover, i think it's unnecessary to show the below warnings. @mernst  @wmdietl  Would that be fine to reproduce just the 1st error ignoring the below ones? Still i would like to know why this unusual behavior occurs?

wmdietlGC: Didn't I already explain why it is intended behavior? You check whether a type is a valid instantiation and that might raise an error. Later uses of the type will still use that incorrect instantiation, not a magically fixed instantiation.

kanha95: @wmdietlGC  Got it now. Sorry i couldn't understand at first. Should i make changes then? Removing those @NonNull statements ?

kanha95:
@mernst  Professor kindly review this work. It's pending since a long time.

mernst:
@wmdietlGC requested changes.  If you have addressed all of those, you should so inform him.

wmdietlGC:
@kanha95 Tests are failing. Once the tests pass and you adapted the formatting, I can have another look.

kanha95:
@wmdietlGC  Tests are probably failing because travis build downloaded previous results instead of processing the new ones. I have tested them locally. There doesn't seem to be an error for which the test would fail. So kindly suggest me changes if any? I would be definitely working on it,

checker/jdk/nullness/src/java/util/concurrent/ConcurrentHashMap.java
```diff
@@ -0,0 +1,41 @@
...
+import org.checkerframework.dataflow.qual.Pure;
+import org.checkerframework.dataflow.qual.SideEffectFree;
+import java.util.*;
+import org.checkerframework.checker.nullness.qual.KeyFor;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public class ConcurrentHashMap<K extends @NonNull Object, V extends @NonNull Object> extends AbstractMap<K, V>
+        implements ConcurrentMap<K, V>, java.io.Serializable 
+{
```
wmdietlGC: Don't put on a separate line.

checker/jdk/nullness/src/java/util/concurrent/ConcurrentHashMap.java
```diff
@@ -0,0 +1,41 @@
...
+
+     public ConcurrentHashMap(int initialCapacity,float loadFactor, int concurrencyLevel) {throw new RuntimeException("skeleton method");}
```
wmdietlGC: Consistently have a space after a comma.

wmdietlGC: Have spaces around the method bodies: `{ throw new RuntimeException("skeleton method"); }`

checker/jdk/nullness/src/java/util/concurrent/ConcurrentHashMap.java
```diff
@@ -0,0 +1,41 @@
...
+     public boolean replace(K key, V oldValue,V newValue) {throw new RuntimeException("skeleton method");}
+     public @Nullable V replace(K key, V value) {throw new RuntimeException("skeleton method");}
+     public void clear() {throw new RuntimeException("skeleton method");}
+     @SideEffectFree public Set<@KeyFor("this") K> keySet() {throw new RuntimeException("skeleton method");}
+     @SideEffectFree public Collection<V> values() {throw new RuntimeException("skeleton method");}
+     @SideEffectFree public Set<Map.Entry<@KeyFor("this") K,V>> entrySet() {throw new RuntimeException("skeleton method");}
+     @SideEffectFree public Enumeration<K> keys() {throw new RuntimeException("skeleton method");}
+     @SideEffectFree public Enumeration<V> elements() {throw new RuntimeException("skeleton method");}
+     @SideEffectFree public Object clone() { throw new RuntimeException("skeleton method"); }
+
```
wmdietlGC: Remove the empty lines before and after the curly.

checker/jdk/nullness/src/java/util/concurrent/ConcurrentMap.java
wmdietlGC: Same comments as for file above.

checker/jdk/nullness/src/java/util/concurrent/ConcurrentMap.java
```diff
@@ -0,0 +1,35 @@
...
+     public @Nullable V replace(K key, V value);
+     public void clear();
+     @SideEffectFree public Set<@KeyFor("this") K> keySet();
+     @SideEffectFree public Collection<V> values();
+     @SideEffectFree public Set<Map.Entry<@KeyFor("this") K,V>> entrySet();
+     @SideEffectFree public Enumeration<K> keys();
+     @SideEffectFree public Enumeration<V> elements();
+     @SideEffectFree public Object clone();
+   
+}
```
wmdietlGC: Do end the file with a newline. Note the red sign.

checker/tests/nullness/Issue1142.java
```diff
@@ -0,0 +1,14 @@
...
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+class Issue1142 {
+
+    void foo() {
+        //:: error: (type.argument.type.incompatible)
+        ConcurrentHashMap<Integer, @Nullable Integer> chm1 = new ConcurrentHashMap<>();
+
+        //:: error: (argument.type.incompatible)
+        chm1.put(1, null);
```
wmdietlGC: Why is this one failing? Won't it use `@Nullable Integer` as the value type and allow this?

kanha95: Yes correct. It shouldn't fail. I will correct it. Thanks a lot for the suggestions.

mernst:
If you pull master into your branch, then the tests should pass.

mernst:
To fix all formatting issues at once, you can just run google-java-format.  One way to do that is via the scripts in this repository: https://github.com/plume-lib/run-google-java-format .  There are other ways to install and run it too.

kanha95:
@mernst  Did i commit any mistake professor? Just the misc error in the end.

wmdietlGC:
One test needs finish running, but that one probably won't fail.
This looks good now. Thanks for all your work on this!

PR #1162, opened 2017-03-20
Rename areSameByClass to hasClass.

smillst:
And areSameByName to hasName and other related methods.

### DIFF
--- a/javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
+++ b/javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
@@ -235,19 +235,21 @@ public class AnnotationUtils {
     }
 
     /** Checks that the annotation {@code am} has the name {@code aname}. Values are ignored. */
-    public static boolean areSameByName(AnnotationMirror am, /*@Interned*/ String aname) {
+    public static boolean hasName(AnnotationMirror am, /*@Interned*/ String aname) {
         // Both strings are interned.
         return annotationName(am) == aname;
     }
 
     /** Checks that the annotation {@code am} has the name of {@code anno}. Values are ignored. */
-    public static boolean areSameByClass(AnnotationMirror am, Class<? extends Annotation> anno) {
-        /*@Interned*/ String canonicalName = annotationClassNames.get(anno);
-        if (canonicalName == null) {
+    public static boolean hasClass(AnnotationMirror am, Class<? extends Annotation> anno) {
+        /*@Interned*/ String canonicalName;
+        if (annotationClassNames.containsKey(anno)) {
+            canonicalName = annotationClassNames.get(anno);
+        } else {
             canonicalName = anno.getCanonicalName().intern();
             annotationClassNames.put(anno, canonicalName);
         }
-        return areSameByName(am, canonicalName);
+        return hasName(am, canonicalName);
     }
 
     /**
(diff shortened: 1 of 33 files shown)


mernst:
Looks good.
The method name `hasClass` is a bit odd to me, because it suggests containment.  Maybe `isOfClass` or `instanceOf` or `isInstanceOf`?  Maybe someone else has a better idea for a name.

javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -235,19 +235,21 @@ public static boolean areSameIgnoringValues(AnnotationMirror a1, AnnotationMirro
...
         return annotationName(am) == aname;
     }
 
     /** Checks that the annotation {@code am} has the name of {@code anno}. Values are ignored. */
-    public static boolean areSameByClass(AnnotationMirror am, Class<? extends Annotation> anno) {
-        /*@Interned*/ String canonicalName = annotationClassNames.get(anno);
-        if (canonicalName == null) {
+    public static boolean hasClass(AnnotationMirror am, Class<? extends Annotation> anno) {
+        /*@Interned*/ String canonicalName;
+        if (annotationClassNames.containsKey(anno)) {
```
wmdietl: This is undoing a recent change to make this code thread safe.

mernst: The method names areSameByName and areSameByClass are similar to areSameIgnoringValues, but the signature is different, and this makes client code confusing when it uses both areSameIgnoringValues and one of the other two.

The method "areSameByName" isn't comparing two things that are the same by name.
If you have better ideas for the names, let us know, but the current ones seem misleading.

smillst: @wmdietl Good catch!   I'll fix it when/if we agree on names.

smillst:
@mernst @wmdietl  Do you discuss this yesterday?

mernst:
We agreed that:
 * the old methods should be deprecated through at least one release, then renamed, to reduce disruption for users
 * we need to find better names, but we don't have great ideas yet.  We are still thinking about it.

mernst:
We think `isNamed` is a better method name than `hasName`.
I gave some suggestions for better names than `hasClass` in an earlier comment.  What do others think?

wmdietlGC:
I went through all methods in AnnotationUtils and think the following renamings would make the interface more consistent:

```
8,9c8,9
<   final String getName(AnnotationMirror);
<   String getSimpleName(AnnotationMirror);
---
>   final String annotationName(AnnotationMirror);
>   String annotationSimpleName(AnnotationMirror);

13,14c13,14
<   boolean hasName(AnnotationMirror, String);
<   boolean hasClass(AnnotationMirror, Class<? extends Annotation>);
---
>   boolean areSameByName(AnnotationMirror, String);
>   boolean areSameByClass(AnnotationMirror, Class<? extends Annotation>);

20,21c20,21
<   boolean containsWithClass(Collection<? extends AnnotationMirror>, Class<? extends Annotation>);
<   AnnotationMirror getWithClass(Collection<? extends AnnotationMirror>, Class<? extends Annotation>);
---
>   boolean containsSameByClass(Collection<? extends AnnotationMirror>, Class<? extends Annotation>);
>   AnnotationMirror getAnnotationByClass(Collection<? extends AnnotationMirror>, Class<? extends Annotation>);
```

Lines starting with `<` are the new names, lines with `>` the old names.
I think this would make the queries, getters, operations on collections, etc. more consistent. Just renaming some of these methods would introduce inconsistencies and not improving the overall API.

I would add the new methods, deprecate the old methods for one release, then remove the old methods.

(There is also a final modifier on a static method, which is useless.)

smillst:
I'm fine with the following renamings:
```
String annotationName(AnnotationMirror) => String getName(AnnotationMirror)
String annotationSimpleName(AnnotationMirror) => String getSimpleName(AnnotationMirror)
```

For the following I prefer "instanceOf" to "withClass" or "hasClass"
```
boolean containsSameByClass(Collection, Class) => boolean containsInstanceOf(Collection, Class)
AnnotationMirror getAnnotationByClass(Collection, Class) => AnnotationMirror getInstanceOf(Collection, Class)
boolean areSameByClass(AnnotationMirror, Class) => boolean isInstanceOf(AnnotationMirror, Class)
```

I don't have a preference between hasName or isNamed 
```
boolean areSameByName(AnnotationMirror, String) => boolean hasName(AnnotationMirror, String)
```

mernst:
We all agree about `getName` and `getSimpleName`.

I agree with Suzanne about `instanceOf` being superior to `withClass`.

I prefer `isNamed` to `hasName`, because `hasName` which sounds too much to me like it is extracting a field or testing for containment.

wmdietlGC:
I don't like `instanceOf` because it is confusing with the normal use of that term. No `AnnotationMirror` is an instance of a Class in the normal sense.

I like the `hasName` and `hasClass` symmetry. `isNamed` and `isClassed` would sound awkward. The AnnotationMirror has a particular name and has a particular class, so `hasName` and `hasClass` seem most compelling.

mernst:
I can live with those names -- they are better than the current ones.
(The code has changed enough that it is probably easier to make a new pull request than to try to update this one.)

smillst:
What if areSameByClass is renamed to hasClass, what should containsSameByClass be renamed to?
Also, getAnnotationByClass?  (Note, there are containsSame and containsSameIgnoringValues methods, too)

wmdietlGC:
I have that case covered in https://github.com/typetools/checker-framework/pull/1162#issuecomment-288807417 . I'm fine with 'By' instead of 'With'. Although 'With' would be more consistent with 'has'.

wmdietlGC:
And I wouldn't add 'Annotation' to the method name, as we're not using it in any of the other methods.

wmdietlGC:
What happened to this?
Should we open an issue to keep track of doing this?

PR #1163, opened 2017-03-20
Implement refinements based on binary comparison operators in value checker

kelloggm:
And also create ArrayLen annotations for arrays when their lengths have more precise IntVal annotations.

Fixes kelloggm#131

Note that the only thing that changed in `ValueAnnotatedTypeFactory#resultAnnotationHandler` is that I moved it from the `TreeAnnotator` to the ATF itself. I did that so that I could write `ValueAnnotatedTypeFactory#createAnnotationFromResults`, which wraps `resultAnnotationHandler` so that `ValueTransfer` can use it.


JasonMrX:
Previously the transfer result of a binary operation is a RegularTransferResult. In this change do you change it to ConditionalTransferResult so that we support control flow refinement? 

Thanks,

Jason

kelloggm:
@JasonMrX Yes, that's an important part of this change.

smillst:
@kelloggm FYI, I'm working on merge master into this branch.

JasonMrX:
Great. So I was having concern that other parts might be using the RegularTransferResult feature, e.g. @IntVal({1, 2}) > 0 => @BoolVal({true}). So now this feature is gone, right? Would it be a problem?

kelloggm:
No, this change preserves that behavior. That information is still added to the store in the same way it was before. I'll add some tests as part of the new, big refinement test case that demonstrate that that functionality is unchanged (though it may be tested elsewhere).

JasonMrX:
I see. Thanks for clarifying!

kelloggm:
Thanks for bringing it up! I hadn't actually checked that it was still working, so it's great that you mentioned it.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -189,6 +193,53 @@ private AnnotationMirror createBooleanAnnotationMirror(List<Boolean> values) {
         return new RegularTransferResult<>(newResultValue, result.getRegularStore());
     }
 
+    private TransferResult<CFValue, CFStore> createNewConditionalResult(
+            TransferResult<CFValue, CFStore> result, List<Boolean> resultValues, Node node) {
```
smillst: Node isn't used, so it should be removed.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -189,6 +193,53 @@ private AnnotationMirror createBooleanAnnotationMirror(List<Boolean> values) {
...
+        AnnotationMirror stringVal = createBooleanAnnotationMirror(resultValues);
+        CFValue newResultValue =
+                analysis.createSingleAnnotationValue(
+                        stringVal, result.getResultValue().getUnderlyingType());
+
+        return new ConditionalTransferResult<CFValue, CFStore>(
```
smillst: This should only be a conditional result of the TransferInput to the visitX method is a conditional result.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -189,6 +193,53 @@ private AnnotationMirror createBooleanAnnotationMirror(List<Boolean> values) {
         return new RegularTransferResult<>(newResultValue, result.getRegularStore());
     }
 
+    private TransferResult<CFValue, CFStore> createNewConditionalResult(
```
smillst: Is a new method really needed or could `createNewResultBoolean` be changed to return a conditional result (when appropriate)?

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -189,6 +193,53 @@ private AnnotationMirror createBooleanAnnotationMirror(List<Boolean> values) {
...
+                    for (Long l : lengthValues) {
+                        arrayLenValues.add(l.intValue());
+                    }
+                    AnnotationMirror newArrayAnno =
+                            ((ValueAnnotatedTypeFactory) atypefactory)
+                                    .createArrayLenAnnotation(arrayLenValues);
+                    FlowExpressions.Receiver arrayRec =
+                            FlowExpressions.internalReprOf(
+                                    analysis.getTypeFactory(), node.getReceiver());
+                    result.getRegularStore().clearValue(arrayRec);
```
smillst: You should glb with the previous value rather than clearing it.  See test case I added in framework/tests/value/Refinement2.java.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -189,6 +193,52 @@ private AnnotationMirror createBooleanAnnotationMirror(List<Boolean> values) {
...
     }
 
+    private TransferResult<CFValue, CFStore> createNewConditionalResult(
+            TransferResult<CFValue, CFStore> result, List<Boolean> resultValues, Node node) {
+        AnnotationMirror stringVal = createBooleanAnnotationMirror(resultValues);
+        CFValue newResultValue =
+                analysis.createSingleAnnotationValue(
+                        stringVal, result.getResultValue().getUnderlyingType());
+
+        return new ConditionalTransferResult<>(
```
smillst: This should only be a conditional result of the TransferInput to the visitX method is a conditional result.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -189,6 +193,52 @@ private AnnotationMirror createBooleanAnnotationMirror(List<Boolean> values) {
...
+                    List<Integer> arrayLenValues = new ArrayList<>(lengthValues.size());
+                    for (Long l : lengthValues) {
+                        arrayLenValues.add(l.intValue());
+                    }
+                    AnnotationMirror newArrayAnno =
+                            atypefactory.createArrayLenAnnotation(arrayLenValues);
+                    Receiver arrayRec =
+                            FlowExpressions.internalReprOf(
+                                    analysis.getTypeFactory(), node.getReceiver());
+                    result.getRegularStore().clearValue(arrayRec);
```
smillst: You should glb with the previous value rather than clearing it. See test case I added in framework/tests/value/Refinement2.java.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -528,56 +579,152 @@ private AnnotationMirror createBooleanAnnotationMirror(List<Boolean> values) {
...
+        // Refine the values of the operands in the store based on the result of the comparison.
+
+        List<Number> thenLeftVals = new ArrayList<>();
+        List<Number> elseLeftVals = new ArrayList<>();
+
+        // Because resultValues contains the (boolean) results of each comparison (i.e. the product of the number
+        // of elements in each list), we have to iterate through parts of the list at a time in order to find all
+        // the results of comparisons involving a particular value in the left list.
+        for (int i = 0; i < lefts.size(); i++) {
+            for (int j = i; j < (1 + i) * rights.size(); j++) {
```
smillst: It's really hard to understand why `(1 + i( * rights.size()` is correct.  Could this logic (the logic for the rightValues)  be moved into the nested for loop above?

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -528,56 +579,152 @@ private AnnotationMirror createBooleanAnnotationMirror(List<Boolean> values) {
...
+                } else {
+                    elseRightVals.add(rights.get(i));
+                }
+            }
+        }
+
+        FlowExpressions.Receiver leftRec =
+                FlowExpressions.internalReprOf(analysis.getTypeFactory(), leftNode);
+        TypeMirror leftUnderlyingType =
+                atypefactory.getAnnotatedType(leftNode.getTree()).getUnderlyingType();
```
smillst: I think you can replace this with `leftNode.getType()`.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -528,56 +579,152 @@ private AnnotationMirror createBooleanAnnotationMirror(List<Boolean> values) {
...
         return resultValues;
     }
 
+    private void createAnnotationFromResultsAndAddToStore(
+            CFStore store,
+            List<?> results,
+            TypeMirror underlyingType,
+            FlowExpressions.Receiver rec) {
+        AnnotationMirror anno = atypefactory.createResultingAnnotation(underlyingType, results);
+        store.clearValue(rec);
```
smillst: Again, don't remove, glb.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -528,56 +579,152 @@ private AnnotationMirror createBooleanAnnotationMirror(List<Boolean> values) {
...
+        store.insertValue(rec, anno);
+    }
+
     @Override
     public TransferResult<CFValue, CFStore> visitLessThan(
             LessThanNode n, TransferInput<CFValue, CFStore> p) {
         TransferResult<CFValue, CFStore> transferResult = super.visitLessThan(n, p);
+        ConditionalTransferResult<CFValue, CFStore> newResult =
```
smillst: Why are you creating a conditional result here and in the createNewConditionalResult call below?

kelloggm: To be entirely honest I don't know why I seem to need to do this. But if I thread the existing transfer result (`transferResult`) through, then the refinements stop working. I'd appreciate it if you could shed some light on why that's happening.

kelloggm:
@smillst I've addressed all of your feedback and fixed the problem you demonstrated with `Refinement2.java`. I'd appreciate it if you could take another look.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -528,56 +589,135 @@ private AnnotationMirror createBooleanAnnotationMirror(List<Boolean> values) {
...
+        AnnotationMirror anno =
+                atypefactory.createResultingAnnotation(
+                        node.getType(), results.size() == 0 ? null : results);
+        AnnotationMirror currentAnno =
+                atypefactory
+                        .getAnnotatedType(node.getTree())
+                        .getAnnotationInHierarchy(atypefactory.BOTTOMVAL);
+        Receiver rec = FlowExpressions.internalReprOf(analysis.getTypeFactory(), node);
+        store.insertValue(
+                rec, atypefactory.getQualifierHierarchy().greatestLowerBound(anno, currentAnno));
```
kelloggm: Am I correct that GLB will only return bottom if we have two inconsistent sets of values? For instance, if one side is `@IntVal({1,2})` and the other is `@IntVal({3})`? In those cases, I think we're doing the right thing: in dead code, the refined types of things should be bottom, since we don't want to issue alarms.

smillst: >Am I correct that GLB will only return bottom if we have two inconsistent sets of values?

Yes

>In those cases, I think we're doing the right thing: in dead code, the refined types of things should be bottom, since we don't want to issue alarms.

I agree, though it seems wrong to allow the following code type check.
```
@IntVal(2) int x = 2;
if (x != 2) {
   @IntVal(100) int y = x;
}
```

But I think the way to rectify the "wrongness" is to issue a warning (not an error) that the if branch is never taken.  Though this is out-of-scope for this pull request.  (I'll submit an issue.)

smillst:
@kelloggm  The tests are crashing: https://travis-ci.org/kelloggm/checker-framework/jobs/213919098
Once they pass, I'll review the code again.

framework/src/org/checkerframework/common/value/ValueTransfer.java
smillst: (I'm looking into to this.)

framework/src/org/checkerframework/common/value/ValueTransfer.java
smillst: FYI, I was wrong about this.  I've fixed it.

framework/src/org/checkerframework/common/value/ValueTransfer.java
kelloggm: So do you think that we should change the code here? As is, it fails two of the tests in `Refinement2.java` in the dead code sections because it assigns the (actually correct) type to the variable and then finds it inconsistent with the expressed annotation. I'm actually okay with dead code issuing that kind of warning, so I think we should just change `Refinement2.java` to expect those assignments to fail.

kelloggm: I think this is the only outstanding issue with this PR @smillst

smillst:
@kelloggm  There's a merge conflict now. Can you resolve it?

framework/src/org/checkerframework/common/value/ValueTransfer.java
smillst: Consider the following code:
```
@IntVal(2) int x = 2;
if (x == 3) {
    // Option 1: x is refined to @BottomVal, the assignment below doesn't issue a warning.
    // Option 2: x is refined to @IntVal(3) and the assignment below fails.
    @IntVal(100) int y = x;
    @IntVal(3) int z = x; // ok for both options
}
if (x != 2) {
    // Option 1: x is refined to @BottomVal, the assignment below doesn't issue a warning.
    // Option 2: x is not refined and remains @IntVal(2) and the assignment below fails.
   @IntVal(100) int y = x;
   @IntVal(2) int z = x; // ok for both options
}
```
@mernst What do you think?

kelloggm:
I resolved the merge conflict

framework/src/org/checkerframework/common/value/ValueTransfer.java
mernst: I like refining to `@BottomVal`.  That is more consistent, it doesn't require special-case code, and I don't mind if the type-checker skips over dead code.  (If we have places where this is confusing to people, we could revisit that decision, but it has worked well so far in other type-checkers.)
Adding an optional warning about dead code is fine with me.

smillst:
@kelloggm  I updated the test cases to not expect warnings from dead code, but warnings are still issued.  (See https://travis-ci.org/kelloggm/checker-framework/builds/214378663)

kelloggm:
Yeah, I see that. I'll take a look and fix it.

intval:
Any chance you start writing @intval without the @ ? I keep receiving notifications, about being tagged, so I looked at the issues and hoped if I solve them for you - the emails and notifications would stop. However I'm of little help here as I have no idea what's going on in the code, sorry. It's a loose loose situation, and while I understand annotations are generally starts with @, GitHub is a bastard and thinks it's a person

smillst:
@intval Sorry for tagging you!  We usually try to put annotations in code blocks `@IntVal` so that people aren't tagged.  You can [unsubscribe](https://help.github.com/articles/unsubscribing-from-conversations/) from notifications and we will avoid tagging you again.

JasonMrX:
@intval I am so sorry..but I can't stop laughing when I saw your comment! For months I've been teasing my brain to think of the reason why @intval is always marked bold while other annotations like @doubleval remains plain..

I will try to be careful in the future and put `@IntVal` in a code block..

kelloggm:
I've addressed all changes and all tests pass.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -185,13 +190,80 @@ private AnnotationMirror createNumberAnnotationMirror(List<Number> values) {
...
+        TransferResult<CFValue, CFStore> result = super.visitFieldAccess(node, in);
+
+        modifyArrayLenBasedOnIntValOnArrayLength(node, result.getRegularStore());
+
+        return result;
+    }
+
+    private void modifyArrayLenBasedOnIntValOnArrayLength(
+            FieldAccessNode arrayLengthNode, CFStore store) {
+        // If array.length is encountered, transform its @IntVal annotation into an @ArrayLen annotation for array.
```
smillst: This would make a good method comment.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -185,13 +190,80 @@ private AnnotationMirror createNumberAnnotationMirror(List<Number> values) {
...
+            FieldAccessNode node, TransferInput<CFValue, CFStore> in) {
+
+        TransferResult<CFValue, CFStore> result = super.visitFieldAccess(node, in);
+
+        modifyArrayLenBasedOnIntValOnArrayLength(node, result.getRegularStore());
+
+        return result;
+    }
+
+    private void modifyArrayLenBasedOnIntValOnArrayLength(
```
smillst: This name is unclear.  I would change it to something like "refineArrayWithConstantLength"  or "refineArrayAtLengthAccess".

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -185,13 +190,80 @@ private AnnotationMirror createNumberAnnotationMirror(List<Number> values) {
...
+            } else {
+                combinedAnno =
+                        atypefactory
+                                .getQualifierHierarchy()
+                                .greatestLowerBound(oldArrayAnno, newArrayAnno);
+            }
+            Receiver arrayRec =
+                    FlowExpressions.internalReprOf(
+                            analysis.getTypeFactory(), arrayLengthNode.getReceiver());
+            store.clearValue(arrayRec);
```
smillst: Why clear the old value?  Isn't combinedAnno always a subtype of the previous annotation?

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -185,13 +190,80 @@ private AnnotationMirror createNumberAnnotationMirror(List<Number> values) {
...
+                    atypefactory.getAnnotationMirror(
+                            arrayLengthNode.getReceiver().getTree(), ArrayLen.class);
+            AnnotationMirror combinedAnno;
+            if (oldArrayAnno == null) {
+                combinedAnno = newArrayAnno;
+            } else {
+                combinedAnno =
+                        atypefactory
+                                .getQualifierHierarchy()
+                                .greatestLowerBound(oldArrayAnno, newArrayAnno);
```
smillst: Please add a comment explaining why GLB.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -185,13 +190,80 @@ private AnnotationMirror createNumberAnnotationMirror(List<Number> values) {
...
+            store.insertValue(arrayRec, combinedAnno);
+        } else {
+            lengthAnno =
+                    AnnotationUtils.getAnnotationByClass(value.getAnnotations(), BottomVal.class);
+            if (lengthAnno != null) {
+                Receiver arrayRec =
+                        FlowExpressions.internalReprOf(
+                                analysis.getTypeFactory(), arrayLengthNode.getReceiver());
+                store.clearValue(arrayRec);
+                store.insertValue(arrayRec, lengthAnno);
```
smillst: I don't see why this code is needed.  Would you please add a comment?

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -555,56 +628,131 @@ private AnnotationMirror createNumberAnnotationMirror(List<Number> values) {
...
+
+        List<Number> thenRightVals = new ArrayList<>();
+        List<Number> elseRightVals = new ArrayList<>();
+
+        // Because resultValues contains the (boolean) results of each comparison (i.e. the product of the number
+        // of elements in each list), we have to iterate through parts of the list at a time in order to find all
+        // the results of comparisons involving a particular value in the left list.
+        for (int i = 0; i < lefts.size(); i++) {
+            for (int j = 0; j < rights.size(); j++) {
+                if (resultValues.get(j + i * rights.size())) {
```
smillst: It still isn't clear why this is the correct index.  Can you populate these lists in the for loop above?

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -555,56 +628,131 @@ private AnnotationMirror createNumberAnnotationMirror(List<Number> values) {
...
+            CFStore store, List<?> results, Node node, boolean canGoToBottom) {
+        AnnotationMirror anno =
+                atypefactory.createResultingAnnotation(
+                        node.getType(), results.size() == 0 && !canGoToBottom ? null : results);
+        AnnotationMirror currentAnno =
+                atypefactory
+                        .getAnnotatedType(node.getTree())
+                        .getAnnotationInHierarchy(atypefactory.BOTTOMVAL);
+        Receiver rec = FlowExpressions.internalReprOf(analysis.getTypeFactory(), node);
+        store.insertValue(
```
smillst: Please add a comment explaining why GLB is used.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -555,56 +628,131 @@ private AnnotationMirror createNumberAnnotationMirror(List<Number> values) {
...
+                elseStore, elseRightVals, rightNode, canGoToBottom);
+
         return resultValues;
     }
 
+    private void createAnnotationFromResultsAndAddToStore(
+            CFStore store, List<?> results, Node node, boolean canGoToBottom) {
+        AnnotationMirror anno =
+                atypefactory.createResultingAnnotation(
+                        node.getType(), results.size() == 0 && !canGoToBottom ? null : results);
```
smillst: I don't understand why this is required.

framework/src/org/checkerframework/common/value/ValueTransfer.java
kelloggm: While I agree that the comment on this loop could be better (I'll rewrite it), I don't think it makes sense to try to do this above. That would require duplicating the same three or four lines of code for each type of comparison (something like: if the result of the comparison was true, put it in this list, else put it in that list). I think the single pair of for loops is clearer, even if the indexing is difficult to grasp. But I think we can make it easier to understand with a better comment.

smillst: >That would require duplicating the same three or four lines of code for each type of comparison

What about this:
```
                boolean result  = false;
                 switch (op) {
                     case EQUAL:
                         result = nmLeft.equalTo(right));
                          break;
                  ... other cases...
              }  // end of switch
             resultValues.add(result);
             if(result) {
                 thenLeftValues.add(left);
                 thenRightValues.add(right);
             } else {
                  ...
```

kelloggm: That would work, you're right. I'll do that.

kelloggm:
@smillst I believe I've addressed your concerns, and I'd appreciate it if you could take another look.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -521,90 +597,165 @@ private AnnotationMirror createNumberAnnotationMirror(List<Number> values) {
...
+                    thenRightVals.add(right);
+                } else {
+                    elseLeftVals.add(left);
+                    elseRightVals.add(right);
+                }
             }
         }
+
+        // The result can only be bottom if both the comparison had some results and one of the four lists is empty.
+        boolean canGoToBottom = resultValues.size() > 0;
```
smillst: The contract of getNumericalValues changed on 3/23 (so after you first made this pull request).   I merged in those changes this morning.  Now, getNumbericalValues returns null if the type is `@BottomVal` or `@UnknownVal`.  And if `lefts` and `rights` are null, the method returns before this point.  So, I think this is always true now.

Also, I think `==` needs special handling, see failing testcase I added.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -691,95 +771,172 @@ private Range calculateRangeUnaryOp(
...
-            TransferInput<CFValue, CFStore> p) {
-        List<Boolean> resultValues = new ArrayList<>();
-        if (!isIntRange(leftNode, p) && !isIntRange(rightNode, p)) {
+            TransferInput<CFValue, CFStore> p,
+            CFStore thenStore,
+            CFStore elseStore) {
+        if (isIntRange(leftNode, p) || isIntRange(rightNode, p)) {
             // TODO:
             // Handle @IntRange annotation when the control flow refinement is implemented
             // typetools/checker-framework#1163
```
smillst: @kelloggm FYI, I think this can be done after this PR is merged. (Maybe by you maybe someone else.)

smillst:
@kelloggm FYI I merged in master, which now includes `@IntRange`.

framework/src/org/checkerframework/common/value/ValueTransfer.java
kelloggm: I've taken a look at the test case you provided. After quite a bit of digging, it turns out that the problem is that the Value Checker isn't handling the mix of `Integer` and `int` quite as precisely as we would like. Here are two simplified test cases:

```java
    void simpleTest1(@BottomVal Integer x, int y) {
	if (x == y) {
	    @BottomVal int z = y;
	}
    }

    void simpleTest2(@BottomVal Integer x, Integer y) {
	if (x == y) {
	    @BottomVal int z = y;
	}
    }
```

`SimpleTest1` issues a compile warning on the assignment to `@BottomVal int z`. `SimpleTest2` does not. I've determined that this happens because when an `Integer` and an `int` are compared, the `Integer`, `x`, is desugared to `x.intValue()`, which is compared to `y`. But `x.intValue()`'s type isn't being properly assigned when `x` is `@BottomVal`. 

I can look for a fix, but I don't think it'll be in the code here, and I wanted to let you know what was going on.

smillst: Thanks for figuring this out. I changed the test case to be two `Integer` and it now passes.  Could you open an issue for the problem you describe above? I don't think you need to fix it.

On a separate note, I still think this line:
`boolean canGoToBottom = resultValues.size() > 0;`
is always true.

kelloggm: I'll open an issue for that. And I think you're right about `canGoToBottom`; I'll remove it.

kelloggm:
@smillst I think this PR is ready to go and that I've addressed everything we've talked about.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -691,95 +771,179 @@ private Range calculateRangeUnaryOp(
...
+        createAnnotationFromResultsAndAddToStore(elseStore, elseLeftVals, leftNode);
+        createAnnotationFromResultsAndAddToStore(thenStore, thenRightVals, rightNode);
+        createAnnotationFromResultsAndAddToStore(elseStore, elseRightVals, rightNode);
+
         return resultValues;
     }
 
+    private void createAnnotationFromResultsAndAddToStore(
+            CFStore store, List<?> results, Node node) {
+        // createResultingAnnotation returns bottom if an empty list is passed. So,
```
smillst: Is this comment out of date now?

PR #994, opened 2016-12-15
Integrate @IntRange annotation with ValueChecker

JasonMrX:
Hello Professor @wmdietl,

Would you kindly review this PR. 

Thanks and regards,

Jason

PR #1165, opened 2017-03-21
getValues methods:  use null, not empty list, to represent top value

mernst:


framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -113,39 +123,46 @@ public ValueTransfer(CFAbstractAnalysis<CFValue, CFStore, CFTransfer> analysis)
...
     }
 
     private AnnotationMirror createStringValAnnotationMirror(List<String> values) {
-        if (values.isEmpty()) {
+        if (values == null) {
             return ((ValueAnnotatedTypeFactory) atypefactory).UNKNOWNVAL;
         }
+        if (values.isEmpty()) {
+            return ((ValueAnnotatedTypeFactory) atypefactory).BOTTOMVAL;
+        }
```
JasonMrX: I got rid of all this checks in typetools/checker-framework#994 because ValueAnnotatedTypeFactory#createStringAnnotation is doing these checks already. Do we still need to keep them?

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -63,26 +62,34 @@ public ValueTransfer(CFAbstractAnalysis<CFValue, CFStore, CFTransfer> analysis)
...
+        if (topAnno != null) {
+            return null;
         }
-        numberAnno = AnnotationUtils.getAnnotationByClass(value.getAnnotations(), BottomVal.class);
-        if (numberAnno != null) {
-            return Collections.singletonList("null");
+        AnnotationMirror bottomAnno =
+                AnnotationUtils.getAnnotationByClass(value.getAnnotations(), BottomVal.class);
+        if (bottomAnno != null) {
+            return new ArrayList<String>();
```
JasonMrX: I think the `Collections.singletonList("null")` is returned on purpose to handle `String + null` situation, e.g., `"str" + null == "strnull"`

mernst: I'm not sure.  @smillst , what do you think?

smillst: Jason is right, but this isn't a good place to do that.  The `null` literal is only converted to `"null"` at string concatenations.  I corrected the code and added another test case in 2692e56af5911e3719ec15fa7eed48fd95a274d0.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -113,38 +124,35 @@ public ValueTransfer(CFAbstractAnalysis<CFValue, CFStore, CFTransfer> analysis)
         return ValueAnnotatedTypeFactory.getCharValues(intAnno);
     }
 
+    /** Returns a list of possible values, or null if no estimate is available. */
```
smillst: This method returns null when the type of subNode is `@BottomVal`.  This seem to go against convention.

smillst: This method returns null when the type of subNode is @BottomVal. This seems to go against convention.

mernst: I changed the documentation to clarify:
"Returns a list of possible values, or null if no estimate is available and any value is possible."
Does that make sense now?

PR #1168, opened 2017-03-21
Allows LBC to correctly handle assignments in comparisons

kelloggm:
Following the example of the nullness checker, which handles this correctly.

Fixes kelloggm#58


kelloggm:
I don't think it's possible to construct an example for MinLen. I've written a test that covers the SameLen and UBC cases, and written code so that the UBC case works correctly. The SameLen case is working without any code changes.

kelloggm:
I'll merge as soon as the tests pass.

PR #1172, opened 2017-03-22
Adds support for switch statements in dataflow refinement

kelloggm:
Fixes most cases with switch statements. There is one situation this change doesn't seem to handle correctly, which is when the `default` case is an else (it doesn't give the appropriate type in the else branch in the value checker tests). I think the problem with this is more fundamental than the level at which I'm making this change - my guess is that the CFG is just wrong, since the default case ought to be treated as an else. If it were, then this change should also make that case work.

Fixes kelloggm#126


framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
```diff
@@ -1042,7 +1042,19 @@ protected void processConditionalPostconditions(
...
     public TransferResult<V, S> visitCase(CaseNode n, TransferInput<V, S> in) {
         S store = in.getRegularStore();
-        return new RegularTransferResult<>(finishValue(null, store), store);
+        TransferResult<V, S> result =
+                new ConditionalTransferResult<>(
+                        finishValue(null, store), in.getThenStore(), in.getElseStore(), false);
+
+        V caseValue = null;
+        if (n.getCaseOperand()
+                != null) { // When encountering a default node, getCaseOperand returns null.
```
smillst: I debugged through this and n.getCaseOperand() is never null.  Looking at the code that creates the CaseNodes (org.checkerframework.dataflow.cfg.CFGBuilder.CFGTranslationPhaseOne.SwitchBuilder#buildCase),  a CaseNode isn't created for the default case.

kelloggm: Ah, I can remove that. I experimented with actually adding a case node for the default case in `SwitchBuilder`, and needed to add that null check there. But since that didn't result in `default` blocks working correctly, I should've removed the null check.

jack17529:
@smillst do you know some bugs in project that are easy but you don't have time doing , then please assign them to me.

mernst:
@jack17592 This bug is not the place to ask such a question.  https://checkerframework.org/gsoc-2017-ideas.html#get-started says  how to get started, for prospective GSOC students.

kelloggm:
I've addressed all comments and all tests pass.

Issue #1147, opened 2017-03-17
Class StringJoiner of java 8 needs to be annotated for add method


kanha95:
import org.checkerframework.checker.nullness.qual.*;
import java.util.StringJoiner;
class Test2{

public static void main(String[] args){

StringJoiner sj=new StringJoiner(",");

    sj.add("a");
    sj.add(null);

    System.out.println(sj);
}
}

It's a correct program but Nullness Checker gives false warnings at compile time. The documentation states that null is a valid argument, so the parameter should be annotated as @Nullable.

mernst:
I agree.  Thanks for pointing out this problem.

smillst:
Close via dcf80de

PR #1176, opened 2017-03-23
added annotated StringJoiner.java and a test file to support the chan…

kanha95:
fixes Issue #1147


checker/tests/nullness/Issue1147.java
```diff
@@ -0,0 +1,16 @@
...
+import java.util.StringJoiner;
+
+class Issue1147 {
+
+    public static void main(String[] args) {
+
+        StringJoiner sj = new StringJoiner(",");
+
+        sj.add("a");
+        // Nullness Checker raises false positive warning
```
mernst: Please remove this comment because it won't be true once your pull request is merged.

kanha95: sure

kanha95:
@mernst Please review the codes and suggest why it is failing in case of jdk8 test.

mernst:
Since `@NonNull` is the default, those annotations are not necessary.  If you aren't going to include the whole original file, there is no need for private members.  It seems that the `add` method is the only one needed.

kanha95:
Yes, Right. I will fix it up. Is that the reason for failure with jdk 8 test? i don't think so.

mernst:
Currently, to same time during builds, Travis-CI does not rebuild the JDK but downloads a previously-built one.
This makes builds much faster, but it suffers spurious failures for any build that adds a test for JDK changes. :-(

kanha95:
@mernst removing the @NonNull tags generates error. I think @NonNull is not set as default type here as StringJoiner class doesn't take any Type parameters. So by default it's extending the  @Nullable Object. The previously annotated code was fine though. Should i revert the changes back?

mernst:
Could you be more specific?  What new error results from removing the `@NonNull` annotations, that does not result from having them present?

kanha95:
@mernst 
I am not sure professor. I am currently analyzing the results thrown by travis build. They make no sense at all. Errors & warnings are forecasted out of some unusual method bodies which are not present in my source code.

mernst:
When you created this pull request, did you allow changes to your fork, as described here:
https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/
If not, could you do so?  Thanks.

mernst:
Did you ensure that all the tests work locally before pushing them to your branch?

mernst:
Can you pull upstream into your branch or give the Checker Framework maintainers permission to do so?

kanha95:
@mernst  Sorry for late reply. I was out for my college. Yes i had checked them locally professor. I have allowed edits from the maintainers.

mernst:
I pulled upstream into your branch.  I am hopeful that this will fix the tests; we'll see.

kanha95:
Thanks again !!! I am sorry for the branch name. i would definitely write descriptive branch names in my future pull requests.

PR #1179, opened 2017-03-23
minor fixes to TreeSet.java class

kanha95:
Minor changes to fix certain missed warnings.

### DIFF
--- a/checker/jdk/nullness/src/java/util/TreeSet.java
+++ b/checker/jdk/nullness/src/java/util/TreeSet.java
@@ -3,9 +3,10 @@ import org.checkerframework.dataflow.qual.Pure;
 import org.checkerframework.dataflow.qual.SideEffectFree;
 import org.checkerframework.checker.nullness.qual.EnsuresNonNullIf;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.checkerframework.checker.nullness.qual.NonNull;
 
 // Subclasses of this interface/class may opt to prohibit null elements
-public class TreeSet<E extends @Nullable Object> extends AbstractSet<E> implements NavigableSet<E>, Cloneable, java.io.Serializable {
+public class TreeSet<E extends @NonNull Object> extends AbstractSet<E> implements NavigableSet<E>, Cloneable, java.io.Serializable {
   private static final long serialVersionUID = 0;
   public TreeSet() { throw new RuntimeException("skeleton method"); }
   public TreeSet(Comparator<? super E> a1) { throw new RuntimeException("skeleton method"); }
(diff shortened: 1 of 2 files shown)


mernst:
Your change seems to prohibit this legal code:
```java
import java.util.Comparator;
import java.util.TreeSet;

public class TreeSetNullable {

        TreeSet<@Nullable Integer> ts =
                new TreeSet<@Nullable Integer>(new IntegerOrNullComparator());

        ts.add(-5);
        ts.add(null);
        ts.add(10);
        ts.add(20);

        System.out.println(ts.size());
    }

    private static class IntegerOrNullComparator implements Comparator<@Nullable Integer> {
        public int compare(@Nullable Integer o1, @Nullable Integer o2) {
            // Treat null as 0
            if (o1 == null) {
                o1 = 0;
            }
            if (o2 == null) {
                o2 = 0;
            }
            return o1.compareTo(o2);
        }
    }
}
```

kanha95:
@mernst  I agree professor with your given example. But the above code overwrites the basic property of SET theory.

Consider the following 2 cases to the above code:

case 1:
        ts.add(null);
        ts.add(10);
        ts.add(20);
        ts.add(-5);
        ts.add(0); 
        System.out.println(ts);

Output: [-5, null, 10, 20]

case 2:
        ts.add(0); 
        ts.add(null);
        ts.add(10);
        ts.add(20);
        ts.add(-5);
        System.out.println(ts);

Output: [-5, 0, 10, 20]

For the same set of input we will have 2 different outputs just based upon the first arrival element takes the place logic. This completely changes the basic logic behind sets.

kanha95:
Also consider this example:

import java.util.*;
import java.lang.*;
import java.io.*;

 class IntegerOrNullComparator implements Comparator<Integer> {
        public int compare(Integer o1, Integer o2) {
            // Treat null as 0
            if (o1 == null) {
                o1 = 0;
            }
            if (o2 == null) {
                o2 = 1;
            }
            return o1.compareTo(o2);
        }
    }

class TreeSetNullable
{
	public static void main (String[] args) throws java.lang.Exception
	{
		  TreeSet<Integer> ts =
                new TreeSet<Integer>(new IntegerOrNullComparator());

       
        ts.add(null); 
        ts.add(null);
        ts.add(10);
        ts.add(20);
        ts.add(-5);
       

        System.out.println(ts);
	}
}

Output: [-5, null, null, 10, 20]

Here we get 2 null elements in the output.
It completely violates the set property. Using such comparators will treat TreeSet just as a container to hold elements ignoring its Set property.

mernst:
These are bugs in the Comparator that I hastily wrote, and they don't affect the main thrust of the concern.  Thanks for pointing them out, though!  Changing the comparator to the following fixes the bugs,  I think.

```java
    private static class IntegerOrNullComparator implements Comparator<@Nullable Integer> {
        public int compare(@Nullable Integer o1, @Nullable Integer o2) {
            // Treat null as last element
            if (o1 == null && o2 == null) {
                return 0;
            }
            if (o1 == null) {
                return 1;
            }
            if (o2 == null) {
                o2 = -1;
            }
            return o1.compareTo(o2);
        }
    }
```

mernst:
Your commit message says,
> Modified as per java docs. TreeSet does not permit null elements.
Can you point out where the `TreeSet` Javadocs prohibit null elements?

kanha95:
[https://docs.oracle.com/javase/7/docs/api/java/util/TreeSet.html](url)

In the above url , just below the add method , it's clearly mentioned that
 "NullPointerException - if the specified element is null and this set uses natural ordering, or its comparator does not permit null elements"

So i think by changing its default comparator we are changing its basic properties too. Still after analyzing your mentioned example i am confused professor. Kindly guide me.

Thanking you

mernst:
That Javadoc specification does not say that null is prohibited in all situations.  It says that a NullPointerException is thrown in certain situation.

The comment says that null is **permitted** when the comparator is provided and the comparator permits null elements (and that null is prohibited in other situations, but not in all situations).

kanha95:
Yes right. But then for my given test code the nullness checker will pass even though at runtime it will lead to nullPointerException. I think this limitation can't be removed now.

mernst:
That reasoning seems correct and justifies the change.
But the statement in the commit message that these annotations reflect the Javadoc is incorrect.

kanha95:
@mernst  I updated that comment. But can you tell me why my jdk 7 & jdk 8 test build are failing ?

mernst:
I will merge the changes.  Thanks!

I have a small request:  in the future, could you please use more descriptive names for your branches than `devr2`?

kanha95:
@mernst  Thanks professor and Sorry for the irrelevant branch names. i would use descriptive names in future.

PR #1181, opened 2017-03-24
Infer array length types from SameLen annotations.

kelloggm:
Given an array `arr` with a `@SameLen` annotation, when `arr.length` is encountered, it should be `@LTEqLengthOf` each array in the `@SameLen` annotation. Because we have this information, we should really be inferring these types correctly.

Fixes part of the problem in kelloggm#134


checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
```diff
@@ -432,10 +433,29 @@ private UBQualifier removeArrayLengths(LessThanLengthOf i, LessThanLengthOf j) {
...
+            List<String> sameLenArrays =
+                    sameLenAnno == null
+                            ? new ArrayList<String>()
+                            : IndexUtil.getValueOfAnnotationWithStringArgument(sameLenAnno);
+
+            if (!sameLenArrays.contains(arrayRec.toString())) {
+                sameLenArrays.add(arrayRec.toString());
+            }
+
+            ArrayList<String> negativeOnes = new ArrayList<>(sameLenArrays.size());
```
mernst: I would call this `offsets` which reflects its semantics rather than `negativeOnes` which reflects its contents.

PR #1182, opened 2017-03-24
Add a checker to handle the contract of the JDK's binary search method

kelloggm:
Fixes kelloggm#121 (modulo samelen bugs).

Doing this as a checker made the most sense, because we want the information to be available to both the lower and upper bound checkers.


checker/src/org/checkerframework/checker/index/searchindex/SearchIndexAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,106 @@
...
+                        .containsAll(IndexUtil.getValueOfAnnotationWithStringArgument(subAnno));
+            }
+            if (AnnotationUtils.areSameByClass(subAnno, NegativeIndexFor.class)) {
+                if (AnnotationUtils.areSameByClass(superAnno, SearchIndex.class)) {
+                    return true;
+                }
+                return IndexUtil.getValueOfAnnotationWithStringArgument(superAnno)
+                        .containsAll(IndexUtil.getValueOfAnnotationWithStringArgument(subAnno));
+            }
+            return false; // dead code
```
mernst: If it's dead code, throw an exception rather than silently returning a value.

checker/src/org/checkerframework/checker/index/searchindex/SearchIndexAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,106 @@
...
+            if (AnnotationUtils.areSameByClass(subAnno, SearchIndex.class)) {
+                if (AnnotationUtils.areSameByClass(superAnno, NegativeIndexFor.class)) {
+                    return false;
+                }
+                return IndexUtil.getValueOfAnnotationWithStringArgument(superAnno)
+                        .containsAll(IndexUtil.getValueOfAnnotationWithStringArgument(subAnno));
+            }
+            if (AnnotationUtils.areSameByClass(subAnno, NegativeIndexFor.class)) {
+                if (AnnotationUtils.areSameByClass(superAnno, SearchIndex.class)) {
+                    return true;
```
mernst: Don't you need to check the argument values in this case?

checker/src/org/checkerframework/checker/index/searchindex/SearchIndexAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,106 @@
...
+                }
+                return IndexUtil.getValueOfAnnotationWithStringArgument(superAnno)
+                        .containsAll(IndexUtil.getValueOfAnnotationWithStringArgument(subAnno));
+            }
+            return false; // dead code
+        }
+    }
+
+    AnnotationMirror createNegativeIndexFor(List<String> arrays) {
```
mernst: Please write a brief comment for each method and field.  This one is pretty obvious, but a one-line comment will still help.

checker/src/org/checkerframework/checker/index/searchindex/SearchIndexChecker.java
```diff
@@ -0,0 +1,24 @@
+package org.checkerframework.checker.index.searchindex;
+
+import java.util.LinkedHashSet;
+import org.checkerframework.common.basetype.BaseTypeChecker;
+import org.checkerframework.common.value.ValueChecker;
+import org.checkerframework.framework.source.SuppressWarningsKeys;
+
+/**
+ * An internal checker that assists the Index Checker in correctly typing the results of calls to
+ * the JDK's binarySearch function.
```
mernst: Add `{@link ...}`

checker/src/org/checkerframework/checker/index/searchindex/SearchIndexTransfer.java
```diff
@@ -0,0 +1,74 @@
...
+    private CFAnalysis analysis;
+
+    public SearchIndexTransfer(CFAnalysis analysis) {
+        super(analysis);
+        this.analysis = analysis;
+        aTypeFactory = (SearchIndexAnnotatedTypeFactory) analysis.getTypeFactory();
+    }
+
+    /**
+     * Special binary search handling: if the left value is exactly comparison, and the right side
```
mernst: English: "exactly comparison".
More generally, I don't understand the documentation.  Oftentimes, giving a concrete example is helpful.  What are examples of the code being matched against, and what is inferred from them?

checker/src/org/checkerframework/checker/index/searchindex/SearchIndexTransfer.java
```diff
@@ -0,0 +1,74 @@
...
+     * is a search index, then convert the right side in the store to a negative index for.
+     */
```
mernst: Please document the formal parameters, especially `comparison` which is obscure.  Should you use an enum rather than literal values?

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -380,10 +391,33 @@ public Void visitMethodInvocation(MethodInvocationTree tree, AnnotatedTypeMirror
...
+                addAnnotationForBitwiseComplement(
+                        getSearchIndexAnnotatedTypeFactory().getAnnotatedType(node.getExpression()),
+                        type);
+            } else {
+                type.addAnnotation(UNKNOWN);
+            }
             return super.visitUnary(node, type);
         }
 
+        private void addAnnotationForBitwiseComplement(
```
mernst: Please document every method and field.  Otherwise I cannot review it, and more seriously future programmers (including you) won't understand it.

docs/manual/index-checker.tex
```diff
@@ -106,13 +109,13 @@ \section{Index Checker structure and annotations\label{index-annotations}}
...
   For example, the
-  \sunjavadoc{java/util/Arrays.html\#binarySearch-java.lang.Object:A-int-int-java.lang.Object-}{Arrays.binarySearch}
+  \sunjavadoc{java/util/Arrays.html\#fill-java.lang.Object:A-int-int-java.lang.Object-}{Arrays.fill}
    method is declared as
 
   \begin{mysmall}
   \begin{Verbatim}
   class Arrays {
-    int binarySearch(Object[] a, @IndexFor("#1") int fromIndex, @IndexOrHigh("#1") int toIndex, Object key)
+    void fill(Object[] a,  @IndexFor("#1") int fromIndex, @IndexOrHigh("#1") int toIndex, Object val)
```
mernst: Spurious whitespace before `@IndexFor`.

docs/manual/index-checker.tex
```diff
@@ -370,6 +373,31 @@ \section{Arrays of the same length\label{index-samelen}}
...
 \end{description}
 
+\section{Special handling for binarySearch\label{index-searchindex}}
+
+The JDK's implementation of
+\sunjavadoc{java/util/Arrays.html\#binarySearch-java.lang.Object:A-int-int-java.lang.Object-}{binary search}
+uses a contract that is not directly
+supported by the Index Checker. Because binary search is a common array operation,
+the Index Checker includes a special type system to check whether a value originated
+from a call to binary search. This type system should never be user-visible, but for
```
mernst: "should never be user-visible" is a bit obscure.

docs/manual/index-checker.tex
```diff
@@ -370,6 +373,31 @@ \section{Arrays of the same length\label{index-samelen}}
   \code{null} has this type.
 \end{description}
 
+\section{Special handling for binarySearch\label{index-searchindex}}
+
+The JDK's implementation of
+\sunjavadoc{java/util/Arrays.html\#binarySearch-java.lang.Object:A-int-int-java.lang.Object-}{binary search}
+uses a contract that is not directly
```
mernst: It *is* supported, because the Search Index Checker is part of the Index Checker.

smillst:
@kelloggm should I review this?

mernst:
@smillst I believe there are still outstanding issues from my review.
You can probably wait until someone assigns it to you.

kelloggm:
I've resolved the issues identified by Mike in his review. @smillst, it should be ready for you.

mernst:
It's up to the reviewer, not the reviewee, to decide whether the changes are adequate.

There are often issues that the reviewer didn't get to -- maybe they didn't make sense but will in the context of better code or comments.

kelloggm:
Did I imply otherwise? My impression from your previous comment was that I should assign it to Suzanne when I though I had addressed your concerns. I can reassign it to you instead if you want to review it again.

mernst:
In general, a PR should be reviewed by one person at a time.  It's usually more efficient to do reviews sequentially rather than in parallel.  Changing the reviewer is a pretty aggressive tactic that says you don't consider the original one useful.

kelloggm:
My apologies, then. That's not what I intended at all; I thought you'd given me blessing to do so.

mernst:
Sorry, on my side, for writing a comment that was open to misinterpretation.

kelloggm:
No worries! As long as everyone had good intentions, there's no problem.

checker/src/org/checkerframework/checker/index/lowerbound/LowerBoundAnnotatedTypeFactory.java
```diff
@@ -261,12 +269,30 @@ public Void visitUnary(UnaryTree tree, AnnotatedTypeMirror typeDst) {
...
+                    break;
                 default:
                     break;
             }
             return super.visitUnary(tree, typeDst);
         }
 
+        /**
+         * Special case for bitwise complement on the result of binary search that has been compared
+         * to zero.
```
mernst: I would be a bit more specific:  if the type of `expr` is `@NegativeIndexFor`, then the type of `~expr` is `@IndexFor`.
That's not quite what this routine does -- explaining why would be helpful in understanding the code and how it interacts with other parts of the checker.

checker/src/org/checkerframework/checker/index/qual/NegativeIndexFor.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import org.checkerframework.framework.qual.JavaExpression;
+import org.checkerframework.framework.qual.SubtypeOf;
+
+/**
+ * The annotated expression is between -1 and -a.length - 1, inclusive, of each sequence a listed in
```
mernst: I personally find ranges/intervals easier to understand if the lower bound is stated before the upper bound.  Also, for a variable name that is also a word, like `a`, wrapping it in `{@code ...}` is helpful.

checker/src/org/checkerframework/checker/index/qual/NegativeIndexFor.java
```diff
@@ -0,0 +1,27 @@
...
+ * java.util.Arrays#binarySearch(Object[],Object) binary search} routine.
+ *
+ * @checker_framework.manual #index-checker Index Checker
+ */
+@SubtypeOf(SearchIndex.class)
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE_USE, ElementType.TYPE_PARAMETER})
+public @interface NegativeIndexFor {
+    /**
+     * Sequences for which this value is a negative index; that is, the expression is in the range
```
mernst: I would put "negative index" in quotes, because it's not actually the negative of an index.

checker/src/org/checkerframework/checker/index/qual/NegativeIndexFor.java
```diff
@@ -0,0 +1,27 @@
...
+     * -1 to -1 * a.length - 1, inclusive, for each sequence a.
```
mernst: I found the multiplication hard to read, maybe because the spacing between and within a formula is the same.  Maybe putting the bounds in order would be helpful, along with dropping the multiplication for consistency with the comment above.

checker/src/org/checkerframework/checker/index/qual/SearchIndex.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import org.checkerframework.framework.qual.JavaExpression;
+import org.checkerframework.framework.qual.SubtypeOf;
+
+/**
+ * The annotated expression evaluates to an integer whose between -a.length - 1 and a.length - 1,
```
mernst: English grammar mistake on this line.

checker/src/org/checkerframework/checker/index/qual/SearchIndex.java
```diff
@@ -0,0 +1,31 @@
...
+ * inclusive, for all sequences a listed in the annotation.
+ *
+ * <p>This type is designed with one goal: to handle the convention of {@link
+ * java.util.Arrays#binarySearch(Object[],Object) binary search} in the JDK, which returns an
+ * integer of this type. Because binary search is common in practice, this type exists only to
```
mernst: Cut the last sentence.

checker/src/org/checkerframework/checker/index/qual/SearchIndex.java
```diff
@@ -0,0 +1,31 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import org.checkerframework.framework.qual.JavaExpression;
+import org.checkerframework.framework.qual.SubtypeOf;
+
+/**
+ * The annotated expression evaluates to an integer whose between -a.length - 1 and a.length - 1,
+ * inclusive, for all sequences a listed in the annotation.
+ *
+ * <p>This type is designed with one goal: to handle the convention of {@link
```
mernst: I would just say, "binary search returns an integer of this type", without the other caveats.

checker/src/org/checkerframework/checker/index/qual/SearchIndex.java
```diff
@@ -0,0 +1,31 @@
...
+ * handle that particular usage.
+ *
+ * @checker_framework.manual #index-checker Index Checker
+ */
+@SubtypeOf(SearchIndexUnknown.class)
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE_USE, ElementType.TYPE_PARAMETER})
+public @interface SearchIndex {
+    /**
+     * Sequences for which the annotated expression is the result of a call to the JDK's {@link
```
mernst: It isn't necessarily a result of such a call, but it has the same type or bounds as that.  (The value couldn't be the result of calls to binary search on two sequences.)

checker/src/org/checkerframework/checker/index/qual/SearchIndexUnknown.java
```diff
@@ -0,0 +1,17 @@
+package org.checkerframework.checker.index.qual;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Target;
+import org.checkerframework.framework.qual.DefaultQualifierInHierarchy;
+import org.checkerframework.framework.qual.SubtypeOf;
+
+/**
+ * The top type for the SearchIndex type system, which helps the Index Checker correctly type calls
```
mernst: The "which" clause doesn't refer to the type, but that's how I read it.  I would drop the dependent clause.

checker/src/org/checkerframework/checker/index/searchindex/SearchIndexAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,117 @@
...
+    public SearchIndexAnnotatedTypeFactory(BaseTypeChecker checker) {
+        super(checker);
+        UNKNOWN = AnnotationUtils.fromClass(elements, SearchIndexUnknown.class);
+        BOTTOM = AnnotationUtils.fromClass(elements, SearchIndexBottom.class);
+        this.postInit();
+    }
+
+    /**
+     * Provides a way to query the Constant Value Checker, which computes the values of expressions
+     * known at compile time (constant prop + folding).
```
mernst: Write out abbreviations.

checker/src/org/checkerframework/checker/index/searchindex/SearchIndexAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,117 @@
...
+                return false;
+            }
+            // All annotations have arguments, because each is either NegativeIndexFor or
+            // SearchIndex. All NegativeIndexFor annotations are subtypes of SearchIndex
+            // annotations whose arguments contain their arguments.
+            if (AnnotationUtils.areSameByClass(subAnno, NegativeIndexFor.class)) {
+                List<String> superArrays =
+                        IndexUtil.getValueOfAnnotationWithStringArgument(superAnno);
+                List<String> subArrays = IndexUtil.getValueOfAnnotationWithStringArgument(subAnno);
+                return superArrays != null
```
mernst: If superArrays or subArrays is null, is that an indication of a bug?  If so, it's better to check it and throw an exception than mask it by returning a value.

You could set superArrays and subArrays outside the if statements so the code is shorter and only has to check for nullness once.

checker/src/org/checkerframework/checker/index/searchindex/SearchIndexAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,117 @@
...
+                        && subArrays != null
+                        && superArrays.containsAll(subArrays);
```
mernst: This looks backwards to me.
A `@SI({a,b})` can be used wherever a `@SI(a)` is needed, but not the reverse.  Therefore, `@SI({a,b})` is a subtype, not a supertype, of `@SI(a)`.

Before proceeding any further, I propose that you write the manual section that explains the type system, its semantics, and shows the subtyping relationships.  Writing those and getting feedback on them is supposed to be the first step, before any other part of implementation.  Skipping that step has caused wasted time.

checker/src/org/checkerframework/checker/index/lowerbound/LowerBoundAnnotatedTypeFactory.java
```diff
@@ -261,12 +269,29 @@ public Void visitUnary(UnaryTree tree, AnnotatedTypeMirror typeDst) {
...
+                            typeDst);
+                    break;
                 default:
                     break;
             }
             return super.visitUnary(tree, typeDst);
         }
 
+        /**
+         * Bitwise complement converts between {@code @NegativeIndexFor} and {@code @IndexOrHigh}.
```
mernst: Update the comment or the code, so they are consistent.  (There seem to be some assumptions here -- are they warranted?)

checker/src/org/checkerframework/checker/index/searchindex/SearchIndexTransfer.java
```diff
@@ -0,0 +1,106 @@
...
+     * <p>For example, this allows the following code to typecheck:
+     *
+     * <pre>{@code
+     * @SearchIndexFor("a") int index = Arrays.binarySearch(a, y);
+     * if (index < 0) {
+     *     @NegativeIndexFor("a") int negInsertionPoint = index;
+     * }
+     * }</pre>
+     */
+    private void specialHandlingForBinarySearch(
```
mernst: This method name doesn't convey what the routine does (which has nothing to do with binarySearch, though it is related to the `@SearchIndex` annotation and to refinement).  The documentation also doesn't say anything about greater-than or less-than tests (except indirectly in the example), which is relevant to the specification.

checker/src/org/checkerframework/checker/index/searchindex/SearchIndexTransfer.java
```diff
@@ -0,0 +1,106 @@
...
+                List<String> arrays = IndexUtil.getValueOfAnnotationWithStringArgument(rightSI);
+                AnnotationMirror nif = aTypeFactory.createNegativeIndexFor(arrays);
+                store.insertValue(
+                        FlowExpressions.internalReprOf(analysis.getTypeFactory(), right), nif);
+            }
+        }
+    }
+
+    @Override
+    protected void refineGT(
```
mernst: The test could be written as either
``` searchIndex < 0```
or as 
``` 0 > searchIndex```
and it would be straightforward to handle both.

File: checker/src/org/checkerframework/checker/index/searchindex/SearchIndexTransfer.java [hunk 26]
```diff
@@ -0,0 +1,106 @@
...
+        super(analysis);
+        this.analysis = analysis;
+        aTypeFactory = (SearchIndexAnnotatedTypeFactory) analysis.getTypeFactory();
+    }
+
+    /**
+     * {@link org.checkerframework.checker.index.qual.SearchIndexFor {@code @SearchIndexFor}} is
+     * refined only when it is compared to these values.
+     */
+    private enum ValidComparisons {
```
mernst: I would just use the `int`s 0 and -1.  All uses are within this code, and within a dozen or two lines of code, so I think the likelihood of error is small, and it avoids some bloat and some overhead for readers to figure out what is going on.

docs/manual/index-checker.tex
```diff
@@ -370,6 +373,35 @@ \section{Arrays of the same length\label{index-samelen}}
...
 
+
+\section{Binary search indices\label{index-searchindex}}
+
+The JDK's
+\sunjavadoc{java/util/Arrays.html\#binarySearch-java.lang.Object:A-java.lang.Object-}{Arrays.binarySearch}
+method returns either where the value was found, or a negative value
+indicating where the value could be inserted.  The Search Index Checker
+represents this concept.
+
```
mernst: Please add a type hierarchy figure to the manual.

checker/src/org/checkerframework/checker/index/searchindex/SearchIndexTransfer.java
kelloggm: This is handled by IndexAbstractTransfer, which this class inherits from. It defines < and <= in terms of > and >=, so that only > and >= need to be written. It's also used by the LBC and UBC. I'll make sure that there is a test that demonstrates that this works, though.

kelloggm:
@mernst I think I've addressed your feedback.

mernst:
Please double-check that the tests pass before re-assigning the pull request.  (I fixed the failure,  Ithink.)
Could you expand the type hierarchy figure in the manual to include an example of `@SearchIndexFor` and one of `@NegativeIndexFor` with multiple arguments?

checker/src/org/checkerframework/checker/index/searchindex/SearchIndexTransfer.java
```diff
@@ -0,0 +1,92 @@
...
+    private CFAnalysis analysis;
+
+    public SearchIndexTransfer(CFAnalysis analysis) {
+        super(analysis);
+        this.analysis = analysis;
+        aTypeFactory = (SearchIndexAnnotatedTypeFactory) analysis.getTypeFactory();
+    }
+
+    /**
+     * If the left value is exactly the value of {@code valueToCompareTo}, and the right side has
```
mernst: This documentation is better; thanks.

It still took me a while to understand it.  The "right side" of what?  "exactly the value" made me think of equality tests.

There isn't a connection between the code example and the method signature.

It might also be helpful to start out with something like, "If a `@SearchIndexFor` value is negative, then refine it to `@NegativeIndexFor`."  That describes the purpose, which isn't apparent from the abstract discussion of `valueToCompareTo`.

checker/src/org/checkerframework/checker/index/searchindex/SearchIndexTransfer.java
```diff
@@ -0,0 +1,92 @@
...
+            Node left, Node right, CFStore store, int valueToCompareTo) {
+        assert valueToCompareTo == 0 || valueToCompareTo == -1;
+        Long leftValue =
+                IndexUtil.getExactValue(
+                        left.getTree(), aTypeFactory.getValueAnnotatedTypeFactory());
+        if (leftValue != null && leftValue == valueToCompareTo) {
+            AnnotationMirror rightSI =
+                    aTypeFactory.getAnnotationMirror(right.getTree(), SearchIndexFor.class);
+            if (rightSI != null) {
+                List<String> arrays = IndexUtil.getValueOfAnnotationWithStringArgument(rightSI);
```
mernst: `arrays` is not tested against null, even though the comment above in `SearchIndexQualifierHierarchy.isSubtype` says that it can be null.  Should there be a check here and elsewhere?

checker/src/org/checkerframework/checker/index/searchindex/SearchIndexTransfer.java
mernst: OK, thanks.

.travis.yml
```diff
@@ -14,23 +14,23 @@ matrix:
 env:
 ## Jobs used when downloading JDK during testing.
 ## all-tests = junit + nonjunit + demos.  Separately, test building the JDK.
-  - BUILDJDK=downloadjdk JDKVER=jdk7 GROUP=all-tests
```
mernst: Be sure to undo this change before merging.

checker/src/org/checkerframework/checker/index/searchindex/SearchIndexAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,112 @@
...
+            }
+            if (AnnotationUtils.areSameByClass(superAnno, SearchIndexBottom.class)) {
+                return false;
+            }
+
+            // Each annotation is either NegativeIndexFor or SearchIndexFor.
+            List<String> superArrays = IndexUtil.getValueOfAnnotationWithStringArgument(superAnno);
+            List<String> subArrays = IndexUtil.getValueOfAnnotationWithStringArgument(subAnno);
+
+            // It's possible that either superArrays or subArrays is null in rare circumstances.
```
kelloggm: I'm not sure if it's a framework bug or not, because I haven't investigated significantly. When the check is removed, the test I cited (all-systems/Issue1006.java) throws an NPE. Since it's a regression test, I assumed that that's the desired behavior, so I wrote the check. I can spend some time investigating if you think it's worthwhile, or I can file a CF issue with the information I have now.

smillst: Somewhere the framework is creating an annotation without a default value.   I would guess this is happening in lub.  You need to implement lub and glb for for this checker.

kelloggm:
@smillst I've explicitly implemented lub and glb, and the NPE from the `all-systems/Issue1006.java` went away. It should now be ready for your review.

Unrelated, the travis build seems to be failing because the JDK isn't being built first. I'm not sure why this happens, but someone with knowledge of the build system should probably look into it. The two failing test lines (`SearchIndexTests.java`, lines 7 and 10) are both values derived from calling binary search that should be properly assigned by the JDK (with a pre-built JDK on my machine, the tests pass, and the only reason they would fail is if the JDK did not have the proper annotations - they literally check if the annotations on the JDK method are correct).

smillst:
>The travis build seems to be failing because the JDK isn't being built first.

See checker-framework/.travis.yml.  You can change whether or not Travis builds the jdk during testing in that file.  I made it to that in https://github.com/typetools/checker-framework/pull/1182/commits/f9a03f1f2a9cb6c29e6c07b0dd6712dcbf85d58b.  Make sure you revert that change before merging.  (Actually, I'll need to remake the jdk that Travis uses to include these annotations before you merge.)

checker/src/org/checkerframework/checker/index/searchindex/SearchIndexAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,152 @@
...
+            if (AnnotationUtils.areSame(a2, UNKNOWN)) {
+                return a1;
+            }
+            if (AnnotationUtils.areSame(a1, BOTTOM)) {
+                return a1;
+            }
+            if (AnnotationUtils.areSame(a2, BOTTOM)) {
+                return a2;
+            }
+            if (isSubtype(a1, a2)) {
```
smillst: This isn't right.  The GLB of `@SearchIndex({"a", "b"})` and `@NegativeIndexFor("a")` is `@NegativeIndexFor({"a", "b"})`.

checker/src/org/checkerframework/checker/index/searchindex/SearchIndexAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,152 @@
...
+            if (AnnotationUtils.areSame(a2, UNKNOWN)) {
+                return a2;
+            }
+            if (AnnotationUtils.areSame(a1, BOTTOM)) {
+                return a2;
+            }
+            if (AnnotationUtils.areSame(a2, BOTTOM)) {
+                return a1;
+            }
+            if (isSubtype(a1, a2)) {
```
smillst: This isn't right. The LUB of @SearchIndex({"a", "b"}) and @NegativeIndexFor("a") is @SearchIndex("a").

checker/src/org/checkerframework/checker/index/searchindex/SearchIndexTransfer.java
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * The transfer function for the SearchIndexFor checker. Allows {@link SearchIndexFor} to be refined
+ * to {@link NegativeIndexFor}.
+ */
+public class SearchIndexTransfer extends IndexAbstractTransfer {
+
+    // The ATF (Annotated Type Factory).
+    private SearchIndexAnnotatedTypeFactory aTypeFactory;
+
+    private CFAnalysis analysis;
```
smillst: You don't need this field.  CFAbstractTransfer has a field of the same name/type.

kelloggm: Removed.

kelloggm:
@smillst I've made LUB and GLB more precise. I think they were sound overapproximations before, and that they're now as precise as they can be. Sorry I forgot that case.

checker/src/org/checkerframework/checker/index/searchindex/SearchIndexAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,194 @@
...
+                    && subArrays.containsAll(superArrays));
+        }
+    }
+
+    /** Create a new {@code @NegativeIndexFor} annotation with the given arrays as its arguments. */
+    AnnotationMirror createNegativeIndexFor(List<String> arrays) {
+        if (arrays.size() == 0) {
+            return UNKNOWN;
+        }
+        AnnotationBuilder builder = new AnnotationBuilder(processingEnv, NegativeIndexFor.class);
```
smillst: Sort and remove duplicates? (Sorting doesn't matter for correctness, but will make error messages easier to read.)

checker/src/org/checkerframework/checker/index/searchindex/SearchIndexAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,194 @@
...
+        builder.setValue("value", arrays);
+        return builder.build();
+    }
+
+    /** Create a new {@code @SearchIndexFor} annotation with the given arrays as its arguments. */
+    AnnotationMirror createSearchIndexFor(List<String> arrays) {
+        if (arrays.size() == 0) {
+            return UNKNOWN;
+        }
+        AnnotationBuilder builder = new AnnotationBuilder(processingEnv, SearchIndexFor.class);
```
smillst: Sort and remove duplicates? (Sorting doesn't matter for correctness, but will make error messages easier to read.)

checker/src/org/checkerframework/checker/index/searchindex/SearchIndexTransfer.java
```diff
@@ -0,0 +1,95 @@
...
+ * to {@link NegativeIndexFor}.
+ */
+public class SearchIndexTransfer extends IndexAbstractTransfer {
+
+    // The ATF (Annotated Type Factory).
+    private SearchIndexAnnotatedTypeFactory aTypeFactory;
+
+    public SearchIndexTransfer(CFAnalysis analysis) {
+        super(analysis);
+        this.analysis = analysis;
```
smillst: This happens in super; don't do it here, too.

checker/src/org/checkerframework/checker/index/lowerbound/LowerBoundAnnotatedTypeFactory.java
smillst: Has this comment been addressed?

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -380,10 +396,56 @@ public Void visitMethodInvocation(MethodInvocationTree tree, AnnotatedTypeMirror
...
+         */
+        private void addAnnotationForBitwiseComplement(
+                AnnotatedTypeMirror searchIndexType, AnnotatedTypeMirror typeDst) {
+            if (AnnotationUtils.containsSameByClass(
+                    searchIndexType.getAnnotations(), NegativeIndexFor.class)) {
+                AnnotationMirror nif = searchIndexType.getAnnotation(NegativeIndexFor.class);
+                List<String> arrays = IndexUtil.getValueOfAnnotationWithStringArgument(nif);
+                List<String> negativeOnes = new ArrayList<>(arrays.size());
+                for (String s : arrays) {
+                    negativeOnes.add("-1");
```
smillst: You can use 
` List<String> negativeOnes = Collections.nCopies(arrays.size(), "-1");`
instead of this for loop.

checker/tests/index/BinarySearchTest.java
```diff
@@ -0,0 +1,43 @@
+import java.util.Arrays;
+import org.checkerframework.checker.index.qual.*;
+
+//@skip-test because of a samelen issue
```
smillst: What's the samelen issue?

docs/manual/manual.tex
```diff
@@ -41,7 +41,8 @@
 \input{map-key-checker.tex}
 \input{interning-checker.tex}
 \input{lock-checker.tex}
-%\input{index-checker.tex}
+% Include index-checker.tex EXCEPT on buffalo.cs.washington.edu
+\IfFileExists{/scratch/secs-jenkins/java/jdk1.7.0}{\relax}{\input{index-checker.tex}}
```
smillst: @mernst did you make this change?  Since I make the release on buffalo, this will cause the Index Checker to appear in the manual in the next release.

mernst: Yes, I made that change.  It works as desired for the PDF version of the manual, but not for the HTML version.  I'll commit a version that works for both PDF and HTML separately.

smillst:
I made a version of the jdk.jars with the new annotations and changed the build to download it. (See https://github.com/typetools/checker-framework/pull/1182/commits/1ac7671f94fc30121f5ad94839511d0587f5081a#diff-8e6f9ff3c77d439c1733f5fd5fe2bb6bR1437) That will need to be reverted before merging.  But hopefully the tests will now pass.

checker/src/org/checkerframework/checker/index/lowerbound/LowerBoundAnnotatedTypeFactory.java
kelloggm: Yes. Do you still think the comment is insufficient?

checker/tests/index/BinarySearchTest.java
kelloggm: This was the viewpoint adaptation issue where I wasn't creating a `DependentTypesTreeAnnotator`. I forgot that it was impacting this PR too. I've reenabled the test and checked that it passes now (at least locally).

checker/src/org/checkerframework/checker/index/lowerbound/LowerBoundAnnotatedTypeFactory.java
smillst: No, I think it's fine.  I just wasn't sure.

kelloggm:
@smillst I think this is ready to be merged. Is there anything else that needs to be addressed? If not, I can remove the JDK things and merge.

PR #1183, opened 2017-03-24
Index Checker: Adds a special case when dividing an array length

kelloggm:
If the array's minimum length is 1 or more, then dividing it by a constant >= 2 now results in a `LTLengthOf` for the array. Before it was (conservatively) a `LTEqLengthOf`.

Fixes kelloggm#137


checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -463,6 +463,20 @@ private void addAnnotationForDivide(
                 result = ((LessThanLengthOf) numerator).divide(divisor.intValue());
             }
             result = result.glb(plusTreeDivideByVal(divisor.intValue(), numeratorTree));
+
+            // If the numerator is an array length access of an array with length greater than 1, and the divisor is
```
mernst: Should this be "greater than or equal to 1", or just "non-zero length"?  That's what the test on line 475 seems to be doing.

kelloggm: The comment was just mixed up. Thanks for pointing it out. I've fixed it and I'm going to merge this PR.

PR #1193, opened 2017-03-27
Fixes an Index Checker bug and a related viewpoint adaptation bug

kelloggm:
This fixes kelloggm#139. There are two parts to the fix, but I'm PR'ing them together because either one is insufficient to fix the whole bug:

First, the SameLen Checker both actively assigns the wrong type to the right side of an assignment like `int @SameLen("otherArray") [] x = new int[otherArray.length]` (it before assigned that `@SameLen("new int[]")` via the common assignment checks, which is obviously incorrect), and fails to assign the proper type to the right hand side (i.e. the `new int...` part). These are solved by the changes to SameLenVisitor and SameLenAnnotatedTypeFactory, respectively.

Applying that fix causes a viewpoint adaptation bug to appear in the regression test case [`SLSubtyping.java` on line 22](https://github.com/typetools/checker-framework/blob/master/checker/tests/index/SLSubtyping.java). The change to `GenericAnnotatedTypeFactory` fixes that bug. The left side of that assignment has a user-written annotation (`@SameLen("f")`) and the right side is `f`, but `f` is a field, so the test was failing and claiming that `@SameLen("f")` and `@SameLen("(this).f")` were not the same.

### DIFF
--- a/checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
+++ b/checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
@@ -245,6 +245,7 @@ public class SameLenAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
     @Override
     public TreeAnnotator createTreeAnnotator() {
         return new ListTreeAnnotator(
+                super.createTreeAnnotator(),
                 new SameLenTreeAnnotator(this),
                 new PropagationTreeAnnotator(this),
                 new ImplicitsTreeAnnotator(this));
--- a/checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
+++ b/checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
@@ -2,6 +2,8 @@ package org.checkerframework.checker.index.samelen;
 
 import static org.checkerframework.checker.index.IndexUtil.getValueOfAnnotationWithStringArgument;
 
+import com.sun.source.tree.MemberSelectTree;
+import com.sun.source.tree.NewArrayTree;
 import com.sun.source.tree.Tree;
 import com.sun.source.util.TreePath;
 import java.lang.annotation.Annotation;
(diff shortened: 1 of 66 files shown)


checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
```diff
@@ -234,6 +242,44 @@ public boolean isSubtype(AnnotationMirror subAnno, AnnotationMirror superAnno) {
...
 
+    @Override
+    public TreeAnnotator createTreeAnnotator() {
+        return new ListTreeAnnotator(
+                new SameLenTreeAnnotator(this),
+                new PropagationTreeAnnotator(this),
+                new ImplicitsTreeAnnotator(this));
+    }
+
+    protected class SameLenTreeAnnotator extends TreeAnnotator {
```
mernst: Could you add documentation to indicate what this is intended to do?  The documentation might go here or elsewhere.

kelloggm: I've added a comment that describes why a tree annotator is needed.

checker/tests/index/SameLenNewArrayWithSameLength.java
```diff
@@ -0,0 +1,7 @@
+import org.checkerframework.checker.index.qual.*;
+
+class SameLenNewArrayWithSameLength {
+    public void m1(int[] a) {
+        int @SameLen("a") [] b = new int[a.length]; // index checker warning
```
wmdietlGC: The `index checker warning` is presumably before your fix. If it is expected, it should be a proper expected output. If not, reformulate to make clear that there used to be a warning.

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1195,6 +1195,10 @@ public AnnotatedTypeMirror getAnnotatedTypeLhs(Tree lhsTree) {
             case MEMBER_SELECT:
             case ARRAY_ACCESS:
                 res = getAnnotatedType(lhsTree);
+                if (dependentTypesHelper != null && lhsTree.getKind() == Kind.VARIABLE) {
```
wmdietlGC: Is this really the right place to perform this standardization? Shouldn't it be part of `getAnnotatedType`?

checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
```diff
@@ -234,6 +242,48 @@ public boolean isSubtype(AnnotationMirror subAnno, AnnotationMirror superAnno) {
         }
     }
 
+    @Override
+    public TreeAnnotator createTreeAnnotator() {
+        return new ListTreeAnnotator(
+                new SameLenTreeAnnotator(this),
+                new PropagationTreeAnnotator(this),
```
smillst: Call super!  You are missing the DependentTypesTreeAnnotator!

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
smillst: The code is wrong.  The problem is not calling super in createTreeAnnotator method.

checker/tests/index/SameLenNewArrayWithSameLength.java
```diff
@@ -0,0 +1,7 @@
+import org.checkerframework.checker.index.qual.*;
+
+class SameLenNewArrayWithSameLength {
+    public void m1(int[] a) {
```
smillst: Please add test cases where `a` is the same length of other arrays (besides `b`).

checker/src/org/checkerframework/checker/index/samelen/SameLenVisitor.java
```diff
@@ -38,6 +38,7 @@ protected void commonAssignmentCheck(
             /*@CompilerMessageKey*/ String errorKey) {
         if (valueType.getKind() == TypeKind.ARRAY
                 && TreeUtils.isExpressionTree(valueTree)
+                && !(valueTree.getKind() == Tree.Kind.NEW_ARRAY)
```
smillst: On line 52 (below), call `isReceiverToStringParsable(rec)` instead of `!(rec instanceof Unknown)`.

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
smillst: Also, in general, if you think you found a framework bug, it's good to confirm it using more than one checker.

kelloggm:
I've addressed all review comments and reverted the incorrect change to `GenericAnnotatedTypeFactory`. @smillst can you take a look at the code again? Thanks.

PR #1196, opened 2017-03-28
@BottomVal should only be converted to @StringVal("null") at StringConversions

smillst:
According to the JLS only String concatentions are StringConversions.  @BottomVal is already converted to @StringVal("null") in that location. See org.checkerframework.common.value.ValueTransfer#stringConcatenation.

Fixes kelloggm#112


wmdietl:
Is the special handling of `"null"` in `convertStringVal` still needed? https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/common/value/ValueCheckerUtils.java#L227

smillst:
@wmdietl I removed that code, too.  Thanks!

PR #1197, opened 2017-03-28
Add FieldInvariant annotation

### LINKED ISSUE #1136: Field Invariants for final fields declared in super classes
Consider the following code:
```
class Super {
    final int @MinLen(1) [] myField;
    public Super( int @MinLen(1) [] o) {
        this.myField = o;
    }
}

class Sub extends Super {
    public Sub( int @MinLen(2) [] o) {
        super(o);
        this.myField[1] = 42;  // should be legal, but currently: error, index possibly too high
    }
}
```

In `Sub`, `field` is has at least 2 elements, but there is no way to express that, so an error is issued.

A `@FieldInvariant` annotation could be used on class `Sub` to express this.  For example,
`@FieldInvariant(field = "myField", annotation = "@MinLen(2)")`,  where the value of `field` is some Java expression for a final field not declared in the class on which the field invariant appears.  (The FlowExpression parser would need to be extended to handle annotations such as `@MinLen(2)`.)

Because the field must be final, the assignment must happen in the superclass.  This will make it difficult to verify that the `@FieldInvariant` is correct.  Being able to document the invariant will be useful nonetheless.

smillst:
See Javadoc on FieldInvariant for explanation. 

Fixes #1136.


mernst:
Needs documentation in the manual.

checker/src/org/checkerframework/checker/index/qual/MinLenFieldInvariant.java
```diff
@@ -0,0 +1,21 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Target;
+import org.checkerframework.framework.qual.JavaExpression;
+
+@Inherited
+@Target(ElementType.TYPE)
+public @interface MinLenFieldInvariant {
+    /* The field which has the qualifier.  The field must
+     * be final and declared in a superclass of the class on which the field invariant applies.*/
```
mernst: applies => is written

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -330,6 +333,106 @@ public void processClassTree(ClassTree classTree) {
         super.visitClass(classTree, null);
     }
 
+    /**
+     * Check that the field invariant meets the following requirements:
```
mernst: Define what a field invariant is, or cross-reference to the relevant annotations?

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -330,6 +333,106 @@ public void processClassTree(ClassTree classTree) {
...
+     *
+     * <p>1.) If the superclass of {@code classTree} has a field invariant, then the field invariant
```
mernst: I would use <ol> instead of paragraphs, since it's a numbered list.

framework/src/org/checkerframework/common/basetype/messages.properties
```diff
@@ -79,3 +79,10 @@ contracts.conditional.postcondition.expression.parameter.name=The conditional po
 
 lambda.unimplemented=This version of the Checker Framework does not type-check lambda expressions.
 methodref.inference.unimplemented=This version of the Checker Framework does not type-check method references with implicit type arguments.
+
+field.invar.not.found=the field invariant annotation refers to fields not found in a superclass\nfields not found: %s
+field.invar.not.final=the field invariant annotation refers to fields that are not final\nfields not final: %s
+field.invar.not.subtype=the qualifier for field %s is not a subtype of the declared type\nfound: %s\ndeclared type: %s
+field.invar.not.wellformed=the field invariant annotation does not have equal numbers of fields and qualifiers.
+field.invar.not.found.superclass=the field invariant annotation is missing fields that are listed in the super class field invariant.\nfields not found: %s
```
mernst: "super class" => "superclass" (here and on next line)

framework/src/org/checkerframework/framework/qual/FieldInvariant.java
```diff
@@ -0,0 +1,37 @@
...
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import javax.lang.model.element.TypeElement;
+import org.checkerframework.framework.type.AnnotatedTypeFactory;
+/**
+ * In the class in which the field invariant holds, then field[i] has annotation[i].
+ *
+ * <p>This annotation is inherited so, if a superclass is annotation with @FieldInvariant, it's
```
mernst: it's => its.  Also on next line.

framework/src/org/checkerframework/framework/qual/FieldInvariant.java
```diff
@@ -0,0 +1,37 @@
...
+import java.lang.annotation.Annotation;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import javax.lang.model.element.TypeElement;
+import org.checkerframework.framework.type.AnnotatedTypeFactory;
+/**
+ * In the class in which the field invariant holds, then field[i] has annotation[i].
```
mernst: State the overall purpose.  State that the annotations provided here are strict subtypes of those in the superclass.  Give an example of how to write an annotation.

framework/src/org/checkerframework/framework/qual/FieldInvariant.java
```diff
@@ -0,0 +1,37 @@
...
+ * system can implement its own field invariant annotation and override {@link
+ * AnnotatedTypeFactory#getFieldInvariantDeclarationAnnotations()} and {@link
+ * AnnotatedTypeFactory#getFieldInvariants(TypeElement)}. See {@link
+ * org.checkerframework.checker.index.qual.MinLenFieldInvariant} for example.
+ */
+@Target({ElementType.TYPE})
+@Inherited
+@Retention(RetentionPolicy.RUNTIME)
+public @interface FieldInvariant {
+    /* The field which has the qualifier.  The field must
```
mernst: which => that

framework/src/org/checkerframework/framework/qual/FieldInvariant.java
```diff
@@ -0,0 +1,37 @@
...
+     * be final and declared in a superclass of the class on which the field invariant applies.*/
+    String[] field();
+
+    /**
+     * The qualifier on field. Must be a subtype of the annotation on the declaration of the field.
```
mernst: field => the field.
Also must be a subtype of any `@FieldAnnotation` on a superclass.  Did you omit that because it didn't seem relevant?

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1285,10 +1287,123 @@ protected void postDirectSuperTypes(
...
     public void postAsMemberOf(
             AnnotatedTypeMirror type, AnnotatedTypeMirror owner, Element element) {
+        if (element.getKind() == ElementKind.FIELD) {
+            addAnnotationFromFieldInvariant(type, owner, (VariableElement) element);
+        }
         addComputedTypeAnnotations(element, type);
     }
 
     /**
+     * Adds the qualifier specified by a field invariant for {@code field} to {@code type}
```
mernst: End sentence with period.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1285,10 +1287,123 @@ protected void postDirectSuperTypes(
...
+     *
+     * @param type annotated type to which the annotation is added
+     * @param accessedVia the annotated type of the receiver of the accessing tree. (Only used to
+     *     get the type element of the underling type.)
+     * @param field element of the field
```
mernst: "of" => "for" ?

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1285,10 +1287,123 @@ protected void postDirectSuperTypes(
...
+            return;
+        }
+        List<AnnotationMirror> invariantAnnos = invariants.getQualifiersFor(field.getSimpleName());
+        type.replaceAnnotations(invariantAnnos);
+    }
+
+    /**
+     * Returns the field invariants for the given class.
+     *
+     * <p>Subclass may override this to implement their own field invariant annotations if {@link
```
mernst: Really, subclasses may implement their own field invariant annotations, and in that case they need to override this too.

smillst:
@mernst Should the documentation for this go in the "Advanced type system features" section of the manual?  Or someplace else?

mernst:
Yes.  It should go near the "unused fields" documentation, which is similar in spirit.

Thanks for improving the documentation -- we need it for the release, and reviewing is more difficult without it.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
smillst: I numbered the list by hand so that I could refer to the numbers in the method body.

smillst:
Note to self: Remaining tasks:
- [ ] Add manual section to "Advanced type system features" near "unused fields"
- [ ] Reference section in FieldInvariant class comment
- [ ] Reference section in BaseTypeVisitor#checkFieldInvariant
- [ ] Add Class Javadoc to MinLenFieldInvariant

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
smillst: I did this at with a reference to the manual section at the bottom of the Javadoc.

mernst: That doesn't help a person to understand this comment; the person doesn't know that in order to define this particular term, he or she needs to read the manual section.
I also feel that a term like "refined field invariant" would be better.  `@FieldInvariant` is a fine name for the annotation, but "field invariant" doesn't suggest the fact that it is different than in the parent.  If someone mentioned a "field invariant" to me, I would think of just the annotation on the field.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
mernst: You can specify numbers in `<li>` items.  But even if you retain the current formatting, please add a `<!-- ... -->` comment telling future maintainers to update the body if they update the list.  Otherwise it will be too easy for someone editing this comment not to realize the damage they are doing.

smillst: Oh, you are suggesting `<li value="1">`.  I didn't know that was possible.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
smillst: "refined field invariant" is confusing to me.  What's being refined, the field or the invariant? It's especially confusing since it's not mentioned anywhere else in describing "field invariants".  Though maybe when you review my documentation improvements you'll suggest some other places that it might go.

mernst: OK, let's keep thinking about good terminology, and reviewing all the documentation at once will help us with that.  Thanks.

javacutil/src/org/checkerframework/javacutil/ElementUtils.java
```diff
@@ -247,6 +247,18 @@ public static VariableElement findFieldInType(TypeElement type, String name) {
         return null;
     }
 
+    /**
+     * Returns the elements of the fields whose simple names are {@code names} and are declared in
```
mernst: Should this be only non-private fields?
Same question for the routines below.

mernst:
One more comment:

The examples in the documentation are written with `qualifier` before `field`:

`@FieldInvariant(qualifier = NonNull.class, field = "field")`

but in the definition of @FieldInvariant, `field` comes before `qualifier`.  Please make them consistent.

javacutil/src/org/checkerframework/javacutil/ElementUtils.java
smillst: Do you mean should this method only return non-private fields?  Or are you asking if the method as implemented only returns non-private fields?

mernst: I think it should only return non-private fields, because private fields don't exist in the subclass so it doesn't seem to make sense for the subclass to specify their behavior.  Does that sound right to you?

smillst: Yes, I agree.  I'll make the change.

smillst: Or rather I agree that this method should only return non-private fields, but the method below, should return private and non-private fields.

PR #1199, opened 2017-03-29
Fix #1198:

### LINKED ISSUE #1198: IntRange on new array lengths causes Value Checker to crash
Consider the following code:

```java

class IntRangeArrayInteraction {
    void test(@IntRange( to = 1, from = Integer.MAX_VALUE ) int length) {
	int[] a = new int[length];
    }
}
```

When the Value Checker is run on this code, it attempts to assign an `@ArrayLen` annotation to `a` in `ValueAnnotatedTypeFactory#ValueTreeAnnotator#handleDimensions`. When it does this, it tries to call `getIntValues` on the annotation on `length`, which fails because `length`'s annotation is a range.

JasonMrX:
Convert @IntRange to @IntVal before creating the corresponding @ArrayLen.
Handle @BottomVal annotated dimension by creating a @BottomVal.
Don't create annotation if the dimension is not @IntRange, @IntVal or @BottomVal.

### DIFF
--- a/framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
+++ b/framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
@@ -547,8 +547,14 @@ public class ValueAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
          * Recursive method to handle array initializations. Recursively descends the initializer to
          * find each dimension's size and create the appropriate annotation for it.
          *
+         * <p>If the annotation of the dimension is {@code @IntVal}, create an {@code @ArrayLen}
+         * with the same set of possible values. If the annotation is {@code @IntRange}, convert it
+         * to {@code @IntVal} before creating the {@code @ArrayLen} by the same method as above. If
+         * the annotation is {@code @BottomVal}, create an {@code @BottomVal} instead. In other
+         * cases, no annotations would be created.
+         *
          * @param dimensions a list of ExpressionTrees where each ExpressionTree is a specifier of
-         *     the size of that dimension (should be an IntVal)
+         *     the size of that dimension
          * @param type the AnnotatedTypeMirror of the array
          */
         private void handleDimensions(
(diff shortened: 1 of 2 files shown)


JasonMrX:
This is a quick fix for the crashing issue. I am still thinking of how to adapt `@ArrayLen` to handle range as we do for `@IntRange`

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -547,8 +547,14 @@ public Void visitNewArray(NewArrayTree tree, AnnotatedTypeMirror type) {
          * Recursive method to handle array initializations. Recursively descends the initializer to
          * find each dimension's size and create the appropriate annotation for it.
          *
+         * <p>If the annotation of the dimension is {@code @IntVal}, create an {@code @ArrayLen}
+         * with the same set of possible values. If the annotation is {@code @IntRange}, convert it
+         * to {@code @IntVal} before creating the {@code @ArrayLen} by the same method as above. If
+         * the annotation is {@code @BottomVal}, create an {@code @BottomVal} instead. In other
+         * cases, no annotations would be created.
```
wmdietl: `would be` -> `are`

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -558,10 +564,12 @@ private void handleDimensions(
                         dimensions.subList(1, dimensions.size()),
                         (AnnotatedArrayType) type.getComponentType());
             }
-
             AnnotationMirror dimType =
                     getAnnotatedType(dimensions.get(0)).getAnnotationInHierarchy(UNKNOWNVAL);
-            if (!AnnotationUtils.areSameIgnoringValues(dimType, UNKNOWNVAL)) {
+            if (AnnotationUtils.areSameByClass(dimType, IntRange.class)) {
+                dimType = convertIntRangeToIntVal(dimType);
```
wmdietl: Wouldn't it be easier to use `ValueCheckerUtils.getValuesFromRange(range, Long.class);` instead of creating a new `IntVal` annotation? That is, move out the `List<Long>` and either initialize it from the IntVal or the IntRange and then have the common logic.

mernst:
As a small comment, please use a descriptive name for your pull request and commits.  Text like "fixes #1198" is good to include, but it can go in the descriptive text.  There is some advice about pull requests at http://homes.cs.washington.edu/~mernst/advice/github-pull-request.html .

JasonMrX:
I see. Will do in the future.

PR #1200, opened 2017-03-29
Make the check-nullness target actually process all source files.

wmdietlGC:
Currently, `ant check-nullness` only checks the JDK annotations instead of all source files.
We already type check the JDK annotations when building the annotated JDK.
This PR type-checks the source code itself, which isn't type-checked yet.
Unfortunately, when trying to do both, either one didn't work. This PR checks the parts that we haven't checked yet.
It also uses Java 8, because the source code uses Java 8 type annotations.

We should add 'ant check-nullness' to the travis tests and ensure that there are no errors. There are currently many warnings, but only two errors, of the form:

```
     [java] error: Upper bounds should never contain declarations.
     [java]   type=@Initialized @NonNull Annotation
     [java]   Compilation unit: .../checker-framework/checker/src/org/checkerframework/checker/guieffect/GuiEffectTypeFactory.java
     [java]   Exception: java.lang.Throwable; Stack trace: org.checkerframework.framework.source.SourceChecker.errorAbort(SourceChecker.java:722)
     [java]   org.checkerframework.javacutil.ErrorReporter.errorAbort(ErrorReporter.java:25)
     [java]   org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedTypeVariable.setUpperBound(AnnotatedTypeMirror.java:1624)
     [java]   org.checkerframework.framework.type.AsSuperVisitor.visitDeclared_Typevar(AsSuperVisitor.java:373)
     [java]   org.checkerframework.framework.type.AsSuperVisitor.visitDeclared_Typevar(AsSuperVisitor.java:29)
     [java]   org.checkerframework.framework.util.AtmCombo.accept(AtmCombo.java:338)
     [java]   org.checkerframework.framework.type.visitor.AbstractAtmComboVisitor.visit(AbstractAtmComboVisitor.java:59)
     [java]   org.checkerframework.framework.type.AsSuperVisitor.visit(AsSuperVisitor.java:86)
```

### DIFF
--- a/checker/build.xml
+++ b/checker/build.xml
@@ -777,6 +777,7 @@
         <antcall target="check-compilermsgs"/>
         <antcall target="check-purity"/>
 
+        <antcall target="check-nullness"/>
     </target>
 
     <target name="aggregate-tests" depends="jar,build-tests"
--- a/checker/build.xml
+++ b/checker/build.xml
@@ -1237,7 +1237,6 @@ So, use our own archived version.
                 <fileset dir="${src}">
                     <include name="**/*.java"/>
                     <exclude name="${checker.jdk8orhigher.sources}"/>
-                    <exclude name=""/>
                 </fileset>
                 <fileset dir="${javacutil.loc}/${src}">
                     <include name="**/*.java"/>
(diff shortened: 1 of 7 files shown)


checker/build.xml
```diff
@@ -777,6 +777,7 @@
         <antcall target="check-compilermsgs"/>
         <antcall target="check-purity"/>
 
+        <antcall target="check-nullness"/>
```
wmdietl: This takes a considerable time and generates >1200 warnings. I wanted to add it to make sure there are no exceptions thrown. However, as we already run check-compilermsgs and check-purity, we might not need this.
Two options:
- disable just check-nullness again
- put all three check-xxx targets into a separate travis configuration

smillst: I vote for disabling the check-nullness target.

wmdietlGC:
Thanks for the review.
I agree with disabling check-nullness until we say it is a priority.

PR #1202, opened 2017-03-29
Make it easier to switch between building and downloading JDK on Travis

smillst:
I couldn't figure out a way to switch which Travis builds occur based on an environment variable.  (When downloading the jdk, "jdk" and "all-tests" needs to build, otherwise "junit" and "nonjunit" need to run.)  

Adds two different .travis.yml files
1. .travis-downloadjdk.yml tests download the jdk rather than building it.
2. .travis-buildjdk.yml tests build the jdk.

.travis.yml should link to which ever version is desired.


mernst:
All that duplicated code in the two files is a code smell for me.

I have a slightly different proposal:  We have just one `.travis.yml` file, and it runs all possible jobs.  However, `.travis-build.sh` has a hard-coded variable defined at the top of the file, to either "buildjdk" or "downloadjdk".  This causes it to skip some tasks (those jobs will trivially succeed, with very little CPU time).  The non-trivial jobs are the same as when spawning different sets of jobs as is done by the current pull request.

What do you think?

smillst:
That would mean 13 jobs in one build, though 4-6 would succeed quickly.  Seems like that would make waiting for the tests to pass rather painful.

What if we move the code in `script` and `after-script` to actual scripts that both yml file could use?

mernst:
Because 4-6 would pass quickly, they wouldn't affect total wait time.  Travis throttles based on CPU use, not based on number of jobs run.

I don't object on principle to factoring out the script and after-script sections, but that doesn't seem to address the problem with the matrix containing extra, fast jobs.

smillst:
>Because 4-6 would pass quickly, they wouldn't affect total wait time. Travis throttles based on CPU use, not based on number of jobs run.

It takes about a minute from the time one job completes until the next queued job starts.  It takes about 30 seconds for a job to start downloading the docker image.  (Presumably, a job that isn't needed would exit before the docker image is download.)   So that's 6-8 minutes of overhead.  In my opinion it is worth some duplicated code to avoid that overhead. 

>I don't object on principle to factoring out the script and after-script sections, but that doesn't seem to address the problem with the matrix containing extra, fast jobs.

The point of factoring out the script would be to address your concern about code duplication without running 13 jobs.

mernst:
The 6-8 minutes of overhead would be distributed among the 5 processors that Travis allocates, so it's 1-2 minutes of wall clock time.  Nonetheless, I agree it is a delay.

I was confused by your proposal, but now think I see:  you are proposing to retain the two .travis.yml files (which was the duplication that I initially objected to, and managing symbolic links sounded confusing), but reduce the amount of duplication by factoring out parts of them into scripts.  I guess that's OK, and it will save the time as discussed.

smillst:
Would you prefer comment/uncomment lines in the yml file to symbolic links?  Something like this:

```
env:
  global:
  - BUILDJDK=downloadjdk
#  - BUILDJDK=buildjdk
  matrix:
## Jobs used when downloading jdk
  - JDKVER=jdk7 GROUP=all-tests
  - JDKVER=jdk7 GROUP=jdk.jar
  - JDKVER=jdk8 GROUP=all-tests
  - JDKVER=jdk8 GROUP=jdk.jar
## Jobs used when building jdk
#  - JDKVER=jdk7 GROUP=junit
#  - JDKVER=jdk7 GROUP=nonjunit
#  - JDKVER=jdk7 GROUP=demos
#  - JDKVER=jdk8 GROUP=junit
#  - JDKVER=jdk8 GROUP=nonjunit
#  - JDKVER=jdk8 GROUP=demos
## Job used when building or downloading jdk
  - JDKVER=jdk7 GROUP=downstream
  - JDKVER=jdk8 GROUP=downstream
  - JDKVER=jdkany GROUP=misc

```

mernst:
Symbolic links is OK, so long as it's clearly documented.  That leads to smaller diffs.

mernst:
Really, whichever way you find simpler and clearer.

smillst:
I prefer commenting line in the yml file.  I pushed those changes.

PR #1203, opened 2017-03-30
Index Checker: Fixes bug in SameLen subtyping rule, and several bugs that were masked by that bug

kelloggm:
The SameLen subtyping rule was too broad. This change restricts it. This fixes kelloggm#133.

I discovered two other bugs that this was hiding, both false positives. Both had tests that failed already, so no new tests are added for either.

The first was SameLen's version of kelloggm#58: when `strengthenEqualTo` is called, `SameLenTranfer` failed to descend into assignments. This is fixed with a call to split assignments, which necessitated some surgery in the `SameLenAnnotatedTypeFactory` to make `createCombinedSameLen` more general; it now takes two corresponding lists instead of two corresponding pairs. I left the pairwise version as well to minimize code churn, but changed it to call the new version.

The second bug occurred when a new array is being typed by the `SameLenTreeAnnotator`. If the new array is created with another array's length as its length (e.g. `int [] a = new int[b.length]`), then the new array's type is the same as the other array (so in the example `new int[b.length]` would have the same type as `b`). The problem was that if `b`'s type was something like `@SameLen("c")`, then the new array is not `@SameLen("b")`. The fix adds the array's name to the type if it is not already there.

### DIFF
--- a/checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
+++ b/checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
@@ -105,7 +105,6 @@ public class SameLenAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
         newValues.addAll(a1Names);
         newValues.addAll(a2Names);
         String[] names = newValues.toArray(new String[newValues.size()]);
-        Arrays.sort(names);
         return createSameLen(names);
     }
 
@@ -293,7 +292,6 @@ public class SameLenAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
                         if (!arrayAnnoArrays.contains(rec.toString())) {
                             arrayAnnoArrays.add(rec.toString());
                             String[] newArrayAnnoArrays = arrayAnnoArrays.toArray(new String[0]);
-                            Arrays.sort(newArrayAnnoArrays);
                             arrayAnno = createSameLen(newArrayAnnoArrays);
                         }
                     }
@@ -307,10 +305,15 @@ public class SameLenAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
     /** Creates a @SameLen annotation whose values are the given strings. */
     public AnnotationMirror createSameLen(String... val) {
         AnnotationBuilder builder = new AnnotationBuilder(processingEnv, SameLen.class);
+        Arrays.sort(val);
         builder.setValue("value", val);
         return builder.build();
     }
 
+    /**
+     * Find all the arrays that are members of the SameLen annotation associated with the array
+     * named in arrayExpression along the current path.
+     */
     public List<String> getSameLensFromString(
             String arrayExpression, Tree tree, TreePath currentPath) {
         AnnotationMirror sameLenAnno = null;
--- a/checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
+++ b/checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
@@ -110,37 +110,46 @@ public class SameLenAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
     }
 
     /**
+     * Combines the given arrays and annotations into a single SameLen annotation. See {@link
+     * #createCombinedSameLen(List, List)}.
+     */
+    public AnnotationMirror createCombinedSameLen(
+            Receiver rec1, Receiver rec2, AnnotationMirror a1, AnnotationMirror a2) {
+        List<Receiver> receivers = new ArrayList<>();
+        receivers.add(rec1);
+        receivers.add(rec2);
+        List<AnnotationMirror> annos = new ArrayList<>();
+        annos.add(a1);
+        annos.add(a2);
+        return createCombinedSameLen(receivers, annos);
+    }
+
+    /**
      * For the use of the transfer function; generates a SameLen that includes a and b, as well as
      * everything in sl1 and sl2, if they are SameLen annotations.
      *
-     * @param aRec receiver representing the first array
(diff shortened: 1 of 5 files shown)


checker/src/org/checkerframework/checker/index/samelen/SameLenTransfer.java
```diff
@@ -149,27 +151,38 @@ private boolean isArrayLengthAccess(Node node) {
...
+        AnnotationMirror combinedSameLen = aTypeFactory.createCombinedSameLen(receivers, annos);
 
         propagateCombinedSameLen(combinedSameLen, left, store);
     }
 
     /** Return n's annotation from the SameLen hierarchy. */
     AnnotationMirror getAnno(Node n) {
-        CFValue cfValue = analysis.getValue(n);
-        if (cfValue == null) {
+        if (n.isLValue()) {
```
smillst: Why?

checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
```diff
@@ -270,15 +279,25 @@ public Void visitNewArray(NewArrayTree node, AnnotatedTypeMirror type) {
...
                         arrayAnno = createSameLen(rec.toString());
+                    } else if (AnnotationUtils.areSameByClass(arrayAnno, SameLen.class)) {
+                        // Ensure that the array whose length is actually being used is part of the
+                        // annotation. If not, add it.
+                        List<String> arrayAnnoArrays =
+                                getValueOfAnnotationWithStringArgument(arrayAnno);
+                        if (!arrayAnnoArrays.contains(rec.toString())) {
+                            arrayAnnoArrays.add(rec.toString());
+                            String[] newArrayAnnoArrays = arrayAnnoArrays.toArray(new String[0]);
+                            Arrays.sort(newArrayAnnoArrays);
```
kelloggm: Probably yes. `Arrays.sort`  is also called by `getCombinedSameLen`, which is why I did it here. I'll remove it from both places and add it to `createSameLen` instead.

checker/src/org/checkerframework/checker/index/samelen/SameLenTransfer.java
kelloggm: `analysis.getValue` fails if called on an lvalue. However, I need the annotations from lvalues when they are arrays in assignments. Consider this code:
`if ( (a = b) == c) {...}`, where `a`, `b`, and `c` are all arrays, and `a` has type `@SameLen("d")`. Afterwards, all three should have the type `@SameLen({"a", "b", "c", "d"})`, but in order to accomplish this we need to get the type of `a`, which is an lvalue.

kelloggm: I'll add a comment in the code that explains this.

checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
```diff
@@ -231,10 +240,10 @@ public boolean isSubtype(AnnotationMirror subAnno, AnnotationMirror superAnno) {
                 return AnnotationUtils.areSameByClass(superAnno, PolySameLen.class);
             } else if (AnnotationUtils.hasElementValue(subAnno, "value")
                     && AnnotationUtils.hasElementValue(superAnno, "value")) {
-                List<String> a1Val = getValueOfAnnotationWithStringArgument(subAnno);
-                List<String> a2Val = getValueOfAnnotationWithStringArgument(superAnno);
+                List<String> subArrays = getValueOfAnnotationWithStringArgument(subAnno);
+                List<String> superArrays = getValueOfAnnotationWithStringArgument(superAnno);
 
-                if (overlap(a1Val, a2Val)) {
+                if (subArrays.containsAll(superArrays)) {
```
kelloggm: I've changed the image in the manual to include three types that aren't bottom or top: `@SameLen("a")` and `@SameLen("b")` at the same level, both of which are subtypes of `@SameLen({"a", "b"})`.

kelloggm:
@smillst, I believe I've addressed all your feedback. I'd appreciate it if you'd take another look.

checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
smillst: The code says that `@SameLen({"a", "b"})` is a subtype of `@SameLen("a")`, so I think the image is wrong.

checker/tests/index/SameLenFormalParameter2.java
```diff
@@ -0,0 +1,13 @@
+import org.checkerframework.checker.index.qual.*;
+
+//@skip-test
```
smillst: Don't skip!

docs/manual/index-checker.tex
```diff
@@ -322,7 +322,8 @@ \section{Array minimum lengths\label{index-minlen}}
 \end{center}
   \caption{The two type hierarchies for array types used by the Index
     Checker.  On the left is a type system for minimum array lengths.  On
-    the right is a type system for arrays of equal length.  Qualifiers
+    the right is a type system for arrays of equal length ("a" and "b" are
+    assumed to be valid arrays).  Qualifiers
```
smillst: What's a valid array?  I think this could just be "assuming "a" and "b" are arrays"

checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
kelloggm: That's definitely true. My image is backwards! I'll fix it.

kelloggm:
This should be ready to be reviewed again @smillst

PR #1206, opened 2017-03-30
Fix #1055 by adding type parameter elements when parsing stub files.

### LINKED ISSUE #1055: Incorrect wildcard bound determination with stub files
Take the following simple class:

```
class NonN<T> {}
```

Compile with normal javac.
Then add a file `NonN.astub` that specifies the upper bound is non-null:

```

class NonN<T extends @NonNull Object> {}
```

Then add this main class:

```
class Wildcards {
  NonN<?> f = new NonN<Object>();

  class LocalNonN<T extends Object> {}
  LocalNonN<?> g = new LocalNonN<Object>();
}
```

Compile with `.../javac -processor nullness -Astubs=NonN.astub Wildcards.java` and you get:

```
Wildcards.java:2: error: [type.argument.type.incompatible] incompatible types in type argument.
  NonN<?> f = new NonN<Object>();
       ^
  found   : ? extends @Initialized @Nullable Object
1 error
```

The behavior for the wildcard is different for `NonN` and `LocalNonN`, which is wrong.
I was able to reproduce it with this stub-file setup, but maybe there is a simpler way to reproduce this.

wmdietlGC:
I still need to add the test case (unless it's already somewhere...) but want to see whether anything breaks.

### DIFF
--- a/checker/jtreg/stubs/wildcards/NonN.astub
+++ b/checker/jtreg/stubs/wildcards/NonN.astub
@@ -1,0 +1,4 @@
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+class NonN<T extends @NonNull Object> {
+}
--- a/checker/jtreg/stubs/wildcards/NonN.java
+++ b/checker/jtreg/stubs/wildcards/NonN.java
@@ -1,0 +1,1 @@
+class NonN<T> {}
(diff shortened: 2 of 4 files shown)


mernst:
Looks fine to me, but you requested Suzanne's feedback.  She's on vacation but will be back at work on Monday.

wmdietlGC:
I had assigned Suzanne because we discussed the issue yesterday. This is a low-risk fix that we could still merge.

mernst:
I agree; that's what it looked like when I reviewed it yesterday.

PR #1208, opened 2017-03-31
Refine IntRange annotations in Value Checker

kelloggm:
Extends typetools#1163 by adding support for `@IntRange` annotations to the refinement rules in `ValueTransfer`.


kelloggm:
@JasonMrX you should probably be aware of this too

framework/src/org/checkerframework/common/value/util/Range.java
```diff
@@ -767,6 +767,38 @@ public Range refineEqualTo(Range right) {
...
+     * Refines this range to reflect that no value in it can be equal to a value in the given range.
+     * This is used for calculating the control-flow-refined result of the != operator. For example:
+     *
+     * <pre>
+     * <code>
+     *     {@literal @}IntRange(from = 0, to = 10) int a;
+     *     {@literal @}IntRange(from = 3, to = 15) int b;
+     *     ...
+     *     if (a != b) {
+     *         // range of <i>a</i> is now refined to [0, 2] because a value in range [3, 10]
```
JasonMrX: I don't think this is the correct logic. `a` should not be refined to `[0, 2]` because even if `b` has range of `[3, 15]` and `a` is not equal to `b`, `a` still has a chance to equal to `3` (or other values) because `b` is not guaranteed to be `3`. 

This is the reason why I didn't implement this api. An equality check can't really provide much information for range refinement. Do you agree?

kelloggm: You're correct; this is my mistake. I was wondering why you'd implemented the other APIs and not that one, but I suppose I didn't think it through. I'll remove this part and add a test to ensure that it isn't reintroduced.

On an equality check we certainly can gain information, and your API gives that information correctly. But on an `!=` we shouldn't do a refinement.

Thanks for pointing this out!

JasonMrX: Ah, my mistake. It's `!=` that doesn't provide too much information. And actually I should have leave a comment in the `Range` class for future reference (like in you case..) 

Thanks for working on this. It's kind of a remaining work of my part. Other than this issue, this PR looks great to me.

kelloggm: Haha no worries! I was just working on the other refinement stuff for the Value Checker, and so this seemed like a natural extension. Plus it'll be really helpful for some of the false positives we're seeing commonly in the Index Checker.

I also didn't actually do much besides call out to your code and write tests. Thanks for writing such a good Range library.

JasonMrX: My pleasure.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -855,6 +852,78 @@ private Range calculateRangeUnaryOp(
...
+                elseRightRange = rightRange.refineLessThan(leftRange);
+                break;
+            case NOT_EQUAL:
+                thenRightRange = rightRange; // Not equals gives almost no information.
+                thenLeftRange = leftRange;
+                elseRightRange = rightRange.refineEqualTo(leftRange);
+                elseLeftRange = leftRange.refineEqualTo(rightRange);
+                break;
+            default:
+                throw new UnsupportedOperationException();
```
wmdietl: Can you replace this (here and elsewhere in this class) and the `throw new RuntimeException("Unrecognized conditional operator " + op);` further down, with the proper `ErrorReporter.errorAbort`... that will result in a nicer error message in case this ever happens and isn't much longer to write.

kelloggm: Sure. I've replaced it everywhere in the class.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -855,6 +856,84 @@ private Range calculateRangeUnaryOp(
...
+        final Range thenRightRange;
+        final Range thenLeftRange;
+        final Range elseRightRange;
+        final Range elseLeftRange;
+
+        switch (op) {
+            case EQUAL:
+                thenRightRange = rightRange.refineEqualTo(leftRange);
+                thenLeftRange = leftRange.refineEqualTo(rightRange);
+                elseRightRange = rightRange; // Not equals gives almost no information.
```
kelloggm: I've added code to handle not-equals in this case, and appropriate tests.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -855,6 +856,84 @@ private Range calculateRangeUnaryOp(
...
+
+        final Range thenRightRange;
+        final Range thenLeftRange;
+        final Range elseRightRange;
+        final Range elseLeftRange;
+
+        switch (op) {
+            case EQUAL:
+                thenRightRange = rightRange.refineEqualTo(leftRange);
+                thenLeftRange = leftRange.refineEqualTo(rightRange);
```
kelloggm: It's always the same. I've fixed this, too.

kelloggm:
@smillst I've addressed your feedback. I'd appreciate it if you could take another look!

build-styleguide.xml
```diff
@@ -139,7 +139,6 @@
     <exec executable="git"
           dir="${checker.loc}/bin-devel/.run-google-java-format">
       <arg value="pull"/>
-      <arg value="-q"/>
```
mernst: What is the reason for this change?

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -442,7 +443,8 @@ private Range calculateRangeBinaryOp(
                     resultRange = leftRange.bitwiseXor(rightRange);
                     break;
                 default:
-                    throw new UnsupportedOperationException();
+                    ErrorReporter.errorAbort("ValueTransfer: unsupported operation: " + op);
+                    resultRange = null;
```
mernst: Is this code dead?  If so, please throw an exception with message "this can't happen".  Using an assignment will lead a reader to think that this statement can get executed.

Same comment elsewhere.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -855,6 +856,85 @@ private Range calculateRangeUnaryOp(
...
+                break;
+            case NOT_EQUAL:
+                thenRightRange = rightRange.refineNotEqualTo(leftRange);
+                thenLeftRange = leftRange.refineNotEqualTo(rightRange);
+                elseRightRange = rightRange.refineEqualTo(leftRange);
+                elseLeftRange = elseRightRange; // Equality only needs to be computed once.
+                break;
+            default:
+                ErrorReporter.errorAbort("ValueTransfer: unsupported operation: " + op);
+                // The next four lines are dead, but if they aren't here then
```
mernst: Throw an exception instead.  It is a shorter way to suppress a javac warning, and its message can take the place of the comment.j

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -1042,7 +1122,8 @@ private void createAnnotationFromResultsAndAddToStore(
                 }
                 return resultValues;
         }
-        throw new RuntimeException("Unrecognized conditional operator " + op);
+        ErrorReporter.errorAbort("ValueTransfer: unsupported operation: " + op);
+        return null; // dead code
```
mernst: This comment is helpful.  It's OK, or throwing an exception.

framework/src/org/checkerframework/common/value/util/Range.java
```diff
@@ -767,6 +767,39 @@ public Range refineEqualTo(Range right) {
...
 
     /**
+     * Refines this range to reflect that some value in it must not be equal to a value in the given
+     * range. This only changes the range if the given range (right) contains exactly one integer,
+     * and that integer is one of the bounds of this range. This is used for calculating the
+     * control-flow-refined result of the != operator. For example:
+     *
+     * <pre>
+     * <code>
+     *     {@literal @}IntRange(from = 0, to = 10) int a;
```
mernst: Need to remove a leading space on this line and the next, to get alignment right in the generated Javadoc.  (Take a look!)

build-styleguide.xml
smillst: Sorry I committed that by mistake!

framework/src/org/checkerframework/common/value/util/Range.java
kelloggm: This was an issue on every one of the refine* methods in `Range.java`, so I've fixed it for all of them.

PR #1210, opened 2017-03-31
Standardize definitions of bottom qualifiers

mernst:


docs/manual/generics.tex
```diff
@@ -424,6 +424,21 @@ \subsection{Method type argument inference and type qualifiers\label{infer-metho
...
+value, dead code, and some erroneous situations.  A programmer should
+rarely write the bottom type.
+
+One use is on a lower bound, to indicate that any type qualifier is
+permitted.  A lower-bounded wildcard indicates that a consumer method can
+accept a collection containing any Java type above some Java type, and you
+can add the bottom type qualifier as well:
+
+\begin{Verbatim}
+public static void addNumbers(List<? super @PolySigned Integer> list) { ... }
```
wmdietl: Shouldn't the `@PolySigned` be `@SignednessBottom`? The former doesn't exist.

mernst: You are right.  Thanks for the catch!  I will update it.

PR #1211, opened 2017-04-01
Annotations added for some JAXP classes

nikshinde1996:
I have added annotation for few related JAXP classes. Please review the PR and suggest if some changes are required.

Thanks,


nikshinde1996:
It would be helpful if mentors review this PR and suggest me necessary changes if needed .
Whenever a document is first made available via DOM, the Text, CharacterData nodes are important in arrangement for data , due to this I thought of working on this PR.

Thanks,

checker/jdk/nullness/src/org/w3c/dom/Attr.java
```diff
@@ -0,0 +1,307 @@
...
+     * <br>On setting, this creates a <code>Text</code> node with the unparsed
+     * contents of the string, i.e. any characters that an XML processor
+     * would recognize as markup are instead treated as literal text. See
+     * also the method <code>Element.setAttribute()</code>.
+     * <br> Some specialized implementations, such as some [<a href='http://www.w3.org/TR/2003/REC-SVG11-20030114/'>SVG 1.1</a>]
+     * implementations, may do normalization automatically, even after
+     * mutation; in such case, the value on retrieval may differ from the
+     * value on setting.
+     */
+    @Pure public @Nullable String getValue();
```
mernst: The documentation does not say that this method can return null.
What is the justification for the `@Nullable` annotation?

checker/jdk/nullness/src/org/w3c/dom/CharacterData.java
```diff
@@ -0,0 +1,185 @@
...
+     * implementation limits may mean that the entirety of a node's data may
+     * not fit into a single <code>DOMString</code>. In such cases, the user
+     * may call <code>substringData</code> to retrieve the data in
+     * appropriately sized pieces.
+     * @exception DOMException
+     *   DOMSTRING_SIZE_ERR: Raised when it would return more characters than
+     *   fit in a <code>DOMString</code> variable on the implementation
+     *   platform.
+     */
+    @Pure public @Nullable String getData()
```
mernst: What justifies this `@Nullable` annotation?

checker/jdk/nullness/src/org/w3c/dom/CharacterData.java
```diff
@@ -0,0 +1,185 @@
...
+     * implementation may not put arbitrary limits on the amount of data
+     * that may be stored in a <code>CharacterData</code> node. However,
+     * implementation limits may mean that the entirety of a node's data may
+     * not fit into a single <code>DOMString</code>. In such cases, the user
+     * may call <code>substringData</code> to retrieve the data in
+     * appropriately sized pieces.
+     * @exception DOMException
+     *   NO_MODIFICATION_ALLOWED_ERR: Raised when the node is readonly.
+     */
+    public void setData(@Nullable String data)
```
mernst: I don't see documentation that says the interface permits null values to be passed in.

checker/jdk/nullness/src/org/w3c/dom/CharacterData.java
```diff
@@ -0,0 +1,185 @@
...
+     * @param offset The character offset at which to insert.
+     * @param arg The <code>DOMString</code> to insert.
+     * @exception DOMException
+     *   INDEX_SIZE_ERR: Raised if the specified <code>offset</code> is
+     *   negative or greater than the number of 16-bit units in
+     *   <code>data</code>.
+     *   <br>NO_MODIFICATION_ALLOWED_ERR: Raised if this node is readonly.
+     */
+    public void insertData(int offset,
+                           @Nullable String arg)
```
mernst: Why is this `@Nullable`?

checker/jdk/nullness/src/org/w3c/dom/Text.java
```diff
@@ -0,0 +1,190 @@
...
+     *                              |
+     *                           /-----\
+     *                           | foo |
+     *                           \-----/
+     * </pre>
+     * <em>Figure: barTextNode.wholeText value is "barfoo"</em>
+     *
+     * @since DOM Level 3
+     */
+    @Pure public @Nullable String getWholeText();
```
mernst: Why is this `@Nullable`?

checker/jdk/nullness/src/org/w3c/dom/Text.java
```diff
@@ -0,0 +1,190 @@
...
+     * contains "bar" fails, because the <code>EntityReference</code> node
+     * "ent" contains an <code>Element</code> node which cannot be removed.</p>
+     * @param content The content of the replacing <code>Text</code> node.
+     * @return The <code>Text</code> node created with the specified content.
+     * @exception DOMException
+     *   NO_MODIFICATION_ALLOWED_ERR: Raised if one of the <code>Text</code>
+     *   nodes being replaced is readonly.
+     * @since DOM Level 3
+     */
+    public Text replaceWholeText(String content)
```
mernst: The return value should be annotated as `@Nullable`, per line 139.

checker/jdk/nullness/src/org/w3c/dom/CharacterData.java
nikshinde1996: This function returns DOMString which can be null, Also this class is derived class of Node which allows null value.

checker/jdk/nullness/src/org/w3c/dom/CharacterData.java
nikshinde1996: CharacterData is subclass of Node, which allows null value.

checker/jdk/nullness/src/org/w3c/dom/CharacterData.java
nikshinde1996: CharacterData is sub class of Node, both Node value and  DOMString can be null. parameter in appendData function should be annotated.

checker/jdk/nullness/src/org/w3c/dom/Text.java
nikshinde1996: I agree, will update this.
Thanks.

checker/jdk/nullness/src/org/w3c/dom/Text.java
nikshinde1996: The node value and DOMString can be null. Text class is derived from parent Node class.

checker/jdk/nullness/src/org/w3c/dom/Attr.java
mernst: Can you quote the exact words in the documentation that justify this?  I don't see it.
There are only 4 occurrences of `null` in the Javadoc that you linked to, and the only possibly-relevant one is
> Thus, the Node attributes parentNode, previousSibling, and nextSibling have a null value for Attr objects.
Even if a Node's fields may be null, that doesn't say anything about an Attr object's fields.
So, you need to connect the dots here.

checker/jdk/nullness/src/org/w3c/dom/CharacterData.java
mernst: `Node` doesn't even have a `getData()` method, so your argument that the specification of `Node` justifies this doesn't make sense to me.

checker/jdk/nullness/src/org/w3c/dom/CharacterData.java
mernst: Node doesn't have a `setData` method, so saying that Null allows null is a non sequitur.  There might be an argument for the formal parameter being nullable, but that is not it.

checker/jdk/nullness/src/org/w3c/dom/Attr.java
nikshinde1996: Initially I used this  [reference](https://docs.oracle.com/javase/7/docs/api/org/w3c/dom/Node.html)  which  mentioned the individual interfaces and possible values for attributes.  Line 209 says that function returns the value of attribute as String.

mernst: You gave a link to an HTML document.  What does line 209 mean in that document?  Can you give a direct link to the relevant part?  I asked for the exact words that justified your decision, and you didn't give the exact words.

If the value of an attribute is returned as a String, that could mean that for null the string "null" is returned, or it could mean that the null value is returned.  Or, it could mean that the client shouldn't call getValue() on a null attribute.  How do you know which one is the intended specification?

Furthermore, as noted before, even if an attribute can be null in a Node, that is unrelated to the specification of Attr's getValue() routine.

nikshinde1996: I apologize for confusing explanation.  The first tabular data in this [reference](https://docs.oracle.com/javase/7/docs/api/org/w3c/dom/Node.html) mentioned about null attributes and other parameters for all interfaces in JAXP. Refer Attr row in table.   

Now line 209 in Attr.java file   
```
 208  +    /**
 209  +     * On retrieval, the value of the attribute is returned as a string.
 210  +     *
``` 
This made me to annotate the return type as `@Nullable`

mernst: Thanks for the clarification!
But, that line doesn't say anything about the treatment of null values, and does not say that `null` can be returned.  Therefore, I don't think we can rely on that sentence to write a `@Nullable` annotation.

checker/jdk/nullness/src/org/w3c/dom/CharacterData.java
nikshinde1996: I have misinterpreted Node.value with CharacterData.data . The oracle reference have mentioned that Node.value can be null, but there is no documentation available for CharacterData.data .

The answer to question `Can CData in XML have null value ?`  will be helpful to decide about Nullable property. 
Please suggest me about CData. Thanks

mernst: I don't know any more about this API than you do, so I cannot make a suggestion about CData beyond reading the documentation.

checker/jdk/nullness/src/org/w3c/dom/Text.java
```diff
@@ -0,0 +1,190 @@
...
+     * length of this node, the new node has no data.
+     * @param offset The 16-bit unit offset at which to split, starting from
+     *   <code>0</code>.
+     * @return The new node, of the same type as this node.
+     * @exception DOMException
+     *   INDEX_SIZE_ERR: Raised if the specified offset is negative or greater
+     *   than the number of 16-bit units in <code>data</code>.
+     *   <br>NO_MODIFICATION_ALLOWED_ERR: Raised if this node is readonly.
+     */
+    public @Nullable Text splitText(int offset)
```
mernst: Why is the return type `@Nullable`?  The specification says that the result is of the same type as this node, which implies it is not null.

nikshinde1996: How would split be if the offset value is terminal value ? i.e ` 0 ` or ` text.length() ` .

mernst: I don't know the answer, but I know that we definitely shouldn't make up annotations that are not justified by the documentation.  If the documentation is incorrect or incomplete, then submit a bug report to Oracle to learn the intended behavior.

nikshinde1996: For the two cases i.e when offset is terminal values. The DOMExcecption : INDEX_SIZE_ERR , DOMSTRING_SIZE_ERR occurs where I thought for NullPointerException.

I have submited bug report for the incomplete documentation in CharacterData and Text Class. Will update changes if required.

mernst:
Could you please check every other annotation you added and ensure that it is supported by the documentation?
Thanks!

nikshinde1996:
I have rechecked the document support for added annotations. Changes are fine. Will update Attr and CharacterData files if I get to know more from bug report.

Should I pull master into my branch and push updated branch ?

checker/jdk/nullness/src/org/w3c/dom/CharacterData.java
```diff
@@ -0,0 +1,185 @@
...
+     *   units to the end of the data are returned.
+     * @exception DOMException
+     *   INDEX_SIZE_ERR: Raised if the specified <code>offset</code> is
+     *   negative or greater than the number of 16-bit units in
+     *   <code>data</code>, or if the specified <code>count</code> is
+     *   negative.
+     *   <br>DOMSTRING_SIZE_ERR: Raised if the specified range of text does
+     *   not fit into a <code>DOMString</code>.
+     */
+    public String substringData(int offset,
```
mernst: This could be annotated as `@SideEffectFree`.

checker/jdk/nullness/src/org/w3c/dom/Text.java
```diff
@@ -0,0 +1,190 @@
...
+     *                              |
+     *                           /-----\
+     *                           | foo |
+     *                           \-----/
+     * </pre>
+     * <em>Figure: barTextNode.wholeText value is "barfoo"</em>
+     *
+     * @since DOM Level 3
+     */
+    @Pure public String getWholeText();
```
mernst: This `@Pure` annotation looks wrong.  A method that returns a String is not `@Deterministic` unless the result is interned.

nikshinde1996: This section of [Manual](https://checkerframework.org/manual/#type-refinement-purity) must  also mention to check if returned results are interned. The statement in manual that explained `@Deterministic` only mentioned about identical results. I will update annotation to `@SideEffectFree` in next commit.

mernst: The only way to ensure that the results are identical is if they are interned.  For example, two Strings could be `.equals()` to one another without being identical.

Nonetheless, I'll add to the manual a caution about the difference between `.equals()` and `==`.

nikshinde1996:
Why travis-ci is showing jdk8 and jdk7 download error ? Do I need to make any additional settings in my system ? Local build is successfull .

smillst:
The error isn't downloading the jdk, it's building it:
https://travis-ci.org/typetools/checker-framework/jobs/225046931#L1171

checker/jdk/nullness/src/org/w3c/dom/CharacterData.java
```diff
@@ -0,0 +1,186 @@
...
+     * implementation limits may mean that the entirety of a node's data may
+     * not fit into a single <code>DOMString</code>. In such cases, the user
+     * may call <code>substringData</code> to retrieve the data in
+     * appropriately sized pieces.
+     * @exception DOMException
+     *   DOMSTRING_SIZE_ERR: Raised when it would return more characters than
+     *   fit in a <code>DOMString</code> variable on the implementation
+     *   platform.
+     */
+    @Pure public String getData()
```
mernst: This is most likely `@SideEffectFree`, but how do you know it is `@Pure`?

checker/jdk/nullness/src/org/w3c/dom/Attr.java
```diff
@@ -0,0 +1,307 @@
...
+ * </tr>
+ * </table>
+ * <p>See also the <a href='http://www.w3.org/TR/2004/REC-DOM-Level-3-Core-20040407'>Document Object Model (DOM) Level 3 Core Specification</a>.
+ */
+public interface Attr extends Node {
+    /**
+     * Returns the name of this attribute. If <code>Node.localName</code> is
+     * different from <code>null</code>, this attribute is a qualified name.
+     */
+    @Pure public String getName();
```
mernst: How do you know this is `@Pure`?

checker/jdk/nullness/src/org/w3c/dom/Attr.java
```diff
@@ -0,0 +1,307 @@
...
+     * <br>On setting, this creates a <code>Text</code> node with the unparsed
+     * contents of the string, i.e. any characters that an XML processor
+     * would recognize as markup are instead treated as literal text. See
+     * also the method <code>Element.setAttribute()</code>.
+     * <br> Some specialized implementations, such as some [<a href='http://www.w3.org/TR/2003/REC-SVG11-20030114/'>SVG 1.1</a>]
+     * implementations, may do normalization automatically, even after
+     * mutation; in such case, the value on retrieval may differ from the
+     * value on setting.
+     */
+    @Pure public String getValue();
```
mernst: How do you know this is `@Pure`?

checker/jdk/nullness/src/org/w3c/dom/Attr.java
nikshinde1996: Every attribute is attribute node. From this [reference](https://www.w3.org/2003/01/dom2-javadoc/org/w3c/dom/Node.html) first row in tabular data, the results of `getName()` and `getValue()` method are similar to `getNodeName()` and `getNodeValue()` from [Node.java](https://github.com/typetools/checker-framework/blob/master/checker/jdk/nullness/src/org/w3c/dom/Node.java) . This both entities `name` and `value` cannot have null value.

checker/jdk/nullness/src/org/w3c/dom/CharacterData.java
nikshinde1996: I agree, I misinterpreted `data` to `value` in Node. I will correct this.

mernst:
The Travis tests are failing.  Can you pull upstream into this branch to see whether that fixes the problem?

checker/jdk/nullness/src/org/w3c/dom/Attr.java
mernst: This question is still waiting for an answer.

nikshinde1996: Every attribute is attribute node. From this [reference](https://www.w3.org/2003/01/dom2-javadoc/org/w3c/dom/Node.html) first row in tabular data, the results of `getValue()` method are similar to `getNodeValue()` from [Node.java](https://github.com/typetools/checker-framework/blob/master/checker/jdk/nullness/src/org/w3c/dom/Node.java) . In Attr class,  `value` cannot have null value.

PR #1217, opened 2017-04-04
Fix Value Checker crash on varargs

smillst:
One crash was a null pointer exception.

The second crash was a problem with statically evaluating vararg methods.

With these fixes, the Index Checker no longer crashes on daikon.

### DIFF
--- a/framework/src/org/checkerframework/common/value/ReflectiveEvaluator.java
+++ b/framework/src/org/checkerframework/common/value/ReflectiveEvaluator.java
@@ -67,7 +67,12 @@ public class ReflectiveEvaluator {
         }
 
         if (method.isVarArgs()) {
-            listOfArguments = normalizeVarargInvocations(listOfArguments, method);
+            List<Object[]> newList = new ArrayList<>();
+            int numberOfParameters = method.getParameterTypes().length;
+            for (Object[] args : listOfArguments) {
+                newList.add(normalizeVararg(args, numberOfParameters));
+            }
+            listOfArguments = newList;
         }
 
         List<Object> results = new ArrayList<>();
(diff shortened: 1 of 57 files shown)


framework/tests/src/testlib/lib/VarArgMethods.java
```diff
@@ -0,0 +1,33 @@
+package testlib.lib;
+
+import org.checkerframework.common.value.qual.StaticallyExecutable;
+
+/** Use by framework/tests/value/VarArgRe.java */
```
mernst: Typo "Use"

framework/tests/value/VarArgRe.java
```diff
@@ -0,0 +1,24 @@
+import org.checkerframework.common.value.qual.IntVal;
+import testlib.lib.VarArgMethods;
+
+public class VarArgRe {
+    // VarArgMethods is declarded in framework/tests/src/testlib/lib.
+    // All the methods return the length of the vararg.
+    public void use() {
```
mernst: Confusing method names.  I would call them `use0`, `use1`, and `use2`, and put them in order.

framework/src/org/checkerframework/common/value/ReflectiveEvalutator.java
```diff
@@ -94,6 +119,40 @@ public ReflectiveEvalutator(
     }
 
     /**
+     * Convert the arguments that are part of the var arg into an array.
```
mernst: "vararg" has no space.

framework/src/org/checkerframework/common/value/ReflectiveEvalutator.java
```diff
@@ -94,6 +119,40 @@ public ReflectiveEvalutator(
...
+        int numberOfParameters = method.getParameterTypes().length;
+        for (Object[] args : arguments) {
+            if (args == null) {
+                // null means no arguments.  For varargs no arguments is an empty array.
+                args = new Object[] {};
+            }
+            Object[] newArgs;
+            int numOfVarArgs = args.length - numberOfParameters + 1;
+            if (numOfVarArgs > 0) {
+                newArgs = new Object[numberOfParameters];
```
mernst: The first two and last 1 line of the then and else bodies are the same.  I would move them outside the `if` statement in order to reduce code duplication.

framework/src/org/checkerframework/common/value/ReflectiveEvalutator.java
mernst: Also, this doesn't document what the method does overall (just one part of its behavior).  Please express the behavior in terms of its inputs and outputs; this overview doesn't mention the list.

Even the `@param` documentation doesn't say how the method is related to the other arguments.

This doesn't have to be really elaborate, but what's currently there is confusing to someone who isn't the author.

framework/src/org/checkerframework/common/value/ReflectiveEvaluator.java
```diff
@@ -94,6 +119,43 @@ public ReflectiveEvalutator(
     }
 
     /**
+     * This method changes a set of varargs invocations (each of which passes an arbitrary number of
```
smillst: Should `set` be `list`?

smillst: I find the use of "invocation" in this comment confusing.  The "invocation" you are referring to is really not an invocation but an array of arguments.   Using invocation in the name of the method makes it extremely confusing at the call site where the method is actually invoked a few lines after.

mernst: Yes, it could be `list`; feel free to change it.  I didn't, because I was already confused about lists vs arrays and didn't want to compound it.  However, feel free to change it if doing so would improve the documentation.

mernst: With regard to "invocation", you can change that too.  Varargs is fundamentally a declaration-site concept, so at first I thought that this was doing something with parameter lists.  It's not:  it is only changing argument lists.  Maybe you can find a better way to express that.

smillst:
@mernst I refactored the code so that the method comment is easier to understand.  (It's no longer a list of arrays, but now only takes an array.)

framework/src/org/checkerframework/common/value/ReflectiveEvaluator.java
```diff
@@ -94,6 +124,36 @@ public ReflectiveEvalutator(
     }
 
     /**
+     * This method normalizes an array of arguments to a varargs method by changing the arguments
+     * associated with the varargs parameter into an array.
+     *
+     * @param arguments an array of arguments for {@code method}. The length is at least {@code
+     *     method.numberOfParameters-1}.
+     * @param method a method whose last formal parameter is a varargs parameter
+     * @return the length of the array is exactly {@code method.numberOfParameters-1}
```
mernst: Shouldn't this be method.numberOfParameters (without "-1")?

I agree that this is clearer as a method that works on one "invocation" at a time.

smillst: Yes.  Also, I need to update the Javadoc.

PR #1219, opened 2017-04-05
Fix #1096 by considering initialization when checking contracts.

### LINKED ISSUE #1096: Precondition checks of non-null fields on fully-initialized objects
Take this example:

```
import org.checkerframework.checker.nullness.qual.RequiresNonNull;

class PreCond {
  Object f;

  PreCond() {
    f = new Object();
    early();
  }

  @RequiresNonNull("this.f")
  void early(@UnknownInitialization PreCond this) {
    f.toString();
  }
}

class User {
  void bar(PreCond pc) {
    pc.early();
  }
}
```

The Nullness Checker gives:

```
PreCond.java:19: error: [contracts.precondition.not.satisfied] the called method 'pc.early()' has a precondition 'pc.f' that is not satisfied
    pc.early();
            ^
1 error
```

Indeed, at the invocation in User, we don't know explicitly that field `f` has been initialized.
However, we do know that `pc` is fully initialized and that the field `f` is non-null. Therefore, the fields are guaranteed to satisfy the pre-condition.
This is particularly useful for a class with multiple fields and a method needs to be called on an partially initialized object or after full initialization.

A pre-condition of a non-null field on a fully-initialized receiver should be considered satisfied.
Does this sound right to you?

wmdietlGC:
Have a look through the test case and let me know if there is a combination that's worth testing.

If you know of a better way to handle all possible receiver kinds, please let me know (see TODO comment).


mernst:
Looks fine to me -- that is, I don't see other test cases or a better way to handle it in the code.
Feel free to merge.  Thanks!

checker/src/org/checkerframework/checker/initialization/InitializationVisitor.java
```diff
@@ -181,6 +184,19 @@ protected boolean checkContract(
                             return true;
                         }
                     }
+                } else if (fa.getReceiver() instanceof LocalVariable) {
+                    // TODO: what is the best way to handle other types of
+                    // receivers?
+                    Element elem = ((LocalVariable) fa.getReceiver()).getElement();
+                    AnnotatedTypeMirror recvType = atypeFactory.getAnnotatedType(elem);
```
smillst: Instead of getting the type from the factory, is it sufficient to get the CFValue from the store?

wmdietl: The store might not contain anything for the receiver and then I would need to do the fallback myself.
The factory will look into the store and use it's value for me, so in case something refined is available, it should will used.
Is there a method that would allow me to go from `fa.getReceiver()` to an AnnotatedTypeMirror, without me having to know it's a local variable, getting the Element, and then asking the factory?

smillst: No, but you can extract the code in GenericAnnotatedTypeFactory#getAnnotationFromJavaExpressionString into such a method.

smillst: I would recommending using the store first because it contains information about more than just LocalVariables.  Then if it does not have a value, then fall back to using the AnnotatedTypeFactory if the Receiver is a LocalVariable or a FieldAccess -- the only two kinds of Receivers with an element. 

Also, I don't think getAnnotatedType(Element) will return a refined type, will it?

wmdietlGC: I tried using the Store before going to the ATF. I'm not sure the end result is really nicer, but if you prefer this I'm fine with it.
Can you please have another look @smillst ?

checker/src/org/checkerframework/checker/initialization/InitializationVisitor.java
```diff
@@ -181,6 +185,35 @@ protected boolean checkContract(
...
+                    Value value = (Value) store.getValue(fa.getReceiver());
+                    if (value != null) {
+                        recvAnnoSet = value.getAnnotations();
+                    } else if (fa.getReceiver() instanceof LocalVariable) {
+                        Element elem = ((LocalVariable) fa.getReceiver()).getElement();
+                        AnnotatedTypeMirror recvType = atypeFactory.getAnnotatedType(elem);
+                        recvAnnoSet = recvType.getAnnotations();
+                    } else {
+                        // Is there anything better we could do?
+                        return false;
```
smillst: Should this be a call to super.checkContracts or will that always return false in this case?

checker/src/org/checkerframework/checker/initialization/InitializationVisitor.java
```diff
@@ -181,6 +185,35 @@ protected boolean checkContract(
...
+                    @SuppressWarnings("unchecked")
+                    Value value = (Value) store.getValue(fa.getReceiver());
+                    if (value != null) {
+                        recvAnnoSet = value.getAnnotations();
+                    } else if (fa.getReceiver() instanceof LocalVariable) {
+                        Element elem = ((LocalVariable) fa.getReceiver()).getElement();
+                        AnnotatedTypeMirror recvType = atypeFactory.getAnnotatedType(elem);
+                        recvAnnoSet = recvType.getAnnotations();
+                    } else {
+                        // Is there anything better we could do?
```
smillst: If `fa.getReceiver()` is a field access, you can get the annotated type via the element as you did for local variables.

wmdietl: Let's see whether this comes up. It could also be a field access on a local variable or anything more complicated. A general-purpose method for how to use Receiver would be very helpful.

checker/src/org/checkerframework/checker/initialization/InitializationVisitor.java
wmdietl: A contract couldn't be resolved, so returning false seems like the conservative answer.

PR #1220, opened 2017-04-05
Address feedback on field invariant feature

smillst:
This pull request addresses comments on https://github.com/typetools/checker-framework/commit/8a4bd6c0bdf1ab5f373c6e8f8135363fe1a30583.


framework/src/org/checkerframework/common/basetype/messages.properties
```diff
@@ -80,9 +80,9 @@ contracts.conditional.postcondition.expression.parameter.name=The conditional po
 lambda.unimplemented=This version of the Checker Framework does not type-check lambda expressions.
 methodref.inference.unimplemented=This version of the Checker Framework does not type-check method references with implicit type arguments.
 
-field.invar.not.found=the field invariant annotation refers to fields not found in a superclass\nfields not found: %s
-field.invar.not.final=the field invariant annotation refers to fields that are not final\nfields not final: %s
-field.invar.not.subtype=the qualifier for field %s is not a subtype of the declared type\nfound: %s\ndeclared type: %s
-field.invar.not.wellformed=the field invariant annotation does not have equal numbers of fields and qualifiers.
-field.invar.not.found.superclass=the field invariant annotation is missing fields that are listed in the superclass field invariant.\nfields not found: %s
-field.invar.not.subtype.superclass=the qualifier for field %s is not a subtype of the qualifier in the superclass field invariant\nfound: %s\nsuperclass type: %s
+field.invariant.not.found=the field invariant annotation refers to fields not found in a superclass\nfields not found: %s
```
smillst: @wmdietl You suggested making `field.invariant` a suffix, but the pattern in this file is to use prefixes.

framework/src/org/checkerframework/framework/util/FieldInvariantObject.java
```diff
@@ -13,7 +13,9 @@
  * this field invariant applies.
  */
 public class FieldInvariantObject {
```
smillst: @wmdietl said:
>'Object' is terrible in a class name. Is there nothing more specific we can use?

What would you suggest? FieldInvariant would be the best name, but that's the name of the annotation.

framework/src/org/checkerframework/common/basetype/messages.properties
mernst: Prefixes are fine

framework/src/org/checkerframework/framework/util/FieldInvariantObject.java
mernst: Here are some possibilities:
InvariantFields
InvariantFieldList
FieldsAndQualifiers

framework/src/org/checkerframework/common/basetype/messages.properties
wmdietl: But only suffixes work. If you have a error key a.b.c, you can suppress it with '@SuppressWarnings("a.b.c")', '@SuppressWarnings("b.c")', and '@SuppressWarnings("c")'.
So we want the unifying, general part as the suffix, to allow suppressing multiple errors at once.
If there are other keys that don't follow this pattern, they should be changed, too.

smillst: According to the documentation, it's prefixes that allow multiple errors to be suppressed; see [SourceChecker#checkSuppressWarnings](https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/source/SourceChecker.java#L1219).

wmdietl: Hmm, in https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/common/basetype/messages.properties we definitely use suffixes, e.g. for all the `type.incompatible` keys.
Could it be that it's neither and it just checks for containment? See https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/source/SourceChecker.java#L1270
I didn't find a place where it splits the message on the `.` and checks for prefix or suffix, nor is there a startsWith or endsWith.

smillst: I agree that it appears both suffixes and prefixes work.  I found prefixes easier to read in this case.

framework/src/org/checkerframework/framework/util/FieldInvariantObject.java
wmdietl: Is this representing a single FieldInvariant or multiple?

framework/src/org/checkerframework/framework/util/FieldInvariantObject.java
```diff
@@ -13,7 +13,9 @@
...
+    /** simple name of the field */
```
wmdietl: This is a List, so it can't be the simple name of a field.

framework/src/org/checkerframework/framework/util/FieldInvariantObject.java
```diff
@@ -13,7 +13,9 @@
...
     private final List<String> fields;
+
```
wmdietl: What qualifiers are these? Do all these qualifiers apply to all the fields? Or do the two lists have to be the same length and align? Instead of Lists should both fields be Sets?
The class javadoc talks about a `A list of fields that have a specific qualifier`, so it's not clear what this List represents.

PR #1221, opened 2017-04-06
Convert @ArrayLen with negative numbers  to bottom

smillst:
Also, issue a warning if an explicit array len has a negative value.  I refactored the code that issue warnings on explicit annotations to use AnnotationMirror so that the code is easier to read.


smillst:
@JasonMrX  and @kelloggm this is a fix for the bug I mentioned in kelloggm#150.

smillst:
@mernst  Your changes look good.  Thanks!

PR #1222, opened 2017-04-06
Refine @UnknownVal to @IntRange when possible.

smillst:
This allows the following code to type check:
```
void test(int x) {
    if ( x > 0 ) {
      @IntRange(from=1) int y = x;
```

### DIFF
--- a/framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
+++ b/framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
@@ -1242,15 +1242,19 @@ public class ValueAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
     }
 
     /**
-     * Create an {@code @IntRange} annotation from the range. May return BOTTOMVAL or UNKNOWNVAL.
+     * Create an {@code @IntRange} or {@code @IntVal} annotation from the range. May return
+     * BOTTOMVAL or UNKNOWNVAL.
      */
     public AnnotationMirror createIntRangeAnnotation(Range range) {
         if (range.isNothing()) {
             return BOTTOMVAL;
         } else if (range.isEverything()) {
             return UNKNOWNVAL;
-        } else {
+        } else if (range.isWiderThan(MAX_VALUES)) {
             return createIntRangeAnnotation(range.from, range.to);
+        } else {
+            List<Long> newValues = ValueCheckerUtils.getValuesFromRange(range, Long.class);
+            return createIntValAnnotation(newValues);
         }
     }
 
--- a/framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
+++ b/framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
@@ -44,8 +44,6 @@ import org.checkerframework.framework.flow.CFValue;
 import org.checkerframework.framework.type.AnnotatedTypeFactory;
 import org.checkerframework.framework.type.AnnotatedTypeMirror;
 import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedArrayType;
-import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedDeclaredType;
-import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedPrimitiveType;
 import org.checkerframework.framework.type.QualifierHierarchy;
 import org.checkerframework.framework.type.treeannotator.ImplicitsTreeAnnotator;
 import org.checkerframework.framework.type.treeannotator.ListTreeAnnotator;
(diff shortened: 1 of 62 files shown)


smillst:
I'm not sure who should review this change.  @kelloggm @JasonMrX  @mernst ?

kelloggm:
I'll review it, though you might want someone else to look, too.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1242,15 +1242,19 @@ public AnnotationMirror createIntRangeAnnotation(long from, long to) {
     }
 
     /**
-     * Create an {@code @IntRange} annotation from the range. May return BOTTOMVAL or UNKNOWNVAL.
+     * Create an {@code @IntRange} or {@code @IntVal} annotation from the range. May return
+     * BOTTOMVAL or UNKNOWNVAL.
      */
     public AnnotationMirror createIntRangeAnnotation(Range range) {
```
kelloggm: Should this be combined with ValueAnnotatedTypeFactory#createIntValAnnotation to create one unified createIntAnnotation function, which can take either a range or a list of values (and then converts one to the other before calling one unified method)?

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -777,8 +783,14 @@ private Range calculateRangeUnaryOp(
             TransferInput<CFValue, CFStore> p,
             CFStore thenStore,
             CFStore elseStore) {
-        if (isIntRange(leftNode, p) || isIntRange(rightNode, p)) {
-            return refineIntRanges(leftNode, rightNode, op, p, thenStore, elseStore);
+        if (isIntRange(leftNode, p)
```
kelloggm: I think there should be a comment here explaining why the check against UnknownVal happens. It's not obvious from reading the code, and I think it would be confusing if I wasn't reviewing a change for exactly this reason.

mernst:
If you want a review from me after you address Martin's comments, please let me know.
Merging based on Martin's review is perfectly fine with me, too.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1242,15 +1242,19 @@ public AnnotationMirror createIntRangeAnnotation(long from, long to) {
...
         if (range.isNothing()) {
             return BOTTOMVAL;
         } else if (range.isEverything()) {
             return UNKNOWNVAL;
-        } else {
+        } else if (range.isWiderThan(MAX_VALUES)) {
             return createIntRangeAnnotation(range.from, range.to);
+        } else {
```
JasonMrX: @mernst and I had a discussion regarding this before -- do we convert `@IntRange` to `@IntVal` when the range is not wider than 10?. I remember Mike's concern was that if we do this conversion, users might be confused if they are intended to use `@IntRange` but it get converted to `@IntVal` easily during constant analysis. For example:
```java
void test(@IntRange(from = 0, to = 1) a) {
  int b = a + 1; // b has @IntVal({1, 2}) instead of @IntRange(from = 1, to = 2);
}
```
So at that time, we decided not to make this conversion for now and would revisit this if the precision enhancement provided by the conversion is really needed.

If we do the conversion here, for consistency, I think we might also need to add the same logic in `ValueAnnotatedTypeFactory#ValueTypeAnnotator#replaceWithNewAnnoInSpecialCases`. And maybe add a warning that `@IntRange` is converted to `@IntVal` for better precision?

mernst: I was motivated by the example when the user writes `@IntRange(from = 0, to = 1)` and is shown instead the more verbose `@IntVal({0, 1, 2, 3, 4, 5, 6, 7, 8, 9})`.  I thought that consistency with what the user wrote was desirable.

I agree with Jason that we should be consistent regardless.

smillst: I agree that keeping the annotation in the warning messages consistent with what the user wrote is important.  However, at this point in the code, there is no way to know if range from an explicitly written annotation or was computing by dataflow (for example `x > 5 && x <7`).  Also, the Range object could have come from an explicitly written `@IntVal` annotation which are converted to Range objects sometimes.

What if instead of using `@IntVal` when the annotation is wider than `MAX_VALUES`, `@IntVal` is used of the range is wider than 2? That way `@IntRange(from=1, to=1)` will be converted to `@IntVal(1)` and  `@IntRange(from=1, to=2)` to `@IntVal({1, 2})`.

mernst: I had thought that, even though some annotations are implicit (computed by dataflow) and some have been converted, it was still better to retain the programmer-written form when possible -- that is, not to gratuitously change them.

Can you remind me the benefit of canonicalizing the annotations into a consistent form here?

smillst: The benefit to doing it here is that this method is called by everywhere an `@IntRange` is computed.

smillst: Basically, this the one places all computed Range objects are converted to annotations.

mernst: I agree that Range objects need to be converted to annotations.
One approach would be to convert all Range objects to IntRange annotations.
Another approach would be to convert some Range objects to IntRange annotations, and convert other Range objects to IntVal annotations.
Can you tell me the advantages of the latter over the former?

smillst: 1.  Annotations are written more compactly and easier to read.  `@IntRange(from=1, to=2)` converted to `@IntVal({1, 2})`

2. If the Range was from an explicitly written `@IntVal`, then it will appear in error messages as the user wrote it.

3.  At binary comparisons, such as `<`, if one side is an `@IntRange` and the other is `@IntVal`, the `@IntVal` is converted to a range.  This causes the refinement of `@IntVal` side to be less precise.  For example,

```
void test(@IntVal({1, 10}) int value, @IntRange(from=3, to=7) int range) {
  if (value < range) {
    // without this conversion: value: @IntVal({1, 10})
    // with this conversion: value: @IntVal(1)
    // in both cases range isn't refined and therefor is still @IntRange(from=3, to=7)
  }
```
So this conversion makes the Value Checker a little more precise without having to write refinement rules when one annotation is a range and the other is a value.  The Transfer function could do this conversion just before refining types and get the same benefit.

smillst: @mernst any more thoughts?

mernst: I don't think this is worth a huge design discussion.  Go ahead and do what you think is right.

PR #1223, opened 2017-04-07
Fix crash in Value Checker

kelloggm:
Fixes a crash uncovered by @as56 in the Index Checker case study.

### DIFF
--- a/framework/src/org/checkerframework/common/value/ValueTransfer.java
+++ b/framework/src/org/checkerframework/common/value/ValueTransfer.java
@@ -108,10 +108,7 @@ public class ValueTransfer extends CFTransfer {
         } else if (isIntRange(subNode, p)) {
             Range range = getIntRange(subNode, p);
             List<Long> longValues = ValueCheckerUtils.getValuesFromRange(range, Long.class);
-            values =
-                    longValues == null
-                            ? null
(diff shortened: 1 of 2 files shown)


framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -108,7 +108,10 @@ public ValueTransfer(CFAbstractAnalysis<CFValue, CFStore, CFTransfer> analysis)
         } else if (isIntRange(subNode, p)) {
             Range range = getIntRange(subNode, p);
             List<Long> longValues = ValueCheckerUtils.getValuesFromRange(range, Long.class);
-            values = NumberUtils.castNumbers(subNode.getType(), longValues);
+            values =
+                    longValues == null
+                            ? null
+                            : NumberUtils.castNumbers(subNode.getType(), longValues);
```
smillst: I changed NumberUtils.castNumbers to except a null parameter in #1217 (which was only merged earlier today).  So this part of the change isn't needed, but we should add the testcase to make sure.

kelloggm: Alright. I'll cancel this build, revert this change, and see if the tests pass.

PR #1225, opened 2017-04-07
Index Checker: Adds a lengthOf annotation that aliases IndexOrHigh

kelloggm:
This improves the situation in kelloggm#151.

### DIFF
--- a/checker/src/org/checkerframework/checker/index/qual/LengthOf.java
+++ b/checker/src/org/checkerframework/checker/index/qual/LengthOf.java
@@ -1,0 +1,13 @@
+package org.checkerframework.checker.index.qual;
+
+/**
+ * An integer that, for each of the given sequences, is equal to the sequence's length. This is
+ * treated as an {@link IndexOrHigh} annotation internally.
+ *
+ * @see IndexOrHigh
+ * @checker_framework.manual #index-checker Index Checker
+ */
+public @interface LengthOf {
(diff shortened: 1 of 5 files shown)


checker/src/org/checkerframework/checker/index/qual/LengthOf.java
```diff
@@ -0,0 +1,13 @@
+package org.checkerframework.checker.index.qual;
+
+/**
+ * An integer that, for each of the given sequences, is equal to the sequence's length. This is
```
mernst: Put the second sentence in a paragraph of its own, and state that it is a temporary implementation detail that will be fixed in the future.

checker/src/org/checkerframework/checker/index/qual/LengthOf.java
```diff
@@ -0,0 +1,13 @@
...
+ * treated as an {@link IndexOrHigh} annotation internally.
+ *
+ * @see IndexOrHigh
```
mernst: Given the `{@link ...}` in the body, I don't think you need an explicit `@see`.

kelloggm:
I haven't updated the manual's figure because this isn't really part of either the UBC's type system or the LBC's type system. Instead, I included it in the list of aliased qualifiers (along with `@IndexOrHigh`). Sorry I forgot to add it to the manual before!

PR #1231, opened 2017-04-10
Add ArrayLenRange annotation to Value Checker

kelloggm:
Adds an `@ArrayLenRange` annotation, which expresses that an array's length must be between the two values the annotation holds. The relationship between `@ArrayLenRange` and `@ArrayLen` is analogous to the relationship between `@IntRange` and `@IntVal`.

Fixes kelloggm#150

### DIFF
--- a/framework/src/org/checkerframework/common/value/RangeOrListOfValues.java
+++ b/framework/src/org/checkerframework/common/value/RangeOrListOfValues.java
@@ -1,0 +1,95 @@
+package org.checkerframework.common.value;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import javax.lang.model.element.AnnotationMirror;
+import org.checkerframework.common.value.qual.ArrayLen;
+import org.checkerframework.common.value.qual.ArrayLenRange;
+import org.checkerframework.common.value.qual.IntVal;
+import org.checkerframework.common.value.util.Range;
+import org.checkerframework.framework.type.AnnotatedTypeMirror;
+
+/**
+ * An abstraction that can be either a range or a list of values that could come from an {@link
+ * ArrayLen} or {@link IntVal}. This abstraction reduces the number of cases that {@link
+ * org.checkerframework.common.value.ValueAnnotatedTypeFactory.ValueTreeAnnotator#handleInitializers(List,
+ * AnnotatedTypeMirror.AnnotatedArrayType)} and {@link
+ * org.checkerframework.common.value.ValueAnnotatedTypeFactory.ValueTreeAnnotator#handleDimensions(List,
+ * AnnotatedTypeMirror.AnnotatedArrayType)} must handle.
+ */
+class RangeOrListOfValues {
+    Range range;
+    List<Integer> values;
+    boolean isRange;
+
+    public RangeOrListOfValues() {
+        this.values = new ArrayList<>();
+        isRange = false;
+    }
+
+    public RangeOrListOfValues(List<Integer> values) {
+        this.values = new ArrayList<>();
+        isRange = false;
+        addAll(values);
+    }
+
+    public RangeOrListOfValues(Range range) {
+        this.range = range;
+        isRange = true;
+    }
+
+    /**
+     * If this is not a range, adds all members of newValues to the list. Otherwise, extends the
+     * range as appropriate based on the max and min of newValues. If adding newValues to a
+     * non-range would cause the list to become too large, converts this into a range.
+     */
+    public void addAll(List<Integer> newValues) {
+        if (isRange) {
+            Range newValueRange = new Range(Collections.min(newValues), Collections.max(newValues));
+            range = range.union(newValueRange);
+        } else {
+            for (Integer i : newValues) {
+                if (!values.contains(i)) {
+                    values.add(i);
+                }
+            }
+            if (values.size() > ValueAnnotatedTypeFactory.MAX_VALUES) {
+                convertToRange();
+            }
+        }
+    }
+
+    /**
+     * Produces the most precise annotation that captures the information stored in this
+     * RangeOrListofValues. The result is either a {@link ArrayLen} or a {@link ArrayLenRange}.
+     */
+    public AnnotationMirror createAnnotation(ValueAnnotatedTypeFactory atypefactory) {
+        if (isRange) {
+            return atypefactory.createArrayLenRangeAnnotation(range);
+        } else {
+            return atypefactory.createArrayLenAnnotation(values);
+        }
+    }
+
+    // To be called before addAll
+    public static List<Integer> convertLongsToInts(List<Long> newValues) {
+        List<Integer> result = new ArrayList<>(newValues.size());
+        for (Long l : newValues) {
+            result.add(l.intValue());
+        }
+        return result;
+    }
+
+    /**
+     * Transforms this into a range. Fails if there are no values in the list. Has no effect if this
+     * is already a range.
+     */
+    public void convertToRange() {
+        if (!isRange) {
+            isRange = true;
+            range = new Range(Collections.min(values), Collections.max(values));
+            values = null;
+        }
+    }
+}
--- a/framework/src/org/checkerframework/common/value/RangeOrListOfValues.java
+++ b/framework/src/org/checkerframework/common/value/RangeOrListOfValues.java
@@ -19,9 +19,9 @@ import org.checkerframework.framework.type.AnnotatedTypeMirror;
  * AnnotatedTypeMirror.AnnotatedArrayType)} must handle.
  */
 class RangeOrListOfValues {
-    Range range;
-    List<Integer> values;
-    boolean isRange;
+    private Range range;
+    private List<Integer> values;
+    private boolean isRange;
 
     public RangeOrListOfValues() {
         this.values = new ArrayList<>();
@@ -37,6 +37,15 @@ class RangeOrListOfValues {
     public RangeOrListOfValues(Range range) {
         this.range = range;
         isRange = true;
+    }
+
+    public void add(Range otherRange) {
+        if (isRange) {
+            range = range.union(otherRange);
+        } else {
+            convertToRange();
+            add(otherRange);
+        }
     }
 
     /**
--- a/framework/src/org/checkerframework/common/value/RangeOrListOfValues.java
+++ b/framework/src/org/checkerframework/common/value/RangeOrListOfValues.java
@@ -55,6 +55,9 @@ class RangeOrListOfValues {
      * If this is not a range, adds all members of newValues to the list. Otherwise, extends the
      * range as appropriate based on the max and min of newValues. If adding newValues to a
      * non-range would cause the list to become too large, converts this into a range.
+     *
+     * <p>If reading from an {@link org.checkerframework.common.value.qual.IntRange} annotation,
+     * {@link #convertLongsToInts(List)} should be called before calling this method.
      */
     public void addAll(List<Integer> newValues) {
         if (isRange) {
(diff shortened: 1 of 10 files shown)


smillst:
You forgot to add/commit `ArrayLenRange.java`.

kelloggm:
I did! Thanks for pointing that out!

smillst:
You'll need to add `@ArrayLenRange` to the manual.  I'll review the code without that, but it can't be merged until that's done.

kelloggm:
I've added documentation to the manual, added some more tests, and fixed the issue with defaults. I'd appreciate it if you could take another look.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -580,26 +636,79 @@ private void handleDimensions(
...
 
+        // An abstraction that can be either a range or a list of values that could come from an
+        // ArrayLen or IntVal. This abstraction reduces the number of cases that handleInitializers
+        // and handleDimensions must handle.
+        private class RangeOrListOfValues {
+            Range range;
+            List<Integer> values;
+            boolean isRange;
+
+            public RangeOrListOfValues(Integer... values) {
```
smillst: This is never called with a variable number of arguments, only the empty array.  Also the next line after the call is a call to addAll.  This should be one 0 argument constructor and a constructor that takes a list.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -580,26 +636,79 @@ private void handleDimensions(
...
+                isRange = false;
+            }
+
+            public RangeOrListOfValues(Range range) {
+                this.range = range;
+                isRange = true;
+            }
+
+            public void addAll(List<Integer> newValues) {
+                for (Integer i : newValues) {
```
smillst: This method does not account for the possibility that that this is a range not a list.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -580,26 +636,79 @@ private void handleDimensions(
...
+                    } else {
+                        newQual = createArrayLenAnnotation(rolv.values);
                     }
-                    AnnotationMirror newQual = createArrayLenAnnotation(new ArrayList<>(lengths));
                     type.replaceAnnotation(newQual);
                 }
             }
         }
 
+        // An abstraction that can be either a range or a list of values that could come from an
```
smillst: This should be a Javadoc comment.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -580,26 +636,79 @@ private void handleDimensions(
...
+                    rolv.addAll(rolv.convertLongsToInts(getIntValues(dimType)));
                 }
-                if (longLengths != null) {
-                    HashSet<Integer> lengths = new HashSet<Integer>(longLengths.size());
-                    for (Long l : longLengths) {
-                        lengths.add(l.intValue());
+                if (rolv != null) {
+                    AnnotationMirror newQual;
+                    if (rolv.isRange) {
+                        newQual = createArrayLenRangeAnnotation(rolv.range);
```
smillst: Make createAnnotation a method in RangeOrListOfValues. There are two occurs that of this pattern that can be replaced with a call to that method.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -620,23 +729,43 @@ private void handleInitializers(
...
-                    } else {
-                        arrayLens = arrayLenOfDimensions.get(dimension);
+                        arrayLenOfDimensions.add(new RangeOrListOfValues());
                     }
                     AnnotationMirror arrayLen = componentType.getAnnotation(ArrayLen.class);
                     if (arrayLen != null) {
+                        rolv = arrayLenOfDimensions.get(dimension);
                         List<Integer> currentLengths = getArrayLength(arrayLen);
-                        arrayLens.addAll(currentLengths);
+                        if (rolv.isRange) {
```
smillst: The method addAll in RangeOrList should take a list and properly add if RangeOrList is a range.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -272,19 +273,37 @@ private void refineArrayAtLengthAccess(FieldAccessNode arrayLengthNode, CFStore
...
-            List<Long> lengthValues = ValueAnnotatedTypeFactory.getIntValues(lengthAnno);
-            List<Integer> arrayLenValues = new ArrayList<>(lengthValues.size());
-            for (Long l : lengthValues) {
-                arrayLenValues.add(l.intValue());
+            Range range;
+            if (isIntRange) {
+                range = ValueAnnotatedTypeFactory.getRange(lengthAnno);
+            } else {
+                List<Long> lengthValues = ValueAnnotatedTypeFactory.getIntValues(lengthAnno);
+                range = new Range(Collections.min(lengthValues), Collections.max(lengthValues));
```
smillst: This loses precision if the array.length is something like `@IntVal({1, 3})`.  Is there are reason for this?  You could use the RangeOrListOfValues class here.

kelloggm:
I've fixed the failing Index Checker tests, and addressed your review comments.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -275,7 +270,22 @@ private void replaceWithNewAnnoInSpecialCases(AnnotatedTypeMirror atm) {
...
+                    long to = AnnotationUtils.getElementValue(anno, "to", Long.class, true);
+                    if (from > to) {
+                        atm.replaceAnnotation(BOTTOMVAL);
+                    }
+                } else if (AnnotationUtils.areSameByClass(anno, ArrayLenRange.class)) {
+                    int from = AnnotationUtils.getElementValue(anno, "from", Integer.class, true);
+                    int to = AnnotationUtils.getElementValue(anno, "to", Integer.class, true);
+                    if (from > to) {
+                        atm.replaceAnnotation(BOTTOMVAL);
                     }
```
smillst: (fixed)

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -620,23 +674,43 @@ private void handleInitializers(
...
-                    } else {
-                        arrayLens = arrayLenOfDimensions.get(dimension);
+                        arrayLenOfDimensions.add(new RangeOrListOfValues());
                     }
                     AnnotationMirror arrayLen = componentType.getAnnotation(ArrayLen.class);
                     if (arrayLen != null) {
+                        rolv = arrayLenOfDimensions.get(dimension);
                         List<Integer> currentLengths = getArrayLength(arrayLen);
-                        arrayLens.addAll(currentLengths);
+                        if (rolv.isRange) {
```
smillst: Call rolv.addAll instead of using this if/else

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -620,23 +674,43 @@ private void handleInitializers(
...
+                        } else {
+                            rolv.addAll(currentLengths);
+                        }
+                    } else {
+                        // Check for an arrayLenRange annotation
+                        AnnotationMirror arrayLenRangeAnno =
+                                componentType.getAnnotation(ArrayLenRange.class);
+                        if (arrayLenRangeAnno != null) {
+                            Range range = getRange(arrayLenRangeAnno);
+                            rolv.convertToRange();
```
smillst: Rolv is always null in this branch.  Please add a test case the covers this code.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -620,23 +674,43 @@ private void handleInitializers(
...
+                            rolv.range = range.union(rolv.range);
```
smillst: All the fields of RangeOrList should be private.  Here, this code should be in an add(Range) method.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -620,23 +674,43 @@ private void handleInitializers(
...
             // A list of arrayLens.  arrayLenOfDimensions.get(i) is the array lengths for the ith
             // dimension.
-            List<List<Integer>> arrayLenOfDimensions = new ArrayList<>();
+            List<RangeOrListOfValues> arrayLenOfDimensions = new ArrayList<>();
             for (ExpressionTree init : initializers) {
                 AnnotatedTypeMirror componentType = getAnnotatedType(init);
                 int dimension = 0;
                 while (componentType.getKind() == TypeKind.ARRAY) {
-                    List<Integer> arrayLens;
+                    RangeOrListOfValues rolv = null;
```
smillst: It's better to leave local variables uninitialized rather than setting them to `null`.  (Unless `null` is an expected value.)  That way javac will issue a warning letting you know it was not properly initialized.

framework/src/org/checkerframework/common/value/RangeOrListOfValues.java
```diff
@@ -0,0 +1,95 @@
...
+        } else {
+            return atypefactory.createArrayLenAnnotation(values);
+        }
+    }
+
+    // To be called before addAll
+    public static List<Integer> convertLongsToInts(List<Long> newValues) {
+        List<Integer> result = new ArrayList<>(newValues.size());
+        for (Long l : newValues) {
+            result.add(l.intValue());
```
smillst: Why is it safe to ignore overflow here?  If l is too big, then the ArrayLen* annotation will end up with negative array lengths.

kelloggm:
I've addressed this set of comments.

framework/src/org/checkerframework/common/value/RangeOrListOfValues.java
```diff
@@ -0,0 +1,116 @@
...
+        } else {
+            convertToRange();
+            add(otherRange);
+        }
+    }
+
+    /**
+     * If this is not a range, adds all members of newValues to the list. Otherwise, extends the
+     * range as appropriate based on the max and min of newValues. If adding newValues to a
+     * non-range would cause the list to become too large, converts this into a range.
```
smillst: Please mention that `convertLongsToInts` should be used if to convert `List<Long>' to 'List<Integer>'.

kelloggm:
@JasonMrX I'm going to go ahead and merge this. If you have comments on it, I'm still happy to address them.

JasonMrX:
Thanks @kelloggm for asking and sorry for not being able to following up with the updates.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1151,7 +1229,7 @@ public AnnotationMirror createArrayLenAnnotation(List<Integer> values) {
         if (values.isEmpty() || Collections.min(values) < 0) {
```
JasonMrX: Minor: seems to be a small negligence in the original code, but the `values.isEmpty()` check seems to be unnecessary, does it?

framework/src/org/checkerframework/common/value/qual/BottomVal.java
```diff
@@ -20,7 +20,15 @@
...
 )
-@SubtypeOf({ArrayLen.class, BoolVal.class, DoubleVal.class, IntVal.class, StringVal.class})
+@SubtypeOf({
+    ArrayLen.class,
+    BoolVal.class,
+    DoubleVal.class,
+    IntVal.class,
+    StringVal.class,
+    ArrayLenRange.class,
+    IntRange.class
```
JasonMrX: Thanks.

PR #1233, opened 2017-04-12
Value Checker no-overflow mode

kelloggm:
This adds a no-overflow mode to the value checker. This mode only takes effect for `@IntRange` annotations (`@ArrayLenRange` support needs to be added when typetools#1231 is merged). It's purpose is to allow range annotations to replace the overflow-ignoring annotations in the MinLen Checker and the Lower Bound Checker, which will be integrated into the Value Checker soon.


framework/tests/value-no-overflow/ValueNoOverflow.java
```diff
@@ -0,0 +1,31 @@
+import org.checkerframework.common.value.qual.*;
+
+class ValueNoOverflow {
```
smillst: Please add more tests.  You could copy checker/tests/index/Transfer*.java and replace the LBC annotations with the proper `@IntRange` annotation.  You should also test with longs rather than ints.  Also, please a few tests with bytes and shorts.

framework/build.xml
```diff
@@ -614,6 +614,13 @@
         </antcall>
     </target>
 
+    <target name="value-tests-no-overflow" depends="jar,build-tests"
```
smillst: Please consistently use "ignore-overflow" rather than "no-overflow"

framework/src/org/checkerframework/common/value/util/Range.java
```diff
@@ -130,6 +130,26 @@ public Range intRange() {
         }
     }
 
+    /**
+     * Either calls {@link #intRange()} if ignoreOverflow is false, or creates a new range whose
+     * from is either this.from or Integer.MIN_VALUE, and whose to is either this.to or
+     * Integer.MAX_VALUE.
+     */
+    public Range intRange(boolean ignoreOverflow) {
```
smillst: Is there any reason that `ignoreOverflow` isn't a field of the Range class?  It seems error prone to have both this method and the `intRange()` method.  When should `intRange()` be called?  (The `intRange()` method should at least be private, but I think the better design would be for `ignoreOverflow` to be a final field.)

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -447,10 +447,11 @@ private Range calculateRangeBinaryOp(
                     throw new RuntimeException("this can't happen");
             }
             // Any integral type with less than 32 bits would be promoted to 32-bit int type during operations.
+            System.out.println(resultRange);
```
smillst: Remove println.

kelloggm:
@smillst Do you think it would be better to rename `-AignoreOverflow` to `-AignoreOverflowForRange` or something similar? That would help convey that it really is just for `*Range` annotations.

smillst:
That's a good idea.  What about `ignoreRangeOverflow`?

kelloggm:
I like `ignoreRangeOverflow`, so I'll change it to that.

kelloggm:
@smillst, I believe I've addressed all your feedback, and I'd appreciate a second review.

docs/manual/constant-value-checker.tex
```diff
@@ -245,6 +245,18 @@ \section{Warnings\label{value-checker-warnings}}
 
 \end{itemize}
 \end{sloppypar}
+
+\section{Overflow}
+\begin{sloppypar}
+The constant value checker normally takes Java's normal overflow rules into account when
```
smillst: Capitalize `Constant Value Checker`

framework/src/org/checkerframework/common/value/util/Range.java
```diff
@@ -853,21 +892,35 @@ private boolean isWithinInteger() {
...
             return EVERYTHING;
         } else {
-            long resultFrom = bigFrom.longValue();
-            long resultTo = bigTo.longValue();
+            long resultFrom;
+            long resultTo;
+
+            if (IGNORE_OVERFLOW) {
+                BigInteger longMin = BigInteger.valueOf(Long.MIN_VALUE);
+                resultFrom = bigFrom.compareTo(longMin) < 0 ? Long.MIN_VALUE : bigFrom.longValue();
```
smillst: Use `bigFrom.max(longMin).longValue();` instead.

framework/src/org/checkerframework/common/value/util/Range.java
```diff
@@ -853,21 +892,35 @@ private boolean isWithinInteger() {
...
+                BigInteger longMax = BigInteger.valueOf(Long.MAX_VALUE);
+                resultTo = bigTo.compareTo(longMax) > 0 ? Long.MAX_VALUE : bigTo.longValue();
```
smillst: Use `bigFrom.min(longMin).longValue();` instead.

framework/src/org/checkerframework/common/value/util/Range.java
```diff
@@ -20,6 +20,9 @@
     /** The upper bound of the interval, inclusive. */
     public final long to;
 
+    /** Should this range take overflow into account or ignore it? */
+    public static boolean IGNORE_OVERFLOW = false;
```
smillst: I changed this to a static field rather, so that it doesn't have to be passed around everywhere.  Let me if you disagree with this change.

framework/src/org/checkerframework/common/value/util/Range.java
```diff
@@ -109,23 +112,35 @@ public boolean isNothing() {
...
      */
     public Range intRange() {
-        if (this.isWiderThan(integerWidth)) {
-            return INT_EVERYTHING;
+        if (this.isNothing()) {
+            return NOTHING;
+        }
+        if (IGNORE_OVERFLOW) {
+            return new Range(
+                    from < Integer.MIN_VALUE ? Integer.MIN_VALUE : from,
```
smillst: Could you just use the max function and the min function below?  I find this a little hard to read.

framework/tests/value-ignore-range-overflow/ValueNoOverflow.java
```diff
@@ -0,0 +1,29 @@
...
+        @IntRange(to = -1) int y = x - 1; // IntRange(from = 0) to GTEN1
+        @IntRange(to = 0) int w = z - 1; // Pos to NN
+    }
+
+    void test_mult(@IntRange(from = 0) int x, @IntRange(from = 1) int z) {
+        @IntRange(from = 0) int y = x * z;
+        @IntRange(from = 1) int w = z * z;
+    }
+
+    void testLong(@IntRange(from = 0) long x, @IntRange(from = 1) long z) {
```
smillst: Plus add bytes and shorts and more long tests.

framework/src/org/checkerframework/common/value/util/Range.java
kelloggm: This seems like it's a much better solution.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -777,7 +800,10 @@ private Range calculateRangeUnaryOp(
             TransferInput<CFValue, CFStore> p,
             CFStore thenStore,
             CFStore elseStore) {
-        if (isIntRange(leftNode, p) || isIntRange(rightNode, p)) {
+        if (isIntRange(leftNode, p)
+                || isIntRange(rightNode, p)
+                || isIntegralUnknownVal(rightNode, p)
```
smillst: I copied the `isIntegralUnknownVal` method to #1222.

kelloggm:
This is now waiting on typetools#1222 being merged, since it's dependent on that.

docs/manual/constant-value-checker.tex
```diff
@@ -251,6 +251,18 @@ \section{Warnings\label{value-checker-warnings}}
 
 \end{itemize}
 \end{sloppypar}
+
+\section{Overflow\label{value-checker-overflow}}
+\begin{sloppypar}
```
mernst: I don't see the point of the sloppypar.  The manual formats identically with and without it.

docs/manual/constant-value-checker.tex
```diff
@@ -251,6 +251,18 @@ \section{Warnings\label{value-checker-warnings}}
...
+The Constant Value Checker normally takes Java's normal overflow rules into account when
+computing the possible values of expressions. However, the overflow rules can be turned off
+for range annotations (\<@IntRange>, for instance) if desired. This is achieved by passing the
+\code{-AignoreRangeOverflow} flag to the Value Checker. Ignoring overflow is useful for some checkers
```
mernst: How is it useful?  How does a programmer make the decision of whether to use the flag or not?

framework/src/org/checkerframework/common/value/util/Range.java
```diff
@@ -853,26 +886,39 @@ private boolean isWithinInteger() {
...
      * return EVERYTHING.
      *
-     * <p>If the BigInteger bounds are out of the Long type scope, convert the bounds to Long type
-     * in accordance with Java overflow rules, e.g., Long.MAX_VALUE + 1 is converted to
-     * Long.MIN_VALUE.
+     * <p>If the BigInteger bounds are out of the Long type scope and this range does not ignore
+     * overflow, convert the bounds to Long type in accordance with Java overflow rules, e.g.,
+     * Long.MAX_VALUE + 1 is converted to Long.MIN_VALUE.
+     *
+     * <p>If the BigInteger bounds are out of the Long type scope and this range ignores overflow,
```
mernst: What is a "type scope"?

smillst:
I just thought of another place that needs to take ignore overflow mode into account: `ValueQualifierHierarchy#widenUpperBound`.  Here's an example of why:

```
void test(int c, int max) {
  int decexp = 0;
  int seendot = 0;
  int i = 0;
  while (true) {
    if (c == '.' && seendot == 0) {
      seendot = 1;
    } else if ('0' <= c && c <= '9') {
      decexp += seendot;
    }

    if(max < i++) {
      break;
    }
  }
// If ignore overflow mode, then decexp should be @IntRange(from=0), 
// but decexp is @IntRange(from=-2147483648, to=2147483647) here.
  @IntRange(from=0) int x = decexp; 
```

So widenUpperBound should only change on bound of the range at a time.  I have an idea how to do this, but it will be faster for me to just change the code directly.  (So I'll do that.)

framework/build.xml
```diff
@@ -614,6 +614,13 @@
         </antcall>
     </target>
 
+    <target name="value-tests-ignore-range-overflow" depends="jar,build-tests"
+            description="Run tests for constant value propagation with no overflow for ranges">
+        <antcall target="-run-tests">
+            <param name="param" value="tests.ValueIgnoreRangeOverflowTest"/>
```
smillst: This file isn't checked in.

kelloggm:
I checked it in. Sorry about that.

On Fri, Apr 14, 2017 at 3:09 PM, Suzanne Millstein <notifications@github.com

framework/src/org/checkerframework/common/value/util/Range.java
kelloggm: I changed the wording to be more clear. This is the wording the original comment on this method used before I changed it, but I agree that I ought to have made it more precise when editing it.

kelloggm:
@mernst I've made the change you requested, and I'd appreciate it if you could take another look.

docs/manual/constant-value-checker.tex
```diff
@@ -251,6 +251,20 @@ \section{Warnings\label{value-checker-warnings}}
...
+
+
+\section{Overflow\label{value-checker-overflow}}
+
+The Constant Value Checker normally takes Java's normal overflow rules into account when
+computing the possible values of expressions.
+The \code{-AignoreRangeOverflow} command-line option
+makes it ignore the possibility of overflow
+for range annotations (\<@IntRange>, for instance).
+Ignoring overflow is useful for some checkers
```
mernst: I had a request on a previous iteration of this code review for an explanation of how a user decides whether to supply this flag.  If we cannot explain how a user should make that decision and what good it does, then it should not be user-visible option.  And, without that information, it's hard for me to review the code to assess whether it achieves that goal.

mernst:
@kelloggm Thanks for the one change, but the more important one is still outstanding.  I have marked it again.

docs/manual/constant-value-checker.tex
mernst: Also, if there is a new user-visible command-line argument, please mention it in the changelog.

kelloggm: Sorry I missed this comment before. I've updated the manual. Should I make an entire new changelog entry? It looks like all of the recent ones are releases, but this isn't a release...

docs/manual/constant-value-checker.tex
```diff
@@ -251,6 +251,39 @@ \section{Warnings\label{value-checker-warnings}}
...
+
+\section{Overflow\label{value-checker-overflow}}
+
+The Constant Value Checker normally takes Java's normal overflow rules into account when
+computing the possible values of expressions.
+The \code{-AignoreRangeOverflow} command-line option
+makes it ignore the possibility of overflow
+for range annotations (\<@IntRange>, for instance), making them more precise.
+You can supply the \code{-AignoreRangeOverflow} command-line option
+when you are confident that no overflow will occur in the program (for
```
mernst: @kelloggm What is the tradeoff between writing upper bound annotations and using `-AignoreRangeOverflow`?  That is, would it always be possible to write upper bound annotations instead of using `-AignoreRangeOverflow`?  Is the reason for the command-line argument that it too much of a hassle to write the upper bound annotations, or is there a different motivation?

kelloggm: I'm not sure it would always be possible to write upper bound annotations instead of using `-AignoreRangeOverflow`. It's possible that a programmer might not know the precise upper bound but are still confident that there will not be overflow; this is the assumption that the Index Checker makes.

From a design perspective, I view this as separation of concerns - checking that there is no overflow is one thing, and doing constant propagation a la the value checker is a different thing. If they're tightly coupled (as they are without the `-AignoreRangeOverflow` flag) then there is no way to get one without the other, which is what we need in the Index Checker.

framework/src/org/checkerframework/common/value/util/Range.java
```diff
@@ -853,26 +886,39 @@ private boolean isWithinInteger() {
...
+            resultFrom = bigFrom.max(longMin).longValue();
+            BigInteger longMax = BigInteger.valueOf(Long.MAX_VALUE);
+            resultTo = bigTo.min(longMax).longValue();
         } else {
-            long resultFrom = bigFrom.longValue();
-            long resultTo = bigTo.longValue();
-            if (resultFrom <= resultTo) {
-                return new Range(resultFrom, resultTo);
-            } else {
+            if (numValues.compareTo(longWidth) == 1) {
```
mernst: `compareTo` is allowed to return any positive value.  So, you should compare its result to 0, but don't assume it returns 1.

docs/manual/manual.bbl
```diff
@@ -34,8 +34,8 @@ Joe Darcy.
 Werner Dietl, Stephanie Dietzel, Michael~D. Ernst, K{\i}van{\c{c}} Mu{\c{s}}lu,
   and Todd Schiller.
 \newblock Building and using pluggable type-checkers.
-\newblock In {\em ICSE: Proceedings of the 33rd International Conference on
-  Software Engineering}, pages 681--690, Waikiki, Hawaii, USA, May 2011.
+\newblock In {\em ICSE 2011, Proceedings of the 33rd International Conference
```
mernst: The changes in this file are unrelated to the main changes in your pull request.  Please make a pull request contain only one set of changes.

docs/manual/constant-value-checker.tex
kelloggm: Thinking about this more, I suppose it is a convenience thing in the end; if the programmer reasons "I'm confident this isn't going to go above 10,000 or so, but I'm not sure exactly what the upper bound is exactly" (which is the sort of thing I'm thinking of), then they _could_ do the extra work to figure out what the bound exactly is, or they could instead use `-AignoreRangeOverflow`. I care more about the second argument that I made above, though - there's not any particular reason to couple overflow checking with some other checker (the value checker or the index checker or any other checker that doesn't explicitly check overflow). We don't try to reason about nullness in the value checker. Why would we try to reason about overflow?

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -331,19 +332,48 @@ public boolean implementsWidening() {
         public AnnotationMirror widenUpperBound(AnnotationMirror a1, AnnotationMirror a2) {
             AnnotationMirror lub = leastUpperBound(a1, a2);
             if (AnnotationUtils.areSameByClass(lub, IntRange.class)) {
-                Range range = getRange(lub);
-                if (range.isWithin(Byte.MIN_VALUE, Byte.MAX_VALUE)) {
+                Range lubRange = getRange(lub);
+                if (Range.IGNORE_OVERFLOW) {
+                    Range range1 = getRangeOrConvertIntVal(a1);
```
kelloggm: @smillst I don't entirely understand why you added this code, either. Can you add the comment Mike requested?

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1403,6 +1436,19 @@ public static Range getRange(AnnotationMirror rangeAnno) {
     }
 
     /**
+     * Returns a {@code Range} bounded by the values specified in the given {@code @IntRange} or
```
smillst: (Fixed in next line.)

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
smillst: I'll address this.  I'm also going to add more test cases.

smillst: This whole methods needs more thought -- both when ignoring and not ignoring overflow.  I'm going to do that in another pull request, but I don't think that should block this pull request from being merged.  We can either revert this change (and skip the test associated with it) or just leave it as is.

docs/manual/constant-value-checker.tex
mernst: The programmer doesn't need to know the exact bound in order to write correct annotations.

The reason the Value Checker reasons about overflow is in order to provide a sound estimate.  It might seem on first blush that initialization is independent of nullness, so the Nullness Checker should not reason about initialization; but if it did not, then it would be unsound.  Likewise, the Value Checker would be unsound if it did not reason about the possibility of overflow.

kelloggm: I agree that this makes the Value Checker unsound - that's the whole point. The Index Checker is built on the assumption that it doesn't have to check overflow, and integrating the significant parts of the Index Checker into the Value Checker means that we need that assumption to be an option in the Value Checker. That's the fundamental reason for this change. I'm not sure how else to describe that in the manual, though - maybe the flag just shouldn't be user visible? I can make that change.

mernst: Even if the Index Checker to be built on the assumption that it doesn't have to check overflow, that's different from it being built on the assumption that overflow does not occur.  You have also changed the discussion from being about the Value Checker to being about the Index Checker -- which is this really about?

I think you can clearly explain what is going on, and I would like you to do so even if it isn't easy.  So far, the patches and the discussion have seemed to be about putting in hacks to prevent warnings, without necessarily understanding the implications of those hacks.  Let's make the design proceed from an understanding of the issues, rather than making changes and then attempting to triangulate back to a design that would justify those changes.

Thanks.

framework/src/org/checkerframework/common/value/util/Range.java
```diff
@@ -20,6 +20,9 @@
     /** The upper bound of the interval, inclusive. */
     public final long to;
 
+    /** Should this range take overflow into account or ignore it? */
```
mernst: Since this is a static variable, the documentation about "this range" is misleading.
Should the variable be instance or the documentation be changed?

framework/src/org/checkerframework/common/value/util/Range.java
```diff
@@ -109,23 +112,33 @@ public boolean isNothing() {
     /**
      * Converts a this range to a 32-bit integral range.
      *
-     * <p>If this range is too wide, i.e., wider than the full range of the Integer class, return
-     * INT_EVERYTHING.
+     * <p>If this range ignores overflow and one of the bounds is outside the Integer range, then
```
mernst: Here again, there is a comment about "this range", but it's controlled by a static variable.  Please clarify.

framework/src/org/checkerframework/common/value/util/Range.java
```diff
@@ -109,23 +112,33 @@ public boolean isNothing() {
...
+     * that bound is set to the bound of the Integer range.
```
mernst: "the bound of the Integer range" is confusing because both "bound" and "range" are being used with different meanings than elsewhere in the sentence (correct English, but still a different meaning).  I would say MIN_VALUE or MAX_VALUE.
Or you can say the extent of the Integer class (here and elsewhere), or some other wording that you use consistently.

framework/src/org/checkerframework/common/value/util/Range.java
kelloggm: The documentation needed to be changed. We originally had this as an instance variable, and decided to switch to a static variable because it makes the implementation cleaner.

kelloggm:
@mernst I believe I've addressed all of your comments on this code, and would appreciate it if you would take another look.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1408,6 +1441,20 @@ public static Range getRange(AnnotationMirror rangeAnno) {
...
+     * {@code @IntVal} annotation. Returns null if the given annotation is not {@code IntRange} or
+     * {@code IntVal}.
+     */
+    public static Range getRangeOrConvertIntVal(AnnotationMirror anno) {
```
mernst: I think this method should be folded into `getRange`.  Then, clients don't have to think about which of the two to call.  The two have essentially identical specifications.  That will also address the problem that this one's name says too much about the implementation.

framework/src/org/checkerframework/common/value/util/Range.java
```diff
@@ -109,23 +112,33 @@ public boolean isNothing() {
...
      */
     public Range intRange() {
-        if (this.isWiderThan(integerWidth)) {
-            return INT_EVERYTHING;
+        if (this.isNothing()) {
+            return NOTHING;
+        }
+        if (IGNORE_OVERFLOW) {
+            return new Range(Math.max(from, Integer.MIN_VALUE), Math.min(to, Integer.MAX_VALUE));
         } else {
```
mernst: In general, I often don't use an `else` clause after a return statement, both because it causes deep indentation and because that indentation makes me keep in mind what else may have occurred, rather than knowing that everything that has gone before is irrelevant.  I see that you did this for the `isNothing` special case above.

kelloggm:
There is still the issue of casts. This PR includes a partial solution (which Mike and I spoke about via email) that we generally find unsatisfactory (the relevant code is in ValueVisitor). There is an issue for this problem in general: #1264. We can either wait to merge this until that issue is solved, or we can remove the code in ValueVisitor and mark the tests that fail as tests for #1264 if we merge now.

smillst:
>We can either wait to merge this until that issue is solved, or we can remove the code in ValueVisitor and mark the tests that fail as tests for #1264 if we merge now.

Let's get this pull request merged and not wait to fix #1264.

kelloggm:
Alright, I've removed the hacky code and added a failing test provided by Mike back when we were talking about this, with a pointer to #1264. When tests pass this should be merge-able.

framework/tests/value/ValueCast.java
```diff
@@ -0,0 +1,11 @@
+// Test case for issue 1264: https://github.com/typetools/checker-framework/issues/1264
```
smillst: #1264 is about casts, but this test case doesn't have a cast in it.  I think it will not be fixed by the fix for that issue.

kelloggm: This is an implicit cast from `short` to `int`. I would expect that a solution to #1264 would also address this issue. If you really think it's a different issue, we can add it as one, but I think that they are two sides of the same coin.

smillst: To fix #1264, the Value Checker only needs to override org.checkerframework.common.basetype.BaseTypeVisitor#checkTypecastSafety and checkTypecastRedundancy so that cast unsafe warnings aren't issued.  

To fix this example, the declared type of `x` needs to be changed to `@IntRange(from=0, to = Short.MAX_VALUE)`.  I think the general problem in this example is that `@IntRange(from = 0) short` should be assumed to be `@IntRange(from = 0, to = Short.MAX_VALUE ) short`

kelloggm: I can see why you view them as separate issues. I will create another issue for this test case and change this reference.

kelloggm: I created a new issue: #1299

smillst: Thank you!

kelloggm: I actually just went ahead and fixed #1299 and opened a PR #1300. We shouldn't merge this until that's in.

framework/tests/value-ignore-range-overflow/Widen.java
```diff
@@ -0,0 +1,25 @@
...
+                    seendot = 1;
+                } else if ('0' <= c && c <= '9') {
+                    decexp += seendot;
+                }
+
+                if (max < i++) {
+                    break;
+                }
+            }
+            // @IntRange(from = 0) int x = decexp;
```
wmdietl: Should this be un-commented or deleted?

framework/tests/value-ignore-range-overflow/TransferTimes.java
```diff
@@ -0,0 +1,27 @@
...
+        @IntRange(from = 1) int j = g * a;
+
+        int k = 0;
+        int l = 1;
+        @IntRange(from = 1) int m = a * l;
+        @IntRange(from = 0) int n = k * l;
+        @IntRange(from = 0) int o = k * k;
+    }
+}
+//a comment
```
wmdietl: Not sure the comment is necessary.

framework/tests/value-ignore-range-overflow/TransferSub.java
```diff
@@ -0,0 +1,73 @@
...
+        //:: error: (assignment.type.incompatible)
+        @IntRange(from = 1) int t1 = --k1;
+
+        int u1 = -1;
+        @IntRange(from = -1) int x = u1--;
+        //:: error: (assignment.type.incompatible)
+        @IntRange(from = -1) int x1 = u1;
+    }
+}
+//a comment
```
wmdietl: Remove.

framework/tests/value-ignore-range-overflow/TransferSub.java
```diff
@@ -0,0 +1,73 @@
...
+
+        @IntRange(from = 0) int r = k--; // after this k = -1
+
+        int k1 = 0;
+        @IntRange(from = 0) int s = k1--;
+
+        //:: error: (assignment.type.incompatible)
+        @IntRange(from = 0) int s1 = k1;
+
+        // transferred to SimpleTransferSub.java
```
wmdietl: Can you open a bug and put this in a skipped test case instead?

framework/tests/value-ignore-range-overflow/TransferMod.java
```diff
@@ -0,0 +1,32 @@
...
+
+        @IntRange(from = 0) int n = c % a;
+        @IntRange(from = 0) int o = b % a;
+
+        @IntRange(from = -1) int p = a % a;
+        @IntRange(from = -1) int q = a % d;
+        @IntRange(from = -1) int r = a % c;
+    }
+}
+//a comment
```
wmdietl: Remove.

framework/tests/value-ignore-range-overflow/TransferDivide.java
```diff
@@ -0,0 +1,59 @@
...
+
+        /** n1p / nn -> n1p */
+        @IntRange(from = -1) int x = a / l;
+    }
+
+    void testDivideByTwo(@IntRange(from = 0) int x) {
+        @IntRange(from = 0) int y = x / 2;
+    }
+}
+//a comment
```
wmdietl: Remove.

framework/tests/value-ignore-range-overflow/TransferAdd.java
```diff
@@ -0,0 +1,83 @@
...
+
+        @IntRange(from = 0) int x1 = n1p2;
+
+        //:: error: (assignment.type.incompatible)
+        @IntRange(from = 1) int y = ++d;
+        //:: error: (assignment.type.incompatible)
+        @IntRange(from = 1) int z = e++;
+    }
+}
+//a comment
```
wmdietl: Remove.

framework/tests/value-ignore-range-overflow/RefinementNEq.java
```diff
@@ -0,0 +1,31 @@
...
+
+        if (1 != s) {
+            //:: error: (assignment.type.incompatible)
+            @IntRange(from = 1) int t = s;
+        } else {
+            @IntRange(from = 1) int u = s;
+        }
+    }
+}
+//a comment
```
wmdietl: Remove.

framework/tests/value-ignore-range-overflow/RefinementLTE.java
```diff
@@ -0,0 +1,57 @@
...
+
+        if (s <= 1) {
+            //:: error: (assignment.type.incompatible)
+            @IntRange(from = 1) int t = s;
+        } else {
+            @IntRange(from = 1) int u = s;
+        }
+    }
+}
+//a comment
```
wmdietl: Remove.

framework/tests/value-ignore-range-overflow/RefinementLT.java
```diff
@@ -0,0 +1,57 @@
...
+
+        if (s < 1) {
+            //:: error: (assignment.type.incompatible)
+            @IntRange(from = 1) int t = s;
+        } else {
+            @IntRange(from = 1) int u = s;
+        }
+    }
+}
+//a comment
```
wmdietl: Remove.

framework/tests/value-ignore-range-overflow/RefinementGTE.java
```diff
@@ -0,0 +1,50 @@
...
+
+        if (1 >= s) {
+            //:: error: (assignment.type.incompatible)
+            @IntRange(from = 1) int t = s;
+        } else {
+            @IntRange(from = 1) int u = s;
+        }
+    }
+}
+//a comment
```
wmdietl: Remove.

framework/tests/value-ignore-range-overflow/RefinementGT.java
```diff
@@ -0,0 +1,59 @@
...
+
+        if (1 > s) {
+            //:: error: (assignment.type.incompatible)
+            @IntRange(from = 1) int t = s;
+        } else {
+            @IntRange(from = 1) int u = s;
+        }
+    }
+}
+//a comment
```
wmdietl: Remove.

framework/tests/value-ignore-range-overflow/RefinementEq.java
```diff
@@ -0,0 +1,29 @@
...
+
+        if (1 == s) {
+            @IntRange(from = 1) int t = s;
+        } else {
+            //:: error: (assignment.type.incompatible)
+            @IntRange(from = 1) int u = s;
+        }
+    }
+}
+//a comment
```
wmdietl: Remove.

framework/src/org/checkerframework/common/value/util/Range.java
```diff
@@ -20,6 +20,9 @@
     /** The upper bound of the interval, inclusive. */
     public final long to;
 
+    /** Should ranges take overflow into account or ignore it? */
```
wmdietl: This public, static mutable state is not nice and in general we try to avoid having this. It prevents multiple checkers from running independently.
One option would be to pass in the Checker instance in the constructor and set an instance field.

framework/tests/src/tests/ValueIgnoreRangeOverflowTest.java
```diff
@@ -0,0 +1,35 @@
...
+import java.io.File;
+import java.util.List;
+import org.checkerframework.common.value.ValueChecker;
+import org.checkerframework.framework.test.CheckerFrameworkPerDirectoryTest;
+import org.junit.runners.Parameterized.Parameters;
+
+/**
+ * Tests the constant value propagation type system without overflow.
+ *
+ * <p>NOTE: $CHECKERFRAMEWORK/framework/tests/value/ needs to be on the classpath. Otherwise
```
wmdietl: Where is this done? I don't see a classpath option here or in framework/build.xml .
If you only need this for one file, you should make it a jtreg test.

framework/tests/src/tests/ValueIgnoreRangeOverflowTest.java
```diff
@@ -0,0 +1,35 @@
...
+ * ExceptionTest will fail because it cannot find the ExceptionTest.class file for reflective method
+ * resolution.
+ *
+ * @author plvines
```
wmdietl: How come?

kelloggm: I don't know why this comment is here. I'm just going to delete it; I don't see any evidence that the claim it's making is true, either.

framework/src/org/checkerframework/common/value/util/Range.java
kelloggm: We avoided doing this because it results in the `Range` class becoming quite bloated: most Range operations produce another Range, so the value of `ignoreOverflow` as an instance field would need to be passed through.

With the current design, any checker that uses the `Range` library should set the value of the `IGNORE_RANGE_OVERFLOW`. This permits reuse. That said, I agree that it's undesirable to force clients to set that field because it allows for the possibility of error.

I view both designs as having non-trivial disadvantages. Do you believe we should switch back to an instance-field design? @smillst, did I forget an important argument for why this should be a static field?

wmdietl: I would add a comment calling out at least this basic trade-off.
Using the static mutable field doesn't allow you to use the Range class in different contexts, where one might care about overflow but the other doesn't. That would make composition of different checkers impossible.
I agree that passing around the option isn't nice either.

kelloggm: I've added a comment that calls out the tradeoff.

kelloggm:
I expect that the test cases should pass now that #1309 has been merged in (as that was the cause of test failures before).

kelloggm:
I had to add some additional code in `ValueVisitor.java` to correctly handle casts where one of the sides of a range could overflow. This is similar to the code introduced in #1309.

kelloggm:
@wmdietl It looks like your review on this is the only outstanding one. Can you take another look? Thanks!

framework/src/org/checkerframework/common/value/ValueVisitor.java
```diff
@@ -181,6 +182,9 @@ public Void visitTypeCast(TypeCastTree node, Void p) {
                         .getAnnotatedType(node.getExpression())
                         .getAnnotationInHierarchy(atypeFactory.UNKNOWNVAL);
 
+        System.out.println("castAnno: " + castAnno);
```
wmdietl: Remove debugging output.

framework/src/org/checkerframework/common/value/ValueVisitor.java
```diff
@@ -202,8 +206,27 @@ public Void visitTypeCast(TypeCastTree node, Void p) {
             if (castType.getKind() == TypeKind.LONG && castRange.isLongEverything()) {
                 return p;
             }
+            if (Range.IGNORE_OVERFLOW) {
+                // Range.IGNORE_OVERFLOW is only set if this checker is ignoring overflow.
+                // In that case, compress casts to the range of the underlying type.
```
wmdietl: I'm not sure what the connection between casts and overflow is. Can you expand on what `compress` means? And what happens if the flag is not given?

framework/tests/value-ignore-range-overflow/RefinementGT.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class RefinementGT {
+
+    void test_forward(int a, int j, int s) {
+        /** forwards greater than */
+        //:: error: (assignment.type.incompatible)
+        @IntRange(from = 0) int aa = a;
+        if (a > -1) {
+            /** a is NN now */
+            @IntRange(from = 0) int b = a;
```
wmdietl: Can you add a test that makes sure -1 and 1 behave as expected? (-1 passes, but 1 fails?)

framework/src/org/checkerframework/common/value/ValueVisitor.java
kelloggm: This is necessary to suppress `cast.unsafe` warnings that appear on code like this (a pathological example from `ValueNoOverflow.java` test case:

```java
void testShort_plus(@IntRange(from = 0) short x, @IntRange(from = -1) short z) {
        @IntRange(from = 1) short y = (short) (x + 1); // IntRange(from = 0) to IntRange(from = 1)
        @IntRange(from = 0) short w = (short) (z + 1); // GTEN1 to NN
    }
```

I can change the comment here to explain this better.

framework/tests/value-ignore-range-overflow/RefinementGT.java
kelloggm: I've added additional tests to all of these tests that should cover these cases.

kelloggm:
I believe I've addressed all comments.

checker/jdk/nullness/src/java/lang/Object.java
```diff
@@ -10,7 +10,7 @@
   public Object() { throw new RuntimeException("skeleton method"); }
   @Pure public boolean equals(@Nullable Object a1) { throw new RuntimeException("skeleton method"); }
   // toString() is @SideEffectFree rather than @Pure because it returns a string
-  // that differs according to ==, and @Deterministic is with respect to ==.
+  // that differs according to ==.  It would be @Deterministic is with respect to ==.
```
wmdietl: Remove `is`?

kelloggm: I don't know why there are changes to this - it's unrelated and probably an artifact of this PR being fairly old. I'll remove this change.

PR #1235, opened 2017-04-17
Update error-prone, fix reported issues, and run on travis

wmdietl:
Most fixes are simply adding default branches to switches.
However, it also uncovered a real issue in the logic in framework/src/org/checkerframework/framework/util/AtmLubVisitor.java . I simplified the code; let's see whether the tests still work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ language: generic
 # Alternative to "language: java", which has node installed
 before_install:
   - sudo apt-get update -qq
-  - sudo apt-get install -qq nodejs
+  - sudo apt-get install -qq nodejs maven
 
 services:
   - docker
--- a/framework/src/org/checkerframework/framework/util/AtmLubVisitor.java
+++ b/framework/src/org/checkerframework/framework/util/AtmLubVisitor.java
@@ -114,6 +114,8 @@ class AtmLubVisitor extends AbstractAtmComboVisitor<Void, AnnotatedTypeMirror> {
         Set<? extends AnnotationMirror> lubSet;
         if (type1.getAnnotations().isEmpty()) {
             lubSet = type2.getAnnotations();
+        } else if (type2.getAnnotations().isEmpty()) {
+            lubSet = type1.getAnnotations();
         } else {
             lubSet =
                     qualifierHierarchy.leastUpperBounds(
--- a/framework/src/org/checkerframework/framework/util/AtmLubVisitor.java
+++ b/framework/src/org/checkerframework/framework/util/AtmLubVisitor.java
@@ -20,7 +20,6 @@ import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedWildcard
 import org.checkerframework.framework.type.QualifierHierarchy;
 import org.checkerframework.framework.type.visitor.AbstractAtmComboVisitor;
 import org.checkerframework.framework.type.visitor.VisitHistory;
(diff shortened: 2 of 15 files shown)


framework/src/org/checkerframework/framework/util/AtmLubVisitor.java
```diff
@@ -115,14 +112,8 @@ private AnnotatedTypeMirror lubWithNull(
     private void lubPrimaryAnnotations(
             AnnotatedTypeMirror type1, AnnotatedTypeMirror type2, AnnotatedTypeMirror lub) {
         Set<? extends AnnotationMirror> lubSet;
-        if (visitingUninferedWildcard
-                        && type1.getAnnotations().size() != type1.getAnnotations().size()
```
wmdietl: This condition uses type1 twice, making it always false, making the whole first clause false, and effectively only leaving the `type1.getAnnotations().isEmpty()`.
Then, because type1 has no annotations, the nested condition can also be simplified.
As this was the only place that used visitingUninferedWildcard, I also removed that variable and the helper method.
@smillst Does this change look ok to you?

wmdietl:
The code tests pass: https://travis-ci.org/typetools/checker-framework/builds/222682038
I'm installing Maven in Ubuntu, let's see whether that fixes the final failing configuration.

.travis.yml
```diff
@@ -5,7 +5,7 @@ language: generic
 # Alternative to "language: java", which has node installed
 before_install:
   - sudo apt-get update -qq
-  - sudo apt-get install -qq nodejs
+  - sudo apt-get install -qq nodejs maven
```
mernst: These programs are for use by the Travis process.
If you want to install mvn for use in the Docker container, that needs to be done in `checker/bin-devel/Dockerfile-ubuntu-jdkany` (and the new Docker container built and uploaded).
I can do that for you.

framework/src/org/checkerframework/framework/util/AtmLubVisitor.java
mernst: Should it have been "type*1*... != type*2*"?  Did you try that change?  From the surrounding code seems like the most likely typo, rather than the typo being that the condition shouldn't have been there at all.

@kelloggm You might want to chime in, and to create a test that relies on this behavior.

smillst: Although @kelloggm is the author this change, I actually made it as part of the Index Checker merge.  (Though, I thought I moved all changes to the framework out of the Index Checker branch before merging. ) Anyways, Mike is right, one of the `type1` should be `type2`.  I'll fix this in a separate PR.

mernst: @wmdietl It sounds like you should undo this change.

smillst: I fixed this in #1238.

wmdietl: I tried with type1 != type2 as condition but that never got triggered.
#1238 doesn't contain a test case that exercises this branch. I think it is much better to simplify this code and find when this complexity is actually needed.
The variable name 'visitingUninferedWildcard' is also rather poorly chosen - usually we use 'uninferred' for method type arguments that weren't inferred, but here it is used for raw types.

PR #1238, opened 2017-04-17
Correct lub for uninfered wildcards 

smillst:
framework/tests/all-systems/LubRawTypes.java tests the case where `getAnnotations()` is empty.  I'm not sure if the case where `type1.getAnnotations().size() != type2.getAnnotations().size()` can happens in practice, but if it does, it shouldn't cause a crash.

This corrects a problem @wmdietl found in #1235.

### DIFF
--- a/framework/src/org/checkerframework/framework/util/AtmLubVisitor.java
+++ b/framework/src/org/checkerframework/framework/util/AtmLubVisitor.java
@@ -119,6 +119,9 @@ class AtmLubVisitor extends AbstractAtmComboVisitor<Void, AnnotatedTypeMirror> {
                 && (type1.getAnnotations().size() != type2.getAnnotations().size()
                         || type1.getAnnotations().isEmpty()
                         || type2.getAnnotations().isEmpty())) {
+            // qualifierHierarchy.leastUpperBounds errors if the sets of annotations passed are
+            // not the same size and non-empty.  But, if either type is from an uninferred
+            // wildcard, then avoid this crash.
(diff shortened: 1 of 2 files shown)


wmdietl:
I would prefer the simpler solution in #1235 . If we find a crash, we can see what the correct solution is. As all tests, including Daikon, work in #1235, maybe this isn't necessary.

smillst:
I added a test case where the two sets are different sizes.  So, this is the correct change.

framework/src/org/checkerframework/framework/util/AtmLubVisitor.java
```diff
@@ -116,8 +116,12 @@ private void lubPrimaryAnnotations(
             AnnotatedTypeMirror type1, AnnotatedTypeMirror type2, AnnotatedTypeMirror lub) {
         Set<? extends AnnotationMirror> lubSet;
         if (visitingUninferedWildcard
-                        && type1.getAnnotations().size() != type1.getAnnotations().size()
-                || type1.getAnnotations().isEmpty()) {
+                && (type1.getAnnotations().size() != type2.getAnnotations().size()
```
wmdietl: Thanks for the additional test!
`visitingUninferedWildcard` can only be true if at least one of the two annotation sets is empty, right?
So you can remove the partialQualifierLub method, see which one of the two is empty and then use the other one as LUB. I've updated #1235 to check whether either side is empty and expanded the test case to cover both cases. The test passed locally.

PR #1242, opened 2017-04-18
Fix minor issues reported by error-prone

mernst:


framework/src/org/checkerframework/common/value/Operators.java
```diff
@@ -399,11 +399,11 @@ public static int UNSIGNED_RIGHT_SHIFT(int a, int b) {
     }
 
     public static boolean AND(Boolean a, Boolean b) {
-        return a & b;
+        return a && b;
```
smillst: The changes in this file are wrong.   (These methods are suppose to compute bitewise operations.  The method name matches the name of the Tree.Kind.)  However, this file isn't used any more, so I'll just remove it.

mernst: Thanks for noticing.

PR #1243, opened 2017-04-18
Move the functions of the Index Checker's MinLen Checker into the Value Checker

kelloggm:
This change makes `@MinLen(n)` an alias for `@ArrayLenRange(from = n)`, and updates the Index Checker to use this version instead of the origin MinLen Checker. The MinLen Checker is removed.

`@PolyMinLen` is removed; it will need to be replaced by `@PolyValue` when that is implemented (see typetools#1236).

I wanted to get this under review sooner rather than later, since it's a large change.


smillst:
The test are failing because the jdk download has `@PolyMinLen` still.  I'll correct this.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -925,7 +927,7 @@ private Range calculateRangeUnaryOp(
                 break;
             case GREATER_THAN_EQ:
                 thenRightRange = rightRange.refineLessThanEq(leftRange);
-                thenLeftRange = leftRange.refineGreaterThan(rightRange);
+                thenLeftRange = leftRange.refineGreaterThanEq(rightRange);
```
smillst: Is this just a bug?

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -1210,4 +1212,28 @@ private void addAnnotationToStore(CFStore store, AnnotationMirror anno, Node nod
...
+     * Node, Node, CFAbstractValue, CFAbstractValue, boolean)}, which is called by {@link
+     * org.checkerframework.framework.flow.CFAbstractTransfer#visitCase(CaseNode, TransferInput)}.
+     * Therefore, we need to explicitly call the handling for binary comparison if we want to do the
+     * extra refinements that the Value Checker applies on switch statements.
+     */
+    @Override
+    public TransferResult<CFValue, CFStore> visitCase(
+            CaseNode n, TransferInput<CFValue, CFStore> p) {
+        TransferResult<CFValue, CFStore> transferResult = super.visitCase(n, p);
+        if (n.getSwitchOperand() instanceof FieldAccessNode) {
```
smillst: Why just field accesses?

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -1210,4 +1212,28 @@ private void addAnnotationToStore(CFStore store, AnnotationMirror anno, Node nod
                 resultValues,
                 transferResult.getResultValue().getUnderlyingType());
     }
+
+    /**
+     * The value checker doesn't override {@link
+     * org.checkerframework.framework.flow.CFAbstractTransfer#strengthenAnnotationOfEqualTo(TransferResult,
+     * Node, Node, CFAbstractValue, CFAbstractValue, boolean)}, which is called by {@link
+     * org.checkerframework.framework.flow.CFAbstractTransfer#visitCase(CaseNode, TransferInput)}.
+     * Therefore, we need to explicitly call the handling for binary comparison if we want to do the
```
smillst: Please rewrite to not use "we".

framework/src/org/checkerframework/common/value/qual/MinLen.java
```diff
@@ -1,18 +1,14 @@
...
 
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Target;
-import org.checkerframework.framework.qual.DefaultQualifierInHierarchy;
-import org.checkerframework.framework.qual.SubtypeOf;
 
 /**
  * The value of the annotated expression is a sequence containing at least the given number of
- * elements.
+ * elements. An alias for an {@link ArrayLenRange} annotation with only a {@code from} field.
```
smillst: Please explain what value the "to" field of ArrayLenRange has.

framework/src/org/checkerframework/common/value/qual/MinLen.java
```diff
@@ -1,18 +1,14 @@
...
  *
  * @checker_framework.manual #index-checker Index Checker
```
smillst: This should be documented in the Value Checker -- not the Index Checker.

framework/src/org/checkerframework/common/value/qual/MinLenFieldInvariant.java
```diff
@@ -1,4 +1,4 @@
-package org.checkerframework.checker.index.qual;
+package org.checkerframework.common.value.qual;
```
smillst: The Javadoc (below) is out-of-date.  Also, would it make more sense for this to be changed to ArrayLenRangeFieldInvariant?  (This is something I could do.)

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -707,10 +707,13 @@ public AnnotationMirror getAnnotationFromJavaExpressionString(
 
         AnnotationMirror annotationMirror = null;
         if (CFAbstractStore.canInsertReceiver(expressionObj)) {
-            Value value = getStoreBefore(tree).getValue(expressionObj);
-            if (value != null) {
-                annotationMirror =
-                        AnnotationUtils.getAnnotationByClass(value.getAnnotations(), clazz);
+            Store store = getStoreBefore(tree);
```
smillst: This should be in a separate PR with a test case.  It's a general bug fix to a framework class.

framework/src/org/checkerframework/common/value/qual/MinLen.java
```diff
@@ -1,18 +1,14 @@
...
  */
-@DefaultQualifierInHierarchy
-@SubtypeOf({})
 @Target({ElementType.TYPE_USE, ElementType.TYPE_PARAMETER})
 public @interface MinLen {
```
smillst: Should we rename this?  It seems out of place with  `@ArrayLen` and `@ArrayLenRange`.  Maybe `@ArrayMinLen`?

docs/manual/index-checker.tex
```diff
@@ -303,23 +303,27 @@ \section{Array minimum lengths\label{index-minlen}}
 annotated as
 \refqualclasswithparams{checker/index/qual}{LTLengthOf}{"arr"}.
 
+The MinLen type system is implemented in practice by the Constant Value Checker
```
smillst: `@MinLen` should be documented in the Value Checker section, not here.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1184,6 +1245,20 @@ private AnnotationMirror convertIntValToDoubleVal(AnnotationMirror intValAnno) {
     }
 
     /**
+     * Builds an ArrayLenRange annotation with from equal to its argument. This is called by the
+     * clients that used the old MinLen Checker, and provides the same interface.
```
smillst: Don't reference the old MinLen Checker -- just document what this does.

framework/src/org/checkerframework/common/value/ValueTransfer.java
kelloggm: Yes, this is just a bug. I probably should have PR'd it separately; I can do that if you'd like.

docs/manual/index-checker.tex
kelloggm: The Index Checker's manual section should at least include an explanation of how to use `@MinLen` and why a programmer needs it while using the Index Checker. I'm fine with moving the documentation of the annotation itself to the Value Checker's manual section, though.

checker/jdk/index/src/java/lang/String.java
```diff
@@ -162,7 +162,7 @@ public String() {
      * @param  original
      *         A {@code String}
      */
-    public @PolyMinLen @PolySameLen String(@PolyMinLen @PolySameLen String original) {
+    public /*@ PolyMinLen*/ @PolySameLen String(/*@ PolyMinLen*/ @PolySameLen String original) {
```
smillst: `@ PolyMinLen` is the same as `@PolyMinLen`, so it's getting compiled into the code and causing the tests to fail.

framework/src/org/checkerframework/common/value/ValueTransfer.java
kelloggm: This problem only came up when trying to refine `array.length` field accesses, but it is more general, so the `if` guard can (and should) probably be removed.

framework/src/org/checkerframework/common/value/ValueTransfer.java
smillst: Please do.  Since this PR is likely to take a few days before it is merged, it would be good to have this fix change in master sooner.

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
kelloggm: Actually this should just be reverted; I didn't mean to include it in the PR (and thought I'd removed it). Sorry about that.

checker/jdk/index/src/java/lang/String.java
kelloggm: Really? We use `/*@ AnnotationName*/` instead of `@AnnotationName` to comment out annotations throughout the Index Annotated JDK. If they're treated the same, I need to go through the whole Index JDK and change many things (especially the annotations on Strings, which I wrote but have since removed because we deferred String support).

framework/src/org/checkerframework/common/value/qual/MinLen.java
kelloggm: I prefer `@MinLen` still, but if you and @mernst prefer `@ArrayMinLen` or something else, we can change it. I think `@MinLen` is preferable because we'd like to eventually expand it to also be useful for Strings and possibly lists in the future.

checker/jdk/index/src/java/lang/String.java
smillst: Try this:
```
void test() {
        @MinLen(4) String s = String.valueOf(true);
    }
```
A annotation must have `@Target(TYPE_USE)` for it to be read from comments.   `@IndexFor` doesn't have that so it is ignored if it is in comments (with or without the spaces).  That's why you didn't run into this problem before.

But yes, you should probably fix that.  And IndexFor (and related annotations) should be have `@Target({TYPE_USE, TYPE_PARAMETER})`

framework/src/org/checkerframework/common/value/qual/MinLen.java
smillst: >I think @MinLen is preferable because we'd like to eventually expand it to also be useful for Strings and possibly lists in the future.

At that time `@ArrayLen` and `@ArrayLenRange` would have to be changed, too.

mernst: Good point.  I'm in favor of renaming them all now.  The longer we wait, the more annotated code will exist that has to be updated.

checker/jdk/index/src/java/lang/String.java
kelloggm: I've issued a PR for this: typetools#1247

kelloggm: The changeset for this PR will be big until typetools#1247 is merged, because it includes all the changes to the JDK in typetools#1247. I did it programmatically over the whole Index JDK directory.

framework/src/org/checkerframework/common/value/qual/MinLen.java
kelloggm: Should that change be part of this PR? I think it would be more appropriate as a separate PR. I can do it, if you'd like.

smillst: I agree, this should be a separate PR. (Probably, after this one is merged!)

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
smillst: I had to put this back in because without it there was a crash in when building the jdk.   It seems like correct code to me -- it just needs to be in a different PR.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -1210,4 +1212,30 @@ private void addAnnotationToStore(CFStore store, AnnotationMirror anno, Node nod
...
+     * Node, Node, CFAbstractValue, CFAbstractValue, boolean)}, which is called by {@link
+     * org.checkerframework.framework.flow.CFAbstractTransfer#visitCase(CaseNode, TransferInput)}.
+     * Therefore, this needs to explicitly call the handling for binary comparison to do the extra
+     * refinements that the Value Checker applies on switch statements.
+     */
+    @Override
+    public TransferResult<CFValue, CFStore> visitCase(
+            CaseNode n, TransferInput<CFValue, CFStore> p) {
+        TransferResult<CFValue, CFStore> transferResult = super.visitCase(n, p);
+        // TODO: this should work for chars too, but it is causing a crash when building the
```
smillst: I think the correct thing to do is make the Value Checker override strengthenAnnotationOfEqualTo.  If the types are numerical, call calculateBinaryComparison; otherwise, it should refine the left and right hand sides to the intersection of the known values.  I'll work on this.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1506,21 +1581,87 @@ public static Range getRange(AnnotationMirror rangeAnno) {
...
+            return null;
+        }
+    }
+
+    /**
+     * Returns the minimum value of an integral type. This is used by the Index Checker's array
+     * bounds checking routines when determining the minimum length of an array from the annotation
+     * on the length of the array.
+     */
+    public Integer getMinLenValueFromLengthType(AnnotatedTypeMirror atm) {
```
smillst: Why `LengthType`?  Currently this is only used for array lengths, but it is (or should be) a general purpose routine that other checkers could use.  I would name it something like `getMinIntegralValue` and remove the mention of the Index Checker from the Javadoc.

Oh maybe it's `LengthType` because this method returns an Integer instead of a Long?  In either case, this should return a long and then the caller of the method can downcast as needed.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1189,6 +1250,20 @@ private AnnotationMirror convertIntValToDoubleVal(AnnotationMirror intValAnno) {
     }
 
     /**
+     * Builds an ArrayLenRange annotation with {@code from} equal to its argument. The {@code to}
+     * field is set to Integer.MAX_VALUE, the default.
+     *
+     * @param val The value to use as the from in the ArrayLenRange annotation created. This value
+     *     must be non-negative.
+     */
+    public AnnotationMirror createMinLen(int val) {
```
smillst: I find this method name confusing since it is actually creating an `@ArrayLenRange`.  This method is only called once, so could you just replace that call with one to `createArrayLenRangeAnnotation(val, MAX_VALUE)`

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1506,21 +1581,87 @@ public static Range getRange(AnnotationMirror rangeAnno) {
...
+     * bounds checking routines when determining the minimum length of an array from the annotation
+     * on the length of the array.
+     */
+    public Integer getMinLenValueFromLengthType(AnnotatedTypeMirror atm) {
+        AnnotationMirror anm = atm.getAnnotationInHierarchy(UNKNOWNVAL);
+        if (AnnotationUtils.areSameByClass(anm, IntVal.class)) {
+            List<Long> possibleValues = getIntValues(anm);
+            Long minlen = Collections.min(possibleValues);
+            if (minlen < 0) {
+                return 0;
```
smillst: The caller should handle this case.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1506,21 +1581,87 @@ public static Range getRange(AnnotationMirror rangeAnno) {
...
+            } else {
+                return minlen.intValue();
+            }
+        } else if (AnnotationUtils.areSameByClass(anm, IntRange.class)) {
+            Range range = getRange(anm);
+            if (range.from < 0) {
```
smillst: The caller should handle this case.

framework/src/org/checkerframework/common/value/qual/MinLenFieldInvariant.java
smillst: Ping.

kelloggm: I'm not sure how to change this, since I don't really understand what it's doing in the first place. Should I change the name, but leave it so that it only accepts a lower bound (i.e. a MinLen)? That doesn't seem like the right thing to do, but I'm not sure what I should add to make it able to handle all of ArrayLenRange.

smillst: I updated the Javadoc.  For now, let's just leave it as is.  After this PR is merged, we can decide if we want to change it.

docs/manual/index-checker.tex
```diff
@@ -303,23 +303,27 @@ \section{Array minimum lengths\label{index-minlen}}
 annotated as
 \refqualclasswithparams{checker/index/qual}{LTLengthOf}{"arr"}.
 
+The MinLen type system is implemented in practice by the Constant Value Checker,
```
smillst: MinLen is no longer a type system. Say something like, `@MinLen` is checked by the CVC.  Then maybe add something about how the other two array annotations can be used, too.  Maybe add an example using `@ArrayLen`.

docs/manual/index-checker.tex
```diff
@@ -303,23 +303,27 @@ \section{Array minimum lengths\label{index-minlen}}
...
+using \<@ArrayLenRange> annotations (see \chapterpageref{constant-value-checker}).
+
 Array minimum lengths use the following type qualifiers
 (Figure~\ref{fig-index-array-types}):
 
 \begin{description}
```
smillst: `@MinLen` is documented in the Value Checker now, so remove this.

docs/manual/figures/minlen.svg
```diff
@@ -176,7 +176,7 @@
       <tspan
          x="460.5"
          y="466.772"
-         id="tspan4346">@MinLenBottom</tspan>
+         id="tspan4346">@BottomVal</tspan>
```
smillst: MinLen is not a type system, so remove this figure.

framework/src/org/checkerframework/common/value/ValueTransfer.java
smillst: I've done this in https://github.com/typetools/checker-framework/pull/1251.  We can't merge this pull request until that one is merged.

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
smillst: I create a PR, #1254, for this.

kelloggm:
@smillst I think I've fixed all the documentation issues you pointed out yesterday. Can you review again?

kelloggm:
Those tweaks look fine - thanks for catching those mistakes. Merging after the release on May 1 makes sense to me.

smillst:
@kelloggm Could you please update the Javadoc in IndexChecker to reflect these changes?

kelloggm:
@smillst Good catch. I've updated those docs.

checker/src/org/checkerframework/checker/index/IndexChecker.java
```diff
@@ -3,43 +3,56 @@
...
  * described, and then what each checker requests from the checkers that run before it is described.
  * The Index Checker itself is just an alias for the Upper Bound Checker, which runs last.
  *
  * <p>The checkers run in this order:
  *
- * <p>SameLen Checker, MinLen Checker, Lower Bound Checker, Upper Bound Checker
+ * <p>Constant Value Checker, SameLen Checker, SearchIndex Checker, Lower Bound Checker, Upper Bound
+ * Checker
+ *
+ * <p>The Constant Value Checker has no dependencies, but it does trust Positive annotations from
```
smillst: Technically, this paragraph should be added in https://github.com/kelloggm/checker-framework/pull/155, but it's fine with me if you want to leave it here, too.

smillst:
IndexChecker Javadoc looks good.  Thanks!

checker/src/org/checkerframework/checker/index/IndexChecker.java
kelloggm: Oops, I forgot which PR this was! I have to go to class now, but I'll change it later today.

kelloggm:
@smillst Now that the release has passed are we okay to merge this?

PR #1244, opened 2017-04-19
Documentation improvements for "How to create a new checker"

Calvin-L:
These changes clarify some points that I think are confusing for new developers.

### DIFF
--- a/docs/manual/Makefile
+++ b/docs/manual/Makefile
@@ -71,7 +71,7 @@ manual.html: manual.pdf CFLogo.png favicon-checkerframework.png ../api
 	sed -i -e "s%<style type=\"text/css\">%<style type=\"text/css\">\nimg { max-width: 100\%; max-height: 100\%; }%" manual.html
 
 ../api:
-	cd .. && ln -s ../checker/api api
+	cd .. && ln -fs ../checker/api api
 
 CFLogo.png: ../logo/Logo/CFLogo.png
 	cp -p $< $@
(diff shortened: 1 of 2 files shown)


docs/manual/Makefile
```diff
@@ -71,7 +71,7 @@ manual.html: manual.pdf CFLogo.png favicon-checkerframework.png ../api
 	sed -i -e "s%<style type=\"text/css\">%<style type=\"text/css\">\nimg { max-width: 100\%; max-height: 100\%; }%" manual.html
 
 ../api:
-	cd .. && ln -s ../checker/api api
+	cd .. && ln -fs ../checker/api api
```
mernst: Why is `-f` needed?  This target runs only if the `../api` directory doesn't exist, so there is no need to overwiret existing destination files.

docs/manual/creating-a-checker.tex
```diff
@@ -174,6 +186,51 @@ \section{The parts of a checker\label{creating-parts-of-a-checker}}
 \end{description}
 
 
+
+
+\section{Compiling and using a custom checker}
```
mernst: Please add a `\label` LaTeX command before the `}`, so that the `check-labels` Ant task will pass and the Travis build will pass.  We cannot merge until the tests pass.

docs/manual/Makefile
Calvin-L: I put a note in my commit message about this. The rule will _also_ run if the symlink `../api` exists but points to a folder that does not exist. That can happen if you have never built `../../checker/api`. The "right" fix here might be for the Makefile to build `../../checker/api` and its contents, but I wasn't confident I could get that working.

mernst: Thanks for the explanation.  The commit message won't be visible to people reading the source, but anyone who chooses to change this in the future will be able to see it.

PR #1249, opened 2017-04-23
Update maven compiler plugin to 3.6.1

avandeursen:
As pointed out by @mernst existing maven setup truncated compiler messages.

This update helps avoiding this, by upgrading from the erroneous 3.3 plugin version to the newer 3.6.1. See also

https://issues.apache.org/jira/browse/MCOMPILER-229


mernst:
Thanks!  I appreciate the fix.

PR #1251, opened 2017-04-24
Value Checker: override strengthenAnnotationOfEqualTo

smillst:
Override strengthenAnnotationOfEqualTo rather than visitEqual and visitNotEqual.

This way switch statements are refined as the same as ==.  Also, fixes a crash when chars are in a switch statement. 

(Blocking #1243 )

### DIFF
--- a/framework/src/org/checkerframework/common/value/ValueTransfer.java
+++ b/framework/src/org/checkerframework/common/value/ValueTransfer.java
@@ -195,21 +195,23 @@ public class ValueTransfer extends CFTransfer {
     }
 
     private Range getIntRangeFromAnnotation(Node node, AnnotationMirror val) {
-        Range range = Range.EVERYTHING;
+        Range range;
         if (val == null || AnnotationUtils.areSameByClass(val, UnknownVal.class)) {
-            return Range.EVERYTHING;
+            range = Range.EVERYTHING;
         } else if (AnnotationUtils.areSameByClass(val, IntRange.class)) {
-            return ValueAnnotatedTypeFactory.getRange(val);
+            range = ValueAnnotatedTypeFactory.getRange(val);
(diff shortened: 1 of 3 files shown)


framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -805,36 +804,49 @@ private Range calculateRangeUnaryOp(
             TransferInput<CFValue, CFStore> p,
             CFStore thenStore,
             CFStore elseStore) {
-        if (isIntRange(leftNode, p)
-                || isIntRange(rightNode, p)
-                || isIntegralUnknownVal(rightNode, p)
-                || isIntegralUnknownVal(leftNode, p)) {
+        return calculateBinaryComparison(
```
kelloggm: What's the reasoning for keeping this version of `calculateBinaryComparison`? Why can't its callers just use the version below and call `getValueOfSubnode` themselves?

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -151,67 +149,69 @@ public ValueTransfer(CFAbstractAnalysis<CFValue, CFStore, CFTransfer> analysis)
...
 
     /** Get possible integer range from annotation. */
     private Range getIntRange(Node subNode, TransferInput<CFValue, CFStore> p) {
-        CFValue value = p.getValueOfSubNode(subNode);
+        AnnotationMirror val = getValueAnnotation(subNode, p);
+        return getIntRangeFromAnnotation(subNode, val);
+    }
+
+    private Range getIntRangeFromAnnotation(Node node, AnnotationMirror val) {
         Range range = Range.EVERYTHING;
```
smillst: Opps, I forgot to change the returns to assignments.  Good catch!

PR #1252, opened 2017-04-25
Consistently use $SLUGOWNER to determine repository org,

wmdietl:
including for downstream tests and when triggering travis builds.


.travis-build.sh
```diff
@@ -71,13 +75,13 @@ if [[ "${GROUP}" == "downstream" || "${GROUP}" == "all" ]]; then
   ##  * daikon-typecheck: (takes 2 hours)
 
   # checker-framework-inference: 18 minutes
-  (cd .. && git clone --depth 1 https://github.com/typetools/checker-framework-inference.git)
+  (cd .. && git clone --depth 1 https://github.com/${SLUGOWNER}/checker-framework-inference.git)
```
smillst: `${SLUGOWNER}/checker-framework-inference`may not exist.  If it doesn't typetools should be used.  (Same comment for `plume-lib`.

.travis-build.sh
```diff
@@ -71,13 +75,13 @@ if [[ "${GROUP}" == "downstream" || "${GROUP}" == "all" ]]; then
...
   export AFU=`pwd`/../annotation-tools/annotation-file-utilities
   export PATH=$AFU/scripts:$PATH
   (cd ../checker-framework-inference && gradle dist && ant -f tests.xml run-tests)
 
   # plume-lib-typecheck: 30 minutes
-  (cd .. && git clone https://github.com/mernst/plume-lib.git)
+  (cd .. && git clone https://github.com/${SLUGOWNER}/plume-lib.git)
```
smillst: Instead of creating a new fork for plume-lib which will have to be kept up to date with `mernst/plume-lib`.  Why not use If `mernst/plume-lib` if `${SLUGOWNER}/plume-lib` doesn't exist?

.travis-build.sh
wmdietl: Thanks for the comment! Yes, adapted it to do the same fall-back as in .travis-build-without-tests.sh.

.travis-build.sh
wmdietl: The typetools version should be the canonical version. I don't see a reason to use the personal repo instead of the organization repo like we do for everything else.

smillst: Are you suggesting that mernst/plume-lib should be deleted and typetools/plume-lib should be used instead?

wmdietl: It doesn't have to be deleted. It can be the repo used for pull requests to typetools/plume-lib.
I also have clones of all typetools repos in wmdietl and use those for local work.

smillst: But every place that use mernst/plume-lib should be changed to typetools/plume-lib?  That's a much bigger change than what you have done here.  For example, the plume-lib documentation would need to be update: https://github.com/mernst/plume-lib/blob/master/README.  Daikon would  need to be update to use typetools/plume-lib. etc.

More importantly plume-lib isn't about types, so it doesn't make sense for it to be in the typetools organization.

wmdietl: But plume-lib is a dependency that keeps breaking. By having a clone in typetools we can isolate such changes and have clones for different organizations.
I don't think we need to change all occurrences of mernst/plume-lib, as long as the ones in typetools are changed.

smillst: So the problem isn't that plume-lib is in `mernst`, it's that changes to plume-lib can cause test failures that have nothing to do with changes to the Checker Framework.  So you are proposing that the tests use a version in typetools that only contains code that type-checks with typetools master.  Is that right?

PR #876, opened 2016-08-19
enable storing extends/implements in TypesIntoElements#store()

CharlesZ-Chen:
This PR resolved the set different sizes issues posted in pascaliUWat/CFI:
- https://github.com/pascaliUWat/checker-framework-inference/issues/15
- https://github.com/pascaliUWat/checker-framework-inference/issues/14

Also fixed bugs caused by enabling storing extends/implements for a class element :

1. add if-check in storeClassExtends(): do nothing when the passed class symbol is an interface or java.lang.Object. This fixed a Nullpointer Exception. 

2. changed `typeOffset` in `SuperTypeAppiler#annotateSupers()` to always be -1. detail reasons are commented in code

The original review of this code is here: 

https://github.com/opprop/checker-framework/pull/5

PR #1253, opened 2017-04-25
Fix #989 by calculating type indices correctly.

### LINKED ISSUE #989: Error:java: ElementAnnotationUtil.getLocationTypeADT: invalid location TYPE_ARGUMENT(0) for type: Object
Simple interfaces compilation produce exception with CF 2.1.6. Works good with CF 2.1.0

ListWrapper.java:

```

import java.util.List;

public interface ListWrapper<E> extends List<@NonNull E> {
    default @Nullable E distinct() {
        if(size() > 1) throw new RuntimeException();
        return isEmpty() ? null : get(0);
    }
}

```

Test.java
```

public interface Test {

    ListWrapper<Boolean> getWrapper();

    default @Nullable Boolean getBoolean(){
        return getWrapper().distinct();
    }
}
```

wmdietl:


wmdietl:
There is another test for #880 in `checker/nullness/annotationsOnExtends/Other.java` It's still passing locally.

checker/tests/nullness/Issue880.java
```diff
@@ -0,0 +1,22 @@
+// Test case for Issue 880:
```
smillst: #880 is a pull request not an issue. The changes reverted #880 (namely PR #876) were trying to enable storing annotations in extends/implements clauses and that is what this test is testing.  But this PR doesn't enable storing annotations in extends/implements clauses.  So I'm confused as to what this is suppose to be testing.  

Also is this test just a duplicate of the tests in https://github.com/typetools/checker-framework/tree/master/checker/jtreg/nullness/annotationsOnExtends?

wmdietlGC: The test case was mentioned in #880, without a reference to the jtreg test. So I added it as an explicit test.
Do you want me to change the "Issue" to "Pull request"?

smillst: >The test case was mentioned in #880, without a reference to the jtreg test. So I added it as an explicit test.

That makes sense.  Is it just a duplicate of the jtreg test? If so, it should just be deleted.

wmdietlGC: The jtreg test uses separate files, whereas this tests within one file. So it covers an additional case.

smillst: So, it's an additional test for storing annotations in extends/implements clause.  We don't have an issue for that.  So maybe we should open one and then add reference that issue along with PR #880 and #876?

wmdietlGC: Issue #989 is caused by an annotation in an extends/implements clause, so the test from #880 seems like a good additional test to have.
I agree that we should add an issue that tries getting #876 fixed-up correctly.

smillst: Sounds good.  So then this test should either be renamed Issue989 or to the number of the new issue for storing annotations in extends/implements.

wmdietl: The test doesn't ensure that the annotation is stored in bytecode. It just tests whether the compiler crashes. I've renamed the file, to make all this clearer.

smillst:
@wmdietl Is this ok to merge?

wmdietlGC:
I was just waiting for Travis to pass after my last push and then forgot... looks good to me.

PR #1254, opened 2017-04-25
getStoreBefore might return null.

smillst:


wmdietl:
Can you add a test that triggers this problem?
Can you improve the getStoreBefore javadoc and state when the result can be null?

smillst:
https://github.com/typetools/checker-framework/pull/1243 has a test for this change.  But the test requires the changes in that pull request to trigger it.

>Can you improve the getStoreBefore javadoc and state when the result can be null?
I don't know all the cases it can be null, so I don't think I can improve the Javadoc.

PR #1255, opened 2017-04-27
Remove unnecessary brackets.

smillst:


wmdietl:
There are a few bottom qualifiers that have the null literal, but not the Void type as default, e.g. SignednessBottom and FBCBottom.
Should these be changed to also use Void? If not, is it worth calling out this irregularity with the qualifier?

PR #1259, opened 2017-05-04
Create MinLen annotations from Positive annotations

kelloggm:
This solves kelloggm#66 and similar bugs by inferring `@MinLen` annotations from arrays created using an `@Positive` length.

This creates a new type in the Value Checker called `@IntRangeFromPositive`. It is only introduced by when an `@Positive` annotation is read (it is aliased), and it is treated by dataflow/lub/glb/isSubtype/etc. as if it were an `@IntRange(from = 1)` annotation. `IntRange`s derived from it are normal `@IntRange` annotations. However, I also changed the common assignment checks so that if the left side has type `@IntRangeFromPositive`, the standard common assignment checks do not occur (in ValueVisitor). Writing `@IntRangeFromPositive` in code directly is an error. `@IntRangeFromPositive` is also not written into bytecode.

I also added a true bottom type to the LBC, because the Value Checker was trying to add `@IntRangeFromPositive` annotations to null.

This pull request has already been (partially) reviewed by @smillst as kelloggm#155, but was waiting on typetools#1243, which it depends on.


smillst:
@mernst Could you please review the documentation?

checker/src/org/checkerframework/checker/index/IndexChecker.java
```diff
@@ -13,7 +13,9 @@
  * <p>Constant Value Checker, SameLen Checker, SearchIndex Checker, Lower Bound Checker, Upper Bound
  * Checker
  *
- * <p>The Constant Value Checker has no dependencies.
+ * <p>The Constant Value Checker has no dependencies, but it does trust Positive annotations from
```
mernst: This comment seems more appropriate for file ValueChecker.java than here.

docs/manual/index-checker.tex
```diff
@@ -156,12 +156,16 @@ \section{Lower bounds\label{index-lowerbound}}
...
 The Index Checker issues an error when
 a sequence is indexed by an integer that might be negative.
-It uses a type system (Figure~\ref{fig-index-int-types}) with the following
+The Lower Bound Checker uses a type system (Figure~\ref{fig-index-int-types}) with the following
 qualifiers:
 
 \begin{description}
 \item[\refqualclass{checker/index/qual}{Positive}]
   The value is 1 or greater, so it is not too low to be used as an index.
+  Note that this annotation is trusted by the Constant Value Checker,
```
mernst: This detail seems distracting at this point in the manual; mentions elsewhere seem adequate.
How about making the Index Checker issue a warning if it ever encounters a programmer-written `@Positive` annotation and the Lower Bound Checker is not being run?  That seems like a less error-prone solution, and another benefit is that it enables all the relevant warnings to be removed from the manual.

docs/manual/index-checker.tex
```diff
@@ -171,6 +175,9 @@ \section{Lower bounds\label{index-lowerbound}}
 \item[\refqualclass{checker/index/qual}{LowerBoundUnknown}]
   There is no information about the value.
   It may not be used as an index for a sequence, because it might be too low.
+\item[\refqualclass{checker/index/qual}{LowerBoundBottom}]
```
mernst: I would be inclined to put qualifiers that the user does not write at the end of the list (that is, after `@PolyLowerBound`).

checker/src/org/checkerframework/checker/index/IndexChecker.java
kelloggm: Mentioning it in ValueChecker.java seems reasonable, but I think it's important to mention it here, too, since this is documenting the way the Index Checker's subcheckers interact, and the Lower Bound - Value Checker interaction is very important.

docs/manual/index-checker.tex
kelloggm: Right now it's not possible to run the Index Checker without running the Lower Bound Checker. We would need to have the Value Checker issue such a warning, which I think isn't feasible (do checkers have access to the list of checkers that will run after them?).

mernst: > do checkers have access to the list of checkers that will run after them?

That's my key question.  Is this information available, or could it be made available?

smillst: This information isn't currently available, but it could be made available.  

What if instead of issuing a warning, what if the Value Checker only converts `@Postive` if the Lower Bound Checker will be run?

mernst: I'm a bit concerned that such behavior would confuse users:  "I see `@Positive` right here, and in other code it is taken account of, but not in this code."  That's why I think a warning (which users could suppress) might be better.

docs/manual/index-checker.tex
kelloggm: I made this change for all of the Index Checker type systems.

checker/src/org/checkerframework/checker/index/IndexChecker.java
kelloggm: Looking at ValueChecker.java, it doesn't mention anything like this in its Javadoc. My bias would then be to avoid adding this information there, since no other information is there.

mernst: That is the logical place for the documentation.  I think it's better to add documentation rather than to omit it because some other documentation is missing.

kelloggm: I'll copy it there, but I think we should retain it in IndexChecker.java

framework/src/org/checkerframework/common/value/ValueChecker.java
```diff
@@ -10,8 +10,8 @@
  * The Constant Value Checker is a constant propagation analysis: for each variable, it determines
  * whether that variable's value can be known at compile time.
  *
- * <p>The Constant Value Checker has no dependencies, but it does trust {@link
- * org.checkerframework.checker.index.qual.Positive} annotations from the {@link
+ * <p>The Constant Value Checker has no dependencies, but it does trust {@code
```
smillst: You should be able to link to things that are on the classpath.  (So long as it is fully qualified.)

PR #1260, opened 2017-05-04
Add a polymorphic qualifier to the Value Checker

kelloggm:
Adds a `@PolyValue` annotation to the Value Checker and associated tests to fix typetools#1236.

Also changes commented out `@PolyMinLen` annotations in the Index JDK to `@PolyValue` and uncomments them.


framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -133,6 +138,8 @@ public ValueAnnotatedTypeFactory(BaseTypeChecker checker) {
         // this line just registers the alias. The BottomVal is never used.
         addAliasedAnnotation(MinLen.class, BOTTOMVAL);
 
+        addAliasedAnnotation(PolyAll.class, POLY);
```
smillst: PolyAll should be treated as PolyValue by the framework.  So this shouldn't be needed.

checker/jdk/index/src/java/lang/String.java
```diff
@@ -163,7 +163,7 @@ public String() {
      * @param  original
      *         A {@code String}
      */
-    public /*!PolyMinLen*/ @PolySameLen String(/*!PolyMinLen*/ @PolySameLen String original) {
+    public @PolyValue @PolySameLen String(@PolyValue @PolySameLen String original) {
```
smillst: Technically these are Value Checker annotations, so they don't belong in the index annotated jdk. They should go in org/checkerframework/common/value/statically-executable.astub.  (Could you also move the MinLen annotations there, too?)

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
kelloggm: This is a bug, then. I removed this and the tests that rely on them being equivalent started failing (`Polymorphic2.java` is the failing test).

smillst: Ok, I'll take a look.

kelloggm:
@smillst I think this is ready for you to look at again.

smillst:
I moved some of the changes to another pull request and then merged master into this branch.  Once the test pass, this pull request is ready to merge.

PR #1263, opened 2017-05-05
Remove extraneous MinLen and PolySameLen annotations on Strings in Index JDK

kelloggm:
This is causing the Value Checker to report warnings that it shouldn't when run on Daikon (because #1243 moved the MinLen annotation into the Value Checker). The extraneous PolySameLen annos aren't causing any problems, but they should be removed anyway in case they do cause problems down the line.

### DIFF
--- a/checker/jdk/index/src/java/io/File.java
+++ b/checker/jdk/index/src/java/io/File.java
@@ -195,7 +195,7 @@ public class File
      * string for convenience.  This string contains a single character, namely
      * <code>{@link #separatorChar}</code>.
      */
-    public static final /*!MinLen(1)*/ String separator = "" + separatorChar;
+    public static final String separator = "" + separatorChar;
 
     /**
      * The system-dependent path-separator character.  This field is
--- a/checker/jdk/index/src/java/io/File.java
+++ b/checker/jdk/index/src/java/io/File.java
@@ -195,7 +195,7 @@ public class File
      * string for convenience.  This string contains a single character, namely
      * <code>{@link #separatorChar}</code>.
      */
-    public static final @MinLen(1) String separator = "" + separatorChar;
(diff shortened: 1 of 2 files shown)


checker/jdk/index/src/java/io/File.java
```diff
@@ -195,7 +195,7 @@ int getPrefixLength() {
      * string for convenience.  This string contains a single character, namely
      * <code>{@link #separatorChar}</code>.
      */
-    public static final @MinLen(1) String separator = "" + separatorChar;
+    public static final /*!MinLen(1)*/ String separator = "" + separatorChar;
```
smillst: Should this just be deleted?

kelloggm: I left it in a comment because it would be correct if the Index Checker could handle Strings. But since MinLen is now a Value Checker annotation, maybe it should just be deleted.

smillst: The problem is that separator has a value type of `@StringVal("/")`, so it can't also have `@MinLen(1)`.

kelloggm: That's going to be a problem when we want to add String support. Since the annotation is commented out, it's not a problem now - but if you'd prefer, I can remove it entirely.

smillst: I don't think any Value Checker annotations should be in the Index jdk.  (Including ones that are commented out.)

kelloggm: We've talked about this before and I agree with you, but that change should be in a separate PR. It's on my TODO list.

kelloggm: I'll remove this one for now though.

PR #1266, opened 2017-05-08
Use areSame when sorting AnnotationMirrors.

smillst:
`compare` should return 0 when comparing `@IntRange(from=0, to=MAX)` and `@IntRange(to=MAX,from=0)`.  

The expected `CompletionFailure` exception issued by `checker/tests/nullness-extra/issue309` now occurs in a different method.  So I added a new catch for it.

(This was causing the following error: https://travis-ci.org/typetests/daikon-typecheck/jobs/229178124#L1038)

### DIFF
--- a/framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
+++ b/framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
@@ -2960,6 +2960,7 @@ public class AnnotatedTypeFactory implements AnnotationProvider {
             try {
                 results.add(annotation);
             } catch (com.sun.tools.javac.code.Symbol.CompletionFailure cf) {
+                // If a CompletionFailure occurs, issue a warning.
                 checker.message(
                         Kind.WARNING,
                         annotation.getAnnotationType().asElement(),
--- a/framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
+++ b/framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
@@ -2955,7 +2955,20 @@ public class AnnotatedTypeFactory implements AnnotationProvider {
 
         Set<AnnotationMirror> results = AnnotationUtils.createAnnotationSet();
         // Retrieving the annotations from the element.
-        results.addAll(elements.getAllAnnotationMirrors(elt));
+        List<? extends AnnotationMirror> fromEle = elements.getAllAnnotationMirrors(elt);
+        for (AnnotationMirror annotation : fromEle) {
+            try {
+                results.add(annotation);
+            } catch (com.sun.tools.javac.code.Symbol.CompletionFailure cf) {
+                checker.message(
+                        Kind.WARNING,
+                        annotation.getAnnotationType().asElement(),
+                        "annotation.not.completed",
+                        ElementUtils.getVerboseName(elt),
+                        annotation);
+            }
+        }
+
         // If declAnnosFromStubFiles == null, return the annotations in the element.
         if (declAnnosFromStubFiles != null) {
             // Adding @FromByteCode annotation to declAnnosFromStubFiles entry with key
--- a/javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
+++ b/javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
@@ -374,9 +374,14 @@ public class AnnotationUtils {
                     if (AnnotationUtils.areSame(a1, a2)) {
                         return 0;
                     }
+
                     String n1 = a1.toString();
                     String n2 = a2.toString();
 
+                    // Because the AnnotationMirror.toString prints the annotation as it appears
+                    // in source code, the order in which annotations of the same class are
+                    // sorted may be confusing.  For example, it might order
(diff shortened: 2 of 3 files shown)


framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -2955,7 +2955,21 @@ private AnnotationMirror getDeclAnnotation(
 
         Set<AnnotationMirror> results = AnnotationUtils.createAnnotationSet();
         // Retrieving the annotations from the element.
-        results.addAll(elements.getAllAnnotationMirrors(elt));
+        List<? extends AnnotationMirror> fromEle = elements.getAllAnnotationMirrors(elt);
+        for (AnnotationMirror annotation : fromEle) {
+            try {
```
wmdietl: Do you know why the CompletionFailure can happen here?
Should we put this try/catch into a helper method and use it everywhere?

javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -369,6 +369,11 @@ public static AnnotationMirror getSameIgnoringValues(
             new Comparator<AnnotationMirror>() {
                 @Override
                 public int compare(AnnotationMirror a1, AnnotationMirror a2) {
+                    // AnnotationMirror.toString() prints the elements of an annotation in the
```
wmdietl: Do we have a way to also sort the elements for comparison? It could be confusing if the order of elements depends on the order of the annotation attributes.

smillst: We don't currently have to do this comparison.
The order of elements only matter if the annotation is the same, but the elements are not (Otherwise, areSame would return true.)  It's probably a bug for the same annotation with different elements to be in an annotation set, so I don't think the order matters.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
smillst: It happens when calling areSame on an annotation mirror whose class isn't on the classpath.  So I guess I can move the try catch there.

wmdietlGC: Yes, that would catch more cases of this problem.

javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
wmdietlGC: I meant if you have `@IntRange(from=1, to=MAX)` and `@IntRange(to=MAX,from=0)` you would expect `from=0` to come before `from=1`. But because string comparison is used, you get the opposite order.
I don't think we have to fix this here, but a comment might be good. In general, doing things with strings is a hack.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
smillst: I can't move it there because there isn't a reference to a checker or an element to which to issue the error.

wmdietlGC: We could extend javacutil/src/org/checkerframework/javacutil/ErrorReporter.java to also support warnings.

smillst: The ErrorReporter throws a CheckerError that is caught by the SourceChecker. Even if we made it a warning, the Checker wouldn't continue to process the current compilation unit.

wmdietlGC: The warning method wouldn't throw an exception, it would just output the warning and continue normally.

smillst: How would it issue a warning (or any diagnostic message) without a reference to the checker?

wmdietlGC: The ErrorReporter has a reference to the ErrorHandler and the ErrorHandler is a checker.
So we could introduce a handler.issueWarning and have that output as a warning.

smillst: That seem like a good idea but out-of-scope for this pull request.  This pull request is fixing a crash on Daikon, so I would like to get it merged.

wmdietlGC: Sure. Can you at least file an issue? This seems like an important cleanup that would prevent future CompletionErrors from creeping up in other places.

javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
wmdietlGC: A todo comment to implement this correctly is fine.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
smillst: #1283  tracks this issue.

PR #1284, opened 2017-05-17
Fix a CompletionFailure and a follow-up NPE

wmdietlGC:


javacutil/src/org/checkerframework/javacutil/ElementUtils.java
```diff
@@ -360,14 +360,25 @@ public static boolean hasReceiver(Element element) {
             // For each direct supertype of the current type element, if it
             // hasn't already been visited, push it onto the stack and
             // add it to our superelems set.
-            TypeMirror supertypecls = current.getSuperclass();
-            if (supertypecls.getKind() != TypeKind.NONE) {
+            TypeMirror supertypecls;
+            try {
+                supertypecls = current.getSuperclass();
```
smillst: There are 18 other calls to getSuperClass in the Checker Framework.  Could all of them also issue a CompletionFailure or just this location?

framework/src/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -939,6 +939,13 @@ public void typeProcess(TypeElement e, TreePath p) {
         } else {
             previousErrorCompilationUnit = null;
         }
+        if (visitor == null) {
+            // typeProcessingStart invokes initChecker, which should
+            // have set the visitor. If the field is still null, an
+            // exception occured during initialization, which was already
+            // logged there. Don't also cause a NPE here.
+            return;
```
smillst: Is this change unrelated to the other one?

wmdietl: The CompletionFailure from the other change leaves the visitor null, which then causes a related NPE here.

javacutil/src/org/checkerframework/javacutil/ElementUtils.java
wmdietl: I don't know.

PR #1292, opened 2017-05-21
 Made it possible to subclass OverrideChecker in subclasses of BaseTypeVisitor.

mcoblenz:
This is needed for the Glacier typechecker.


framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -2218,6 +2218,30 @@ protected boolean testTypevarContainment(
     }
 
     /**
+     * Create an OverrideChecker.
+     *
+     * <p>This exists so that subclasses can subclass OverrideChecker and use their subclass instead
+     * of using OverrideChecker itself.
+     */
+    protected OverrideChecker makeOverrideChecker(
```
wmdietl: I would rename this method to `createOverrideChecker` to be in line with the other `createXXX` methods we have.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -2218,6 +2218,30 @@ protected boolean testTypevarContainment(
...
+                overrider,
+                overridingType,
+                overridingReturnType,
+                overridden,
+                overriddenType,
+                overriddenReturnType);
+    }
+
+    /**
      * Type checks that a method may override another method. Uses the OverrideChecker class.
```
wmdietl: Please adapt this javadoc to point to createOverrideChecker instead of the standard OverrideChecker class.

framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java
```diff
@@ -1440,6 +1440,12 @@ public AnnotatedArrayType getErased() {
             at.setComponentType(ct);
             return at;
         }
+
+        @Override
+        public void clearAnnotations() {
```
wmdietl: The javadoc for the super method only specifies that the main modifier should be cleared.
This change now also clears array component types. Should this also be done for other types, e.g. `AnnotatedDeclaredType`?
Why is this change necessary? If we agree this is a good change, the javadoc needs to be updated.

mcoblenz: This is somehow an artifact of how I created the pull request. Let me retry.

mcoblenz:
I will recreate this pull request.

PR #1293, opened 2017-05-21
Made it possible to subclass OverrideChecker in subclasses of BaseTyp…

mcoblenz:
…eVisitor.

### DIFF
--- a/framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
+++ b/framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
@@ -2218,7 +2218,32 @@ public class BaseTypeVisitor<Factory extends GenericAnnotatedTypeFactory<?, ?, ?
     }
 
     /**
-     * Type checks that a method may override another method. Uses the OverrideChecker class.
+     * Create an OverrideChecker.
+     *
+     * <p>This exists so that subclasses can subclass OverrideChecker and use their subclass instead
+     * of using OverrideChecker itself.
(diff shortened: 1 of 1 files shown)


mcoblenz:
The failure doesn't obviously have anything to do with my patch: 

* What went wrong:
Error resolving plugin [id: 'com.github.sherter.google-java-format', version: '0.6']
> Could not GET 'https://plugins.gradle.org/api/gradle/2.10/plugin/use/com.github.sherter.google-java-format/0.6'.
   > sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target

Is this an issue with the test framework, or is it somehow my bug?

mernst:
It's probably a transient failure.  I have re-started the job.

mcoblenz:
I've identified a couple of additional changes that I think are needed here. I'll post them once I have my dependent code working properly. For now, it might make sense to hold off integrating this.

mcoblenz:
I believe this is ready for review and merging (if you approve) now. I suggest squashing. Thanks!

wmdietl:
Thanks for the improvement!

mcoblenz:
Thank you! Any prediction as to when you’ll release the next version? I have a Maven project ready to deploy but I need a version number to depend on.

Michael

wmdietl:
We usually release on the first workday on or after the 1st of each month. So the next release should come out June 1st.

PR #1298, opened 2017-05-23
Add Index Checker error message for compile time constants

kelloggm:
This addresses kelloggm#157. The goal is to issue an error message that is meaningful. To do that, this PR adds a new error message that gives the array's Value Checker type and compares it to the constant value of the erroneous index.

If any of the conditions is true, the new error message is not displayed (i.e. the existing behavior is not changed):

1. The maximum constant value for the index is not known
2. The minimum length of the array is not known
3. The constant has an upperbound qualifier that is not `@UpperBoundUnknown`

The last condition is added to avoid strange error messages that come from the lengths of arrays, which have known constant ranges (i.e. they cannot be negative) but which also have upper bound types which are usually more meaningful.


checker/src/org/checkerframework/checker/index/upperbound/messages.properties
```diff
@@ -1,2 +1,3 @@
 ### Error messages for the Upper Bound Checker
 array.access.unsafe.high=Potentially unsafe array access: the index could be larger than the array's bound\nfound   : %s\nrequired: an integer less than %s's length (@LTLengthOf("%s"))
+array.access.unsafe.high.constant=Potentially unsafe array access: %s is larger than %s's known minimum length\nfound   : %s\nrequired: an array with a minimum length of at least %s
```
smillst: The following is an example of this warning:
```
Constants.java:9: error: [array.access.unsafe.high.constant] Potentially unsafe array access: 3 is larger than arr's known minimum length
        int j = arr[3];
                    ^
  found   : @UnknownVal int @ArrayLen(3) []
  required: an array with a minimum length of at least 4
```

The other `array.access.unsafe` warnings are written in terms of what's wrong with the index, rather than the array.  It seems to be a more common error to get the index wrong, rather than the length on the array incorrect.  

What about something like the following?
```
Constants.java:9: error: [array.access.unsafe.high.constant] Potentially unsafe array access: 3 is larger than arr's known minimum length (@ArrayLen(3))
        int j = arr[3];
                    ^
  value of index :  3
  array length    : @ArrayLen(3)
1 error
```
@mernst might have a better suggestion.

mernst:
As a general comment, a positive statement is usually clearer than a double negative.  I had trouble understanding the comment describing this pull request, which is

"the new error message is not displayed if ... 1. ...  is not known ... 2.  ... is not known" etc.

I'm hoping that the code logic is clearer.

kelloggm:
In the code, it is the "double negative" that's checked: we issue the old error message if any of three variables is null, and the new error message if all are non-null. But the comment in the code expresses the same thing in a positive way; I'm not sure why I chose to express it negatively in the PR.

checker/src/org/checkerframework/checker/index/upperbound/messages.properties
kelloggm: I updated the error message like you suggested.

kelloggm:
@mernst I've updated these error messages based on our discussion earlier

mernst:
Thanks.
I made a pull request against your branch, with a suggested change to the error messages.

smillst:
What's the status of this pull request?

kelloggm:
This should probably be merged.

Issue #1299, opened 2017-05-23
Value Checker issues warnings on implicit up-cast


kelloggm:
Consider the following code:

```java
void testShort_plus(@IntRange(from = 0) short x) {
    @IntRange(from = 1, to = Short.MAX_VALUE + 1) int y = x + 1;
}
```
It generates the following error:
```
ValueCast.java:9: error: [assignment.type.incompatible] incompatible types in assignment.
        @IntRange(from = 1, to = Short.MAX_VALUE + 1) int y = x + 1;
                                                                ^
  found   : @IntRange(from=-2147483648, to=2147483647) int
  required: @IntRange(from=1L, to=32768L) int
```

The Value Checker will issue an error for this code, because `@IntRange(from = 0)` is implicitly treated as `@IntRange(from = 0, to = Long.MAX_VALUE)`, even when the type being annotated is not a `long`. The Value Checker should fill in the appropriate default max value (or min value, when the specified value is `to`) for ranges when only one value is written by the programmer.

(Test case framework/tests/value/ValueCast.java)

PR #1300, opened 2017-05-23
Correct defaults for IntRange annotations on non-Long types

kelloggm:
This pull request solves #1299. More generally, this PR makes the defaults for `@IntRange` annotations that have no `to` or `from` field be determined by the underlying type of the variable that the annotation applies to.

### DIFF
--- a/framework/src/org/checkerframework/common/value/qual/IntRange.java
+++ b/framework/src/org/checkerframework/common/value/qual/IntRange.java
@@ -12,7 +12,9 @@ import org.checkerframework.framework.qual.SubtypeOf;
  * represents the four values 6, 7, 8, and 9.
  *
  * <p>If only one of the {@code to} and {@code from} fields is set, then the other will default to
- * the max/min value of the underlying type of the variable that is annotated.
+ * the max/min value of the underlying type of the variable that is annotated. Note that there must
+ * be "default" values (Long.MIN_VALUE and Long.MAX_VALUE below) even though they will be replaced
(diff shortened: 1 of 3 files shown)


framework/tests/value/ValueCast.java
```diff
@@ -0,0 +1,35 @@
+// Test case for issue 1299: https://github.com/typetools/checker-framework/issues/1299
+
+//@skip-test
```
smillst: No skip!

kelloggm: Ah, silly mistake!

kelloggm:
Added javadoc and fixed an errant skip-test

framework/src/org/checkerframework/common/value/qual/IntRange.java
```diff
@@ -11,14 +11,17 @@
...
+ * the max/min value of the underlying type of the variable that is annotated.
+ *
  * @checker_framework.manual #constant-value-checker Constant Value Checker
  */
 @SubtypeOf(UnknownVal.class)
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.TYPE_PARAMETER, ElementType.TYPE_USE})
 public @interface IntRange {
     /** Smallest value in the range, inclusive */
-    long from() default Long.MIN_VALUE;
```
smillst: Javac will complain if an annotation field doesn't have a default and a value isn't specified in an annotation.  In other words, Java is going to complain about `@IntRange(from=0)`.

PR #1301, opened 2017-05-24
Dealing with Value Checker integration problems

kelloggm:
@cmackie reported an issue integrating the signedness checker with the value checker yesterday via email - when typechecking code that interacts with unannotated libraries, the value checker issues spurious warnings. I've looked into it, and it seems that the value checker is assigning `@BottomVal` to some code that it shouldn't be. I didn't see why, and what I've got here is an awful hack, but it does resolve the problem. I'd like to know why, but to be honest I don't know why it works. 

@smillst, I'd appreciate your input here. The relevant test (that this "fixes") is `checker/tests/signedness/TestUncheckedByteCode.java`, which fails on lines 8, 13, and 14 because the constants being passed to the function are expected to be `@BottomVal` for some reason.

### DIFF
--- a/framework/src/org/checkerframework/common/value/ValueVisitor.java
+++ b/framework/src/org/checkerframework/common/value/ValueVisitor.java
@@ -1,5 +1,9 @@
 package org.checkerframework.common.value;
 
+/*>>>
+import org.checkerframework.checker.compilermsgs.qual.CompilerMessageKey;
+*/
+
 import com.sun.source.tree.AnnotationTree;
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.LiteralTree;
@@ -18,6 +22,7 @@ import org.checkerframework.common.value.qual.IntRange;
 import org.checkerframework.common.value.qual.IntVal;
 import org.checkerframework.common.value.qual.StringVal;
 import org.checkerframework.framework.source.Result;
+import org.checkerframework.framework.type.AnnotatedTypeMirror;
 import org.checkerframework.javacutil.AnnotationUtils;
 import org.checkerframework.javacutil.InternalUtils;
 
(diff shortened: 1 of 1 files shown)


framework/src/org/checkerframework/common/value/ValueVisitor.java
```diff
@@ -129,6 +134,20 @@ public Void visitAnnotation(AnnotationTree node, Void p) {
...
 
     @Override
+    protected void commonAssignmentCheck(
+            AnnotatedTypeMirror varType,
+            ExpressionTree valueExp,
+            /*@CompilerMessageKey*/ String errorKey) {
+
+        // This is definitely a hack. However, it does prevent some spurious warnings
+        // from the Value Checker when interacting with unannotated bytecode.
+        if (varType.toString().contains("@")
```
smillst: varType.toString() won't contain an annotation if the annotation on the varType is invisible.  An annotation is invisible if it has the meta-annotation org.checkerframework.framework.qual.InvisibleQualifier.  `@BottomVal` has this annotation, so that's why this "fixes" that warning.    I would guess this will also cause test failures (or should).

smillst:
`checker/tests/signedness/TestUncheckedByteCode.java` doesn't exist.

kelloggm:
My mistake. The test in question is `checker/tests/signedness-unchecked-defaults/TestUncheckedByteCode.java`

framework/src/org/checkerframework/common/value/ValueVisitor.java
kelloggm: Just `varType.toString().contains("@")` does cause a bunch of tests to fail, but adding the second condition causes them not to. I don't actually think this code is correct; I just wanted some help understanding what's going on here.

smillst:
The errors are occurring because -AuseDefaultsForUncheckedCode=-source,bytecode is passed as an option when those test are type-checked.  (See https://checkerframework.org/manual/#defaults-classfile for an explanation of that option.)  So those errors are expected given the Value Checker defaults for unchecked code.

We probably should change the Value Checker defaults for unchecked code.  @cmackie you can suppress those warnings for now.

framework/src/org/checkerframework/common/value/ValueVisitor.java
smillst: Oh, the second condition is different because it is checking if the underlying (Java) type has an annotation.  That happens when the type has an explicit annotation.  I would guess that it prevents test failures by chance.  Does that make sense?

kelloggm:
Thanks for explaining. I'm going to close this PR and open an issue that reminds us to reexamine the defaults for unchecked code in the Value Checker.

kelloggm:
#1303 describes the problem

PR #1306, opened 2017-05-26
Change VisitHistory to use a set of pairs of ATM.

smillst:
The "two properties of wildcards" mentioned in its Javadoc aren't true any more.

(Diakon still type checks with this change. https://travis-ci.org/smillst/daikon-typecheck/builds/236454679)


framework/src/org/checkerframework/framework/type/visitor/VisitHistory.java
```diff
@@ -3,43 +3,21 @@
...
- * // and each call to getEffectiveSuperBound returns a copy of the original bound
- * }</pre>
- * </ol>
- *
- * When we encounter types with property 1, property 2 ensures we cannot stop recursively comparing
- * the bounds because the equals method will not return true when we encounter a copy of a bound we
- * have already explored. This class defines a "Visit" inner class which consists of a pair of
- * AnnotatedTypeMirrors. Its equalityCompare method compares AnnotatedTypeMirrors in a way that
- * identifies wildcards that have already been compared.
+ * not they have visited an equivalent pair of AnnotatedTypeMirrors already. This is necessary in
```
wmdietl: Can you explain why this needs to be a pair of types? To prevent recursion, wouldn't it be enough to keep track of visited types? E.g. in `org.checkerframework.framework.type.visitor.AnnotatedTypeScanner.visitedNodes` we simply keep track of the types and results.
Then instead of passing around this object, we could just pass the `Set<AnnotatedTypeMirror>` around.

smillst: This class is primarily used to implement `isSubtype(ATM, ATM)`.  So the pair of types corresponds to the subtype and the supertype being checked.  A single subtype may be visited more than once, but with a different supertype.  

For example, if the two types are `@A T extends @B Number` and `@C Number`, then isSubtype is first called one those types and then on `@B Number` and `@C Number`.

wmdietl: It would be nice if you could tie in the explanation with the "necessary... on recursive bounds" aspect.
Maybe if you use `@A T extends @B Serializable<T>` as example?
Either way, some additional context would be helpful.

Issue #1264, opened 2017-05-05
Value Checker issues warnings when casting between integral types with range annotations


kelloggm:
Consider the following code:

```java
byte foo(@IntRange(from = Integer.MIN_VALUE, to = Integer.MAX_VALUE) int x) {
    return (byte) x;
}
```

The Value Checker issues a warning that the cast from `@IntRange(from = Integer.MIN_VALUE, to = Integer.MAX_VALUE)` to `@IntRange(-128, 127)` is unsafe. This warning is a false positive.

The same problem occurs whenever there is a downcast between integral types, not just between `int` and `byte`

smillst:
When this issue is fixed, this daikon change should be reverted:
https://github.com/codespecs/daikon/commit/2ac5315e02aa40d042686011d3ed53bf9b8ba86a

smillst:
Implementation note: 

The ValueVisitor should probably override org.checkerframework.common.basetype.BaseTypeVisitor#checkTypecastSafety and checkTypecastRedundancy.

wmdietl:
Out of curiosity: why is the cast a false positive? The cast goes from a larger range to a smaller range, which would seem like exactly what the Value Checker should warn about.

smillst:
The cast is from an int to a byte.  A byte's range is at most -128 to 127. So casting from `@IntRange(from = Integer.MIN_VALUE, to = Integer.MAX_VALUE) int` to `@IntRange(-128, 127) byte` is always safe.

wmdietl:
I would agree that casting from `@IntRange(-128, 127) int` to `@IntRange(-128, 127) byte` is safe.
But casting from all possible integers to `@IntRange(-128, 127) byte` looses information. I must be missing something.

smillst:
What Value Checker type should '(byte) x' have if not '@IntRange(-128, 127)'?

kelloggm:
Or, put another way, even if the integer is out of the range of the byte type, the resulting byte will still have type `@IntRange(-128, 127)` because the value will overflow, and a byte cannot hold values outside that range.

wmdietl:
Do you get a warning when you assign/cast a `@IntRange(from = Integer.MIN_VALUE, to = Integer.MAX_VALUE) int` to a `@IntRange(-128, 127) int`?

smillst:
[...]

PR #1309, opened 2017-05-31
Suppress false positive warnings in the Value Checker on some (safe) type casts.

kelloggm:
Solves the issue described in #1264. The Value Checker currently issues `cast.unsafe` warnings even when the programmer has written an explicit cast and the annotation on the expression being cast covers the whole range of the target type (i.e. a cast from an integer covering the whole `int` range to a `byte`.


framework/src/org/checkerframework/common/value/ValueVisitor.java
```diff
@@ -171,6 +173,40 @@ public Void visitTypeCast(TypeCastTree node, Void p) {
...
+        // of the new type. I.e. do not warn if an @IntRange(from = Integer.MIN_VALUE,
+        // to = Integer.MAX_VALUE) int is casted to a @IntRange(from = Byte.MIN_VALUE,
+        // to = Byte.MAX_VALUE byte).
+        if (castAnno != null
+                && exprAnno != null
+                && atypeFactory.isIntRange(castAnno)
+                && atypeFactory.isIntRange(exprAnno)) {
+            Range castRange = ValueAnnotatedTypeFactory.getRange(castAnno);
+            Range exprRange = ValueAnnotatedTypeFactory.getRange(exprAnno);
+            if (exprRange.contains(castRange)) {
```
smillst: I don't understand why this is needed.  If `castType` is a byte and is "byte everything", then isn't it a safe typecast?

kelloggm: I think you're correct. I was thinking of casting a value that didn't cover the whole `byte` range into a `byte` - for instance, casting an `@IntRange(-200, -100)` int into a `byte`. That won't cover the whole `byte` range, but it will be _in_ the `byte` range, so the cast should be okay. I'll fix this.

framework/src/org/checkerframework/common/value/ValueVisitor.java
```diff
@@ -171,6 +173,37 @@ public Void visitTypeCast(TypeCastTree node, Void p) {
...
         }
+
+        AnnotatedTypeMirror castType = atypeFactory.getAnnotatedType(node);
+        AnnotationMirror castAnno = castType.getAnnotationInHierarchy(atypeFactory.UNKNOWNVAL);
+        AnnotationMirror exprAnno =
+                atypeFactory
+                        .getAnnotatedType(node.getExpression())
+                        .getAnnotationInHierarchy(atypeFactory.UNKNOWNVAL);
+
+        // Don't check a type cast if it is to a declared IntRange that covers the whole
```
smillst: This sentence is a bit confusing.  For example, what does it mean to "cover the whole type"?   What about something like:
It is always legal to cast a type to a type that includes all possible values for that type.

kelloggm: I changed the wording so that it should be easier to understand.

PR #1311, opened 2017-06-01
Fix #867: Dataflow: postfix increment and decrements are treated like prefix postfix increment and decrements

### LINKED ISSUE #867: Dataflow: postfix increment and decrements are treated like prefix postfix increment and decrements

bitterfox:
In this fix, I've changed building cfg for postfix operation to:
- First, create a temp variable
- Second, escape the current value of the expression to the temp variable(Assign the current value to the temp variable)
- Third, add 1 and assign it to the variable
- Fourth, use the temp variable as a result of postfix increment or decrement

For example, considering `y = x++ + x;`, the nodes will be like following:
```
x : LocalVariable
temp = x : Declare variable and assign
x + 1 : NumericalAddition
x = x + 1 : Assign
x : LocalVariable
temp + x : NumericalAddition
y = temp + x : Assign
```

It's also including refactoring because most of codes for postfix inc&dec and prefix inc&dec are duplicated.


dataflow/src/org/checkerframework/dataflow/cfg/CFGBuilder.java
```diff
@@ -4301,6 +4253,61 @@ public Node visitUnary(UnaryTree tree, Void p) {
             return result;
         }
 
+        private AssignmentNode createIncrementOrDecrementAssign(
```
wmdietl: Could you add some brief javadoc to the two new methods?

framework/tests/value/Issue867.java
```diff
@@ -0,0 +1,28 @@
+import org.checkerframework.common.value.qual.*;
```
wmdietl: We usually start test casses with a standardized header:
```
// Test case for Issue 867:
// https://github.com/typetools/checker-framework/issues/867
```

bitterfox:
Hi Werner,
I've just fixed what are you mentioned.

Please review this again!

Regards

framework/tests/value/Issue867.java
```diff
@@ -0,0 +1,30 @@
+// Test case for Issue 867:
+// https://github.com/typetools/checker-framework/issues/867
+import org.checkerframework.common.value.qual.*;
+
+class Issue867 {
+    void test1() {
+        @IntVal({0, 1}) int x = 0;
+        @IntVal(0) int y = x++;
+        @IntVal(1) int w = x;
+    }
```
wmdietl: Can you also add a few tests with expected errors?
E.g. what happens if you ++ x again? You should get an error as you leave the 0-1 range.
Having a few expected errors you check for makes sure that your code is catching mistakes, too.

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1275,15 +1274,8 @@ protected void addComputedTypeAnnotations(
         defaults.annotate(tree, type);
 
         if (iUseFlow) {
-            Value as;
-            if (tree.getKind() == Kind.POSTFIX_DECREMENT
-                    || tree.getKind() == Kind.POSTFIX_INCREMENT) {
-                // Dataflow incorrectly treats postfix as prefix.
-                // See Issue 867: https://github.com/typetools/checker-framework/issues/867
-                as = getInferredValueFor(((UnaryTree) tree).getExpression());
```
smillst: [BaseTypeVisitor](https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java#L1343) references this work around.  Now that the work around is removed, the comment in BaseTypeVisitor can also be removed.

bitterfox:
Hi @wmdietl and @smillst ,

I've now removed the unnecessary comments in BaseTypeVisitor and added some expected error tests.
I've added some codes and modified some tests because tests are failed.

Now it's prepared to be reviewed again.
Please review this again!!
Thank you.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1359,13 +1359,25 @@ public Void visitUnary(UnaryTree node, Void p) {
...
                 || (node.getKind() == Tree.Kind.POSTFIX_INCREMENT)) {
             AnnotatedTypeMirror varType = atypeFactory.getAnnotatedTypeLhs(node.getExpression());
-            // For postfix increments/decrements, the value type is incorrect due to the workaround
-            // in GenericAnnotatedTypeFactory.addComputedTypeAnnotations(Tree, AnnotatedTypeMirror, boolean)
-            // for the following bug:
-            // See Issue 867: https://github.com/typetools/checker-framework/issues/867
-            // This means could result in a false warning (false positive) in some cases and a lack
-            // of a warning in other cases (false negative).
-            AnnotatedTypeMirror valueType = atypeFactory.getAnnotatedType(node);
+            Tree newValue = node;
```
wmdietlGC: Why introduce this variable? The name is also rather confusing, given that it is a `Tree`, but named `newValue`. There is a `value` local variable further down, which is not related at all.
Can't you just use `node`?

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1359,13 +1359,25 @@ public Void visitUnary(UnaryTree node, Void p) {
...
+            AnnotatedTypeMirror valueType = atypeFactory.getAnnotatedType(newValue);
+
+            // When the operation is postfix, valueType is a type of a value before the
+            // assignment. We override valueType with a type from a store, if it's provided.
+            if (node.getKind() == Tree.Kind.POSTFIX_DECREMENT
+                    || node.getKind() == Tree.Kind.POSTFIX_INCREMENT) {
+                CFAbstractStore<?, ?> store = atypeFactory.getStoreAfter(newValue);
+                if (store != null) {
+                    CFAbstractValue<?> value =
```
wmdietlGC: Why can't you use code more similar to how the `GenericATF` worked below, e.g. use `Value as = getInferredValueFor(node.getExpression());` ?

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1359,13 +1359,25 @@ public Void visitUnary(UnaryTree node, Void p) {
...
+                            store.getValue(
+                                    FlowExpressions.internalReprOf(
+                                            atypeFactory, node.getExpression()));
+                    if (value != null) {
+                        valueType.clearAnnotations();
```
wmdietlGC: `valueType.replaceAnnotations(value.getAnnotations())` is probably better and doesn't remove unrelated annotations.

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
wmdietlGC: @bitterfox I think you mis-interpreted @smillst comment. I think it was nicer to have this logic in the GenericATF instead of moving it to the BaseTypeVisitor. The visitors shouldn't modify the types.
The suggestion was to just update the comment in BaseTypeVisitor.

wmdietlGC:
@bitterfox The last comment about the change to BaseTypeVisitor is the most important one, as the other comments will no longer matter.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
bitterfox: I'll explain this logic and the logic which I removed in `GenericATF` little bit more using example:
```
int x = 0;
@IntVal(0) int y = x++;
```

Before this PR, the inferred value of `x++` is `1` but it should be `0`. 
To accept `@IntVal(0) int y = x++;`, the workaround in `GenericATF`is required.

But in this PR, it has been fixed, so it's no longer necessary.

However when we consider following code:
```
@IntVal(0) int zero;
zero++;
```

The compound assignment by `zero++` should be error because `1` is assigned to `zero`.
But the inferred value of `zero++` and `zero` are `0`, so `getInferredValueFor(node.getExpression())` or `getInferredValueFor(node)` doesn't works well in this situation.
The value of of `zero` in the store after `zero++` is `1`.
So I've add the logic to `visitUnary`, which takes a store after the node and takes a value from the store and updates the type.
I think this logic should be in `visitUnary` because this logic is necessary just only in `visitUnary` and it is close to what `visitUnary` want to do.

wmdietlGC: I'm wondering whether we should do something similar to boxing/unboxing of primitives.
If we do the special treatment in the Visitor, everybody who uses the CFG needs to be aware of the distinction and implement it.
For boxing/unboxing we keep a reference to the Tree before conversion. See Section 3.3.2 in the dataflow manual and https://github.com/typetools/checker-framework/blob/master/dataflow/src/org/checkerframework/dataflow/cfg/ControlFlowGraph.java#L53

Can you argue why unary increment/decrement shouldn't be handled similarly?
This reminds me of another thing: this PR should also update the dataflow manual and document what tree transformation we're performing.

mernst:
@bitterfox Shinya, this pull request is still assigned to you.  Have you made the requested changes?  If so, please reassign the pull request to a reviewer.  If not, could you do so, so that we can resolve this pull request before we forget about it?  Thanks!

bitterfox:
Hi, 
I've just updated my PR.
Please review this again.

It's my understanding that `convertedTreeLookupMap` is used during dataflow analysis and `treeLookupMap` is used in type checking.
The value of the node which is corresponding unary postfix operation in both of maps should be the value which is before compound assignment.
So we couldn't use both map for compound assignment.

In this change, to handle it similarly with boxing/unboxing, I've added a map which has `UnaryTree` and corresponding compound `AssignmentNode` in CFGBuilder, CFG and AnalysisResult.

dataflow/src/org/checkerframework/dataflow/analysis/AnalysisResult.java
```diff
@@ -31,6 +32,9 @@
     /** Map from AST {@link Tree}s to {@link Node}s. */
     protected final IdentityHashMap<Tree, Node> treeLookup;
 
+    /** Map from AST {@link UnaryTree}s to compound {@link AssignmentNode}s. */
+    protected final IdentityHashMap<UnaryTree, AssignmentNode> compoundAssignTreeLookupMap;
```
wmdietl: Why do we need a separate map for the UnaryTree? Can the same UnaryTree be both in `treeLookup` and in `compoundAssignTreeLookupMap`? If so, describe what the two different values mean. If not, why have the separate map?

wmdietl: `compoundAssignTreeLookupMap` seems like a bad name for this map.
A compound assignment is something like `a += b`, but this map stores a UnaryTree to AssignmentNode mapping. Is the assignment node the artificially introduced one for the pre/post inc/dec? In any event, there is no compound assignment involved, is there?

dataflow/src/org/checkerframework/dataflow/cfg/CFGBuilder.java
```diff
@@ -4198,91 +4212,56 @@ public Node visitUnary(UnaryTree tree, Void p) {
...
-                        Node operNode;
-                        if (kind == Tree.Kind.PREFIX_INCREMENT) {
-                            operNode = new NumericalAdditionNode(operTree, exprRHS, one);
+                        boolean isIncrement =
+                                kind == Tree.Kind.POSTFIX_INCREMENT
+                                        || kind == Kind.PREFIX_INCREMENT;
+                        boolean isPostfix =
+                                kind == Tree.Kind.POSTFIX_INCREMENT
+                                        || kind == Kind.POSTFIX_DECREMENT;
+                        AssignmentNode compoundAssign =
```
wmdietl: Also here the name `compoundAssign` is confusing, as it is an assignment node and not a compound assignment (for which there wouldn't be a Node type.)

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1359,13 +1359,10 @@ public Void visitUnary(UnaryTree node, Void p) {
...
             AnnotatedTypeMirror varType = atypeFactory.getAnnotatedTypeLhs(node.getExpression());
-            // For postfix increments/decrements, the value type is incorrect due to the workaround
-            // in GenericAnnotatedTypeFactory.addComputedTypeAnnotations(Tree, AnnotatedTypeMirror, boolean)
-            // for the following bug:
-            // See Issue 867: https://github.com/typetools/checker-framework/issues/867
-            // This means could result in a false warning (false positive) in some cases and a lack
-            // of a warning in other cases (false negative).
-            AnnotatedTypeMirror valueType = atypeFactory.getAnnotatedType(node);
+            AnnotatedTypeMirror valueType = atypeFactory.getAnnotatedTypeRhsCompoundAssign(node);
+            if (valueType == null) {
```
wmdietl: Why not put this logic into `getAnnotatedTypeRhsCompoundAssign`?

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -847,6 +848,9 @@ public Store getStoreBefore(Node node) {
     /** @return the store immediately after a given {@link Tree}. */
     public Store getStoreAfter(Tree tree) {
         if (analyses.isEmpty()) {
+            if (flowResult == null) {
```
wmdietl: Why is this new check needed?

dataflow/src/org/checkerframework/dataflow/cfg/CFGBuilder.java
```diff
@@ -4301,6 +4280,75 @@ public Node visitUnary(UnaryTree tree, Void p) {
...
+                handleArtificialTree(target);
+            }
+
+            AssignmentNode assignNode = new AssignmentNode(target, expr, narrowed);
+            assignNode.setInSource(false);
+            return extendWithNode(assignNode);
+        }
+
+        /**
+         * Find nearest owner element(Method or Class) which holds current tree
```
wmdietl: Add a whitespace after `element`.

dataflow/src/org/checkerframework/dataflow/analysis/AnalysisResult.java
bitterfox: Because `treeLookup` and `convertedTreeLookup` are reserved for a node for unary operation 
self and a conversion node like unbox or widening or something like that.

Please consider following code:
```
@IntVal(0)
short n = 0;
@IntVal(0)
int m = n++;
```

At `visitUnary` for `n++`, artificial tree and node for assignment(`n = n + 1`) will be created.
And `temp = n`(where `n` is a node of before the assignment) will be created and stored in `treeLookup` because it's postfix.

At `visitAssignment` for `int m = n++`, `WideningConversionNode` for `temp = n` will be created and stored into `convertedTreeLookup`.

We need `n = n + 1` to type check in `BaseTypeVisitor`, I need to store it into `unaryAssignNodeLookupMap`

bitterfox:
Hi @wmdietl ,
I've updated dataflow manual and fixed other comments.
Please review this again.

I also commented for the question which is hidden due to outdated:
https://github.com/typetools/checker-framework/pull/1311#discussion_r131914170

dataflow/src/org/checkerframework/dataflow/analysis/AnalysisResult.java
```diff
@@ -100,6 +100,11 @@ public void combine(AnalysisResult<A, S> other) {
         return treeLookup.get(tree);
     }
 
+    /** @return the corresponding {@link AssignmentNode} for a given {@link UnaryTree}. */
```
wmdietl: When can this return null? Shouldn't all UnaryTrees be processed before this method is invoked? Therefore, wouldn't it be better to fail if a given tree is not in the map?

dataflow/src/org/checkerframework/dataflow/cfg/ControlFlowGraph.java
```diff
@@ -216,6 +223,11 @@ public UnderlyingAST getUnderlyingAST() {
         return new IdentityHashMap<>(treeLookup);
     }
 
+    /** @return lookup map of the assign node for unary operation */
```
wmdietl: Can you clarify here and in `getTreeLookup` above that a copy is returned.

dataflow/manual/content.tex
```diff
@@ -930,6 +930,17 @@ \subsubsection{Assignment}
 left-to-right (cf.\ \jlsref{15.26.1}),
 which is properly handled by the translation.}
 
+\subsubsection{Postfix/Prefix Increment/Decrement}
+Both of postfix and prefix increment or decrement have a side effect to
```
wmdietl: Can you add a `\label` for this section?

bitterfox:
Hi Werner,
Thank you for your review.
I've fixed. Please check this again.

wmdietl:
@smillst Could you look through the Index & Value Checker tests? They looked right to me.
Any comments on the other changes also appreciated. I think this can be merged.

PR #1317, opened 2017-06-05
Correctly type a cast to a type variable. (Fixes 1315)

### LINKED ISSUE #1315: Cast of @Nullable value to a type parameter
When we have a cast from a nullable value, the result has the same nullability.
In this example:

```
class NonGenDemo {
  void take(String s) {}
  void baz(@Nullable Object p) {
    take((String) p);
  }
}
```

We get:
```
Casts.java:22: error: [argument.type.incompatible] incompatible types in argument.
    take((String) p);
         ^
```
because String is automatically `@Nullable String`.
Good!

Now take this example with type parameters:

```

class Box<T> {
  T f;
  Box(T p) { f = p; }
  T unsound(@Nullable Object p) {
    return (T) p;
  }
}

public class Casts {
    Box<String> bs = new Box<String>("");
    bs.f = bs.unsound(null);
    bs.f.toString();
  }
}
```

This class compiles and only raises an `uses unchecked or unsafe operations` note from javac.
However, this results in a NullPointerException.

The expression `(T) p` in `unsound` should have type `@Nullable T` and should therefore be incompatible with the return type of the method.
I acknowledge that there is an unchecked cast, but i think we should treat this as an error.

smillst:
Fixes #1315.

### DIFF
--- a/framework/tests/all-systems/TypeVars.java
+++ b/framework/tests/all-systems/TypeVars.java
@@ -3,7 +3,7 @@ class TypeVars {
 
     class Test1<T> {
         void m() {
-            @SuppressWarnings("unchecked")
+            @SuppressWarnings({"unchecked", "assignment.type.incompatible"})
             T x = (T) new Object();
 
             Object o = x;
--- a/framework/tests/all-systems/TypeVars.java
+++ b/framework/tests/all-systems/TypeVars.java
@@ -3,7 +3,10 @@ class TypeVars {
 
     class Test1<T> {
         void m() {
-            @SuppressWarnings({"unchecked", "assignment.type.incompatible"})
+            @SuppressWarnings({
+                "unchecked",
+                "units:assignment.type.incompatible" // Units doesn't use dataflow, so "x" isn't defaulted to top.
+            })
             T x = (T) new Object();
 
             Object o = x;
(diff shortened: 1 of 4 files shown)


framework/tests/all-systems/TypeVars.java
```diff
@@ -3,7 +3,7 @@
 
     class Test1<T> {
         void m() {
-            @SuppressWarnings("unchecked")
+            @SuppressWarnings({"unchecked", "assignment.type.incompatible"})
```
wmdietl: What is the assignment error here? Is the cast type `@NonNull T`?
As x is a local variable, shouldn't the assignment always work?

smillst: This assignment works if the default for local variables is `@Top`.  The Unit Checker issued an assignment.type.incompatible error.  (The Nullness Checker does not issue a warning here and the cast type is `@NonNull T`.)

wmdietl: If it's just the units checker, we could suppress the warning only for it. A short comment might be helpful to avoid confusion why this is necessary.

smillst:
We decided that instead of change the type of `(T) expr`, that a cast unsafe warning should be issued.  So, I'm going to close this pull request and open another one with that change.

PR #1318, opened 2017-06-05
Correct unchecked defaults for the Value Checker.

### LINKED ISSUE #1303: Defaults for unchecked code in the Value Checker
The Value Checker issues warnings when run on `checker/tests/signedness-unchecked-defaults/TestUncheckedByteCode.java` because it uses conservative defaulting for unchecked code. This might not be the correct behavior for the Value Checker, since it's not intended to be run as a verification tool. Instead, it would be better to be conservative about issuing warnings and allow the later analysis to issue warnings.

smillst:
Fixes #1303

### DIFF
--- a/framework/tests/src/tests/ValueUncheckedDefaultsTest.java
+++ b/framework/tests/src/tests/ValueUncheckedDefaultsTest.java
@@ -1,0 +1,29 @@
+package tests;
+
+import java.io.File;
+import java.util.List;
+import org.checkerframework.common.value.ValueChecker;
+import org.checkerframework.framework.test.CheckerFrameworkPerDirectoryTest;
+import org.junit.runners.Parameterized.Parameters;
+
+/** Tests unchecked code defaults the constant value propagation type system. */
+public class ValueUncheckedDefaultsTest extends CheckerFrameworkPerDirectoryTest {
+
+    public ValueUncheckedDefaultsTest(List<File> testFiles) {
+        super(
+                testFiles,
+                ValueChecker.class,
+                "value",
+                "-Anomsgtext",
+                "-Astubs=statically-executable.astub",
+                "-AuseDefaultsForUncheckedCode=btyecode",
+                "-A" + ValueChecker.REPORT_EVAL_WARNS);
+    }
+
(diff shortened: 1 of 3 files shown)


framework/tests/src/tests/ValueUncheckedDefaultsTest.java
```diff
@@ -0,0 +1,29 @@
+package tests;
+
+import java.io.File;
+import java.util.List;
+import org.checkerframework.common.value.ValueChecker;
+import org.checkerframework.framework.test.CheckerFrameworkPerDirectoryTest;
+import org.junit.runners.Parameterized.Parameters;
+
+/** Tests unchecked code defaults the constant value propagation type system. */
```
wmdietl: Nit: add "for" to make "for the constant...". Or "in"?

framework/tests/src/tests/ValueUncheckedDefaultsTest.java
```diff
@@ -0,0 +1,29 @@
...
+public class ValueUncheckedDefaultsTest extends CheckerFrameworkPerDirectoryTest {
+
+    public ValueUncheckedDefaultsTest(List<File> testFiles) {
+        super(
+                testFiles,
+                ValueChecker.class,
+                "value",
+                "-Anomsgtext",
+                "-Astubs=statically-executable.astub",
+                "-AuseDefaultsForUncheckedCode=btyecode",
```
wmdietl: Hi @smillst , we've just been looking at this code and notice that `btyecode` is not spelled correctly.
Should there be an error message for incorrectly spelled arguments?

mernst: Yes, there should be.  Do you want to implement that, or are you asking Suzanne to do so?
Thanks for noticing this.

wmdietl: I assume that with the invalid flag this whole test case doesn't do what was intended.
So it would be best for @smillst to have a look through the test and in the process implement an error for invalid keys.

Issue #159, opened 2015-07-03
Modify Maven plugin artifact ID to conform to naming conventions


GoogleCodeExporter:
```
The artifact ID for the Maven plugin should be something like 
"checkers-maven-plugin" - by following the naming conventions, the prefix 
mapping for the plugin happens automatically (i.e. without additional 
configuration): 
http://maven.apache.org/guides/introduction/introduction-to-plugin-prefix-mappin
g.html#Specifying_a_Plugins_Prefix. This then permits command-line invocations 
like "mvn checkers:check" instead of the much longer "mvn 
types.checkers:checkersplugin:check".
```

Original issue reported on code.google.com by `atomknig...@gmail.com` on 24 Sep 2012 at 3:19

PR #1320, opened 2017-06-05
Adds a missing annotation in the Index JDK, and resolves a related issue with SameLen 

### LINKED ISSUE #146: Applying @PolymorphicQualifier to fields or classes results in unsoundness
```
1. Declare @Bottom, @Top, @Bottom <: @Top, and @PolymorphicQualifier @Poly
2. Write a class with a @Poly field.
3. Write an assignment from @Top to the polymorphic field
4. Run checker.

I would expect an error, either for the checker to issue an error about invalid 
assignment or for the checker framework to throw an exception because a 
polymorphic qualifier was applied to a field type.  If the field were 
instantiated to @Bottom, this would be unsound.

What version of the product are you using? On what operating system?
HEAD, Linux.

Please provide any additional information below.
The crux of the issue would seem to be that polymorphically-qualified fields 
don't make sense in a system that only support qualifiers polymorphic methods, 
and not qualifier-polymorphic types, but no warning is issues. 
```

Original issue reported on code.google.com by `colin.s....@gmail.com` on 14 May 2012 at 4:05

kelloggm:
When a variable declaration is reached, ensures that the name of the variable on the left hand side of the declaration (i.e. the variable being declared) is present in the SameLen annotation on the right hand side. This means that an assignment will never fail because a variable is not SameLen to itself when it is being declared.

Resolves #146 except for the issues detailed in #159.


checker/src/org/checkerframework/checker/index/samelen/SameLenVisitor.java
```diff
@@ -25,6 +25,23 @@ public SameLenVisitor(BaseTypeChecker checker) {
...
+
+    protected void commonAssignmentCheck(
+            Tree varTree, ExpressionTree valueExp, /*@CompilerMessageKey*/ String errorKey) {
+        if (varTree.getKind() == Tree.Kind.VARIABLE) {
+            VariableTree vTree = (VariableTree) varTree;
+            lastVar = vTree.getName().toString();
+        } else {
+            lastVar = null;
+        }
+        super.commonAssignmentCheck(varTree, valueExp, errorKey);
```
smillst: Should probably set lastVar to null again here.

checker/src/org/checkerframework/checker/index/samelen/SameLenVisitor.java
```diff
@@ -25,6 +25,23 @@ public SameLenVisitor(BaseTypeChecker checker) {
         super(checker);
     }
 
+    // This variable is used to store the name of the variable being assigned to on
+    // the left hand side of an assignment/pseudo-assignment, so that that variable
+    // can later be included in the SameLen type. This pattern is a bit uncomfortable,
+    // but there is not a clear other way to save this information.
```
smillst: Would it work to remove the variable from the SameLen rather than copying it to the RHS?  You could do this in SameLenAnnotatedTypeFactory#getAnnotatedTypeLhs.

checker/jdk/index/src/java/lang/Object.java
```diff
@@ -211,7 +211,7 @@ public boolean equals(Object obj) {
      *               be cloned.
      * @see java.lang.Cloneable
      */
-    protected native Object clone() throws CloneNotSupportedException;
+    protected native @PolySameLen Object clone(@PolySameLen Object this) throws CloneNotSupportedException;
```
kelloggm: I added this because a test was failing without it; I don't remember which one. Either way, I think leaving this as an explicit annotation makes what's occurring clearer, so even if it isn't entirely necessary I'd advocate keeping it.

checker/src/org/checkerframework/checker/index/samelen/SameLenVisitor.java
kelloggm: I tried this, and I couldn't quite get it to work. The problem I ran into is that to remove it you need to get the canonical name of the expression from the lhs' Tree. This is usually a `VariableTree`, but the `ExpressionTree` that I ought to get by calling `VariableTree#getNameExpression()` is always `null`, even though calling `getName()` on the same `VariableTree` results in a reasonable (but not canonical) name for the new variable. If `getNameExpression` actually returned an `ExpressionTree`, I could call `FlowExpressions.getInternalReprOf` to get a `Receiver`, which would have the canonical name if it was parsable to a String.

kelloggm: ...But this is also a problem with the code as is. So in principle I don't see any reason we couldn't do it this way, but regardless I need to be able to get the canonical name of the lhs of a `VariableTree`. Do you know how to do this? @smillst

smillst: Convert the VariableTree to an Element.  If the ElementKind is  a LOCAL_VARIABLE, RESOURCE_VARIABLE, EXCEPTION_PARAMETER, or PARAMETER, `getName()` returns the canonical name.  Otherwise, do something like this:

```
Receiver  receiverF = FlowExpressions.internalRepOfImplicitReceiver(ele);
FlowExpressionContext context = new FlowExpressionContext(receiverF, null, factory.getContext());
Receiver result =FlowExpressionParseUtil.parse(tree.getName(), context, getCurrentTreePath(), false);
```

This should probably be added as a new method to FlowExpressions. (`getInternalReprOf(Variable tree,...)`)

smillst:
Is this ready for me to review again?

kelloggm:
Yes. I thought I had reassigned you but it looks like I didn't. My apologies.

checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
```diff
@@ -73,6 +77,52 @@ public QualifierHierarchy createQualifierHierarchy(MultiGraphFactory factory) {
         return new SameLenQualifierHierarchy(factory);
     }
 
+    @Override
+    public AnnotatedTypeMirror getAnnotatedTypeLhs(Tree tree) {
+        AnnotatedTypeMirror atm = super.getAnnotatedTypeLhs(tree);
+        if (tree.getKind() == Tree.Kind.VARIABLE) {
+            String varName;
```
smillst: Could you please move this code to org.checkerframework.dataflow.analysis.FlowExpressions, to a new method `internalReprOfVariable(AnnotationProvider provider, FieldAccessNode node) `?

kelloggm: If I did, I wouldn't be able to call it from here, because I don't have a `FieldAccessNode`, which is why I didn't write this in `FlowExpressions` in the first place.

smillst: Sorry, I meant `internalReprOfVariable(AnnotationProvider provider, VariableTree tree)`

kelloggm: Sure, I can do that.

smillst: Ok, you can't because org.checkerframework.framework.util.FlowExpressionParseUtil.FlowExpressionContext uses a Context which is in the framework package.

smillst: I guess you could put it in FlowExpressionParseUtil...I can't find a better place than that.

kelloggm: I remember now that this was the problem with doing this in `FlowExpressions` I ran into originally (I distinctly remember you making this suggestion before). I'll move it to `FlowExpressionParseUtil`

smillst: Thanks!

checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
```diff
@@ -73,6 +77,42 @@ public QualifierHierarchy createQualifierHierarchy(MultiGraphFactory factory) {
         return new SameLenQualifierHierarchy(factory);
     }
 
+    @Override
+    public AnnotatedTypeMirror getAnnotatedTypeLhs(Tree tree) {
+        AnnotatedTypeMirror atm = super.getAnnotatedTypeLhs(tree);
+        if (tree.getKind() == Tree.Kind.VARIABLE) {
+            String varName = null;
+            Element elt = TreeUtils.elementFromDeclaration((VariableTree) tree);
+            if (elt.getKind() == ElementKind.LOCAL_VARIABLE
```
smillst: Could you please put this code in `internalReprOfVariable` so that is can handle all kinds of VariableTrees?

kelloggm: That code doesn't result in a `Receiver`, though - it results in a `String` (which is what I need here; I only get a `Receiver` in the other cases because that's the most straightforward way to get the name.

kelloggm: Because of that, I'm not sure what you're asking for here - should `internalReprOfVariable` return a `String`? That seems strange. I'm not sure how to go from an `Element` to a `Receiver`, either.

kelloggm: No worries. Thanks for your help!

PR #1321, opened 2017-06-06
Return unique values from methods in ValueAnnotatedTypeFactory

panacekcz:
I noticed that in `ValueAnnotatedTypeFactory`, there are cases where the result of `ValueCheckerUtils.removeDuplicates` is ignored, making it a no-op.

This apparently unintended behavior is probably not observable, but the javadoc of `getIntValues`, `getDoubleValues` and `getArrayLength` requires that the returned lists are sorted and have no duplicates.

This change uses the return value of `ValueCheckerUtils.removeDuplicates` to ensure this behavior and fixes a typo in the javadoc describing it.


wmdietl:
@mernst How about we start using http://errorprone.info/bugpattern/CheckReturnValue ? The result of `removeDuplicates` should never be ignored.

mernst:
I don't object to that (though it would be a shame to have to use the obsolete JSR 305 annotations that error-prone assumes).

There is a check-errorprone target, but it issues warnings:  apparently no one thought it was worth fixing those errors.  Doing so would enable us to fix enable error-prone on Travis.

PR #1322, opened 2017-06-06
Don't apply @StringVal to byte or char arrays.

smillst:
The `@ArrayLen` annotation was originally added to the Value Checker to track the length of Class arrays passed to Method.invoke in order to determine which method was beginning invoked. 

In case studies, we found the byte and char arrays were often converted to/from constant Strings, so adding `@StringVal` to them improved those case studies.  (We were only interested in constant propagation of Strings.)  Now that we are using the Value Checker for estimating array lengths, adding `@StringVal` to byte and char arrays interferes with estimating array lengths.  This pull requests removes this functionality. 

Also, several methods that use byte or char arrays are no longer statically executable, so was updated statically-executable.astub.  All constructors of a primitive wrapper or String were hard coded as statically-executable, but now some of the String constructors are not.


wmdietl:
Doesn't this harm the case studies that relied on having these `@StringVal` annotations?
Would an alternative be that we treat a `@StringVal("abc")` as a subtype of `@ArrayLen(3)`?
StringVal on an array strictly gives more information than ArrayLen and it seems like a petty to throw that functionality away, in particular as we used it in case studies.

panacekcz:
> Would an alternative be that we treat a `@StringVal("abc")` as a subtype of `@ArrayLen(3)`?

I am implementing that and it does not work well with `@StringVal` on `byte[]`. That's why I brought it up in the gsoc mailing list.

Let me state arguments against the feature proposed for removal:
1. Conversion between `String` and `byte[]` depends on the platform's character set. 
2. The `byte[]` can have a different length than the corresponding string.
3. Arrays can be mutated to have a different value than specified by the annotation. 

The first point means that the checker can currently give different results on different platforms. As an example, `byte@StringVal("\u00C3\u00A1")[]x={-61,-95};` gives an error on Linux but not on Windows. And that's not even bringing in the possibility that the character set at run time may be different.

The second point is why it is not compatible with making `@StringVal` a subtype of `@ArrayLen`. The following code currently gives no error on Linux (utf-8 character set):
```java
byte@StringVal("\u00e1")[]x={-61,-95};
char@StringVal("\u00e1")[]y={0xe1};
```

If `@StringVal("\u00e1")` is a subtype of `@ArrayLen(1)`, then `x` would be convertible to `byte@ArrayLen(1)[]`, but the correct type is `byte@ArrayLen(2)[]`.

wmdietl:
Thanks for giving this detailed motivation!
Points 1 and 2 only apply to byte arrays, right? Are they arguments for removing support for char arrays?
Point 3 could be handled by either updating the value of `StringVal` or by losing information and only retaining the length.

wmdietl:
@panacekcz Adding a mention, b/c I can't assign you.

panacekcz:
> Points 1 and 2 only apply to byte arrays, right?

Yes.

> Are they arguments for removing support for char arrays?

No, the only problems I see with `char[]` is point 3 and some implementation details.

(Regarding my work, I would be happy if the support for `byte[]` was removed soon and the mutability could be dealt with any time later.)

wmdietl:
I'm fine with removing support for byte arrays and opening a new issue that we need to correctly handle mutation of char arrays.

There is no change to the manual in this PR, so maybe neither byte nor char array support is documented. A short addition would be good and keep track of the discussion we had.

@smillst How does this sound to you?

smillst:
I agree with this comment https://github.com/typetools/checker-framework/pull/1322#issuecomment-307440896

smillst:
I'm going to open a new pull request that only removes support of `@StringVal` on byte arrays.

PR #1327, opened 2017-06-08
Issue a warning for cast to type variables,

smillst:
where the type of the casted expression is not a type variable and its qualifer is not a subtype of the tyep variable's lower bound.


framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1429,63 +1438,69 @@ protected void checkTypecastSafety(TypeCastTree node, Void p) {
...
-                } else {
-                    newCastType = castType;
-                }
-                AnnotatedTypeMirror newExprType;
-                if (exprType.getKind() == TypeKind.TYPEVAR) {
-                    newExprType = ((AnnotatedTypeVariable) exprType).getUpperBound();
-                } else {
-                    newExprType = exprType;
-                }
+        if (checker.hasOption("checkCastElementType")) {
```
smillst: This code is the same, I only removed the surround if statement.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1429,63 +1438,69 @@ protected void checkTypecastSafety(TypeCastTree node, Void p) {
...
-                                ((AnnotatedDeclaredType) newExprType).getTypeArguments().size();
-
-                        if (castSize != exprSize) {
-                            // Always warn if the cast and expression contain a different number of
-                            // type arguments, e.g. to catch a cast from "Object" to "List<@NonNull Object>".
-                            // TODO: the same number of arguments actually doesn't guarantee anything.
-                            isSubtype = false;
-                        }
-                    }
+            if (!atypeFactory.getTypeHierarchy().isSubtype(newExprType, newCastType)) {
```
smillst: This code is also the same -- except it returns false rather than setting isSubtype to false.

smillst:
We decided to only issue a `cast.unsafe` in these cases if the `checkCastElementType`option is passed.

checker/jtreg/tainting/NewClass.java
```diff
@@ -0,0 +1,20 @@
+/*
+ * @test
+ * @summary Test for bug where arguments to constructors were visited twice.
+ *
+ * @compile/fail/ref=NewClass.out -XDrawDiagnostics -processor org.checkerframework.checker.tainting.TaintingChecker -AprintErrorStack -Alint NewClass.java
+ */
+
+import org.checkerframework.checker.tainting.qual.Untainted;
+
+public class NewClass {
```
wmdietlGC: Do you understand why this file shows up in this PR? It's already merged in master?

smillst: Not really sure why that happened.  But I merged master again and the file went away.  Thanks for pointing it out!

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1429,63 +1443,74 @@ protected void checkTypecastSafety(TypeCastTree node, Void p) {
...
-                                        castType.getEffectiveAnnotations());
+                castAnnos = castType.getAnnotations();
             }
-        }
 
-        if (!isSubtype) {
-            checker.report(Result.warning("cast.unsafe", exprType, castType), node);
+            return qualifierHierarchy.isSubtype(exprType.getEffectiveAnnotations(), castAnnos);
         }
+        return qualifierHierarchy.isSubtype(
```
wmdietlGC: I would move this into an `else` block, to clarify the control flow a little. The only way to get here is if `checkCastElementType` isn't provided, right? Maybe add that as a comment in the else block.

PR #1328, opened 2017-06-08
Change test of TreeSet into test of HashSet

mernst:
TreeSet may permit nulls, but HashSet does not.


mernst:
This pull request requires that the annotated JDK be updated (the prebuilt one that the tests use).
At that point, master will start to fail and this pull request should be merged.

smillst:
A TreeSet only allows null if its comparator allows null.  See discussion:
https://github.com/typetools/checker-framework/pull/1179

The following code issues a NullPointerException.
```
TreeSet<@Nullable Integer> ts = new TreeSet<>();
ts.add(null);
```

mernst:
A TreeSet can permit null.  Do you propose that `TreeSet<@Nullable Integer>` is always an illegal type and users have to suppress the warning if they are sure that no null pointer exception will occur?

mernst:
A better solution could be that `new TreeSet<>()` has type `TreeSet<@NonNull T>`, but other treesets can have nullable elements.

smillst:
>A better solution could be that `new TreeSet<>()` has type `TreeSet<@NonNull T>`

I agree that this is the better solution, though we don't currently have a way to express this. (You asked about this to [the mailing list](https://groups.google.com/d/msg/checker-framework-dev/kXaB0qRNmX0/5wNgJRY9EQAJ).  I think we discussed it in person, which is why there is no reply.)

Since we can't currently implement the best solution, then either `TreeSet<@Nullable T>` should always issue a warning or we allow a false negatives, such as this test case.  Either way, we should add a comment to TreeSet, so that we remember our reasoning.

mernst:
I created issue #1330 to track this.  I propose that we leave TreeSet's annotation as it is (unsound) for now, but fixing issue #1330 will solve the problem.

smillst:
Sounds good.  I'll remake the annotated JDK.

PR #1338, opened 2017-06-11
Fix copypasta in "Inherited wildcard annotations"

aabdagic:
If `class MyList<@Nullable T extends @Nullable Object> {}`, then `T` can only be `@Nullable` so the two examples wouldn't compile. But if the lower bound is `@NonNull` (or omitted), then they make sense.


mernst:
Thanks for noticing the error in the manual, and thanks for the patch.

PR #1341, opened 2017-06-12
Don't apply @StringVal to byte arrays.

smillst:
Also, several methods that use byte or char arrays are no longer statically executable, so was updated statically-executable.astub. All constructors of a primitive wrapper or String were hard coded as statically-executable, but now some of the String constructors are not.

@panacekcz is implementing treating `@StringVal("abc")` as a subtype of `@ArrayLen(3)`.

(This is a follow up to PR #1322 which removed `@StringVal` from both byte and char arrays.)

### DIFF
--- a/framework/src/org/checkerframework/common/value/ReflectiveEvaluator.java
+++ b/framework/src/org/checkerframework/common/value/ReflectiveEvaluator.java
@@ -288,7 +288,7 @@ public class ReflectiveEvaluator {
             for (Object[] arguments : listOfArguments) {
                 try {
                     results.add(constructor.newInstance(arguments));
-                } catch (ReflectiveOperationException e) {
+                } catch (Exception e) {
                     if (reportWarnings) {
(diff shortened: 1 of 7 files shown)


framework/src/org/checkerframework/common/value/ReflectiveEvaluator.java
```diff
@@ -288,7 +288,7 @@ public Object evaluateStaticFieldAccess(
             for (Object[] arguments : listOfArguments) {
                 try {
                     results.add(constructor.newInstance(arguments));
-                } catch (ReflectiveOperationException e) {
+                } catch (Exception e) {
```
wmdietlGC: This try/catch is nested within another try/catch for `ReflectiveOperationException`.
Should the catch type of the outer try/catch also be changed to `Exception`?
Can the scope of the outer try/catch be limited to whichever other method might throw this exception?

PR #1335, opened 2017-06-09
Return uninferred type arguments when inference fails.

smillst:
Fixes #1308, Fixes #1312, Fixes  #1313, Fixes #1331, Fixes  #1332

PR #1343, opened 2017-06-13
Issue an error when type argument inference fails.

smillst:


smillst:
@wmdietl [said](https://github.com/typetools/checker-framework/pull/1335#discussion_r121551878)
>I like the idea of outputting a message if we couldn't infer some type arguments! Should it be a warning for every time we failed doing this? Or should it be a single notice for the whole compilation, with an option to issue a warning for every instance (similar to how unchecked warnings work)? The latter would allow us to work even with -werror.
Maybe if -AconservativeUninferredTypeArguments it is a warning for each instance and without -AconservativeUninferredTypeArguments it is a single notice?

I think it should be an error at each location.  That way the user can either explicitly write the type argument or suppress the error.

smillst:
@wmdietl  I changed it so the warnings is only issued if conservativeUninferredTypeArguments is used.  I haven't implemented issuing the warning once if conservativeUninferredTypeArguments is not used.  Any ideas on how best to do this?

wmdietlGC:
Regarding issuing a warning only once: I'm not sure whether we have done that anywhere yet.
I would look at `DeferredDiagnosticHandler` in `jsr308-langtools/src/share/classes/com/sun/tools/javac/util/Log.java` and how it is used in `jsr308-langtools/src/share/classes/com/sun/tools/javac/comp/Check.java`.

smillst:
I haven't gotten back around to issuing a warning if `conservativeUninferredTypeArguments` is not passed.  I opened an issue for it #1355.  I'm going to merge this pull request.

PR #1346, opened 2017-06-15
Fix Index Checker soundness bug

kelloggm:
The Index Checker currently does not correctly invalidate types when an array is reassigned. This pull request fixes that problem for the Upper Bound Checker (but not for the SameLen Checker). The relevant issue is kelloggm#30.

To do so, it creates a new `UpperBoundStore` class that replaces `CFStore` for the Upper Bound Checker, which overwrites `updateForAssignment` and `updateForMethodCall`. This new store then checks that the reassignment is safe, and rejects code that is not (following the proposal in [this document](https://docs.google.com/document/d/1820riCds1WEHiOWqWPHi4Kiif7lyGtERgDY4HKvRCEY/edit). 

This PR also issues a warning when the programmer writes an annotation on a local variable, because I could not find a way to find the annotations on in-scope local variables given an `Element`. This required a significant refactoring of many of the test cases for the UBC, which is reflected in this PR as well.

There are a number of nasty implementation choices in this code, and I'd appreciate some help in removing them. Notable ones:
* ~~UpperBoundTransfer no longer can inherit from `IndexAbstractTransfer` due to generics, so I just copied the code there, which isn't a great thing to do.~~ Fixed 7b91b6920
* The code in `UpperBoundStore` that actually handles checking annotations is difficult to read and slow to run. Any ideas for how to improve it are welcome.


checker/tests/index/Reassignment.java
```diff
@@ -3,12 +3,46 @@
 // problem but still permits some unsoundness is worse than no solution and
 // an obvious issue.
 
+import org.checkerframework.checker.index.qual.*;
```
smillst: Unskip?

kelloggm:
Yeah, I meant to unskip this test. Sorry about that.

On Thu, Jun 15, 2017 at 1:12 PM, Suzanne Millstein <notifications@github.com

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundStore.java
```diff
@@ -0,0 +1,337 @@
...
+        // Do reassignment things here.
+        super.updateForAssignment(n, val);
+
+        // This code determines the list of dependences in types that are to be invalidated
+        CheckController checkController = null;
+
+        if (n.getType().getKind() == TypeKind.ARRAY) {
+            if (n instanceof LocalVariableNode) {
+                // Do not warn about assigning to a final variable. javac handles this.
+                if (!ElementUtils.isFinal(((LocalVariableNode) n).getElement())) {
```
smillst: Should this be a call to effectivelyFinal?

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundStore.java
```diff
@@ -0,0 +1,337 @@
...
+        }
+
+        // Find all possibly-invalidated types
+        if (checkController != null) {
+
+            Element elt;
+
+            // So that assignments into arrays are treated correctly, as well as type casts
+            Node oldN = n;
+            while (n instanceof TypeCastNode || n instanceof ArrayAccessNode) {
```
smillst: If `n` is not a FieldAccessNode nor a LocalVariableNode, then `checkController` is null. So this condition is always false.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundStore.java
```diff
@@ -0,0 +1,337 @@
...
+import org.checkerframework.dataflow.cfg.node.TypeCastNode;
+import org.checkerframework.framework.flow.CFAbstractStore;
+import org.checkerframework.framework.flow.CFValue;
+import org.checkerframework.framework.source.Result;
+import org.checkerframework.framework.type.AnnotatedTypeFactory;
+import org.checkerframework.framework.type.AnnotatedTypeMirror;
+import org.checkerframework.javacutil.AnnotationUtils;
+import org.checkerframework.javacutil.ElementUtils;
+
+public class UpperBoundStore extends CFAbstractStore<CFValue, UpperBoundStore> {
```
smillst: Somewhere in this pull request, you should include the contents of the google doc describing this scheme.  It makes more sense to me to put it as a class comment here.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundStore.java
```diff
@@ -0,0 +1,337 @@
...
+                List<? extends Element> rgparam = ee.getParameters();
+                for (Element param : rgparam) {
+                    AnnotatedTypeMirror atmP = analysis.getTypeFactory().getAnnotatedType(param);
+                    enclosedTypes.add(atmP);
+                }
+            }
+        }
+    }
+
+    private boolean checkAnno(
```
smillst: A comment here would be helpful.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundStore.java
```diff
@@ -0,0 +1,337 @@
...
+                                        == CheckController.NONFINAL_REFS_AND_METHOD_CALLS_METHOD) {
+                            if (!((FlowExpressions.FieldAccess) r)
+                                            .getReceiver()
+                                            .toString()
+                                            .equals("this")
+                                    && !((FlowExpressions.FieldAccess) r).isFinal()) {
+                                if (report
+                                        && checkController
+                                                != CheckController
+                                                        .NONFINAL_REFS_AND_METHOD_CALLS_METHOD) {
```
smillst: This whole method is unreadable.  Could you please do something to reduce the indentation level so that the code formats with less line breaks?  (Extract code to helper methods?  Only put code that might throw an exception in the try block?

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundStore.java
```diff
@@ -0,0 +1,337 @@
...
+                                                n.getTree());
+                            }
+                            return true;
+                        }
+                        for (FlowExpressions.Receiver param :
+                                ((FlowExpressions.MethodCall) r).getParameters()) {
+                            canonicalStrings.add(param.toString());
+                        }
+                    }
+                } catch (Exception e) { //FlowExpressionParseUtil.FlowExpressionParseException e) {
```
smillst: EVIL!

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundStore.java
smillst: (You can address the other comments before doing this and then reassign to me.)

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundStore.java
kelloggm: Yes. These should all be calls to `isEffectivelyFinal`, and I'm not sure how I screwed that up.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundStore.java
```diff
@@ -0,0 +1,348 @@
...
+    private static /*@CompilerMessageKey*/ String NO_REASSIGN_FIELD =
+            "reassignment.field.not.permitted";
+    private static /*@CompilerMessageKey*/ String SIDE_EFFECTING_METHOD =
+            "side.effect.invalidation";
+    private static /*@CompilerMessageKey*/ String NO_REASSIGN_FIELD_METHOD =
+            "reassignment.field.not.permitted.method";
+    private SourceChecker checker;
+
+    private enum CheckController {
+        NAME_ONLY,
```
smillst: NAME of what?

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundStore.java
```diff
@@ -0,0 +1,348 @@
...
+        NAME_AND_METHOD_CALLS,
+        NONFINAL_REFS_AND_METHOD_CALLS,
+        NONFINAL_REFS_AND_METHOD_CALLS_METHOD
```
smillst: I don't understand METHOD_CALLS_METHOD

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundStore.java
```diff
@@ -0,0 +1,348 @@
...
+            }
+            canonicalDependencies.add(r.toString());
+            while (r instanceof FlowExpressions.FieldAccess) {
+                // I included an exception here for "this", because otherwise the rules don't make sense - any
+                // field, including final ones, would be invalidated, since it's impossible to know if this is
+                // final or not. But this can't be modified by calling other code, so it's fine.
+                if (checkController == CheckController.NONFINAL_REFS_AND_METHOD_CALLS
+                        || checkController
+                                == CheckController.NONFINAL_REFS_AND_METHOD_CALLS_METHOD) {
+                    if (!((FlowExpressions.FieldAccess) r).getReceiver().toString().equals("this")
```
smillst: This be` instanceOf ThisReference` rather than comparing strings. Or would `FlowExpressions#isUnmodifiableByOtherCode` work instead of `isFinal`?  That method accounts for receivers correctly.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundStore.java
```diff
@@ -0,0 +1,348 @@
...
+    void checkForRemainingAnnotations(
+            List<AnnotatedTypeMirror> enclosedTypes,
+            String canonicalTargetName,
+            Node n,
+            CheckController checkController) {
+        for (AnnotatedTypeMirror atm : enclosedTypes) {
+            for (AnnotationMirror anno : atm.getAnnotations()) {
+                Result r = checkAnno(anno, canonicalTargetName, n, checkController);
+                if (r.isFailure()) {
+                    checker.report(r, n.getTree());
```
smillst: Typically only the Visitor issues errors.  I don't understand the code well enough to suggest how these error might be moved to the Visitor.

kelloggm: I thought about that and couldn't see any way to do it, but you know the codebase better than I do. We should revisit this issue once we've improved this code.

mernst:
@kelloggm The current 11 errors are all of the form:
```
CharSequenceTest.java:56: warning: [local.variable.unsafe.dependent.annotation] Dependent type annotations on local variables are not supported by the Upper Bound Checker.
        @IndexOrHigh("cs") int l = cs.length();
                               ^
```

Could you adjust the tests and see whether you can complete this pull request?  Thanks!

mernst:
There is a problem with this pull request that is not apparent in the failing tests.

When building the annotated JDK, I got 6 occurrences of
"error: SourceChecker.typeProcess: unexpected Throwable (NullPointerException)":

```
error: SourceChecker.typeProcess: unexpected Throwable (NullPointerException) while processing src/java/lang/Class.java
error: SourceChecker.typeProcess: unexpected Throwable (NullPointerException) while processing src/java/lang/ClassLoader.java
error: SourceChecker.typeProcess: unexpected Throwable (NullPointerException) while processing src/java/io/ObjectStreamClass.java
error: SourceChecker.typeProcess: unexpected Throwable (NullPointerException) while processing src/java/util/regex/Matcher.java
error: SourceChecker.typeProcess: unexpected Throwable (NullPointerException) while processing src/java/sun/util/LocaleServiceProviderPool.java
error: SourceChecker.typeProcess: unexpected Throwable (NullPointerException) while processing src/java/util/Collections.java
```

Here is one of the stack traces:

```
error: SourceChecker.typeProcess: unexpected Throwable (NullPointerException) while processing src/java/lang/Class.java
  Compilation unit: src/java/lang/Class.java
  Exception: java.lang.NullPointerException; Stack trace: org.checkerframework.javacutil.InternalUtils.typeOf(InternalUtils.java:233)
  org.checkerframework.framework.type.GenericAnnotatedTypeFactory.getReceiverFromJavaExpressionString(GenericAnnotatedTypeFactory.java:761)
  org.checkerframework.checker.index.upperbound.UpperBoundUtil.getDependentReceivers(UpperBoundUtil.java:143)
  org.checkerframework.checker.index.upperbound.UpperBoundUtil.getDependentReceivers(UpperBoundUtil.java:160)
  org.checkerframework.checker.index.upperbound.UpperBoundStore.clearFromStore(UpperBoundStore.java:134)
  org.checkerframework.checker.index.upperbound.UpperBoundStore.updateForAssignment(UpperBoundStore.java:112)
  org.checkerframework.checker.index.upperbound.UpperBoundStore.updateForAssignment(UpperBoundStore.java:64)
  org.checkerframework.framework.flow.CFAbstractTransfer.processCommonAssignment(CFAbstractTransfer.java:857)
  org.checkerframework.framework.flow.CFAbstractTransfer.visitAssignment(CFAbstractTransfer.java:795)
  org.checkerframework.checker.index.upperbound.UpperBoundTransfer.visitAssignment(UpperBoundTransfer.java:52)
  org.checkerframework.checker.index.upperbound.UpperBoundTransfer.visitAssignment(UpperBoundTransfer.java:37)
  org.checkerframework.dataflow.cfg.node.AssignmentNode.accept(AssignmentNode.java:63)
  org.checkerframework.dataflow.analysis.Analysis.callTransferFunction(Analysis.java:404)
  org.checkerframework.dataflow.analysis.Analysis.performAnalysis(Analysis.java:200)
  org.checkerframework.framework.type.GenericAnnotatedTypeFactory.analyze(GenericAnnotatedTypeFactory.java:1116)
  org.checkerframework.framework.type.GenericAnnotatedTypeFactory.analyze(GenericAnnotatedTypeFactory.java:1073)
  org.checkerframework.framework.type.GenericAnnotatedTypeFactory.performFlowAnalysis(GenericAnnotatedTypeFactory.java:1003)
  org.checkerframework.framework.type.GenericAnnotatedTypeFactory.checkAndPerformFlowAnalysis(GenericAnnotatedTypeFactory.java:1340)
  org.checkerframework.framework.type.GenericAnnotatedTypeFactory.preProcessClassTree(GenericAnnotatedTypeFactory.java:224)
  org.checkerframework.common.basetype.BaseTypeVisitor.visitClass(BaseTypeVisitor.java:281)
  org.checkerframework.common.basetype.BaseTypeVisitor.visitClass(BaseTypeVisitor.java:160)
  com.sun.tools.javac.tree.JCTree$JCClassDecl.accept(JCTree.java:720)
  com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:50)
  org.checkerframework.framework.source.SourceVisitor.visit(SourceVisitor.java:66)
  org.checkerframework.framework.source.SourceChecker.typeProcess(SourceChecker.java:962)
  org.checkerframework.common.basetype.BaseTypeChecker.typeProcess(BaseTypeChecker.java:499)
  org.checkerframework.javacutil.AbstractTypeProcessor$AttributionTaskListener.finished(AbstractTypeProcessor.java:185)
  com.sun.tools.javac.api.ClientCodeWrapper$WrappedTaskListener.finished(ClientCodeWrapper.java:681)
  com.sun.tools.javac.api.MultiTaskListener.finished(MultiTaskListener.java:111)
  com.sun.tools.javac.main.JavaCompiler.flow(JavaCompiler.java:1342)
  com.sun.tools.javac.main.JavaCompiler.flow(JavaCompiler.java:1296)
  com.sun.tools.javac.main.JavaCompiler.compile2(JavaCompiler.java:901)
  com.sun.tools.javac.main.JavaCompiler.compile(JavaCompiler.java:860)
  com.sun.tools.javac.main.Main.compile(Main.java:523)
  com.sun.tools.javac.main.Main.compile(Main.java:381)
  com.sun.tools.javac.main.Main.compile(Main.java:370)
  com.sun.tools.javac.main.Main.compile(Main.java:361)
  com.sun.tools.javac.Main.compile(Main.java:56)
  com.sun.tools.javac.Main.main(Main.java:42)
```

Could you please reproduce the error with a minimal test case, commit it to the repository, and fix it?

kelloggm:
I determined the cause of the null pointer exceptions while building the JDK. In [this commit](https://github.com/typetools/checker-framework/commit/0712903c13910dbd7d228730a7232a325b0ad337) special handling was added for finally blocks that might throw exceptions: a generated tree is included in the CFG. This causes the enclosing class of the tree associated with the node in `UpperBoundStore#clearFromStore` to be null. That causes an NPE while trying to parse the a variable with a Java expression in its annotation.

There is a minimized test case in `tests/index/ObjectOutputStreamCrash.java`. The `finally` block is what causes the NPE directly, but everything else needs to be present. The variable that's being unrefined when the crash occurs is `i`.

I added code to ignore the NPE, but I'm not sure that's the right thing to do. I think that it's unsound in the case where the `finally` block throws an exception, but that doesn't seem that important, since the program is already crashing.

kelloggm:
@smillst, do you mind taking another look at this PR. I've (finally) fixed all the issues with the Travis build, so I think this PR is finally again a candidate for being merged into master. Thanks :)

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundStore.java
```diff
@@ -0,0 +1,153 @@
...
+                                entry.getValue().getAnnotations(),
+                                path,
+                                (UpperBoundAnnotatedTypeFactory) analysis.getTypeFactory())) {
+                    if (UpperBoundUtil.isSideEffected(dependent, reassignedVar, sideEffect)
+                            != UpperBoundUtil.SideEffectError.NO_ERROR) {
+                        this.clearValue(r);
+                    }
+                }
+            } catch (NullPointerException e) {
+                // This happens in the second round of type-checking of a finally block,
```
smillst: Is it `path` that is null?  If so, then this is related to #1520.  Also, if that's the case, check if `path` is  null and just return if it is.  (And move the comment there.)

(Technically, it's not the second round of type-checking, it's performing the dataflow analysis on the finally block that has been duplicated to represent the exceptional case.)

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -36,6 +55,129 @@ public UpperBoundVisitor(BaseTypeChecker checker) {
...
+     * which must unrefine all qualifiers, but which cannot collect all in-scope local variables. So
+     * that there are no local variables with qualifiers that are not in the store, this method
+     * forbids programmers from writing such qualifiers. This may introduce false positives if the
+     * programmer would like to write a non-primary annotation on a local variable.
+     *
+     * <p>Also issues warnings if the expression in a dependent annotation is not permitted.
+     * Permitted expressions in dependent annotations are:
+     *
+     * <ul>
+     *   <li>final or effectively final variables
```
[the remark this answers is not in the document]
kelloggm: No. This is any final or effectively final variable, in any scope. Because such variables can't be reassigned, there's no need to prevent users from writing annotations that are dependent on them: they can never be invalidated by a side-effect, since no side-effect can reassign them!

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -36,6 +55,129 @@ public UpperBoundVisitor(BaseTypeChecker checker) {
...
+    @Override
+    public Void visitVariable(VariableTree node, Void p) {
+        AnnotatedTypeMirror atm = atypeFactory.getAnnotatedTypeLhs(node);
+        AnnotationMirror anm = atm.getAnnotationInHierarchy(atypeFactory.UNKNOWN);
+        if (anm != null && AnnotationUtils.hasElementValue(anm, "value")) {
+            // This is a dependent annotation. If this is a local variable,
+            // issue a warning; dependent annotations should not be written on local variables.
+            Element elt = TreeUtils.elementFromDeclaration(node);
+            if (elt.getKind() == ElementKind.LOCAL_VARIABLE) {
+                checker.report(Result.warning(LOCAL_VAR_ANNO), node);
```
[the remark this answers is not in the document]
kelloggm: I think so, for the same reason we can allow any final variable in a dependent type. I'll make this change.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundStore.java
kelloggm: `path` is null, but that's not the direct cause of the NPE. Checking if `path` is `null` is too broad - I tried that, and it introduces additional unsoundness (possibly related to #1520) as `path` is `null` sometimes even when no NPE is thrown.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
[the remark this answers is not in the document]
smillst: I mean what kinds of variables have to be final or effectively final? Local variables and private fields are allowed even if they are not final or effectively final.  So which variables have to be final or effectively final?  Non-private fields and methods parameters?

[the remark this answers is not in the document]
kelloggm: I see the confusion. If a variable has any one of the properties on this list, it is permitted. So, being final or effectively final is sufficient by itself; so is being a private field; so is being a local variable; etc. I can rephrase this comment to make this clear.

smillst:
Tests are failing again.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundUtil.java
```diff
@@ -0,0 +1,166 @@
+package org.checkerframework.checker.index.upperbound;
+
+/*>>>
```
smillst: The Checker Framework is now compiled with Java 8, so annotations don't have to be in comments.  (Unless they are annotations in the framework package that are declared in the checker package.)  [The manual](https://checkerframework.org/manual/#uncommenting-annotations) gives Unix commands to easily uncomment the annotations.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -33,6 +52,150 @@ public UpperBoundVisitor(BaseTypeChecker checker) {
...
+            Element elt = TreeUtils.elementFromDeclaration(node);
+            boolean allVariablesFinal = true;
+
+            UBQualifier qual = UBQualifier.createUBQualifier(anm);
+            if (qual.isLessThanLengthQualifier()) {
+                LessThanLengthOf ltl = (LessThanLengthOf) qual;
+                for (String array : ltl.getSequences()) {
+                    Receiver rec = getReceiverForCheckingDependentTypes(array, node);
+                    // covers final and effectively final variables
+                    boolean isFinal =
```
smillst: All LocalVariable are not final, so this variable's name isn't correct.  (Use org.checkerframework.javacutil.ElementUtils#isEffectivelyFinal for local variables.)

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -33,6 +52,150 @@ public UpperBoundVisitor(BaseTypeChecker checker) {
...
+
+        return super.visitVariable(node, p);
+    }
+
+    private Receiver getReceiverForCheckingDependentTypes(String expr, Tree tree) {
+        try {
+            return atypeFactory.getReceiverFromJavaExpressionString(expr, getCurrentPath());
+        } catch (FlowExpressionParseUtil.FlowExpressionParseException e) {
+            // issue warning
+            checker.report(
```
smillst: If the expression isn't parsable, then shouldn't the checker already have issued a warning?  Is this really a duplicate warning?

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -33,6 +52,150 @@ public UpperBoundVisitor(BaseTypeChecker checker) {
...
+            return false;
+        }
+
+        if (rec.isUnmodifiableByOtherCode() || rec instanceof FlowExpressions.LocalVariable) {
+            return true;
+        }
+
+        if (rec instanceof FlowExpressions.FieldAccess) {
+            FlowExpressions.FieldAccess faRec = (FlowExpressions.FieldAccess) rec;
+            if (faRec.getField().getModifiers().contains(Modifier.PRIVATE)
```
smillst: Why isn't the receiver of a private field checked to see if it is permitted?

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -33,6 +52,150 @@ public UpperBoundVisitor(BaseTypeChecker checker) {
...
+                    || (faRec.getField().getModifiers().contains(Modifier.PUBLIC)
```
smillst: What about protected fields and package private fields?  Are they never permitted?

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -142,6 +305,94 @@ private void visitAccess(ExpressionTree indexTree, ExpressionTree arrTree) {
...
+
+    private void checkAnnotationsInClass(
+            Element elt,
+            Receiver canonicalTargetName,
+            Tree tree,
+            TreePath path,
+            SideEffectKind sideEffectKind) {
+        List<? extends Element> enclosedElts =
+                ElementUtils.enclosingClass(elt).getEnclosedElements();
+        List<AnnotatedTypeMirror> enclosedTypes = findEnclosedTypes(enclosedElts);
```
smillst: It seems extremely costly to recalculate this list for each call.  (Plus calling `UpperBoundUtil.getDependentReceivers(atm.getAnnotations(), path, atypeFactory)` on each type every time.) . Can you calculate the list of receivers once per class?

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -142,6 +305,94 @@ private void visitAccess(ExpressionTree indexTree, ExpressionTree arrTree) {
...
+            } else {
+                return;
+            }
+            checkAnnotationsInClass(elt, field, varTree, getCurrentPath(), kind);
+        }
+    }
+
+    private void checkAnnotationsInClass(
+            Element elt,
+            Receiver canonicalTargetName,
```
smillst: This parameter name doesn't make sense.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -142,6 +305,94 @@ private void visitAccess(ExpressionTree indexTree, ExpressionTree arrTree) {
...
+            } else if (typeKind == TypeKind.ARRAY) {
+                kind = SideEffectKind.LOCAL_VAR_REASSIGNMENT;
+            } else {
+                return;
+            }
+            checkAnnotationsInClass(elt, field, varTree, getCurrentPath(), kind);
+        }
+    }
+
+    private void checkAnnotationsInClass(
```
smillst: Documentation?  I'm not sure what this method is suppose to be doing.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -142,6 +305,94 @@ private void visitAccess(ExpressionTree indexTree, ExpressionTree arrTree) {
...
+            }
+        }
+    }
+
+    private List<AnnotatedTypeMirror> findEnclosedTypes(List<? extends Element> enclosedElts) {
+        List<AnnotatedTypeMirror> enclosedTypes = new ArrayList<>();
+        for (Element e : enclosedElts) {
+            AnnotatedTypeMirror atm = atypeFactory.getAnnotatedType(e);
+            enclosedTypes.add(atm);
+            if (e.getKind() == ElementKind.METHOD) {
```
smillst: What about enclosed class or interfaces?

checker/tests/index/ArrayAssignmentSameLen.java
```diff
@@ -10,8 +10,10 @@
         i_index = index;
     }
 
+    @SuppressWarnings("local.variable.unsafe.dependent.annotation")
```
smillst: Are all of these warning suppressions still needed?

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -142,6 +305,94 @@ private void visitAccess(ExpressionTree indexTree, ExpressionTree arrTree) {
...
+                    kind = UpperBoundUtil.SideEffectKind.NON_ARRAY_FIELD_REASSIGNMENT;
+                } else {
+                    return;
+                }
+            } else if (typeKind == TypeKind.ARRAY) {
+                kind = SideEffectKind.LOCAL_VAR_REASSIGNMENT;
+            } else {
+                return;
+            }
+            checkAnnotationsInClass(elt, field, varTree, getCurrentPath(), kind);
```
smillst: Why is this called for reassignments of all array types?

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -142,6 +305,94 @@ private void visitAccess(ExpressionTree indexTree, ExpressionTree arrTree) {
...
+            Receiver canonicalTargetName,
+            Tree tree,
+            TreePath path,
+            SideEffectKind sideEffectKind) {
+        List<? extends Element> enclosedElts =
+                ElementUtils.enclosingClass(elt).getEnclosedElements();
+        List<AnnotatedTypeMirror> enclosedTypes = findEnclosedTypes(enclosedElts);
+        for (AnnotatedTypeMirror atm : enclosedTypes) {
+            List<Receiver> rs =
+                    UpperBoundUtil.getDependentReceivers(atm.getAnnotations(), path, atypeFactory);
```
smillst: The path is used to resolve local variables and parameters, `#1`, correctly, so it should be a path to the location of the annotation or to a place when the same scope.  This is causing the follow code:
```

public class Parameter {
    public void method(@IndexFor("#2") int a, String[] array) {
    }

    public void method2(int a, String[] array) {
        array = new String[]{""};  // an error is issued here
    }
}
```

To issue the following error:
```
Parameter.java:11: error: Invalid format string: "[reassignment.not.permitted] Reassigning the variable %s is not permitted, because the annotation %s depends on it. Reassigning %s would invalidate that annotation." args: [@LTLengthOf(value="array", offset={}) int]
        array = new String[]{""};
        ^
1 error
```
But that code shouldn't issue a warning.

Also, note that the error was not formatted properly.

smillst: Here's another example:
```
import org.checkerframework.checker.index.qual.IndexOrLow;

public class Parameter {
    public void method(@IndexFor("#2") int a, String[] array) {
    }

    public void method2(int a, String[] b) {
        b = new String[]{""};
        int[] array = {1};
        array = new int[0]; // error here
    }
}
```

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
kelloggm: I don't think so. `FlowExpressionParseUtil#parse` and its helper methods just throw an exception, and don't issue a checker warning. So catching this is appropriate.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
kelloggm: Because private fields are always permitted in dependent annotations. In order to write a private field in a dependent annotation, we need to be in the class that the field is defined in, so all the code that might side-effect the field will be visited.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
kelloggm: That's a mistake. This code should cover all non-private fields.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
kelloggm: Would those have more than a single annotated type associated with them? Methods have parameter and return types that need to be calculated separately, but classes/interfaces should only have the annotation that's written on the declaration.

Actually, I might need to handle generic bounds here. Is that what you mean?

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
kelloggm: Any array reassignment invalidates any annotations that have it as an argument, so we always have to check if there are any such annotations.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
smillst: The exception needs to be caught, but BaseTypeVisitor also parses the expression and issues a warning `expression.unparsable.type.invalid` warning.  For example:
```

class Unparsable {
   //::error: (expression.unparsable.type.invalid)
   @IndexFor("sdlfj") int x;
}
```

kelloggm: Makes sense. I'll remove the extra warning.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
smillst: I mean methods in an enclosed class or interface.

kelloggm: I was under the impression that `enclosedElts` should already contain those methods.

smillst: Oh and anonymous classes?

smillst: >I was under the impression that enclosedElts should already contain those methods.

From `getEnclosedElements` Javadoc:

>Returns the fields, methods, constructors, and member types that are directly declared in this class or interface.

kelloggm: So do you think I should be using `Elements#getAllMembers` instead? That takes a `TypeElement`; could I pass the result of `ElementUtils#getEnclosingClass` to that method? If not (I suspect not), how would I go about making the conversion?

smillst: A method of an inner class isn't a member of the outer class, so `Elements#getAllMembers` won't work either.  I would recur in the the case of an inner class.

smillst: Also, `ElementUtils#getEnclosingClass` returns a `TypeElement`.

smillst: This should work:
```
            } else if(e.getKind().isClass()) {
                enclosedTypes.addAll(findEnclosedTypes(e.getEnclosedElements()));
            }
```
Plus add a test case.

kelloggm: Alright, will do. Thanks.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
smillst: If it is a public non-final field then it's reassignment can't invalidate any annotation since it can't appear in one.

kelloggm: I think you're correct that I could apply this code less often - it doesn't need to be called for non-final, public fields. I thought your original comment was about the fact that local variables are handled directly above, so I'm sorry if my answer wasn't clear.

I don't think it's really necessary to add a specific case for public, non-final fields, since, as you say, they should never appear.

smillst: Yes, at first I was just confused as to what was being checked.  Your answered clarified what was going on. 

Calling `checkAnnotationsInClass` for non-private fields won't issue any false positives, but calling it is costly.  

I think in general I'm having a hard time matching the spec we wrote in the Google Doc to the implementation.

kelloggm: Maybe we should sit down and talk about this on Monday?

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundUtil.java
```diff
@@ -0,0 +1,166 @@
...
+    /**
+     * Checks if the {@code receiver} may be affected by the side effect.
+     *
+     * @param receiver expression to check if the side effect affects it
+     * @param reassignedVariable field that is reassigned as the side effect, possible null
+     * @param sideEffectKind {@link SideEffectKind}
+     * @return non-null result
+     */
+    public static SideEffectError isSideEffected(
+            Receiver receiver, Receiver reassignedVariable, SideEffectKind sideEffectKind) {
```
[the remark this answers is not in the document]
kelloggm: I agree that this is significantly clearer, but I don't think it contains all the same logic that the original did. In particular, I don't think this correctly handles chains of field accesses and/or method calls (like `a.b.c().d()`, or similar). The original version of this code wrapped parts of this logic in a while loop to handle that case. I think I can adapt this version to do the same, though.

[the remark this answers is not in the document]
smillst: `isUnmodifiableByOtherCode` applies recursively to the receiver/field access expression.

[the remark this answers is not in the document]
mernst: @kelloggm Does a test case for this situation exist?  If not, please add one.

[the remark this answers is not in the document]
kelloggm: I've added an explicit test for this in `tests/index/ReassignmentChain.java`. This new test confirmed that the loop *is* required: removing it results in the test failing. I discovered an analogous problem in the code that detects dependent annotations that aren't permitted.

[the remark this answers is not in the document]
mernst: Great, thanks for obtaining that valuable information.

kelloggm:
@smillst, I think this is ready for another review.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
smillst: This is still a problem.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -2,34 +2,220 @@
...
+     * Find the Receiver associated with the given String, on the current path. Issues a checker
+     * warning if {@code expr} is unparseable.
+     *
+     * @param expr a String containing a Java expression that is currently in scope
+     * @param tree the current tree, for error reporting
+     * @return the Receiver for the Java expression named by {@code expr} in the current scope
+     */
+    private Receiver getReceiverForCheckingDependentTypes(String expr, Tree tree) {
+        try {
+            return atypeFactory.getReceiverFromJavaExpressionString(expr, getCurrentPath());
```
smillst: See https://github.com/typetools/checker-framework/pull/1346#discussion_r143780363.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -141,6 +327,123 @@ private void visitAccess(ExpressionTree indexTree, ExpressionTree arrTree) {
...
+        }
+        List<AnnotatedTypeMirror> enclosedTypes = findEnclosedTypes(enclosedElts);
+        for (AnnotatedTypeMirror atm : enclosedTypes) {
+            List<Receiver> rs =
+                    UpperBoundUtil.getDependentReceivers(atm.getAnnotations(), path, atypeFactory);
+            for (Receiver r : rs) {
+                ReassignmentError result =
+                        UpperBoundUtil.isSideEffected(r, possiblyInvalidatedRec, reassignmentKind);
+                if (result != ReassignmentError.NO_ERROR) {
+                    checker.report(Result.failure(result.errorKey, atm), tree);
```
smillst: The error keys have require different number of arguments, so this causes errors such as:
```
 Invalid format string: "[reassignment.not.permitted] Reassigning the variable %s is not permitted, because the annotation %s depends on it. Reassigning %s would invalidate that annotation." args: [@LTLengthOf(value="array", offset={}) int]
```

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
smillst: Ping

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -141,6 +327,123 @@ private void visitAccess(ExpressionTree indexTree, ExpressionTree arrTree) {
...
+            Tree tree,
+            TreePath path,
+            UpperBoundUtil.ReassignmentKind reassignmentKind) {
+        List<? extends Element> enclosedElts =
+                enclosingElementsCache.containsKey(elt) ? enclosingElementsCache.get(elt) : null;
+        if (enclosedElts == null) {
+            enclosedElts = ElementUtils.enclosingClass(elt).getEnclosedElements();
+            enclosingElementsCache.put(elt, enclosedElts);
+        }
+        List<AnnotatedTypeMirror> enclosedTypes = findEnclosedTypes(enclosedElts);
```
smillst: Here's another comment from the last review that was lost:
https://github.com/typetools/checker-framework/pull/1346#discussion_r143769405

kelloggm:
@smillst I've pushed a draft solution to the problem we've been discussing in 88fd34f9802d0519d03620d3e6b69b38cdc2b0d9. Do you mind looking it over? Thanks!

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -176,6 +197,74 @@ public void addComputedTypeAnnotations(Tree tree, AnnotatedTypeMirror type, bool
...
 
+    @Override
+    public void preProcessClassTree(ClassTree classTree) {
+        super.preProcessClassTree(classTree);
+        Element elt = TreeUtils.elementFromDeclaration(classTree);
+        List<? extends Element> enclosedElts =
+                ElementUtils.enclosingClass(elt).getEnclosedElements();
+        List<FlowExpressions.Receiver> rs = new ArrayList<>();
+        for (Element e : enclosedElts) {
+            if (!ElementUtils.isEffectivelyFinal(e)) {
```
smillst: Why does it matter if the element is final?   The expression in annotation on the element may not be.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -176,6 +197,74 @@ public void addComputedTypeAnnotations(Tree tree, AnnotatedTypeMirror type, bool
...
+                List<AnnotatedTypeMirror> enclosedTypes = findEnclosedTypes(e);
```
smillst: Instead of recursively finding the enclosed types then parsing the expressions, I would do both at the same time.  Also, if you use a TreePathScanner, then you will have access to the TreePath, without have to recompute it via `getPath()`.  (Also, the path is still wrong for fields or methods of enclosed classes, because it is computed from the current element rather than the element on which the annotation appears.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -176,6 +197,74 @@ public void addComputedTypeAnnotations(Tree tree, AnnotatedTypeMirror type, bool
...
+                for (AnnotatedTypeMirror atm : enclosedTypes) {
+                    List<FlowExpressions.Receiver> atmRs =
+                            UpperBoundUtil.getDependentReceivers(
+                                    atm.getAnnotations(), getPath(declarationFromElement(e)), this);
+                    rs.addAll(atmRs);
+                    for (FlowExpressions.Receiver rec : atmRs) {
+                        dependentAnnotationByReceiver.put(rec, atm);
```
smillst: These are the expressions allowed on a declared dependent type:
 * final or effectively final variable
 * local variable
 * private field
 * pure method call all of whose arguments (including the receiver expression) are composed only of expressions in this list
 * access of a public final field whose access expression (sometimes called the receiver) is composed only of expressions in this list

So, only expressions for those kinds need to be kept.  And then only those expression that would cause an error to be issued by `UpperBoundUtil#isSideEffected`.  (I think that is just expressions containing a non-final private field access or a method call.)

Also, I don't see in the code where an error is used if a declared type contains an expression not in that list.  (For annotations on variables, it happens in UBVisitor#visitVariable.)

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -61,19 +67,32 @@
...
 
     public final AnnotationMirror UNKNOWN, BOTTOM, POLY;
 
     private final IndexMethodIdentifier imf;
 
+    /**
+     * A map that relates an element representing a class to all the receivers of elements of the
+     * class that are dependend on by user-written annotations.
+     */
+    private final HashMap<Element, List<FlowExpressions.Receiver>> dependentReceiversByClass;
```
smillst: Caches should be cleared between compilation in a method overriding AnnotatedTypeFactory#setRoot.  Otherwise, these caches will be huge!

kelloggm:
Closing for now because I don't have time right now to work on this.

PR #1350, opened 2017-06-15
Call hasFieldInvariantAnnotation instead of checking annotations.

### LINKED ISSUE #1347: Initialization Checker confused by delegating constructors and generic fields
The following test case:

```java
class Test<T> {
    T t;

    Test(T t) {
        this(t, 0);
    }

    Test(T t, int i) {
        this.t = t;
    }
}
```

produces the following error:

```
error: [initialization.fields.uninitialized] the constructor does not initialize fields: t
    Test(T t) {
    ^
```

If the field is of a specific type rather than `T`, the checker is happy.

smillst:
Fixes #1347

### DIFF
--- a/checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
+++ b/checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
@@ -170,7 +170,22 @@ public abstract class InitializationAnnotatedTypeFactory<
      * @param field field that might have invariant annotation
      * @return whether or not field has the invariant annotation
      */
-    protected abstract boolean hasFieldInvariantAnnotation(VariableTree field);
+    protected boolean hasFieldInvariantAnnotation(VariableTree field) {
+        AnnotatedTypeMirror type = getAnnotatedType(field);
+        return hasFieldInvariantAnnotation(type);
+    }
+
+    /**
+     * Returns whether or not {@code type} has the invariant annotation.
+     *
+     * <p>If the {@code type} is a type variable, this method returns true if any possible
+     * instantiation of the type parameter could have the invariant annotation. See {@link
+     * NullnessAnnotatedTypeFactory#hasFieldInvariantAnnotation(VariableTree)} for an example.
+     *
+     * @param type of field that might have invariant annotation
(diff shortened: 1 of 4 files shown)


checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -170,7 +170,22 @@ protected boolean isInitializationAnnotation(AnnotationMirror anno) {
      * @param field field that might have invariant annotation
      * @return whether or not field has the invariant annotation
      */
-    protected abstract boolean hasFieldInvariantAnnotation(VariableTree field);
+    protected boolean hasFieldInvariantAnnotation(VariableTree field) {
```
wmdietl: I would suggest making this method `final`. So far this signature was the extension point and subclasses could implement this method. Now these implementation would stop working, because they should override the newly introduced abstract method below.
By making the method `final` we break such subclasses, which seems nicer than silently failing.
Do you agree?

checker/tests/nullness/init/Issue1347.java
```diff
@@ -0,0 +1,23 @@
+// Issue1347 case for Issue 1347.
```
wmdietl: Adapt header.

checker/tests/nullness/init/Issue1347.java
```diff
@@ -0,0 +1,23 @@
...
+    Issue1347(T t) {
+        this(t, 0);
+    }
+
+    Issue1347(T t, int i) {
+        this.t = t;
+        this.t2 = t;
+        this.o = new Object();
+    }
+    //:: error: (initialization.fields.uninitialized)
```
wmdietl: Nit: add empty line before this.

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
wmdietl: At the moment it isn't clear how to choose between these two methods and which one to override.
Adding final will clarify which one to override.
Maybe adding a comment to this method can clarify that it's just a convenience method for the other one.

PR #1352, opened 2017-06-16
Apply @ArrayLen and @ArrayLenRange to strings

panacekcz:
This is a first part of the Bounded-size strings GSoC project. I submit this PR to start a review process after discussion with @kelloggm.

Summary of changes:
- Meaning of `@ArrayLen` and `@ArrayLenRange` is extended to string values of the specifed lengths.
- `@StringVal` is a subtype of `@ArrayLen` of the corresponding lengths.
- `@StringVal` with too many values reduces to `@ArrayLen` or `@ArrayLenRange`.
- For the result of `String.length()`, `@IntVal` or `@IntRange` is inferred from the string annotations on the receiver.
- The result of `String.length()` is used to refine the length annotation of the receiver.
- `@StringVal`, `@ArrayLen` and `@ArrayLenRange` are added to return types of methods in `statically-executable.astub`.
- Value checker subtyping diagram is updated with a new part showing subtyping of `@StringVal`, `@ArrayLen` and `@ArrayLenRange`.

I ran the modified value checker on daikon and plume-lib.
Running the checker produces errors such as:
```
src/plume/RegexUtil.java:184: warning: [type.invalid] [@ArrayLen(1), @UnknownVal] may not be applied to the type "@ArrayLen(1) @UnknownVal String"
    return isRegex(Character.toString(c));
```
This is caused by `@ArrayLen` annotations added to `statically-executable.astub`. More information can be found on the checker-framework-gsoc mailing list. The problem should be solved before merging.


kelloggm:
@panacekcz is there a small test case that fails showing the error you described above? I have some theories about what may be causing it that I'd like to try out.

kelloggm:
One possible theory is that `isSubtype`, GLB, or LUB are implemented incorrectly - I've had problems with that in the past. If an `@UnknownVal` is, for instance, LUB'ed with a `@StringVal` and the result is wrong, the framework can automatically add more `@UnknownVal` annotations. Without a small test case I'm not able to check, but that's a good place to look (I often make errors when implementing those functions, and they are used a lot internally by the framework).

panacekcz:
A test case is:

```java
public class CharacterToString {
	void m(){
		String s = Character.toString('a');
	}
}
```

Run the checker as: `checker/bin/javac -processor org.checkerframework.common.value.ValueChecker CharacterToString.java`

kelloggm:
So I've spent some time investigating this and I can't pin down the root cause. I checked over `isSubtype` and LUB, and I can't find anything wrong. @smillst, do you have any ideas where the extra annotation might be coming from? I was expecting that the problem would be that the `ArrayLen` annotation would fail to be a subtype of `UnknownVal`, which is the problem we've had before, but I checked and that's definitely not the case.

smillst:
I can't reproduce the problem.  I committed the test case.  If Travis fails, then I can use the Docker image that Travis uses to reproduce the problem locally and try to fix it.

smillst:
I know what is causing this problem.  It's not related to this pull request. 
```
src/plume/RegexUtil.java:184: warning: [type.invalid] [@ArrayLen(1), @UnknownVal] may not be applied to the type "@ArrayLen(1) @UnknownVal String"
    return isRegex(Character.toString(c));
```

docs/manual/constant-value-checker.tex
```diff
@@ -57,10 +57,11 @@ \subsection{Type Annotations\label{constant-value-checker-type-annotations}}
 could be more than 10 different values, the Constant Value
 Checker gives up and its type becomes
 \refqualclass{common/value/qual}{IntRange} for integral types,
-\refqualclass{common/value/qual}{ArrayLenRange} for array types, and
+\refqualclass{common/value/qual}{ArrayLenRange} for array types,
```
kelloggm: It would be helpful to explain here how `StringVal` is actually converted into `ArrayLen` or `ArrayLenRange` when there are too many values.

docs/manual/figures/value-subtyping.svg
```diff
@@ -1,791 +1,183 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
kelloggm: It looks like the text on the top row (the one with all the annotations next to each other) is a little too compressed on `ArrayLen` and `ArrayLenRange`. Are the `@` symbols actually inside the box entirely? It would be great if we could make the boxes just a little bigger.

framework/src/org/checkerframework/common/basetype/flow.astub
```diff
@@ -125,6 +125,7 @@ public class Object {
 public final class String {
     @Pure public String intern();
     @Pure public int lastIndexOf(int ch);
+    @Pure int length();
```
kelloggm: What's the rationale for this change?

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -412,14 +430,75 @@ public ValueQualifierHierarchy(MultiGraphQualifierHierarchy.MultiGraphFactory fa
             super(factory);
         }
 
+        /**
```
kelloggm: Can you document why a more precise GLB is needed for StringVal but not for other types?

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -686,6 +787,37 @@ public boolean isSubtype(AnnotationMirror subAnno, AnnotationMirror superAnno) {
...
+            } else if (AnnotationUtils.areSameByClass(superAnno, ArrayLen.class)
+                    && AnnotationUtils.areSameByClass(subAnno, StringVal.class)) {
+                // StringVal is a subtype of ArrayLen, if all the strings have one of the correct lengths
+                List<String> subValues =
+                        AnnotationUtils.getElementValueArray(subAnno, "value", String.class, true);
+                List<Integer> superValues =
+                        AnnotationUtils.getElementValueArray(
+                                superAnno, "value", Integer.class, true);
+                for (String value : subValues) {
+                    if (!superValues.contains(value.length())) return false;
```
kelloggm: We normally use the `{` format even for single-line `if` statements. Can you switch this to be 
```
 if (!superValues.contains(value.length())) {
     return false;
}```

Comment (kelloggm) [inline framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java]: The rationale for this is that we might want to add more logic later, and a common mistake is to forget to add the `{`, so it's easier to just always include them.

File: framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java [hunk 6]
```diff
@@ -686,6 +787,37 @@ public boolean isSubtype(AnnotationMirror subAnno, AnnotationMirror superAnno) {
...
+                }
+                return true;
+            } else if (AnnotationUtils.areSameByClass(superAnno, ArrayLenRange.class)
+                    && AnnotationUtils.areSameByClass(subAnno, StringVal.class)) {
+                // StringVal is a subtype of ArrayLenRange, if all the strings have a length from the range
+                List<String> subValues =
+                        AnnotationUtils.getElementValueArray(subAnno, "value", String.class, true);
+                Range superRange = getRange(superAnno);
+                for (String value : subValues) {
+                    if (!superRange.contains(value.length())) return false;
```
kelloggm: See above comment about `{`

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1527,12 +1697,38 @@ public AnnotationMirror convertArrayLenToArrayLenRange(AnnotationMirror arrayLen
...
+        List<Integer> lengths = ValueCheckerUtils.getLengthsForStringValues(values);
+        return createArrayLenRangeAnnotation(Collections.min(lengths), Collections.max(lengths));
+    }
+
+    /** Converts an {@code @StringVal} annotation to an {@code @ArrayLen} annotation. */
+    public AnnotationMirror convertStringValToArrayLen(AnnotationMirror stringValAnno) {
+        List<String> values = getStringValues(stringValAnno);
+        return createArrayLenAnnotation(ValueCheckerUtils.getLengthsForStringValues(values));
+    }
+
```
kelloggm: Why the distinction between `convertStringValToArrayLen` and `convertStringValToArrayLenRange`? A better design might be to have one method that accepts a `StringVal` and returns either an `ArrayLen` or an `ArrayLenRange` as appropriate. Then you don't have to do a check everytime these are called (and future programmers won't either).

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1527,12 +1697,38 @@ public AnnotationMirror convertArrayLenToArrayLenRange(AnnotationMirror arrayLen
...
     /** Converts an {@code @IntVal} annotation to an {@code @IntRange} annotation. */
     public AnnotationMirror convertIntValToIntRange(AnnotationMirror intValAnno) {
         List<Long> intValues = getIntValues(intValAnno);
         return createIntRangeAnnotation(Collections.min(intValues), Collections.max(intValues));
     }
 
+    /** Returns a {@code Range} bounded by the array lengths specified in the given annotation. */
+    public static Range getArrayLenRange(AnnotationMirror arrayLenRangeAnno) {
```
kelloggm: Why create this method instead of calling `getRange`?

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -70,6 +72,98 @@ public ValueTransfer(CFAbstractAnalysis<CFValue, CFStore, CFTransfer> analysis)
...
+        CFValue value = p.getValueOfSubNode(subNode);
+
+        AnnotationMirror arrayLenRangeAnno =
+                AnnotationUtils.getAnnotationByClass(value.getAnnotations(), ArrayLenRange.class);
+
+        if (arrayLenRangeAnno != null)
+            return ValueAnnotatedTypeFactory.getArrayLenRange(arrayLenRangeAnno);
+
+        // @StringVal, @UnknownVal, @BottomVal
+        AnnotationMirror topAnno =
```
kelloggm: For `UnknownVal` and `BottomVal`, you're not using the result of the call to `getAnnotationByClass` so you can just use `AnnotationUtils#hasAnnotation` instead

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -70,6 +72,98 @@ public ValueTransfer(CFAbstractAnalysis<CFValue, CFStore, CFTransfer> analysis)
...
+    /**
+     * Returns a list of possible lengths for {@code subNode}, as casted to a String. Returns null
+     * if {@code subNode}'s type is top/unknown. Returns an empty list if {@code subNode}'s type is
+     * bottom.
+     *
+     * @param stringValues a list of values for subNode, or null
+     */
+    private List<Integer> getStringLengths(
+            Node subNode, TransferInput<CFValue, CFStore> p, List<String> stringValues) {
+        if (stringValues != null) {
```
kelloggm: For both this method and `getStringLengthRange` you have this pattern where something is passed in, and if it's non-null you make a very different decision. I think this would be more readable if you made this decision at the call sites, and remove the `stringValues` parameter.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -70,6 +72,98 @@ public ValueTransfer(CFAbstractAnalysis<CFValue, CFStore, CFTransfer> analysis)
...
+        CFValue value = p.getValueOfSubNode(subNode);
+
+        AnnotationMirror arrayLenAnno =
+                AnnotationUtils.getAnnotationByClass(value.getAnnotations(), ArrayLen.class);
+
+        if (arrayLenAnno != null)
+            return AnnotationUtils.getElementValueArray(arrayLenAnno, "value", Integer.class, true);
+
+        // @StringVal, @UnknownVal, @BottomVal
+        AnnotationMirror topAnno =
```
kelloggm: Again for these you can use `AnnotationUtils#hasAnnotation`

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -70,6 +72,98 @@ public ValueTransfer(CFAbstractAnalysis<CFValue, CFStore, CFTransfer> analysis)
...
+            return null;
+        }
+        AnnotationMirror bottomAnno =
+                AnnotationUtils.getAnnotationByClass(value.getAnnotations(), BottomVal.class);
+        if (bottomAnno != null) {
+            return new ArrayList<Integer>();
+        }
+
+        // @IntRange
+        if (subNode instanceof StringConversionNode) {
```
kelloggm: Can you add a comment explaining this? I think I know what's going on here but the comment `// @IntRange` doesn't explain why you make this split.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -70,6 +72,98 @@ public ValueTransfer(CFAbstractAnalysis<CFValue, CFStore, CFTransfer> analysis)
...
+            if (valueRange.contains(0)) {
+                lowerLength = 1;
+            }
+
+            int upperLength = Math.max(fromLength, toLength);
+
+            Range lengthRange = new Range(lowerLength, upperLength);
+
+            return ValueCheckerUtils.getValuesFromRange(lengthRange, Integer.class);
+
```
kelloggm: Why no `IntVal` case?

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -292,8 +436,7 @@ private void refineArrayAtLengthAccess(FieldAccessNode arrayLengthNode, CFStore
         if (AnnotationUtils.areSameByClass(lengthAnno, BottomVal.class)) {
             // If the array's length is bottom, then this is dead code, so the array's type
```
kelloggm: Should probably change this comment not to mention arrays

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -278,8 +380,50 @@ private void refineArrayAtLengthAccess(FieldAccessNode arrayLengthNode, CFStore
...
+     * Transform @IntVal or @IntRange annotations of a array or string length into an @ArrayLen
+     * or @ArrayLenRange annotation for the array or string.
+     */
+    private void refineAtLengthAccess(Node lengthNode, Node receiverNode, CFStore store) {
         Receiver arrayLenRec =
-                FlowExpressions.internalReprOf(analysis.getTypeFactory(), arrayLengthNode);
+                FlowExpressions.internalReprOf(analysis.getTypeFactory(), lengthNode);
+
+        // If the expression is not representable (for example if String.length() for some reason is not marked @Pure, then do not refine.
+        if (arrayLenRec instanceof FlowExpressions.Unknown) {
```
kelloggm: Probably should change the name of `arrayLenRec` so that it no longer mentions arrays. You changed some of these names but not all of them.

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -350,33 +483,86 @@ private void refineArrayAtLengthAccess(FieldAccessNode arrayLengthNode, CFStore
...
+
+    /**
+     * Calculates a range of possible lengths of a result of string concatenation of strings with
+     * known ranges of lengths.
+     */
+    private Range calculateLengthRangeAddition(Range leftLengths, Range rightLengths) {
+        return leftLengths.plus(rightLengths).intersect(Range.INT_EVERYTHING);
+    }
+
+    private AnnotationMirror createAnnotationForStringConcatenation(
```
kelloggm: Comments describing the different sections of this method (i.e. first check if both are StringVals, then try lengths, etc.) would make it easier to read.

framework/src/org/checkerframework/common/value/qual/ArrayLen.java
```diff
@@ -7,9 +7,9 @@
 import org.checkerframework.framework.qual.SubtypeOf;
```
kelloggm: Because `ArrayLen` now applies to Strings, we should probably rename it. Perhaps to `@Len`? Or `@Length`?

framework/src/org/checkerframework/common/value/qual/ArrayLenRange.java
```diff
@@ -7,9 +7,9 @@
 import org.checkerframework.framework.qual.SubtypeOf;
```
kelloggm: See comment on `ArrayLen` about renaming. This name should just be change to *Range, where * is whatever we rename `ArrayLen` to.

framework/tests/value/Basics.java
```diff
@@ -277,11 +277,13 @@ void tooManyValuesString() {
...
 
-        a = c; // This should succeed if a is treated as @UnknownVal
+        //:: error: (assignment.type.incompatible)
+        a = c; // This should not succeed if a is treated as @ArrayLen(1)
 
         //:: warning: (too.many.values.given)
         @StringVal({"a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l"}) String d = "h";
 
-        d = "b" + d; // This should succeed since d is @UnknownVal
+        //:: error: (assignment.type.incompatible)
```
kelloggm: Can you add a test that checks that `a` and `d` in this test are actually `@ArrayLen(1)`, in addition to what's already present?

framework/tests/value/CharacterToString.java
```diff
@@ -0,0 +1,5 @@
+public class CharacterToString {
```
kelloggm: This test should probably include a comment about the issue that it exposes.

framework/src/org/checkerframework/common/value/ValueTransfer.java
panacekcz: Cannot find that method, do you mean `containsSameByClass`?

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
panacekcz: Currently the methods are used to convert to the required type in `leastUpperBound`. The incoming `StringVal` should already have a limited number of values, so there would be no need to chose `ArrayLenRange` over `ArrayLen`. It would add an additional conversion step (from `StringVal` to `ArrayLen` and then to `ArrayLenRange`) when doing a lub of  `StringVal` and `ArrayLenRange`.
I don't do a check everytime these are called.

kelloggm: I agree that the way they are used right now is fine, but I'm concerned with maintenance. I expect that we will want a unified way to convert an arbitrary `StringVal` to an `ArrayLen` or `ArrayLenRange`, and, more importantly, I don't think that future programmers should be presented with two public methods that have identical arguments, very similar names, but important differences in their preconditions. Maybe a compromise is to make these two methods private, and add a public method that takes a `StringVal` and dispatches to these two methods as appropriate?

panacekcz: But there are no differences in the preconditions.
Making the methods private is fine, they were public only for consistency with similar methods such as `convertIntValToIntRange`. I don't know why they are public, eclipse didn't find any references from outside.

kelloggm: Is there really no difference in the precondition? One expects a `StringVal` that can be converted to an `ArrayLen` without having too many values, and the other does not. I view that difference as meaningful.

Whether `convertIntValToIntRange` should be private is a different matter (though it probably should). But I suppose I can accept these methods as-is on that basis, though I still think it will be a maintenance headache down the road.

panacekcz: If `convertStringValToArrayLen` receives too many values, it returns `ArrayLenRange`, despite its name. I will note that in the comment.

panacekcz:
The travis build time-out was caused by infinite (or almost-infinite) loop in the checker because widening was not implemented on `@ArrayLenRange`. It happened on the method `java.util.logging.Logger.entering(String sourceClass, String sourceMethod, Object params[])`. I got confused by the checker output, which stopped in a different file. Perhaps some output buffering is involved.

A simple widening of `@ArrayLenRange` should resolve this.

panacekcz:
I addressed the review comments. @kelloggm, could you review the PR again, please?
The widening I implemented for `ArrayLenRange` is maybe too simple, because it forgets the lower bound even if only the upper bound is growing. Do you think I should implement a more precise widening (expanding only the bounds which actually grow)?

docs/manual/figures/value-subtyping.svg
kelloggm: This comment doesn't appear to have been addressed. The `@` signs are still hard to read because the boxes aren't large enough - can you make the boxes on `@ArrayLen` and `@ArrayLenRange` a little bigger so that the text fits? Thanks.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
kelloggm: I still don't see a comment either here or on `greatestLowerBound` where this is called that explains why StringVal needs a more precise GLB than the other Value Checker types. Please include one.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
kelloggm: It also seems like this hasn't been addressed. @panacekcz you said that this was ready for a re-review, but it appears that many of my comments haven't been addressed, even when they are very simple to address. Did you forget to push some changes, perhaps?

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
kelloggm: Again, not fixed.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1143,6 +1284,40 @@ private boolean handledByValueChecker(AnnotatedTypeMirror type) {
         }
     }
 
+    /** Determines whether a method is the {@code String.length()} method. */
+    boolean isStringLengthMethod(ExecutableElement method) {
+        return method == lengthMethod;
```
kelloggm: Elements are not interned. See [the javadoc here](https://docs.oracle.com/javase/8/docs/api/javax/lang/model/element/Element.html). You should always compare them with `equals` instead of `==`

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1143,6 +1284,40 @@ private boolean handledByValueChecker(AnnotatedTypeMirror type) {
...
+    }
+
+    /**
+     * Returns a constant value annotation for a length of an array or string type with a constant
+     * value annotation
+     */
+    AnnotationMirror createArrayLengthResultAnnotation(AnnotatedTypeMirror receiverType) {
+        AnnotationMirror arrayAnno = receiverType.getAnnotation(ArrayLen.class);
+        if (arrayAnno != null) {
```
kelloggm: I don't think this code was originally yours, but since it's under review I'm going to suggest that this pattern of calling `getAnnotation` and then checking for `null` is an anti-pattern, since it's rather obfuscated. Do you mind changing this method to use `AnnotatedTypeMirror#getAnnotationInHierarchy` to get whatever annotation it is, and then guard each `if` block with a call to `AnnotationUtils.areSameByClass` with the appropriate annotation? I think that would be much cleaner code.

docs/manual/constant-value-checker.tex
smillst: (Addressed below.)

docs/manual/figures/value-subtyping.svg
smillst: @panacekcz Has this been addressed?

framework/tests/value/CharacterToString.java
smillst: (Addressed above.)

framework/tests/value/Basics.java
smillst: (Addressed below.)

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
smillst: @panacekcz Has this been addressed?

docs/manual/figures/value-subtyping.svg
panacekcz: Addressed in d32a700c83489360e6dbeda163d0c297abffb0f5, made the boxes 10px wider.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
panacekcz: Adressed in be1fb6326bd54061c4c84b89195c1085d17bc9ae.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -505,6 +589,15 @@ public AnnotationMirror leastUpperBound(AnnotationMirror a1, AnnotationMirror a2
                     newValues.addAll(a1Values);
                     newValues.addAll(a2Values);
                     return createArrayLenAnnotation(newValues);
+                } else if (AnnotationUtils.areSameByClass(a1, StringVal.class)) {
+                    List<String> a1Values =
+                            AnnotationUtils.getElementValueArray(a1, "value", String.class, true);
```
panacekcz: @smillst, there are also similar calls not written by me. Should I change them too or just the new ones?

smillst: Please change the all.  Thanks!

panacekcz:
@smillst, I have adressed the review comments.

smillst:
Thanks!  Everything looks good now.  I noticed that we haven't fixed #1356 yet, so I added a workaround for the Value Checker.  (See  ac43117.)  Once the tests pass again, I'll merge.

PR #1360, opened 2017-06-23
Use TypeAnnotationUtils.unannotatedType instead of Type.unannotatedType;

wmdietlGC:
the latter will be removed in Java 9.
Implement TypeAnnotationUtils without reflection. This makes the code easier to read, but unfortunately didn't speed up execution.


wmdietlGC:
This works as intended. @smillst can you have a look?

Issue #1098, opened 2017-03-03
Nullness: upper bounds for j.u.Optional and j.l.Class


wmdietlGC:
Take this simple example:

```
import java.util.Optional;

class OptionalDemo {
  <T> void opt(Optional<T> p1, T p2) {}
  <T> void cls(Class<T> p1, T p2) {}

  void use() {
    opt(Optional.empty(), null);
    cls(this.getClass(), null);
  }
}
```

The declaration of method `opt` doesn't pass the Nullness Checker, because the upper bound of T is incompatible with the non-null bound of Optional.
For j.l.Class we used nullable as upper bound, conveniently allowing the `cls` declaration.
The invocation of method `cls` is however forbidden, which doesn't seem desirable.

For both Optional and Class one can make the argument that the nullity of the type argument doesn't really matter and that any instantiation is fine.

Attached a possible change to the signature of Optional that works conveniently in more code.

[Optional-Nullable.zip](https://github.com/typetools/checker-framework/files/815407/Optional-Nullable.zip)

mernst:
I agree that the inconsistency between `Optional` and `Class` is undesirable.

I would be OK with permitting unbounded type variables as in `Optional<T>`.  However, I would prefer that programmers do not write `Optional<@Nullable String>` which is misleading to readers.  Is there any way to achieve that?

We'll eventually need more annotations than those in your diff.
Consider the `Optional.ifPresent` routine:
```java
  public void ifPresent(Consumer<? super T> consumer) {
```

It's currently commented out because the `Consumer` class is defined in JDK 8.
However, we would want to change `T` to `@NonNull T` so that the signatures of these routines are the same:
```java
Optional<@NonNull String>.ifPresent()
Optional<@Nullable String>.ifPresent()
```

mernst:
Making the change permits the examples in this issue to type-check, but it causes other previously-passing code to fail to type-check under the nullness type system.  Here is an example:

```
    void m(Date d) {
        Class<? extends Date> c = d.getClass();
    }
```

```
error: [assignment.type.incompatible] incompatible types in assignment.
        Class<? extends Date> c = d.getClass();
                                            ^
  found   : @Initialized @NonNull Class<? extends @Initialized @Nullable Date>
  required: @UnknownInitialization @Nullable Class<? extends @Initialized @NonNull Date>
```

Do we have evidence about which type of code is more common, and therefore more desirable to support while forcing user annotations for the other?

wmdietlGC:
[...]

PR #1361, opened 2017-06-24
Optional checker

mernst:
Tests are failing because the annotated JDK used by Travis has not been updated.
I think this is ready for review, though.


checker/build.xml
```diff
@@ -1476,9 +1483,13 @@ So, use our own archived version.
     </target>
 
     <target name="download-jdk" description="Downloads jdk7.jar and jdk8.jar from checkerframework.org/dev-jdk to checker/jdk and checker/dist">
+<!--
         <get src="https://checkerframework.org/dev-jdk/jdk7.jar" dest="jdk/jdk7.jar"/>
         <get src="https://checkerframework.org/dev-jdk/jdk8.jar" dest="jdk/jdk8.jar"/>
+-->
 
+        <get src="https://github.com/smillst/checker-framework-jdk/releases/download/mernst%2Foptional-checker/jdk7.jar" dest="jdk/jdk7.jar"/>
```
smillst: This change must be reverted before merging.

mernst:
@smillst The tests passed.

wmdietlGC:
We discussed this in our meeting today. I'll summarize what I had proposed in the meeting and then suggest doing something else.

My proposal was to make the Optional Checker completely independent of the Nullness Checker and add support for `java.util.Optional` to the Nullness Checker itself.
To do this, we would use the type argument of Optional to designate whether there is a value attached.
A `Optional<@Nullable C>` means that we don't know whether the Optional contains a value.
A `Optional<@NonNull C>` means that there definitely is a C value attached.
As Optional is immutable, we can make it covariant in the type argument; therefore, it's safe to make `Optional<@NonNull C>` a subtype of `Optional<@Nullable C>`, which will allow passing around optionals easily.
We could restrict the receiver of `Optional.get` to only allow non-null arguments: `T get(Optional<@NonNull T> this)`.
The only feature that we don't have yet is refining the type argument from an `isPresent` check: `isPresent` on an `Optional<@Nullable C>` will change the type in the then branch to `Optional<@NonNull C>` and therefore allow invoking `get`.

All of this would probably work nicely, with not too much additional effort.
However, there is one big drawback with this suggestion: usually, when we write `Optional<C>` we want that this means that we don't know whether a value is attached. That would mean we want to interpret `Optional<C>` as `Optional<@Nullable C>` which is not our usual default. Changing the default for the type argument of Optional may be possible, but it would definitely be inconsistent.

So instead, how about the following changes to this PR:
- simplify the Optional Checker to be independent of the Nullness Checker and only enforce isPresent/get usage. This will be a simple, hopefully fast checker that will be easy to use.
- Make the Optional Checker a subchecker of the Nullness Checker, to get the additional guarantee about correct isPresent/get usage. This might need some tweaking, in case the two need a tighter integration, in particular if the Optional Checker needs to access nullness information. I would then suggest building a combined "Nullness Optional Checker" that is more tightly integrated. 

As argued in #1098 I think we should make the upper bound of Optional `@Nullable Object`, to flexibly allow arbitrary instantiations. Annotations on individual methods will express whether parameters/returns are non-null.

mernst:
I'm separating our discussion into three distinct, actionable proposals.
I agree with two and am not convinced about the third.


1. `Optional` should permit instantiating as both 
  `Optional<@NonNull String>`
  `Optional<@Nullable String>`
and the JDK annotations for `Optional` should be rewritten to change the upper bound on the type argument no be `@Nullable`, with `@NonNull` made explicit in the Javadoc where it is currently implicit.

Do the same for Class.

The motivation is to make it easier to instantiate `Optional<T>` where T is a type variable.
Google found it necessary to use `@Nullable` as the upper bound on `Optional` in their stub files, so that suggests we should do it for all users (even though we haven't noticed the problem in other case studies, with `Class` or `Optional`).

A use case is method type inference.  We don't want to force users to always write `Optional<@NonNull T>`, which is too heavy an annotation burden.

I agree with this proposal and will implement it.


2. Create two versions of the Optional Checker.
 * One is a lightweight checker that is independent of the Nullness Checker.  It only checks properties #2-7, not anything about nullness.
   (I need to figure out all the places that the Optional Checker is coupled to the Nullness Checker.  One example is the precise type-checking for the return type of ofNullable() is an example, but this probably it isn't so important a feature:  a user can use of() instead of ofNullable() if the user knows the arguments are non-null.)
 * The other is the Nullness Checker, which will do checking of Optional just as it currently does for Initialization and KeyFor.  It makes sense to call this the Nullness Checker rather than the Optional Checker or the heavyweight Optional Checker, in terms of what most of the implementation is and in terms of getting people to use it (since the heavyweight Optional Checker wouldn't be as attractive as the Nullness Checker to a potential user).
   The implementation of the Nullness Checker will be the Optional Checker, with the Nullness Checker as a subchecker; this may momentarily surprise someone who reads the implementation, but a // comment will clarify the issue.

I agree with this proposal and will implement it.


3. Change the syntax for expressing "present `Optional`" and "possibly-present `Optional`", in at least the heavyweight Nullness Checker.
  `Optional<@NonNull X>`   replaces  `@Present Optional<X>`
  `Optional<@Nullable X>`  replaces  `@MaybePresent Optional<X>`

Points in favor:
 * Fewer annotations (don't introduce two new annotations `@Present` and `@MaybePresent`).
 * If users have to write few annotations (as we hope), it is easy to transition from one syntax to the other when transitioning between the lightweight and heavyweight Optional checkers.

Points opposed:
 * Inconistent syntax in the two versions of the Optional Checker.
 * A user who sees
     `Optional<@NonNull X>`
     `Optional<@Nullable X>`
   won't mentally associate those with present and not-necessarily present `Optional`s.  By contrast, 
     `@Present Optional<X>`
     `@MaybePresent Optional<X>`
   are clear and readable.
   More generally, the new syntax uses the annotations `@NonNull` and `@Nullable` for a different meaning than anywhere else that they are used.  The `Optional` documentation is worded in terms of present and not-present values, and it is clear that these are distinct from nullness.  (The existence of a field that contains null within the `Optional` class is an implementation detail.)
 * Poor defaulting leads to many extraneous annotations.
   The default for a type annotation is `@NonNull`.  Most `Optional`s (and almost every declaration in source code) are `@MaybePresent`, and this would mean that a user would be forced to edit every occurrence of `Optional<String>` in a program to make it `Optional<@Nullable String>`.  By contrast, the `@MaybePresent` syntax does not require these edits, because `@MaybePresent` is the default type annotation.
 * If lightweight and heavyweight Optional Checkers use the same syntax, it is trivial for users to start with the lightweight one and transition to the heavyweight one.
 * Minor: The Checker Framework does not currently support refining type arguments, such as flow-sensitive refinement of a variable from `Optional<@Nullable String>` to `Optional<@NonNull String>` as would be needed in the heavyweight Optional Checker (that is, in the Nullness Checker).  We could do something similar to KeyFor, adding new information to the store.
 * Minor: Different representations mean a bit more work to share code between the lightweight and heavyweight Optional Checkers.

If we retain the `@Present` and `@MaybePresent` syntax, then we need to tell users that `Optional<@NonNull String>` and `Optional<@Nullable String>` mean the same thing but aren't interoperable.


I'm concerned about this proposal, particularly because the syntax won't be obvious to users and because it will lead to the need to write a lot of annotations.

wmdietl:
I think we agree on all points.
We'll see whether the Optional Checker should contain the Nullness Checker or the other way around. It would be a bit easier if the Optional Checker would be a sub-checker of the current Nullness Checker and I would try that way around first.

mernst:
This is blocked on pull request #1383

mernst:
This should be broken into multiple separate issues (and perhaps even closed in favor of those).

smillst:
I think the test failures are because the Optional Checker uses the Nullness Checker when creating its annotated JDK, so nullness annotations are added by the Optional Checker and then other nullness annotations are added when the just the Nullness Checker compiles jdk/nullness/*/Optional.java.

mernst:
Among the three tasks, the first is done and the third is not worth doing.  I propose that this pull request can be merged, and the second opened as an issue.

changelog.txt
```diff
@@ -113,6 +113,8 @@ Closed issues:
 
 Version 2.1.13, 3 July 2017
 
+Adds Optional Checker to verify uses of Java 8's Optional class.
```
wmdietl: This should be moved to the next upcoming release.

checker/src/org/checkerframework/checker/optional/OptionalChecker.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.LinkedHashSet;
+import org.checkerframework.checker.nullness.NullnessChecker;
+import org.checkerframework.common.basetype.BaseTypeChecker;
+
+/**
+ * A type-checker that prevents misuse of the {@link java.util.Optional} class.
+ *
+ * @checker_framework.manual #optional-checker Optional Checker
+ */
+// Uncomment when the Checker Framework assumes Java 8: @RelevantJavaTypes(Optional.class)
```
wmdietl: Can be uncommented.

checker/src/org/checkerframework/checker/optional/OptionalChecker.java
```diff
@@ -0,0 +1,22 @@
...
+public class OptionalChecker extends BaseTypeChecker {
+
+    @Override
+    protected LinkedHashSet<Class<? extends BaseTypeChecker>> getImmediateSubcheckerClasses() {
```
wmdietl: I think the Optional Checker would be much better without the dependency on the Nullness Checker, as described in previous review comments.
This will only have an impact on `Optional.of` and `Optional.ofNullable` calls, which we can leave as future issues.
The Optional Checker would then be a fast checker for that specific property. Handling the interaction with the Nullness Checker can then be done in another iteration.
The impact to this PR should be small: remove this method, the uses of nullnessTypeFactory, and move a few tests to a new issue.

checker/src/org/checkerframework/checker/optional/OptionalVisitor.java
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    /**
+     * Part of rule #3.
+     *
+     * <p>Pattern match for: {@code VAR.isPresent() ? VAR.get().METHOD() : VALUE}
+     *
+     * <p>Prefer: {@code VAR.map(METHOD).orElse(VALUE);}
+     */
+    public void handleTernaryIsPresentGet(ConditionalExpressionTree node) {
```
wmdietl: This logic for `isPresent` followed by `get` would more generally be handled by a transfer function, instead of coding the pattern on Trees for ternaries and ifs separately.
As the current approach works fine for the tests, I'm fine with merging it. But this would be a good future generalization.

checker/src/org/checkerframework/checker/optional/OptionalVisitor.java
```diff
@@ -0,0 +1,340 @@
...
+            return;
+        }
+        ExpressionTree trueReceiver = TreeUtils.getReceiverTree(trueExpr);
+        if (!isCallToGet(trueReceiver)) {
+            return;
+        }
+        ExpressionTree getReceiver = TreeUtils.getReceiverTree(trueReceiver);
+
+        // What is a better way to do this than string comparison?
+        if (receiver.toString().equals(getReceiver.toString())) {
```
wmdietl: This is OK for now. When going to transfer functions this would be solved using the correct Store entries. (Comparing the strings of the expressions also doesn't take care of side-effecting methods as receivers...)

checker/src/org/checkerframework/checker/optional/OptionalVisitor.java
```diff
@@ -0,0 +1,340 @@
...
+                            // TODO: add this to the error message.
+                            //ElementUtils.getQualifiedClassName(ele);
+                            ele.getSimpleName(),
+                            falseExpr),
+                    node);
+        }
+    }
+
+    /** Return true if the two trees represent the same expression. */
+    private boolean sameExpression(ExpressionTree tree1, ExpressionTree tree2) {
```
wmdietl: This method is currently not called. Should it be used above in `if (receiver.toString().equals(getReceiver.toString()))`?

javacutil/src/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -355,6 +355,29 @@ public static ExpressionTree skipParens(final ExpressionTree tree) {
...
     }
 
+    /**
+     * If the given tree is a block tree with a single element, return the enclosed non-block
+     * statement. Otherwise, return the same tree.
+     *
+     * @param tree a statement tree
+     * @return the single enclosed statement, if it exists; otherwise, the same tree
+     */
+    public static StatementTree skipBlocks(final StatementTree tree) {
```
wmdietl: This method is only needed because the OptionalVisitor operates over Trees instead of over the CFG. I would rather move this to a private helper method in OptionalVisitor. A checker shouldn't write code like this.

checker/src/org/checkerframework/checker/optional/OptionalVisitor.java
mernst: OK, we'll leave it for now and improve it in the future.  The comment was there so that during code review someone would tell me the right way to do it, because I knew this was not it. :-)

javacutil/src/org/checkerframework/javacutil/TreeUtils.java
mernst: Does the same comment apply to skipParens and possibly other methods in TreeUtils?

wmdietl:
As discussed, I removed the dependency on the Nullness Checker.
I needed to disable three tests; I added TODO comments for them. If you prefer, maybe move them to a `@skip-test`ed file.
The stream API call didn't work for me, but I'm not sure this was caused by this change.
I didn't get to look to see whether the manual needs an update. I might get a chance to do this later tonight.

checker/src/org/checkerframework/checker/optional/OptionalChecker.java
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * A type-checker that prevents misuse of the {@link java.util.Optional} class.
+ *
+ * @checker_framework.manual #optional-checker Optional Checker
+ */
+/* TODO:
+ * For a call to ofNullable, if the argument has type @NonNull, make the return type have type @Present.
+ * Make Optional Checker a subchecker of the Nullness Checker, or vice versa.
+ */
+// Uncomment when the Checker Framework assumes Java 8: @RelevantJavaTypes(Optional.class)
```
wmdietl: This can be uncommented as we're on Java 8 now.

smillst:
(FYI, the tests are failing because the Optional Checker was adding nullness annotations to the annotated JDK.  Since Werner removed the Nullness Checker as a sub checker, I'll rebuild jdk8.jar.)

javacutil/src/org/checkerframework/javacutil/TreeUtils.java
wmdietl: Possibly for `skipParens`, yes. There are very few uses of that method and some in InterningVisitor might be better implemented over the CFG.
`skipParens` would still be useful to print shorter error messages and for some Tree-based checks. I don't think we ever need to do that with blocks.

checker/src/org/checkerframework/checker/optional/OptionalVisitor.java
mernst: Done.

smillst:
I noticed this issue #1425. Is an enhancement request for the adding two optional checkers.  It should probably be updated to reflect that one of the Optional Checkers now exists.

mernst:
@wmdietl I have updated the manual, so you don't have to look into that.

mernst:
I have made enough changes that this could probably stand another round of review.

checker/src/org/checkerframework/checker/optional/OptionalUtils.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+import javax.lang.model.element.ExecutableElement;
+import org.checkerframework.framework.type.AnnotatedTypeFactory;
+import org.checkerframework.javacutil.ErrorReporter;
+import org.checkerframework.javacutil.TreeUtils;
+
+/** Utility methods used by OptionalVisitor, OptionalAnnotatedTypeFactory, etc. */
+public class OptionalUtils {
```
smillst: This class isn't used.  Should it be removed?

checker/tests/optional/JdkCheck.java
```diff
@@ -0,0 +1,73 @@
...
+
+    boolean isPresentTest1(@Present Optional<String> pos) {
+        return pos.isPresent();
+    }
+
+    boolean isPresentTest2(Optional<String> mos) {
+        return mos.isPresent();
+    }
+
+    // No JDK annotations on methods using Java 8 types yet.
```
smillst: I'm rebuilding the JDK again.  Maybe these can be uncommented after that is done?

mernst: Yes, I think they can.  Could you please try that when you get a chance this morning?

smillst:
I think this is ready to be merged once the test pass!

mernst:
@wmdietl If you want to take one more pass over this code you are welcome to do so.  If not, that's OK too.
Thanks, everyone, for helping this pull request get finished.
With the change to not support Java 7, it became much simpler!

wmdietl:
I had a quick look and things look great. It will be good to merge this and make the type system available.

PR #1363, opened 2017-06-26
Fix #1180: Default case in switch statements not handled correctly

### LINKED ISSUE #1180: Default case in switch statements not handled correctly
Consider a switch statement:

```java
switch(x) {
  case 0:
  case 1:
      // do some code...
      break;
   //... more cases
  default:
      // some code
}
```

When the CFG is built, a series of `CaseNode` objects are built for each case - except the default case. This happens in `org.checkerframework.dataflow.cfg.CFGBuilder.CFGTranslationPhaseOne.SwitchBuilder`. In Java, the semantics of the `default` statement should be that it is executed if either: it is reached via fall-through from one of the previous cases, or if none of the previous cases were triggered.

With no `CaseNode` for default, it is impossible to refine a type in the default case based on what came before. See typetools#1172 for some test cases that demonstrate this (and more exposition on the motivation for desiring this change).

bitterfox:
The cause of the problem is not referring the current value of the switch expression.
Due to it, another case node captures the variable of the wrong type and failed to determine the most specific value to update the store.

Consider following switch statement:
```
     void test(@IntVal({1, 2, 3}) int x) {
         switch (x) {
             case 1:
             case 2:
                 @IntVal({1, 2}) int w = x;
                 break;
             default:
                 @IntVal({3}) int q = x;
                 break;
         }
```

Previously of this PR, the CFG will be: (A number of after @ is a label)
```
Load x@1
Literal 1
Case x@1, 1
|-THEN_TO_BOTH(@IntVal({1, 2}) int w = x;)
|-ELSE_TO_BOTH
  Literal 2
  Case x@1, 2
  |-THEN_TO_BOTH(@IntVal({1, 2}) int w = x;)
  |-ELSE_TO_BOTH(@IntVal({3}) int q = x;)
```

`x@1` is resolved as `@IntVal({1, 2, 3})` and stores are updated correctly at first case like:
```
thenStore: {x: @IntVal(1)}
elseStore: {x: @IntVal(2, 3)}
```
In second case, the type of the local variable node is still `@IntVal({1, 2, 3})`(because it still refer `x@1 : @IntVal({1, 2, 3})`). After comparing each value of 1, 2, 3 with 2, ValueTransfer try to update store with current store(`x: @IntVal(2, 3)`)

```
thenStore: {x: @IntVal(2)}
elseStore: {x: @IntVal(1, 3)} // will fail
```
Updating thenStore will succeed, but another will not happened because:

BOTTOMVAL is selected to the candidate for the new type because @IntVal(1, 3) is not a sub type of @IntVal(2, 3) and opposite case is also not.
https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/common/value/ValueTransfer.java#L992
https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java#L413

Updating don't occur because it failed to detect most specific of BOTTOMVAL and @IntVal(2, 3) because they don't have sub type relation for each other
https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/flow/CFAbstractStore.java#L297
https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/flow/CFAbstractValue.java#L269
https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/flow/CFAbstractValue.java#L258

In this fix, I propose to fix take the current value from current store by keeping the current value in store as a value of template variable


dataflow/src/org/checkerframework/dataflow/cfg/CFGBuilder.java
```diff
@@ -2458,6 +2451,21 @@ protected VariableTree getAssertionsEnabledVariable() {
         }
 
         /**
+         * Find nearest owner element(Method or Class) which holds current tree
+         *
+         * @return Nearest owner element of current tree
+         */
+        private Element findOwner() {
```
wmdietl: Don't forget to use the `findOwner` from #867 in the end.

framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
```diff
@@ -736,8 +736,9 @@ protected void addInformationFromPreconditions(
                         } else {
                             thenStore.insertValue(secondInternal, firstValue);
                         }
-                        return new ConditionalTransferResult<>(
-                                res.getResultValue(), thenStore, elseStore);
+                        res =
```
wmdietl: Can you add a comment what this is doing?

framework/tests/value/Switch.java
```diff
@@ -138,4 +151,25 @@ void testStrings(String s) {
...
+                @StringVal("Good") String x1 = a;
+                @StringVal("Good") String x2 = s;
+            case "Bye":
+                @StringVal({"Good", "Bye"}) String y1 = a;
+                @StringVal({"Good", "Bye"}) String y2 = s;
+                break;
+            case "Hello":
+                @StringVal("Hello") String z1 = a;
+                @StringVal("Hello") String z2 = s;
+                break;
```
wmdietl: Can you add a test, for ints, Strings or something else, where a case falls through into the default case?

bitterfox:
I've added a test case and updated dataflow manual.
Please review this again.

bitterfox:
Oops, I've failed to commit a pdf due to following error:
```
% git cmm "Update dataflow manual"
Buildfile: /home/bitter_fox/dev/gsoc/checker-framework/build.xml
Some files have trailing whitespace: dataflow/manual/examples/graphs/CFGSwitch.pdf
```

Do you have any idea to fix this?

dataflow/manual/content.tex
```diff
@@ -875,11 +875,6 @@ \subsection{Formal Definition of the Control-Flow Graph}
 \end{workinprogress}
 
 \begin{workinprogress}
-There is a \code{CaseNode}. How is the default case of a switch
-handled?
-\end{workinprogress}
-
-\begin{workinprogress}
 There is a \code{StringConcatenateAssignmentNode}.
 What about other compound assignments? Are they desugared?
```
wmdietl: Now that you looked at unary assignments and other CFG nodes, can you, in a separate PR, address this `workinprogress` about compound assignments?

dataflow/manual/content.tex
```diff
@@ -1083,6 +1078,23 @@ \subsubsection{Assert statements}
 is no alternate control flow?
 \end{workinprogress}
 
+\subsubsection{Default case and fall through for switch statement}
```
wmdietl: Can you add a `\label` for this section?

dataflow/manual/content.tex
```diff
@@ -1083,6 +1078,23 @@ \subsubsection{Assert statements}
...
+Note that the expression of switch statement must be executed just only
+once at first of switch statement. To refer the value of the expression,
+a fake variable is created and the expression is assigned to a fake
+variable. \code{THEN_TO_BOTH} edge goes to nodes for the case and \code{ELSE_TO_BOTH}
+edge goes to next \code{CaseNode}. When a next is default case, it goes
+to nodes for default case. If a break statement in nodes for a case create
+an edge to next node of the switch statement. When it doesn't exist, it means
+that it's fall-through and the next node of nodes of the case is nodes of
+the next case. For example, the code in \autoref{lst:CFGSwitch} produces
+the control flow graph in \autoref{fig:CFGSwitch}. The face variable named
```
wmdietl: Should the `face` be `fake`? Or something else?
Can you do another pass through the English of this paragraph?

framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
```diff
@@ -736,8 +736,11 @@ protected void addInformationFromPreconditions(
                         } else {
                             thenStore.insertValue(secondInternal, firstValue);
                         }
-                        return new ConditionalTransferResult<>(
-                                res.getResultValue(), thenStore, elseStore);
+                        // To handle `(a = b = c) == x`, repeat for all insertable receivers of
+                        // splitted assignments instead of returning.
```
wmdietl: Was there already a test for this and it failed without this change?
If not, can you add a dedicated test that triggers this problem?

bitterfox: Yes. I already have a test case. `testStrings2` in `framework/tests/value/Switch.java` fails without this change.

dataflow/manual/content.tex
bitterfox: I'll update dataflow manual about this point in another new PR.

bitterfox:
Hi Werner,
I've fixed. Please check this again.
If you merged postfix inc dec PR, please wait for merging this PR to remove `findOwner` in this PR.

wmdietl:
@smillst If you want to have a look through the Value Checker tests.

PR #1364, opened 2017-06-26
Signedness value

cmackie:
This branch integrates the Value Checker into the Signedness Checker and uses it to add new introduction rules for identifiers.


checker/src/org/checkerframework/checker/signedness/SignednessAnnotatedTypeFactory.java
```diff
@@ -109,6 +121,14 @@ protected TreeAnnotator createTreeAnnotator() {
     }
 
     /**
+     * Provides a way to query the Constant Value Checker, which computes the values of expressions
+     * known at compile time (constant propagation and folding).
+     */
+    ValueAnnotatedTypeFactory getValueAnnotatedTypeFactory() {
```
mernst: How about caching this value for efficiency?  Also, I think this method should be private.

checker/src/org/checkerframework/checker/signedness/SignednessAnnotatedTypeFactory.java
```diff
@@ -154,5 +174,42 @@ public Void visitCompoundAssignment(CompoundAssignmentTree tree, AnnotatedTypeMi
             annotateBoolean(type);
             return null;
         }
+
+        @Override
+        public Void visitIdentifier(IdentifierTree tree, AnnotatedTypeMirror type) {
+            TypeMirror javaType = type.getUnderlyingType();
+            ValueAnnotatedTypeFactory valFact = getValueAnnotatedTypeFactory();
```
mernst: For efficiency, should these two lines be executed only if javaType is a primitive?

checker/src/org/checkerframework/checker/signedness/SignednessAnnotatedTypeFactory.java
```diff
@@ -154,5 +174,42 @@ public Void visitCompoundAssignment(CompoundAssignmentTree tree, AnnotatedTypeMi
             annotateBoolean(type);
             return null;
         }
+
+        @Override
+        public Void visitIdentifier(IdentifierTree tree, AnnotatedTypeMirror type) {
```
mernst: Please write a one-line `//` comment to say what this overriding implementation does.

checker/src/org/checkerframework/checker/signedness/SignednessAnnotatedTypeFactory.java
```diff
@@ -154,5 +174,42 @@ public Void visitCompoundAssignment(CompoundAssignmentTree tree, AnnotatedTypeMi
...
+            TypeMirror javaType = type.getUnderlyingType();
+            ValueAnnotatedTypeFactory valFact = getValueAnnotatedTypeFactory();
+            Range treeRange = IndexUtil.getPossibleValues(valFact.getAnnotatedType(tree), valFact);
+
+            if (treeRange != null) {
+                switch (javaType.getKind()) {
+                    case BYTE:
```
mernst: This only handles primitives, but the Value Checker handles wrappers such as `Integer`.  Please support them or write a comment explaining why they are not supported.

checker/src/org/checkerframework/checker/signedness/SignednessAnnotatedTypeFactory.java
smillst: The annotated type factory of a subchecker should not be cached.  See [the documentation of getTypeFactoryOfSubchecker](https://checkerframework.org/api/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.html#getTypeFactoryOfSubchecker-java.lang.Class-).

mernst:
Chris, can you answer my question about why the rules are only applied to identifiers?
Thanks!

cmackie:
The new rules can absolutely be applied to expressions, and should be. I ran out of time to implement those rules for expressions, but I figured the rules as applied to literals could be pushed now since they are sound and still pretty helpful.

mernst:
If you won't have time to expand the rules beyond expressions, could you document in the code that doing so is desirable, and also add an issue so that someone fixes this in the future?

mernst:
A ping on the documentation and new issue I suggested.  Then we can close this issue, in advance of the August 1 release.

smillst:
The new issue is https://github.com/typetools/checker-framework/issues/1430.

PR #1370, opened 2017-06-27
Fix ValueCheckerUtils.getValuesFromRange

panacekcz:
The method `ValueCheckerUtils.getValuesFromRange` contains code that does not work properly. Furthemore, the method may cause an `OutOfMemoryError` if the upper bound of the input range is `Long.MAX_VALUE`, because in that case the condition of the loop `value<=range.to` always succeeds and the loop variable `value` overflows. Then the list of values grows until memory is exhausted.
This problem can be reproduced by running the value checker on this code:
```java
public class LongMax {
 public void longMaxRange() {
  @IntRange(from = 9223372036854775807l, to = 9223372036854775807l) long i = 9223372036854775807l;
  @IntRange(from = 9223372036854775807l, to = 9223372036854775807l) long j = i;
 }
}
```
Run the checker by:
```
checker/bin/javac -processor org.checkerframework.common.value.ValueChecker LongMax.java
```
The error output is:
```
error: SourceChecker.typeProcess: unexpected Throwable (OutOfMemoryError) while processing LongMax.java; message: Java heap space; invoke the compiler with -AprintErrorStack to see the stack trace.
```

This PR changes the implementation to:
- not cause `OutOfMemoryError` if `range.to` is `Long.MAX_VALUE`. This is done by looping from 0 to `range.to-range.from` instead of looping from `range.from` to `range.to`. This makes the upper bound of the loop very small (limited by `ValueAnnotatedTypeFactory.MAX_VALUES`), so the loop variable will not overflow before hitting the bound.
- support conversion to `Integer`, `Short`, `Byte` and `Float`
- remove support (which didn't work) for primitive types
- avoid unnecessary boxing and unboxing

An earlier similar change of this method (not fixing the out-of-memory error) was a part of #1352. This PR should be merged before #1352, which requires these fixes.

### DIFF
--- a/framework/src/org/checkerframework/common/value/ValueCheckerUtils.java
+++ b/framework/src/org/checkerframework/common/value/ValueCheckerUtils.java
@@ -134,11 +134,10 @@ public class ValueCheckerUtils {
     }
 
     /**
-     * Get all possible values from the given type and cast them into Long type, Double type, or
-     * Character type accordingly. Only support casting to integral type and double type.
+     * Get all possible values from the given type and cast them into a boxed primitive type.
      *
      * @param range the given range
-     * @param expectedType the expected type
+     * @param expectedType the expected type (must be a class type)
      * @return a list of all the values in the range
      */
     public static <T> List<T> getValuesFromRange(Range range, Class<T> expectedType) {
@@ -149,27 +148,33 @@ public class ValueCheckerUtils {
         if (range.isNothing()) {
             return values;
         }
-        if (expectedType == Integer.class
-                || expectedType == int.class
-                || expectedType == Long.class
-                || expectedType == long.class
-                || expectedType == Short.class
-                || expectedType == short.class
-                || expectedType == Byte.class
-                || expectedType == byte.class) {
-            for (Long value = range.from; value <= range.to; value++) {
-                values.add(expectedType.cast(value.longValue()));
+        if (expectedType == Integer.class) {
+            for (long value = range.from; value <= range.to; value++) {
+                values.add(expectedType.cast((int) value));
             }
-        } else if (expectedType == Double.class
-                || expectedType == double.class
-                || expectedType == Float.class
-                || expectedType == float.class) {
-            for (Long value = range.from; value <= range.to; value++) {
-                values.add(expectedType.cast(value.doubleValue()));
+        } else if (expectedType == Short.class) {
+            for (long value = range.from; value <= range.to; value++) {
+                values.add(expectedType.cast((short) value));
             }
-        } else if (expectedType == Character.class || expectedType == char.class) {
-            for (Long value = range.from; value <= range.to; value++) {
-                values.add(expectedType.cast((char) value.intValue()));
+        } else if (expectedType == Byte.class) {
+            for (long value = range.from; value <= range.to; value++) {
+                values.add(expectedType.cast((byte) value));
+            }
+        } else if (expectedType == Long.class) {
+            for (long value = range.from; value <= range.to; value++) {
+                values.add(expectedType.cast(value));
+            }
+        } else if (expectedType == Double.class) {
+            for (long value = range.from; value <= range.to; value++) {
+                values.add(expectedType.cast((double) value));
+            }
+        } else if (expectedType == Float.class) {
+            for (long value = range.from; value <= range.to; value++) {
+                values.add(expectedType.cast((float) value));
+            }
+        } else if (expectedType == Character.class) {
+            for (long value = range.from; value <= range.to; value++) {
+                values.add(expectedType.cast((char) value));
             }
         } else {
             throw new UnsupportedOperationException(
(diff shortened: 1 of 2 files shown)


framework/src/org/checkerframework/common/value/ValueCheckerUtils.java
```diff
@@ -134,11 +134,10 @@ public static Range getRangeFromValues(List<? extends Number> values) {
     }
 
     /**
-     * Get all possible values from the given type and cast them into Long type, Double type, or
-     * Character type accordingly. Only support casting to integral type and double type.
+     * Get all possible values from the given type and cast them into a boxed primitive type.
      *
      * @param range the given range
-     * @param expectedType the expected type
+     * @param expectedType the expected type (must be a boxed type, not a primitive type)
```
wmdietlGC: Can you make this a separate paragraph and explain, why primitive types don't work?
I guess it's because a `List` can only contain boxed types, but it would be good to make explicit.

framework/src/org/checkerframework/common/value/ValueCheckerUtils.java
```diff
@@ -149,31 +148,35 @@ public static Range getRangeFromValues(List<? extends Number> values) {
...
-            for (Long value = range.from; value <= range.to; value++) {
-                values.add(expectedType.cast((char) value.intValue()));
+
+        // Does not overflow, because the width has already been checked.
+        long length = range.to - range.from;
+
+        // Each value is computed as a sum of the first value and an offset within the range,
+        // to avoid comparing to range.to, which may be Long.MAX_VALUE
+        for (long offset = 0; offset <= length; offset++) {
+            long value = range.from + offset;
```
wmdietlGC: I don't see how this rewrite is any different than the logic before and this new code has an additional addition.
If `range.from` is `0` and `range.to` is `Long.MAX_VALUE`, both versions iterate over `Long.MAX_VALUE` elements. What is the benefit of using this offset?

framework/tests/value/LongMax.java
```diff
@@ -0,0 +1,9 @@
+import org.checkerframework.common.value.qual.*;
+// Test for handling value annotations involving Long.MAX_VALUE
```
wmdietlGC: Can you describe what fails before this fix?

framework/src/org/checkerframework/common/value/ValueCheckerUtils.java
panacekcz: This code is executed only for ranges not wider than `ValueAnnotatedTypeFactory.MAX_VALUES`, so here the upper bound of the loop is always a very small number. I will change the comments to explain that. Or I can implement it in a different way if you think that would be better.

wmdietlGC: What is the difference between going from `range.from` to `range.to` and going from `0` to `range.to - range.from` and then adding `range.from` to each value?
In the PR description, can you add what problem you are trying to fix and be explicit about what goes wrong?

panacekcz:
I have addressed the comments and questions in the code and updated the PR description to explain the problem.

framework/tests/value/LongMax.java
```diff
@@ -0,0 +1,15 @@
+import org.checkerframework.common.value.qual.*;
```
wmdietlGC: Only import the IntRange annotation you actually need.

framework/src/org/checkerframework/common/value/ValueCheckerUtils.java
```diff
@@ -149,31 +151,38 @@ public static Range getRangeFromValues(List<? extends Number> values) {
...
+            } else if (expectedType == Character.class) {
+                convertedValue = (char) value;
+            } else {
+                throw new UnsupportedOperationException(
+                        "ValueCheckerUtils: unexpected class: " + expectedType);
             }
-        } else {
-            throw new UnsupportedOperationException(
-                    "ValueCheckerUtils: unexpected class: " + expectedType);
+            values.add(expectedType.cast(convertedValue));
```
wmdietlGC: Why do you need to do both the cast in the if-checks above and this `expectedType.cast`?
Can't you keep the basic structure of the previous code, simply removing the primitive cases?
Having the conditions outside of the loop would also make the code more readable. The duplication of the loop logic doesn't seem so bad.

wmdietlGC:
Thanks, the PR description is much clearer now!

framework/src/org/checkerframework/common/value/ValueCheckerUtils.java
panacekcz: `Class.cast` throws `ClassCastException` if the argument is not of the correct type, so a cast is needed to perform the conversion. `Class.cast` is used for casting to `T` without needing to suppress an unchecked cast warning.
I thought it would be too much duplication to have the loop 7 times there. If this code is not readable enough, I think extracting conversion logic to a separate method should help.

panacekcz:
I have addressed the comments and tried to make the code more readable by extracting the conversion to a separate method. If you still think duplicating the loop would be better, here the code: https://github.com/panacekcz/checker-framework/commit/cbc0c3c1bc7880fa73b5d8f0bc91c5dc901480f1

PR #1373, opened 2017-06-29
Fix 1366 by checking for uninferred type arguments in constraints 

### LINKED ISSUE #1366: Nullness Checker crash: AsSuperVisitor: type is not an erased subtype of supertype 4
After lots of trial and error I finally managed to reproduce the crash mentioned in https://github.com/typetools/checker-framework/issues/1312#issuecomment-309198631 and the source is quite simple:

```java
abstract class Foo<T> {
  abstract <S> Foo<S> m1(Foo<S> p1, Foo<?> p2);

  abstract <S> Foo<S> m2(Foo<? extends S> p);

  abstract void m3(Foo<Number> p);

  void foo(Foo<Number> s) {
    s.m3(s.m2(s.m1(s, s)));
  }
}
```

Running the Nullness Checker results in the following crash (more after stack trace):

```
  type: @Initialized @Nullable Object
  superType: @Initialized @NonNull Number
  Compilation unit: ACrash.java
  Exception: java.lang.Throwable; Stack trace: org.checkerframework.framework.source.SourceChecker.errorAbort(SourceChecker.java:727)
  org.checkerframework.javacutil.ErrorReporter.errorAbort(ErrorReporter.java:41)
  org.checkerframework.framework.type.AsSuperVisitor.errorTypeNotErasedSubtypeOfSuperType(AsSuperVisitor.java:155)
  org.checkerframework.framework.type.AsSuperVisitor.visitDeclared_Declared(AsSuperVisitor.java:340)
  org.checkerframework.framework.type.AsSuperVisitor.visitDeclared_Declared(AsSuperVisitor.java:29)
  org.checkerframework.framework.util.AtmCombo.accept(AtmCombo.java:308)
  org.checkerframework.framework.type.visitor.AbstractAtmComboVisitor.visit(AbstractAtmComboVisitor.java:59)
  org.checkerframework.framework.type.AsSuperVisitor.visit(AsSuperVisitor.java:86)
  org.checkerframework.framework.type.AsSuperVisitor.asSuper(AsSuperVisitor.java:68)
  org.checkerframework.framework.util.AnnotatedTypes.asSuper(AnnotatedTypes.java:131)
  org.checkerframework.framework.util.QualifierPolymorphism$PolyCollector.visitDeclared(QualifierPolymorphism.java:458)
  org.checkerframework.framework.util.QualifierPolymorphism$PolyCollector.visitDeclared(QualifierPolymorphism.java:342)
  org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedDeclaredType.accept(AnnotatedTypeMirror.java:900)
  org.checkerframework.framework.type.visitor.SimpleAnnotatedTypeVisitor.visit(SimpleAnnotatedTypeVisitor.java:64)
  org.checkerframework.framework.util.QualifierPolymorphism$PolyCollector.visit(QualifierPolymorphism.java:393)
  org.checkerframework.framework.util.QualifierPolymorphism$PolyCollector.visitDeclared(QualifierPolymorphism.java:477)
  org.checkerframework.framework.util.QualifierPolymorphism$PolyCollector.visitDeclared(QualifierPolymorphism.java:342)
  org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedDeclaredType.accept(AnnotatedTypeMirror.java:900)
  org.checkerframework.framework.type.visitor.SimpleAnnotatedTypeVisitor.visit(SimpleAnnotatedTypeVisitor.java:64)
  org.checkerframework.framework.util.QualifierPolymorphism$PolyCollector.visit(QualifierPolymorphism.java:393)
  org.checkerframework.framework.util.QualifierPolymorphism.annotate(QualifierPolymorphism.java:202)
  org.checkerframework.framework.type.GenericAnnotatedTypeFactory.methodFromUse(GenericAnnotatedTypeFactory.java:1369)
  org.checkerframework.checker.nullness.NullnessAnnotatedTypeFactory.methodFromUse(NullnessAnnotatedTypeFactory.java:299)
```

It turns out that using `Number` as type argument makes a difference. Using `String` instead results in a type error:

```
ACrash.java:9: error: [argument.type.incompatible] incompatible types in argument.
    s.m3(s.m2(s.m1(s, s)));
             ^
  found   : @Initialized @NonNull Foo<@Initialized @Nullable Object>
  required: @Initialized @NonNull Foo<@Initialized @NonNull String>
1 error
```

smillst:
Also, fixes bug in `InternalUtils.lub` uncovered by this fix.

Fixes #1366.

### DIFF
--- a/checker/src/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
+++ b/checker/src/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
@@ -55,6 +55,7 @@ import org.checkerframework.framework.type.typeannotator.ImplicitsTypeAnnotator;
 import org.checkerframework.framework.type.typeannotator.ListTypeAnnotator;
 import org.checkerframework.framework.type.typeannotator.PropagationTypeAnnotator;
 import org.checkerframework.framework.type.typeannotator.TypeAnnotator;
+import org.checkerframework.framework.type.visitor.SimpleAnnotatedTypeScanner;
 import org.checkerframework.framework.util.AnnotatedTypes;
 import org.checkerframework.framework.util.DependentTypes;
 import org.checkerframework.framework.util.MultiGraphQualifierHierarchy.MultiGraphFactory;
@@ -301,11 +302,23 @@ public class NullnessAnnotatedTypeFactory
                 super.methodFromUse(tree);
         AnnotatedExecutableType method = mfuPair.first;
 
-        if (method.getReturnType().getKind() == TypeKind.WILDCARD
-                && ((AnnotatedWildcardType) method.getReturnType()).isUninferredTypeArgument()
-                && !checker.hasOption("conservativeUninferredTypeArguments")) {
-            method.getReturnType().replaceAnnotations(qualHierarchy.getBottomAnnotations());
-        }
+        // Change the type any uninferred type arguments to @NonNull @Initialized.
+        SimpleAnnotatedTypeScanner<Void, Void> scan =
+                new SimpleAnnotatedTypeScanner<Void, Void>() {
+                    @Override
+                    protected Void defaultAction(AnnotatedTypeMirror type, Void aVoid) {
+                        if (type == null) {
+                            return null;
+                        }
+                        if (type.getKind() == TypeKind.WILDCARD
+                                && ((AnnotatedWildcardType) type).isUninferredTypeArgument()
+                                && !checker.hasOption("conservativeUninferredTypeArguments")) {
+                            type.replaceAnnotations(qualHierarchy.getBottomAnnotations());
+                        }
+                        return null;
+                    }
+                };
+        scan.visit(method.getReturnType());
 
         systemGetPropertyHandler.handle(tree, method);
         collectionToArrayHeuristics.handle(tree, method);
--- a/checker/src/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
+++ b/checker/src/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
@@ -55,7 +55,6 @@ import org.checkerframework.framework.type.typeannotator.ImplicitsTypeAnnotator;
 import org.checkerframework.framework.type.typeannotator.ListTypeAnnotator;
 import org.checkerframework.framework.type.typeannotator.PropagationTypeAnnotator;
 import org.checkerframework.framework.type.typeannotator.TypeAnnotator;
-import org.checkerframework.framework.type.visitor.SimpleAnnotatedTypeScanner;
 import org.checkerframework.framework.util.AnnotatedTypes;
 import org.checkerframework.framework.util.DependentTypes;
 import org.checkerframework.framework.util.MultiGraphQualifierHierarchy.MultiGraphFactory;
@@ -302,23 +301,11 @@ public class NullnessAnnotatedTypeFactory
                 super.methodFromUse(tree);
         AnnotatedExecutableType method = mfuPair.first;
 
-        // Change the type any uninferred type arguments to @NonNull @Initialized.
-        SimpleAnnotatedTypeScanner<Void, Void> scan =
-                new SimpleAnnotatedTypeScanner<Void, Void>() {
-                    @Override
-                    protected Void defaultAction(AnnotatedTypeMirror type, Void aVoid) {
-                        if (type == null) {
-                            return null;
-                        }
-                        if (type.getKind() == TypeKind.WILDCARD
-                                && ((AnnotatedWildcardType) type).isUninferredTypeArgument()
-                                && !checker.hasOption("conservativeUninferredTypeArguments")) {
-                            type.replaceAnnotations(qualHierarchy.getBottomAnnotations());
-                        }
-                        return null;
-                    }
-                };
-        scan.visit(method.getReturnType());
+        if (method.getReturnType().getKind() == TypeKind.WILDCARD
+                && ((AnnotatedWildcardType) method.getReturnType()).isUninferredTypeArgument()
+                && !checker.hasOption("conservativeUninferredTypeArguments")) {
+            method.getReturnType().replaceAnnotations(qualHierarchy.getBottomAnnotations());
+        }
 
         systemGetPropertyHandler.handle(tree, method);
         collectionToArrayHeuristics.handle(tree, method);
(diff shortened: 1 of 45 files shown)


framework/src/org/checkerframework/framework/util/typeinference/constraint/AFReducingVisitor.java
```diff
@@ -585,4 +594,15 @@ public void visitWildcardAsSuperType(
...
         addInverseConstraint(subtype, supertype.getExtendsBound(), constraints);
     }
+
+    /**
+     * Checks that the wildcard is not an uninferred type argument. If it is, errorAbort will be
+     * called.
+     */
+    private void checkForUninferredTypes(AnnotatedWildcardType wildcardType) {
+        if (wildcardType.isUninferredTypeArgument()) {
+            ErrorReporter.errorAbort(
```
wmdietlGC: Won't this now simply terminate in a different location with a different error message?
How does this make the test case pass? Or is the important fix the fix in InternalUtils and this change is to detect similar problem faster in the future?

framework/src/org/checkerframework/framework/util/typeinference/constraint/AFReducingVisitor.java
```diff
@@ -585,4 +594,15 @@ public void visitWildcardAsSuperType(
...
+                    "Can't make a constraint that includes an uninferred " + "type argument.");
```
wmdietlGC: Should this include the `wildcardType`? Or why the `+` in the middle?

framework/src/org/checkerframework/framework/util/typeinference/constraint/AFReducingVisitor.java
smillst: I updated the comment on this method to clarify.

checker/tests/nullness/java8inference/Issue980.java
```diff
@@ -24,6 +24,8 @@ void m(List<String> strings) {
         // This assignment issues a warning due to incompatible types:
```
wmdietlGC: Related to your comment below: why is there no expected warning about these incompatible types?

framework/src/org/checkerframework/framework/util/typeinference/constraint/AFReducingVisitor.java
```diff
@@ -585,4 +595,16 @@ public void visitWildcardAsSuperType(
         // in fact
         addInverseConstraint(subtype, supertype.getExtendsBound(), constraints);
     }
+
+    /**
+     * Checks that the wildcard is not an uninferred type argument. If it is, errorAbort will be
+     * called. The error will be caught in {@link DefaultTypeArgumentInference#infer} and inference
+     * will be aborted, but type-checking will continue.
```
wmdietlGC: Ah, thanks for clarifying what is going on.

checker/tests/nullness/java8inference/Issue980.java
```diff
@@ -24,6 +24,8 @@ void m(List<String> strings) {
...
         List<String> collectedStrings = s.collect(Collectors.toList());
 
+        // This issues an error because type argument inference fails on the above statement.
```
wmdietlGC: This error is a pity, because this is a false positive caused by the type argument inference failure.
On the statement above, can we use a more optimistic type (unless we provide conservative flags)?

checker/tests/nullness/java8inference/Issue980.java
smillst: It appears that the warning went away when the conservative type argument option was added in #1065.  (This code should not issue a warning, I think that's what this issue was about originally.)  I'll correct the comment.

wmdietlGC:
For the failure on this line: https://github.com/typetools/checker-framework/blob/master/checker/tests/nullness/generics/Issue329.java#L37
we have two options:
- Mark that this is a false negative and make sure that the error appears again when inference is properly implemented
- Run test cases with the conservative flag, which would then require adding expected errors for the false positives we would detect.

I have a slight preference for the latter, but don't feel strongly about this.

checker/src/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
```diff
@@ -299,6 +301,12 @@ protected AnnotatedTypeFormatter createAnnotatedTypeFormatter() {
                 super.methodFromUse(tree);
         AnnotatedExecutableType method = mfuPair.first;
 
+        if (method.getReturnType().getKind() == TypeKind.WILDCARD
```
wmdietlGC: Can we make this a deep check, to cover if the uninferred wildcard occurs as an array component type or as a type argument? Those cases also happen.

checker/src/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
```diff
@@ -299,6 +302,24 @@ protected AnnotatedTypeFormatter createAnnotatedTypeFormatter() {
                 super.methodFromUse(tree);
         AnnotatedExecutableType method = mfuPair.first;
 
+        // Change the type any uninferred type arguments to @NonNull @Initialized.
```
smillst: This caused new failures in plume-lib: https://travis-ci.org/smillst/checker-framework/jobs/249713675#L1820.

wmdietlGC: Yes, this isn't working well and I also see lots of errors.
In the plume-lib log all cases seem to be with the un-inferred wildcard in the subtype and the supertype being a type parameter. Maybe that combination needs some extra love in the qualifier hierarchy?

smillst:
@wmdietl  As discussed, I removed the code that is attempting to reduce false positives.  I'll open a separate pull request with those changes (and more changes as dicussed in https://github.com/typetools/checker-framework/pull/1373#discussion_r125345861).

checker/tests/nullness/java8inference/Issue980.java
```diff
@@ -21,9 +21,9 @@ void m(List<String> strings) {
         @SuppressWarnings("nullness")
```
wmdietlGC: Is this `@SuppressWarnings` unnecessary? As `collectedStrings` no longer produces a warning, I would expect that this one also doesn't.

checker/tests/nullness/java8inference/Issue980.java
```diff
@@ -21,9 +21,9 @@ void m(List<String> strings) {
...
         List<String> collectedStrings3 = s.collect(Collectors.toList());
 
-        // This assignment issues a warning due to incompatible types:
         List<String> collectedStrings = s.collect(Collectors.toList());
 
+        //:: error: (dereference.of.nullable)
```
wmdietlGC: This is a false positive, right? Can you add a comment to that effect?

PR #1374, opened 2017-06-30
Nullness JDK upper bounds

mernst:


checker/jdk/nullness/src/java/lang/reflect/Constructor.java
```diff
@@ -6,7 +6,7 @@
 import org.checkerframework.checker.nullness.qual.NonNull;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
-public final class Constructor<T> extends AccessibleObject implements GenericDeclaration, Member {
+public final class Constructor<T extends @Nullable Object> extends AccessibleObject implements GenericDeclaration, Member {
```
smillst: Should this be `@Covariant(0)`, too?

smillst:
Adding `@Covariant` to Class and Optional made all the tests pass except for two involving type argument inference.  (See https://github.com/typetools/checker-framework/pull/1374/commits/9bbd89c1ad8440d8fae453a703b6c2ce40741fbd.)

checker/jdk/nullness/src/java/lang/reflect/Constructor.java
mernst: We don't have a test case that demonstrates the need for this, but I agree it's a good idea for consistency.  I will make the change and push it after local tests pass.

mernst:
To follow up to @smillst's comment about the two failing tests, she has suppressed the warnings on those two tests, so perhaps this branch will pass tests and be mergeable.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1972,7 +1972,7 @@ public final AnnotatedTypeMirror getReceiverType(ExpressionTree expression) {
      * @param receiverType the receiver type of the method invocation (not the declared receiver
      *     type)
      */
-    protected static void adaptGetClassReturnTypeToReceiver(
+    public void adaptGetClassReturnTypeToReceiver(
```
wmdietl: Can't this method stay protected? I don't see a new invocation.

checker/jdk/nullness/src/java/lang/Class.java
```diff
@@ -54,7 +59,7 @@ protected Class() {}
   public boolean desiredAssertionStatus() { throw new RuntimeException("skeleton method"); }
   @Pure public boolean isEnum() { throw new RuntimeException("skeleton method"); }
   public @NonNull T @Nullable [] getEnumConstants() { throw new RuntimeException("skeleton method"); }
-  java.util.Map<String, T> enumConstantDirectory() { throw new RuntimeException("skeleton method"); }
+  java.util.Map<String, @NonNull T> enumConstantDirectory() { throw new RuntimeException("skeleton method"); }
```
wmdietl: Can you add a test case for enumConstantDirectory and why the value is non-null?
Does this work:

<T> Map<String, T> getECD(Class<T> p) {
  return p.enumConstantDirectory();
}

checker/jdk/nullness/src/java/lang/Class.java
```diff
@@ -6,8 +6,13 @@
 import org.checkerframework.checker.nullness.qual.PolyNull;
 import org.checkerframework.dataflow.qual.Pure;
 import org.checkerframework.dataflow.qual.SideEffectFree;
+import org.checkerframework.framework.qual.Covariant;
 
-public final class Class<T> extends Object implements java.io.Serializable, java.lang.reflect.GenericDeclaration, java.lang.reflect.Type, java.lang.reflect.AnnotatedElement {
+// The type annotation on an instantiation of Class is meaningless.
+// Class<@NonNull String> and Class<@Nullable String> have the same
+// meaning, but unfortunately are not interchangeable.
+@Covariant(0)
```
wmdietl: Just to be absolutely clear, maybe expand on the `interchangeable` and say that the `@Covariant` allows one direction, but not the other.

checker/jdk/nullness/src/java/lang/Class.java
```diff
@@ -6,8 +6,13 @@
 import org.checkerframework.checker.nullness.qual.PolyNull;
 import org.checkerframework.dataflow.qual.Pure;
 import org.checkerframework.dataflow.qual.SideEffectFree;
+import org.checkerframework.framework.qual.Covariant;
 
-public final class Class<T> extends Object implements java.io.Serializable, java.lang.reflect.GenericDeclaration, java.lang.reflect.Type, java.lang.reflect.AnnotatedElement {
+// The type annotation on an instantiation of Class is meaningless.
```
wmdietl: I would rather call it the type argument to Class.

checker/jdk/nullness/src/java/lang/Class.java
```diff
@@ -6,8 +6,13 @@
...
+// Class<@NonNull String> and Class<@Nullable String> have the same
+// meaning, but unfortunately are not interchangeable.
+@Covariant(0)
+public final class Class<T extends @Nullable Object> extends Object implements java.io.Serializable, java.lang.reflect.GenericDeclaration, java.lang.reflect.Type, java.lang.reflect.AnnotatedElement {
```
wmdietl: Super long line.

framework/tests/all-systems/GetClassTest.java
```diff
@@ -6,11 +8,29 @@ void context() {
         Integer i = 4;
         i.getClass();
         Class<?> a = i.getClass();
-        // Type agruments don't match
-        @SuppressWarnings("fenum:assignment.type.incompatible")
+        // Type arguments don't match
+        @SuppressWarnings({
+            "fenum:assignment.type.incompatible",
+            "nullness:assignment.type.incompatible"
```
wmdietl: Shouldn't this work for nullness? adaptGetClassReturnTypeToReceiver changes the upper bound of getClass to be non-null, so `Class<? extends Object>` should be a supertype of `Class<? extends Integer>`.
Same issues with other tests.

checker/jdk/nullness/src/java/lang/Class.java
mernst: I don't think every annotation in the JDK merits a test case.  An enum constant cannot be null, so this annotation is correct and necessary.

wmdietlGC: I provided a concrete test case that doesn't work with this annotation. You didn't provide a case that requires it.

PR #1380, opened 2017-07-06
Store extra keyfor information in KeyForValue.

### LINKED ISSUE #961: KeyFor refinement of type variables
```
Map<T, Object> map = ...;
<T> T method(T param) {
  if (map.contains(param) {
      @NonNull Object o = map.get(param);
      return param;
   }
}
````

The declared type of T is `@KeyForUnknown`.  Within the `if` statement, there are two possibilities:
 * the type of `T` is refined, in which case the get succeeds (it returns a `@NonNull` value), but the  `return param;` fails.
 * the type of `T` is not refined, in which case the get fails, but the `return param;` succeeds.

We would like both lines to succeed.

Here is a proposed design that would accommodate both:

Essentially, instead of representing a `@KeyFor` annotation as a set of keys, it would represent a `@KeyFor` annotation as a pair:
 * an optional type being refined
 * an explicit set

For an ordinary `@KeyFor` annotation, the type being refined is missing/empty.
The type of `param` within the `if` statement would be:
 * refining `T`
 * also `map`
Thus, the call to `get` works, but also `param` is a subtype of the return type and the return statement works.

### PR DESCRIPTION
**Pull request by smillst**
Fixes #961

### CODE REVIEW

File: checker/src/org/checkerframework/checker/nullness/KeyForAnalysis.java [hunk 1]
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Set;
+import javax.lang.model.element.AnnotationMirror;
+import javax.lang.model.element.VariableElement;
+import javax.lang.model.type.TypeMirror;
+import org.checkerframework.common.basetype.BaseTypeChecker;
+import org.checkerframework.framework.flow.CFAbstractAnalysis;
+import org.checkerframework.framework.flow.CFAbstractValue;
+import org.checkerframework.javacutil.Pair;
+
+public class KeyForAnalysis extends CFAbstractAnalysis<KeyForValue, KeyForStore, KeyForTransfer> {
```
wmdietl: A quick note that everything in this class is just to glue together the other classes would safe time in the future. I looked through everything and didn't find any actual logic (which is something we should try to automate...). If I missed something, even more reason to document here.

checker/src/org/checkerframework/checker/nullness/KeyForAnnotatedTypeFactory.java
```diff
@@ -251,15 +254,19 @@ public AnnotationMirror createKeyForAnnotationMirrorWithValue(String value) {
...
+        Collection<String> maps = null;
         AnnotatedTypeMirror type = getAnnotatedType(tree);
         AnnotationMirror keyForAnno = type.getAnnotation(KeyFor.class);
-        if (keyForAnno == null) {
-            return false;
+        if (keyForAnno != null) {
+            maps = AnnotationUtils.getElementValueArray(keyForAnno, "value", String.class, false);
+        } else {
+            KeyForValue value = getInferredValueFor(tree);
+            if (value != null && value.keyForMaps != null) {
```
wmdietl: The second null check seems redundant, b/c maps is checked again below.

checker/src/org/checkerframework/checker/nullness/KeyForTransfer.java
```diff
@@ -11,21 +11,18 @@
...
  * cause the appropriate @KeyFor annotation to be added to the key.
  */
-public class KeyForTransfer extends CFTransfer {
+public class KeyForTransfer extends CFAbstractTransfer<KeyForValue, KeyForStore, KeyForTransfer> {
 
     protected final AnnotationMirror UNKNOWNKEYFOR, KEYFOR;
 
-    public KeyForTransfer(CFAnalysis analysis) {
+    public KeyForTransfer(KeyForAnalysis analysis) {
         super(analysis);
```
wmdietl: Can we change the super implementation of this method to do `this(analysis, false);` Passing false to forceConcurrentSemantics will not force and then the two constructors share the same implementation.
Then similarly, in this constructor I would like to use `this(analysis, false);` to not have to duplicate the field initialization, simplifying this code.

checker/src/org/checkerframework/checker/nullness/KeyForValue.java
```diff
@@ -0,0 +1,113 @@
...
+import javax.lang.model.type.TypeMirror;
+import org.checkerframework.checker.nullness.qual.KeyFor;
+import org.checkerframework.framework.flow.CFAbstractAnalysis;
+import org.checkerframework.framework.flow.CFAbstractValue;
+import org.checkerframework.javacutil.AnnotationUtils;
+
+/**
+ * KeyForValue holds additional information about which maps this value is a key for. This extra
+ * information is required when adding the @KeyFor qualifier to the type is not a refinement of
+ * type. For example,
```
wmdietl: Add `the` to make `... not a refinement of the type.` which made the sentence easier to read. Maybe rewording this part can also make this easier.

checker/src/org/checkerframework/checker/nullness/KeyForValue.java
```diff
@@ -0,0 +1,113 @@
...
+ * }</pre>
+ *
+ * Inside the if statement, {@code param} is a key for "map". This would normally be represented as
+ * {@code @KeyFor("map") T}, but this is not a subtype of {@code T}, so the type cannot be refined.
+ * Instead, the value for {@code param} includes "map" in the list of keyForMaps. This information
+ * is used in {@link KeyForAnnotatedTypeFactory#isKeyForMap(String, ExpressionTree)}.
+ */
+public class KeyForValue extends CFAbstractValue<KeyForValue> {
+    /**
+     * If the underlying type is a type variable or a wildcard, then this is list of maps this for
```
wmdietl: I can't quite parse the `then` part of the sentence.

checker/src/org/checkerframework/checker/nullness/KeyForValue.java
```diff
@@ -0,0 +1,113 @@
...
+     * which this value is a key. Otherwise, it's null.
+     */
+    Set<String> keyForMaps = null;
```
wmdietl: Field visibility missing. Make private and add a getter that returns an immutable copy? Or document who is supposed to access?

checker/src/org/checkerframework/checker/nullness/KeyForValue.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+    }
+
+    @Override
+    public KeyForValue leastUpperBound(KeyForValue other) {
+        KeyForValue lub = super.leastUpperBound(other);
+        if (other == null || other.keyForMaps == null || this.keyForMaps == null) {
+            return lub;
+        }
+        // Lub the keyFoMaps by intersecting the sets.
```
wmdietl: keyFoMaps -> keyForMaps

checker/src/org/checkerframework/checker/nullness/KeyForValue.java
```diff
@@ -0,0 +1,113 @@
...
+        return lub;
+    }
+
+    @Override
+    public KeyForValue mostSpecific(KeyForValue other, KeyForValue backup) {
+        KeyForValue mostSpecific = super.mostSpecific(other, backup);
+        if (mostSpecific == null) {
+            // mostSpecific is null if the two types are not comparable.  This is normally
+            // because one of this or other is a type variable and annotations is empty, but the
+            // other annotations are not empty.  In this case copy the keyForMaps and
```
wmdietl: And do what?

checker/src/org/checkerframework/checker/nullness/KeyForValue.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+        return mostSpecific;
+    }
+
+    private void addKeyFor(Set<String> newKeyForMaps) {
+        if (newKeyForMaps == null) {
+            return;
+        }
+        if (keyForMaps == null) {
+            keyForMaps = new LinkedHashSet<>();
```
wmdietl: Slightly restructure the logic to not allocate the linked list and then set the field back to null.

checker/tests/nullness/Issue961.java
```diff
@@ -4,18 +4,19 @@
...
         if (map.containsKey(param)) {
             @NonNull Object o = map.get(param);
             return param;
         }
-        return null;
+        return param;
     }
 
     abstract class MapContains<K, V> {
+        @SuppressWarnings("method.invocation.invalid")
```
wmdietl: Add a comment that we want to allow invocation in field initializer. Why suppression instead of expected error?

checker/src/org/checkerframework/checker/nullness/KeyForTransfer.java
smillst: I think subclasses only need to override the single argument constructor unless forcing forcing concurrent semantics is required.  I'll address this in a separate pull request.

PR #1383, opened 2017-07-07
Fix #1382 by moving @Covariant into the framework proper.

### LINKED ISSUE #1382: @Covariant not implemented for Nullness Checker
The Nullness Checker provides a `@Covariant` annotation.
I would expect the following to compile:

```java
import org.checkerframework.checker.nullness.qual.Covariant;

@Covariant(0)
class Box<T> {}

class Demo {
    void foo(Box<@Nullable Object> p) {}
    void bar(Box<Object> p) {
        foo(p);
    }
}
```

However, I get:

```
Error: [argument.type.incompatible] incompatible types in argument.
  found : @Initialized @NonNull Box<@Initialized @NonNull Object>
  required: @Initialized @NonNull Box<@Initialized @Nullable Object>
```

I see some code to handle Covariant in `https://github.com/typetools/checker-framework/blob/master/checker/src/org/checkerframework/checker/nullness/KeyForAnnotatedTypeFactory.java`, but nothing in the NullnessATF.
https://github.com/typetools/checker-framework/blob/master/checker/src/org/checkerframework/checker/nullness/qual/Covariant.java#L9 mentions to move into the framework, but I had expected it to work for all Nullness Checker annotations.
I can work on fixing this, but wanted a second pair of eyes whether I'm doing something wrong.

https://github.com/typetools/checker-framework/blob/master/checker/src/org/checkerframework/checker/nullness/KeyForAnnotatedTypeFactory.java#L187 looks like it wouldn't work with the Covariant annotation being in a stub file, right?

wmdietlGC:


framework/src/org/checkerframework/framework/type/DefaultTypeHierarchy.java
```diff
@@ -537,14 +539,26 @@ public Boolean visitTypeArgs(
...
+            if (covam != null) {
+                covariantArgIndexes =
+                        AnnotationUtils.getElementValueArray(covam, "value", Integer.class, false);
+            }
+
+            for (int i = 0; i < supertypeTypeArgs.size(); i++) {
+                final AnnotatedTypeMirror superTypeArg = supertypeTypeArgs.get(i);
+                final AnnotatedTypeMirror subTypeArg = subtypeTypeArgs.get(i);
+                final boolean covariant =
+                        this.covariantTypeArgs
```
wmdietlGC: I didn't find a use of the constructor that sets this field to true, not within the Checker Framework and not in sparta.
When does it make sense to set `covariantTypeArgs` true? It seems rather risky to allow co-variance for all types.
Does anyone know a reason? Otherwise, I would like to remove it.

smillst: [Glacier](https://github.com/mcoblenz/Glacier/blob/master/src/edu/cmu/cs/glacier/GlacierTypeHierarchy.java#L21) uses it.  But it wouldn't be hard to adapt Glacier to override `compareTypeArgs`.

I agree that it's risky to allow covariance for all types and so the field should be removed.

wmdietlGC: Thanks, I removed the field. It seems much nicer to override that method.
I also added the old annotation back again but deprecated it, to make the transition easier.

All tests pass locally for me and the Travis error doesn't really make sense.
@smillst do you see what is wrong on Travis? Only thing I can think of is downloading the JDK, but that also doesn't make sense, in particular now with the deprecated annotation.

checker/src/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
```diff
@@ -154,6 +154,13 @@ public NullnessAnnotatedTypeFactory(BaseTypeChecker checker, boolean useFbc) {
                 org.checkerframework.checker.nullness.compatqual.MonotonicNonNullType.class,
                 MONOTONIC_NONNULL);
 
+        // Alias deprecated annotation
```
smillst: This should be in the KeyFor Checker.  It's also not copying the values in `@Covariant`, so it's causes crashes:
https://travis-ci.org/typetools/checker-framework/jobs/251223815#L1378

wmdietlGC: But here I can't copy the annotation values, because I don't have a concrete instance yet.
Are you suggesting enhancing the handling of declaration annotation aliases to copy over all values? Do we do that for type annotations?

smillst: I wasn't suggesting anything, just pointing why Travis is crashing.  

We don't copy values for aliased type annotations. Some checkers, such as the Upper Bound Checker, override `AnnotatedTypeFactory#aliasedAnnotation` to copy the values over.

Instead of aliasing, it might be easier to check for both annotations in DefaultTypeHierarchy.

wmdietlGC: I can't check for the keyfor version because that would introduce a dependency from the framework project to the checker project... unless I do a Class.forName, which would be rather ugly. Any objection to such an ugly hack?

smillst: That's fine with me.  Or you could make AnnotatedTypeFactory#getDeclAnnotation(javax.lang.model.element.Element, java.lang.String, boolean) public.  I don't have a preference.

PR #1385, opened 2017-07-07
Building Checker Framework from source on Mac OS

feliperodri:
@mernst,

I believe that the documentation could be a bit more straightforward and point out directly the small differences between building from source on Mac OS and Ubuntu. In particular, I would suggest changes in *33.3.1.* and *33.3.3.* sections.

Thank you,
Felipe.


.gitignore
```diff
@@ -75,6 +75,7 @@ docs/manual/figures/signature-types-with-canonicalname.dia
 docs/manual/*.png
 docs/manual/*.svg
 docs/examples/**/*.class
+docs/manual/manual.html-e
```
mernst: What process creates this file?

docs/manual/troubleshooting.tex
```diff
@@ -772,6 +772,18 @@ \subsection{Install prerequisites\label{building-prerequisites}}
  texlive-latex-base texlive-latex-recommended texlive-latex-extra texlive-fonts-recommended unzip
 \end{Verbatim}
 
+% Building Checker Framework from source on Mac OS
+
+The installation of prerequisites is a bit different on Mac OS, but still easy most of the time with \ahref{https://brew.sh}{Homebrew} through the following commands:
```
mernst: I would avoid the wordy text.  It can be brief like the version for Ubuntu.

docs/manual/troubleshooting.tex
```diff
@@ -838,6 +852,16 @@ \subsection{Build the Type Annotations compiler\label{building-compiler}}
 %END LATEX
 
 \item
+{\bf [Mac OSX 10.5 or later]} It is recommended to set the \code{\$JAVA\_HOME} variable to \</usr/libexec/java\_home>, just export \code{\$JAVA\_HOME} in file \code{.bash\_profile} or \code{.profile}.
```
mernst: Is this guaranteed to work even if the user has installed his or her own JDK, such as JDK 9?
It's easy to write a hard-code command that works in the usual case, but it is better to write one that always works.
If this only works in certain situations, then at least say that.

Also, rather than "it is recommended", say why.

mernst:
Ping.

.gitignore
feliperodri: Every time that a run `make`, `manual.html-e` file is created, but I couldn't identify which specific part creates it. Perhaps this file is only created on Mac OS. Should I remove this line from `.gitignore`?

feliperodri:
Thanks for the feedback @mernst. Could you please re-review this pull request?

.gitignore
mernst: OK, thanks for the explanation.  I wish I knew what created the file, but given your report, it's best to retain the line in the `.gitignore` file.

docs/manual/troubleshooting.tex
```diff
@@ -764,14 +764,28 @@ \section{Building from source\label{build-source}}
 \subsection{Install prerequisites\label{building-prerequisites}}
 
 You need to install several packages in order to build the Checker
-Framework.  The following command installs them for Ubuntu:
+Framework. The following command installs them for Ubuntu:
```
mernst: Please don't make unnecessary whitespace changes.  They bloat the diffs.

docs/manual/troubleshooting.tex
```diff
@@ -764,14 +764,28 @@ \section{Building from source\label{build-source}}
...
+
+\begin{Verbatim}
+brew update
+brew install caskroom/cask/brew-cask
+brew install git ant hevea mercurial librsvg unzip make
+brew cask install java
+brew cask install mactex
+\end{Verbatim}
+
+Mac OS users also need to copy \<hevea.sty> into \<\$HOME/Library/texmf/tex/latex> (you probably need to make that directory first).
```
mernst: Please make this a concrete pair of commands:
```
mkdir -p $HOME/Library/texmf/tex/latex
cp -p .../hevea.sty $HOME/Library/texmf/tex/latex
```

It's generally best to give concrete commands that can be cut-and-pasted, rather than a general description, which requires more effort from users and is prone to misinterpretation.

docs/manual/troubleshooting.tex
```diff
@@ -764,14 +764,28 @@ \section{Building from source\label{build-source}}
...
 
 \begin{Verbatim}
 sudo apt-get update
 sudo apt-get install --yes ant dia git hevea junit4 librsvg2-bin make mercurial openjdk-8-jdk \
  texlive-latex-base texlive-latex-recommended texlive-latex-extra texlive-fonts-recommended unzip
 \end{Verbatim}
 
+% Building Checker Framework from source on Mac OS
+
+If you are building on Mac OS, then the following commands install the aforementioned packages:
```
mernst: I think it would be best to remark that a Mac user should use these commands only if the user already employs homebrew to install packages.  The manual could say that instructions are forthcoming for macports users.

docs/manual/troubleshooting.tex
```diff
@@ -823,7 +837,9 @@ \subsection{Build the Type Annotations compiler\label{building-compiler}}
 (The \<JAVA\_HOME> environment
 variable might already be set, because it is needed for Ant to work.)
 
-In the bash shell, putting the following command in your \code{.bashrc} file
+\begin{enumerate}
```
mernst: Write brief introductory text saying to choose one of the following alternatives.

docs/manual/troubleshooting.tex
```diff
@@ -838,6 +854,22 @@ \subsection{Build the Type Annotations compiler\label{building-compiler}}
 %END LATEX
 
 \item
+{\bf [Mac OSX 10.5 or later]} Similarly to Ubuntu, add one of the following commands in your \code{.bash\_profile} or \code{.profile} file.
```
mernst: Drop "Simillarly to Ubuntu".  Mac users shouldn't have even read the Ubuntu instructions, so this is just confusing and might make them read more than they need to.

feliperodri:
@mernst. Could you please re-review this pull request?

PR #1388, opened 2017-07-07
Update purity and nullness annotations. Fix #1365.

### LINKED ISSUE #1365: Incorrect annotation on JDK Constructor.newInstance
In the annotated JDK (v2.1.12), the `java.lang.reflect.Constructor.newInstance` should support passing a null array. However, it doesn't currently. This should work:
```java
java.lang.reflect.Constructor c = String.class.getConstructor(String.class);
Object[] p = null;
c.newInstance(p);
```
but the Nullness Checker returns with a:

```
Error: [argument.type.incompatible] incompatible types in argument.
found : @Initialized @NonNull Object @FBCBottom @Nullable []
required: @Initialized @Nullable Object @Initialized @NonNull []
```
It looks like the JDK was annotated to support a Nullable element, but not a Nullable array (see #336).

Cheers

wmdietlGC:


checker/jdk/nullness/src/java/lang/Class.java
```diff
@@ -39,7 +39,7 @@ protected Class() {}
   public java.lang.reflect.Method[] getMethods() throws SecurityException { throw new RuntimeException("skeleton method"); }
   public java.lang.reflect.Constructor<?>[] getConstructors() throws SecurityException { throw new RuntimeException("skeleton method"); }
   public java.lang.reflect.Field getField(String a1) throws NoSuchFieldException, SecurityException { throw new RuntimeException("skeleton method"); }
-  public java.lang.reflect.Method getMethod(String a1, Class<?> @Nullable ... a2) throws NoSuchMethodException, SecurityException { throw new RuntimeException("skeleton method"); }
+  public java.lang.reflect.Method getMethod(String a1, Class<?> ... a2) throws NoSuchMethodException, SecurityException { throw new RuntimeException("skeleton method"); }
```
wmdietlGC: The method might require parameters, so to be consistent with the other reflective methods, the parameter is changed to be non-null.

checker/jdk/nullness/src/java/lang/reflect/Constructor.java
```diff
@@ -19,7 +19,7 @@
     @Pure public int hashCode() { throw new RuntimeException("skeleton method"); }
     @SideEffectFree public String toString() { throw new RuntimeException("skeleton method"); }
     public String toGenericString() { throw new RuntimeException("skeleton method"); }
-    public @NonNull T newInstance(@Nullable Object ... initargs) throws InstantiationException,IllegalAccessException,IllegalArgumentException,InvocationTargetException { throw new RuntimeException("skeleton method"); }
+    public @NonNull T newInstance(Object ... initargs) throws InstantiationException,IllegalAccessException,IllegalArgumentException,InvocationTargetException { throw new RuntimeException("skeleton method"); }
```
wmdietlGC: This is the change discussed in #1365 .

framework/src/org/checkerframework/common/basetype/flow.astub
```diff
@@ -7,6 +7,8 @@ package java.lang;
 
 public final class Class<T> {
     @Pure public String getName();
+    @Pure public java.lang.reflect.Method getMethod(String a1, Class<?> ... a2);
```
wmdietlGC: Making these two methods pure makes reflective code easier to use, as less information is lost.

PR #1390, opened 2017-07-09
Fix #1218: Incomplete handling of VarArgs in dataflow

### LINKED ISSUE #1218: Incomplete handling of VarArgs in dataflow
Consider a method that has a varargs parameter:

```java
void foo(String... args) {}
```

`javac` implicitly converts a call to `foo` with any number of string arguments into an array. For instance, `foo()` is actually `foo(new String[0])`, while `foo("bar", "baz")` is actually `foo(new String {"bar", "baz"})`. If `foo` is called with a literal array of strings, then that's what's passed.

When the CFG is constructed in `CFGBuilder`, the Checker Framework appropriately creates a new, artificial `ArrayCreationNode` to represent the new array, if one is implicitly created. However, the Checker Framework does **not** create a new AST to go along with it: instead, a `null` tree is inserted. Most of the time, this is fine. However, consider a situation where a programmer writes an annotation on the varargs parameter (borrowing the annotation from the Index Checker):

```java
void foo(String @MinLen(1) ... args) {
   // Do something with args[0]
}
```

The annotation says that the implicitly created array must have at least one element. If an actual array is passed when the function is called, it's checked correctly:
```java
String[] args = new String[0];
foo(args); // incompatible assignment, warning issued
```

But, if the function is called using varargs syntax, no warning is issued - because there is no tree to visit. The annotated type factory can't assign a type to a null tree, and the transfer function can't insert a type into the store for a node that has a null tree. So, for instance, the call `foo()` never has the type of its argument checked, and is accepted without a warning by the Checker Framework.

I think that the solution is to create an artificial tree in `CFGBuilder` using the `TreeBuilder` interface, in much the same way that artificial trees are created for enhanced for loops and other constructions that include implicit CFG nodes.

bitterfox:
@wmdietl @wmdietlGC 
Hi, Here is fix for #1218.
I've added a creation of tree for a varargs array in DFGBuilder and check the type of it with expected type in BaseTypeVisitor.

This is required to change in checker-framework-inference. Please also review corresponding changes:
https://github.com/typetools/checker-framework-inference/pull/50


framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -957,6 +958,48 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
...
+        // because type checking for elements will be done in checkArguments.
+        AnnotatedTypeMirror objectType =
+                atypeFactory.getAnnotatedType(elements.getTypeElement("java.lang.Object"));
+        lastParamAnnotatedType.setComponentType(objectType);
+        if (wrappedVarargsType instanceof AnnotatedArrayType) {
+            ((AnnotatedArrayType) wrappedVarargsType).setComponentType(objectType);
+        }
+
+        commonAssignmentCheck(
+                lastParamAnnotatedType, wrappedVarargsType, tree, "argument.type.incompatible");
```
wmdietl: Should we use a distinct error key, e.g. `varargs.argument.type.incompatible`, to make the error message easier to understand?

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -957,6 +958,48 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
     }
 
     /**
+     * A helper method to check that the array type of actual varargs is a subtype of the
+     * corresponding required varargs, and issues "argument.invalid" error for it's not a subtype of
```
wmdietl: `for` should be `if`.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -957,6 +958,48 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
...
+     * the required one.
+     *
+     * <p>Note this method requires that type checking for the each elements of varargs is executed
```
wmdietl: Does that type checking need to happen before this one? Or is this only a statement about what this method guarantees.
Try clarifying this sentence.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -957,6 +958,48 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
...
+            return;
+        }
+
+        List<AnnotatedTypeMirror> formals = invokedMethod.getParameterTypes();
+        int numFormals = formals.size();
+        int lastArgIndex = numFormals - 1;
+        AnnotatedArrayType lastParamAnnotatedType = (AnnotatedArrayType) formals.get(lastArgIndex);
+        AnnotatedTypeMirror wrappedVarargsType = atypeFactory.getAnnotatedTypeVarargsArray(tree);
+
+        if (wrappedVarargsType == null) {
```
wmdietl: When can this happen? Add a comment why it is ok to just return here.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -957,6 +958,48 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
...
+        }
+
+        // If the component type is intersection or including type variable by type inference,
+        // the comparison doesn't work well.
+        // We can consider that the both component type of actual and formal is Object,
+        // because type checking for elements will be done in checkArguments.
+        AnnotatedTypeMirror objectType =
+                atypeFactory.getAnnotatedType(elements.getTypeElement("java.lang.Object"));
+        lastParamAnnotatedType.setComponentType(objectType);
+        if (wrappedVarargsType instanceof AnnotatedArrayType) {
```
wmdietl: Use `getKind()` checks instead of instanceof checks on `AnnotatedTypeMirrors` (like for `TypeMirror`).

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1323,6 +1357,29 @@ protected void checkAndPerformFlowAnalysis(Tree tree) {
     }
 
     /**
+     * Returns the inferred value (by the org.checkerframework.dataflow analysis) for a given node.
+     */
+    private Value getInferredValueFor(Node node) {
```
wmdietl: The method is private and I don't see a caller in this class. What is it used for?

framework/tests/value/Issue1218.java
```diff
@@ -0,0 +1,132 @@
...
+    @SafeVarargs
+    final <T> void anys(T @MinLen(3) ... anys) {}
+
+    void testMethodCall() {
+        //:: error: (argument.type.incompatible)
+        strs();
+        //:: error: (argument.type.incompatible)
+        strs("");
+        strs("", "");
+        strs((String[]) null);
```
wmdietl: Can you add a comment why this call is allowed?

javacutil/src/org/checkerframework/javacutil/trees/TreeBuilder.java
```diff
@@ -652,4 +655,26 @@ public BinaryTree buildBinary(
         binary.setType((Type) type);
         return binary;
     }
+
+    /**
+     * Builds an AST Tree to create a new array with initializers.
+     *
+     * @param componentType a type for component of a new array
```
wmdietl: `component type of the new array`

wmdietl:
Can you also make sure the Dataflow Manual is updated?

bitterfox:
Hi, I've fixed this. Please confirm this again.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -957,6 +960,73 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
     }
 
     /**
+     * A helper method to check that the array type of actual varargs is a subtype of the
+     * corresponding required varargs, and issues "argument.invalid" error if it's not a subtype of
+     * the required one.
+     *
+     * <p>Note it's required that type checking for the each elements of varargs is executed by
```
wmdietl: `for the each elements of varargs` -> `for each element in varargs`
`by caller` -> `by the caller`

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -957,6 +960,73 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
...
+     * caller before or after calling this method.
+     *
+     * @see #checkArguments(List, List)
+     * @param invokedMethod the method type to be invoked
+     * @param tree method or constructor invocation tree
+     */
+    private void checkVarargs(AnnotatedExecutableType invokedMethod, Tree tree) {
```
wmdietl: Like `checkArguments` I would make this method protected, in case a subchecker needs to adapt the behavior.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -957,6 +960,73 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
...
+            if (lastArgType.getKind() == TypeKind.ARRAY
+                    && getArrayDepth(lastParamAnnotatedType)
+                            == getArrayDepth((AnnotatedArrayType) lastArgType)) {
+                return;
+            }
+        }
+
+        AnnotatedTypeMirror wrappedVarargsType = atypeFactory.getAnnotatedTypeVarargsArray(tree);
+
+        // When dataflow analysis is not enabled, it will be null and we can suppose to there is no
```
wmdietl: Remove `to`.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -957,6 +960,73 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
...
+        // annotation to be checked for generated varargs array.
+        if (wrappedVarargsType == null) {
+            return;
+        }
+
+        // If the component type is intersection or including type variable by type inference,
```
wmdietl: The comment and the code seem disconnected - I don't see a check for intersection types or type variables in the following code. Can you clarify the comment?

bitterfox:
Fixed. Please confirm this again.

wmdietl:
I've merged the downstream checker-framework-inference change, so this should now be clean.

PR #1392, opened 2017-07-09
Fix #536: Poor Nullness Checker performance when code has many assignments

### LINKED ISSUE #536: Poor Nullness Checker performance when code has many assignments
Type-checking file checker-framework/checker/tests/nullness/AssignmentPerformanceTest.java takes about 10 minutes even though the only code in it is 150 assignments.

bitterfox:
Hi @wmdietl @wmdietlGC 
Here is fix for #536 
I've tried to fix this by caching.

This is caused by running dataflow analysis(CFTransfer) for O(n^2) by calling getStoreBefore for each of nodes( where n is a number of nodes).

By this change, compiling of "checker/tests/nullness/AssignmentPerformanceTest.java" is done in a few ten seconds, even though it took about 10 or more minutes before this change.

Before:
bitter_fox@bitterfox-MacBookPro% time javac -processor NullnessChecker checker/tests/nullness/AssignmentPerformanceTest.java
javac -processor NullnessChecker   585.82s user 1.02s system 102% cpu 9:33.23 total

After:
bitter_fox@bitterfox-MacBookPro% time javac -processor NullnessChecker checker/tests/nullness/AssignmentPerformanceTest.java
javac -processor NullnessChecker   18.44s user 0.23s system 171% cpu 10.876 total


checker/jtreg/nullness/AssignmentPerformanceTest.java
```diff
@@ -1,6 +1,9 @@
-// @skip-test
-// Test case for issue #536: https://github.com/typetools/checker-framework/issues/536
-
+/*
+ * @test
+ * @summary Test case for issue #536: https://github.com/typetools/checker-framework/issues/536
+ *
+ * @compile/timeout=60 -J-Xmx256M -XDrawDiagnostics -Xlint:unchecked -processor org.checkerframework.checker.nullness.NullnessChecker -AprintErrorStack -Alint AssignmentPerformanceTest.java
```
bitterfox: To confirm my change doesn't make terrible memory usage, I've added `-Xmx256M` option for JVM.

wmdietl: The `@test` style only works for jtreg style tests. You would need to move the test to a different directory for this.
I'm not sure we need to do this, so I would undo these changes and just remove the `@skip-test`.

dataflow/src/org/checkerframework/dataflow/analysis/AnalysisResult.java
```diff
@@ -142,7 +148,16 @@ protected S runAnalysisFor(Node node, boolean before) {
         if (transferInput == null) {
             return null;
         }
-        return runAnalysisFor(node, before, transferInput);
+        IdentityHashMap<Node, TransferResult<A, S>> cache = this.cache.get(transferInput);
```
wmdietl: Use a different name for the local variable to make the code easier to read.

dataflow/src/org/checkerframework/dataflow/analysis/AnalysisResult.java
```diff
@@ -142,7 +148,16 @@ protected S runAnalysisFor(Node node, boolean before) {
...
+        if (cache == null) {
+            this.cache.put(transferInput, cache = new IdentityHashMap<>());
```
wmdietl: Don't assign to a variable as an argument to something else.

dataflow/src/org/checkerframework/dataflow/analysis/AnalysisResult.java
```diff
@@ -142,7 +148,16 @@ protected S runAnalysisFor(Node node, boolean before) {
...
+        }
+        return runAnalysisFor(node, before, transferInput, cache);
+    }
+
+    public static <A extends AbstractValue<A>, S extends Store<S>> S runAnalysisFor(
+            Node node, boolean before, TransferInput<A, S> transferInput) {
+        return runAnalysisFor(node, before, transferInput, null);
```
wmdietl: How many places call this method? Can we instead pass the right cache in those locations?

dataflow/src/org/checkerframework/dataflow/analysis/AnalysisResult.java
```diff
@@ -37,6 +37,10 @@
     /** The stores before every method call. */
     protected final IdentityHashMap<Block, TransferInput<A, S>> stores;
 
+    /** The results for each node. */
+    protected IdentityHashMap<TransferInput<A, S>, IdentityHashMap<Node, TransferResult<A, S>>>
```
wmdietl: Can this field be final?

checker/jtreg/nullness/AssignmentPerformanceTest.java
bitterfox: I think this should be needed to be a jtreg test to confirm it'll be done until the timeout.
Please note that this test case is already moved to jtreg directory.

wmdietl: Yes, sorry for missing that the file was moved. This file looks good.

bitterfox:
Hi,
I've fixed for the comments.

I have some questions.

1. Should I remove `runAnalysisFor(Node node, boolean before, TransferInput<A, S> transferInput)`?
As your suggestion, I've fixed all of the callers of this method to pass the cache.
So in `checker framework`, nobody calls this method.
Regarding backward compatibility and that this method is public, I'm not sure whether I should remove or not is preferred.

Do you have a suggestion?

2. Should I respect `shouldCache` and `atfCacheSize` option for this cache at everywhere and make it LRU cache?
I've realized the question by investigating #1438.
Do you have any suggestion

dataflow/src/org/checkerframework/dataflow/analysis/AnalysisResult.java
```diff
@@ -37,6 +37,11 @@
     /** The stores before every method call. */
     protected final IdentityHashMap<Block, TransferInput<A, S>> stores;
 
+    /** The results for each node. */
```
wmdietl: Please expand this documentation and state what the different keys and values to the maps are and what they are caching.

dataflow/src/org/checkerframework/dataflow/analysis/AnalysisResult.java
```diff
@@ -142,7 +149,29 @@ protected S runAnalysisFor(Node node, boolean before) {
         if (transferInput == null) {
             return null;
         }
-        return runAnalysisFor(node, before, transferInput);
+        return runAnalysisForWithCacheLookupMap(node, before, transferInput, cacheLookupMap);
+    }
+
+    public static <A extends AbstractValue<A>, S extends Store<S>>
+            S runAnalysisForWithCacheLookupMap(
```
wmdietl: Do we really want the `WithCacheLookupMap` in the name? Isn't the version with cache the one that most people will use?

dataflow/src/org/checkerframework/dataflow/analysis/AnalysisResult.java
```diff
@@ -142,7 +149,29 @@ protected S runAnalysisFor(Node node, boolean before) {
         if (transferInput == null) {
             return null;
         }
-        return runAnalysisFor(node, before, transferInput);
+        return runAnalysisForWithCacheLookupMap(node, before, transferInput, cacheLookupMap);
+    }
+
+    public static <A extends AbstractValue<A>, S extends Store<S>>
```
wmdietl: Please add javadoc to this public method, in particular highlighting how to choose between the different `runAnalysisFor` methods.

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -165,6 +165,11 @@
     /** An empty store. */
     private Store emptyStore;
 
+    protected final IdentityHashMap<
```
wmdietl: Provide some documentation and explain what is different to the cache in AnalysisResult.

framework/src/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1099,6 +1107,7 @@ protected void analyze(
         if (emptyStore == null) {
             emptyStore = newAnalysis.createEmptyStore(transfer.usesSequentialSemantics());
         }
+        cacheLookupMap.clear();
```
wmdietl: Should the cache also be cleared in `setRoot`? Most other maps are cleared there.

wmdietl:
Regarding 2: definitely make sure the cache is LRU.
Reusing `shouldCache` and `atfCacheSize` sounds good to me. @smillst what do you think?

bitterfox:
Hi, I've fixed. Please review this again.

dataflow/src/org/checkerframework/dataflow/analysis/AnalysisResult.java
```diff
@@ -47,6 +55,7 @@ public AnalysisResult(
         this.treeLookup = new IdentityHashMap<>(treeLookup);
         this.stores = stores;
         this.finalLocalValues = finalLocalValues;
+        analysisCaches = new IdentityHashMap<>();
```
wmdietl: For consistency, write `this.analysisCaches = ...`

wmdietl: Note how the three constructors are inconsistent... I would just make each one locally consistent.

dataflow/src/org/checkerframework/dataflow/analysis/AnalysisResult.java
```diff
@@ -55,6 +64,17 @@ public AnalysisResult() {
...
+    }
+
+    /** Initialize empty result with specified cache. */
+    public AnalysisResult(
+            Map<TransferInput<A, S>, IdentityHashMap<Node, TransferResult<A, S>>> analysisCaches) {
+        nodeValues = new IdentityHashMap<>();
+        treeLookup = new IdentityHashMap<>();
+        stores = new IdentityHashMap<>();
+        finalLocalValues = new HashMap<>();
+        this.analysisCaches = analysisCaches;
```
wmdietl: Ether add `this` to the others or remove here.

bitterfox:
Hi, I've fixed. Please check this again.

smillst:
Once the test pass, this is ok to merge.

PR #1393, opened 2017-07-10
Fix #1386: copy values when getting alias annotation from canonical annotation

### LINKED ISSUE #1386: Aliased type and declaration annotations should handle annotation values
Motivated by discussion at https://github.com/typetools/checker-framework/pull/1383/files/55d89e985d12346c727dc8ceb3bfd8508983c8f4#r126213420

Currently, each type system needs to take care of copying annotation values from the alias to the original, e.g. as done in `UpperBoundAnnotatedTypeFactory.aliasedAnnotation` https://github.com/typetools/checker-framework/blob/master/checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java#L241

It would be nice if there were a default implementation that copied over annotation values 1:1, which seems like the desired default behavior.
If attribute names need to be changed, it should be easy for a type system to customize this behavior.

### LINKED ISSUE #1389: Crash when use Value Checker on `android.support.annotation.IntRange`
The follow code should be handled by value checker aliasing, but crash
```java
import android.support.annotation.*;
class AliasedAnnotations {
    void useIntRangeAnnotation() {
        IntRange(from=0, to=10) int i = 12;
    }
}
```
Seems to be regressed by this commit e125e19 when changing `ValueAnnotatedTypeFactory#getIntRange` to `ValueAnnotatedTypeFactory#getRange`.

A fix for #1386 should address this issue more fundamentally

JasonMrX:
Also fix #1389


framework/tests/value/Alias.java
```diff
@@ -0,0 +1,10 @@
+import android.support.annotation.*;
```
wmdietl: Just import the single annotation you're using.

framework/src/org/checkerframework/framework/util/AnnotationBuilder.java
```diff
@@ -147,6 +148,25 @@ public String toString() {
...
+        for (Map.Entry<? extends ExecutableElement, ? extends AnnotationValue> entry :
+                annotation.getElementValues().entrySet()) {
+            Name elemName = entry.getKey().getSimpleName();
+            for (ExecutableElement elt :
+                    ElementFilter.methodsIn(annotationElt.getEnclosedElements())) {
+                if (elt.getSimpleName().contentEquals(elemName)) {
+                    elementValues.put(elt, entry.getValue());
+                    break;
+                }
+            }
```
wmdietl: I agree with Martin that some error handling would be better than silently ignoring attributes.
It should be enough to keep track whether for each outer loop element you found an inner loop element; if not, raise an error. Some javadoc above would also need updating. The error message should say what to do in that case - i.e. write your own addAliasedAnnotation implementation.

framework/src/org/checkerframework/framework/util/AnnotationBuilder.java
```diff
@@ -147,6 +148,25 @@ public String toString() {
         };
     }
 
+    /**
+     * Copies every element value from the given annotation to the annotation to be built if the
+     * element name match
```
wmdietl: `@param` documentation.

framework/src/org/checkerframework/framework/util/AnnotationBuilder.java
```diff
@@ -147,6 +148,25 @@ public String toString() {
...
+     */
+    public AnnotationBuilder copyElementValuesFromAnnotation(AnnotationMirror annotation) {
+        for (Map.Entry<? extends ExecutableElement, ? extends AnnotationValue> entry :
```
wmdietl: I would pick clearer names for the variables to distinguish better which ones are input and which ones are modified.

JasonMrX:
Thanks for the feedback @kelloggm and @wmdietl.

I silently ignore those values that are not copied over on purpose. Originally I was trying to tackle the issue with the `android.annotation.support.@IntRange` alias in Value Checker, which has the exact same set of parameters as the checker framework `@IntRange`. So I did what you guys are suggesting -- fail the checker if the element name doesn't match. However, the Index Checker is broken because some annotations are aliased with annotations that do not have the same set of parameters on purpose. For example, in the Lower Bound Checker, the `@IndexFor` is an alias of `@NonNegative`, but we certainly don't want to copy the value. Therefore I silently ignore the value if the canonical one doesn't contain it. 

I could change it as you suggested but in that case the Index Checker would still need to keep the special handling for aliases, which kind of trivialize this fix (`@IntRange` can still be benefited though).

wmdietlGC:
I think we should only do the mapping automatically if for every source attribute there is a destination attribute with the same name. The destination can have more attributes (but will need to have default values to work).
Dropping attributes from the source seems like a simple error.

In the future, we can provide a mapping argument, that describes how source to target attributes should be mapped, e.g. if renaming of attributes is required. A declarative way might also be useful.

This does limit the usefulness for the Index Checker aliases at the moment, but e.g. the `@Covariant` annotation in #1383 could already use this feature.

@JasonMrX does this sound good to you?

JasonMrX:
@wmdietlGC I see. I've updated the behaviour as requested, however, I put the error raising part at `annotatedTypeFactory` instead of within the `annotationBuilder` which I think should not be aware of what's going on with the alias.

JasonMrX:
I just realised from the failures that this doesn't work for some cases -- in Lower Bound Checker, `@IndexFor(values)` is an alias of `@NonNegative` which doesn't not take any arguments. In this case, error would be raised during `aliasedAnnotation` because we cannot copy the `values` to the `@NonNegative`.

Two options here I can come up with:
1. override `aliasedAnnotation` in the Lower Bound Checker (further limit the use of this change)
2. add a `suppressValueCopy` option to `aliasedAnnotation`

Any ideas?

JasonMrX:
@wmdietlGC quite a few checkers have aliased annotation with the similar pattern -- an annotation with annotation with attributes is an alias to an annotation without attribute (e.g. `@IndexFor(values)` => `@NonNull`, `@NotNull(message)` => `@NonNull`). So if we don't have an attribute mapping strategy that can handling mismatching atrributes, it's difficult to make this feature useful (in terms of avoiding code duplication)

wmdietl:
Hi @JasonMrX,
sorry for dropping the ball on this issue. Let's wrap it up and get this merged.
How hard would it be to add an argument to `addAliasedAnnotation` that gives the attribute names that can be dropped. So for `@NotNull(message) => @NonNull` one would have to write `addAliasedAnnotation(javax.validation.constraints.NotNull.class, NONNULL, "groups", "message", "payload");` listing the three attributes that can be ignored. If a new attribute is added, an error would be raised if it isn't mapped. Maybe that granularity isn't needed and we just need a boolean to turn this check on or off?
Another option would be to add meta-annotations on the aliased annotation which describes how the mapping should be performed, which would be most flexible.

mernst:
@JasonMrX , there is a failing merge -- just one small part of one file.  Could you address that, and then the other suggestions so we can merge this?
Thanks!

JasonMrX:
Wasn't following for a while. Will take a look during the weekend. Thanks.

mernst:
@JasonMrX Thanks for merging master.  Now there are test failures of the form:
> Tried to copy attributes from the aliased annotation @javax.validation.constraints.NotNull(message="hi") to the canonical annotation @org.checkerframework.checker.nullness.qual.NonNull, however, the alias contains attribute(s) that does not exist in the canonical. You should try to override the aliasedAnnotation in the specific checker for special cases.

mernst:
@JasonMrX The Travis build is now passing.  Is the pull request ready to be re-assigned back to reviewers?

JasonMrX:
Thanks @mernst for the Javadoc fix. Yes, I think it's ready to be reviewed. Assigned back to @wmdietl.

JasonMrX:
@wmdietl I think the new set of changes can achieve what you proposed:
1. By default, the elements would not be copied over when building the alias
2. If user explicitly inform the type factory to copy over the elements, an error would be raised if elements do not match unless specially specified as in 3.
3. User can also provide a list of elements that can be safely ignored when copying over the elements, so that the checker wouldn't raise error if the alias annotation does not contain these elements.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -2411,18 +2432,46 @@ protected void addAliasedAnnotation(String canonicalName, AnnotationMirror type)
...
+        aliasesCopyElements.put(canonicalName, copyElements);
+        aliasesIgnorableElements.put(canonicalName, ignorableElements);
+    }
+
+    /**
      * Returns the canonical annotation for the passed annotation if it is an alias of a canonical
      * one in the framework. If it is not an alias, the method returns null.
      *
-     * <p>Returns an aliased type of the current one
+     * <p>Returns an aliased type of the current one. The attributes contained by the passed
```
wmdietl: The `current one` is strange. And this method returns the canonical annotation for an alias, not an alias.
How about removing this `<p>` block, as the sentence before is much clearer? And then update the `@return` to make clear that the canonical annotation is returned.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -2411,18 +2432,46 @@ protected void addAliasedAnnotation(String canonicalName, AnnotationMirror type)
...
      * @return the alias or null if none exists
      */
     public /*@Nullable*/ AnnotationMirror aliasedAnnotation(AnnotationMirror a) {
         TypeElement elem = (TypeElement) a.getAnnotationType().asElement();
         String qualName = elem.getQualifiedName().toString();
-        return aliases.get(qualName);
+        AnnotationMirror frameworkAnno = aliases.get(qualName);
+        if (frameworkAnno != null && a.getElementValues().size() > 0) {
+            AnnotationBuilder builder = new AnnotationBuilder(processingEnv, frameworkAnno);
+            if (aliasesCopyElements.getOrDefault(qualName, false)) {
```
wmdietl: As you only care whether or not to copy elements, you could make `aliasesCopyElements` a `Set` and check for containment here. That should be a bit simpler to check and store.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -2411,18 +2432,46 @@ protected void addAliasedAnnotation(String canonicalName, AnnotationMirror type)
...
+     * AnnotationMirror, boolean, String...)}
      *
      * @param a the qualifier to check for an alias
      * @return the alias or null if none exists
      */
     public /*@Nullable*/ AnnotationMirror aliasedAnnotation(AnnotationMirror a) {
         TypeElement elem = (TypeElement) a.getAnnotationType().asElement();
         String qualName = elem.getQualifiedName().toString();
-        return aliases.get(qualName);
+        AnnotationMirror frameworkAnno = aliases.get(qualName);
```
wmdietl: To follow the other naming, maybe call this `canonicalAnno`?

javacutil/src/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -153,6 +154,39 @@ public AnnotationMirror build() {
         return new CheckerFrameworkAnnotationMirror(annotationType, elementValues);
     }
 
+    /**
+     * Copies every element value from the given annotation. If an element in the given annotation
+     * doesn't exist in the annotation to be built, an error would be raised unless the element is
```
wmdietl: `would be` -> `is`

javacutil/src/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -153,6 +154,39 @@ public AnnotationMirror build() {
...
+     * specified in {@code ignorableElements}.
+     *
+     * @param valueHolder the annotation that holds the values to be copied
+     * @param ignorableElements the elements that can be safely dropped
+     */
+    public AnnotationBuilder copyElementValuesFromAnnotation(
```
wmdietl: Return `void`? You don't say anything about the returned value, so I would have assumed you create a new AnnotationBuilder. But this is directly modifying the current object. And you don't need this as part of some fluid API, so just make it void. Or document what is returned.

javacutil/src/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -153,6 +154,39 @@ public AnnotationMirror build() {
...
+            AnnotationMirror valueHolder, String... ignorableElements) {
+        for (Map.Entry<? extends ExecutableElement, ? extends AnnotationValue> eltValToCopy :
+                valueHolder.getElementValues().entrySet()) {
+            Name eltNameToCopy = eltValToCopy.getKey().getSimpleName();
+            if (Arrays.asList(ignorableElements).contains(eltNameToCopy.toString())) {
```
wmdietl: Move the `Arrays.asList` out of the loop.

javacutil/src/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -153,6 +154,39 @@ public AnnotationMirror build() {
...
+                continue;
+            }
+            boolean hasElement = false;
+            for (ExecutableElement elt :
+                    ElementFilter.methodsIn(annotationElt.getEnclosedElements())) {
```
wmdietl: Move the `ElementFilter...` out of the loop.

javacutil/src/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -153,6 +154,39 @@ public AnnotationMirror build() {
...
+    public AnnotationBuilder copyElementValuesFromAnnotation(
+            AnnotationMirror valueHolder, String... ignorableElements) {
+        for (Map.Entry<? extends ExecutableElement, ? extends AnnotationValue> eltValToCopy :
+                valueHolder.getElementValues().entrySet()) {
+            Name eltNameToCopy = eltValToCopy.getKey().getSimpleName();
+            if (Arrays.asList(ignorableElements).contains(eltNameToCopy.toString())) {
+                continue;
+            }
+            boolean hasElement = false;
+            for (ExecutableElement elt :
```
wmdietl: Can't you simplify the rest of this by calling `AnnotationBuilder setValue(CharSequence key, Object value)`? That method finds the right method and adds the value, additionally ensures that it's of the right type, and outputs the same error message.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -2397,11 +2409,20 @@ public boolean isSupportedQualifier(/*@Nullable*/ AnnotationMirror a) {
...
+    protected void addAliasedAnnotation(
+            Class<?> alias,
+            AnnotationMirror type,
+            boolean copyElements,
+            String... optionalElements) {
+        addAliasedAnnotation(alias.getCanonicalName(), type, copyElements, optionalElements);
+    }
+
     /**
      * Add the annotation whose canonical name is given by {@code canonicalName} as an alias for the
```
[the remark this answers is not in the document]
JasonMrX: I found it confusing too. That's why I followed the trend. I agree with your interpretation of "alias" and "canonical". Will modify accordingly.

javacutil/src/org/checkerframework/javacutil/AnnotationBuilder.java
JasonMrX: It's been too long that I forgot why I did this. I tried to use `setValue` but the `checkSubtype` method can't be passed.  I was trying to replace this chunk of code with the following:
```java
setValue(eltNameToCopy, (Object) eltValToCopy.getValue());
```
It keeps saying that `given value differs from expected; found: com.sun.tools.javac.code.Attribute.Array; expected: java.lang.String[]`

Do you have any insight on what happens here?

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -2411,18 +2432,46 @@ protected void addAliasedAnnotation(String canonicalName, AnnotationMirror type)
     }
 
     /**
+     * Add the annotation whose canonical name is given by {@code canonicalName} as an alias for the
+     * annotation {@code type}. An {@code copyElements} flag is passed in to indicate whether we
+     * want to copy the elements over when getting the alias from the canonical one. An optional
+     * {@code ignorableElements} contains a list of elements that can be safely dropped when the
+     * elements are being copied over.
+     */
+    protected void addAliasedAnnotation(
```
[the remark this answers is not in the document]
JasonMrX: The `String` version is provided because in some case the alias class cannot be referenced directly (e.g., the usage of the 2-param `String` version of this method). I still keep this version here since the implementation cannot be moved to the other version without code duplication (cannot get class from string). I've updated the javadoc to explain this)

javacutil/src/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -153,6 +156,28 @@ public AnnotationMirror build() {
...
+    public void copyElementValuesFromAnnotation(
+            AnnotationMirror valueHolder, String... ignorableElements) {
+        Set<String> ignorableElementsSet = new HashSet<>(Arrays.asList(ignorableElements));
+        for (Map.Entry<? extends ExecutableElement, ? extends AnnotationValue> eltValToCopy :
+                valueHolder.getElementValues().entrySet()) {
+            Name eltNameToCopy = eltValToCopy.getKey().getSimpleName();
+            if (ignorableElementsSet.contains(eltNameToCopy.toString())) {
+                continue;
+            }
+            elementValues.put(findElement(eltNameToCopy), eltValToCopy.getValue());
```
JasonMrX: I omit the sub type check here as I couldn't find a way to make it pass by passing the key value pair into `setValue` method. Any suggestion here would be great.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -249,6 +249,15 @@
      */
     private final Map<String, AnnotationMirror> aliases = new HashMap<String, AnnotationMirror>();
 
+    /** Set contains class names of aliases whose elements need to be copied over. */
+    private final Set<String> aliasesCopyElements = new HashSet<>();
+
+    /**
+     * Map from class name (canonical name) of an annotation, to the ignorable elements that the
```
wmdietl: Shouldn't the `canonical name` be `alias name`? Or to make it more symmetric to the other documentation: `Map from class names of aliases of an annotation to the ignorable...`?

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -249,6 +249,15 @@
...
+     * framework can safely drop when copying over the elements.
+     */
+    private final Map<String, String[]> aliasesIgnorableElements = new HashMap<>();
+
     /**
      * A map from the class name (canonical name) of an annotation to the set of class names
      * (canonical names) for annotations with the same meaning (i.e., aliases), as well as the
```
wmdietl: If you also want to fix an existing inconsistency, I would remove the second `canonical names` here, as it is an alias name, not a canonical name.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -2397,32 +2406,88 @@ public boolean isSupportedQualifier(/*@Nullable*/ AnnotationMirror a) {
...
-    /** Add the annotation {@code clazz} as an alias for the annotation {@code type}. */
-    protected void addAliasedAnnotation(Class<?> alias, AnnotationMirror type) {
-        addAliasedAnnotation(alias.getCanonicalName(), type);
+    /** Adds the annotation {@code aliasClass} as an alias for the annotation {@code type}. */
+    protected void addAliasedAnnotation(Class<?> aliasClass, AnnotationMirror type) {
+        addAliasedAnnotation(aliasClass.getCanonicalName(), type);
+    }
+
+    /**
+     * Adds the annotation whose canonical name is given by {@code aliasName} as an alias for the
```
wmdietl: Change the `canonical name` to `fully-qualified name`? Isn't it the fully-qualified class name of the alias that is given as the first argument?

javacutil/src/org/checkerframework/javacutil/AnnotationBuilder.java
wmdietl: The code looks much more readable now. I don't think a comment about `setValue` is necessary any more, as you don't duplicate that logic.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
JasonMrX: I think the `canonical` here means `#Class.getCanonicalName`. To avoid confusion I will remove it anyway.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
JasonMrX: same reason here as above

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -244,15 +244,23 @@
     protected final BaseTypeChecker checker;
 
     /**
-     * Map from class name (canonical name) of an annotation, to the annotation in the Checker
-     * Framework that will be used in its place.
+     * Map from class names of aliases, to the annotations in the Checker Framework that will be
```
kelloggm: Does class name here mean fully-qualified name? There's a good explanation of the various kinds of javac names here: https://checkerframework.org/manual/#signature-checker

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -244,15 +244,23 @@
...
+     * Map from class names of aliases, to the ignorable elements that the framework can safely drop
+     * when copying over the elements.
+     */
+    private final Map<String, String[]> aliasesIgnorableElements = new HashMap<>();
+
     /**
-     * A map from the class name (canonical name) of an annotation to the set of class names
-     * (canonical names) for annotations with the same meaning (i.e., aliases), as well as the
-     * annotation mirror that should be used.
+     * A map from the class name (canonical name) of an annotation to the set of class names for
```
kelloggm: Same question as above. What does "canonical name" mean?

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -2397,32 +2405,88 @@ public boolean isSupportedQualifier(/*@Nullable*/ AnnotationMirror a) {
...
 
     /**
      * Returns the canonical annotation for the passed annotation if it is an alias of a canonical
      * one in the framework. If it is not an alias, the method returns null.
      *
-     * <p>Returns an aliased type of the current one
-     *
      * @param a the qualifier to check for an alias
-     * @return the alias or null if none exists
+     * @return the canonical annotation or null if none exists
```
kelloggm: Same question about what "canonical name" means.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -84,14 +84,14 @@ public UpperBoundAnnotatedTypeFactory(BaseTypeChecker checker) {
...
-        addAliasedAnnotation(LengthOf.class, createLTEqLengthOfAnnotation());
-        addAliasedAnnotation(PolyAll.class, POLY);
-        addAliasedAnnotation(PolyIndex.class, POLY);
+        addAliasedAnnotation(IndexFor.class, createLTLengthOfAnnotation(), true);
+        addAliasedAnnotation(IndexOrLow.class, createLTLengthOfAnnotation(), true);
+        addAliasedAnnotation(IndexOrHigh.class, createLTEqLengthOfAnnotation(), true);
+        addAliasedAnnotation(SearchIndexFor.class, createLTLengthOfAnnotation(), true);
+        addAliasedAnnotation(NegativeIndexFor.class, createLTLengthOfAnnotation(), true);
+        addAliasedAnnotation(LengthOf.class, createLTEqLengthOfAnnotation(), true);
+        addAliasedAnnotation(PolyAll.class, POLY, true);
```
kelloggm: Why do polymorphic qualifiers need to copy values? If they do, this change should also happen in every other checker that uses polymorphic qualifiers (almost all checkers). If not, then it shouldn't happen here.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
JasonMrX: I think @wmdietl and I shared the same understanding here on the names:
* Alias means the annotation to be replaced when running the checker framework. For example, `@NotNull` is an alias of `@NonNull` in checker framework. `@IndexFor` is an alias of the combination of `@NonNegative` and `@LTLengthOf`. 
* Canonical means the annotation to replace with when running the checker framework. For example `@NonNull` (`@org.checkerframework.checker.nullness.qual.NonNull`) is the canonical annotation for all kinds of `@NonNull`s in other packages.

Correct me if my understanding is incorrect.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
JasonMrX: See above

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
JasonMrX: See above

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
JasonMrX: My bad. Will remove in next iteration. It doesn't do any harm here thought, since there is no annotation arguments to copy.

JasonMrX:
Sorry for delaying this for so long. I will pick it up ASAP. My Linux environment is down and I need sometime to set it up again.

As for your last comment, I will try to add more test cases to verify the behaviors and provides explanations. I've already added a test for `@IntRange` since one of the reason for this PR is to fix this issue. I will add more tests for the Index checker since it is also using this feature.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
mernst: This discussion suggests that  "canonical name" is a misleading term.  "Name of the canonical annotation" would be more accurate.
Perhaps it's worth a phrase to define "canonical annotation" -- I'm not sure about that.

Even after clarifying that issue, the documantation should indicate the string format of the class name:  whether it is a fully-qualified name, or a simple name, or something else.  That would address the code review comment.

JasonMrX:
I've tried to make the javadoc clearer according to @kelloggm's and @mernst's suggestions. Any further suggestions would be appreciated. 

For the tests/manual part @kelloggm mentioned, could you please provide more details on what you think? In my humble opinion, a crystal clear javadoc for all aliase related methods should be more beneficial for understanding the underlying logics.

kelloggm:
Putting the documentation in the Javadoc instead of the manual is fine. I'm still concerned that the failure case here isn't well documented, though. I don't see anything in the Javadoc about what happens if an annotation has a value that isn't in the set of values to be safely dropped. Can make sure that the Javadoc for the aliasing methods mention this?

My concern is someone developing a new checker who wants to alias an annotation with an annotation with different values. How will they figure out what they did wrong?

JasonMrX:
Hi @kelloggm, I put more detailed explanation on how this "ignorable elements" work (and what happen if fail) in the javadocs of the main methods. Would you mind taking another look and see if this is what you expected?

Thanks,

Jason

kelloggm:
Yes, those changes make the javadoc much more clear. Thank you.

PR #1396, opened 2017-07-12
Fix #753: Flow Expression parser: parse method arguments containing dots

### LINKED ISSUE #753: Flow Expression parser: parse method arguments containing dots
Currently, if given the flow expression
`"this.getField(this.field.field).field"`
the flow expression parser does not recognize the method call and attempts to parse the following two substrings
`"getField(this"`
`"field.field).field"`
The flow expression parser needs to be fixed so that subexpressions representing method actual parameters can contain dots.

bitterfox:
@wmdietlGC 
I've refined the parser of flow expression to accept a field access with a method invocation.


bitterfox:
Hi Suzanne,
Thank you for your kindly review and sorry for late reply.
I'm now addressing to your comments.
I've noticed similar issues exist for other syntaxes like array access or something like it.
Now I'm also addressing to it and will update the PR soon.
I'll reply when I updated the PR. So please review this again at that time.

bitterfox:
Hi @smillst, thank you for your review.
I've just updated the PR. Please review this again.
As I wrote above, I also fix similar issues for other syntaxes.

PR #1403, opened 2017-07-18
- Fixed the bash script in pre-commit. Removed quotation marks that a…

Bohdankm22:
…re not needed.


Bohdankm22:
I experienced an issue in the pre-commit bash file while tried to commit to the Checker Framework.
Hence I fixed it by removing redundant quotation marks.

mernst:
Thanks, and sorry for my typo.

PR #1409, opened 2017-07-18
Add nullness annotation for JDK method Enum.name().  Fixes #1402.

### LINKED ISSUE #1402: name() call from enum constructor is wrongly assumed to be @UnderInitialization
Suppose we have an enum class with a constructor:

```java
public enum MyEnum {
  TEST_ONE("abc"),
  TEST_TWO("def");
  
  private final String newName;

  MyEnum(String customData) {
    this.newName = DoSomething.withName(name());
  }
}
```

Now, nullness checker complains that the call to `name()` is not initialized:

```
error: [method.invocation.invalid] call to name() not allowed on the given receiver.
        this.newName = DoSomething.withName(name());
                                                ^
  found   : @UnderInitialization(java.lang.Enum<MyEnum>.class) @NonNull Enum
  required: @Initialized @NonNull Enum
```

This call being made inside an enum constructor, there is not much we can do to avoid this error other than marking the whole constructor to suppress warning. It will be great if the checker framework can handle cases involving enum constructors.

mernst:


wmdietl:
Test failures are because of the change to the annotated JDK and the old version is downloaded.

smillst:
Passed with jdk built with this branch. https://travis-ci.org/typetools/checker-framework/builds/255375985?utm_source=github_status&utm_medium=notification

Issue #1411, opened 2017-07-19
AssertionError in substituteMethodReturnType


panacekcz:
This error appeared when testing a modified index checker on plume-lib, in [MultiVersionControl.java](https://github.com/mernst/plume-lib-check-index-new/blob/3f82f73faaf4c01bee92944a1ea16d8e854bf2fb/java/src/plume/MultiVersionControl.java#L887). This report also uses index checker to trigger this error through an inferred `@SameLen` qualifier in the same way, but the error seems to be in the framework.
Minimalized test case:
```java
import org.checkerframework.dataflow.qual.Pure;
interface IGeneric<V>{
	@Pure public V get();
}
interface IConcrete extends IGeneric<char[]>{
}
public class SubstituteCrash {
	static void m(IConcrete ic) {
		char[] val = ic.get();
	}
}
```
Run the checker as:
```
checker/bin/javac -processor org.checkerframework.checker.index.IndexChecker -AprintErrorStack  SubstituteCrash.java
```
Error output:
```
error: SourceChecker.typeProcess: unexpected Throwable (AssertionError) while processing SubstituteCrash.java
  Compilation unit: SubstituteCrash.java
  Exception: java.lang.AssertionError; Stack trace: org.checkerframework.javacutil.InternalUtils.substituteMethodReturnType(InternalUtils.java:392)
  org.checkerframework.framework.util.FlowExpressionParseUtil.parseMethod(FlowExpressionParseUtil.java:541)
  org.checkerframework.framework.util.FlowExpressionParseUtil.parseHelper(FlowExpressionParseUtil.java:166)
  org.checkerframework.framework.util.FlowExpressionParseUtil.parseMemberSelect(FlowExpressionParseUtil.java:219)
  org.checkerframework.framework.util.FlowExpressionParseUtil.parseHelper(FlowExpressionParseUtil.java:168)
  org.checkerframework.framework.util.FlowExpressionParseUtil.parse(FlowExpressionParseUtil.java:131)
  org.checkerframework.framework.type.GenericAnnotatedTypeFactory.getReceiverFromJavaExpressionString(GenericAnnotatedTypeFactory.java:757)
  org.checkerframework.checker.index.samelen.SameLenTransfer.propagateCombinedSameLen(SameLenTransfer.java:134)
  org.checkerframework.checker.index.samelen.SameLenTransfer.visitAssignment(SameLenTransfer.java:110)
[...]

PR #1413, opened 2017-07-19
Support strings in the Index checker

panacekcz:
This is a second part of the Bounded-size strings GSoC project. I submit this PR to start a review process. Could you please review the changes, @kelloggm?

Summary of changes:
- in the index checker, treat `string.length()` the same as `array.length`
- enable and change annotations on string indices in the index JDK
- enable annotations on `CharSequence` indices in the index JDK. Only preconditions are enabled, in order to avoid getting index qualifiers on mutable sequences such as `StringBuilder`.
- refer to "sequences" instead of "arrays" in the manual, comments, and variable and method names

Remaining issues:
- test `CheckAgainstNegativeOne` has been disabled, see panacekcz#4
- ~~this PR ignores the fact that `String` implements `CharSequence`. See panacekcz#3 for details and possible approaches. All comments and questions are welcome there. The current situation corresponds to solution 2.~~
- ~~`.travis.yml` is edited to build the jdk~~
- ~~should not be merged before #1352~~
- `build.xml` is edited to download the annotated jdk built from this code


panacekcz:
Junit tests fail on Travis, because one test requires the changes done in #1352, which are not included in this PR.
They pass in a [branch](https://github.com/panacekcz/checker-framework/tree/bounded-strings-index-dev) which combines this and #1352: https://travis-ci.org/panacekcz/checker-framework/builds/255544911

panacekcz:
I have tested the changes on on [plume-lib](https://github.com/mernst/plume-lib-check-index-new/tree/3f82f73faaf4c01bee92944a1ea16d8e854bf2fb). I created a version with suppressed warnings and links to issues: https://github.com/mernst/plume-lib-check-index-new/compare/master...panacekcz:master

The issues encountered are:
- #1411: Index checker crash (1 occurence).
- kelloggm#165: Upper bound not inferred for a string literal (1 occurence).
- kelloggm#56: Calling `substring` after `startsWith` or `endsWith` returns `true` (12 occurences).
- panacekcz#4: Calling `substring` with an index returned from `indexOf` plus the length of the string (6 occurences).
- panacekcz#5: Calling `substring` with end `offset+length` (1 occurence). Calling `write(String,int,int)` with offset `0` (1 occurence). 
- kelloggm#164: Warning on clone method, not related to strings (2 occurences).
- 1 [known bug](https://github.com/mernst/plume-lib-check-index-new/blob/3f82f73faaf4c01bee92944a1ea16d8e854bf2fb/java/src/plume/SimpleLog.java#L188) calling `substring`
- 1 [call to `substring`](https://github.com/mernst/plume-lib-check-index-new/blob/3f82f73faaf4c01bee92944a1ea16d8e854bf2fb/java/src/plume/ICalAvailable.java#L534) I don't know if it is correct or a bug

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -76,7 +76,7 @@
 
     public final AnnotationMirror UNKNOWN, BOTTOM, POLY;
 
-    private final IndexMethodIdentifier imf;
+    protected final IndexMethodIdentifier imf;
```
kelloggm: Why is this `protected` instead of package-private (no modifier)? It's not used anywhere outside the package, and package-private prevents subclasses outside the package from seeing the field. I'm not sure why anyone would want to subclass `UpperBoundAnnotatedTypeFactory`, but we can expose this if there's a good reason.

checker/src/org/checkerframework/checker/index/lowerbound/LowerBoundAnnotatedTypeFactory.java
```diff
@@ -309,21 +309,39 @@ public Void visitMethodInvocation(MethodInvocationTree tree, AnnotatedTypeMirror
...
+                return getMinLenFromTree(tree);
+            }
+            return null;
+        }
+
+        /**
+         * Looks up the minlen of a mehtod invocation tree. Returns null if the tree doesn't
+         * represent an string length method.
+         */
+        private Integer getMinLenFromMethodInvocationTree(MethodInvocationTree tree) {
```
panacekcz: This would mean the method would have 4 parameters `(MethodInvocationTree, ValueAnnotatedTypeFactory, IndexMethodIdentifier, ProcessingEnvironment)`. I think it would be less readable, given how short the method is. Is it worth it?

kelloggm: That's a good point. I'll defer to your judgment.

panacekcz:
@kelloggm, I have addressed the review comments. Also, I enabled annotations for `CharSequence` to avoid unsoundness when accessing string through that interface.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
```diff
@@ -322,7 +322,7 @@ private void refineNeqSequenceLength(
                             atypeFactory, (FieldAccessNode) lengthAccess);
             receiver = fa.getReceiver();
 
-        } else if (IndexUtil.isStringLengthInvocation(lengthAccess, atypeFactory.imf)) {
+        } else if (atypeFactory.imf.isStringLengthInvocation(lengthAccess)) {
```
kelloggm: I made a similar comment on panacekcz#6, but this looks like a code smell because you're reading through a field. You should add a getter to the type factory that just returns the Index Method Identifier and use it here.

panacekcz:
I have addressed the review comment. Test should pass after #1352 is merged.

kelloggm:
The tests appear to be failing due to timing out. I restarted the ones that timed out (sometimes they'll pass the second time), but if this continues we'll have to either up the timeout or look into whether there's a significant performance regression in this PR somewhere.

mernst:
Even if re-running these tests multiple times manages to get them to finish under 50 minutes, pushing the changes would mean that master would have these same recurring timeout issues.  Therefore, restarting the tests can help determine whether there is a functionality error, but does not enable the pull request to be merged.

The Index Checker already imposes a slowdown, and adding it to the Checker Framework is the reason that Checker Framework Travis jobs already sometimes time out.  Adding new functionality beyond that make take us over the 50-minute timeout.  Two possibilities are to profile performance (both in the Index Checker and is these proposed changes) or to split the Travis jobs into smaller pieces.

panacekcz:
The `.travis.yml` file in this PR is changed to build the jdk, because there are changes to the index jdk. So I think that may cause the timeouts. I wouldn't expect a massive slowdown due to this PR, but I don't know how to test it on travis in this situation.

mernst:
Rather than building the JDK, it's better to prebuild it, put it somewhere, and then download it in your fork.
You can do this, or @smillst can help with this.

panacekcz:
I published the jdk built from this branch and changed the url to point at it. The Travis build is now passing.
Changes to `build.xml` should be reverted before merging.

mernst:
Great, thanks.
Also, before merging the canonical prebuilt JDK should be updated to reflect your changes.

smillst:
FYI, I'm making the August release now, so I'll merge this later today or tomorrow.

smillst:
I'm rebuilding the JDK to upload to checker-framework.org/dev-jdk; once that's done I'll merge.

PR #1418, opened 2017-07-23
Fix #1214: CFG construction does not have branches to catch blocks for unchecked exceptions

### LINKED ISSUE #1214: CFG construction does not have branches to catch blocks for unchecked exceptions
The following code passes the Nullness Checker but fails with a null pointer exception at run time.

```java
public class TryCatch {

    foo1().toString();
  }

  static String foo1() {
    String s = "not null";
    try {
      int data = 50 / 0;
    } catch (Exception e) {
      s = null;
    }
    return s;
  }
}
```

I suspect the problem is with dataflow's construction of the CFG -- the fact that `s` may be null is not reflected at the merge point after the catch clause.

The CFG construction doe.s not choose to add an outgoing edge to every operation, to reflect that the operation might throw an unchecked exception such as OutOfMemoryException.  Doing so would more accurately reflect the possible control flow, but might lead to many false positives.  (We should evaluate whether it really does lead to more false positives.)

A compromise solution would be that, if there is a catch block that catches unchecked exceptions, then edges are created to it from every operation that might throw any caught exception.  This still might lead to false positives, but it's bad style to write large try blocks or expansive catch clauses, so they should be rare.

bitterfox:
The cause of issue #1214 is that there is no `ArithmeticException` path in CFG for division operation tough it could occur.
By adding such path at every operation which it could occur, this issue is fixed.

Note that OOME which is shown as an example in the issue is an `Error` and there is no path to `catch (Exception e) {...}`even if OOME occurred.
And I believe that it's not necessary to consider `Error` path for all tree because `Error` is usually not recoverable and applications shouldn't catch it.

@mernst @wmdietl


bitterfox:
I've noticed that `AIOOBE` path and `NPE` path for an array access and `NPE` path for enhanced-for-loop for an array are missing.
I've just updated my PR to fix it and add test cases.

checker/tests/nullness/flow/Issue1214.java
```diff
@@ -0,0 +1,126 @@
...
+        String s = "not null";
+        try {
+            data /= 0;
+        } catch (Exception e) {
+            s = null;
+        }
+        return s;
+    }
+
+    static String ok4(double data) {
```
wmdietl: Can you include a test with `float`?

checker/tests/nullness/flow/Issue1214.java
```diff
@@ -0,0 +1,126 @@
+// Test case for issue #1214:
+// https://github.com/typetools/checker-framework/issues/1214
+
+public class Issue1214 {
+    static String ng1() {
+        String s = "not null";
+        try {
+            int data = 50 / 0;
```
wmdietl: All tests are with the literal `0`. What happens if the 0 is in a variable? What about:
```
static String ng2(int x) {
    String s = "not null";
    try {
        short data = (short) (50 / x);
     } catch (Exception e) {
        try {
            s = null;
        } catch (Exception ee) {
        }
    }
    return s;
}
```

wmdietl:
Can you also make sure the Dataflow Manual is updated?

bitterfox:
Hi, I've updated test case and dataflow manual. Please review again.
Thank you.

dataflow/manual/content.tex
```diff
@@ -842,6 +837,35 @@ \subsection{Formal Definition of the Control-Flow Graph}
         \label{tab:nodes}
     \end{longtable}
 
+\autoref{tab:nodesWithException} shows all node types whish can possibly throw an exception and type to be thrown.
```
wmdietl: `whish` -> `which`

wmdietl: `and type` -> `and the exception type`

dataflow/manual/content.tex
```diff
@@ -842,6 +837,35 @@ \subsection{Formal Definition of the Control-Flow Graph}
...
+Java class name of nodes are simplified as with \autoref{tab:nodes}.
+All of specific Exceptions type in table are in package \code{java.lang}.
```
wmdietl: What do you mean with `specific`?
Why is `Exceptions` upper case?
I think the `of` can be removed, but some reformulation of that sentence would be good.

dataflow/manual/content.tex
```diff
@@ -842,6 +837,35 @@ \subsection{Formal Definition of the Control-Flow Graph}
...
+
+    \begin{longtable}{lp{0.6\linewidth}l}
+        \midrule
+        \multicolumn{2}{c}{\autoref{tab:nodesWithException}: All node types could throw Exception and types to be thrown..} \\ \\
```
wmdietl: Why the two dots `..` at the end?

dataflow/manual/content.tex
```diff
@@ -842,6 +837,35 @@ \subsection{Formal Definition of the Control-Flow Graph}
...
+        \textbf{Node type} & \textbf{Exception type} \\ \midrule \endfirsthead
+
+        \textbf{Node type} & \textbf{Exception type} \\ \midrule \endhead
+        \hline \multicolumn{2}{|c|}{{Continued on next page}} \\ \hline \endfoot
+        \endlastfoot
+
+        \code{ArrayAccess} & NullPointerException, ArrayIndexOutOfBoundsException \\
+        \code{FieldAccess} & NullPointerException \\
+        \code{MethodAccess} & NullPointerException \\
+        \code{MethodInvocation} & Throwable, Types in throws clause of the signature \\
```
wmdietl: I would also put the class names of the exception in `\code{}`.
The `Types` isn't a class name, so don't capitalize.

dataflow/manual/content.tex
```diff
@@ -842,6 +837,35 @@ \subsection{Formal Definition of the Control-Flow Graph}
...
+        \code{ObjectCreation} & Throwable, Types in throws clause of the signature \\
+        \code{TypeCast} & ClassCastException \\
+        \code{Throw} & Type of \code{e} when \code{throw e} \\
+        \code{AssertionError} & AssertionError \\
+        \midrule
+
+        \caption{All node types could throw Exception and types to be thrown.}
+        \label{tab:nodesWithException}
+    \end{longtable}
+
```
wmdietl: Can you add a comment about `Errors`, like the OOME discussed above, and `NoSuchFieldError` and others.

bitterfox:
I've fixed dataflow manual. Please confirm this again.

PR #1432, opened 2017-08-02
Refine minimum string length after startsWith and endsWith

panacekcz:
Previous conversation on this PR is in panacekcz#6. This is a part of the Bounded-size strings GSoC project. 

With these changes, the minimum length of the receiver string in the then-branch is refined to the minimum length known for the argument (based on a stored annotation or the length of a string literal).

The purpose is to support code like:
```java
if(str.startsWith("x")){
 String after = str.substring(1);
}
```
This pattern appears often in [plume-lib](https://github.com/typetools/plume-lib) and causes warnings from the index checker with support for strings (#1413). With the added refinement, the checker will see that the string is at least as long as the prefix or suffix and so the index is valid for the string.

Unlike outlined in kelloggm#56, this PR does not use annotations, but adds special handling of `String.startsWith(String)` and `String.endWith(String)` to `ValueTransfer`.


framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -1324,6 +1326,58 @@ private void addAnnotationToStore(CFStore store, AnnotationMirror anno, Node nod
...
+    protected void processConditionalPostconditions(
+            MethodInvocationNode n,
+            ExecutableElement methodElement,
+            Tree tree,
+            CFStore thenStore,
+            CFStore elseStore) {
+        // For String.startsWith(String) and String.endsWith(String), refine the minimum length
+        // of the receiver to the minimum length of the argument.
+        // Basically, MinLen annotations are transferred from the argument to the receiver.
+        // However, a polymorphic qualifier cannot be used to express this, because
```
smillst: Actually, a polymorphic qualifier would not work because it does not refine the type of receiver expressions.  ( You should just delete this comment, but keep the first two lines.)

framework/src/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -1324,6 +1326,58 @@ private void addAnnotationToStore(CFStore store, AnnotationMirror anno, Node nod
...
+        // MinLen is in the value hierarchy, and using PolyValue would mean that the receiver
```
smillst: If the method was annotated like:

```
boolean startsWith(@PolyValue String this, @PolyValue String arg);
```
```
@StringVal("hello") String hello = "hello";
@StringVal("bye!!") String bye = "bye!!";
hello.startsWith(by); // legal, the @PolyValue is replaced with @ArrayLen(5). 
```

framework/src/org/checkerframework/common/value/ValueTransfer.java
panacekcz: I added this comment on this request: https://github.com/panacekcz/checker-framework/pull/6#pullrequestreview-52566329. I did not investigate this in detail because the reason stated in this comment was enough to tell that the polymorphic qualifier is not a solution. If it cannot work in principle, then I don't know why it was proposed in the first place.

panacekcz:
@smillst, @kelloggm, I removed the comment about the polymorphic qualifier. If that approach doesn't work in principle, then there is probably no point in explaining that here in the code.

framework/src/org/checkerframework/common/value/ValueTransfer.java
smillst: Thanks for the explanation.   Issue  https://github.com/kelloggm/checker-framework/issues/56 isn't very clear.  The proposal was to use the polymorphic qualifier in an `@EnsuresQualifierIf` annotation.    So `startsWith` would be annotated like:
```
@EnsuresQualifierIf( result = true, expression = "this", qualifier = PolyValue.class)
boolean startsWith(@PolyValue String arg);
```
However `@EnsuresQualifierIf` doesn't support polymorphic qualifiers, so the special casing you implemented was required.

PR #1434, opened 2017-08-03
Support string literals in ValueAnnotatedTypeFactory.getMinLenFromString

panacekcz:
This PR improves `ValueAnnotatedTypeFactory.getMinLenFromString` to return the length of the string if the expression is a string literal.

This fixes a warning from index checker with string support (#1413) in plume-lib ([`TestPlume.java`](https://github.com/mernst/plume-lib-check-index-new/blob/3f82f73faaf4c01bee92944a1ea16d8e854bf2fb/java/src/plume/TestPlume.java#L2585)), reported as kelloggm#166.
This can be merged independently of #1413.

### DIFF
--- a/checker/tests/index/Index165StringLiteral.java
+++ b/checker/tests/index/Index165StringLiteral.java
@@ -1,0 +1,15 @@
+// Test case for Issue 165, where the argument is a string literal:
+// https://github.com/kelloggm/checker-framework/issues/165
+
+import org.checkerframework.checker.index.qual.IndexFor;
+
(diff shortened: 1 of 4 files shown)


panacekcz:
I don't think I can, because the method `ValueAnnotatedTypeFactory.getMinLenFromString` is (according to eclipse) never called from the value checker, only from the upper bound checker.

smillst:
I agree with @panacekcz that the additional test isn't really possible, so I'm going to merge this.

PR #1436, opened 2017-08-06
Update README-developers file

hrishabh23:
Directory named 'tutorial' was moved inside the directory 'docs' which also contained manuals.


mernst:
Thank you!  You're right that it was out of date.
I've changed this in a slightly different way (yours had a missing space character, for example) in commit 7adb3d3f.

PR #1437, opened 2017-08-07
Ignore top annotations on the output for compound types

feliperodri:
I updated the addAnnotationsToATypeElement and removeIgnoredAnnosFromATypeElement methods on WholeProgramInferenceScenesHelper.java in order to ignore top annotations on the output for compound types.

Signed-off-by: Felipe R. Monteiro <rms.felipe@gmail.com>


feliperodri:
@mernst, could you please review this pull request?

feliperodri:
@mernst, I added CompoundTypesTest.java as a test case. Could you please review it?

I also created the following package that should store all test cases for testing WPI with Interning Checker: `framework/tests/src/testlib/wholeprograminference/interning/`

feliperodri:
@mernst,

Frist, we need to run WPI in this test case. Before the merge, WPI will infer @Interned annotations on this code, which is wrong. After the merge, WPI will not infer any annotation on this code, which is the correct result. Is there any way that I can indicate this in the test case? Should I write this test in another format?

I agree with you, this test case should be integrated with the tests in `framework/tests/whole-program-inference/`.

mernst:
The test framework used by `framework/tests/whole-program-inference/` has ways to indicate which annotations are inferred and which are not.  Reading/improving its documentation and creating a test there should enable you to create a test.

feliperodri:
@mernst could you please re-review this pull request? Thanks a lot!

Issue #1371, opened 2017-06-28
Stub parser error reporting


kelloggm:
The stub parser does not report errors in a way friendly to users. Consider the following two examples:

- invalid syntax in a stub (such as a semicolon at the very end, right after the class close-bracket) can cause a `NullPointerException` in the stub parser ([here](https://github.com/typetools/checker-framework/blob/8f0a2f38464a7a06a1ffc27ffb0f99f104668e64/framework/src/org/checkerframework/framework/stub/StubParser.java#L381)) instead of reporting an error to the user.
- even in debug mode, misspelling an annotation (e.g. `PolyAll` as `PolyAl` or similar) fails silently without warning the user that the annotation they wrote doesn't exist

These sort of problems make using the stub parser difficult, especially for novices.

wmdietl:
I agree with the general sentiment and a GSoC student is working on general improvements to the stubparser.
Regarding the second bullet: what do you mean with "in debug mode"? Did passing `stubWarnIfNotFound` not give a warning for this?

mernst:
Fixed by 8997a34.

Issue #571, opened 2016-01-28
Update the version of JavaParser used by the stub parser


smillst:
The stub parser uses version 1.0.7 of [JavaParser](http://javaparser.github.io/javaparser/).  It should be updated to use the latest version, which is currently [2.3.0](https://github.com/javaparser/javaparser/releases/tag/javaparser-parent-2.3.0).  This will fix (or at the very least help fix) the limitations of the stub parser that violate Java 8 syntax.

(This should also fix some issues that have been filed: #407, #375, #297, #220, #798.)

mernst:
Fixed by 8997a34.

Issue #375, opened 2015-07-03
StubParser assertion failure


GoogleCodeExporter:
```
> What steps will reproduce the problem?
Run the following 3 commands:
7z x demo-example.7z
cd error-version
$CHECKERFRAMEWORK/checker/bin/javac -AprintErrorStack -processor 
org.checkerframework.checker.guieffect.GuiEffectChecker -cp 
../equinox.jar:../swt-win.jar demo/ImageViewer.java 

> What is the expected output? What do you see instead?
The type-checker should run to completion.  Instead, the StubParser suffers an 
assertion error (see stack trace below).  Even if the fault is with the stub 
libraries, the Checker Framework should issue a human-friendly error message 
without a stack trace.

Is the problem that all supertypes of a given type need to be in the stub file? 
 (If so, that could be fixed by augmenting the stub file, but is that a truly 
necessary requirement?)

error: InvocationTargetException when invoking constructor for class 
org.checkerframework.checker.guieffect.GuiEffectVisitor; Underlying cause: 
java.lang.AssertionError: StubParser: could not find superinterface Drawable 
from type /*DECL*/ Display
  Exception: java.lang.reflect.InvocationTargetException; Stack trace: sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
  sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:57)
  sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
  java.lang.reflect.Constructor.newInstance(Constructor.java:526)
  org.checkerframework.common.basetype.BaseTypeChecker.invokeConstructorFor(BaseTypeChecker.java:174)
  org.checkerframework.common.basetype.BaseTypeChecker.createSourceVisitor(BaseTypeChecker.java:113)
  org.checkerframework.common.basetype.BaseTypeChecker.createSourceVisitor(BaseTypeChecker.java:80)
  org.checkerframework.framework.source.SourceChecker.initChecker(SourceChecker.java:713)
  org.checkerframework.framework.source.SourceChecker.typeProcessingStart(SourceChecker.java:674)
  org.checkerframework.javacutil.AbstractTypeProcessor$AttributionTaskListener.finished(AbstractTypeProcessor.java:183)
  com.sun.tools.javac.api.ClientCodeWrapper$WrappedTaskListener.finished(ClientCodeWrapper.java:681)
[...]

PR #1441, opened 2017-08-09
Upgrading the stubparser

Bohdankm22:
Upgrading the stubparser in the Checker Framework to use the newest JavaParser project. The stubparser will be removed and instead of this javapasrer-core.jar lib will be added in framework/lib


mernst:
What is the status of all the [issues labeled StubParser](https://github.com/typetools/checker-framework/issues?q=is%3Aopen+is%3Aissue+label%3AStubParser)?  Does this fix all of them?  Which ones does it fix?  I would think that it makes sense to merge a pull request that fixes all those issues, or as many as possible.

Regarding checker-framework-inference:  First, I am hopeful there will be little or nothing to do for it.  Second, it would not be acceptable to merge a pull request that breaks the build, only planning to fix it later.  The pull request needs to be complete.

build-common.properties
```diff
@@ -62,6 +62,7 @@ hamcrest.lib=${framework.loc}/tests/hamcrest-core-1.3.jar
 
 checker.loc=${checkerframework}/checker
 checker.lib=${checker.loc}/dist/checker.jar
+checker.lib.tmp=${checker.loc}/dist/checker-tmp.jar
```
wmdietl: Can you provide a comment what this file is for? The name implies it is temporary, but I don't see when it is removed.

build-common.properties
```diff
@@ -72,6 +73,7 @@ checker.javadoc.lib=${checker.loc}/dist/checker-javadoc.jar
 
 checker.jdk8orhigher.sources=org/checkerframework/checker/nullness/Opt.java
 
+javaparser.core.lib=${checker.loc}/lib/javaparser-core.jar
```
wmdietl: `checker/lib` contains annotated versions of libraries. This is a dependency, not an annotated library, so a different location should be used.
Also, the `framework` project depends on the javaparser, so this library shouldn't be in the `checker` project.

checker/bin-devel/javac
```diff
@@ -35,7 +35,7 @@ annoToolsDir="${cfDir}"/../annotation-tools
 # Put afu jar files last, as they might contain out-of-date CF files.
 # Put "checker" after the other sub-projects, as "ant bindist" puts
 # other projects into the checker/build directory.
-buildDirs="${cfDir}"/dataflow/build:"${cfDir}"/javacutil/build:"${cfDir}"/stubparser/build:"${cfDir}"/framework/build:"${cfDir}"/checker/build:"${annoToolsDir}"/scene-lib/bin:"${annoToolsDir}"/annotation-file-utilities/annotation-file-utilities.jar
+buildDirs="${cfDir}"/dataflow/build:"${cfDir}"/javacutil/build:"${cfDir}"/stubparser/build:"${cfDir}"/framework/build:"${cfDir}"/checker/build:"${cfDir}"/checker/lib/javaparser-core.jar:"${annoToolsDir}"/scene-lib/bin:"${annoToolsDir}"/annotation-file-utilities/annotation-file-utilities.jar
```
wmdietl: Shouldn't the whole `stubparser` directory and all of its uses be removed?

Bohdankm22:
@mernst 
I believe the issue #375 should be fixed already.
This pull request should fix #1371, #571, #407, #297, #220.
However, the issues #798 and #1356 still require efforts to be resolved.
Should I work on these issue resolution before merging the pull request?

build-common.properties
Bohdankm22: I am not sure about this one. 
It was added in https://github.com/Bohdankm22/checker-framework/commit/edc02b4fa2e2a6507d814d666337db4fc4fc7122
@mernst could you please comment?

mernst:
Looks good; I approve the pull request.
Don't squash-and-merge until tests pass and the CF developers have approved requiring Java 8 for the next release.

mernst:
Note:  when we squash-and-merge this pull request, include the text:
Fixes #220. Fixes #297. Fixes #375. Fixes #407. Fixes #571. Fixes #1371.

Bohdankm22:
I fixed it in the next commit https://github.com/typetools/checker-framework/pull/1441/commits/d7c6191c89669f25fd08614fcf5cc12c8eb1957e

Bohdankm22:
No differences in Java limitations and stub parser requirements regarding the enums declarations.
Hence the entire "Enum declarations:" item should be removed.
I will take care of it.

checker/bin-devel/javac
```diff
@@ -35,7 +35,7 @@ annoToolsDir="${cfDir}"/../annotation-tools
 # Put afu jar files last, as they might contain out-of-date CF files.
 # Put "checker" after the other sub-projects, as "ant bindist" puts
 # other projects into the checker/build directory.
-buildDirs="${cfDir}"/dataflow/build:"${cfDir}"/javacutil/build:"${cfDir}"/stubparser/build:"${cfDir}"/framework/build:"${cfDir}"/checker/build:"${annoToolsDir}"/scene-lib/bin:"${annoToolsDir}"/annotation-file-utilities/annotation-file-utilities.jar
+buildDirs="${cfDir}"/dataflow/build:"${cfDir}"/javacutil/build:"${cfDir}"/framework/build:"${cfDir}"/checker/build:"${cfDir}"/checker/lib/javaparser-core.jar:"${annoToolsDir}"/scene-lib/bin:"${annoToolsDir}"/annotation-file-utilities/annotation-file-utilities.jar
```
panacekcz: Is the `"${cfDir}"/checker/lib/javaparser-core.jar` forgotten here from some older version of this change?
I just got a little bit confused by this addition. Not really a problem, but perhaps better clean it up.

mernst: @panacekcz , the pull request removed the stub parser source code from the Checker Framework repository and added, in its place, a dependency on a .jar file built from an external repository.  So, adding that dependence is correct.

@feliperodri , should we rename the `javaparser-core.jar` file to `stubparser-core.jar`, to emphasize that this is not the real JavaParser but our modified version of it?  (However, all the package names are not changed, so it isn't currently possible to use real JavaParser at the same time as the Stub Parser.)

feliperodri: @mernst , I would definitely suggest you rename the javaparser-core.jar file to stubparser-core.jar. Futures developers might find this confusing, so it is better to make clear it's a different version (even though it isn't possible to use both at the same time right now.)

mernst: Sorry, that question should have been aimed at @Bohdankm22 .  My mistake.

panacekcz: @mernst In that case, I think the path should be changed to `"${cfDir}"/framework/lib/javaparser-core.jar` and added to javac-debug as well.

Bohdankm22: @panacekcz, @mernst it's true, the path should be changed because the lib is in the framework/lib/. I will change it and rename to the stubparser-core.jar.

mernst: And keep javac in sync with javac-debug, as Vlastimil pointed out.  (Thanks; I missed that in my review.)

Bohdankm22: @mernst, sure.
Doing the changes in https://github.com/typetools/checker-framework/pull/1451

PR #1445, opened 2017-08-11
Improve index annotations in java.util.regex.Matcher

panacekcz:
This PR strengthens the index annotations on return values of functions in `java.util.regex.Matcher`.
The methods `start()`, `end()`, `regionStart()`, and `regionEnd()` all return `@NonNegative`.
This should eliminate some false positives.
Fixes panacekcz#8.
checker/build.xml is edited to download an annotated jdk with the change.


smillst:
I'll revert and merge after I've rebuilt the jdk.

Issue #1224, opened 2017-04-07
Support for Java 9


youk:
As far as I understand, there is no support for upcoming Java 9 as of now. Is this on the roadmap? Or maybe there are workarounds to get going with JDK 9?

wmdietl:
Half a year ago I worked a bit on this: https://github.com/eisop/checker-framework
Modules changed again since then and I need to update all the repositories.
This is on my to-do list for the next few weeks.

youk:
Thank you. Do I get this right that annotated JDK 9 is currently non-existent?

mernst:
That's correct.  Many of the existing JDK annotations would transfer over, though the rearrangement of packages/modules makes that a bit harder than when upgrading from (say) JDK 7 to JDK 8.

shivanshsoni:
I think jdk9 is not much in use right now...

youk:
What difference does it make? It is complete and will be released soon. Its adoption should not be hindered by incompatible tools. Isn't this an obvious thing?

shivanshsoni:
ya you are right ut it will take some time for Users to adopt jdk9 and i think other issues are more important to improve checker framework...

youk:
It solely depends on what boat you're in. Anyone should speak for themselves only.

mernst:
@shivanshsoni Can we please keep the discussion on topic?
Unproductive conversations make the issue tracker harder for everyone to read in the future.
If you want to claim this issue, please do so.  Otherwise, your help is not needed in prioritizing it.

Pr0methean:
Since this still isn't fixed, I'm going to have to disable Checker Framework on my project. :cry:

rkraneis:
JDK 9 (and soon 10) support would be very useful.

mernst:
I agree.  It's an important issue for us.  If you wish to help, we would welcome assistance.

avandeursen:
As a work around a possibility is to keep the framework enabled for jdk8, and disable it for jdk9, using maven profiles. See my pom.xml:

https://github.com/SERG-Delft/jpacman-framework/commit/6894f594cc5bfaafce3efdd49e5794ce7100b129#diff-600376dffeb79835ede4a0b285078036

rgoldberg:
What must be updated for Checker to support Java 9 and/or 10?

Will Checker then be able to work on multiple java versions (possibly via a multi-release jar), or just one?

wmdietl:
[...]

PR #1446, opened 2017-08-15
Update manual to reflect that Java 8 is required to run the Checker Framework

mernst:


docs/manual/introduction.tex
```diff
@@ -196,7 +196,7 @@ \section{Installation\label{installation}}
 
 % Not "\url" because it looks bad in the printed manual.
 \textbf{Requirement:}
-You must have \textbf{JDK 7 or later} installed.  You can get the JDK from
+You must have \textbf{JDK 8 or later} installed.  You can get the JDK from
```
smillst: The Checker Framework can't be built with Java 9 (See https://github.com/typetools/checker-framework/issues/1224). So, we should probably make this JDK 8.

mernst:
I have made the suggested changes.

PR #1452, opened 2017-08-16
Correctly initialize extends bounds of wildcards

### LINKED ISSUE #1427: Nullness Checker crash: AsSuperVisitor: type is not an erased subtype of supertype 12
This is a minimal repro for a crash @wmdietl discovered.

The instructions below use version 2.1.13, but it also reproduces with a build that I think has all of the recent fixes in this area applied, so it looks like a distinct issue.

```java
// B.java
class One {
  abstract static class B<T extends B<T, U>, U extends One> {
    abstract U build();
    T f() { throw new AssertionError(); }
  }
}

class Two extends One {
  static class B<T extends B<T>> extends One.B<T, Two> {
    @Override Two build() { throw new AssertionError(); }
  }
}

class Three<T extends Three<T>> extends Two.B<T> {
  static Three<?> c() { throw new AssertionError(); }
  T g() { throw new AssertionError(); }
}
```

```java
// T.java
class Test {
  {
    Three.c().g().f().build();
  }
}
````

```
# compile B.java separately with stock javac8
javac -fullversion B.java 
javac full version "1.8.0_152-ea-b05"

# run the nullness checker on T.java
javac -processor org.checkerframework.checker.nullness.NullnessChecker -AprintErrorStack -ve
rsion T.java
...
javac 1.8.0-jsr308-2.1.13
  type: B<T extends B<T, U extends One>, U extends One>
  superType: B<T extends B<T extends B<T>>>
  Compilation unit: T.java
  Exception: java.lang.Throwable; Stack trace: org.checkerframework.framework.source.SourceChecker.errorAbort(SourceChecker.java:727)
  org.checkerframework.javacutil.ErrorReporter.errorAbort(ErrorReporter.java:41)
  org.checkerframework.framework.type.AsSuperVisitor.errorTypeNotErasedSubtypeOfSuperType(AsSuperVisitor.java:155)
  org.checkerframework.framework.type.AsSuperVisitor.visitDeclared_Declared(AsSuperVisitor.java:340)
  org.checkerframework.framework.type.AsSuperVisitor.visitDeclared_Declared(AsSuperVisitor.java:29)
  org.checkerframework.framework.util.AtmCombo.accept(AtmCombo.java:308)
  org.checkerframework.framework.type.visitor.AbstractAtmComboVisitor.visit(AbstractAtmComboVisitor.java:59)
  org.checkerframework.framework.type.AsSuperVisitor.visit(AsSuperVisitor.java:86)
  org.checkerframework.framework.type.AsSuperVisitor.visitTypevar_NotTypevarNorWildcard(AsSuperVisitor.java:586)
  org.checkerframework.framework.type.AsSuperVisitor.visitTypevar_Declared(AsSuperVisitor.java:593)
  org.checkerframework.framework.type.AsSuperVisitor.visitTypevar_Declared(AsSuperVisitor.java:29)
  org.checkerframework.framework.util.AtmCombo.accept(AtmCombo.java:632)
...
  type: @Initialized @NonNull B<T extends @Initialized @NonNull B<T, U extends @Initialized @NonNull One>, U extends @Initialized @NonNull One>
  superType: @Initialized @NonNull B<T extends @Initialized @NonNull B<T extends @Initialized @NonNull B<T>>>
  Compilation unit: T.java
  Exception: java.lang.Throwable; Stack trace: org.checkerframework.framework.source.SourceChecker.errorAbort(SourceChecker.java:727)
  org.checkerframework.javacutil.ErrorReporter.errorAbort(ErrorReporter.java:41)
  org.checkerframework.framework.type.AsSuperVisitor.errorTypeNotErasedSubtypeOfSuperType(AsSuperVisitor.java:155)
  org.checkerframework.framework.type.AsSuperVisitor.visitDeclared_Declared(AsSuperVisitor.java:340)
  org.checkerframework.framework.type.AsSuperVisitor.visitDeclared_Declared(AsSuperVisitor.java:29)
  org.checkerframework.framework.util.AtmCombo.accept(AtmCombo.java:308)
```

### LINKED ISSUE #1428: Nullness Checker crash: AsSuperVisitor: type is not an erased subtype of supertype 13
This is another minimal repro for a crash @wmdietl discovered. It appears to be distinct from #1427.

```java
// B.java
import java.util.List;

interface B { 
  List<L<?>> getItems();
}

interface V {}

interface L<T extends V> {}
```

```java
// T.java
import java.util.Collections;
import java.util.List;

class T {
  public List<L<?>> f(B b) {
    return true ? Collections.<L<?>>emptyList() : b.getItems();
  }
}
````

```
# compile B.java separately
javac B.java 

# run the nullness checker on T.java
javac -processor org.checkerframework.checker.nullness.NullnessChecker -AprintErrorStack T.java
...
  type: Object
  superType: V
  Compilation unit: T.java
  Exception: java.lang.Throwable; Stack trace: org.checkerframework.framework.source.SourceChecker.errorAbort(SourceChecker.java:727)
  org.checkerframework.javacutil.ErrorReporter.errorAbort(ErrorReporter.java:41)
  org.checkerframework.framework.type.AsSuperVisitor.errorTypeNotErasedSubtypeOfSuperType(AsSuperVisitor.java:155)
  org.checkerframework.framework.type.AsSuperVisitor.visitDeclared_Declared(AsSuperVisitor.java:340)
  org.checkerframework.framework.type.AsSuperVisitor.visitDeclared_Declared(AsSuperVisitor.java:29)
  org.checkerframework.framework.util.AtmCombo.accept(AtmCombo.java:308)
  org.checkerframework.framework.type.visitor.AbstractAtmComboVisitor.visit(AbstractAtmComboVisitor.java:59)
  org.checkerframework.framework.type.AsSuperVisitor.visit(AsSuperVisitor.java:86)
  org.checkerframework.framework.type.AsSuperVisitor.visitWildcard_Wildcard(AsSuperVisitor.java:790)
  org.checkerframework.framework.type.AsSuperVisitor.visitWildcard_Wildcard(AsSuperVisitor.java:29)
  org.checkerframework.framework.util.AtmCombo.accept(AtmCombo.java:716)
  org.checkerframework.framework.type.visitor.AbstractAtmComboVisitor.visit(AbstractAtmComboVisitor.java:59)
  org.checkerframework.framework.type.AsSuperVisitor.visit(AsSuperVisitor.java:86)
  org.checkerframework.framework.type.AsSuperVisitor.asSuper(AsSuperVisitor.java:68)
  org.checkerframework.framework.util.AnnotatedTypes.asSuper(AnnotatedTypes.java:131)
  org.checkerframework.framework.util.AtmLubVisitor.lubTypeArgument(AtmLubVisitor.java:198)
  org.checkerframework.framework.util.AtmLubVisitor.visitDeclared_Declared(AtmLubVisitor.java:184)
  org.checkerframework.framework.util.AtmLubVisitor.visitDeclared_Declared(AtmLubVisitor.java:31)
  org.checkerframework.framework.util.AtmCombo.accept(AtmCombo.java:308)
  org.checkerframework.framework.type.visitor.AbstractAtmComboVisitor.visit(AbstractAtmComboVisitor.java:59)
  org.checkerframework.framework.util.AtmLubVisitor.lubTypeArgument(AtmLubVisitor.java:243)
  org.checkerframework.framework.util.AtmLubVisitor.visitDeclared_Declared(AtmLubVisitor.java:184)
  org.checkerframework.framework.util.AtmLubVisitor.visitDeclared_Declared(AtmLubVisitor.java:31)
  org.checkerframework.framework.util.AtmCombo.accept(AtmCombo.java:308)
  org.checkerframework.framework.type.visitor.AbstractAtmComboVisitor.visit(AbstractAtmComboVisitor.java:59)
```

### PR DESCRIPTION
**Pull request by smillst**
The extends bound of a unbound or super bounded wildcard should be initialized to the upper bound of the type parameter to which it is an argument.  (If the wildcard has an extends bound, then use that even if it is a supertype of the upper bound of the type parameter.)

Fixes #1427 and Fixes #1428

### CODE REVIEW

File: framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java [hunk 1]
```diff
@@ -1966,6 +1963,24 @@ private void fixupBoundAnnotations() {
             }
         }
 
+        void setTypeVariable(TypeParameterElement typeParameterElement) {
```
wmdietl: Explicit document why this method should have default visibility, same for overload below.

framework/src/org/checkerframework/framework/type/BoundsInitializer.java
```diff
@@ -391,36 +456,62 @@ public void initializeExtendsBound(final AnnotatedWildcardType wildcard) {
...
                 }
+                typeArgReplacements.add(replaceOrVisit(typeArg));
+                removePathNode(node);
+            }
+
+            declaredType.setTypeArguments(typeArgReplacements);
+        }
 
-                declaredType.setTypeArguments(typeArgReplacements);
+        Map<TypeVariable, WildcardType> rawTypeWildcards = new HashMap<>();
```
wmdietl: Can you add some documentation for this field, please.
Why is the field default visibility?

framework/src/org/checkerframework/framework/type/BoundsInitializer.java
```diff
@@ -391,36 +456,62 @@ public void initializeExtendsBound(final AnnotatedWildcardType wildcard) {
...
+
+        private TypeMirror getJavaType(
```
wmdietl: Can this method be together with the similarly named `getJavaTypeArg`? Can they be static? Some Javadoc would be good.

framework/tests/all-systems/WildcardForEach.java
```diff
@@ -0,0 +1,10 @@
+import java.util.List;
+
+public class WildcardForEach {
+    static class Gen<T extends Gen<?>> {}
+
+    void test(List<Gen<?>> x) {
+        // This use to crash in org.checkerframework.framework.flow.CFTreeBuilder#buildAnnotatedType
```
wmdietl: use -> used ? Or us of `x`?

framework/src/org/checkerframework/framework/type/BoundsInitializer.java
```diff
@@ -391,36 +465,76 @@ public void initializeExtendsBound(final AnnotatedWildcardType wildcard) {
...
+         * Returns the underlying java type of the {@code i} type parameter of {@code type}. If
+         * {@code type} is raw, then a new wildcard is created or returned from {@code
+         * rawTypeWildcards}.
+         *
+         * @param type declared type
+         * @param parameters elements of the type parameters
+         * @param i index of the type parameter
+         * @return the underlying java type of the {@code i} type parameter of {@code type}
+         */
+        private TypeMirror getJavaType(
```
wmdietl: Can you please move `getJavaType` and `getJavaTypeArg` next to each other? Their javadoc is basically the same, but this version caches the created wildcard.
How does one choose between the two? Are the wildcards computed differently? Why are both needed?
Add a note that the implementation of the two should be kept in sync?

framework/src/org/checkerframework/framework/type/BoundsInitializer.java
```diff
@@ -391,36 +465,76 @@ public void initializeExtendsBound(final AnnotatedWildcardType wildcard) {
...
+
+            declaredType.setTypeArguments(typeArgReplacements);
+        }
 
-                declaredType.setTypeArguments(typeArgReplacements);
+        /**
+         * Store the wildcards created as type arguments to raw types. This way if the raw type is
+         * visited again, the same wildcard can be used.
+         */
+        private Map<TypeVariable, WildcardType> rawTypeWildcards = new HashMap<>();
```
wmdietl: Can the field be final?
When can the map be cleared? Or should it be a LRU map to limit its size?

framework/src/org/checkerframework/framework/type/BoundsInitializer.java
```diff
@@ -47,6 +50,71 @@
...
+                    // do nothing
+            }
+            typeArgs.add(typeArg);
+        }
+
+        declaredType.typeArgs = Collections.unmodifiableList(typeArgs);
+    }
+
+    /**
+     * Returns the underlying java type of the {@code i} type parameter of {@code type}. If {@code
```
wmdietl: The `type parameter` should be `type argument` to match the method name.
Changing it to `{@code i}-th` makes it a bit easier to read.
Same comments for `@return` below.

framework/src/org/checkerframework/framework/type/BoundsInitializer.java
smillst: I'll make it final.  There's no need to limit or clear the cache as a new `InitializerVisitor` is created for each type that is initialized.

framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java
```diff
@@ -925,29 +911,34 @@ public void setTypeArguments(List<? extends AnnotatedTypeMirror> ts) {
...
-                    for (TypeMirror t : getUnderlyingType().getTypeArguments()) {
-                        typeArgs.add(createType(t, atypeFactory, declaration));
-                    }
+            if (typeArgs != null) {
+                return typeArgs;
+            } else if (wasRaw()) {
+                // Initialize the type arguments with uninferred wildcards
+                BoundsInitializer.initializeTypeArgs(this);
+
+                // Copy annotations from the declaration to the wildcards.
```
wmdietl: Could this be moved into `initializeTypeArgs`? It seems strange to first have a dedicated function to initialize type arguments and then modify the result again. If not, please add a comment why this logic needs to be here and not there.

framework/src/org/checkerframework/framework/type/BoundsInitializer.java
```diff
@@ -47,6 +50,60 @@
...
+                case TYPEVAR:
+                    initializeBounds((AnnotatedTypeVariable) typeArg);
+                    break;
+                default:
+                    // do nothing
+            }
+            typeArgs.add(typeArg);
+        }
+
+        declaredType.typeArgs = Collections.unmodifiableList(typeArgs);
```
wmdietl: The return is also wrapped in an unmodifiable list `getTypeArguments`. Add to the javadoc of this method that the results are unmodifiable and remove the wrapping from all callers?

framework/src/org/checkerframework/framework/type/BoundsInitializer.java
```diff
@@ -391,36 +454,76 @@ public void initializeExtendsBound(final AnnotatedWildcardType wildcard) {
...
+                removePathNode(node);
+            }
+
+            declaredType.setTypeArguments(typeArgReplacements);
+        }
 
-                declaredType.setTypeArguments(typeArgReplacements);
+        /**
+         * Store the wildcards created as type arguments to raw types. This way if the raw type is
+         * visited again, the same wildcard can be used.
```
wmdietl: You said

> There's no need to limit or clear the cache as a new InitializerVisitor is created for each type that is initialized.

I'm now confused what the `visited again` means. Is it for a type like `Map<Raw, Raw>` where we have two raw types? Why should they get the same wildcard?
Or is the same InitializerVisitor used on the same type multiple times?

smillst:
```
class Foo<T extends Foo> {}
```

The upper bound of the wildcard in `Foo<?>` is `Foo`.   The type argument of `Foo` is initialized to `? extends Foo`.  The type argument of `Foo` in  `? extends Foo` needs to be initialized to the same type argument as the first `Foo` so that `BoundsInitializer.InitializerVisitor#replaceOrVisit` will return the cached AnnotatedWildcardType.

framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java
smillst: The BoundsInitializer creates types without adding annotations, so this code doesn't belong there.

framework/src/org/checkerframework/framework/type/BoundsInitializer.java
wmdietl: How about adding this explanation somewhere, either here or with the class? Otherwise, the "visited again" isn't clear at all.

framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java
```diff
@@ -925,29 +911,34 @@ public void setTypeArguments(List<? extends AnnotatedTypeMirror> ts) {
...
+                    wct.replaceAnnotations(typeParam.getAnnotations());
                 }
                 typeArgs = Collections.unmodifiableList(typeArgs);
+                return typeArgs;
+            } else if (getUnderlyingType().getTypeArguments().isEmpty()) {
+                typeArgs = Collections.emptyList();
+                return typeArgs;
+            } else {
+                BoundsInitializer.initializeTypeArgs(this);
+                typeArgs = Collections.unmodifiableList(typeArgs);
```
wmdietl: You said that `initializeTypeArgs` doesn't add annotations to the types. Why doesn't this branch need to add annotations?

framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java
```diff
@@ -925,29 +911,34 @@ public void setTypeArguments(List<? extends AnnotatedTypeMirror> ts) {
...
+                            .replaceAnnotations(typeParam.getLowerBound().getAnnotations());
+                    wct.replaceAnnotations(typeParam.getAnnotations());
                 }
                 typeArgs = Collections.unmodifiableList(typeArgs);
+                return typeArgs;
+            } else if (getUnderlyingType().getTypeArguments().isEmpty()) {
+                typeArgs = Collections.emptyList();
+                return typeArgs;
+            } else {
+                BoundsInitializer.initializeTypeArgs(this);
```
wmdietl: Can you add a comment what case this branch covers?

framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java
smillst: This type might have explicit annotations on the type arguments or the annotations are added via defaulting or some other process.  (Note, adding annotations to raw types in this method is the previous behavior.)

PR #1454, opened 2017-08-16
Add MinLen(1) to StringTokenizer.nextToken

panacekcz:
This PR adds a `@MinLen(1)` to the `nextToken` methods of `java.util.StringTokenizer`. These methods always return a non-empty token or throw an exception.
Inconveniently, the [javadoc](https://docs.oracle.com/javase/7/docs/api/java/util/StringTokenizer.html) does not explicitly state that tokens are non-empty. It seems to implicitly assume this property.
Fixes panacekcz#16.


kelloggm:
In general, I'm against including an annotation in the JDK unless the Javadoc alone would justify its inclusion. I understand that there is code that relies on this fact, but I'd argue that if the Javadoc doesn't guarantee it, client code shouldn't rely on it.

If you really think that this property is guaranteed, I suggest trying to convince the OpenJDK folks to include it the JDK10: [http://openjdk.java.net/contribute/](http://openjdk.java.net/contribute/)

If that contribution is accepted, then I'm more than happy to include this here. However, I think that for now the better thing to do is to communicate with the owners of the library you were typechecking (IIRC it was Daikon, so Mike will do) and file a bug against their code. They shouldn't depend on implementation details of the JDK.

For our standard policy on this, see [the manual](https://checkerframework.org/manual/#libraries-specification).

mernst:
I entirely agree with Martin's sentiment.

In this particular case, I think that the annotations are acceptable.
The property is documented in the Javadoc (even though not in the Javadoc for the method).  The class Javadoc says:

> An instance of StringTokenizer behaves in one of two ways, depending on whether it was created with the returnDelims flag having the value true or false:
>
> * If the flag is false, delimiter characters serve to separate tokens. A token is a maximal sequence of consecutive characters that are not delimiters.
> * If the flag is true, delimiter characters are themselves considered to be tokens. A token is thus either one delimiter character, or a maximal sequence of consecutive characters that are not delimiters.

A maximal sequence of characters contains at least one character.

mernst:
@smillst , could you rebuild the JDK and merge this pull request?  Thanks!

smillst:
I updated the jdk, so once the test pass, this can be merged.

checker/jdk/index/src/java/util/StringTokenizer.java
```diff
@@ -332,7 +333,7 @@ public boolean hasMoreTokens() {
      * @exception  NoSuchElementException  if there are no more tokens in this
      *               tokenizer's string.
      */
-    public String nextToken() {
+    public @MinLen(1) String nextToken() {
```
smillst: Because `@MinLen` is a Value Checker annotation, this annotations should be in 
org/checkerframework/common/value/statically-executable.astub rather than the Index Checker's annotated jdk.

panacekcz:
@smillst, I have moved the annotations to statically-executable.astub.
Travis build fails for unknown reason not caused by this change (the same failure is in master too).

PR #1461, opened 2017-08-18
Add IndexOf hierarchy for annotating String.indexOf in the index checker

panacekcz:
This PR adds support in the index checker for a pattern of using the result of `String.indexOf`, where the index is checked not to be -1 and then the length of the substring searched for is added to it.
This PR fixes false positives reported by the upper bound checker, which currently cannot prove that adding the length results in a valid index:
```java
public static String remove(String l, String s) {
 int i = l.indexOf(s);
 if (i != -1) {
  return l.substring(i + s.length()); //argument.type.incompatible
 }
 return l;
}
```
This pattern appears often in daikon and plume-lib, and in one index checker test that had to be disabled in #1413 (issue panacekcz#4).

The property of the return value of `String.indexOf` cannot be expressed by the exisiting index checker annotations (see https://github.com/typetools/checker-framework/pull/1461#issuecomment-326268978 ~https://github.com/panacekcz/checker-framework/issues/4#issuecomment-320232816~). Therefore, a separate IndexOf hierarchy is added. The annotation `IndexOfIndexFor` is similar to `LTLengthOf` with the same parameters, but additionally allowing the value to be -1.
When the upper bound checker processes the `+` operator, it looks at the IndexOf qualifiers of the operands. If an operand is `@IndexOfIndexFor(value="l",offset="j")` and `@NonNegative` at the same time, it is treated as `@LTLengthOf(value="l",offset="j")` (`@IndexOfIndexFor` is converted to an `UBQualifier` in the same way and then combined with the qualifier from the upper bound checker using `glb`).

The IndexOf hierarchy reuses `UBQualifier` to implement the lattice operators.

- `OffsetDependentTypesHelper` is moved out of `UpperBoundAnnotatedTypeFactory` for reuse, with no changes.
- `build.xml` is changed to download a changed annotated jdk.
- I haven't updated the manual yet

### Annotation syntax

In this PR, the result of `l.indexOf(s)` is annotated `@IndexOfIndexFor(value="l",offset="s.length()-1")`.
Perhaps a better annotation from users' point of view (more readable and with clearer meaning) would be an annotation `@IndexOf(value="s",in="l")` which would replace `@IndexOfIndexFor` and could also imply `@GTENegativeOne` and `@LTEqLengthOf("l")` for convenience.
However, this PR attempts to achieve the goal with least implementation effort and tries to reuse the existing code where possible.

Supporting `@IndexOf` would mean:
- the IndexOf qualifier hierarchy would be simpler. A simple flat hierarchy should work here.
- the conversion from the `@IndexOfIndexFor` annotation to the corresponding UBQualifier would be more complicated. The conversion code could not be reused and would have to be duplicated. There would have to be a conversion from the expression `"s"` to `"s.length()-1"`.
- optionally making the lower bound and upper bound checker treat `@IndexOf` as `@NonNegative` and `@LTEqLengthOf`.


checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -128,11 +124,19 @@ ValueAnnotatedTypeFactory getValueAnnotatedTypeFactory() {
      * Provides a way to query the Search Index Checker, which helps the Index Checker type the
      * results of calling the JDK's binary search methods correctly.
      */
-    private SearchIndexAnnotatedTypeFactory getSearchIndexAnnotatedTypeFactory() {
+    SearchIndexAnnotatedTypeFactory getSearchIndexAnnotatedTypeFactory() {
```
kelloggm: I don't see anywhere else in this PR that needs a `SearchIndexAnnotatedTypeFactory`; maybe this is left over from before you factored out the IndexOf checker?

checker/src/org/checkerframework/checker/index/indexof/IndexOfAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,130 @@
...
+
+    public AnnotationMirror convertUBQualifierToAnnotation(UBQualifier qualifier) {
+        if (qualifier.isUnknown()) {
+            return UNKNOWN;
+        } else if (qualifier.isBottom()) {
+            return BOTTOM;
+        }
+
+        LessThanLengthOf ltlQualifier = (LessThanLengthOf) qualifier;
+        return ltlQualifier.convertToIndexOfAnnotationMirror(processingEnv);
```
mernst: The call is to a method whose name ends `to...AnnotationMirror`, but it's within a method whose name ends `toAnnotation`.  Please make the names consistent throughout the codebase.

checker/src/org/checkerframework/checker/index/qual/IndexOfIndexFor.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import org.checkerframework.framework.qual.JavaExpression;
+import org.checkerframework.framework.qual.SubtypeOf;
+
+/**
+ * The annotated expression evaluates to either -1 or a value compatible with a {@link LTLengthOf}
```
mernst: What does "compatible with" mean?

checker/src/org/checkerframework/checker/index/qual/IndexOfIndexFor.java
```diff
@@ -0,0 +1,37 @@
...
+ * annotation with the same parameters-
```
mernst: Weird terminating punctuation.

checker/src/org/checkerframework/checker/index/qual/IndexOfIndexFor.java
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * <p>This is the return type of {@link java.lang.String#indexOf(String) String.indexOf} and {@link
```
mernst: How are the annotation's argument related to the method call's arguments?  A concrete example would be helpful.

checker/src/org/checkerframework/checker/index/qual/IndexOfIndexFor.java
```diff
@@ -0,0 +1,37 @@
...
+ * java.lang.String#lastIndexOf(String) String.lastIndexOf} in the JDK.
+ *
+ * @checker_framework.manual #index-checker Index Checker
+ */
+@SubtypeOf(IndexOfUnknown.class)
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE_USE, ElementType.TYPE_PARAMETER})
+public @interface IndexOfIndexFor {
```
mernst: The documentation should somewhere explain the name; what is a mnemonic for, or why does this name convey the desired meaning?  I don't understand that from the documentation.

checker/src/org/checkerframework/checker/index/qual/IndexOfIndexFor.java
```diff
@@ -0,0 +1,37 @@
...
+    /** Sequences, each of which is longer than the annotated expression's value. */
+    @JavaExpression
+    public String[] value();
+
+    /**
+     * If the annotated expression is not -1, then this expression plus the annotated expression is
```
mernst: "This expression" doesn't make sense, because the documentation refers to an array.  Maybe it means that this is true of each element of the array?

mernst: Is not lexically `-1`, or isn't a compile-time constant that evaluates to -1, or can't be proven by the Constant Value checker to be -1?

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -133,6 +129,14 @@ private SearchIndexAnnotatedTypeFactory getSearchIndexAnnotatedTypeFactory() {
     }
 
     /**
+     * Provides a way to query the IndexOf Checker, which helps the Index Checker type the results
```
mernst: Throughout, comments like this don't help me much.  It might be better to say what it represents.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -184,62 +188,7 @@ private void addUpperBoundTypeFromValueType(
...
-                        // adapt it based on the declaration of length; it will fail.
-                        if (TreeUtils.isArrayLengthAccess(tree)) {
-                            return null;
-                        }
-                        return super.visitMemberSelect(tree, type);
-                    }
-                };
-            }
-        };
+        return new OffsetDependentTypesHelper(this);
```
mernst: Abstracting this into a new class is a fine choice.  Why did you create a new file for it rather than making it a nested class in this file?

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
```diff
@@ -519,6 +521,37 @@ private UBQualifier removeSequenceLengths(LessThanLengthOf i, LessThanLengthOf j
     }
 
     /**
+     * Returns the UBQualifier for node, with additional refinement useful for integer addition. It
```
mernst: This comment describes the implementation, which is a fine thing to have as a code comment.  However, the Javadoc should give the specification.  This Javadoc doesn't do so.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
panacekcz: `OffsetDependentTypesHelper` does not use any internals of `UpperBoundAnnotatedTypeFactory`, it is not used exclusively by it and there is no logical relationship between these classes. Therefore I don't see any reason for nesting it.
Also, I put it in the common `org.checkerframework.checker.index` package, because it is used by both the Upper bound and IndexOf checkers. However, the dependence on `OffsetEquation` might be a reason to put it in the `org.checkerframework.checker.index.upperbound` package.

panacekcz:
@mernst, I have adressed the review comments.

docs/manual/index-checker.tex
```diff
@@ -75,6 +75,9 @@ \section{Index Checker structure and annotations\label{index-annotations}}
 \item
   whether an integer came from calling the JDK's binary search routine on
   an array (Section~\ref{index-searchindex})
+\item
+  whether an integer came from calling the JDK's \<String.indexOf> or
```
mernst: This does not seem like a distinction as important as the others; it seems to just combine lower and upper bounds, so I would not list it separately.

docs/manual/index-checker.tex
```diff
@@ -457,8 +460,65 @@ \section{Binary search indices\label{index-searchindex}}
   This is the top type, and programmers should rarely need to write it.
 \end{description}
 
+\section{IndexOf indices\label{index-indexof}}
```
mernst: I found the manual section confusing.

What is an "IndexOf index"?  The term never appears in the manual again, so this section title will confuse users.

`@IndexOfIndexFor` should be listed in section 7.1 along with other annotations.

The manual does not explicitly describe the relationship between `@LTLengthOf and `@IndexOfIndexFor`.
I would think that `@IndexOfIndexFor` should appear in section 7.2 and 7.3 to make these relationships clear.

More seriously, the manual does not explain the distinction between `@IndexOrLow` and `@IndexOfIndexFor`.  These seem to be exactly the same except for the addition of a parameter, and this should be discussed and the names should be similar.

The mnemonic behind the name `@IndexOfIndexFor` is never explained.  It's also ungainly.  I suggest renaming it.  Maybe `@IndexOf`, though that could be confused with `@IndexFor`.  Something more like `@IndexOrLow` would be appropriate.

Does the implementation require a separate checker, or is `@IndexOfIndexFor` just syntactic sugar?
The manual description strongly suggests the latter, and that is the implementation I would recommend.  It is likely to lead to fewer special cases and more correct code in general.

Even if the implementation requires a separate checker, the manual need not make the distinction.
Describing this together with the other upper bound and lower bound annotations, as syntactic sugar (exactly as `@IndexFor` is described), is probably most helpful to users.

docs/manual/index-checker.tex
```diff
@@ -457,8 +460,65 @@ \section{Binary search indices\label{index-searchindex}}
...
+
+The sole purpose of the IndexOf Checker is to allow expressing the
```
mernst: This mention of "the IndexOf Checker" (an implementation detail) comes out of the blue.  What is it, and why should the user care?  The manual should start with things that the user cares about, in terms that the user can understand.

docs/manual/index-checker.tex
```diff
@@ -457,8 +460,65 @@ \section{Binary search indices\label{index-searchindex}}
...
+properties of the return value of the JDK's
+\sunjavadoc{java/lang/String.html\#indexOf-java.lang.String-}{String.indexOf} and
+\sunjavadoc{java/lang/String.html\#lastIndexOf-java.lang.String-}{String.lastIndexOf} methods.
+These methods return an index of a substring, or -1 if the substring is not found.
+
+Because of the value -1, the return type of these methods cannot be
+annotated by a \<@LTLengthOf> with a positive offset.
+The \<@IndexOfIndexFor> annotation solves it by having the same
+meaning as \<@LTLengthOf> with the same parameters, except additionally
```
mernst: This is the most important information in this entire section, and it should be the most prominent.
The description is actually incorrect, since `@LTLengthOf` already permits -1.  Did you mean `@IndexFor`?

docs/manual/index-checker.tex
```diff
@@ -457,8 +460,65 @@ \section{Binary search indices\label{index-searchindex}}
...
+allowing the -1 value. The Upper Bound Checker can then infer
+the corresponding \<@LTLengthOf> annotation for expressions that
+are known to be non-negative.
+
+The \<@IndexOfIndexFor(value="this", offset="\#1.length()-1")> annotation on the return value of
+\<String.indexOf> allows writing code such this without warnings from the Upper Bound Checker.
```
mernst: What would go wrong otherwise?  It usually isn't helpful to readers to write a code snippet and expect the reader to understand the behavior with and without a feature.  (I realize this is clear to you and that it concisely motivates the feature, but you have more context and understanding of the feature than readers do, so it doesn't work for other people.)  I would just cut the example, and as discussed above the whole section.  Organizing the manual in terms of implementation details such as subcheckers is not helpful to a reader.

panacekcz:
Apparently, the `@IndexOfIndexFor(value="l",offset="s.length()-1")` annotation is too confusing. 
So I think rather than trying to motivate and explain this, it will be better to invest a little bit more into the implementation and use the `@IndexOf(value="s", in="l")` syntax. The correspondence between `@IndexOfIndexFor` and `@LTLengthOf` will disappear, but there will be better separation of the concepts and the annotation will be easy to understand and use.

mernst:
A surface change to the syntax is orthogonal to all of my questions.  I never even tried to understand the `offset=` annotation argument (which seemed to be undocumented in the manual).  So, changing the syntax will not address any of my concerns.  That doesn't mean that the change is necessarily a bad idea.  However, I feel that clarifying and documenting the concepts are the most important tasks to perform.  Thanks!

panacekcz:
To answer the questions about why it is not "syntactic sugar" and has its own sub-checker, here is an explanation. I am going to include it in the manual.
## Typing the result of `String.indexOf(String)`
The methods `String.indexOf(String)` and `String.lastIndexOf(String)` return an index of a given substring within a given string, or -1 if no such substring exists.
The index `i` returned from `receiver.indexOf(substring)` satisfies this property: `i==-1 || i>=0 && i<=receiver.length() - substring.length()`.

This tells us what Index Checker annotations can be applied to `i`:
- `i` is `@GTENegativeOne`, because `i>=-1`.
- `i` is `@LTEqLengthOf("receiver")`, because `i<=receiver.length()`.
- `i` is not `@IndexOrLow("receiver")`, because for `receiver="", substring="", i="0"`, the property `i>=-1 && i<receiver.length()` does not hold.
- `i` is not `@IndexOrHigh("receiver")`, because for `receiver="", substring="b", i=-1`, the property `i>=0 && i<=receiver.length()` does not hold.
- `i` is not `@LTLengthOf(value="receiver", offset="substring.length()-1")`, because for `receiver="",substring="abc",i=-1`, the property `i+substring.length()-1<receiver.length()` does not hold.

The last annotation in the list above, `@LTLengthOf(value="receiver", offset="substring.length()-1")`, is the correct and precise upper bound for all values of `i` except -1. The offset expresses the fact that we can add `substring.length()` to this index and still get a valid index for `receiver`. That is useful for type-checking code, where we add the length of the substring to the index to get the rest of the string after the occurence we found. However, the upper bound applies only after the index is explicitly checked not to be -1:
```java
 int i = receiver.indexOf(substring);
 // i is @GTENegativeOne and @LTEqLengthOf("receiver")
 // i is not @LTLengthOf(value="receiver", offset="substring.length()-1")
 if (i != -1) {
  // i is @NonNegative and @LTLengthOf(value="receiver", offset="substring.length()-1")
  int j = i + substring.length();
  // j is @IndexOrHigh("receiver")
  return receiver.substring(j); // this call is safe
 }
```

The property of the result of `indexOf` cannot be expressed by any combination of lower- and upper-bound annotations. It's because the upper-bound annotations apply independently of the lower-bound annotations, but in this case, the upper bound `i<=receiver.length() - substring.length()` holds only if `i>=0`.

Therefore, to express this property and make the example type-check without false positives, a new annotation such as `@IndexOfIndexFor(value="receiver", offset="substring.length()-1")` or `@IndexOfSubstring(value="substring", in="receiver")` is necessary. This new annotation is similar to `@LTLengthOf(value="receiver", offset="substring.length()-1")`, but explicitly allows the index to be -1 even if the upper bound would not allow it.

kelloggm:
I looked over this explanation, and I think it does an excellent job of explaining why this new annotation is necessary. I think that this should definitely go into the manual mostly as-is, and be followed by a short (1 or 2 paragraphs at most) explanation that a dedicated checker is used. You should also document the annotations. I think that should be sufficient for readers of the manual.

I have a few minor grammatical fixes I'd like to apply. Since the document is a github comment, though, I can't do this in a way that's easy to review. Could you check it in as part of the manual so that I can easily make those changes and keep the history?

docs/manual/index-checker.tex
```diff
@@ -207,7 +207,7 @@ \section{Lower bounds\label{index-lowerbound}}
     Checker.  On the left is a type system for lower bounds.  On the right
     is a type system for upper bounds.  Qualifiers written in gray should
     never be written in source code; they are used internally by the type
-    system. ''myArray'' in the Upper Bound qualifiers is an example of
+    system. \<"myArray"> in the Upper Bound qualifiers is an example of
```
kelloggm: Why were these quotes changed from what was already in the manual? Was there something wrong with what was here already? It seems unrelated to the rest of the PR.

docs/manual/index-checker.tex
```diff
@@ -457,8 +457,119 @@ \section{Binary search indices\label{index-searchindex}}
   This is the top type, and programmers should rarely need to write it.
 \end{description}
 
+\section{Typing the result of \<String.indexOf(String)>\label{index-indexof}}
+
+The methods \sunjavadoc{java/lang/String.html\#indexOf-java.lang.String-}{String.indexOf}
+and \sunjavadoc{java/lang/String.html\#lastIndexOf-java.lang.String-}{String.lastIndexOf} return an index
+of a given substring within a given string, or -1 if no such substring exists.
+The index \<i> returned from \<receiver.indexOf(substring)> satisfies this property:
+\code{i==-1 || i>=0 \&\& i<=receiver.length() - substring.length()}.
```
kelloggm: I'd prefer to use spaces between operators (so `i == -1` instead of `i==-1`, etc.). This makes the code more readable.

kelloggm: This appears a few other times below, as well.

docs/manual/index-checker.tex
```diff
@@ -457,8 +457,119 @@ \section{Binary search indices\label{index-searchindex}}
...
+\item
+\<i> is not \<@LTLengthOf(value="receiver", offset="substring.length()-1")>,
+because for \code{receiver="", substring="abc", i=-1},
+the property \code{i+substring.length()-1<receiver.length()} does not hold.
+\end{itemize}
+The last annotation in the list above, \<@LTLengthOf(value="receiver", offset="substring.length()-1")>,
+is the correct and precise upper bound for all values of \<i> except -1.
+The offset expresses the fact that we can add \<substring.length()> to
+this index and still get a valid index for \<receiver>.
+That is useful for type-checking code, where we add the length of the substring
```
kelloggm: No need for the comma on this line

docs/manual/index-checker.tex
```diff
@@ -457,8 +457,119 @@ \section{Binary search indices\label{index-searchindex}}
...
+ if (i != -1) {
+  // i is @NonNegative and @LTLengthOf(value="receiver", offset="substring.length()-1")
+  int j = i + substring.length();
+  // j is @IndexOrHigh("receiver")
+  return receiver.substring(j); // this call is safe
+ }
+\end{Verbatim}
+
+The property of the result of \<indexOf> cannot be expressed by any combination of lower- and upper-bound annotations.
+It's because the upper-bound annotations apply independently of the lower-bound annotations,
```
kelloggm: There's no need for this to be two sentences. Just say "The property...annotations, because the...". This is cleaner and easier to read than having to remember what the antecedent of "it" is.

docs/manual/index-checker.tex
```diff
@@ -457,8 +457,119 @@ \section{Binary search indices\label{index-searchindex}}
...
+but in this case, the upper bound \code{i<=receiver.length() - substring.length()} holds only if \code{i>=0}.
+Therefore, to express this property and make the example type-check without false positives,
+a new annotation such as
+\<@IndexOfIndexFor(value="receiver", offset="substring.length()-1")>
+%Alternaive:
+%\<@IndexOfSubstring\allowbreak(value="substring", in="receiver")>
+is necessary.
+
+This new annotation is similar to \<@LTLengthOf\allowbreak(value="receiver", offset="substring.length()-1")>,
+but explicitly allows the index to be -1 even if the upper bound would not allow it.
```
kelloggm: I would say "even if the upper bound's offset would not allow it" or something along those lines, to help the reader recall the case above where the correct annotation is an `LTLengthOf` with an offset as long as the index is not -1.

docs/manual/index-checker.tex
```diff
@@ -457,8 +457,119 @@ \section{Binary search indices\label{index-searchindex}}
...
+The Upper Bound Checker can infer
+the corresponding \<@LTLengthOf> annotation for expressions that have a
+\<@IndexOfIndexFor>
+%\<@IndexOfSubstring>
+annotation and at the same time are known to be non-negative (according to the Lower Bound Checker).
+
+% The \<@IndexOfIndexFor(value="this", offset="\#1.length()-1")> annotation on the return value of
```
kelloggm: Why is this commented out? I would either include it or remove it entirely.

docs/manual/index-checker.tex
panacekcz: I don't know, this is from https://github.com/typetools/checker-framework/pull/1461/commits/ec20ff96bc689003a47c05e15449c20e9bdf5d95.

panacekcz:
@kelloggm, I have addressed the review comments. I included the example that was commented out. Or should I move that to the beginning of the section as suggested by https://github.com/typetools/checker-framework/pull/1461#discussion_r135767362?

docs/manual/index-checker.tex
kelloggm: Okay, we'll leave it.

checker/src/org/checkerframework/checker/index/qual/IndexOfIndexFor.java
```diff
@@ -0,0 +1,52 @@
...
+ * java.lang.String#lastIndexOf(String) String.lastIndexOf} in the JDK. The return type of these
+ * methods is annotated {@code @IndexOfIndexFor(value="this",offset="#1.length()-1")}, meaning that
+ * the returned value is either -1 or it is less than or equal to the length of the receiver
+ * sequence minus the length of the sequence passed as the first argument.
+ *
+ * <p>The annotation {@code @IndexOfIndexFor(args)} is like <code>
+ * {@literal @}{@link NonNegative} {@literal @}{@link LTLengthOf}(args)</code>, except that
+ * {@code @IndexOfIndexFor(args)} additionally permits the value -1.
+ *
+ * <p>The name of this annotation, "indexOf index for", is intended to mean that the annotated
```
mernst: I don't feel that this conveys the intended meaning.  The annotated expression is not necessarily an index for the given sequence, because it might be -1.  The documentation calls it an "indexOf result", which would suggest an annotation named `@IndexOfResult`.  Can you think of a better name than that?

mernst: Another possibility would be something like `@IndexOfSubstring`.

mernst: Do you have any thoughts on naming?

kelloggm: I might call it `SubstringIndex` to extend the naming convention that we used for `SearchIndex`. In general, I'm partial to a name that includes `substring` somehow, because I think that most effectively conveys its intended use.

mernst: `SubstringIndex` is fine with me.  Thanks for the idea.

panacekcz: Thank you, I think it is a good suggestion, so the hierarchy would be `SubstringIndexUnknown`, `SubstringIndexFor(value, offset)` (or alternatively `IndexOfSubstring(value,in)`) and `SubstringIndexBottom`.
Should I rename the whole checker to org.checkerframework.checker.index.substringindex.SubstringIndexChecker?

mernst: I think the names should be consistent, so I wouldn't use both `SubstringIndexUnknown` and `IndexOfSubstring(value,in))`, and I would rename the checker for consistency as well.

panacekcz:
I have renamed the IndexOf checker to Substring Index checker.

kelloggm:
It looks to me as if this PR is ready to merge. @mernst, are there any other outstanding issues with this PR other than the name of the checker? If not, I'll go ahead and merge it.

checker/build.xml
```diff
@@ -1425,7 +1425,7 @@ So, use our own archived version.
     </target>
 
     <target name="download-jdk" description="Downloads jdk8.jar from checkerframework.org/dev-jdk to checker/jdk and checker/dist">
-        <get src="https://checkerframework.org/dev-jdk/jdk8.jar" dest="jdk/jdk8.jar"/>
+        <get src="https://github.com/panacekcz/checker-framework/releases/download/strings-indexof-2/jdk8.jar" dest="jdk/jdk8.jar"/>
```
panacekcz: Don't forget to revert this.

checker/jdk/index/src/java/lang/String.java
```diff
@@ -1732,7 +1732,7 @@ private int lastIndexOfSupplementary(int ch, int fromIndex) {
      * @return  the index of the first occurrence of the specified substring,
      *          or {@code -1} if there is no such occurrence.
      */
-    public @GTENegativeOne @LTEqLengthOf("this") int indexOf(String str) {
+    public @GTENegativeOne @LTEqLengthOf("this") @SubstringIndexFor(value="this",offset="#1.length()-1") int indexOf(String str) {
```
mernst: Why are the `@GTENegativeOne` and `@LTEqLengthOf` annotations needed?  Doesn't the `@SubstringIndexFor` annotation imply them, so that they can be omitted from the JDK source code for brevity and clarity?
(Same comment for other library annotations.)

checker/src/org/checkerframework/checker/index/OffsetDependentTypesHelper.java
```diff
@@ -0,0 +1,89 @@
...
+        if (equation.hasError()) {
+            return equation.getError();
+        }
+        try {
+            equation.standardizeAndViewpointAdaptExpressions(context, localScope, useLocalScope);
+        } catch (FlowExpressionParseUtil.FlowExpressionParseException e) {
+            return new DependentTypesError(expression, e).toString();
+        }
+
+        return equation.toString();
```
mernst: It's surprising that `super.standardizeString` is not called (either before or after the main logic of this method), which is inconsistent with line 35 above.

checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
```diff
@@ -287,14 +289,39 @@ public boolean isLessThanLengthOf(String sequence) {
...
          * @param env ProcessingEnvironment
          * @return the AnnotationMirror that represents this qualifier
          */
-        public AnnotationMirror convertToAnnotationMirror(ProcessingEnvironment env) {
+        public AnnotationMirror convertToAnnotation(ProcessingEnvironment env) {
+            return convertToAnnotation(env, false);
+        }
+
+        /**
+         * Returns the @{@link SubstringIndexFor} AnnotationMirror that represents this qualifier in
```
mernst: The returned AnnotationMirror doesn't represent this qualifier.  Maybe it's an overapproximation to it, or maybe there is some other relationship.  Please clarify what this does.
(It can't be an overapproximation, because @SubstringIndexFor forbids some values that @LTLengthOf permits.  Maybe there is some other invariant about how the values are used that makes this sound, such as the fact that this method is only ever called from SubstringIndexAnnotatedTypeFactory.convertUBQualifierToAnnotation, and that should only be called on annotations that have been created from @SubstringIndexFor annotations?  In any event, please explain those tricky bits of code.)

checker/src/org/checkerframework/checker/index/upperbound/UBQualifier.java
```diff
@@ -287,14 +289,39 @@ public boolean isLessThanLengthOf(String sequence) {
...
+         * the Substring Index hierarchy.
+         *
+         * @param env ProcessingEnvironment
```
mernst: This documentation is not informative (same comment for other methods in the class where the comment is just the type).

mernst:
Can you address the last few comments so that we can merge this?
Thanks!

checker/jdk/index/src/java/lang/String.java
panacekcz: Currently it doesn't. Implementing that for `@GTENegativeOne` should be easy.

checker/src/org/checkerframework/checker/index/OffsetDependentTypesHelper.java
panacekcz: I didn't write this code, but `super.standardizeString` is not called because the work of parsing is done in `equation.standardizeAndViewpointAdaptExpressions` for individual terms. I added comments to explain this.

panacekcz:
@mernst, I have addressed the review comments.
`@SubstringIndexFor` is now an alias for `@GTENegativeOne`.

mernst:
This pull request is not yet ready for review and merging, because the tests are not yet passing.  A commit gate is that plume-lib and daikon should type-check cleanly.  The Index Checker is incomplete, so Daikon has never type-checked, but plume-lib did.  It is failing now, however, due to changes over the summer:  https://travis-ci.org/mernst/plume-lib-check-index-new .  Apparently we forgot to check that job.

Could you please help Martin to get that job passing again (hopefully by just writing annotations, but in the worst case, by suppressing warnings and adding links to issues in the issue tracker)?  Then, you can check that the job will continue to work after this pull request is merged, and ping me about review.  Thanks!

mernst:
I have made https://travis-ci.org/mernst/plume-lib-check-index-new use this branch.
Thus, this branch should not be merged until that job passes, and once this branch is merged, then plume-lib-check-index-new can be merged into plume-lib.

kelloggm:
I manually tested this code against the plume-lib code annotated with Index Checker annotations. It does not cause any regressions. I think this branch should not be held up any longer on plume-lib.

PR #1465, opened 2017-08-21
Fixed annotating fields in stub files that contain an array of primitive types

### LINKED ISSUE #1456: Error parsing a field with array of primitive types in a stub file
Consider following stub file:
`class Myclass {
    @MyAnnotation int[] field;
}
`

It is being parsed by the stubparser project successfully. However, it leads to the ClassCastException in the framework subproject of the Checker Framework.
`[java] error: SourceChecker.typeProcessingStart: unexpected Throwable (ClassCastException); message: com.github.javaparser.ast.type.PrimitiveType cannot be cast to com.github.javaparser.ast.type.ReferenceType; The Checker Framework crashed.  Please report the crash.  To see the full stack trace invoke the compiler with -AprintErrorStack`

Errored Travis job:
https://travis-ci.org/Bohdankm22/sparta/builds/264818673

I think the problem is somewhere in the StubParser.java in the framework subproject. It tries to cast the object to the ReferenceType.
The exception is probably thrown in https://github.com/typetools/checker-framework/blob/8997a341784732c42444ba5a0123e6c514348c0e/framework/src/org/checkerframework/framework/stub/StubParser.java#L763

Checked the ArrayType.java in the stubparser project and it extends ReferenceType. Hence it should be possible to cast ArrayType to the ReferenceType.

Also, need to mention that the issue happens only for the arrays of primitive types.

A possible reason could that it tries to cast to the ReferenceType underlying in the ArrayType type instead of casting the ArrayType itself.

Bohdankm22:
Annotating fields of type array of primitive types when parsing stabfiles will fail due to the issue #1456.
The issue happens when the StubParser tries to annotate the field that contains the array of primitive types. StubParser expects to get the field type which is ArrayType and tries to cast it to the ReferenceType, but StubParser gets an underlying type which is primitive. And primitive type is not a reference type which is why ClassCastException happens when casting.

I have added a test case in https://github.com/typetools/checker-framework/commit/d6c6162987dc41bc84f2d66dfda64d3b6cd48331
It does not matter what annotation is used in stubfile when trying to annotate the field of an array of primitives, it should fail anyway.

I added some description in the last commits.

Fixes #1456


Bohdankm22:
After I added a test case for the issue in https://github.com/typetools/checker-framework/pull/1465/commits/d6c6162987dc41bc84f2d66dfda64d3b6cd48331 Travis job was failed on it:
https://travis-ci.org/Bohdankm22/checker-framework/jobs/266896294

The test was fixed in later commits.

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -858,7 +858,12 @@ private void parseField(
         // StubParser parses all annotations in type annotation position as type annotations
         annotateDecl(declAnnos, elt, decl.getElementType().getAnnotations());
         AnnotatedTypeMirror fieldType = atypeFactory.fromElement(elt);
-        annotate(fieldType, decl.getElementType());
+        // Getting type from the first variable in a line
+        // (all the variables have the same type as they written in one line of code).
```
wmdietl: Can you clarify what the difference between `decl.getElementType()` and the new code is?
Why is the type of the field declaration not the same as the type of the first variable in the field declaration?
The comment about `same type ... in one line of code` is also not correct: `@Foo int @Bar [] a, b @Baz [] , c @Nope [];` is a valid field declaration. `@Foo` should be applied to `int` in all three arrays.

smillst:
@Bohdankm22 Can you please address Werner's comment?  We need this pull request merged in order to complete the stub parser changes.

smillst:
Closing in favor of #1489.

Issue #1466, opened 2017-08-21
Remove stubparser-core.jar and instead add dependency to the stubpaser project


Bohdankm22:
As suggested by @wmdietl https://github.com/typetools/checker-framework/commit/d216f67e1520df198b45e56d419b3f911374faf3#commitcomment-23775365 it is possible to transfer the dependency from the jar file to the stubparser project the same way as it works for annotation-tools and jsr308-langtools dependencies.

wmdietl:
PR: #1476

smillst:
Fixed via https://github.com/typetools/checker-framework/commit/f4444eee9f23d91dbfe7fb06be677b08fff5e3c6.

PR #1467, opened 2017-08-21
Upgrading the stubpaser-core.jar

Bohdankm22:
This pull request should be merged together with https://github.com/typetools/stubparser/pull/6
It should update the stubparser-core.jar lib and add documentation on how to update the stubparser project.


docs/manual/troubleshooting.tex
```diff
@@ -1062,6 +1062,70 @@ \subsection{Code style\label{code-style}}
 in the top-level directory that sets the property \code{maxparallel} to 4000.  If the error is still issued, reduce the number and
 try again.
 
+\section{Upgrading the stubparser module\label{upgrading-stubparser}}
```
wmdietl: I think these instructions are better in typetools/stubparser.
Also, instead of committing new jar files, let's fix #1466 and properly integrate this project.

docs/manual/troubleshooting.tex
```diff
@@ -1062,6 +1062,70 @@ \subsection{Code style\label{code-style}}
...
+\item
+Resolve the conflicts if required and commit it.
+\item
+Run maven tests in the root stubparser directory.
+\begin{Verbatim}
+mvn test
+\end{Verbatim}
+If something failed in tests the stubparser project should be fixed before continuing.
+\item
+Build the stubparser-core.jar file from the javapaser-core module.
```
wmdietl: Details about this step would be great. After `mvn test` I have no jar files.

docs/manual/troubleshooting.tex
```diff
@@ -1062,6 +1062,70 @@ \subsection{Code style\label{code-style}}
...
+\item
+Fork \ahref{https://github.com/typetools/checker-framework}{the stubparser project} to your github account.
+\item
+Clone the Checker Framework repository to your local environment.
```
wmdietl: It would be better if we could have the Checker Framework as a down-stream dependency for the stubparser project.
Then for each change/PR to the stubparser we'll immediately see whether the CF continues to work.

docs/manual/troubleshooting.tex
wmdietl: `mvn package` builds jar files, but none of them is called `stubparser-core.jar`.
Should I take `./javaparser-core/target/javaparser-core-3.3.1-SNAPSHOT.jar`?

wmdietl:
https://github.com/typetools/stubparser/pull/7 integrates stubparser itself better.
Now it should be easy to integrate stubparser like we do for annotation-tools.
Please let me know whether you can do this to finish #1466

wmdietl:
https://github.com/typetools/stubparser/pull/8 updated the stubparser version and #1466 will be merged soon. This can be closed.

Bohdankm22:
Thank you @wmdietl

PR #1468, opened 2017-08-22
Add index upper bound annotations to Math.max

panacekcz:
This PR adds `@PolyUpperBound` annotations to `Math.max` methods, similar to `@PolyLowerBound` annotations that are on `Math.min`.
If both arguments satisfy a certain upper bound, then the result does too.

Fixes panacekcz#20.
The file build.xml is changed to download the modified jdk.


kelloggm:
It doesn't look like GitHub will let me reopen this PR, so I'm going to create another one with the same content. Sorry about my hasty merging before.

PR #1470, opened 2017-08-22
Generalize index upper bound refinement by !=

panacekcz:
This PR generalizes the refinement of upper bounds in the index checker by the `!=` operator.
Currently, if `i` is `@LTEqLengthOf("receiver")` or equivalently `@LTLengthOf(value="receiver",offset="-1")`, then after checking `i != receiver.length`, it is refined to `@LTLengthOf(value="receiver")`.
With this change, if `i` is `@LTLengthOf(value="receiver",offset=c-1)` then after checking `i != receiver.length - c`, it is refined to `@LTLengthOf(value="receiver", offset=c)` for any ~literal~ constant `c`.

Fixes panacekcz#12.
Should not be merged before #1469. ~Tests fail because of kelloggm#167 and should pass after that is merged.~

### DIFF
--- a/checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
+++ b/checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
@@ -23,7 +23,6 @@ import org.checkerframework.dataflow.analysis.TransferResult;
 import org.checkerframework.dataflow.cfg.node.ArrayCreationNode;
 import org.checkerframework.dataflow.cfg.node.AssignmentNode;
 import org.checkerframework.dataflow.cfg.node.FieldAccessNode;
-import org.checkerframework.dataflow.cfg.node.IntegerLiteralNode;
 import org.checkerframework.dataflow.cfg.node.MethodInvocationNode;
 import org.checkerframework.dataflow.cfg.node.Node;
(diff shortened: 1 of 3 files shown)


checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
```diff
@@ -307,14 +308,29 @@ private void refineEq(
...
             Node lengthAccess, Node otherNode, AnnotationMirror otherNodeAnno, CFStore store) {
 
         Receiver receiver = null;
+        // If lengthAccess is "receiver.length - c" where c is an integer literal, stores c
+        // into lengthOffset
+        int lengthOffset = 0;
+        if (lengthAccess instanceof NumericalSubtractionNode) {
+            NumericalSubtractionNode subtraction = (NumericalSubtractionNode) lengthAccess;
+            Node offsetNode = subtraction.getRightOperand();
+            if (offsetNode instanceof IntegerLiteralNode) {
```
kelloggm: Instead of only checking for an `IntegerLiteralNode`, why not use the Value Checker and check if there's a single estimate for the value? That will be even more general.

panacekcz:
Now offsets that are constant according to the value checker are also supported.

PR #1477, opened 2017-08-25
Upgrade error-prone to version 2.1.0

wmdietl:
And fix a few errors it points out.

### DIFF
--- a/framework/src/org/checkerframework/framework/util/typeinference/constraint/FIsA.java
+++ b/framework/src/org/checkerframework/framework/util/typeinference/constraint/FIsA.java
@@ -21,7 +21,7 @@ public class FIsA extends AFConstraint {
     @Override
     protected FIsA construct(
             AnnotatedTypeMirror newArgument, AnnotatedTypeMirror newFormalParameter) {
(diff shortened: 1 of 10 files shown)


framework/src/org/checkerframework/framework/util/typeinference/constraint/FIsA.java
```diff
@@ -21,7 +21,7 @@ public TUConstraint toTUConstraint() {
     @Override
     protected FIsA construct(
             AnnotatedTypeMirror newArgument, AnnotatedTypeMirror newFormalParameter) {
-        return new FIsA(newArgument, newFormalParameter);
+        return new FIsA(newFormalParameter, newArgument);
```
mernst: For consistency and to avoid confusion, how about changing the order of arguments to either the `FIsA` constructor or to `AFConstraint.construct`?

wmdietl: I was thinking about doing that, but that would have required more changes and potential confusion. This was the one location where error-prone complained and it seemed like a valid issue.

PR #1478, opened 2017-08-27
Add missing @Pure annotation for Class.getEnclosingClass()

nikshinde1996:
`Class.getEnclosingClass()` should be annotated as `@Pure`. It returns only class object and has no visible side effects on other variable. Also the returned value is identical (according to` ==`).


wmdietl:
Thanks!

PR #1479, opened 2017-08-28
Fix #1438: nullness checker performance pathology

### LINKED ISSUE #1438: nullness checker performance pathology
The following program takes ~7m30s to compile with version 2.1.14 of the checker framework and the nullness checker enabled, and about 1.5s with it disabled. (The full program is [here](https://gist.githubusercontent.com/cushon/4f6ba1f0269802a9a4784837bbd3780b/raw/ff456f57b2f6d682f5b81d5afdfec1f2e4d06b76/gistfile1.txt).)

The example was boiled down from a real program, but I hope this isn't a very common pattern. Still, there might be something with super-linear performance that's worth optimizing here.

```java
import java.util.HashMap;
import java.util.Map;

public class S {
  static Integer v0;
  static Integer v1;
  static Integer v2;
  static Integer v3;
  // ...
  static Integer v999;

  static Map<String, Integer> f() {
    Map<String, Integer> map = new HashMap<>();
    map.put("k0", v0);
    map.put("k1", v1);
    map.put("k2", v2);
    map.put("k3", v3);
   // ...
    map.put("k999", v999);
    return map;
  }
}
```

Repro:

```
$ curl https://gist.githubusercontent.com/cushon/4f6ba1f0269802a9a4784837bbd3780b/raw/ff456f57b2f6d682f5b81d5afdfec1f2e4d06b76/gistfile1.txt > S.java
$ ~/jsr308/checker-framework-2.1.14/checker/bin/javac S.java
```

bitterfox:
The core cause is that it determines invariant fields by computing declared types of all fields for every method invocation. I've fixed this to remember invariant fields in the store with related optimizations.

The changes in `SourceChecker` is another optimization. Invocation of `getSupportedOptions` could be heavy because it calls `Class#getCanonicalName` and `Class#getSimpleName` so many times and these methods are slow(See an attachment in [JDK-8187123](https://bugs.openjdk.java.net/browse/JDK-8187123) I'm not sure who should cache this. If `BaseTypeChecker` should do, please let me know.

Now the original source in the issue could be compiled in about 30 sec by `NullnessChecker`. I think this is acceptable. For another checker can compile it in about 10 sec like `ValueChecker`. `NullnessChecker` will bring about an execution of KeyForSubchecker.
So we can consider it takes about 15 sec averagely for each checker and this is similar to other checkers.

@wmdietl, in the previous mail, I wrote it's compiled in about 20 sec. However, it had a regression. It became about 30 sec when I fix the regression.


checker/src/org/checkerframework/checker/initialization/InitializationStore.java
```diff
@@ -85,30 +106,22 @@ public void updateForMethodCall(
...
-        // store?
-        List<FlowExpressions.FieldAccess> invariantFields = new ArrayList<>();
-        for (Entry<FlowExpressions.FieldAccess, V> e : fieldValues.entrySet()) {
-            FlowExpressions.FieldAccess fieldAccess = e.getKey();
-            Set<AnnotationMirror> declaredAnnos =
-                    atypeFactory.getAnnotatedType(fieldAccess.getField()).getAnnotations();
-            if (AnnotationUtils.containsSame(declaredAnnos, fieldInvariantAnnotation)) {
-                invariantFields.add(fieldAccess);
-            }
+        // Remove invariant annotated fields for optimization
```
wmdietl: It's not clear what `for optimization` means here. Can you clarify?

checker/src/org/checkerframework/checker/initialization/InitializationStore.java
```diff
@@ -154,12 +167,37 @@ protected boolean supersetOf(CFAbstractStore<V, S> o) {
 
     @Override
     public S leastUpperBound(S other) {
+        // Remove invariant annotated fields for optimization
```
wmdietl: It's not clear what `for optimization` means here. Can you clarify?

framework/src/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -1731,7 +1734,11 @@ public final boolean hasOption(String name) {
     @Override
     public final String getOption(String name, String def) {
 
-        if (!this.getSupportedOptions().contains(name)) {
+        if (supportedOptions == null) {
```
wmdietl: Wouldn't it be better to move this cache into `getSupportedOptions`? Then all callers of the method benefit from the cache. This method would stay unchanged.

checker/src/org/checkerframework/checker/initialization/InitializationStore.java
```diff
@@ -154,12 +167,37 @@ protected boolean supersetOf(CFAbstractStore<V, S> o) {
...
 
+        // Set intersection for invariantFields.
+        for (Entry<FieldAccess, V> e : invariantFields.entrySet()) {
+            if (other.invariantFields.containsKey(e.getKey())) {
+                result.invariantFields.put(e.getKey(), e.getValue());
+            }
+        }
+        // Add invariant annotation again.
+        result.fieldValues.putAll(result.invariantFields);
+
```
wmdietl: Can you go through the other methods that use `initializedFields` and either also handle `invariantFields` or add a comment why this is not needed?
E.g. for `superSetOf` is some logic missing or do we not need to do anything, because it is just an optimization?

For `internalVisualize` you could just add `invariantFields` to the output.

wmdietl:
@bitterfox Will you have a chance to address these comments?

bitterfox:
I've now fixed. Please check this again.

bitterfox:
Now it can be compiled in about 20 sec - 30 sec.
```

[/home/bitter_fox/dev/gsoc/checker-framework]
bitter_fox@bitterfox-MacBookPro% time javac -processor NullnessChecker checker/jtreg/nullness/Issue1438.java                                                                             (git)-[fix_1438]  (m4):S44
Error message is omited
javac -processor NullnessChecker checker/jtreg/nullness/Issue1438.java  54.75s user 1.77s system 239% cpu 23.577 total

[/home/bitter_fox/dev/gsoc/checker-framework]
bitter_fox@bitterfox-MacBookPro% time javac -processor NullnessChecker checker/jtreg/nullness/Issue1438.java                                                                             (git)-[fix_1438]  (m4):S44
Error message is omited
javac -processor NullnessChecker checker/jtreg/nullness/Issue1438.java  72.18s user 1.30s system 257% cpu 28.558 total

[/home/bitter_fox/dev/gsoc/checker-framework]
bitter_fox@bitterfox-MacBookPro% time javac -processor NullnessChecker checker/jtreg/nullness/Issue1438.java                                                                             (git)-[fix_1438]  (m4):S44
Error message is omited
javac -processor NullnessChecker checker/jtreg/nullness/Issue1438.java  51.31s user 1.50s system 227% cpu 23.241 total                                    
```

dataflow/src/org/checkerframework/dataflow/analysis/FlowExpressions.java
```diff
@@ -588,17 +591,17 @@ public boolean equals(Object obj) {
                 return false;
             }
             ClassName other = (ClassName) obj;
-            return getType().toString().equals(other.getType().toString());
+            return typeString.equals(other.typeString.toString());
```
wmdietl: `other.typeString.toString()` can be simplified to `other.typeString`.

framework/src/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -1730,7 +1730,6 @@ public final boolean hasOption(String name) {
      */
     @Override
     public final String getOption(String name, String def) {
-
```
wmdietl: Undo this left-over change.

wmdietl:
I fixed my two small comments.

Issue #1319, opened 2017-06-05
Method reference type inference warning given despite lack of inference


astiob:
This code produces a `methodref.inference.unimplemented` warning (as in #802):

```java
java.util.stream.Stream<String> stream = java.util.stream.Stream.of("foo");
Iterable<String> iterable = stream::iterator;
```

```
warning: [methodref.inference.unimplemented] This version of the Checker Framework does not type-check method references with implicit type arguments.
Iterable<String> iterable = stream::iterator;
                            ^
```

(It does not matter what the stream’s actual element type is or how the stream is obtained.)

I don’t understand how this warning is relevant to this line or why a warning is given at all.

I’ve tried Checker Framework 2.1.12 and 2.1.11, both with Oracle’s javac 1.8.0_40.

PR #1480, opened 2017-08-28
Move jsr308 javac from bootclasspath to normal classpath

CharlesZ-Chen:
This PR had a first round of review by @wmdietl at here:

https://github.com/opprop/checker-framework/pull/47

As CheckerFramework decide to not support java 7 anymore, so it would be good to move jsr308 javac.jar from bootclasspath to normal classpath.

Reason of this moving: Originally put javac.jar on bootclasspath is mainly to be able compile CF under java 7 by overriding normal compiler with jsr308 compiler. Since we don't support java 7 anymore, move javac.jar to normal classpath would be better --- we will only have annotated JDK on the bootclasspath, all other dependencies would be placed on normal classpath.

Changes include in this PR:

- Ant build files: move javac.jar from bootclasspath to classpath for several compilation targets.
- CheckerMain and CheckerDevelMain: move javac.jar to classpath when launching CF.


dataflow/build.xml
```diff
@@ -162,8 +161,7 @@
                  additionalParam="-Xdoclint:all,-missing"
                  failonerror="true"
                  executable="${javadoc.bin}"
-                 classpath="${build}:${javacutil.lib}:${javac.lib}:${javadoc.lib}:${java.home}/../lib/tools.jar"
-                 bootclasspath="${javac.lib}:${javadoc.lib}:${java.home}/lib/rt.jar">
+                 classpath="${build}:${javacutil.lib}:${javac.lib}:${javadoc.lib}:${java.home}/../lib/tools.jar">
```
wmdietl: Can you see whether the `tools.jar` can be removed, here and in `javacutil/build.xml`? javac and javadoc should be the only required dependencies.
Also, `checker/build.xml` mentions `tools.jar` in a comment, but then is not used in the javadoc target.
Hopefully this can be simplified a bit.

framework/tests/all-systems/java8/memberref/MemberReferences.java
```diff
@@ -31,7 +31,7 @@ Object func1(Object o) {
         void context() {
             FunctionMR<Object, Object> f1 = super::func1;
             // TODO: Issue 802: type argument inference
-            //:: warning: (methodref.inference.unimplemented)
+            // TODO: should expect warning: (methodref.inference.unimplemented)
```
wmdietl: @smillst It's not clear to us why the move of javac from the bootclasspath to the normal classpath had an effect on this single expected error.
Do you understand this change in behavior? Do you mind if we disable this warning for now?

smillst: Disabling is fine.  I don't think the code that issues that warning is correct.  See https://github.com/typetools/checker-framework/issues/1319.

smillst:
@CharlesZ-Chen Would you please resolve the merge conflict?

smillst:
@wmdietl  Is this ready to merge?  I haven't reviewed that changes, but I'm fine with merging if you are.

CharlesZ-Chen:
@smillst No problem:) By the way, travis test will failed due to downstream project test, as this PR change some method signatures used in CF Inference project. The travis failure should be resolved once https://github.com/typetools/checker-framework-inference/pull/53 has been merged.

wmdietl:
I merged the CFI PR, so the next rebuild should pass.
@smillst It would be good if you could look at the Gradle changes. They all look good to me, but I'm not quite sure about the gradle setup. I'm for merging as soon as possible, because these are nice cleanups.

docs/examples/GradleExamples/GradleJava7Example/build.gradle
```diff
@@ -53,10 +48,5 @@ allprojects {
                 //'-Awarns',
                 "-Xbootclasspath/p:${configurations.checkerFrameworkAnnotatedJDK.asPath}"
         ]
-        if (targetJavaVersion.isJava7()) {
-            compile.options.compilerArgs += ['-source', '7', '-target', '7']
-            options.fork = true
-            options.forkOptions.jvmArgs += ["-Xbootclasspath/p:${configurations.checkerFrameworkJavac.asPath}"]
```
smillst: This example uses annotations in comments, so it needs to use jsr308 javac.

smillst: Also, this example is duplicated in the manual, so the manual example also needs to be updated.

wmdietl: I think there are two issues:
1. The way that the `targetJavaVersion` is determined with:  `ext.targetJavaVersion = JavaVersion.current().isJava7()` makes it depend on the JDK that is used to run the compiler. We now require Java 8 to run the Checker Framework, so this can never be 7, right?
2. Even if we want the jsr308 javac, we don't need it on the bootclasspath and it already is on the normal classpath.

So don't we need a new way to decide whether to pass `source/target 7` ?

smillst:
I accidentally merged this.  I reverted the merge 02a9a09202e1aab63b9e316daa10c8b87c113965.

docs/examples/GradleExamples/GradleJava7Example/build.gradle
smillst: Right.  So I guess this example should just always pass -source 7 -target 7?

wmdietl: I'm a bit confused why the original code had the conditional check, if this should always be a Java 7 example.
I'm fine with just always adding `-source 7 -target 7`.

smillst: I guess we didn't realize it was using annotations in comments.  I'll make the change.

wmdietl: Annotations in Comments don't need source/target 7.

smillst:
I opened a new pull request since this was merged too soon  #1484.

docs/examples/GradleExamples/GradleJava7Example/build.gradle
smillst: Let's continue this discussion on #1484.

Issue #1459, opened 2017-08-17
Checker Framework gradle setup


zhenqi:
Hi,

I face an issue when add checkerframework in gradle build, I follow the instructions: https://checkerframework.org/manual/#gradle, do all the steps to add these config settings into build.gradle(add all the items from instructions in certain part), and then do a build, here is the error log:

18:40:40 Download http://<company-url>:9999/proximity/repository/public/org/checkerframework/checker/2.1.14/checker-2.1.14.pom
18:40:41 Download http://<company-url>:9999/proximity/repository/public/org/springframework/boot/spring-boot-starter-parent/1.5.4.RELEASE/spring-boot-starter-parent-1.5.4.RELEASE.pom
18:40:44 Download http://<company-url>:9999/proximity/repository/public/org/checkerframework/checker/2.1.14/checker-2.1.14.jar
18:40:44 Download http://<company-url>:9999/proximity/repository/public/org/checkerframework/jdk8/2.1.14/jdk8-2.1.14.pom
18:40:45 Download http://<company-url>:9999/proximity/repository/public/org/checkerframework/jdk8/2.1.14/jdk8-2.1.14.jar
18:40:45 
18:40:45 FAILURE: Build failed with an exception.
18:40:45 
18:40:45 * Where:
18:40:45 Build file '/var/builds/workspace/111592-v3-PR-824/BUILD_CONTAINER/oracle-jdk8-current/label/DOCKER-LOW/app_root/build.gradle' line: 185
18:40:45 
18:40:45 * What went wrong:
18:40:45 A problem occurred evaluating root project '<project-namw>'.
18:40:45 > Cannot change strategy of configuration ':checkerFramework' after it has been resolved.
18:40:45 

I am not sure where this error comes from, does someone have ideas?

Thanks,

Ace

wmdietl:
@zhenqi points out in #1474  a discussion on the gradle issue tracker: https://github.com/gradle/gradle/issues/2747
We should see what the gradle recommendations are and adapt our setup.

Also, with our recent switch to Java 8, there no longer is a reason to put javac on the bootclasspath.
@CharlesZ-Chen is going through all instructions and removes those bootclasspath setups.
So hopefully this issue will have a nice solution with the next release. Sorry for not updating this issue.

zhenqi:
OK, thanks, so is that means the checker cannot be attached into gradle build and work well until the next release?

wmdietl:
[...]

PR #1484, opened 2017-08-30
Move jsr308 javac from bootclasspath to normal classpath

smillst:
~~Continuation of #1480 that was merged before it was complete.~~  I merged the other changes so that master would stop failing its tests.  This pull request is just the changes to the Gradle example.

(Note this should be manually merged and the author changed to @CharlesZ-Chen.)

### DIFF
--- a/docs/examples/GradleExamples/GradleJava7Example/build.gradle
+++ b/docs/examples/GradleExamples/GradleJava7Example/build.gradle
@@ -15,11 +15,6 @@ repositories {
 
 
 configurations {
-    if (targetJavaVersion.isJava7()) {
-        checkerFrameworkJavac {
-            description = 'a customization of the Open JDK javac compiler with additional support for type annotations'
-        }
-    }
     checkerFrameworkAnnotatedJDK {
        description = 'a copy of JDK classes with Checker Framework type qualifers inserted'
     }
@@ -53,10 +48,5 @@ allprojects {
                 //'-Awarns',
                 "-Xbootclasspath/p:${configurations.checkerFrameworkAnnotatedJDK.asPath}"
         ]
-        if (targetJavaVersion.isJava7()) {
-            compile.options.compilerArgs += ['-source', '7', '-target', '7']
-            options.fork = true
-            options.forkOptions.jvmArgs += ["-Xbootclasspath/p:${configurations.checkerFrameworkJavac.asPath}"]
-        }
     }
 }
--- a/docs/examples/GradleExamples/GradleJava7Example/build.gradle
+++ b/docs/examples/GradleExamples/GradleJava7Example/build.gradle
@@ -4,17 +4,15 @@ apply plugin: 'java'
 
 assert (JavaVersion.current().isJava7() || JavaVersion.current().isJava8Compatible())
 
-
-// Indicates the version of Java to which the source and class files should conform
-// If set to JavaVersion.VERSION_1_7 then type annotations must be in comments.
-ext.targetJavaVersion = JavaVersion.current().isJava7() ? JavaVersion.VERSION_1_7 : JavaVersion.VERSION_1_8
-
 repositories {
     mavenCentral()
 }
 
 
 configurations {
+    checkerFrameworkJavac {
+       description = 'a customization of the Open JDK javac compiler with additional support for type annotations'
+    }
     checkerFrameworkAnnotatedJDK {
        description = 'a copy of JDK classes with Checker Framework type qualifers inserted'
     }
@@ -28,9 +26,7 @@ dependencies {
     ext.jdkVersion = JavaVersion.current().isJava7() ? 'jdk7' : 'jdk8'
     checkerFrameworkAnnotatedJDK "org.checkerframework:${jdkVersion}:${checkerFrameworkVersion}"
 
-    if (targetJavaVersion.isJava7()) {
-        checkerFrameworkJavac "org.checkerframework:compiler:${checkerFrameworkVersion}"
-    }
+    checkerFrameworkJavac "org.checkerframework:compiler:${checkerFrameworkVersion}"
     checkerFramework "org.checkerframework:checker:${checkerFrameworkVersion}"
     compile "org.checkerframework:checker-qual:${checkerFrameworkVersion}"
 
@@ -48,5 +44,7 @@ allprojects {
                 //'-Awarns',
                 "-Xbootclasspath/p:${configurations.checkerFrameworkAnnotatedJDK.asPath}"
         ]
+        options.fork = true
+        options.forkOptions.jvmArgs += ["-Xbootclasspath/p:${configurations.checkerFrameworkJavac.asPath}"]
     }
 }
--- a/checker/bin-devel/javac
+++ b/checker/bin-devel/javac
@@ -63,7 +63,7 @@ eval "java" \
      "-DCheckerDevelMain.cp=${buildDirs} " \
      "-DCheckerDevelMain.pp=${buildDirs} " \
      "-DCheckerDevelMain.compile.bcp=${jdkPaths} " \
-     "-DCheckerDevelMain.runtime.bcp=${ltBinDir} " \
+     "-DCheckerDevelMain.runtime.cp=${ltBinDir} " \
      "-DCheckerDevelMain.binary=${binaryDir}" \
      "-classpath ${buildDirs} " \
      "org.checkerframework.framework.util.CheckerDevelMain" \
--- a/checker/build.xml
+++ b/checker/build.xml
@@ -212,7 +212,6 @@
               failonerror="true"
               classpath="${javac.lib}:${framework.lib}:${javaparser.core.lib}:${junit.lib}:${hamcrest.lib}:${annotation-file-utilities.lib}"
               classname="com.sun.tools.javac.Main">
-            <jvmarg value="-Xbootclasspath/p:${javac.lib}"/>
             <arg value="-g"/>
             <!-- Make sure we only have Java 8 source code and generate Java 8 bytecode. -->
             <arg value="-source"/>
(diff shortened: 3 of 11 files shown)


docs/examples/GradleExamples/GradleJava7Example/build.gradle
```diff
@@ -53,10 +44,7 @@ allprojects {
                 //'-Awarns',
                 "-Xbootclasspath/p:${configurations.checkerFrameworkAnnotatedJDK.asPath}"
         ]
-        if (targetJavaVersion.isJava7()) {
-            compile.options.compilerArgs += ['-source', '7', '-target', '7']
-            options.fork = true
-            options.forkOptions.jvmArgs += ["-Xbootclasspath/p:${configurations.checkerFrameworkJavac.asPath}"]
-        }
+        options.fork = true
+        options.forkOptions.jvmArgs += ["-Xbootclasspath/p:${configurations.checkerFrameworkJavac.asPath}"]
```
smillst: The expected errors for this example are:
```
/Users/smillst/jsr308/checker-framework/docs/examples/GradleExamples/GradleJava7Example/src/main/java/org/checkerframework/example/GradleExample.java:33: error: [assignment.type.incompatible] incompatible types in assignment.
        /*@NonNull*/ Object nn = nullable; // error on this line
                                 ^
  found   : @Initialized @Nullable Object
  required: @UnknownInitialization @NonNull Object
/Users/smillst/jsr308/checker-framework/docs/examples/GradleExamples/GradleJava7Example/src/main/java/org/checkerframework/example/GradleExample.java:36: error: [argument.type.incompatible] incompatible types in argument.
        list.add(null); // error on this line
                 ^
  found   : null
2 errors
:compileJava FAILED
```

Without this line the errors are:
```
/Users/smillst/jsr308/checker-framework/docs/examples/GradleExamples/GradleJava7Example/src/main/java/org/checkerframework/example/GradleExample.java:26: error: [assignment.type.incompatible] incompatible types in assignment.
    public static /*@Nullable*/ Object nullable = null;
                                                  ^
  found   : null
/Users/smillst/jsr308/checker-framework/docs/examples/GradleExamples/GradleJava7Example/src/main/java/org/checkerframework/example/GradleExample.java:36: error: [argument.type.incompatible] incompatible types in argument.
        list.add(null); // error on this line
                 ^
  found   : null
2 errors
:compileJava FAILED
```
I don't know understand why this is the case.  @wmdietl any ideas?

wmdietl: The first error message is when annotations-in-comments are correctly used; the second error messages ignore annotations-in-comments.

Is it somehow possible to put `${configurations.checkerFrameworkJavac.asPath}` on the normal classpath for the JVM that runs the compiler?
It shouldn't be necessary to use the bootclasspath, unless that is the only way to change which compiler Gradle uses.
@CharlesZ-Chen can you try this example?

CharlesZ-Chen: Hi @smillst @wmdietl,

I tried running this example by enter directory `docs/examples/GradleExamples/GradleJava7Example` and then run `gradle build`, but I got an strange exception from gradle:

```
java.lang.BootstrapMethodError: call site initialization exception
        at java.lang.invoke.CallSite.makeSite(CallSite.java:341)
        at java.lang.invoke.MethodHandleNatives.linkCallSiteImpl(MethodHandleNatives.java:307)
        at java.lang.invoke.MethodHandleNatives.linkCallSite(MethodHandleNatives.java:297)
        at java.io.ObjectInputStream.<clinit>(ObjectInputStream.java:3578)
        at org.gradle.process.internal.worker.child.SystemApplicationClassLoaderWorker.call(SystemApplicationClassLoaderWorker.java:104)
        at org.gradle.process.internal.worker.child.SystemApplicationClassLoaderWorker.call(SystemApplicationClassLoaderWorker.java:62)
        at worker.org.gradle.process.internal.worker.GradleWorkerMain.run(GradleWorkerMain.java:62)
        at worker.org.gradle.process.internal.worker.GradleWorkerMain.main(GradleWorkerMain.java:67)
Caused by: java.lang.ClassCastException: bootstrap method failed to produce a CallSite
        at java.lang.invoke.CallSite.makeSite(CallSite.java:332)
        ... 7 more
```

I found this post through google, it seems someone tried use Checker Framework with gradle and encountered this problem before:
https://discuss.gradle.org/t/bootstrapmethoderror-with-custom-bootclasspath-in-android/21703

But there were no response to that post:-(

Any ideas on this exception?

Thanks!

CharlesZ-Chen: If I removed this line,

```
options.forkOptions.jvmArgs += ["-Xbootclasspath/p:${configurations.checkerFrameworkJavac.asPath}"]
```

then I got same result as @smillst  got (the case without putting jsr308 javac on bootclasspath).

wmdietl: Do you get this exception only after the change? Does master work?

CharlesZ-Chen: master worked. Below is the result I got on master branch:

```
checker-framework/docs/examples/GradleExamples/GradleJava7Example/src/main/java/org/checkerframework/example/GradleExample.java:26: error: [assignment.type.incompatible] incompatible types in assignment.
    public static /*@Nullable*/ Object nullable = null;
                                                  ^
  found   : null
checker-framework/docs/examples/GradleExamples/GradleJava7Example/src/main/java/org/checkerframework/example/GradleExample.java:36: error: [argument.type.incompatible] incompatible types in argument.
        list.add(null); // error on this line
                 ^
  found   : null
```

wmdietl: Are you using a Java 8 JVM? The version in master doesn't put the 308 compiler on the BCP when on 8.
So it would be nice to find a way to use the 308 compiler on a 7 or 8 JVM, so that in either setting we get annotations-in-comments.

CharlesZ-Chen:
Yes, I used JVM 8. Sure, will have an investigation on this.

Werner Dietl <notifications@github.com>于2017年8月30日 周三22:48写道：

docs/examples/GradleExamples/GradleJava7Example/build.gradle
smillst: I'm not sure about that error, but I used Gradle 3.5 to build the example.

CharlesZ-Chen: Hi Prof. Dietl @wmdietl , I investigated this issue a bit, and I think it should be able to set 308 javac on the classpath when starting the JVM for compilation, and I tried as following:

(Instead of setting javac on bootclasspath, set it to classpath)

```gradle
options.fork = true
options.forkOptions.jvmArgs += ["-classpath", "${configurations.checkerFrameworkJavac.asPath}"]
```

However, the gradle build result still kept same as before. It seems ignore this classpath setting...

I'm not sure if I set the classpath for gradle fork JVM correctly (by adding classpath in `jvmArgs`), but it seems this is the most possible way of doing that after I searching the gradle APIs doc.

Also, I'm not sure if my gradle/ OS env might has some influence on it, as I even cannot get the bootclasspath case working on my machine. (Even I switched to gradle 3.5, I still got the strange bootstrap Exception).

@smillst , could you please try setting `options.forkOptions.jvmArgs += ["-classpath", "${configurations.checkerFrameworkJavac.asPath}"]` on your machine to see if that works? Theoretically it should work as this seems is the same thing we do in `CheckerMain`.

Thanks!

smillst: I tried that, but annotations in comments are still ignored.  I think we should just leave it on the bootclasspath.

wmdietl: I'm fine with keeping jsr308-langtools on the bootclasspath to keep Gradle working. It is rather strange that it works without BCP in our other uses.

wmdietl: One hope of removing it from the BCP was that https://github.com/typetools/checker-framework/issues/1459 would be fixed. Let's just use that issue for future Gradle work.

wmdietl: The problem is that the compiler jar file contains java/lang classes, e.g.:

```
     3037  2017-08-03 09:06   com/sun/tools/javac/util/UnsharedNameTable.class
     2115  2017-08-03 09:06   com/sun/tools/javac/util/Warner.class
      439  2017-08-03 09:06   java/lang/FunctionalInterface.class
     1479  2017-08-03 09:06   java/lang/annotation/ElementType.class
      551  2017-08-03 09:06   java/lang/annotation/Repeatable.class
     1045  2017-08-03 09:06   java/lang/invoke/LambdaConversionException.class
     1463  2017-08-03 09:06   java/lang/invoke/LambdaMetafactory.class
     5068  2017-08-03 09:06   javax/annotation/processing/AbstractProcessor.class
      192  2017-08-03 09:06   javax/annotation/processing/Completion.class
```

We need these classes to run on a Java 7 JVM, because the compiler depends on these classes.
However, when we run on a Java 8 JVM, as we now must, this creates an inconsistency between the Java 8 classes in the rest of the JDK and these few classes that ship with our compiler, causing the strange exception.

I removed these extra classes in jsr308-langtools, so with the next release, we will be able to put our compiler on the bootclasspath and annotations-in-comments should work.
I'm not sure whether there is an easy way to test this before the release.

If we wanted to also support Java 7 execution, we would need to split out these files into a separate module and selectively include it. As we don't care about 7 any longer, we can drop this.
I'll push this change to this PR.

wmdietl: https://github.com/typetools/checker-framework/pull/1487 also includes the required manual updates.

smillst: Thanks for this explanation!

smillst:
Merged #1487  instead.

PR #1489, opened 2017-09-01
Correctly apply annotations in stub files to types of fields.

### LINKED ISSUE #1456: Error parsing a field with array of primitive types in a stub file

smillst:
Fixes #1456.

### DIFF
--- a/checker/jtreg/stubs/issue1456lib/Lib.java
+++ b/checker/jtreg/stubs/issue1456lib/Lib.java
@@ -1,0 +1,8 @@
+package issue1456lib;
+
+public class Lib {
+    public Object object;
+    public byte[] byteArray;
+    public Object object1, object2;
+    public byte byte1, byteArray2[];
+}
--- a/checker/jtreg/stubs/issue1456lib/Lib.java
+++ b/checker/jtreg/stubs/issue1456lib/Lib.java
@@ -1,8 +1,9 @@
 package issue1456lib;
 
 public class Lib {
+    public Object object1, object2;
     public Object object;
     public byte[] byteArray;
-    public Object object1, object2;
     public byte byte1, byteArray2[];
+    public byte[][] byteArray3;
 }
--- a/framework/src/org/checkerframework/framework/stub/StubParser.java
+++ b/framework/src/org/checkerframework/framework/stub/StubParser.java
@@ -871,11 +871,27 @@ public class StubParser {
         annotate(fieldType, fieldVarDecl.getType());
 
         if (fieldType.getKind() == TypeKind.ARRAY) {
-            annotate(((AnnotatedArrayType) fieldType).getComponentType(), decl.getAnnotations());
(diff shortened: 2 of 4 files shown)


framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -858,7 +859,22 @@ private void parseField(
...
+            if (var.getName().toString().equals(elt.getSimpleName().toString())) {
+                fieldVarDecl = var;
+            }
+        }
+        assert fieldVarDecl != null;
+
+        annotate(fieldType, fieldVarDecl.getType());
+
+        if (fieldType.getKind() == TypeKind.ARRAY) {
+            annotate(((AnnotatedArrayType) fieldType).getComponentType(), decl.getAnnotations());
```
wmdietl: Shouldn't this go to the component type that is no longer an array?
I.e. in `@A int [] [] f` I think the current code attaches the annotation on the nested array, instead of on the actual component type. I would have expected a loop here (or maybe we already have a helper for this...).
If that happens in `annotate`, a comment would be helpful.

checker/jtreg/stubs/issue1456lib/Lib.java
```diff
@@ -0,0 +1,8 @@
+package issue1456lib;
+
+public class Lib {
+    public Object object;
+    public byte[] byteArray;
+    public Object object1, object2;
+    public byte byte1, byteArray2[];
```
wmdietl: To test whether my other comment is true, you could add a two-dimensional array.
As a tiny comment, keeping the order of the declaration and the stubfile in sync makes them easier to read.

PR #1490, opened 2017-09-01
Unskip test that was fixed.

### LINKED ISSUE #548: Handling of breaks in try/finally leads to unsoundness
Take the following test case:

```
class TryFinallyBreak {
    String foo() {
        String ans = "x";
        while (true) {
            try {
                ans = "foo";
                break;
            }
            finally {
                ans = null;
            }
        }
        return ans;
    }
}
```

The Nullness Checker should raise an error for the return statement, as the finally branch sets the value to null. However, it currently does not.
The code without 'break' works as expected.

Issue raised in "field initialization vs try block" discussion by Ed Price. Thanks!

Related to Issue #293 and #341 .

smillst:
https://github.com/typetools/checker-framework/commit/0712903c13910dbd7d228730a7232a325b0ad337 also fixed #548.

### DIFF
--- a/checker/tests/initialization/fbc/TryFinallyBreak.java
+++ b/checker/tests/initialization/fbc/TryFinallyBreak.java
@@ -1,6 +1,5 @@
 // Test case for Issue 548:
 // https://github.com/typetools/checker-framework/issues/548
-// @skip-test
 
 class TryFinallyBreak {
     String foo() {


smillst:
Wasn't fixed.

wmdietl:
@bitterfox is preparing another PR to fix #548.

PR #1491, opened 2017-09-01
Correctly read annotations on varargs in stubfiles.

smillst:


framework/.classpath
```diff
@@ -12,5 +12,6 @@
 	<classpathentry combineaccessrules="false" exported="true" kind="src" path="/javacutil"/>
 	<classpathentry combineaccessrules="false" kind="src" path="/annotation-file-utilities"/>
 	<classpathentry combineaccessrules="false" kind="src" path="/scene-lib"/>
+	<classpathentry combineaccessrules="false" kind="src" path="/javaparser-core"/>
```
wmdietl: This change seems unrelated. Adding Eclipse support is a good idea, however I would call the project "stubparser-core" and add the corresponding configuration to the stubparser project.

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -628,6 +628,7 @@ private void parseMethod(
                 // workaround
                 assert paramType.getKind() == TypeKind.ARRAY;
                 annotate(((AnnotatedArrayType) paramType).getComponentType(), param.getType());
```
wmdietl: Isn't it strange to use the component type for one, but the top-level type for the other?

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -628,6 +628,7 @@ private void parseMethod(
                 // workaround
```
wmdietl: Is this change fixing the 'workaround"?

PR #1493, opened 2017-09-04
updating removeIgnoredAnnosFromScene method to remove inner-type annotations

feliperodri:
During whole-program inference, there was a particular case when inner types annotations that should be ignored were not removed from the process. Now, the removeIgnoredAnnosFromScene method checks in all locations (i.e., FIELD, RETURN, RECEIVER, and PARAMETER) whether there are annotations that should be removed from inner types.

@mernst could you please review this pull request? Thanks!

Signed-off-by: feliperodri <rms.felipe@gmail.com>


smillst:
Please fix the test failure.  
```
Setting project property: check.format.stderr -> 
    [apply] Result: 1
    [apply] Applied python to 2640 files and 0 directories.
     [echo] Improper formatting: /checker-framework/framework/src/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenesHelper.java
     [echo] Try running:  ant reformat
Class org.apache.tools.ant.types.resources.selectors.Not loaded from parent loader (parentFirst)
Class org.apache.tools.ant.taskdefs.condition.Equals loaded from parent loader (parentFirst)
     [fail] failing due to condition satisfied
```

mernst:
@feliperodri Can you add a test case that fails before the change and succeeds after?

feliperodri:
@mernst I added the test cases. could you please re-review this pull request? Thanks.

mernst:
The tests are failing again.

It's a good policy to wait for them to pass before you ask for a review.

feliperodri:
Sorry @mernst, my bad! This time all tests are passing. Could you please review the pull request?

framework/src/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenesHelper.java
```diff
@@ -221,15 +221,36 @@ private void removeIgnoredAnnosFromScene(AScene scene) {
         for (AClass aclass : scene.classes.values()) {
             for (AField field : aclass.fields.values()) {
                 removeIgnoredAnnosFromATypeElement(field.type, TypeUseLocation.FIELD);
+                if (field.type.innerTypes.size() != 0) {
```
mernst: Since this identical code appears 4 times, how about ceating a `removeIgnoredAnnosFromATypeElementRecursively` routine that consists of the call to `removeIgnoredAnnosFromATypeElement` and the following 5 lines?

PR #1494, opened 2017-09-05
Fix #1482 by changing an anonymous inner class to a

### LINKED ISSUE #1482: memory leak from AnnotationUtils.annotationMirrorNames cache when using Checker in daemon build process
`AnnotationUtils` uses static fields for various caches, which is undesirable for a number of reasons (see #1108).  Though the state is global, caches like `annotationMirrorNames` have a bounded size of 500, so one might think that memory leaks have been prevented.  Unfortunately, this is not the case.  When running a daemon build process that re-uses the same JVM across multiple invocations of javac, `AnnotationUtils` can end up caching entries from javac invocations that have already completed.  These cache entries are useless, and worse, the `AnnotationMirror` type used as a key for `annotationMirrorNames` can transitively keep alive a huge amount of state.  We have observed this behavior leading to OOM errors in our builds.

The right solution is to get rid of the static state, but that is potentially a huge refactor (though, IMO, a worthwhile one to tackle moving forward).  As a temporary fix, I am wondering if it would be possible to have a different key type for this cache than `AnnotationMirror`, so even if there are stale entries from old javac invocations, they don't keep so much state alive?  I could take a crack at this if no one sees any significant roadblocks.  In the meantime, the only workaround I can see is to call `AnnotationUtils.clear()` periodically.  This will reduce the effectiveness of the cache, though, as we have multiple concurrent instances of javac running in the same JVM.

wmdietl:
static inner class to prevent capturing all enclosing state.


framework/src/org/checkerframework/framework/util/AnnotationBuilder.java
```diff
@@ -97,54 +97,65 @@ private void assertNotBuilt() {
     public AnnotationMirror build() {
         assertNotBuilt();
         wasBuilt = true;
-        return new AnnotationMirror() {
```
smillst: `org.checkerframework.javacutil.AnnotationUtils#fromName` also uses an anonymous inner class.

wmdietl: Yes, but that use is from a static method. So the anonymous class there can't capture an enclosing instance and can't cause an object to stick around.
But I'll look into moving everything into javacutils and having only one class that creates AnnotationMirrors.

smillst: Oh right!

wmdietl:
Downstream failures should be fixed by https://github.com/typetools/checker-framework-inference/pull/58

wmdietl:
@smillst What do you think of this? The interesting change is in https://github.com/typetools/checker-framework/pull/1494/commits/2b76b6b5555071e1cd93b8a4ad74b1e49f40ade5 ; the other changes are just adapting the import/names.

PR #1501, opened 2017-09-07
Updated the stubparser.jar name according to the target jar name chan…

Bohdankm22:
…ges in the stubparser module.

Should be merged right after merging of the https://github.com/typetools/stubparser/pull/9 as the change is related.

### DIFF
--- a/docs/manual/troubleshooting.tex
+++ b/docs/manual/troubleshooting.tex
@@ -930,7 +930,7 @@ This is simply done by:
 \begin{Verbatim}
   cd $JSR308/stubparser
   mvn package -Dmaven.test.skip=true
-  cp -i ./javaparser-core/target/javaparser-core-3.3.4-SNAPSHOT.jar stubparser.jar
+  cp -i ./javaparser-core/target/stubparser.jar stubparser.jar
 \end{Verbatim}
 
 You do not need to add the StubParser to the path, as
--- a/docs/manual/troubleshooting.tex
+++ b/docs/manual/troubleshooting.tex
@@ -934,8 +934,6 @@ This is simply done by:
 
 You do not need to add the StubParser to the path, as
 the Checker Framework build finds it using relative paths.
-However, note that you need to copy the generatreed jar file
-to the stable location \<\$JSR308/stubparser/stubparser.jar>.
 
 
 \subsection{Build the Checker Framework\label{building}}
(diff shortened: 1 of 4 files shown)


docs/manual/troubleshooting.tex
```diff
@@ -930,7 +930,7 @@ \subsection{Build the StubParser\label{stubparser-building}}
 \begin{Verbatim}
   cd $JSR308/stubparser
   mvn package -Dmaven.test.skip=true
-  cp -i ./javaparser-core/target/javaparser-core-3.3.4-SNAPSHOT.jar stubparser.jar
+  cp -i ./javaparser-core/target/stubparser.jar stubparser.jar
```
wmdietl: How about we remove these copies: we only need a version independent path to the jar file. We can just use `./javaparser-core/target/stubparser.jar` instead of the top-level `stubparser.jar`. There is no benefit to copying it now that there is a consistent file name.
It should be enough to update https://github.com/typetools/checker-framework/blob/master/build-common.properties#L42 and then we can just remove these copy commands.

Bohdankm22:
Travis builds for this PR are going to fail till typetools/stubparser#9 is not merged.
I have checked the Travis build on my fork with this changes and it's successfully done.
https://travis-ci.org/Bohdankm22/checker-framework/builds/273868367?utm_source=email&utm_medium=notification

docs/manual/troubleshooting.tex
```diff
@@ -930,12 +930,11 @@ \subsection{Build the StubParser\label{stubparser-building}}
 \begin{Verbatim}
   cd $JSR308/stubparser
   mvn package -Dmaven.test.skip=true
-  cp -i ./javaparser-core/target/javaparser-core-3.3.4-SNAPSHOT.jar stubparser.jar
 \end{Verbatim}
 
 You do not need to add the StubParser to the path, as
 the Checker Framework build finds it using relative paths.
-However, note that you need to copy the generated jar file
+However, note that you need to copy the generatreed jar file
```
wmdietl: Shouldn't these two lines be deleted? `generatreed` isn't a word.

Bohdankm22:
@wmdietl I made the required change. Could we merge both PRs typetools/stubparser#9 and this?

wmdietl:
Done. Thanks for the fixes!

PR #1502, opened 2017-09-07
Handle enclosing types correctly in SuperTypeVisitor.

### LINKED ISSUE #1442: AsSuperVisitor: type is not an erased subtype of supertype
This seems to affect Spring Security.

The crash-inducing calls are `.and().httpBasic()`.

I've uploaded a test case to https://github.com/entonio/AsSuperVisitor.

smillst:
Fixes #1442.

### DIFF
--- a/framework/src/org/checkerframework/framework/type/SupertypeFinder.java
+++ b/framework/src/org/checkerframework/framework/type/SupertypeFinder.java
@@ -165,9 +165,12 @@ class SupertypeFinder {
             while (enclosing != null) {
                 TypeElement enclosingTypeElement =
                         (TypeElement) enclosing.getUnderlyingType().asElement();
+                List<AnnotatedTypeMirror> typeArgs = enclosing.getTypeArguments();
+                List<? extends TypeParameterElement> typeParams =
+                        enclosingTypeElement.getTypeParameters();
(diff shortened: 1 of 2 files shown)


framework/src/org/checkerframework/framework/type/SupertypeFinder.java
```diff
@@ -161,10 +161,17 @@
...
-                AnnotatedTypeMirror typArg = type.getTypeArguments().get(i);
-                TypeParameterElement ele = typeElement.getTypeParameters().get(i);
-                mapping.put(ele, typArg);
+            AnnotatedDeclaredType enclosing = type;
+            while (enclosing != null) {
+                TypeElement enclosingTypeElement =
+                        (TypeElement) enclosing.getUnderlyingType().asElement();
+                for (int i = 0; i < enclosing.getTypeArguments().size(); ++i) {
+                    AnnotatedTypeMirror typArg = enclosing.getTypeArguments().get(i);
+                    TypeParameterElement ele = enclosingTypeElement.getTypeParameters().get(i);
```
wmdietl: Note that `getTypeParameters` allocates a fresh list on every call. I would suggest putting both the type arguments and the type parameters into locals outside the loop.

PR #1503, opened 2017-09-08
Fix #1500: Incorrect initialization and nullness in finally block

### LINKED ISSUE #1500: Incorrect initialization and nullness in finally block
This appears to be a regression sometime between 2.1.14 and 2.2.0.

Consider the following source code:
```java
// Test case for Issue 1500:
// https://github.com/typetools/checker-framework/issues/1500

// @skip-test until the issue is fixed

import java.io.InputStream;

public class TryFinally2 {

    @SuppressWarnings("nullness") // dummy implementation
    Process getProcess() {
        return null;
    }

    void performCommand() {
        Process proc = null;
        InputStream in = null;
        try {
            proc = getProcess();
            in = proc.getInputStream();
            return;
        } finally {
            closeQuietly(in);
            if (proc != null) {
                proc.destroy();
            }
        }
    }

    public static void closeQuietly(final @Nullable InputStream input) {}
}
```

Running the Nullness Checker yields the following false positive warnings:

```
$ch/bin/javac -processor nullness TryFinally2.java
TryFinally2.java:19: error: [argument.type.incompatible] incompatible types in argument.
            closeQuietly(in);
                         ^
  found   : @UnknownInitialization @Nullable InputStream
  required: @Initialized @Nullable InputStream
TryFinally2.java:21: error: [dereference.of.nullable] dereference of possibly-null reference proc
                proc.destroy();
                ^
2 errors
```

There is a test case committed in file `checker/tests/initialization/fbc/TryFinally2.java`.

bitterfox:
When both of `allReachableTrees` and the generated trees contain `null`, unreachable generated trees are not removed and it becomes a cause of the false positive.

### DIFF
--- a/checker/tests/initialization/fbc/TryFinally2.java
+++ b/checker/tests/initialization/fbc/TryFinally2.java
@@ -1,8 +1,6 @@
 // Test case for Issue 1500:
 // https://github.com/typetools/checker-framework/issues/1500
 
-// @skip-test until the issue is fixed
-
 import java.io.InputStream;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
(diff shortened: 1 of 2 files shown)


mernst:
Thanks for the fix!

PR #1514, opened 2017-09-12
Document AnnotatedTypeScanner.

smillst:


framework/src/org/checkerframework/framework/type/visitor/AnnotatedTypeScanner.java
```diff
@@ -112,10 +132,26 @@ protected R scanAndReduce(Iterable<? extends AnnotatedTypeMirror> types, P p, R
...
 
+    /**
+     * Scans {@code type} with the parameter {@code p} and reduces the result with {@code r}.
+     *
+     * @param type type to scan
+     * @param p parameter to use for when scanning {@code type}
+     * @param r result to combine with the result of scanning {@code type}
+     * @return the combination of {@code r} with the result of scanning {@code type}
+     */
     public R scanAndReduce(AnnotatedTypeMirror type, P p, R r) {
```
wmdietl: To document current behavior, shouldn't this `type` also be `@Nullable`?

wmdietl: Unrelatedly, shouldn't this method be protected?

PR #1516, opened 2017-09-13
Add support on the dataflow CFG for visiting lambda result expressions

lazaroclapp:
This adds a new node to the cfg, LambdaResultExpressionNode, which corresponds
to the result of a lambda expression with a single expression as its body, and
which is added at the end of the cfg when CFGBuilder is passed a CFGLambda of
this kind as the underlaying ast.

This is needed by some analyses to handle the result of a lambda expression in
the case where the body is a single ExpressionTree. The case where the body is
a block is already handled by matching ReturnNode.

### DIFF
--- a/dataflow/src/org/checkerframework/dataflow/cfg/node/LambdaResultExpressionNode.java
+++ b/dataflow/src/org/checkerframework/dataflow/cfg/node/LambdaResultExpressionNode.java
@@ -21,20 +21,23 @@ public class LambdaResultExpressionNode extends Node {
     protected ExpressionTree tree;
     protected /*@Nullable*/ Node result;
 
-    public LambdaResultExpressionNode(
-            ExpressionTree t, /*@Nullable*/ Node result, Types types /*,
-            LambdaExpressionTree lambda,
-            Symbol.MethodSymbol methodSymbol*/) {
+    public LambdaResultExpressionNode(ExpressionTree t, /*@Nullable*/ Node result, Types types) {
         super(InternalUtils.typeOf(t));
         this.result = result;
         tree = t;
-        //result.setAssignmentContext(new AssignmentContext.LambdaReturnContext(methodSymbol));
     }
 
+    /**
+     * Returns the final node of the CFG corresponding to the lambda expression body (see {@link
+     * LambdaResultExpressionNode.getTree()}).
+     */
     public Node getResult() {
         return result;
     }
-
+    /**
+     * Returns the {@link ExpressionTree} corresponding to the body of a lambda expression with an
+     * expression body (e.g. X for (o -> X) where X is an expression and not a {...} block).
+     */
     @Override
     public ExpressionTree getTree() {
         return tree;
@@ -55,6 +58,8 @@ public class LambdaResultExpressionNode extends Node {
 
     @Override
     public boolean equals(Object obj) {
+        // No need to compare tree, since in a well-formed LambdaResultExpressionNode, result will be the same only
+        // when tree is the same (this is similar to ReturnNode).
         if (obj == null || !(obj instanceof LambdaResultExpressionNode)) {
             return false;
         }
@@ -67,6 +72,8 @@ public class LambdaResultExpressionNode extends Node {
 
     @Override
     public int hashCode() {
+        // No need to incorporate tree, since in a well-formed LambdaResultExpressionNode, result will be the same only
+        // when tree is the same (this is similar to ReturnNode).
         return HashCodeUtils.hash(result);
     }
 
--- a/dataflow/src/org/checkerframework/dataflow/cfg/node/LambdaResultExpressionNode.java
+++ b/dataflow/src/org/checkerframework/dataflow/cfg/node/LambdaResultExpressionNode.java
@@ -1,0 +1,81 @@
+package org.checkerframework.dataflow.cfg.node;
+
+import com.sun.source.tree.ExpressionTree;
+import java.util.Collection;
+import java.util.Collections;
+import javax.lang.model.util.Types;
+import org.checkerframework.dataflow.util.HashCodeUtils;
+import org.checkerframework.javacutil.InternalUtils;
+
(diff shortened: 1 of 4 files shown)


lazaroclapp:
I am also not sure if there is any other code internally in the checkerframework that needs to be updated to treat `LambdaResultExpressionNode` in ways similar to how `ReturnNode` is currently handled.

dataflow/src/org/checkerframework/dataflow/cfg/node/LambdaResultExpressionNode.java
```diff
@@ -0,0 +1,81 @@
...
+ * @author Lazaro Clapp
+ */
+public class LambdaResultExpressionNode extends Node {
+
+    protected ExpressionTree tree;
+    protected /*@Nullable*/ Node result;
+
+    public LambdaResultExpressionNode(
+            ExpressionTree t, /*@Nullable*/ Node result, Types types /*,
+            LambdaExpressionTree lambda,
```
wmdietl: Can these comments be removed?

dataflow/src/org/checkerframework/dataflow/cfg/node/LambdaResultExpressionNode.java
```diff
@@ -0,0 +1,81 @@
...
+*/
+
+/**
+ * A node for the single expression body of a single expression lambda
+ *
+ * @author Lazaro Clapp
+ */
+public class LambdaResultExpressionNode extends Node {
+
+    protected ExpressionTree tree;
```
wmdietl: Can you add some documentation what the tree and result represent, either here or, maybe better, with the public getter methods?

dataflow/src/org/checkerframework/dataflow/cfg/node/LambdaResultExpressionNode.java
```diff
@@ -0,0 +1,81 @@
...
+    protected /*@Nullable*/ Node result;
+
+    public LambdaResultExpressionNode(
+            ExpressionTree t, /*@Nullable*/ Node result, Types types /*,
+            LambdaExpressionTree lambda,
+            Symbol.MethodSymbol methodSymbol*/) {
+        super(InternalUtils.typeOf(t));
+        this.result = result;
+        tree = t;
+        //result.setAssignmentContext(new AssignmentContext.LambdaReturnContext(methodSymbol));
```
wmdietl: Did you run into any problems with the assignment context? Does this require more work or can this comment be removed?

dataflow/src/org/checkerframework/dataflow/cfg/node/LambdaResultExpressionNode.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == null || !(obj instanceof LambdaResultExpressionNode)) {
+            return false;
+        }
+        LambdaResultExpressionNode other = (LambdaResultExpressionNode) obj;
+        if ((result == null) != (other.result == null)) {
+            return false;
+        }
```
wmdietl: Can you add a comment why the `tree` doesn't need to be compared?

dataflow/src/org/checkerframework/dataflow/cfg/node/LambdaResultExpressionNode.java
```diff
@@ -0,0 +1,81 @@
...
+        return (result == null || result.equals(other.result));
+    }
+
+    @Override
+    public int hashCode() {
+        return HashCodeUtils.hash(result);
```
wmdietl: Similarly, why does the hash only depend on the `result`, not on the `tree`? Maybe this is something to document with the fields, also.

wmdietl:
Regarding other code changes: I tested your changes a bit and everything seemed to work as expected - we already had test coverage for lambdas with single expressions.
I opened one issue for a possible follow-up, #1607, but that shouldn't block this PR.

dataflow/src/org/checkerframework/dataflow/cfg/node/LambdaResultExpressionNode.java
lazaroclapp: Adding it to the methods, let me know if its clear enough (working on the code as I reply to the comments, will update the PR shortly).

dataflow/src/org/checkerframework/dataflow/cfg/node/LambdaResultExpressionNode.java
lazaroclapp: Nope. Removing, this is left-over code from another approach at this.

dataflow/src/org/checkerframework/dataflow/cfg/node/LambdaResultExpressionNode.java
lazaroclapp: You know what, let me add a comparison for `tree` too. It shouldn't matter if we are constructing `LambdaResultExpressionNode`'s correctly always (`result` is supposed to be `scan([path_to_tree], null);`, but we are not enforcing this within the class, so lets both check here and use it to compute the hash)

dataflow/src/org/checkerframework/dataflow/cfg/node/LambdaResultExpressionNode.java
lazaroclapp: See above.

dataflow/src/org/checkerframework/dataflow/cfg/node/LambdaResultExpressionNode.java
lazaroclapp: Although, actually, `ReturnNode`, which is what I am basing this class on, doesn't use `tree` as part of its comparison... mmm...

lazaroclapp:
Rebased and addressed the above comments. Thanks!! :)

PR #1523, opened 2017-09-17
Cleanup caching

### LINKED ISSUE #1482: memory leak from AnnotationUtils.annotationMirrorNames cache when using Checker in daemon build process

wmdietl:
More radical approach to fix #1482.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,15 @@
 ---------------------------------------------------------------------------
 
+Version 2.2.1, September 29, 2017
+
+Deprecated some methods in AnnotatedTypeMirror and AnnotationUtils, to
+be removed after the 2.2.1 release.
+
+Closed issues:
+1482
+
+---------------------------------------------------------------------------
+
 Version 2.2.0, September 5, 2017
 
 A Java 8 JVM is required to run the Checker Framework.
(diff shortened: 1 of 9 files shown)


smillst:
Since this is a user visible change, would you please add a note to the change log?

wmdietl:
I added something to the changelog and also linked the deprecated methods to their replacements: https://github.com/typetools/checker-framework/pull/1523/commits/d4f3ed4b4fe649294ba2aa79f9e6c05b7d332f83
I'm not sure what user visible changes you meant, I assumed you meant the deprecations.

smillst:
Yes I meant the deprecated methods.  Thanks!

PR #1518, opened 2017-09-13
Added handler for simply and fully named classes literals in annotation

Bohdankm22:
Fixes #1496.

PR #1524, opened 2017-09-18
Added handler for simply and fully named classes literals in annotation

Bohdankm22:
Addressed part of the suggested by @wmdietl comments in https://github.com/typetools/checker-framework/pull/1518.


framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1602,4 +1642,8 @@ private void handleExpr(AnnotationBuilder builder, String name, Expression expr)
         faexprcache.put(faexpr, res);
         return res;
     }
+
+    public Map<String, TypeElement> getImportedTypes() {
```
wmdietl: Why add a public accessor for this field? There are just two invocations, which should simply directly access the field.

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -105,6 +106,9 @@
     /** A list of imports that are not annotation types. Used for importing enums. */
     private final List<String> imports;
 
+    /** A map of imported types names and types in a stub file. */
+    private final Map<String, TypeElement> importedTypes;
```
wmdietl: Can you explain what the difference is between `imports` and `importedTypes`? Aren't the elements in the former the keys in the latter? What is different about importing enums vs. class literals? Do we need both?
Why store the `TypeElement` as value, when in the usage of the map you always call `asType()`? Is it to safe the `asType` calls that might never be needed? That is, you probably put more things into the map than will ever be used as class literals later on. So it's easier to store the TypeElements instead of the result of the Types from the `asType()` calls.

smillst:
I incorporated these changes and @wmdietl  feed back into #1547, which more generally fixes expressions in annotations.  Closing this in favor of #1547.

PR #1534, opened 2017-09-21
Annotate Field.get* and Field.set*'s Object param @Nullable

Pr0methean:
It's nullable if the field is static.

### DIFF
--- a/checker/jdk/index/src/java/lang/reflect/Field.java
+++ b/checker/jdk/index/src/java/lang/reflect/Field.java
@@ -1,0 +1,1168 @@
+/*
+ * Copyright (c) 1996, 2013, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package java.lang.reflect;
+
+import sun.reflect.CallerSensitive;
+import sun.reflect.FieldAccessor;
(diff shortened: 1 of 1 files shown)


wmdietl:
But for non-static fields you need a non-null receiver. Our goal is to be conservative and catch potential NPE issues.
We previously discussed this for other reflection methods, e.g. `Method.invoke`: https://github.com/typetools/checker-framework/blob/master/checker/jdk/nullness/src/java/lang/reflect/Method.java#L109

```
    // The method being invoked might be one that requires non-null
    // arguments (including the receiver obj), or might be one that permits
    // null.  We don't know which.  Therefore, the Nullness Checker should
    // conservatively issue a warning whenever null is passed, in order to
    // give a guarantee that no nullness-related exception will be thrown
    // by the invoked method.
```

The same comment applies to these methods in `Field`.
So instead of your change:

```
public Object get(@Nullable Object obj)
```

we would want:

```
public @Nullable Object get(Object obj)
```

with a similar comment about why the receiver is non-null.

Pr0methean:
Okay then, can we maybe have a `@Static` annotation that I can put on Field variables to indicate that they refer to a static field and that the parameter is therefore `@Nullable`? What I don't like is having to annotate `@SuppressWarnings` on the entire class (initializer blocks can't take annotations, at least in JDK 8) just because of one call to `Field.get(null)`. (I'm not sure, in the case of a static field, that the Object parameter even has to be an instance of the class the Field belongs to, but any workaround that relied on assuming otherwise would seem even uglier to me.)

Pr0methean:
We could even add a `castStaticField` helper method to a util class, to check at runtime that the field actually was static. (I'd prefer that over a `getStaticField`, because I'd be able to `-assumenosideeffects` it in Proguard.)

mernst:
Your second proposal of a `@Static` annotation on fields sounds reasonable.
Is this how you would like it to work?

```Field f;
@Static Field sf;
@Nullable Object nullable;
@NonNull Object nonnull;

f.get(nullable);  // warning
f.get(nonnull);   // OK
sf.get(nullable); // OK
sf.get(nonnull);  // OK
```

We would welcome a patch that added that functionality.
If you know what the field is, you could also use/extend the Constant Value Checker instead of a new type system that contains `@Static`.

wmdietl:
I like the `@Static` proposal and @mernst 's examples. This should equally work for the static method case.
Instead of adding this to the Constant Value Checker I would suggest the https://checkerframework.org/manual/#methodval-and-classval-checkers ClassVal and MethodVal Reflection Checker as a suitable place to extend. `@MethodVal` currently describes only the name and number of parameters of a method. `@StaticMethodVal` and `@StaticFieldVal` could describe static methods or field.

Could you change this PR to remove your current `@Nullable` annotations and instead add one to the return type of `getField`? Then we can merge this PR and open a new one for `@Static`.

Pr0methean:
I don't see how the return type of getField would make a difference, since getField itself doesn't return null; you must be talking about a more advanced part of the functionality than I'm familiar with.

wmdietl:
Sorry, I meant `get` should be:

```java
public @Nullable Object get(Object obj)
```

to mark that the field value might be null.
The second parameter to `set` however shouldn't be `@Nullable`, to not allow setting a possibly non-null field to null.

Pr0methean:
Done.

mernst:
Nullness annotations belong in the nullness JDK.  You have added a new file in the index JDK.
In the nullness JDK, the return type of get is [already annotated](https://github.com/typetools/checker-framework/blob/master/checker/jdk/nullness/src/java/lang/reflect/Field.java#L23) as `@Nullable`.
So, I believe that resolves this pull request, since that was the only proposed change.
There is still a feature idea that someone could implement in the future, but we won't keep the issue open just for that.
Thanks!

PR #1536, opened 2017-09-25
Fix #548: Handling of breaks in try/finally leads to unsoundness

### LINKED ISSUE #548: Handling of breaks in try/finally leads to unsoundness

bitterfox:


dataflow/src/org/checkerframework/dataflow/cfg/CFGBuilder.java
```diff
@@ -3397,19 +3402,42 @@ public Node visitConditionalExpression(ConditionalExpressionTree tree, Void p) {
...
+            assert finalliesLabelsStack.contains(l);
+
+            Deque<Object> oldFinalliesLabelsStack = finalliesLabelsStack;
+            finalliesLabelsStack = new LinkedList<>(finalliesLabelsStack);
+            while (!finalliesLabelsStack.isEmpty()) {
+                Object o = finalliesLabelsStack.pop();
+                if (l.equals(o)) {
+                    break;
+                } else if (o instanceof BlockTree) {
+                    BlockTree finallyBlock = (BlockTree) o;
```
wmdietl: Some comment for what happens here would also be good. In particular, what happens with the copied tree?

dataflow/manual/content.tex
```diff
@@ -1141,6 +1141,12 @@ \subsubsection{Default case and fall through for switch statement}
 graph in \autoref{fig:CFGSwitch}. The fake variable named \code{switch#num0}
 is created and each of case nodes creates the branches.
 
+\subsubsection{Break and continue in try block}
+Break and continue statements for while-loop or labeled statements in try
+block will cause executing finally blocks until the target of the jump.
+Dataflow framework will emulate this sort of execution by duplicating
+finally blocks until the target of jump at break and continue statements.
```
wmdietl: These two sentences are not very clear. Let me try to rephrase. Please correct if I got something wrong:

```
Break and continue statements within try blocks will cause execution of the finally block
before continuing at the target of the jump.
The Dataflow Framework models this behavior by duplicating the finally block to the
location before the jump to the target of the break or continue statement.
```

I'm not sure what the `until` meant in both sentences.

dataflow/src/org/checkerframework/dataflow/cfg/CFGBuilder.java
```diff
@@ -4577,6 +4633,8 @@ public Node visitVariable(VariableTree tree, Void p) {
             return node;
         }
 
+        Deque<Object> finalliesLabelsStack = new LinkedList<>();
```
wmdietl: Can you move this field to all the other fields in the class?
Can you make the field private or protected, like other fields?

PR #1537, opened 2017-09-25
checker-qual.jar now uses Java 8 byte code

### LINKED ISSUE #1510: Can we have a checker-qual version with annotations only?
The `checker-qual.jar` file uses [annotations in comments][1], like this code from [NullnessUtil.java][2]:

```java
    public static @EnsuresNonNull("#1") <T extends /*@Nullable*/ Object> /*@NonNull*/ T castNonNull(
            T ref) {
        assert ref != null : "Misuse of castNonNull: called with a null argument";
        return (/*@NonNull*/ T) ref;
    }
```

The standard java 8 compiler just sees `/*@Nullable*/` as a comment. Thus, we must use the special `javac` compiler from the checker-framework.  
To setup the build chain to use it can be tricky (see e.g. the [gradle related issues][3]).

If we had a pure annotation version of the library, like this:

```java
    public static @EnsuresNonNull("#1") <T extends @Nullable Object> @NonNull T castNonNullA(T ref) {
        assert ref != null : "Misuse of castNonNull: called with a null argument";
        return (@NonNull T) ref;
    }
``` 
we would not need the special `javac`, right?

So would you consider to publish an annotation-only version of checker-qual in the future?

  [1]: https://checkerframework.org/manual/#backward-compatibility
  [2]: https://github.com/typetools/checker-framework/blob/549bf8f9621f19bd89876ecd7cf544851d064f74/checker/src/org/checkerframework/checker/nullness/NullnessUtil.java#L73
  [3]: https://github.com/typetools/checker-framework/issues?utf8=%E2%9C%93&q=is%3Aissue%20gradle

smillst:
`checker-qual.jar` used to use Java 7 byte code.  The utility classes, such as `NullnessUtil`, use annotations in comments which are included in the byte code, but ignored by the standard Java 8 compiler.  So, `checker-qual.jar` uses Java 8 byte code so that it can be used with the standard Java 8 compiler.  This fixes #1510. 

Additionally, this adds a new jar, `checker-qual7.jar`, that is exactly the same as `checker-qual.jar`, but uses Java 7 byte code.  The annotations in the utility classes remain in  comments to that this jar can be built.  They are removed from comments when building `checker-qual-sources.jar`.


Pr0methean:
😍 Now I can re-enable Checker Framework for BetterRandom when this releases.

PR #1539, opened 2017-09-25
Correct Vector.copyInto code

smillst:
Don't crash when a method is named `copyInto`, but isn't `Vector#copyInto` or a method that overrides it.  Fixes #1522.

Fixing the above caused an "AsSuper" crash in the case where the Java type of the component  argument isn't a super type of Vector's type argument.  (See line 16 in  checker/tests/nullness/Issue1522.java for an example.) The BaseTypeVisitor now issues a `vector.copyinto.type.incompatible` error in that case.  It could alternatively issue no warning or a warning only if the primary annotations are not in the correct relationship.

Issue #1522, opened 2017-09-16
Crash if method is called `copyInto` and has one parameter


wmdietl:
Take

```java
class CopyInto {
  void copyInto(String p) {}
  void bar() {
    copyInto("Hi");
  }
}
```

and run any type system on the file to get the crash at the bottom of this issue.
In `framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java` the check is only for the method name and number of parameters, ignoring whether the receiver is really a subtype of Vector.
See BaseTypeVisitor#isVectorCopyInto and BaseTypeVisitor#typeCheckVectorCopyIntoArgument.

This situation should be handled more gracefully, but without adding a large performance overhead for each method invocation check.

```
error: SourceChecker.typeProcess: unexpected Throwable (ClassCastException) while processing CopyInto.java; message: org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedDeclaredType cannot be cast to org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedArrayType
  Compilation unit: CopyInto.java
  Exception: java.lang.ClassCastException: org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedDeclaredType cannot be cast to org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedArrayType; Stack trace: org.checkerframework.common.basetype.BaseTypeVisitor.typeCheckVectorCopyIntoArgument(BaseTypeVisitor.java:1165)
  org.checkerframework.common.basetype.BaseTypeVisitor.visitMethodInvocation(BaseTypeVisitor.java:965)
  org.checkerframework.common.basetype.BaseTypeVisitor.visitMethodInvocation(BaseTypeVisitor.java:1)
  com.sun.tools.javac.tree.JCTree$JCMethodInvocation.accept(JCTree.java:1477)
  com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:68)
  org.checkerframework.common.basetype.BaseTypeVisitor.scan(BaseTypeVisitor.java:267)
  org.checkerframework.common.basetype.BaseTypeVisitor.scan(BaseTypeVisitor.java:1)
  com.sun.source.util.TreeScanner.visitExpressionStatement(TreeScanner.java:243)
  com.sun.tools.javac.tree.JCTree$JCExpressionStatement.accept(JCTree.java:1302)
  com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:68)
[...]

PR #1540, opened 2017-09-25
Modernize handling of Vector#copyInto in BaseTypeVisitor

kelloggm:
`BaseTypeVisitor` has special handling for `Vector#copyInto`, because the semantics of that method aren't expressible in Java's type system (as enforced by `javac`). This special case has existed for at least 8 years, based on `git blame`. However, the check used to determine if a given method is actually `Vector#copyInto` just checked if the method was named "copyInto" and that it had one parameter.

Recently, the Index Checker experienced a crash while trying to do a case-study on the JodaTime project that was caused by this bug. JodaTime's `ConverterSet` class contains a method called "copyInto" that takes only one parameter. This caused `BaseTypeVisitor` to trigger the special case, and throw extraneous errors (this other `copyInto` method didn't need special handling...).

This change uses `ElementUtils#isMethod` to check that the method that triggers the special handling is actually either `Vector#copyInto` or a method that overrides it.

**Testing**
A minimized version of the failure from JodaTime is included as both an Index Checker test and an all-systems test. This test fails without this change, but passes with it.

### DIFF
--- a/checker/tests/index/JodaFail.java
+++ b/checker/tests/index/JodaFail.java
@@ -1,0 +1,12 @@
+class Converter {}
+
+class PeriodConverter extends Converter {}
+
+public final class JodaFail {
+
+    void copyInto(Converter[] converters) {}
+
+    public void getPeriodConverters(PeriodConverter[] converters) {
+        copyInto(converters);
+    }
+}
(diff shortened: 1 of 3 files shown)


smillst:
This Issue was reported in #1522.  I just posted a pull request (#1539) that includes the same changes you made, plus additional changes that were required.   I'll add your test case to that branch.

kelloggm:
Ah, sorry I missed that. Sorry to duplicate work, and funny that after many years it happens to two different people within a few weeks...

PR #1541, opened 2017-09-25
Correct inference when type arg is type parameter

### LINKED ISSUE #1526: Failed to infer correct type argument of method invocation
Test case:

```java
public class C {
    public <T> T get(T t) {
            return this.get(t);
    }
}
```

Running Checker Framework (version 2.0) with any checker would raise below type check errors:

```
C.java:3: error: [return.type.incompatible] incompatible types in return.
            return this.get(t);
                           ^
  found   : @UnknownKeyFor Object
  required: T[ extends @UnknownKeyFor Object super @KeyForBottom Void]
C.java:3: error: [return.type.incompatible] incompatible types in return.
            return this.get(t);
                           ^
  found   : @Initialized @Nullable Object
  required: T extends @Initialized @Nullable Object
2 errors
```

It is expected to infer return type of `this.get(t)` to be `T`, but somehow CF infer it as `Object`, result in above unexpected type check errors.

smillst:
For example:

```
    public <T> T get(T t) {
        return this.get(t);
    }
```

Fix this by noting if a type in a constraint is from an argument and therefore should not be considered a "target" type parameter.

Fixes #1526 .


framework/src/org/checkerframework/framework/util/typeinference/constraint/TIsU.java
```diff
@@ -12,6 +12,11 @@ public TIsU(AnnotatedTypeVariable typeVariable, AnnotatedTypeMirror relatedType)
         super(typeVariable, relatedType, 173);
```
wmdietl: Here and for the other new constructors: how about replacing the old constructors with `this(..., false)` to avoid duplicating the magic constants, removing the chance of introducing an inconsistency.

Issue #1543, opened 2017-09-26
SourceChecker.typeProcess: unexpected Throwable (IllegalArgumentException)


CharlesZ-Chen:
Given code:

``` java
class BClass<T> {
}

interface AInterface<T> {
}

class GClass<T extends BClass<?> & AInterface<?>> {
}

class Test {
    GClass gClass;
}
```

Running Checker Framework (version `2.2.0`) with any checker on above code would cause CF crashed. Below is the error stack trace:

```
error: SourceChecker.typeProcess: unexpected Throwable (IllegalArgumentException) while processing deltaTestCase/TypeUtilsTest.java; message: BClass<? extends T>&AInterface<AInterface<? super T>>
  Compilation unit: deltaTestCase/TypeUtilsTest.java
  Exception: java.lang.IllegalArgumentException: BClass<? extends T>&AInterface<AInterface<? super T>>; Stack trace: com.sun.tools.javac.model.JavacTypes.getWildcardType(JavacTypes.java:209)
  org.checkerframework.framework.type.BoundsInitializer.initializeTypeArgs(BoundsInitializer.java:76)
  org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedDeclaredType.getTypeArguments(AnnotatedTypeMirror.java:926)
  org.checkerframework.framework.type.visitor.AnnotatedTypeScanner.visitDeclared(AnnotatedTypeScanner.java:168)
  org.checkerframework.framework.type.AnnotatedTypeFactory$InheritedFromClassAnnotator.visitDeclared(AnnotatedTypeFactory.java:1597)
  org.checkerframework.framework.type.AnnotatedTypeFactory$InheritedFromClassAnnotator.visitDeclared(AnnotatedTypeFactory.java:1555)
  org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedDeclaredType.accept(AnnotatedTypeMirror.java:894)
  org.checkerframework.framework.type.visitor.AnnotatedTypeScanner.scan(AnnotatedTypeScanner.java:104)
  org.checkerframework.framework.type.visitor.AnnotatedTypeScanner.visit(AnnotatedTypeScanner.java:92)
  org.checkerframework.framework.type.AnnotatedTypeFactory.annotateInheritedFromClass(AnnotatedTypeFactory.java:1527)
  org.checkerframework.framework.type.AnnotatedTypeFactory.fromMember(AnnotatedTypeFactory.java:1174)
[...]

PR #1544, opened 2017-09-27
Fix problem with creating wildcards for raw types (Issue 1543)

smillst:
This fixes the first crash reported in #1543.

### DIFF
--- a/framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java
+++ b/framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java
@@ -932,9 +932,7 @@ public abstract class AnnotatedTypeMirror {
                     AnnotatedTypeVariable typeParam =
                             (AnnotatedTypeVariable) declaration.getTypeArguments().get(i);
                     AnnotatedWildcardType wct = (AnnotatedWildcardType) typeArgs.get(i);
-                    if (typeParam.getUpperBound().getKind() == wct.getExtendsBound().getKind()) {
-                        AnnotatedTypeMerger.merge(typeParam.getUpperBound(), wct.getExtendsBound());
-                    }
+                    AnnotatedTypeMerger.merge(typeParam.getUpperBound(), wct.getExtendsBound());
                     wct.getSuperBound().replaceAnnotations(typeParam.getLowerBound().annotations);
                     wct.replaceAnnotations(typeParam.annotations);
                 }
--- a/framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java
+++ b/framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java
@@ -932,7 +932,12 @@ public abstract class AnnotatedTypeMirror {
                     AnnotatedTypeVariable typeParam =
                             (AnnotatedTypeVariable) declaration.getTypeArguments().get(i);
                     AnnotatedWildcardType wct = (AnnotatedWildcardType) typeArgs.get(i);
-                    AnnotatedTypeMerger.merge(typeParam.getUpperBound(), wct.getExtendsBound());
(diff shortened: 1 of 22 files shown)


framework/src/org/checkerframework/framework/type/BoundsInitializer.java
```diff
@@ -73,7 +73,28 @@ public static void initializeTypeArgs(AnnotatedDeclaredType declaredType) {
...
+                                declaredType.atypeFactory.types.getWildcardType(upperBound, null);
+                        break;
+                    case INTERSECTION:
+                        TypeMirror object =
+                                TypesUtils.getObjectTypeMirror(
+                                        declaredType.atypeFactory.processingEnv);
+                        javaTypeArg = declaredType.atypeFactory.types.getWildcardType(object, null);
+                        break;
+                    default:
+                        assert false;
```
wmdietl: Instead of the `assert` could you do an `errorAbort` and remove the next two assignments? I'm not quite sure why these statements are here.

javacutil/src/org/checkerframework/javacutil/TypesUtils.java
```diff
@@ -347,6 +347,13 @@ public static Type wildLowerBound(ProcessingEnvironment env, TypeMirror tm) {
             return TypeAnnotationUtils.unannotatedType(t);
         }
     }
+
+    /** Returns the {@link TypeMirror} for Object. */
```
wmdietl: Let's expand the `Object` to `java.lang.Object` to prevent confusion.

wmdietl: There is an `isObject` method already. Should this method be just `getObject`? Would it help to keep these two methods next to each other?

framework/src/org/checkerframework/framework/type/BoundsInitializer.java
```diff
@@ -73,7 +73,28 @@ public static void initializeTypeArgs(AnnotatedDeclaredType declaredType) {
...
+                    case DECLARED:
+                    case TYPEVAR:
+                        javaTypeArg =
+                                declaredType.atypeFactory.types.getWildcardType(upperBound, null);
+                        break;
+                    case INTERSECTION:
+                        TypeMirror object =
+                                TypesUtils.getObjectTypeMirror(
+                                        declaredType.atypeFactory.processingEnv);
+                        javaTypeArg = declaredType.atypeFactory.types.getWildcardType(object, null);
```
wmdietl: A comment why going to j.l.Object is the right thing to do here would be nice.

framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java
```diff
@@ -932,7 +932,12 @@ public void setTypeArguments(List<? extends AnnotatedTypeMirror> ts) {
                     AnnotatedTypeVariable typeParam =
                             (AnnotatedTypeVariable) declaration.getTypeArguments().get(i);
                     AnnotatedWildcardType wct = (AnnotatedWildcardType) typeArgs.get(i);
-                    AnnotatedTypeMerger.merge(typeParam.getUpperBound(), wct.getExtendsBound());
+                    if (typeParam.getUpperBound().getKind() == wct.getExtendsBound().getKind()) {
+                        AnnotatedTypeMerger.merge(typeParam.getUpperBound(), wct.getExtendsBound());
+                    } else {
+                        wct.getExtendsBound()
```
wmdietl: A comment why we don't need to merge in the other cases would be nice.

wmdietl:
In https://github.com/typetools/checker-framework/pull/1544/commits/9af52bb581244fda8310afcf09f3952b71bb99be I tried fixing the second bug.
I also somehow polluted the PR with extra changes from #1522... not sure why they show up as changes, as they are already merged in master.

Issue #1542, opened 2017-09-26
Stub parser not able to parse integer and long arguments in type annotations


kelloggm:
Consider the following pair of Java classes:

```java
import org.checkerframework.common.value.qual.IntRange;

class NeedsIntRange {
    static int range(boolean big) {
	if (big) {
	    return 20000;
	} else {
	    return 3;
	}
    }
}

class UseIntRange {
    void do_things() {
	@IntRange(from = 3, to = 20000) int x = NeedsIntRange.range(true);
    }
}
```
The first class needs an IntRange annotation in order to permit the value checker to typecheck it. Suppose it is supplied by a stub file:
```java
import org.checkerframework.common.value.qual.IntRange;
class NeedsIntRange {
      @IntRange(from = 3, to = 20000) int range(boolean big);
}
```
Then, executing these two commands results in an error in the stubparser:
```bash
javac NeedsIntRange.java
javac -processor org.checkerframework.common.value.ValueChecker -Astubs=intrange.astub NeedsIntRange.java
```
The error looks like this:
```
error: given value differs from expected; found: java.lang.Integer; expected: java.lang.Long
```
`@IntRange` does require a long, but changing the values in the stub file to longs (which shouldn't actually be required, as integers should be convertible) results in a different error:
```
error: StubParser: unhandled annotation attribute type: 3L class: class com.github.javaparser.ast.expr.LongLiteralExpr
```
With both of these errors present, there doesn't seem to be any way to write an `@IntRange` annotation in a stub file.

wmdietl:
Fixed by https://github.com/typetools/checker-framework/commit/cfba539a5c0573920dd431b36557fd348078d9ab

Issue #1496, opened 2017-09-06
InvocationTargetException: org.checkerframework.framework.source.SourceChecker$CheckerError


maggu2810:
I have created a very small example to reproduce the error.
You can find it here: https://github.com/maggu2810/checker-framework-demo/tree/master/package-info-stub

The problem occurs using a third party jar and a stub file for the package the third party jar is using.
The file that is compiled needs to use a class of the third party jar.

```Java
import org.eclipse.smarthome.automation.handler.BaseModuleHandlerFactory;

public class Test extends BaseModuleHandlerFactory {
}
```

Compilation is done by using the following command:

```text
javac \
    -version \
    -verbose \
    -AprintErrorStack -AprintAllQualifiers \
    -d out \
    -classpath jars/org.eclipse.smarthome.automation.api-0.9.0-SNAPSHOT.jar:jars/checker-2.1.14.jar \
    src/Test.java \
    -processor org.checkerframework.checker.nullness.NullnessChecker \
    -g \
    -target 1.8 -source 1.8 \
    -encoding UTF-8 \
    -Astubs=stubs -AstubWarnIfNotFound -AstubWarnIfOverwritesBytecode -AstubDebug -AshowSuppressWarningKeys \
    -Xbootclasspath/p:jars/org.checkerframework.jdk8-2.1.14.jar
```

The stubs directory contains one file that contains an annotated stub for the package `org.eclipse.smarthome`.

```Java
import org.checkerframework.checker.nullness.qual.Nullable;
import org.checkerframework.framework.qual.DefaultQualifier;

@DefaultQualifier(Nullable.class)
package org.eclipse.smarthome;
```

Without the "-Astubs" the error does not occur.

```text
javac 1.8.0_131
[parsing started RegularFileObject[src/Test.java]]
[parsing completed 13ms]
[search path for source files: jars/org.eclipse.smarthome.automation.api-0.9.0-SNAPSHOT.jar,jars/checker-2.1.14.jar]
[...]

PR #1547, opened 2017-09-28
Correct Stubparser handling of expressions in annotations

smillst:
This fixes both #1542  and #1496.   This uses code written by @Bohdankm22  in #1524 to parse class literals.

### DIFF
--- a/checker/jtreg/stubs/issue1542/UsesIntRange.java
+++ b/checker/jtreg/stubs/issue1542/UsesIntRange.java
@@ -3,6 +3,6 @@ import org.checkerframework.common.value.qual.IntRange;
 class UsesIntRange {
     void do_things() {
         @IntRange(from = 3, to = 20000) int x = NeedsIntRange.range(true);
-        @IntRange(from = 3L, to = 20000) int x = NeedsIntRange.range(true);
+        @IntRange(from = 3L, to = 20000) int y = NeedsIntRange.range(true);
     }
 }
--- a/checker/jtreg/stubs/issue1542/UsesIntRange.java
+++ b/checker/jtreg/stubs/issue1542/UsesIntRange.java
@@ -1,3 +1,5 @@
+package issue1542;
+
 import org.checkerframework.common.value.qual.IntRange;
 
 class UsesIntRange {
--- a/checker/jtreg/stubs/issue1542/UsesIntRange.java
+++ b/checker/jtreg/stubs/issue1542/UsesIntRange.java
@@ -1,8 +1,8 @@
-import issue1542.*;
 import org.checkerframework.common.value.qual.IntRange;
 
 class UsesIntRange {
     void do_things() {
         @IntRange(from = 3, to = 20000) int x = NeedsIntRange.range(true);
+        @IntRange(from = 3L, to = 20000) int x = NeedsIntRange.range(true);
     }
 }
--- a/checker/jtreg/stubs/issue1542/UsesIntRange.java
+++ b/checker/jtreg/stubs/issue1542/UsesIntRange.java
@@ -1,3 +1,4 @@
+import issue1542.*;
 import org.checkerframework.common.value.qual.IntRange;
 
 class UsesIntRange {
--- a/checker/jtreg/stubs/issue1542/UsesIntRange.java
+++ b/checker/jtreg/stubs/issue1542/UsesIntRange.java
@@ -1,0 +1,7 @@
+import org.checkerframework.common.value.qual.IntRange;
+
+class UsesIntRange {
+    void do_things() {
+        @IntRange(from = 3, to = 20000) int x = NeedsIntRange.range(true);
+    }
+}
--- a/framework/src/org/checkerframework/framework/stub/StubParser.java
+++ b/framework/src/org/checkerframework/framework/stub/StubParser.java
@@ -1506,8 +1506,6 @@ public class StubParser {
 
     /*
      * Handles expressions in annotations.
-     * Supports String, int, and boolean literals, but not other literals
-     * as documented in the stub file limitation section of the manual.
      */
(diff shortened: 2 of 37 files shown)


javacutil/src/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -374,7 +374,7 @@ private VariableElement findEnumElement(Enum<?> value) {
         return null; // dead code
     }
 
-    private AnnotationBuilder setValue(CharSequence key, Object value) {
+    public AnnotationBuilder setValue(CharSequence key, Object value) {
```
smillst: I'm going to change this back to private.

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1387,148 +1422,159 @@ private AnnotationMirror getAnnotation(
...
-                                + " and expected: "
-                                + expected);
+                builderSetValue(builder, name, value);
             }
+        }
+    }
+
+    /**
+     * Cast to non-array values so that correct the correct AnnotationBuilder#setValue method is
+     * called.
```
wmdietl: Can you add a comment why this is preferable to making the `AnnotationBuilder#setValue(CharSequence key, Object value)` method visible? I guess it is to be less error-prone with method overloading. Maybe making this a helper method somewhere closer to AnnotationBuilder would be useful?

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1576,8 +1622,8 @@ private void handleExpr(AnnotationBuilder builder, String name, Expression expr)
         }
         TypeElement rcvElt = elements.getTypeElement(faexpr.getScope().toString());
         if (rcvElt == null) {
-            // Search imports for full annotation name.
-            for (String imp : imports) {
+            // Search importedConstants for full annotation name.
+            for (String imp : importedConstants) {
                 String[] import_delimited = imp.split("\\.");
```
wmdietl: Above on 1589 a `StubUtil.partitionQualifiedName` helper method is used. Would it be easier to do the same here, instead of splitting the String?

checker/jtreg/stubs/issue1542/UsesIntRange.java
```diff
@@ -0,0 +1,8 @@
+import issue1542.*;
```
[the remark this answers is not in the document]
smillst: I must have forgotten to git add it.  I have more test classes too, but they are on the computer in my office.

framework/src/org/checkerframework/framework/stub/StubParser.java
smillst: I'm not familiar with this code.  I only renamed the constant here, so I'm not sure.

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1387,148 +1422,159 @@ private AnnotationMirror getAnnotation(
...
+            try {
+                clazz = Class.forName(className);
+            } catch (ClassNotFoundException e) {
+                ErrorReporter.errorAbort("StubParser: unknown class name " + className);
+                throw new Error("dead code; this can't happen");
             }
-        } else if (expr instanceof ArrayInitializerExpr) {
-            ExecutableElement var = builder.findElement(name);
-            TypeMirror expected = var.getReturnType();
+            return TypesUtils.typeFromClass(
```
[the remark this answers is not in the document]
smillst: It's possible, but I don't have time to address this tonight.

smillst:
@wmdietl  I've address the comments I could.  If you have time to review this before tomorrow and approve it, please merge so that the tests can run.  Otherwise, we need to reopen #1496.

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1387,148 +1425,173 @@ private AnnotationMirror getAnnotation(
...
+        } else {
+            ErrorReporter.errorAbort("Unexpected annotation expression: " + expr);
+            return null; // dead code
+        }
+    }
+
+    /**
+     * Converts {@code number} to {@code expectedKind}.
+     * <p>
+     * {@code @interface Anno { long value();})
```
wmdietl: I changed the annotation attribute type to `long`. Otherwise the comment below on line 1500 doesn't make sense to me. If I'm misunderstanding, can you clarify what is happening?

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1387,148 +1425,173 @@ private AnnotationMirror getAnnotation(
...
+            if (expr instanceof NameExpr) {
                 elem = findVariableElement((NameExpr) expr);
+            } else {
+                elem = findVariableElement((FieldAccessExpr) expr);
             }
-
             if (elem == null) {
-                // A warning was already issued by findVariableElement;
-                return;
+                // TODO: should this be stubAlwaysWarn, as it is an input stub file error?
```
wmdietl: I added a few TODO comments for reporting input stub file errors properly. At the moment the user would get a rather scary looking error about a crash in the compiler. If the error is in the input stub file, stubAlwaysWarn seems more appropriate. But I didn't know what to return for that case and didn't want to restructure all this code.

smillst:
@wmdietl Thanks for the improvements!  I'll look into your comments and TODOs after the release.  (It would probably be good to code review and clean up this whole class.)

framework/src/org/checkerframework/framework/stub/StubParser.java
smillst: Yes, `int` should have been `long`. Thanks!

PR #1556, opened 2017-10-02
Change the signature of castNonNull to work around false positive

### LINKED ISSUE #1555: Error when applying castNonNull() to a @MonotonicNonNull type
E.g.

```
import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
import org.checkerframework.checker.nullness.NullnessUtil;

class Test {

    @MonotonicNonNull
    private String x;

    String test() {
        return NullnessUtil.castNonNull(x);
    }
}
```

Results in the following error:

```
Test.java:10: error: [monotonic.type.incompatible] cannot assign org.checkerframework.checker.nullness.qual.MonotonicNonNull to org.checkerframework.checker.nullness.qual.MonotonicNonNull (monotonic type).
        return NullnessUtil.castNonNull(x);
                                        ^
  full type found: @Initialized @MonotonicNonNull String
1 error
```

smillst:
Fixes #1555.

### DIFF
--- a/checker/tests/nullness/Issue1555.java
+++ b/checker/tests/nullness/Issue1555.java
@@ -1,0 +1,11 @@
+import org.checkerframework.checker.nullness.NullnessUtil;
+import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
+
+class Issue1555 {
+
+    @MonotonicNonNull private String x;
+
+    String test() {
+        return NullnessUtil.castNonNull(x);
+    }
+}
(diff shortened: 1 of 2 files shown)


checker/tests/nullness/Issue1555.java
```diff
@@ -0,0 +1,11 @@
+import org.checkerframework.checker.nullness.NullnessUtil;
+import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
+
+class Issue1555 {
+
+    @MonotonicNonNull private String x;
```
wmdietl: For good style I would write `private @MonotonicNonNull String`

checker/tests/nullness/Issue1555.java
```diff
@@ -0,0 +1,11 @@
+import org.checkerframework.checker.nullness.NullnessUtil;
```
wmdietl: Should we stop having the standard issue header with a link to the issue?

Issue #1546, opened 2017-09-28
NPE thrown from TreeUtils.getAssignmentContext() during SourceChecker.typeProcess


CharlesZ-Chen:
Take code:

```java
class Clazz {

    <T> void m(T t) { }

    {
        try {
            new Runnable() {
                public void run() {
                }
            };
        }

        finally {
            m("Hi");
        }
    }
   
}
```

Running Checker Framework `2.2.0` with any checker would cause CF crashes:

```
error: SourceChecker.typeProcess: unexpected Throwable (NullPointerException) while processing C.java
  Compilation unit: C.java
  Exception: java.lang.NullPointerException; Stack trace: org.checkerframework.javacutil.TreeUtils.getAssignmentContext(TreeUtils.java:351)
  org.checkerframework.framework.util.typeinference.TypeArgInferenceUtil.assignedTo(TypeArgInferenceUtil.java:131)
  org.checkerframework.framework.util.typeinference.DefaultTypeArgumentInference.inferTypeArgs(DefaultTypeArgumentInference.java:129)
  org.checkerframework.framework.util.AnnotatedTypes.findTypeArguments(AnnotatedTypes.java:577)
  org.checkerframework.framework.type.AnnotatedTypeFactory.methodFromUse(AnnotatedTypeFactory.java:1984)
  org.checkerframework.framework.type.AnnotatedTypeFactory.methodFromUse(AnnotatedTypeFactory.java:1947)
  org.checkerframework.framework.type.GenericAnnotatedTypeFactory.methodFromUse(GenericAnnotatedTypeFactory.java:1469)
  org.checkerframework.framework.type.TypeFromExpressionVisitor.visitMethodInvocation(TypeFromExpressionVisitor.java:170)
  org.checkerframework.framework.type.TypeFromExpressionVisitor.visitMethodInvocation(TypeFromExpressionVisitor.java:45)
  com.sun.tools.javac.tree.JCTree$JCMethodInvocation.accept(JCTree.java:1477)
  com.sun.source.util.SimpleTreeVisitor.visit(SimpleTreeVisitor.java:54)
  org.checkerframework.framework.type.TypeFromTree.fromExpression(TypeFromTree.java:36)
  org.checkerframework.framework.type.AnnotatedTypeFactory.fromExpression(AnnotatedTypeFactory.java:1202)
  org.checkerframework.framework.type.AnnotatedTypeFactory.getAnnotatedType(AnnotatedTypeFactory.java:976)
[...]

Issue #1520, opened 2017-09-14
Null TreePaths in finally blocks


CharlesZ-Chen:
Below code is a minimum test case:

```java
public class C {
    void start() {
        new Runnable() {
            public void run() {
                try {
                    _run();
                } finally {
                    signal(); //Evaluating this node type as member of implicit `this` will throw NPE
                }
            }
        };
    }

    void signal() {
    }

    void _run() {
    }
}
```

Running Checker Framework (version `2.2.0`) on above code with any checker, will cause a Null Pointer Exception thrown from `AnnotatedTypes.asMemberOfImpl()`.

The reason might be `AnnotatedTypeFactory#getPath(tree)` failed to find the tree path of method invocation tree `signal()` --- which cause `ATF#getImplicitReceiverType(tree)` return `null` for `signal()`, and this cause `AnnotatedTypes.asMemberOfImpl()` throws the NPE.

At first I suspected if `treePathCache` didn't cache path correctly in [`AnnotatedTypeFactory#getPath(tree)`](https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java#L2715), and I tried modify the last return statement from `return treePathCache.getPath(root, node);` to `TreePath.get(root, node)` (just as the way before introducing tree path cacheing), and surprisedly found even `TreePath` failed to find the path of node `signal()`, given the `root` is the complete compilation unit (i.e. the complete CompilationUnitTree of `MultipointServer`).

Consider `TreePath` is a class from package `com.sun.source.util`, so is this a javac bug?

Any ideas are highly appreciated!

wmdietl:
Thanks for the report!
Verified the issue with master and can reproduce the crash.

CharlesZ-Chen:
Update: It seems this crash was introduced recently --- running CF 2.1.14 works fine with above code, while CF 2.2.0 crashes on above code.

wmdietl:
@bitterfox Could you have a look at this issue?

CharlesZ-Chen:
Another similar code seems related to this issue was found:

```java
class TopLevelClass {
    protected class InnerClass implements Runnable {
        public void run() {
            try {
                if (true) {
                    throw new RuntimeException();
                }
            } finally {
                topLevelClassMethod();
            }
        }
    }
    void topLevelClassMethod() {
    }
}
```

Running CF `2.2.0` with any Checker on above code will cause CF crashes:

```
[...]

PR #1557, opened 2017-10-03
Work around for null tree paths in finally blocks

smillst:
This is a work around for issue #1520 and #1546.   It still isn't clear why `AnnotatedTypeFactory#getPath` can't find the tree path in these cases.

### DIFF
--- a/checker/src/org/checkerframework/checker/nullness/KeyForAnnotatedTypeFactory.java
+++ b/checker/src/org/checkerframework/checker/nullness/KeyForAnnotatedTypeFactory.java
@@ -2,8 +2,6 @@ package org.checkerframework.checker.nullness;
 
 import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.NewClassTree;
-import com.sun.source.tree.Tree;
-import com.sun.source.util.TreePath;
 import java.lang.annotation.Annotation;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -17,7 +15,6 @@ import javax.lang.model.element.AnnotationValue;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
-import org.checkerframework.checker.nullness.KeyForPropagator.PropagationDirection;
 import org.checkerframework.checker.nullness.qual.KeyFor;
 import org.checkerframework.checker.nullness.qual.KeyForBottom;
 import org.checkerframework.checker.nullness.qual.PolyKeyFor;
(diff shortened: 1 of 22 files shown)


checker/src/org/checkerframework/checker/nullness/KeyForAnnotatedTypeFactory.java
```diff
@@ -95,7 +96,11 @@ public KeyForAnnotatedTypeFactory(BaseTypeChecker checker) {
         Pair<Tree, AnnotatedTypeMirror> context = getVisitorState().getAssignmentContext();
 
         if (returnType.getKind() == TypeKind.DECLARED && context != null && context.first != null) {
-            AnnotatedTypeMirror assignedTo = TypeArgInferenceUtil.assignedTo(this, getPath(tree));
+            TreePath path = getPath(tree);
```
wmdietl: How about putting this logic, which is duplicated in KeyForPropagationTreeAnnotator, into TypeArgInferenceUtil.assignedTo? If the path is null that method can just return null and then here things would behave the same.

framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
```diff
@@ -171,6 +171,10 @@ protected V getValueFromFactory(Tree tree, Node node) {
         Tree preTree = analysis.getCurrentTree();
         Pair<Tree, AnnotatedTypeMirror> preCtxt = factory.getVisitorState().getAssignmentContext();
         analysis.setCurrentTree(tree);
+        TreePath previousTreePath = factory.getVisitorState().getPath();
+        if (node instanceof MethodInvocationNode) {
```
wmdietl: Can you add a comment why this is only needed for MethodInvocationNode? The assignment context also seems to be set for field updates; should the path also be set?

smillst: (Comment added below this line.)

checker/src/org/checkerframework/checker/nullness/KeyForPropagator.java
```diff
@@ -134,6 +138,31 @@ public void propagate(
         }
     }
 
+    public void propagateNewClassTree(
```
wmdietl: Should this just be default visibility?
Some javadoc might still be nice.

framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
wmdietl: @bitterfox When you look at #1520, can you undo this change?

PR #1560, opened 2017-10-05
Install binutil into dockers.

CharlesZ-Chen:
Currently we are working on integrating Z3 as a new solver to CF Inference, which requires `ar` tool be installed during building phase. Since CF Inference uses the same dockers here to running it's travis test, could we have `binutils` (this is the package provide `ar` tool) be installed in these dockers?

I've tested updated DockerFiles locally and it works.

Thanks a lot!

### DIFF
--- a/checker/bin-devel/Dockerfile-ubuntu-jdk8
+++ b/checker/bin-devel/Dockerfile-ubuntu-jdk8
@@ -20,5 +20,6 @@ RUN apt-get -qqy update && apt-get -qqy install \
   unzip \
   wget \
   default-jdk \
+  binutils \
 && apt-get clean \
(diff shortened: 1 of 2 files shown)


checker/bin-devel/Dockerfile-ubuntu-jdk8
```diff
@@ -20,5 +20,6 @@ RUN apt-get -qqy update && apt-get -qqy install \
   unzip \
   wget \
   default-jdk \
+  binutils \
```
wmdietl: The list of packages seems mostly alphabetically sorted. I would add `binutils` after `ant`, in both files.

PR #1561, opened 2017-10-05
Install build-essential in dockers

CharlesZ-Chen:
Sorry for bothering again. After installed `ar`, I've found `z3` also need c++ compilers be installed in dockers. Could we also install `build-essential` (this provides c++ compiler) into these dockers?

I've tested CFI  travis build script in updated dockers locally, and this time Z3 successfully built. So after merging this PR, I think dockers should be ready for Z3 integration.

Thanks again!

### DIFF
--- a/checker/bin-devel/Dockerfile-ubuntu-jdk8
+++ b/checker/bin-devel/Dockerfile-ubuntu-jdk8
@@ -11,6 +11,7 @@ MAINTAINER Michael Ernst <mernst@cs.washington.edu>
 RUN apt-get -qqy update && apt-get -qqy install \
   ant \
   binutils \
+  build-essential \
   cpp \
   git \
   gradle \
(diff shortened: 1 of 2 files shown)


mernst:
Why are you building Z3?  It might be faster to just download it from https://github.com/Z3Prover/z3/releases/download/z3-4.5.0/z3-4.5.0-x64-ubuntu-14.04.zip .

CharlesZ-Chen:
The reason is for different OS, we need a specific Z3 build result compatible for that OS since Z3 was written in c++. As building Z3 is a one time cost when building CFI, so I just build Z3 instead of checking underlying OS version and download corresponding build results...

mernst:
Building Z3 is not a one-time cost:  it will occur every time that Travis runs.
It's easy to check the operating system and download the proper version.
checker-framework-inference is only building on one OS under Travis, anyway.

CharlesZ-Chen:
Um.. make sense, will change it to check OS version. Thanks for the suggestion!

PR #1562, opened 2017-10-06
Fix a bug with wildcards that extend type variables.

smillst:


framework/src/org/checkerframework/framework/type/DefaultInferredTypesApplier.java
```diff
@@ -110,6 +111,12 @@ private void removePrimaryAnnotationTypeVar(
             TypeMirror typeMirror,
             AnnotationMirror top,
             final AnnotationMirror notInferred) {
+        if (typeMirror.getKind() == TypeKind.WILDCARD) {
```
wmdietl: Can you add a bit of javadoc to this method?
The method name suggests that it should only be called on type variables, but now it is also legal to call on wildcards. Would a different name express this better?
What is the relationship between `type` and `typeMirror`? I'm a bit afraid that the `typeMirror` now changes from a wildcard to a type variable, but the `type` stays a wildcard, which would lead the cast on line 129 to fail. Should the loop update both `typeMirror` and `type`? Is `typeMirror` the underlying type of `type`? Some documentation would help understand this.

framework/src/org/checkerframework/framework/type/DefaultInferredTypesApplier.java
```diff
@@ -106,26 +108,31 @@ private void removePrimaryTypeVarApplyUpperBound(
...
-            AnnotatedTypeMirror type,
-            TypeMirror typeMirror,
+            AnnotatedTypeVariable annotatedTypeVariable,
+            TypeMirror inferredTypeMirror,
             AnnotationMirror top,
-            final AnnotationMirror notInferred) {
-        if (typeMirror.getKind() != TypeKind.TYPEVAR) {
+            AnnotationMirror previousAnnotation) {
+        if (inferredTypeMirror.getKind() == TypeKind.WILDCARD) {
+            // Dataflow might infer a wildcard the extends a type variable for types that are actually
```
wmdietl: `the extends` -> `that extends`

PR #1563, opened 2017-10-06
Add additional override in ValueVisitor to avoid spurious errors

kelloggm:
See kelloggm#171 for a description of this issue: basically, overrides of JDK methods aren't being handled correctly wrt `@IntRangeFromX` annotations.

`BaseTypeVisitor` doesn't actually expose the interface needed to fix the problem, so I made a backwards-compatible change to `BaseTypeVisitor` that splits the `checkOverride` method so that there are two entry points, one of which `ValueVisitor` needs.

### DIFF
--- a/checker/tests/value-index-interaction/MethodOverrides3.java
+++ b/checker/tests/value-index-interaction/MethodOverrides3.java
@@ -1,5 +1,5 @@
 // This class should not issues any errors, since these annotations are identical to the ones
-// on the class in java.io.PrintWriter.
+// on java.io.PrintWriter in the Index JDK.
 
 import java.io.File;
 import java.io.FileNotFoundException;
--- a/checker/tests/value-index-interaction/MethodOverrides3.java
+++ b/checker/tests/value-index-interaction/MethodOverrides3.java
@@ -1,0 +1,17 @@
+// This class should not issues any errors, since these annotations are identical to the ones
+// on the class in java.io.PrintWriter.
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.PrintWriter;
+import org.checkerframework.checker.index.qual.IndexFor;
+import org.checkerframework.checker.index.qual.IndexOrHigh;
+
+public class MethodOverrides3 extends PrintWriter {
+    public MethodOverrides3(File file) throws FileNotFoundException {
+        super(file);
+    }
+
+    @Override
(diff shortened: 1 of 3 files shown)


smillst:
Looks good to me.  @wmdietl Do you want to review the framework change?

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -2417,6 +2417,31 @@ protected boolean checkOverride(
         // Get the type of the overriding method.
         AnnotatedExecutableType overrider = atypeFactory.getAnnotatedType(overriderTree);
 
+        // Call the other version of the method, which takes overrider. The split is
```
wmdietl: Would some version of this comment and an `@see` be useful in the javadoc of this method?
I'm not sure what you mean with the compatibility reasons, can you expand?
Can you add a brief comment about which of the two methods to override in which situation?
Also the `the other method used to include this call` doesn't make sense to me. The other method didn't exist so far, so couldn't include this call.

checker/tests/value-index-interaction/MethodOverrides3.java
```diff
@@ -0,0 +1,17 @@
+// This class should not issues any errors, since these annotations are identical to the ones
+// on the class in java.io.PrintWriter.
```
wmdietl: I found the 'on the class' confusing, as there are no annotations on the `PrintWriter` class.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
kelloggm: I'll add something to the javadoc that differentiates the two and suggests when it is appropriate to call one or the other. 

I agree that this comment could be better explained. By "compatibility" I really meant "backwards-compatibility" - I wanted to make sure that calls to the original method would still behave identically, and this comment is trying to explain that. I'll revise it.

checker/tests/value-index-interaction/MethodOverrides3.java
kelloggm: This comment is referring to the annotations in the Index JDK on `java.io.PrintWriter`. I'll clarify the comment.

kelloggm:
@wmdietl, I've made the changes you requested. Would you mind taking another look? Thanks!

PR #1564, opened 2017-10-06
Restore the filenames option

smillst:


framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -252,6 +252,13 @@ public final Factory getTypeFactory() {
 
     @Override
     public void setRoot(CompilationUnitTree root) {
+        if (checker.hasOption("filenames")) {
```
wmdietl: Is there a reason why you do this in `BaseTypeVisitor` instead of the superclass `SourceVisitor`? The latter would be a bit more general and match better with the options in `SourceChecker`.

smillst: The only sub class for SourceVisitor besides BaseTypeVisitor is org.checkerframework.common.util.count.Locations.Visitor.  There's no reference to the SourceChecker in SourceVisitor.  Is it worth it to add a field to track the checker just for this option?

wmdietl: Ah, good point. How about outputting the visitor name instead of the checker name, that is, using `this.getClass().getSimpleName()`? For most checkers these are the same and for complicated checkers it might actually be interesting to know which of the multiple visitors is running. Then this code can be moved to `SourceVisitor`.

smillst: This code needs the checker to check the option and to print the message.  So, I don't think that will work.

wmdietl: True... how about directly adding the message in `SourceChecker`, before/after the call to `setRoot`? https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/source/SourceChecker.java#L957

smillst: Done.

PR #1566, opened 2017-10-08
Version of TreeUtils.getMethod that checks parameter types

mernst:
Also adds error-checking to version that takes a number of parameters


javacutil/src/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -734,6 +734,42 @@ public static ExecutableElement getMethod(
...
+        TypeElement typeElt = env.getElementUtils().getTypeElement(typeName);
+        for (ExecutableElement exec : ElementFilter.methodsIn(typeElt.getEnclosedElements())) {
+            if (exec.getSimpleName().contentEquals(methodName)
+                    && exec.getParameters().size() == paramTypes.length) {
+                boolean typesMatch = true;
+                List<? extends VariableElement> params = exec.getParameters();
+                for (int i = 0; i < paramTypes.length; i++) {
+                    VariableElement ve = params.get(i);
+                    TypeMirror tm = ve.asType();
+                    // TODO: Is there a better way to do this?
```
wmdietl: You could use `TypesUtils.isDeclaredOfName(tm, paramTypes[i])` instead of comparing the strings yourself.

javacutil/src/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -702,29 +704,27 @@ public static boolean isMethodInvocation(
 
     /**
      * Returns the ExecutableElement for a method declaration of methodName, in class typeName, with
-     * params parameters.
-     *
-     * <p>TODO: to precisely resolve method overloading, we should use parameter types and not just
-     * the number of parameters!
+     * params parameters. Errs if there is more than one matching method.
```
wmdietl: Now that the TODO is resolved, how about adding a cross-reference to the method that takes the parameter types and explaining the trade-offs?

wmdietl: More precisely: `Errs if there is no match or more than one matching method.`

javacutil/src/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -734,6 +734,42 @@ public static ExecutableElement getMethod(
     }
 
     /**
+     * Returns the ExecutableElement for a method declaration of methodName, in class typeName, with
+     * parameters of the given types.
```
wmdietl: How about adding `Errs if there is no matching method.` to make explicit that the method never returns null.

javacutil/src/org/checkerframework/javacutil/TreeUtils.java
wmdietl: Sorry, that doesn't work for primitive types. So you either continue doing the string comparison yourself or add a new utility method to `TypesUtils` that works for both declared and primitive types.

PR #1567, opened 2017-10-08
Add @InternedDistinct annotation

mernst:


checker/src/org/checkerframework/checker/interning/qual/InternedDistinct.java
```diff
@@ -0,0 +1,24 @@
...
+ * <p>This is a stronger property than {@link Interned}, but a weaker property than every value of a
+ * Java type being interned.
+ *
+ * @see org.checkerframework.checker.interning.InterningChecker
+ * @checker_framework.manual #interning-checker Interning Checker
+ */
+@SubtypeOf(Interned.class)
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE_USE, ElementType.TYPE_PARAMETER})
```
wmdietl: As this is the new bottom in the interning type hierarchy, the `@DefaultFor(value = {TypeUseLocation.LOWER_BOUND})` should be moved from Interned to here.
It probably also makes sense to make InternedDistinct implicitfor `typeNames = {Void.class}`, instead of leaving that with Interned.

mernst: Fixed in commit bd5f660

PR #1568, opened 2017-10-08
Handle string conversion between internal form and binary name

mernst:


mernst:
Review this after #1566 has been merged

PR #1571, opened 2017-10-09
Only store erased and boxed types in annotations.

### LINKED ISSUE #1565: Problem writing stubs for generic classes in 2.2.1
I tried switching from 2.1.14 to 2.2.1 but I have run into a problem rewriting my stubs for the new parser, which I think may be a bug (or at the least, could do with a manual section explaining my mistake).  I've made a minimal working example showing my problem.  I've made two classes (attached): MyListGeneric and MyBitSet which are identical except that the former extends a generic class, AbstractList, while the latter extends BitSet which is not generic.  Both classes call clear() in the constructor, and by default, both get a compile error about the initialization state of the super-class:

```
[ERROR] MyListGeneric.java:[7,14] [method.invocation.invalid] call to clear() not allowed on the given receiver.
[ERROR] found   : @UnderInitialization(MyListGeneric<T>.class) @NonNull AbstractList
[ERROR] required: @Initialized @NonNull AbstractList
[ERROR] MyBitSet.java:[7,14] [method.invocation.invalid] call to clear() not allowed on the given receiver.
[ERROR] found   : @UnderInitialization(MyBitSet.class) @NonNull BitSet
[ERROR] required: @Initialized @NonNull BitSet
```

Which is as I expect.  However, I then add what I think are the required annotations in a stub file (attached, but reproduced here):

```

package java.util;
class AbstractList<T>
{
    void clear(@UnknownInitialization(java.util.AbstractList.class) AbstractList<T> this);
}
class BitSet
{
    void clear(@UnknownInitialization(java.util.BitSet.class) BitSet this);
}
```

(Incidentally: might be worth mentioning in the manual that the class literals need explicit qualifying, unlike all other type uses in the file.)  This fixes the error in MyBitSet, but MyListGeneric still complains:

```
[ERROR] MyListGeneric.java:[7,14] [method.invocation.invalid] call to clear() not allowed on the given receiver.
[ERROR] found   : @UnderInitialization(MyListGeneric<T>.class) @NonNull AbstractList
[ERROR] required: @UnknownInitialization(java.util.AbstractList<E>.class) @NonNull AbstractList
```

I'm fairly sure that the error is illogical.  MyListGeneric<T> extends AbstractList<T> so the requirement is satisfied.  It looks to me like a generic erasure issue: I suspect the T is somehow not lining up with the E reported in the error?  Especially since the non-generic analogue in MyBitSet compiles fine.

Files in attached zip should form a complete Maven project: MyBitSet.java and MyListGeneric.java in src/main/java/, pom.xml in top level, and MyStub.astub which in stubs/
[testchecker.zip](https://github.com/typetools/checker-framework/files/1366462/testchecker.zip)

smillst:
Fixes #1565.


javacutil/src/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -215,8 +215,16 @@ public AnnotationBuilder setValue(CharSequence elementName, String value) {
         return setValue(elementName, (Object) value);
     }
 
+    private TypeMirror getErasedAndBoxedType(TypeMirror type) {
```
wmdietl: Should this be `getErasedOrBoxedType`?

PR #1574, opened 2017-10-09
StubParser: Correctly parse class literals that aren't fully qualified.

### LINKED ISSUE #1570: StubParser: class literals should not require fully qualified names
As reported in #1565, if a class literal references a class declared in the current package, it should not have to be fully qualified.  For example:

```

package java.util;
class AbstractList<T>
{
    void clear(@UnknownInitialization(AbstractList.class) AbstractList<T> this);
}
class BitSet
{
    void clear(@UnknownInitialization(BitSet.class) BitSet this);
}
```

smillst:
Also, adds `-Werrors` and ` -AstubWarnIfNotFound` to all jtreg stub file tests.

Fixes #1570.


framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1506,6 +1507,55 @@ private Object getValueOfExpressionInAnnotation(
...
+        while (!enclosingClass.equals(packageName)) {
+            typeElement = elements.getTypeElement(enclosingClass + "." + name);
+            if (typeElement != null) {
+                importedTypes.put(enclosingClass + "." + name, typeElement);
+                return typeElement;
+            }
+
+            int lastDot = enclosingClass.lastIndexOf('.');
+            if (lastDot == -1) {
+                enclosingClass = "";
```
wmdietl: When is `""` useful as value for `enclosingClass`? As it will later be appended with `"."`, no legal class name can be produced. Instead, should this break out of the loop?

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1506,6 +1507,55 @@ private Object getValueOfExpressionInAnnotation(
     }
 
     /**
+     * Returns the TypeElement with the fully qualified name {@code name}, if one exists. Otherwise,
+     * checks {@code enclosingClass} and {@code packageName} for such a class with {@code name}.
+     *
+     * @param packageName name of package being parsed
+     * @param enclosingClass fully-qualified name of the class being parsed
+     * @param name classname (Simple or fully qualified)
+     * @return the TypeElement for {@code name}
```
wmdietl: Add `or null if not found`. And, if this class uses nullness annotations, let's add `@Nullable` to the return type...

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1506,6 +1507,55 @@ private Object getValueOfExpressionInAnnotation(
...
+     */
+    private TypeElement findTypeOfName(String packageName, String enclosingClass, String name) {
+        TypeElement typeElement = elements.getTypeElement(name);
```
wmdietl: It's a bit annoying that these 5 lines are repeated 4 times. How about rewriting it as:

```java
private TypeElement checkCandidate(String name) {
    TypeElement typeElement = elements.getTypeElement(name);
     if (typeElement != null) {
         importedTypes.put(name, typeElement);
     }
     return typeElement;
}

private TypeElement findTypeOfName(String packageName, String enclosingClass, String name) {
    TypeElement typeElement = checkCandidate(name);
    while (typeElement == null && !enclosingClass.equals(packageName)) {
        typeElement = checkCandidate(enclosingClass + "." + name);
        int lastDot = enclosingClass.lastIndexOf('.');
        if (lastDot == -1) {
            break;
        } else {
            enclosingClass = enclosingClass.substring(0, lastDot);
        }
    }
    if (typeElement == null && !packageName.isEmpty()) {
        typeElement = checkCandidate(packageName + "." + name);
    }
    if (typeElement == null && !packageName.equals("java.lang")) {
        typeElement = checkCandidate("java.lang." + name);
    }
    return typeElement;
}
````

Maybe a few more null checks, but less repetition and only a single return.

PR #1577, opened 2017-10-10
Index JDK changes based on mistakes found while type-checking plume-lib

kelloggm:
This PR includes four changes to the Index JDK that came up while type-checking plume-lib. One is mentioned in kelloggm#164 and another in kelloggm#172. The other two are that Appendable and CharSequence confused IndexFor and IndexOrHigh annotations. I fixed the errors. 

The PR is currently downloading its JDK from a non-standard location. That will need to be fixed before merging.


checker/jdk/index/src/java/lang/Appendable.java
```diff
@@ -105,7 +105,7 @@
      * @throws  IOException
      *          If an I/O error occurs
      */
-    Appendable append(CharSequence csq, @IndexOrHigh("#1") int start, @IndexOrHigh("#1") int end) throws IOException;
+    Appendable append(CharSequence csq, @IndexFor("#1") int start, @IndexOrHigh("#1") int end) throws IOException;
```
panacekcz: This was not a mistake. It is `@IndexOrHigh`, because `csq.length()` is a valid value for `start` if `end` is also `csq.length()`.  The case `start=csq.length()` is only invalid if `start>end`, which can happen for other values of `start` as well and it is a property which is not checked by the index checker.
On the other hand, `@IndexFor` would report false positives in code such as:
```java
void m(Appendable a, @MinLen(10) CharSequence csq){
  a.append(csq, 10, csq.length());
}
```

kelloggm: You're right. I misunderstood the documentation, and I'll revert this change. Thanks for pointing this out.

kelloggm:
@panacekcz, would you be willing to validate the rest of these changes? I removed the ones you pointed out were incorrect.

panacekcz:
@kelloggm, regarding kelloggm#172, I generally agree with this change, but for consistency it should be also applied where the methods are overriden, such as in `OutputStreamWriter`. 

Additionally, there are other occurences of the same pattern in the jdk, for example `OutputStream.write`. The same reasoning should apply there as well, unless there is something that makes it different. So if we really want consistency, this would be a big change. I don't know if you want to make that change, there might be cases where `@IndexFor` is more practical than `@IndexOrHigh`.
And then kelloggm#170 makes it even more complicated.
I have written my thoughts about annotating index ranges in panacekcz#5. I think we should have some guideline on how to annotate these ranges and put that into the manual.

I am not sure if I am any help here, but I'd say, in this PR, do the same change to all subclases of `Writer` and it should be ok. The other things can be sorted out later.

kelloggm:
I agree that doing the subclasses of `Writer` should go in this PR.

kelloggm:
@panacekcz I updated all the subclasses of `Writer`. I think we should leave the rest out of this PR. Would you mind re-reviewing? Thanks.

PR #1581, opened 2017-10-13
Update Checker Framework version in build.gradle

utsavoza:


mernst:
Thanks for noticing this.

Issue #1585, opened 2017-10-15
unhelpful stub parser diagnostic


cushon:
The following annotation was being tolerated by the stub file parser before version 2.2.0:

`@DefaultQualifier(value = Nullable, locations = DefaultLocation.UPPER_BOUNDS)`

Afterwards it was rejected (which is good!), but it failed with a stack trace that didn't include the file name or source position, so I had to turn on `-AstubDebug` to figure out what happened. I'd be nice if it reported a friendlier error.

```
error: InvocationTargetException when invoking constructor for class org.checkerframework.checker.nullness.KeyForAnnotatedTypeFactory; Underlying cause: Field not found: Nullable
  Exception: java.lang.reflect.InvocationTargetException; Stack trace: sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
  sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
...
  Underlying Exception: org.checkerframework.framework.source.SourceChecker$CheckerError: Field not found: Nullable; Stack trace: org.checkerframework.framework.source.SourceChecker.errorAbort(SourceChecker.java:727)
  org.checkerframework.javacutil.ErrorReporter.errorAbort(ErrorReporter.java:25)
  org.checkerframework.framework.stub.StubParser.getValueOfExpressionInAnnotation(StubParser.java:1439)
  org.checkerframework.framework.stub.StubParser.handleExpr(StubParser.java:1552)
  org.checkerframework.framework.stub.StubParser.getAnnotation(StubParser.java:1405)
  org.checkerframework.framework.stub.StubParser.annotateDecl(StubParser.java:959)
  org.checkerframework.framework.stub.StubParser.parseMethod(StubParser.java:642)
```

wmdietl:
Thanks for the comment. We had noticed these issues shortly before the last release and fixed them in #1554, so this particular issue is already fixed in master.
I've reviewed that code some more and improved the reporting for parse errors in #1588.

PR #1588, opened 2017-10-15
Add test for #1585. Ensure syntactically invalid astub files are

wmdietl:
reported in a user-friendly way.


framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1370,20 +1385,6 @@ private void stubDebug(String warning) {
     }
 
     /**
-     * Throw an error if the message contains line separator characters. (Any characters after a
-     * line separator are silently discarded and not shown to the user.)
-     */
-    private void ensureSingleLine(String message, String methodName) {
```
wmdietl: I can't reproduce this issue. I ran the test cases manually and correctly see multi-line warnings.
Some existing uses of stub messages contained newlines, which would have resulted in this rather bizarre error.
I removed all this code. Please let me know if you know why this should be kept.

mernst: @wmdietl This was definitely a problem and I believe the code should be retained.

Maybe it's Maven that only prints the first line (Maven does some other violence to error messages as well).  If so, the comment, when reinstated, should be more specific about what tool is at fault.  (It should have been in the first place; sorry about that.)

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -177,24 +178,34 @@ public StubParser(
         try {
             parsedStubUnit = JavaParser.parseStubUnit(inputStream);
         } catch (ParseProblemException e) {
-            ErrorReporter.errorAbort(
-                    "StubParser: exception from StubParser.parse for file "
+            StringBuilder message =
+                    new StringBuilder(
+                            "exception from StubParser.parse for file "
```
smillst: After #1574 is merged, you could add the class currently being parsed.

framework/src/org/checkerframework/framework/stub/StubParser.java
wmdietl: @mernst Why would that only be a problem here? We don't enforce this for any of the other warnings or errors that we issue. The warnings from the parsing errors will be much easier to read when they contain newlines.

mernst: It might be a problem elsewhere as well.  I encountered it here.  In the future we could be consistent throughout.  Or, if the problem is related to Maven, we could convince the Maven maintainers to output the whole message instead of just part.

wmdietl: I have verified that the multi-line StubParser messages, both at WARNING and at NOTE level, are correctly output in the Maven build.
This PR should be merged. If another tool has problems with error messages this should be fixed for all output, not just the stubparser.

framework/src/org/checkerframework/framework/stub/StubParser.java
mernst: #1574 has been merged

wmdietl: I merged in master and output the current class below.

PR #1591, opened 2017-10-16
Refactor postcondition code.

smillst:
So that both postconditions and conditional postconditions use the same code.  But keep the override points seperate.

### DIFF
--- a/framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
+++ b/framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
@@ -1009,15 +1009,13 @@ public abstract class CFAbstractTransfer<
             }
 
             try {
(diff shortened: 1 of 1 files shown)


framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
```diff
@@ -1033,18 +1017,21 @@ protected void processConditionalPostconditions(
                                 flowExprContext,
                                 analysis.atypeFactory.getPath(tree),
                                 false);
-                if (result) {
-                    thenStore.insertValue(r, anno);
+                if (p instanceof ConditionalPostcondition) {
```
wmdietl: It would be a bit cleaner & faster to compare the kind field against the enum.

PR #1595, opened 2017-10-17
Add a different error key for parsing postconditions at call sites.

### LINKED ISSUE #1582: private fields and flowexpr.parse.error
Suppose we have

```java
package foo;

import org.checkerframework.checker.nullness.qual.EnsuresNonNullIf;

public class Foo {

  @Nullable Object theObject;

  Foo(@Nullable Object theObject) {
    this.theObject = theObject;
  }

  @SuppressWarnings("contracts.conditional.postcondition.not.satisfied")
  @EnsuresNonNullIf(
    expression = {"theObject", "getTheObject()"},
    result = true
  )
  public boolean hasTheObject() {
    return theObject != null;
  }

  @Nullable
  public Object getTheObject() {
    return theObject;
  }
}
```

```java
package mainrepropkg;

import foo.Foo;

public class FlowExprParseError {

  public void printAThing(Foo foo) {
    if (foo.hasTheObject()) {
      System.out.println("Print false: " + foo.getTheObject().equals(new Object()));
    }
  }
}
```

private fields (like `Foo.theObject`) usually do not affect separate compilation. Bazel relies on that fact to optimize builds, and prunes private fields and methods from the classes that appear on the compile-time class path (see [1]). If `Foo` and `FlowExprParseError` are compiled separately, the version of `Foo` on `FlowExprParseError`'s classpath does not contain the field `theObject`, and the checker framework reports an error:

```
mainrepropkg/FlowExprParseError.java:8: error: [flowexpr.parse.error] cannot parse the expression 'theObject: identifier not found'
    if (foo.hasTheObject()) {
                        ^
1 error
```

It would be nice if the checker framework tolerated missing symbols referenced by `@EnsuresNonNullIf` annotations, provided the symbols were not accessed by the current compilation.

[1] https://github.com/bazelbuild/bazel/tree/80a34dc97799961201e6dce20fd58dd08022c032/third_party/ijar

smillst:
Fixes #1582.

This pull request includes the commits in #1591 .  That pull request should be merged first.


checker/jtreg/nullness/issue1582/foo/Foo.java
```diff
@@ -0,0 +1,28 @@
+package foo;
+
+import org.checkerframework.checker.nullness.qual.EnsuresNonNullIf;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public class Foo {
+
+    //  @Nullable Object theObject;
```
wmdietl: I'm not sure what the commented-out lines in this file are for. Can they be removed? Or can you add a comment to document?

framework/src/org/checkerframework/common/basetype/messages.properties
```diff
@@ -64,6 +64,8 @@ purity.more.deterministic=the method %s could be declared as @Deterministic
 
 flowexpr.parse.index.too.big=the method does not have a parameter %s
 flowexpr.parse.error=cannot parse the expression '%s'
+# Same as above, but only issued at call sites for a method with a postcondition.
+flowexpr.parse.postcondition.error=cannot parse the expression '%s'
```
wmdietl: If we were to use `flowexpr.parse.error.postcondition` or `postcondition.flowexpr.parse.error` we would automatically suppress this new warning when suppressing `flowexpr.parse.error`, which I think would be nice.

wmdietl: Can the error message provide more information, to make it easier to diagnose the problem? E.g. look at postconditions of method `C.foo` to fix this.

framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
```diff
@@ -1033,18 +1019,30 @@ protected void processConditionalPostconditions(
...
+                        thenStore.insertValue(r, anno);
+                    } else {
+                        elseStore.insertValue(r, anno);
+                    }
                 } else {
-                    elseStore.insertValue(r, anno);
+                    thenStore.insertValue(r, anno);
                 }
             } catch (FlowExpressionParseException e) {
+                Result r = e.getResult();
```
wmdietl: Would it make sense to more nicely encapsulate this within the `FlowExpressionParseException`? At the moment there are two kinds of parse exceptions and we could either have separate subclasses or one subclass with an enum kind. Or simply add a helper method in that class to encapsulate this logic?
Inspecting the result and comparing the message key seems a brittle.

framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
```diff
@@ -1025,8 +1026,21 @@ private void processPostconditionsAndConditionalPostconditions(
                     thenStore.insertValue(r, anno);
                 }
             } catch (FlowExpressionParseException e) {
+                Result result;
+                if (e.isFlowParseError()) {
+                    Object[] args = new Object[e.args.length + 1];
+                    args[0] = ElementUtils.getVerboseName(TreeUtils.elementFromUse(n.getTree()));
+                    for (int i = 0; i < e.args.length; i++) {
```
wmdietl: You might want to use https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#arraycopy-java.lang.Object-int-java.lang.Object-int-int- instead, to make explicit and shorter that you're just copying things around.
(There currently is always just one argument and the error message expects exactly 2 parameters, so maybe this generality is overkill...)

PR #1597, opened 2017-10-17
 Correct subtyping of @UnknownVal and @ArrayLenRange

smillst:


smillst:
> I was concerned that we ought to call this function in LUB and GLB, too, but both of those appear to rely directly on isSubtype.

I hadn't thought of that.  Thanks for checking!

PR #1599, opened 2017-10-18
 Don't put literal strings in inferred SameLen annotations.

smillst:
String s = "hello";
Before this change the type of s was `@SameLen({"s", ""hello""})` with this change, it is just `@SameLen("s")`.

Literal strings are handled by the Value Checker.  I couldn't come up with a situation where adding a  literal gives more precision.  @panacekcz @kelloggm Can either of you?

The reason for this change is that really big string values cause a stack overflow in FlowExpressionParseUtil#parse when calling pattern.match().


kelloggm:
This only situation where I can imagine that this might cause a loss of precision is something like the following:

```java
void test(@SameLen("#2") String s1, @SameLen("#1") String s2);
...
String s = "hello";
test(s, "hello");
```
This example seems both contrived and not the sort of code that anyone ought to write. In principle, I think this change is fine.

checker/src/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
```diff
@@ -179,10 +179,12 @@ public AnnotationMirror createCombinedSameLen(
         return res;
     }
 
-    public static boolean isReceiverToStringParsable(Receiver receiver) {
+    public static boolean shouldUseInAnnotation(Receiver receiver) {
```
kelloggm: This is a much better name.

smillst:
@kelloggm Thanks for the test case.  I thought that the SameLen Checker would use the Value Checker so that a test case like the one above would work, but it doesn't use the Value Checker at all.  For example:
```
import org.checkerframework.checker.index.qual.SameLen;

public class SameLenStrings {
    void test(@SameLen("#2") String s1, @SameLen("#1") String s2) {
    }

    void use(@StringVal("hello") String s) {
        test(s, "hello");
    }
}
```
Issues these warnings:

```
SameLenStrings.java:13: error: [argument.type.incompatible] incompatible types in argument.
        test(s, "hello");
             ^
  found   : @SameLen("s") String
  required: @SameLen(""hello"") String
SameLenStrings.java:13: error: [argument.type.incompatible] incompatible types in argument.
        test(s, "hello");
                ^
  found   : @SameLen(""hello"") String
  required: @SameLen("s") String
2 errors
```

I could fix this crash by limiting the length of literal strings in SameLen annotations.

kelloggm:
Alternatively, I can modify the SameLen Checker to actually use the Value Checker to handle cases like that test case. The modification would be quite straightforward.

smillst:
> Alternatively, I can modify the SameLen Checker to actually use the Value Checker to handle cases like that test case. The modification would be quite straightforward.

That's probably the correct long term solution.  But it's not clear how much this comes up in practice.  I can open an issue in kelloggm/checkerframework for this.

kelloggm:
Opening an issue in the kelloggm fork is the right thing to do. We can then merge this PR.

smillst:
I submitted an issue for SameLen Checker to use the Value Checker. https://github.com/kelloggm/checker-framework/issues/179.

smillst:
Closed in favor of fixing the problem in FlowExpressionParseUtil. (See issue #1609.)

Issue #1587, opened 2017-10-15
AsSuperVisitor: type is not an erased subtype of supertype 14


cushon:
```java
class One implements Two<One, Three, Four, Four> {}

interface Two<E extends Two<E, K, C, I>, K extends Five<K>, C extends Enum<C>, I extends Enum<I>> {}

abstract class Three implements Five<Three> {}

enum Four {}

interface Five<K extends Five<K>> extends Comparable<K> {}

interface Six {
  <E extends Two<E, ?, ?, I>, I extends Enum<I>> Seven<?, E> e(Class<E> entity);
}

interface Seven<T extends Seven<T, E>, E extends Two<E, ?, ?, ?>> extends Eight<T, E, E> {}

interface Eight<T extends Eight<T, R, E>, R, E extends Two<E, ?, ?, ?>> extends Nine<T, R> {}

interface Nine<T extends Nine<T, R>, R> {
  T d();

  Iterable<R> q();
}

abstract class T {
  public Iterable<One> f(Six e) {
    return g(e.e(One.class).d().q());
  }

  abstract Iterable<One> g(Iterable<One> r);
}
```

```
$ ~/jsr308/checker-framework-2.2.1/checker/bin/javac -processor Nullness T.java -AprintErrorStack
...
error: AsSuperVisitor: type is not an erased subtype of supertype.
  type: Two<E extends Two<E, ? extends Five<K extends Five<K extends Five<K>>>, ? extends Enum<C extends Enum<C extends Enum<C>>>, ? extends Enum<I extends Enum<I extends Enum<I>>>>, ? extends Five<K extends Five<K extends Five<K>>>, ? extends Enum<C extends Enum<C extends Enum<C>>>, ? extends Enum<I extends Enum<I extends Enum<I>>>>
  superType: One
  Compilation unit: /Users/cushon/src/error-prone/core/src/main/java/com/google/errorprone/bugpatterns/X.java
  Exception: java.lang.Throwable; Stack trace: org.checkerframework.framework.source.SourceChecker.errorAbort(SourceChecker.java:727)
  org.checkerframework.javacutil.ErrorReporter.errorAbort(ErrorReporter.java:41)
  org.checkerframework.framework.type.AsSuperVisitor.errorTypeNotErasedSubtypeOfSuperType(AsSuperVisitor.java:155)
  org.checkerframework.framework.type.AsSuperVisitor.visitDeclared_Declared(AsSuperVisitor.java:340)
[...]

PR #1610, opened 2017-10-24
Correct unbound wildcard crash.

smillst:
When initializing type arguments, use the mapping from type parameter to type arguments rather than creating a new AnnotatedTypeMirror when a type parameter is encountered.

Fixes Issue #1587.


smillst:
This is crashing on the JDK.  So, I'll fix that crash then repost.

PR #1611, opened 2017-10-24
Remove visitBinary from Lower Bound Checker Annotated Type Factory; move functionality to Lower Bound Transfer

kelloggm:
The Index Checker is known to suffer performance issues. This is the first step towards resolving that problem.

This PR removes code from `LowerBoundTreeAnnotator#visitBinary` and replaces the functionality with code in the transfer function. The previous version of `visitBinary` required calls to `getAnnotatedType`, which is known to be exponentially slow when processing nested binary expressions.

Running `ant index-tests` on my laptop takes 2 minutes 1 second with this PR, and 2 min 29 seconds without it (~18% faster).

Running `javac -processor lowerbound tests/index/BigBinaryExpr.java` with this PR takes 11.2 seconds, while master takes 16.0 seconds (~30% faster). Note that this test is currently commented out. I uncommented it to get these numbers.


kelloggm:
Your interpretation is correct. I just wanted you to look over it to make sure I'd done the conversion in a reasonable way!

PR #1613, opened 2017-10-24
Move annotations for JavaDoc classes to a stub file

### LINKED ISSUE #1602: Move javadoc annotations into a dedicated annotated tools.jar
When we construct an annotated JDK (e.g. by running `ant jdk.jar`), the resulting jar file does not contain class files for every class with an annotation in the JDK source. This appears to only affect JDKs that are built from source (index, nullness, lock). In particular, the generated jar is missing classes under the `com/sun/javadoc` directory, which the nullness JDK contains annotated source for, and I'd like to include in the index JDK. At least some of these classes do have annotations: consider `com/sun/javadoc/RootDoc.java`.

Steps to reproduce:
1. Build or download an annotated JDK.
2. unzip it: `unzip jdk8.jar`
3. Observe that no `com` directory is generated
4. Compare to the structure of the nullness JDK, in `checker/jdk/nullness/src`, which contains a `com` directory, with some annotated classes

kelloggm:
Fixes #1602. Because `ct.sym` in the JDK doesn't expose these APIs, they have to be treated like an external library.


checker/lib/README
```diff
@@ -3,20 +3,25 @@ This directory contains annotated versions of libraries.
 
 Instructions for users:
 
-During type-checking, put these versions on your classpath,
+During type-checking, put these jars on your classpath,
 to obtain more precise type-checking with fewer false positive warnings.
 
+If the file for the library you need is a stub file (its extension is
```
mernst: Shouldn't this be automatically handle by the checker, unless the user has created a new .astub file?
I would note that.

checker/lib/README
```diff
@@ -3,20 +3,25 @@ This directory contains annotated versions of libraries.
...
+.astub instead of .jar), you can supply it to the checker by using
+the -Astubs option, like this: -Astubs=javadoc.astub.
 
 Instructions for people creating a new annotated library:
 
 If you annotate a new library, please inform the Checker Framework developers
  so that they can include it in the Checker Framework distribution.
 
-When adding a .jar file to this directory, update the below list and also
+When adding a .jar or .astub file to this directory, update the below list and also
```
mernst: Is it also necessary to tweak the source code of the checker to know about the .astub file, or are all astub files automatically discovered?

checker/lib/README
```diff
@@ -3,20 +3,25 @@ This directory contains annotated versions of libraries.
...
 the list in ../../docs/manual/annotating-libraries.tex .
 In general, old files need not be removed; that way, we accommodate users
 regardless of which version of a library they wish to use.
 
+Files that
```
mernst: This seems like stray text.

mernst:
This can be merged when the tests pass.

mernst:
Don't merge until review comments are addressed. Then, merge when tests pass.

kelloggm:
My intention here was to make these astub files available, but not load them automatically - which is my impression of how annotated libraries are supposed to work. From your comments, it seems like you expect them to be loaded automatically. 

Since these annotations were in the JDK before, you might expect that making loading them optional would break backwards compatibility. This isn't the case, though, since when they were in the JDK they were ignored (that's what this PR is supposed to solve!).

mernst:
Putting aside for the moment the question of how the system works now, how should it work?
I think it would be easier to use if users didn't have to remember about .stub files, but stub files were used automatically.  I don't think this causes any problem even when the user isn't using the annotated libraries.

kelloggm:
I don't disagree with you, but I think it's unintuitive if some annotated libraries are loaded automatically and others aren't. I can open an issue that suggests loading the distributed annotated libraries automatically, but I don't think we should make com.sun.javadoc's classes do this but, for instance, not BCEL's classes.

mernst:
I was suggesting this only for stub files.  That's how the Checker Framework already behaves.

kelloggm:
> That's how the Checker Framework already behaves
I'm not sure what you mean by this. 

Do you mean "the Checker Framework already loads stub files automatically"? That's definitely not true - the only stub file it loads automatically is `jdk.astub`, whose location is hardcoded into the framework. All other stub files (i.e. those for annotated libraries) require a command line option to load.

Do you mean "the Checker Framework already loads annotated libraries automatically"? If that's true (I currently believe it is not), then I agree that this stub file should automatically be loaded, also. This is what I thought we were talking about, but you said

> I was suggesting this only for stub files

which suggests to me that this isn't what you're saying.

Can you clarify what you meant?

mernst:
The entire discussion is about stub files.

> Do you mean "the Checker Framework already loads stub files automatically"? That's definitely not true - the only stub file it loads automatically is jdk.astub, whose location is hardcoded into the framework.  All other stub files (i.e. those for annotated libraries) require a command line option to load.

A checker can have a `@StubFiles` annotation to indicate stub files that are read without the user needing to indicate them on the command line.  Here are examples:

```
framework/src/org/checkerframework/common/aliasing/AliasingChecker.java:11:1:@StubFiles({"android.astub"})
framework/src/org/checkerframework/common/reflection/MethodValChecker.java:14:1:@StubFiles({"reflection.astub"})
framework/src/org/checkerframework/common/reflection/ClassValChecker.java:14:1:@StubFiles({"reflection.astub"})
framework/src/org/checkerframework/common/value/ValueChecker.java:22:1:@StubFiles("statically-executable.astub")
checker/src/org/checkerframework/checker/guieffect/GuiEffectChecker.java:8:1:@StubFiles({"org-eclipse.astub", "org-osgi.astub", "org-swtchart.astub"})
checker/src/org/checkerframework/checker/regex/RegexChecker.java:14:1:@StubFiles("apache-xerces.astub")
checker/src/org/checkerframework/checker/fenum/FenumChecker.java:29:1:@StubFiles("jdnc.astub")
checker/src/org/checkerframework/checker/interning/InterningChecker.java:20:1:@StubFiles({"com-sun.astub", "org-jcp.astub", "org-xml.astub", "sun.astub"})
```

I was suggesting that you do the same with the new stub files you have created.

kelloggm:
Thank you for clarifying. I wasn't aware of this feature, and it seems I missed the mention of it in the manual (probably because it's only mentioned under the instructions for creating an annotated JDK, where I wasn't looking, since I was explicitly trying to avoid doing that). I'll take advantage of this feature and automatically load the stub files.

framework/src/org/checkerframework/common/value/javadoc.astub
```diff
@@ -0,0 +1,7 @@
+import org.checkerframework.common.value.qual.MinLen;
+
+package com.sun.javadoc;
+
+interface RootDoc {
+    String @MinLen(1) [][] options();
```
wmdietl: Should this be `String [] @MinLen(1) []`?
https://docs.oracle.com/javase/8/docs/jdk/api/javadoc/doclet/com/sun/javadoc/RootDoc.html#options--
doesn't indicate that there always are options. The classdoc says `Also represents the command line information -- the packages, classes and options specified by the user.` which seems to suggest that if I only call `javadoc mypackage` then options is empty?
But for each option, the second dimension is always at least 1 element long.

checker/src/org/checkerframework/checker/nullness/javadoc.astub
```diff
@@ -0,0 +1,97 @@
...
+    FieldDoc[] serializableFields();
+    boolean definesSerializableFields();
+    ClassDoc superclass();
+    Type superclassType();
+    boolean subclassOf(ClassDoc cd);
+    ClassDoc[] interfaces();
+    Type[] interfaceTypes();
+    TypeVariable[] typeParameters();
+    ParamTag[] typeParamTags();
+    @NonNull FieldDoc @NonNull [] fields();
```
wmdietl: What defaults does this file assume and where is that default set? astub files use the normal source defaults, which would be non-null here. Is that changed somewhere? Are all the other methods really returning nullable?

checker/src/org/checkerframework/checker/nullness/NullnessChecker.java
```diff
@@ -1,6 +1,9 @@
 package org.checkerframework.checker.nullness;
 
+import org.checkerframework.framework.qual.StubFiles;
+
 /** A concrete instantiation of {@link AbstractNullnessChecker} using freedom-before-commitment. */
+@StubFiles("javadoc.astub")
```
wmdietl: Parsing stub files is slow. Adding this new stub file here adds this overhead to everyone that wants to use the Nullness Checker, regardless of whether they use javadoc or not.
Is this really a good idea? Why not provide a script that produces an annotated `tools.jar`, similar to the process we use to produce an annotated `jdk8.jar`?
Then people that use javadoc can add that library, like they need to do for the other nullness-annotated libraries in `checker/lib`.

mernst: The javadoc.astub file is only 100 lines long, so I hope the overhead isn't too high.
In the future moving the annotations into a proper library is a good idea, for these and all other stub files.  I don't think we should delay this pull request for that change, though.

framework/src/org/checkerframework/common/value/javadoc.astub
mernst: Thanks for catching the misplaced annotation.  I fixed this.

checker/src/org/checkerframework/checker/nullness/NullnessChecker.java
wmdietl: I would strongly prefer making it optional to include that file.
Nobody complained about the missing annotations. The only useful annotations are `@Pure`. I don't see anything nullness related.
I think this can wait until the annotations are cleaned up. If this important before then, we can include the file, but not include it for every run.

mernst: > Nobody complained about the missing annotations.
Maybe not to you. :-)

wmdietl: Ok, if you have a concrete user, they can use `-Astubs` to use them.

checker/src/org/checkerframework/checker/nullness/javadoc.astub
mernst: It seems that writing `@NonNull` does not have any effect here, because it is the default.
So, it seems that the old annotations that you copied over were written with a different assumption, or are out of date.

Please remove all the annotations that are the same as the default, which is `@NonNull` annotations on a top-level type.

checker/src/org/checkerframework/checker/nullness/NullnessChecker.java
mernst: Let's start this comment thread over.  The speed issue that sparked this thread is not important, but there is an important issue that motivates removing this `@StubFiles("javadoc.astub")` annotation.

If a user uses `-AstubDebug` but does not have `tools.jar` on the classpath, then there will be a flood of warnings about classes that are not found.  This is annoying and will hinder people in debugging their stub files.

Since a user has to specify `tools.jar` explicitly when compiling, it seems reasonable to require them to add an explicit stub file on the command line.

I had not thought about the problem of warnings about classes not found in stub files.  Sorry about suggesting use of `@StubFiles` here.

kelloggm:
@mernst @wmdietl After reading your comments, I've switched back to the original strategy of providing the stub files in the lib directory, along with instructions on how to use them. This avoids the problems that you all noticed with the `@StubFiles` approach. I've also removed redundant `@NonNull` annotations from the stub file.

checker/lib/javadoc.astub
```diff
@@ -0,0 +1,107 @@
+import org.checkerframework.common.value.MinLen;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.checkerframework.dataflow.qual.Pure;
+
+package com.sun.javadoc;
+
+// Index Checker annotations
+
+interface RootDoc {
```
mernst: RootDoc appears twice in this stub file.  Does that work?
Is it necessary to separate the different types of annotations rather than making each class appear just once?

kelloggm: The documentation in [sec. 28.4.1 of the manual](https://checkerframework.org/manual/#stub) indicates that when the same element occurs multiple times in a stub file (or in multiple different stub files), then the annotations are merged. So this does work.

A different question is whether we should separate the annotations. I chose to separate them because they were previously in annotated JDKs, where the source files for each checker are distinct. I separated the annotations in the stub file in a similar way, but one could argue that having all the annotations on a class/method in the same place is easier to read or use. I'm not partial to either position.

mernst: OK, it's good that it works.  Thanks for checking that.

I don't think we should organize the file based on a previous organization, but what makes most sense in this file.  I favor having each class appear just once in the file.

PR #1622, opened 2017-10-31
Fix crash in Guava

smillst:
The fix for #1587 caused another crash when type-checking Guava.  This pull request fixes that crash.

### DIFF
--- a/framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java
+++ b/framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java
@@ -943,10 +943,9 @@ public abstract class AnnotatedTypeMirror {
                             (AnnotatedTypeVariable) declaration.getTypeArguments().get(i);
                     TypeVariableSubstitutor varSubstitutor = atypeFactory.getTypeVarSubstitutor();
                     // The upper bound of a type parameter may refer to other type parameters.
-                    // Substitute those references with the type argument, but make a copy so that
-                    // the merge happens without a concurrent modification exception.
+                    // Substitute those references with the type argument.
                     AnnotatedTypeMirror typeParamUpperBound =
-                            varSubstitutor.substitute(map, typeParam.getUpperBound()).deepCopy();
+                            varSubstitutor.substitute(map, typeParam.getUpperBound());
 
                     AnnotatedWildcardType wct = (AnnotatedWildcardType) typeArgs.get(i);
                     AnnotatedTypeMerger.merge(typeParamUpperBound, wct.getExtendsBound());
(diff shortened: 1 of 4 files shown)


framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java
```diff
@@ -943,9 +943,10 @@ public void setTypeArguments(List<? extends AnnotatedTypeMirror> ts) {
...
                     // The upper bound of a type parameter may refer to other type parameters.
-                    // Substitute those references with the type argument.
+                    // Substitute those references with the type argument, but make a copy so that
+                    // the merge happens without a concurrent modification exception.
                     AnnotatedTypeMirror typeParamUpperBound =
-                            varSubstitutor.substitute(map, typeParam.getUpperBound());
+                            varSubstitutor.substitute(map, typeParam.getUpperBound()).deepCopy();
 
                     AnnotatedWildcardType wct = (AnnotatedWildcardType) typeArgs.get(i);
                     AnnotatedTypeMerger.merge(typeParamUpperBound, wct.getExtendsBound());
```
smillst: Concurrent modification exception throw in this call.

framework/src/org/checkerframework/framework/type/AnnotatedTypeMirror.java
```diff
@@ -943,9 +943,10 @@ public void setTypeArguments(List<? extends AnnotatedTypeMirror> ts) {
                             (AnnotatedTypeVariable) declaration.getTypeArguments().get(i);
                     TypeVariableSubstitutor varSubstitutor = atypeFactory.getTypeVarSubstitutor();
                     // The upper bound of a type parameter may refer to other type parameters.
-                    // Substitute those references with the type argument.
+                    // Substitute those references with the type argument, but make a copy so that
+                    // the merge happens without a concurrent modification exception.
                     AnnotatedTypeMirror typeParamUpperBound =
-                            varSubstitutor.substitute(map, typeParam.getUpperBound());
+                            varSubstitutor.substitute(map, typeParam.getUpperBound()).deepCopy();
```
wmdietl: It is a bit surprising that the `deepCopy` is necessary.
The return type for `substitute` is documented as `a copy of typeMirror with its type variables substituted`.
Is that copy not deep enough?
Is there only one particular case that doesn't get copied correctly? Then, wouldn't it be better to fix this in `TypeVariableSubstitutor`?

smillst: TypeVariableSubstitutor was only doing a shallow copy of the type arguments substituted.  (See TypeVariableSubstitutor#substituteTypeVariable.)  I changed it to a deep copy, but I'm not sure if there is other code that relies on the shallow copy.

PR #1624, opened 2017-10-31
Fix GuiEffectChecker method override checks

### LINKED ISSUE #1495: GuiEffectChecker has broken override checks for @PolyUI
As noticed in @lazaroclapp's PR #1481, the GuiEffectChecker has a bug that is roughly equivalent to unchecked covariant subtyping on generic parameters.

For a concrete example:
```java
import org.checkerframework.checker.guieffect.qual.*;

public class Q {

    @PolyUIType
    public static interface I {
        @PolyUIEffect
        public void m();
    }

    public static void reqSafe(@AlwaysSafe I i) {}
    @UIEffect
    public static void reqUI(@UI I i) {}

    @PolyUIType
    public static interface Doer {
        public void doStuff(@PolyUI Doer this, @PolyUI I i);
    }

    public void q(@AlwaysSafe Doer doer, @UI I i) {
        doer.doStuff(i);
    }

        @AlwaysSafe Doer d = new @AlwaysSafe Doer() {
                                @SafeEffect
                                public void doStuff(@AlwaysSafe I i) {
                                    reqSafe(i);
                                }
                              };
        @UI I ui = new @UI I() {
                        @UIEffect
                        public void m() {
                            reqUI(null);
                        }
                    };
        Q q = new Q();
        q.q(d, ui);
    }
}
```
The ```main``` method above feeds an ```@UI I``` into ```reqSafe```.  Some part of the code above should be rejected.  The critical point of failure is that at the call to method ```doStuff```, the ```@AlwaysSafe Doer``` is upcast to ```@UI Doer```, and ```@UI``` is chosen as the instantiation of ```@PolyUI```, in order to permit the passing of the ```@UI I```.  Then inside the concrete implementation of ```doStuff```, the method is permitted to continue believing the receiver and argument are both ```@AlwaysSafe```.

Essentially, the choice of implicit effect variable instantiation is made once outside the class, but we then permit it to be changed when we use ```@PolyUI``` receiver annotations.

I initially wondered if the bug was specific to the GuiEffectChecker or inherent to the handling of receivers annotated with ```@PolymorphicQualifer``` qualifiers.  However, the tainting checker does not have the same error.  Consider:

```java

public class Tainting {

    public static void reqUntainted(@Untainted Object o) {
        System.out.println("Should be untainted: "+o.toString());
    }

     public static interface ITaint {
         public void m(@PolyTainted ITaint this, @PolyTainted ITaint other);
     }
     @Untainted
     public static class UTaint implements ITaint {
         public void m(@Untainted UTaint this, @Untainted ITaint other) {
             reqUntainted(other);
         }
     }
     @Tainted
     public static class TTaint implements ITaint {
         public void m(@PolyTainted TTaint this, @PolyTainted ITaint other) {
         }
         @Override
         public String toString() { return "<Tainted object>"; }
     }

        @Untainted ITaint un = new @Untainted UTaint();
        @Tainted ITaint t = new @Tainted TTaint();
        un.m(t);
     }
}
```
When compiling this with the TaintingChecker:
```
$ javac -processor org.checkerframework.checker.tainting.TaintingChecker Tainting.java 
Tainting.java:15: error: [type.invalid] [@Tainted] may not be applied to the type "@Tainted UTaint"
     public static class UTaint implements ITaint {
                   ^
Tainting.java:16: error: [override.param.invalid] void m(@Untainted UTaint this, @Untainted ITaint p0) in Tainting.UTaint cannot override void m(@PolyTainted ITaint this, @PolyTainted ITaint p0) in Tainting.ITaint; attempting to use an incompatible parameter type
         public void m(@Untainted UTaint this, @Untainted ITaint other) {
                                                                 ^
  found   : @Untainted ITaint
  required: @PolyTainted ITaint
Tainting.java:16: error: [override.receiver.invalid] void m(@Untainted UTaint this, @Untainted ITaint p0) in Tainting.UTaint cannot override void m(@PolyTainted ITaint this, @PolyTainted ITaint p0) in Tainting.ITaint; attempting to use an incompatible receiver type
         public void m(@Untainted UTaint this, @Untainted ITaint other) {
                     ^
  found   : @Untainted UTaint
  required: @PolyTainted ITaint
3 errors
```

So some part of the GuiEffectChecker has broken override checking.

There are two main options:

1. Restrict declarations: If we disallow using ```@PolyUI``` parameters with ```@PolyUIEffect``` methods (except for the receiver), this will avoid conflating a class-level "effect variable" with a method-level one, as currently happens.  It might, however, break some existing code.  
2. Restrict subtyping: If a class uses the class of methods the first option would reject, instead allow it but prohibit coercing ```@AlwaysSafe``` instances of that class to ```@UI```.  This would break a different subset of possible code, and possibly also make the checker's behavior harder to explain to non-type-theorists.

Based on the SWT and AWT code I've seen, the first option wouldn't be a problem, making it preferable.  But we should ask @msridhar and @lazaroclapp as representatives of the most substantial use we're aware of at the moment, on very different UI infrastructure.

@lazaroclapp, is the problematic case (the case you noticed is permitted when it should be rejected) suggestive of actual code you've seen, or were you just being thorough working through the possibilities?  Since either solution requires detecting the same class of problematic methods, I can prototype the first one for you to try out if you're unsure.

csgordon:
This should fix #1495 

I'll file a separate issue on myself to update the documentation and release notes, as this is a slightly breaking change.


mernst:
I think it would be better to commit (also to review) all the changes together.
In any event, there isn't time to include this in the November 1 release, but users will be able to access it from the version control repository once it is merged.
Thanks!

mernst:
Minor: could you please change the issue title to be more descriptive?  It will be the default commit message when this pull request is merged.

csgordon:
While this worked in a previous CI build (https://travis-ci.org/typetools/checker-framework/builds/276672318, which failed due to hygiene, but passed functional tests), my merge from master broke the new tests for this issue.  So I'm off to see if some part of the override checking changed...

csgordon:
Ah, it's because the format for expected diagnostics changed, so I had to switch
```
//:: error: (override.param.invalid)
```
to
```
// :: error: (override.param.invalid)
```

@mernst @wmdietl @smillst Do we want to issue some kind of complaint somewhere if someone misses that space?  This seems like a really easy thing to get wrong, and it only even occurred to me because I remembered seeing a related commit message fly by in the last couple weeks.

mernst:
@csgordon , OK, I have done that.

I had initially permitted either format, then I decided that would just encourage too much variety -- we would like the tests to be uniform.

checker/src/org/checkerframework/checker/guieffect/GuiEffectVisitor.java
```diff
@@ -65,14 +66,102 @@ protected void checkMethodInvocability(
         // TODO: Undo this hack!
     }
 
+    protected class GuiEffectOverrideChecker extends OverrideChecker {
+        /**
+         * Extend the receiver part of the method override check. We extend the standard check, to
+         * additionally permit narrowing the receiver's permission to <code>@AlwaysSafe</code> in a
```
mernst: Minor:  use `{@code ...}` instead of `<code>...</code>` in Javadoc comments.

checker/src/org/checkerframework/checker/guieffect/GuiEffectVisitor.java
```diff
@@ -65,14 +66,102 @@ protected void checkMethodInvocability(
...
+         * safe instantiation of a <code>@PolyUIType</code> Returns true if the override is
+         * permitted.
+         */
+        @Override
+        protected boolean checkReceiverOverride() {
+            // We cannot reuse the inherited method because it directly issues the failure, but we want a more permissive check.  So this is copied down and modified from BaseTypeVisitor.OverrideChecker.checkReceiverOverride.
```
mernst: Please wrap comments to either 80 or 100 columns (your choice).

(GitHub brokenly refuses to show comments longer than that!  But it's mostly for consistency with the rest of the codebase, and readability.)

checker/src/org/checkerframework/checker/guieffect/GuiEffectVisitor.java
```diff
@@ -203,6 +331,8 @@ public Void visitMethod(MethodTree node, Void p) {
             System.err.println("\nVisiting method " + methElt);
         }
 
+        //checkSaveCovariantReceiver(node, methElt);
```
mernst: Is this necessary?

csgordon:
Unless I'm missing something, we're still waiting for @wmdietl to take a
look at the issue with the inability to persist updated @UI annotations on
lambdas, per email.

On Nov 4, 2017 2:32 PM, "Michael Ernst" <notifications@github.com> wrote:

Assigned #1624 <https://github.com/typetools/checker-framework/pull/1624>
to @csgordon <https://github.com/csgordon>.

—
You are receiving this because you were assigned.

Reply to this email directly, view it on GitHub
<https://github.com/typetools/checker-framework/pull/1624#event-1326156306>,
or mute the thread

mernst:
OK, thanks for the clarification about the status.  I'll re-assign to Werner.

csgordon:
Whoops, I replied to the wrong thread. This one is still mine

mernst:
OK, I changed the assignment back.

csgordon:
@mernst I've made the changes you requested, and cleaned out some dead code while I was at it.

One particular point I'm wondering about in the review is the handling of generic receivers in ```GuiEffectVisitor.GuiEffectOverrideChecker.checkReceiverOverride```, which is based on ```BaseTypeVisitor.OverrideChecker.checkReceiverOverride```, but also contains a note that its handling of generic receivers may be incomplete: https://github.com/typetools/checker-framework/blob/b937933b46ee28d3644fbbc5e3bcf94c88169edd/framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java#L2906

csgordon:
I'll merge as is. I'll open a separate issue about receiver generics; as I noted I think this issue applies to all checkers, since the same comment is in the BaseTypeVisitor version as well.

PR #1629, opened 2017-11-03
Always recalculate direct super types for AnnotatedDeclaredTypes

### LINKED ISSUE #1605: possible issue with missing default annotations on supertypes
Unfortunately, I don't have a small test case to reproduce this issue, and I'm not even sure I can explain it properly.  We are seeing weird crashes like those seen in #1415 (MultiGraphQualifierHierarchy found an unqualified type), and it seems to be due to missing default annotations on supertypes.  I see some possibly-suspicious code in [this method](https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java?utf8=%E2%9C%93#L935) of `AnnotatedTypeFactory`.  That code gets an `AnnotatedTypeMirror` for an `Element` and then calls `addComputedTypeAnnotations` which adds default annotations for some type systems (like for the GUI effect checker).  But if the type has supertypes, it seems that it is possible the computed type annotations are not added for the supertypes here?  With the GUI effect checker we are seeing, e.g., `class B extends A` where the type for `B` is `@AlwaysSafe B` but the supertype is just `A`, rather than `@AlwaysSafe A`.

Sorry for the somewhat vague report and lack of a proper test input.  I'm hoping the above code pointers will be enough for someone to see if something is problematic here.  I can try to produce a reduced input for our problem, but it may take a while.

smillst:
If the annotations on the AnnotatedDeclaredType or enclosing type changes or if the type arguments change, then the super types must be recomputed.  Instead of keeping track of these changes, just recompute.  

Fixes #1605.

I'm not sure if this is the best fix. It's not clear to me why the super types were stored for the AnnotatedDeclaredTypes in the first place.  (The only other types they are stored for is AnnotatedIntersectionTypes which makes sense because the super types can be directly annotated.) My guess is for performance.  I'll run this branch on ~Guava and~ the JDK to see if there is any difference.  There was no impact with Guava.

### DIFF
--- a/framework/src/org/checkerframework/framework/type/AnnotatedTypeCopier.java
+++ b/framework/src/org/checkerframework/framework/type/AnnotatedTypeCopier.java
@@ -130,14 +130,6 @@ public class AnnotatedTypeCopier
             copy.setTypeArguments(copyTypeArgs);
         }
 
-        if (original.supertypes != null) {
-            final List<AnnotatedDeclaredType> copySupertypes = new ArrayList<>();
-            for (final AnnotatedDeclaredType supertype : original.supertypes) {
-                copySupertypes.add((AnnotatedDeclaredType) visit(supertype, originalToCopy));
-            }
-            copy.supertypes = Collections.unmodifiableList(copySupertypes);
-        }
-
         return copy;
     }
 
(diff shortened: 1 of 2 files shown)


smillst:
Both Guava and JDK type-check in about the same time as before.  The need to recompute the super types seems to be offset by no longer having to copy the super types.

PR #1631, opened 2017-11-03
Cache annotation names for non Checker Framework annotations

smillst:
@kelloggm profiled the Index Checker and found that `AnnotationUtils#annotationName` was consuming 40% of the time using [jvmtop](https://github.com/patric-r/jvmtop).  Using a cache from the types of the AnnotationMirrors to the name fixes this.  `AnnotationUtils#annotationName` now takes around 1.5% of the time.  Guava which normally takes around 11 minutes to type-check with the Index Checker on my machine now takes 7.5 minutes.


mernst:
No, this does not re-introduce #1482.
There were two problems that contributed to #1482:
 * The key of the cache had a pointer to the compiler itself.
 * Multiple runs of javac used the same JVM, so multitple compilers were being pinned (kept in memory, prevented from being garbage collected) by the cache.

This pull request solves the first problem, and that is enough to prevent the performance pathology in #1482.  Solving the second problem would be of value, but is not necessary in order to merge this pull request.

wmdietl:
What makes you so sure that using a `DeclaredType` as key is different from using an `AnnotationMirror`?
Looking at `ClassType` and `TypeCompound` doesn't make it obvious to me why one would capture the compiler and the other wouldn't.
What argument do you have that `ClassType` doesn't capture a reference to the compiler?

smillst:
I tried removing the cache and not interning the result and changing areSame to use `equals`.  This has about the same performance improvement as the cache, so I'll do that instead.   Also, remove interning annotation names elsewhere.  (For example, `AnnotatedTypeFactory#getDeclAnnotation` is where ~15% of time is spent and it interns the annotation name, too.)

javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -104,7 +99,7 @@ public static boolean areSame(
             if (a1 == a2) {
                 return true;
             }
-            if (annotationName(a1) != annotationName(a2)) {
+            if (!annotationName(a1).equals(annotationName(a2))) {
```
smillst: @wmdietl Could we compare by `a1.getAnnotationType() == a2.getAnnotatedType()` instead of by name?

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -3303,8 +3303,7 @@ protected boolean isAccessAllowed(
             return true;
         }
 
-        String when =
-                AnnotationUtils.getElementValueClassName(unused, "when", false).toString().intern();
+        String when = AnnotationUtils.getElementValueClassName(unused, "when", false).toString();
```
wmdietl: How about replacing these uses of `String` by `Name`? I see a few places where we get a `Name` from the compiler, call `toString().intern()` and then pass around the `String`.
We're already changing APIs in this release, so we might as well clean this up.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -265,8 +265,10 @@
      * A map from the class name (canonical name) of an annotation to the set of class names for
      * annotations with the same meaning, as well as the annotation mirror that should be used.
      */
-    private final Map<String, Pair<AnnotationMirror, Set</*@Interned*/ String>>> declAliases =
-            new HashMap<>();
+    private final Map<
```
wmdietl: I would expect some corresponding change to the documentation.

javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -68,13 +63,13 @@ public static void clear() {
...
     /** @return the fully-qualified name of an annotation as a String */
-    public static final /*@Interned*/ String annotationName(AnnotationMirror annotation) {
+    public static final String annotationName(AnnotationMirror annotation) {
         if (annotation instanceof AnnotationBuilder.CheckerFrameworkAnnotationMirror) {
             return ((AnnotationBuilder.CheckerFrameworkAnnotationMirror) annotation).annotationName;
         }
         final DeclaredType annoType = annotation.getAnnotationType();
         final TypeElement elm = (TypeElement) annoType.asElement();
-        /*@Interned*/ String name = elm.getQualifiedName().toString().intern();
+        String name = elm.getQualifiedName().toString();
```
wmdietl: This is an example for my first comment: let's change the return type of this method to `Name` and directly return `elm.getQualifiedName()`.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
smillst: I tried this, but didn't see any speed up.  (`containsSameByName` isn't one of the top 20 methods.)  I also tried comparing by `Name` in `areSameIgnoringValues`, but also didn't see any speed up.

javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
smillst: I tried this, but it didn't help with performance.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
wmdietl: Even without an immediate speed-up, wouldn't it be the nicer API? The compiler already uses `Name` everywhere, so why convert to `String`.
If you've already done the work, why not commit it?

smillst: I changed some of the code, but I didn't update the documentation or make the change in every place possible. To really make this change, every `SameByName` method needs to be duplicated, because there are still places where only we have a String but not a Name.  Since it doesn't help with performance, this change is at the bottom of my list.  (And shouldn't hold up this pull request.)

wmdietl: I don't think you can make a fair assessment of performance if you didn't convert all locations.
I wouldn't duplicate all these methods. The places that use `String`s should use `com.sun.tools.javac.util.Names.fromString(String)` to generate a `Name`. If that only happens in few locations, we would safe those String instantiations and comparisons.
But I'm fine with this PR if it gives us a large performance benefit.

PR #1632, opened 2017-11-04
Clean up implementation and documentation of areSame

mernst:


javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -100,38 +100,39 @@ public static String annotationSimpleName(AnnotationMirror annotation) {
...
-                return false;
-            }
-
-            Map<? extends ExecutableElement, ? extends AnnotationValue> elval1 =
-                    getElementValuesWithDefaults(a1);
-            Map<? extends ExecutableElement, ? extends AnnotationValue> elval2 =
-                    getElementValuesWithDefaults(a2);
+        if (a1 == null || a2 == null) {
+            return a1 == a2;
+        }
```
wmdietl: One important case is missing: both `a1` and `a2` being non-null and `==`.
This changed logic now requires converting the same object to a string twice, which will slow things down.

mernst:
Thanks for the catch.  Can you take another look now that I have fixed it?

javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -100,38 +100,45 @@ public static String annotationSimpleName(AnnotationMirror annotation) {
...
-                    getElementValuesWithDefaults(a2);
+        if (a1 == a2) {
+            return true;
+        }
+        if (a1 == null || a2 == null) {
+            return false;
+        }
 
-            return elval1.toString().equals(elval2.toString());
+        if (!areSameIgnoringValues(a1, a2)) {
```
wmdietl: I would probably keep the previous `annotationName(a1) != annotationName(a2)`. The method call doesn't add much readability and adds redundant checks. But I'm fine with changing it, if you feel this makes it much easier to read.

PR #1643, opened 2017-11-08
Remove false negatives in Value Checker

kelloggm:
This fixes kelloggm#183.

Annotations with non-sensical ranges were being replaced with bottom, even when they were not user written. So, for instance, an `@ArrayLenRange(from=0, to=-1)` annotation would be replaced with `@BottomVal` by the annotated type factory, and an error would be issued by the visitor. This PR changes the annotated type factory's behavior to instead replace these annotations with top.

The previous behavior caused false negatives related to overflow - adding 1 to a `@NonNegative` integer produces the range `@IntRange(from=1, to=Integer.MIN_VALUE)`. Changing this to bottom allows an array to be created with this value and then **any** MinLen annotation to be written on the array, which is clearly incorrect.

This change may introduce some false positives, since if a user writes a non-sensical annotation they might get warnings that both directly warn about the nonsensical annotation and errors related to incompatible types derived from their annotation being replaced with top. I believe this is an acceptable tradeoff, since those warnings will disappear once the offending annotation is removed.


smillst:
I'll review once the tests pass.

panacekcz:
> adding 1 to a `@NonNegative` integer produces the range `@IntRange(from=1, to=Integer.MIN_VALUE)`

Can't we rather prevent this from happening?

kelloggm:
@panacekcz The Value Checker intentionally does not ignore overflow. This is the expected behavior; the Value Checker also has an "ignore overflow" setting for which this is explicitly prevented. But, that's inherently unsafe: errors related to overflow are ignored.

panacekcz:
I think that replacing `@ArrayLenRange(from=0, to=-1)` by `@BottomVal` is correct.
I think the real problem is this condition: https://github.com/typetools/checker-framework/pull/1643/files#diff-660753b764271c91741187a29866f6cbL439
```java
if (from > to || from < 0) {
//               ^^^^^^^^
  atm.replaceAnnotation(BOTTOMVAL);
}
```
That means that also `@ArrayLenRange(from=-1, to=1)` is replaced by `@BottomVal`. It should be replaced by `@ArrayLenRange(from=0, to=1)`.

After an overflow, the range is `@IntRange(from=Integer.MIN_VALUE, to=Integer.MAX_VALUE)`. 
In `ValueTreeAnnotator.handleDimensions`, the array length is set to `@ArrayLenRange(from=Integer.MIN_VALUE, to=Integer.MAX_VALUE)`. Already at this point, we should avoid creating an `@ArrayLenRange` with a negative `from` and make it `@ArrayLenRange(from=0, to=Integer.MAX_VALUE)`.

kelloggm:
Mike and I had a discussion about this this morning. We came to the following conclusions:
- replacing `@ArrayLenRange(from=0, to=-1)` by `@BottomVal` is correct
- replacing `@ArrayLenRange(from=0, to=-1)` by `@BottomVal` **silently** is incorrect; an error should be issued when we make this replacement, because it indicates a mistake in the checker's semantics, because:
- we should avoid creating an `@ArrayLenRange` with a negative `from`

I'll address these issues in this PR. In particular, my intention is to:
- prevent `@ArrayLenRange` annotations from ever having a negative value. Arrays can never actually have a negative length, so a negative value for either field of an `@ArrayLenRange` annotation indicates an imprecision in the checker, not an error in the program.
- add an error to the code that converts both `@IntRange` annotations and `@ArrayLenRange` annotations to bottom. Relatedly, I might remove the error that the visitor currently issues. The reasoning here is that we must never convert something to bottom without issuing a warning, since that might create false negatives.

kelloggm:
@smillst, I believe this PR is ready for your review

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -972,6 +991,21 @@ public ValueTreeAnnotator(ValueAnnotatedTypeFactory factory) {
...
+        @Override
+        protected Void defaultAction(Tree node, AnnotatedTypeMirror p) {
+            AnnotationMirror bottom = p.getAnnotation(BottomVal.class);
+            if (bottom != null) {
+                @SuppressWarnings(
+                        "assignment.type.incompatible") // Only compiler message keys are placed into these annotations
+                /*@CompilerMessageKey*/ String errorKey =
+                        AnnotationUtils.getElementValue(bottom, "value", String.class, true);
+                if (!"".equals(errorKey)) {
+                    checker.report(Result.failure(errorKey), node);
```
smillst: Warnings should only be issued by the visitor.  (This is because methods in the factory may be called more than once and thus the error might be issued more than once.)  This error should probably in an override of BaseTypeVisitor#validateType.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1817,6 +1851,12 @@ public AnnotationMirror createArrayLenRangeAnnotation(Range range) {
         }
     }
 
+    public AnnotationMirror createErrorBottom(String errorKey) {
+        AnnotationBuilder builder = new AnnotationBuilder(processingEnv, BottomVal.class);
```
smillst: Currently, "from.greater.than.to" is the only error key. So you could create that annotation once and then use it directly at the call sites to this method.

kelloggm:
@smillst I've addressed your review comments. The result is much cleaner; thanks! It should be ready for another review, when you get the chance.

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -306,6 +306,68 @@ public FieldInvariants getFieldInvariants(TypeElement element) {
...
+     *     the domain of the underlying type (i.e. Integer.MIN_VALUE if the underlying type is int)
+     */
+    public long getFromValueFromIntRange(AnnotatedTypeMirror atm) {
+        AnnotationMirror anno = atm.getAnnotation(IntRange.class);
+        long from;
+        if (AnnotationUtils.hasElementValue(anno, "from")) {
+            from = AnnotationUtils.getElementValue(anno, "from", Long.class, false);
+        } else {
+            switch (atm.getUnderlyingType().getKind()) {
+                case INT:
```
smillst: CHAR, too?

framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -306,6 +306,68 @@ public FieldInvariants getFieldInvariants(TypeElement element) {
...
+     *     domain of the underlying type (i.e. Integer.MAX_VALUE if the underlying type is int)
+     */
+    public long getToValueFromIntRange(AnnotatedTypeMirror atm) {
+        AnnotationMirror anno = atm.getAnnotation(IntRange.class);
+        long to;
+        if (AnnotationUtils.hasElementValue(anno, "to")) {
+            to = AnnotationUtils.getElementValue(anno, "to", Long.class, false);
+        } else {
+            switch (atm.getUnderlyingType().getKind()) {
+                case INT:
```
smillst: CHAR, too?

framework/src/org/checkerframework/common/value/ValueVisitor.java
```diff
@@ -260,4 +260,33 @@ public Void visitTypeCast(TypeCastTree node, Void p) {
         }
         return super.visitTypeCast(node, p);
     }
+
+    /**
+     * Overridden to issue errors at the appropriate place if an IntRange or ArrayLenRange
+     * annotation has from > to. from > to either indicates a user error when writing an annotation
```
smillst: The Javadoc isn't building because of `>`.  You can use `&lt;` or `{@code from > to}`.

PR #1644, opened 2017-11-08
Don't issue methodref.inference.unimplemented warnings

smillst:
If conservativeUninferredTypeArguments is used; otherwise, never issue the warning.


framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -2589,11 +2586,10 @@ private boolean checkMethodReferenceInference(
                 }
             }
             if (requiresInference) {
-                if (!typeArgumentInferenceCheck) {
+                if (checker.hasOption("conservativeUninferredTypeArguments")) {
```
wmdietl: Can you update the test in `./framework/tests/all-systems/java8/memberref/MemberReferences.java` to make explicit why or why not to expect the error message.

smillst: Done.

PR #1646, opened 2017-11-09
Improve what tests get executed on travis.

wmdietl:
This is expected to currently fail in two of the travis jobs. That's a good thing, as it exposes that we didn't have good coverage.


wmdietl:
This failure https://travis-ci.org/typetools/checker-framework/jobs/299427445
will be fixed by removing `@Interned` from `EqualityTests.java` and by #1644 .

Not sure why the `index-tests` work here but not for me locally...

.travis-build.sh
```diff
@@ -62,6 +62,7 @@ if [[ "${GROUP}" == "nonjunit" || "${GROUP}" == "all" ]]; then
 fi
 
 if [[ "${GROUP}" == "all-tests" || "${GROUP}" == "all" ]]; then
+  (cd framework && ant all-tests)
```
smillst: This is running tests that are already run in `(cd checker && ant all-tests-nobuildjdk)`.

smillst: Aren't the checker annotations build in  by `./.travis-build-without-test.sh ${BUILDJDK}` on line 48 of this script?

wmdietl: PR #1660 passes, whereas this PR fails with the expected error https://travis-ci.org/typetools/checker-framework/jobs/301634837#L1604
So maybe the checker annotations are already built earlier, but they are not on the classpath when only the framework tests are executed.
So this clearly improves the quality of the Travis test coverage and should be merged (after undoing the failing test).

smillst: So, the bug is that the qualifiers are on the classpath of the framework tests when they are invoked via checker/build.xml `all-tests`, but not when they are invoke via framework/build.xml  `all-tests`.  This pull request works around that bug by running the framework tests twice.  That should be documented, so we don't remove this line in the future.

Also is there any reason to run the jtreg tests twice?  If not, you could use `all-tests-nojtreg`, so the jtreg tests aren't run again.

smillst:
I finally figured out what the actual problem was.  The checker properties were passed to the framework build file.

wmdietl:
Thanks for tracking this down! I've reverted the invalid test and this PR should be ready to merge once tests pass.

Issue #1642, opened 2017-11-07
Making `checker/lib` available to Maven users


wmdietl:
As noticed in #1641 we currently don't provide access to the annotated libraries contained in `checker/lib` to Maven users.
We should add new artefacts, e.g. org.checkerframework.guava, that wrap these annotated libraries.
Then Maven users can specify these dependencies.

smillst:
The artefact name should be org.checkerframework.annotatedlib/guava.

wmdietl:
Looking at https://search.maven.org/#search%7Cga%7C1%7Cg%3A%22org.checkerframework%22 I also don't see a separate `framework` artifact, which should be separate from `checker`. A newly developed type system would only depend on `framework` and doesn't need all CF type systems.

PR #1647, opened 2017-11-09
Move libraries into a dedicated repository.

wmdietl:
As discussed a week ago, I've created a dedicated repository at:
https://github.com/typetools/annotated-libraries/

which contains both the pre-built `jdk8.jar` and all the annotated libraries that were previously in `checker/lib`.
A few references in `annotating-libraries.tex` still need to be updated.

This also relates to #1642 and how we're distributing these libraries as part of a release. Let's discuss.


wmdietl:
As discussed, let's merge this soon and then work on the release steps separately.
What remains:
- for a release, clone the `annotated-libraries` repository into `checker-framework/checker/annotated-libraries/`

Then two options:
- move the `checker-framework/checker/annotated-libraries/jdk8.jar` file into `checker-framework/checker/dist/`
- remove `checker-framework/checker/annotated-libraries/jdk8.jar` and rebuild it

The download-jdk target now downloads jdk8.jar of a particular commit. Should the release process use `master` of jdk8.jar? Or also that commit version?

smillst:
I meant that we should just make the changes for downloading the JDK from the git repo.  This will make it easier to make changes to it.  Then worry about the other libs when we have time to make the release script changes.

> Should the release process use master of jdk8.jar? Or also that commit version?

Currently the release processes creates its own JDK.  But I guess it makes more sense to use the version that were used in the Travis tests.  So that would be the commit version.

PR #1649, opened 2017-11-09
Avoid calling X.toString().equals(Y)

mernst:


javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -377,6 +381,90 @@ public static boolean hasInheritedMeta(AnnotationMirror anno) {
...
+        // Can't use deepEquals() to compare val1 and val2, because they might have mismatched
+        // AnnotationValue vs. CheckerFrameworkAnnotationValue, and AnnotationValue doesn't override
+        // equals().  So, write my own version of deepEquals().
+        if ((val1 instanceof List<?>) && (val2 instanceof List<?>)) {
+            List<?> list1 = (List<?>) val1;
+            List<?> list2 = (List<?>) val2;
+            if (list1.size() != list2.size()) {
+                return false;
+            }
+            // TODO: Should arrays/lists be compared setwise, in a way that ignores order?
```
smillst: No, order may matter.  For example, `@LTLengthOf(value={"a1","a2"}, offest={"0", "1"})` isn't the same as `@LTLengthOf(value={"a2","a1"}, offest={"0", "1"})`

checker/tests/nullness/PureTest.java
```diff
@@ -52,15 +52,13 @@ public void test() {
     }
 
     static void shortCircuitAnd(PureTest pt) {
-        if ((pt.getSuperclass() != null)
-                && pt.getSuperclass().toString().equals("java.lang.Enum")) {
+        if ((pt.getSuperclass() != null) && pt.getSuperclass().equals(Enum.class)) {
```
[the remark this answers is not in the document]
mernst: I agree tt doesn't make much difference.  However, I don't think it does harm, and with the change this line won't show up in future searches for the `.toString().equals(...)` code pattern.

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -638,10 +638,11 @@ private void annotateSupertypes(
                 // TODO: Java 7 added a few AutoCloseable superinterfaces to classes.
                 // We specify those as superinterfaces in the jdk.astub file. Let's ignore
                 // this addition to be compatible with Java 6.
+                String superTypeString = superType.toString();
                 assert foundType != null
-                                || (superType.toString().equals("AutoCloseable")
-                                        || superType.toString().equals("java.io.Closeable")
-                                        || superType.toString().equals("Closeable"))
+                                || (superTypeString.equals("AutoCloseable")
```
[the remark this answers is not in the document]
wmdietl: Ah, I was confused what is being checked for here... not sure it's worth checking for both.

framework/src/org/checkerframework/framework/util/DependentTypes.java
```diff
@@ -55,7 +54,9 @@ private AnnotationMirror findDependent(Element element) {
         // TODO: does this work with a .astub file?
         List<TypeCompound> tas = ((Symbol) element).getRawTypeAttributes();
         for (TypeCompound ta : tas) {
-            if (ta.getAnnotationType().toString().equals(Dependent.class.getCanonicalName())) {
+            if (ta.getAnnotationType()
+                    .toString()
+                    .equals("org.checkerframework.framework.qual.Dependent")) {
```
wmdietl: I would prefer class literals over string literals.

javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -377,6 +381,92 @@ public static boolean hasInheritedMeta(AnnotationMirror anno) {
...
+            return false;
+        }
+        return sameAnnotationValueValue(av1.getValue(), av2.getValue());
+    }
+
+    /**
+     * Return true if the two annotation values are the same. The arguments to this method are
+     * values that are returned by {@code AnnotationValue.getValue()}.
+     */
+    public static boolean sameAnnotationValueValue(Object val1, Object val2) {
```
wmdietl: Should this be a `private` helper method?

javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -377,6 +381,92 @@ public static boolean hasInheritedMeta(AnnotationMirror anno) {
...
+        } else if ((val1 instanceof Type.ClassType) && (val2 instanceof Type.ClassType)) {
+            // Type.ClassType does not override equals
+            return sameClassType((Type.ClassType) val1, (Type.ClassType) val2);
+        } else {
+            return Objects.equals(val1, val2);
+        }
+    }
+
+    // Type.ClassType does not override equals
+    public static boolean sameClassType(Type.ClassType t1, Type.ClassType t2) {
```
wmdietl: If this is meant as a public method, I would move it to `TypesUtils`.
The existing `org.checkerframework.javacutil.TypesUtils.areSamePrimitiveTypes(TypeMirror, TypeMirror)` has a similar purpose for primitives.
This could be `areSameDeclaredTypes`.
If it is not meant for general consumption, make it private.

javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -98,12 +99,15 @@ public static boolean areSame(
             return false;
         }
 
-        Map<? extends ExecutableElement, ? extends AnnotationValue> elval1 =
-                getElementValuesWithDefaults(a1);
-        Map<? extends ExecutableElement, ? extends AnnotationValue> elval2 =
-                getElementValuesWithDefaults(a2);
+        // This commented implementation is less efficient.  It is also wrong:  it requires a
+        // particular order for fields, and it distinguishes the long constants "33" and "33L".
```
wmdietl: How is the new code handling "33" and "33L" the same way? Wouldn't it reach the `Objects.equals(val1, val2)` and be different, if those two literals are in fact represented differently?

wmdietl: It's good to have these comments here for the review. But for the final PR they should be removed.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1352,10 +1351,10 @@ public Void visitAnnotation(AnnotationTree node, Void p) {
         }
 
         Element anno = TreeInfo.symbol((JCTree) node.getAnnotationType());
-        if (anno.toString().equals(DefaultQualifier.class.getName())
-                || anno.toString().equals(SuppressWarnings.class.getName())) {
-            // Skip these two annotations, as we don't care about the
-            // arguments to them.
+        String annoQualifiedName = anno.toString();
+        if (annoQualifiedName.equals("org.checkerframework.framework.qual.DefaultQualifier")
```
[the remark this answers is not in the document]
mernst: Using tring literals avoids recomputing toString on each and every call.
These particular annotations will never change their package.
But I can change it back if you really dislike it.

javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
mernst: The old code was treating `33` and `33L` differently, and the new code treats them as the same.  I don't have time to write a long comment about it, but this comment lets an interested reader do that investigation.

mernst: I want to retain them.  They give insight into what was wrong with the old implementation and why we went to the trouble to change it.  They caution against someone reinstating the old implementation.

javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
mernst: I was thinking it might possibly be useful to a caller who didn't want to use `sameAnnotationValue` for some reason.  We could make it `private` if you feel that it pollutes the `AnnotationUtils` interface.

mernst:
@wmdietl , I believe I have addressed your comments.

javacutil/src/org/checkerframework/javacutil/TypesUtils.java
```diff
@@ -163,6 +163,21 @@ public static boolean isPrimitive(TypeMirror type) {
...
+    /**
+     * Returns true iff the arguments are both the same primitive types.
+     *
+     * <p>This is needed because class {@code Type.ClassType} does not override equals.
+     *
+     * @return whether the arguments are the same primitive types
+     */
+    public static boolean areSameDeclaredTypes(Type.ClassType t1, Type.ClassType t2) {
+        // Do a cheaper test first
+        if (!t1.tsym.name.toString().equals(t2.tsym.name.toString())) {
```
wmdietl: Couldn't you further optimise this to `if (t1.tsym.name != t2.tsym.name)`?
The two `Name` instances are comparable by reference equality, as they are created by the same compiler instance.

PR #1650, opened 2017-11-09
Fix typo in javadoc.astub and add testing infrastructure

kelloggm:
I did a bad thing. In typetools#1613, I added a javadoc.astub file but I didn't add automated tests for it. This PR fixes that problem, and also removes a typo in the javadoc.astub file itself that the testing regime would have caught.

### DIFF
--- a/checker/build.xml
+++ b/checker/build.xml
@@ -827,10 +827,17 @@
         </antcall>
     </target>
 
-    <target name="index-tests" depends="jar,build-tests"
+    <target name="index-tests" depends="jar,build-tests, index-javadoc-astub-tests"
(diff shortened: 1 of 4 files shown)


wmdietl:
In #1647 I want to move all these annotated external libraries and the pre-built annotated JDK to a separate repository.
It's not a good idea to add a dependency on that astub file in this new test. So far the tests are independent of these external libraries.
However, instead of just downloading the annotated JDK, we could download all annotated libraries and then use them for the tests.

kelloggm:
That seems reasonable. If you'd like to close this until #1647 is resolved, feel free.

PR #1651, opened 2017-11-10
Clean up AnnotatedTypeMirror: remove commented out code and actually

wmdietl:
prevent adding annotations to AnnotatedExecutableType.

### DIFF
--- a/checker/src/org/checkerframework/checker/guieffect/GuiEffectTypeFactory.java
+++ b/checker/src/org/checkerframework/checker/guieffect/GuiEffectTypeFactory.java
@@ -469,16 +469,16 @@ public class GuiEffectTypeFactory extends BaseAnnotatedTypeFactory {
         public Void visitMethod(MethodTree node, AnnotatedTypeMirror type) {
             AnnotatedTypeMirror.AnnotatedExecutableType methType =
                     (AnnotatedTypeMirror.AnnotatedExecutableType) type;
-            Effect e = getDeclaredEffect(methType.getElement());
+            // Effect e = getDeclaredEffect(methType.getElement());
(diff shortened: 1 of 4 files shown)


checker/src/org/checkerframework/checker/guieffect/GuiEffectTypeFactory.java
```diff
@@ -469,16 +469,16 @@ public boolean hasExplicitEffect(ExecutableElement methElt) {
         public Void visitMethod(MethodTree node, AnnotatedTypeMirror type) {
             AnnotatedTypeMirror.AnnotatedExecutableType methType =
                     (AnnotatedTypeMirror.AnnotatedExecutableType) type;
-            Effect e = getDeclaredEffect(methType.getElement());
+            // Effect e = getDeclaredEffect(methType.getElement());
```
smillst: This doesn't seem to be related to cleaning up AnnotatedTypeMirror.  Nor does the other change to the GUI Effect Checker.

wmdietl: They are: on line 480 in this file and on line 243 in the other file an annotation is added to an `AnnotatedExecutableType`.
Neither of these assignments had any effects, so I commented out this practically dead code.
The change in ATM will start throwing an exception for these calls, which is the reason why I commented them out.
No other code used `e`, so I also commented that out.
There already was some commented-out code in both places. Would you prefer completely removing this code?

smillst: Oh, I see now.  The pull request comment didn't mention anything about commenting out dead code, so I was confused.

PR #1652, opened 2017-11-10
Update AnnotationClassLoader for consistency and clarity

jyluo:
- Updated loadExternalAnnotationClass(), used in -Aquals, to use the same low level loading process as loadExternalAnnotationClassesFromDirectory(), namely via loadAnnotationClass(). Prior to this, qualifiers explicitly named using -Aquals were assumed to be valid qualifiers and did not undergo the same consistency and support checks defined in loadAnnotationClass() (that it has `@Target` and `ElementType.TYPE_USE` / `ElementType.TYPE_PARAMETER`).

- Updated the checks in AnnotationClassLoader so that it only loads type annotation classes.

- Added factory method createAnnotationClassLoader() in ATF for creating a class loader. Subclasses can override and return their own AnnotationClassLoader.

- Added SubtypingAnnotationClassLoader to allow `@Unqualified` to be loaded for the Subtyping Checker. That annotation is forced to not have any `@Target` values and is used in SubtypingEncryptedTests.

- Renamed getAppClassLoader() to getClassLoader(), since the class loader returned can be the system class loader

- Updated loadAnnotationClass():
	- Added null check for the return of getClassLoader()
	- Casts the loaded class into an Annotation before checking it for @Target and ElementTypes

- Updated default implementation of isSupportedAnnotationClass() to always return true:
	- previously it would attempt to build an annotation class via AnnotationBuilder, which would fail for qualifiers that have parameters with no default values and cause such annotations to not be included in the supported qualifier set

- Few other variable renaming, removal of java type arguments, docs improvement, and formatting fixes

Tested: checker/all-tests, checker/javadoc, checker/example-tests-nobuildjdk


checker/src/org/checkerframework/checker/fenum/FenumAnnotatedTypeFactory.java
```diff
@@ -50,7 +50,11 @@ public FenumAnnotatedTypeFactory(BaseTypeChecker checker) {
         // load individually named qualifiers
         if (qualNames != null) {
             for (String qualName : qualNames.split(",")) {
-                qualSet.add(loader.loadExternalAnnotationClass(qualName));
+                Class<? extends Annotation> loadedClass =
```
jyluo: Undo null checks, because its never null, internally caught in loadExternalAnnotationClass.

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
```diff
@@ -418,13 +418,13 @@ private final boolean checkDirForPackage(
      * @return the classloader used to load the checker class, or the system classloader, or null if
      *     both are unavailable
      */
-    private final /*@Nullable*/ ClassLoader getAppClassLoader() {
+    private final /* @Nullable */ ClassLoader getClassLoader() {
         return InternalUtils.getClassLoaderForClass(checker.getClass());
     }
 
     /** Debug Use Displays all classpaths */
```
jyluo: add colon after Use, document more on what it does

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
```diff
@@ -460,23 +460,23 @@ private final void printPaths() {
...
-     * classes, into an immutable set.
+     * Gets the set of the annotation classes in the qual directory of each checker shipped with the
+     * Checker Framework. Note that the returned set from this method is mutable. This method is
+     * intended to be called within {@link AnnotatedTypeFactory#createSupportedTypeQualifiers()
+     * createSupportedTypeQualifiers()} (or its helper methods) to help define the set of supported
+     * qualifiers. {@link AnnotatedTypeFactory#createSupportedTypeQualifiers()
+     * createSupportedTypeQualifiers()} must return an immutable set, and it is the responsibility
+     * of that method (or helper methods it calls) to convert the set returned by this method, along
+     * with any additional annotation classes, into an immutable set.
      *
```
jyluo: add @see createSupportedTypeQualifiers

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
```diff
@@ -561,33 +561,26 @@ private final void printPaths() {
...
             final String annoName) {
-        try {
-            final Class<? extends Annotation> annoClass =
-                    Class.forName(annoName, true, getAppClassLoader()).asSubclass(Annotation.class);
-            return annoClass;
-        } catch (ClassNotFoundException e) {
+        Pair<Boolean, /* @Nullable */ Class<? extends Annotation>> loadResult =
+                loadAnnotationClass(annoName);
+        if (!loadResult.first) {
             checker.userErrorAbort(
```
jyluo: get rid of Pair, raise this elsewhere.

Remove method?

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
```diff
@@ -683,43 +676,52 @@ public int compare(File o1, File o2) {
...
-                Class<? extends Annotation> annoClass = cls.asSubclass(Annotation.class);
-
-                // see if the annotation is supported by a checker
+        // ensure that the freshly loaded class is an annotation
+        if (cls != null && cls.isAnnotation()) {
+            Class<? extends Annotation> annoClass = cls.asSubclass(Annotation.class);
+
+            // make sure the loaded annotation class has the @Target meta-annotation, and is a well
+            // defined qualifier or is the special qualifier @Unqualified
+            if (annoClass.getAnnotation(Target.class) != null
```
jyluo: Extract to helper method, to filter
see if we already have this logic somewhere

jyluo:
@wmdietl Ready for review!

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
```diff
@@ -683,40 +652,53 @@ public int compare(File o1, File o2) {
...
+     *     required ElementType values, and is a supported annotation by a checker. If the
+     *     annotation is not supported by a checker, null is returned.
      */
-    private final /*@Nullable*/ Class<? extends Annotation> loadAnnotationClass(
-            final String fullyQualifiedClassName) {
-        Class<?> cls = null;
+    protected final /*@Nullable*/ Class<? extends Annotation> loadAnnotationClass(
+            final String fullyQualifiedClassName, boolean issueWarnings) {
 
         try {
```
wmdietl: Limit the scope of try blocks, to only cover the lines that can actually throw that exception.

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
```diff
@@ -683,40 +652,53 @@ public int compare(File o1, File o2) {
...
-                    cls.getAnnotation(Target.class).value(), cls)) {
-                // if it is supported, then subclass it as an Annotation
-                // class
-                Class<? extends Annotation> annoClass = cls.asSubclass(Annotation.class);
-
-                // see if the annotation is supported by a checker
-                if (isSupportedAnnotationClass(annoClass)) {
-                    return annoClass;
+            ClassLoader classLoader = getClassLoader();
+            if (classLoader != null) {
```
wmdietl: You can reduce indentation by writing
```
if (classLoader == null) {
    return null;
}
```
Isn't that the only path that can reach the `return null` on line 704?
Make the control flow and returns easier to read.

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
```diff
@@ -683,40 +652,53 @@ public int compare(File o1, File o2) {
...
+                Class<?> cls = Class.forName(fullyQualifiedClassName, true, classLoader);
+
+                // see if the freshly loaded class is an annotation
+                if (cls != null && cls.isAnnotation()) {
```
wmdietl: Same here. Invert the condition and directly return null.

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
```diff
@@ -727,20 +709,26 @@ public int compare(File o1, File o2) {
...
         Set<Class<? extends Annotation>> loadedClasses =
                 new LinkedHashSet<Class<? extends Annotation>>();
 
         if (fullyQualifiedAnnoNames != null && !fullyQualifiedAnnoNames.isEmpty()) {
             // loop through each class name & load the class
             for (String fullyQualifiedAnnoName : fullyQualifiedAnnoNames) {
-                Class<? extends Annotation> annoClass = loadAnnotationClass(fullyQualifiedAnnoName);
-                if (annoClass != null) {
-                    loadedClasses.add(annoClass);
+                Class<? extends Annotation> anno =
```
wmdietl: Why not keep the `annoClass` name? Other local variables of this type have that name.

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
```diff
@@ -89,10 +86,11 @@
     private final URL resourceURL;
 
     /**
-     * The loaded annotation classes. Call {@link #getLoadedAnnotationClasses} rather than using
-     * this field directly as it may be null.
+     * The loaded annotation classes, which includes bunded and external annotations that are deemed
```
wmdietl: `bunded` -> `bundled`

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
```diff
@@ -89,10 +86,11 @@
...
+     * supported by each checker. Call {@link #getBundledAnnotationClasses()} rather than using this
```
wmdietl: What does the `each` quantify over? Should it simply be `the`?

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
```diff
@@ -126,18 +124,16 @@ public AnnotationClassLoader(final BaseTypeChecker checker) {
                         Pattern.compile(Character.toString(DOT), Pattern.LITERAL)
                                 .split(packageName)));
 
-        // Only load annotations if requested.  This avoids issuing an error
-        // if the qual package contains an annotation that is not a qualifier,
-        // but the checker does not try to use it as a qualifier.
-        loadedAnnotations = null;
+        // Only load annotations if requested.
```
wmdietl: This comment isn't really clear. Requested how?
Why is it initialized to `null` instead of to `getBundledAnnotationClasses()`?

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
```diff
@@ -561,43 +555,26 @@ private final void printPaths() {
      *
      * @param annoName canonical name of an external annotation class, e.g.
      *     "myproject.qual.myannotation"
-     * @return the loaded annotation class
+     * @return the loaded annotation class, or null if it was not a supported annotation as decided
```
wmdietl: Decided how?

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
jyluo: Fixed

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
jyluo: Fixed

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
jyluo: I've modified this part to just load the bundled annotation classes, thus the field will never be null.

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
jyluo: Updated comment to point to `isSupportedAnnotationClass()`

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
jyluo: Fixed

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
jyluo: Fixed

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
jyluo: Fixed

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
jyluo: Fixed

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
mernst: GitHub review tip: There is no need to answer "fixed".  That just clutters the conversation.
When you push a commit that changes this line, Werner's comment will be automatically closed.
Answer "fixed" only if the change is to some other line, so that GitHub will not automatically close the comment.
Thanks for the changes!

framework/src/org/checkerframework/framework/type/AnnotationClassLoader.java
jyluo: Update: some checkers like the index checker explicitly declares its supported qualifiers in each of its subcheckers, whist sharing a single qual directory across its subcheckers. If we load all of the bundled qualifiers here during the constructor of AnnotationClassLoader then it issues warnings for some of the qualifiers such as [@LengthOf](https://github.com/typetools/checker-framework/blob/master/checker/src/org/checkerframework/checker/index/qual/LengthOf.java) which requires that its `@Target` has `METHOD` in addition to `TYPE_USE`.

`getBundledAnnotationClasses()` is called by checkers that want to load and use its bundled annotations without explicitly declaring them. I'll clarify the comment to indicate this, but the initialization to null and subsequent replacement in `getBundledAnnotationClasses()` is intended.

jyluo:
Ready for review @wmdietl

checker/src/org/checkerframework/checker/i18n/I18nAnnotatedTypeFactory.java
```diff
@@ -26,6 +27,12 @@ public I18nAnnotatedTypeFactory(BaseTypeChecker checker) {
         this.postInit();
     }
 
+    @Override
+    protected AnnotationClassLoader createAnnotationClassLoader() {
+        // I18nChecker does not use a class loader
```
wmdietl: Just to be clear: this is only an optimization to not instantiate the `AnnotationClassLoader`, which would cause some initialization work.
All of that work would be wasted, because `createSupportedTypeQualifiers` below returns a fixed set of qualifiers.

So this could be simplified by not setting the `loader` field in the constructor at all. Then subclasses that override `createSupportedTypeQualifiers` would not incur an overhead, even without also having to override `createAnnotationClassLoader`.
Method `loadTypeAnnotationsFromQualDir` would first call `loadTypeAnnotationsFromQualDir` to set the `loader` field.
The three classes that do something with the loader would be the only ones that need to override this method.

What do you think?

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -851,8 +865,15 @@ public TypeArgumentInference getTypeArgumentInference() {
     @SuppressWarnings("varargs")
     private final Set<Class<? extends Annotation>> loadTypeAnnotationsFromQualDir(
             Class<? extends Annotation>... explicitlyListedAnnotations) {
-        // add the loaded annotations to the annotation set
-        Set<Class<? extends Annotation>> annotations = loader.getLoadedAnnotationClasses();
+        Set<Class<? extends Annotation>> annotations;
+
+        // If the loader is not disabled, add the annotations bunded in the qual directory to the
```
wmdietl: `bunded` -> `bundled`

checker/src/org/checkerframework/checker/i18n/I18nAnnotatedTypeFactory.java
jyluo: There's a few more classes that do use a loader, but yes I'll change it to instantiate a loader within `createSupportedTypeQualifiers`.

jyluo:
Ready for further review @wmdietl

smillst:
>Disabled units external demo make file as it used the loader to load units relationships, the mechanism for specifying relationships externally is being updated in a future PR.

This PR breaks the Units Checker, so we shouldn't merge it until the fix is also ready.  @jyluo do you have a branch that fixes the Units Checker?  Or do you still need to implement the fix?

jyluo:
@smillst The fix, and a big overhaul of Units Checker is under review here https://github.com/opprop/checker-framework/pull/57. I'm finishing up that PR and will file it to here when ready.

smillst:
>@smillst The fix, and a big overhaul of Units Checker is under review here opprop#57. I'm finishing up that PR and will file it to here when ready.

Great! Let's wait to merge this PR until opprop#57 has been approved.

jyluo:
@smillst: per discussions with @wmdietl I've extracted the fix from opprop#57 into here to not hold this PR back. A new -AunitsRelations option is added to Units Checker to load externally defined units relationships, and the external demo and manual are updated to reflect this change.

jyluo:
Update: Turns out it is not necessary to use a new command line flag, as long as the externally defined units relationship classes are not in the qual package it works.

docs/examples/units-extension/relations/FrequencyRelations.java
```diff
@@ -1,4 +1,4 @@
-package qual;
+package relations;
```
smillst: Why did this change package?

jyluo: Hmm upon further testing it was unnecessary to change the package. I've reverted this change.

checker/src/org/checkerframework/checker/units/UnitsAnnotatedTypeFactory.java
```diff
@@ -282,8 +284,8 @@ private boolean isUnitsMultiple(AnnotationMirror metaAnno) {
     }
 
     /**
-     * Look for an @UnitsRelations annotation on the qualifier and add it to the list of
-     * UnitsRelations.
+     * Look for an {@link org.checkerframework.checker.units.qual.UnitsRelations} annotation on the
```
smillst: Are the changes in this method related to this pull request or do they belong in a separate Units Checker pull request?

jyluo: In general I was cleaning up any loader related code, in this case minimizing the scope of the try/catch blocks to where the exceptions can be thrown. I'll move the changes here to another PR.

jyluo:
Description updated

checker/src/org/checkerframework/checker/units/UnitsAnnotatedTypeFactory.java
smillst: Thanks!  It makes it much easier to review a pull request if it only contains changes related to a single goal.

PR #1659, opened 2017-11-13
Make sure jdk exists before running tests that use it.

smillst:
Also, copy jdk.jar to dist after it is created.

### DIFF
--- a/checker/build.xml
+++ b/checker/build.xml
@@ -827,7 +827,7 @@
         </antcall>
     </target>
 
-    <target name="index-tests" depends="jar,build-tests"
+    <target name="index-tests" depends="jar,jdk.jar.exists,build-tests"
             description="Run tests for the Index Checker">
         <antcall target="-run-tests">
             <param name="param" value="tests.IndexTest"/>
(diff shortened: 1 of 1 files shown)


checker/build.xml
```diff
@@ -1508,6 +1508,8 @@
          <env key="LT_BIN" value="${LT_BIN}"/>
       </exec>
       <move file="jdk/jdk.jar" tofile="jdk/${jdkName}"/>
+        <!-- Copy jars to dist. If dist doesn't exist then the copy task creates it.-->
```
wmdietl: The indentation seems off.

checker/build.xml
```diff
@@ -1508,6 +1508,8 @@
...
+        <copy file="jdk/${jdkName}" tofile="dist/${jdkName}"/>
```
wmdietl: This should have `overwrite="true"` and maybe also `failonerror="false"` to account for an already existing file.

wmdietl: Target `dist-nobuildjdk` at the moment also copies around jdks. Those copies be removed now, right?

smillst: By default, if the file exists and is older than the copied file, it is overwritten.  Using `overwrite="true"`  overwrites existing files if they are newer than the copy.  I don't think we want that.

wmdietl: It's what `dist-nobuildjdk` was doing. Fine with me if it works without that option.

PR #1661, opened 2017-11-14
Reduce duplicated code in the stubparser

smillst:
Also, refactor some of the code so that it is clearer.

(The diffs make it hard to see the changes.  I recommend using the "split" view.)

### DIFF
--- a/framework/src/org/checkerframework/framework/stub/StubParser.java
+++ b/framework/src/org/checkerframework/framework/stub/StubParser.java
@@ -655,7 +655,7 @@ public class StubParser {
         }
     }
 
-    /** Adds type and declaration annotations to from {@code decl}. */
+    /** Adds type and declaration annotations from {@code decl}. */
     private void parseMethod(
             MethodDeclaration decl,
             ExecutableElement elt,
@@ -682,7 +682,7 @@ public class StubParser {
         typeParameters.removeAll(methodType.getTypeVariables());
     }
 
-    /** Adds type and declaration annotations to from {@code decl}. */
+    /** Adds type and declaration annotations from {@code decl}. */
     private void parseConstructor(
             ConstructorDeclaration decl,
             ExecutableElement elt,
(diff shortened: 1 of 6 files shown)


framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -841,111 +854,95 @@ private ClassOrInterfaceType unwrapDeclaredType(Type type) {
...
-            for (AnnotatedTypeVariable typePar : typeParameters) {
-                if (typePar.getUnderlyingType() == atype.getUnderlyingType()) {
-                    AnnotatedTypeMerger.merge(typePar.getUpperBound(), typeVarUse.getUpperBound());
-                    AnnotatedTypeMerger.merge(typePar.getLowerBound(), typeVarUse.getLowerBound());
-                }
-            }
-        }
-        if (typeDef.getAnnotations() != null && atype.getKind() != TypeKind.WILDCARD) {
-            annotate(atype, typeDef.getAnnotations());
+        // Primary annotations for the type of a variable declaration are not store in typeDef, but
```
wmdietl: store -> stored

framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -841,111 +854,95 @@ private ClassOrInterfaceType unwrapDeclaredType(Type type) {
...
+                    if (typePar.getUnderlyingType() == atype.getUnderlyingType()) {
+                        AnnotatedTypeMerger.merge(
+                                typePar.getUpperBound(), typeVarUse.getUpperBound());
+                        AnnotatedTypeMerger.merge(
+                                typePar.getLowerBound(), typeVarUse.getLowerBound());
+                    }
+                }
+                break;
+            default:
+                // No annotations to add.
```
wmdietl: Maybe `No further annotations to add.` would make it more explicit that the primary annotation was already added earlier.

wmdietl:
The downstream failure seems like a timeout only.

PR #1662, opened 2017-11-14
StubParser: Check that an annotation is a declaration annotation

### LINKED ISSUE #1356: Type annotation in stub file stored as declaration annotation
The stubparser stores some type annotations as declaration annotations.   The annotation is put in `AnnotatedTypeFactory#declAnnosFromStubFiles` which is then written out to bytecode.  This cause errors `type.invalid` when the byte code is later used. (Example below or `checker/jtreg/stubs/issue1356`.)  This only seems to happen if the type annotation has some argument.

~~This issue is blocking #1352, so we may want to add some work around such as removing type annotations from  `AnnotatedTypeFactory#declAnnosFromStubFiles`.~~  The ValueAnnotatedTypeFactory has been changed to work around this problem.  When this issue is fixed, please remove the work around.

File named MyClass.astub
```
import org.checkerframework.checker.regex.qual.*;

package mypackage;

class MyClass {
   static @ArrayLen(1) @Regex(1) String toString(char c);
}
```
```java
package mypackage;

public class MyClass {
    static String toString(char c) {
        return c + "";
    }
}
```
```java
package mypackage;

public class UseMyClass {
    void test() {
        String s = MyClass.toString('a');
    }
}
```

Compile MyClass:
```
javacheck -processor org.checkerframework.checker.regex.RegexChecker -Astubs=stub.astub mypackage/MyClass.java
```
Compile UseMyClass
```
javacheck -processor org.checkerframework.checker.regex.RegexChecker mypackage/UseMyClass.java
```
Output:

```
mypackage/UseMyClass.java:5: error: [type.invalid] [@Regex(1), @UnknownRegex] may not be applied to the type "@Regex(1) String"
        String s = MyClass.toString('a');
                                   ^
1 error
```

smillst:
This fixes #1356, so the workaround for that issue is removed. 

This fix also found that `@StaticallyExecutable` was missing the `CONSTRUCTOR` target.

### DIFF
--- a/framework/src/org/checkerframework/framework/stub/StubParser.java
+++ b/framework/src/org/checkerframework/framework/stub/StubParser.java
@@ -1019,7 +1019,7 @@ public class StubParser {
             if (annoMirror != null) {
                 Target target =
                         annoMirror.getAnnotationType().asElement().getAnnotation(Target.class);
-                // Only add the declaration annotation if the annotation applies ot the element.
+                // Only add the declaration annotation if the annotation applies to the element.
                 if (AnnotationUtils.getElementKindsForTarget(target).contains(elt.getKind())) {
                     annos.add(annoMirror);
                 }
--- a/framework/src/org/checkerframework/framework/stub/StubParser.java
+++ b/framework/src/org/checkerframework/framework/stub/StubParser.java
@@ -43,6 +43,7 @@ import com.github.javaparser.ast.type.Type;
 import com.github.javaparser.ast.type.TypeParameter;
 import com.github.javaparser.ast.type.WildcardType;
 import java.io.InputStream;
+import java.lang.annotation.Target;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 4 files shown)


framework/src/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1016,7 +1017,12 @@ private void annotateDecl(
         for (AnnotationExpr annotation : annotations) {
             AnnotationMirror annoMirror = getAnnotation(annotation, supportedAnnotations);
             if (annoMirror != null) {
-                annos.add(annoMirror);
+                Target target =
+                        annoMirror.getAnnotationType().asElement().getAnnotation(Target.class);
+                // Only add the declaration annotation if the annotation applies ot the element.
```
wmdietl: ot -> to

javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -347,6 +351,53 @@ public static boolean hasInheritedMeta(AnnotationMirror anno) {
...
+                        ElementKind.LOCAL_VARIABLE,
+                        ElementKind.RESOURCE_VARIABLE,
+                        ElementKind.EXCEPTION_PARAMETER);
+            case ANNOTATION_TYPE:
+                return EnumSet.of(ElementKind.ANNOTATION_TYPE);
+            case PACKAGE:
+                return EnumSet.of(ElementKind.PACKAGE);
+            case TYPE_PARAMETER:
+                return EnumSet.of(ElementKind.TYPE_PARAMETER);
+            case TYPE_USE:
```
wmdietl: Doesn't `TYPE_USE` apply to the same ElementKinds as `TYPE`? A comment here and maybe also in the javadoc would be useful in either case.

javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -347,6 +351,53 @@ public static boolean hasInheritedMeta(AnnotationMirror anno) {
...
+            default:
+                return EnumSet.noneOf(ElementKind.class);
```
wmdietl: How about raising an error in the default case? At the moment this is complete and if something gets added, we should explicitly handle it. Silently ignoring a new case is trickier to track down.

PR #1670, opened 2017-11-18
 Simplify PurityCheckerHelper

mernst:

PR #1671, opened 2017-11-18
 Permit "throw new SomeException();" in @Pure methods; fixes #951

### LINKED ISSUE #951: Purity and thrown exceptions
Consider this code: 

``` java
  @Pure
  public static int min(int[] a) {
    if (a.length == 0) {
      throw new ArrayIndexOutOfBoundsException("Empty array passed to min(int[])");
    }
    int result = a[0];
    for (int i = 1; i < a.length; i++) {
      result = Math.min(result, a[i]);
    }
    return result;
  }
```

(This test case appears in the repository as `framework/tests/flow2/Issue951.java`.)

This causes an error:

`warning: [purity.not.deterministic.not.sideeffectfree.object.creation] object creation not allowed in deterministic side-effect-free method`

The [documentation for `@Deterministic`](http://types.cs.washington.edu/checker-framework/api/org/checkerframework/dataflow/qual/Deterministic.html) says:

> A method is called deterministic if it returns the same value (according to ==) every time it is called with the same parameters and in the same environment.

but makes no mention of exceptions.

I propose that:
 - [x] The documentation should be clarified to indicate that `@Deterministic` is concerned with the return value during non-exceptional executions.  If an exception is thrown, it is allowed to be a different exception on each invocation (and generally should be, to capture the correct stack trace).
 - [ ] Relax the current conservative checking of `@Deterministic`.  Currently, a `@Deterministic` method `dm` is not allowed to contain any object creation or call to any non-deterministic method.  (The theory is that `dm` might use the result somehow to create `dm`'s return value.)  Instead, the expression in a `throw _expr_` statement that is not contained within any `try` statement should be allowed to be non-deterministic (for example, it can be a `new` expression or can call a non-deterministic method).  This is safe, because if the `throw` statement is executed, then that statement will throw some exception (even if the expression itself throws an exception).  So long as the thrown exception (and no other exception that `_expr_` might throw) is not caught within the same method (the checker should verify this), then the method will throw an exception, in which case determinism imposes no constraints.

mernst:
Code-review this after merging #1670


dataflow/src/org/checkerframework/dataflow/util/PurityChecker.java
```diff
@@ -183,15 +183,49 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void ignore) {
...
         public Void visitNewClass(NewClassTree node, Void ignore) {
-            Element methodElement = InternalUtils.symbol(node);
-            boolean sideEffectFree =
-                    (assumeSideEffectFree
-                            || PurityUtils.isSideEffectFree(annoProvider, methodElement));
-            if (sideEffectFree) {
-                purityResult.addNotDetReason(node, "object.creation");
-            } else {
-                purityResult.addNotBothReason(node, "object.creation");
+            // The expression in "throw EXPR;" is allowed to be non-@Deterministic, so long as it is
```
smillst: I was very confused when reading this comment because it is about throw expressions, but this method is `visitNewClass`.   It would be clearer to move this comment and the implementation to `visitThrow`.  (Just check if the thrown expression is a NewClassTree and if it is side effect free.)

mernst: Why is it necessary to check if the NewClassTree is side-effect-free?

mernst: The visitor must visit the arguments to the `new` expression, because one of its arguments might have a side effect.  Therefore, even if this code were moved into `visitThrow`, then there would still need to be a special case in `visitNew`.  (Or the code in throw would have to iterate over and visit each of the arguments to `new`; is that what you are suggesting?)

smillst: Oh right, the NewClassTree doesn't have to be side effect free, but the arguments should be.

smillst: If you move the code to visitThrow, then it would need to iterate over the arguments.  Which would make it possible to implement this TODO:
```
  // TODO: if okThrow, permit arguments to the newClass to be non-deterministic (don't add
  // those to purityResult), but still don't permit them to have side effects.
```

If you don't want to do that right now, that's fine.  I was just confused about why there was a long comment about throw expressions in `visitNewClass`.

mernst:
@smillst, I added some comments indicating the desirability of moving the code to a different method in the future.  More importantly, I added a check of whether the constructor itself is `@SideEffectFree`.  Could you take another look?  Thanks.

dataflow/src/org/checkerframework/dataflow/util/PurityChecker.java
```diff
@@ -183,15 +183,55 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void ignore) {
...
-            } else {
+            if (!sideEffectFree && !deterministic) {
                 purityResult.addNotBothReason(node, "object.creation");
+            } else if (!deterministic) {
+                purityResult.addNotDetReason(node, "object.creation");
+            } else if (!sideEffectFree) {
+                purityResult.addNotSEFreeReason(node, "object.creation");
             }
+
+            // TODO: if okThrow, permit arguments to the newClass to be non-deterministic (don't add
```
smillst: `okThrow` -> `okThrowDeterministic`

PR #1675, opened 2017-11-20
All characters are NonNegative (JLS 4.2)

kelloggm:
By JLS 4.2, characters are unsigned 16 bit integers. This means that the Index Checker can treat them as `@NonNegative` implicitly, removing some false positives. See kelloggm#192 and [JLS 4.2](https://docs.oracle.com/javase/specs/jls/se7/html/jls-4.html#jls-4.2)

### DIFF
--- a/checker/build.xml
+++ b/checker/build.xml
@@ -1452,8 +1452,11 @@
     </target>
 
     <target name="download-jdk" description="Downloads jdk8.jar from checkerframework.org/dev-jdk to checker/jdk and checker/dist">
-        <get src="https://checkerframework.org/dev-jdk/jdk8.jar" dest="jdk/jdk8.jar"
+        <get src="https://github.com/kelloggm/checker-framework/releases/download/character-nn/jdk8.jar" dest="jdk/jdk8.jar"
              usetimestamp="true"/>
+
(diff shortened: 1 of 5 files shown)


mernst:
@kelloggm The test failure is because the annotated JDK has changed since you started work.  You need to pull from master, rebuild the JDK, and re-upload it; then I believe the tests will pass.

kelloggm:
Thanks for checking this. I'll do that.

kelloggm:
Mike was correct about why the tests weren't passing. @panacekcz, could you review this, please? Thanks!

kelloggm:
I've pushed the test case you suggested into this branch, so I expect the tests to fail. I also fixed two of the problems (the ones with `PolyLowerBound` and `LowerBoundBottom`. I think I should also fix the others if we are going to merge this pull request. Do you think it's worthwhile to do so, or would it be more prudent to simply abandon this and just suppress these warnings?

panacekcz:
I think the linked issues should be fixed sooner or later anyway, after that we can come back here.

kelloggm:
This PR is closed for now, as it needs significantly more work to bring to a mergable state that I don't have time to do for now. I will reopen it once that work is completed.

PR #1679, opened 2017-11-21
Stop infinite recursion in sameAnnotationValue.

smillst:
This fixes the current sparta build failure. (https://travis-ci.org/typetools/sparta/builds/305055418) 

When the AnnotationValue is an annotation, the underlying object is a `com.sun.tools.javac.code.Attribute.Compound`which extends `Attribute` which extends `AnnotationValue`. `Compound#getValue` returns itself, so an infinite recursion occurs.  Since `Compound` also implements `AnnotationMirror`, use `areSame` instead.

### DIFF
--- a/javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
+++ b/javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
@@ -510,6 +510,8 @@ public class AnnotationUtils {
                 }
             }
             return true;
+        } else if ((val1 instanceof AnnotationMirror) && (val2 instanceof AnnotationMirror)) {
+            return areSame((AnnotationMirror) val1, (AnnotationMirror) val2);
         } else if ((val1 instanceof AnnotationValue) && (val2 instanceof AnnotationValue)) {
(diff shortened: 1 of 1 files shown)


javacutil/src/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -510,6 +510,8 @@ private static boolean sameAnnotationValueValue(Object val1, Object val2) {
                 }
             }
             return true;
+        } else if ((val1 instanceof AnnotationMirror) && (val2 instanceof AnnotationMirror)) {
+            return areSame((AnnotationMirror) val1, (AnnotationMirror) val2);
         } else if ((val1 instanceof AnnotationValue) && (val2 instanceof AnnotationValue)) {
```
smillst: @mernst I'm not sure when/if an AnnotationValue's value is an AnnotationValue.  Do you know?

mernst: Here is an example from the [JLS](https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.7.3):
```@Author(@Name(first = "Joe", last = "Hacker"))```

PR #1680, opened 2017-11-21
Manual instructions for annotating a custom fixed-size collection

kelloggm:
Users of the Index Checker will likely want to annotate their own collections so that the Index Checker treats them as if they were arrays or Strings. Since it's not obvious how to do that, this PR includes an example of doing so (and the associated test case) as a new manual section.

### DIFF
--- a/checker/tests/index/ArrayWrapper.java
+++ b/checker/tests/index/ArrayWrapper.java
@@ -1,0 +1,43 @@
+import org.checkerframework.checker.index.qual.IndexFor;
+import org.checkerframework.checker.index.qual.LengthOf;
+import org.checkerframework.checker.index.qual.NonNegative;
+import org.checkerframework.checker.index.qual.SameLen;
+
+/**
+ * A test for https://github.com/kelloggm/checker-framework/issues/154
+ *
+ * <p>This class wraps an array, but doesn't expose the array in its public interface. This test
+ * ensures that indexes for this new collection can be annotated as if the collection were an array.
+ *
+ * <p>ArrayWrapper is a fixed-size, weakly typed generic collection.
+ */
+@SuppressWarnings("unchecked")
+public class ArrayWrapper<T> {
+    private final Object @SameLen("this") [] internalRep;
+
+    ArrayWrapper(@NonNegative int size) {
+        @SuppressWarnings(
+                "index") // use the SameLen checker to ensure that anything that's an index for "this" is an
+        // index for the internal array representation. This suppress warnings asserts this fact.
+        final Object @SameLen("this") [] tmp = new Object[size];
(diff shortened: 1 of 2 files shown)


checker/tests/index/ArrayWrapper.java
```diff
@@ -0,0 +1,47 @@
+// Test case for https://github.com/kelloggm/checker-framework/issues/154
+// This class wraps an array, but doesn't expose the array in its public interface. This test
+// ensures that indexes for this new collection can be annotated as if the collection were an array.
+
+import org.checkerframework.checker.index.qual.IndexFor;
+import org.checkerframework.checker.index.qual.LengthOf;
+import org.checkerframework.checker.index.qual.NonNegative;
+import org.checkerframework.checker.index.qual.SameLen;
+
+/** ArrayWrapper is a fixed-size, weakly-typed generic collection. */
```
mernst: In what sense is this weakly-typed?
The abstraction looks strongly-typed to me.  (The implementation uses a cast, but that doesn't jeopardize type safety.)
This remark also appears in the manual section; keep it in sync with this code.

docs/manual/index-checker.tex
```diff
@@ -46,6 +46,12 @@ \chapter{Index Checker for sequence bounds (arrays and strings)\label{index-chec
 write an \refqualclasswithparams{checker/index/qual}{IndexFor}{myArray}
 annotation on the formal parameter's types.
 
+The Index Checker checks fixed-size data structures, whose size is never
+changed after creation.  A fixed-size data structure has no \<add()> or
+\<remove> operation.  Examples are strings and arrays, and you can add
```
panacekcz: Why inconsistent parentheses `add()` vs `remove`?

checker/tests/index/ArrayWrapper.java
kelloggm: I think this was just a mistake in my memory. For some reason, I remembered Java collection classes being classified as either "strongly" typed (backed by a generic array - so `internalRep` would be a `T[]`) or "weakly" typed (backed by an array of `Object`s). I can't corroborate this distinction with some quick googling, so I'll just remove the reference (it's not really important anyway).

checker/tests/index/ArrayWrapper.java
```diff
@@ -2,12 +2,15 @@
 // This class wraps an array, but doesn't expose the array in its public interface. This test
 // ensures that indexes for this new collection can be annotated as if the collection were an array.
 
+// Note that there is a copy of this code in the manual in index-checker.tex. If this code is
+// updated, you MUST update that copy, as well.
```
mernst: Should cross-reference from the manual to the `.java` file, too.

PR #1683, opened 2017-11-23
Makes code handling constructor invocations more uniform.

wmdietl:


wmdietl:
I filed #1684 with a few follow-up clean-ups. Mier recently ran into this mismatch between `symbol` and `constructor`. I highlight this now in the javadoc, but further cleanups should follow.

smillst:
Nice clean up! Thanks!

PR #1691, opened 2017-11-29
Fix crash on raw types.

### LINKED ISSUE #1690: Crash on generics with intersection types
The following is a valid Java program that compiles with `javac 1.8.0_131` (and with version `2.1.14` of the Checker Framework):

```
package com.example;

import java.io.Serializable;
import java.util.List;

public class Example<T extends Runnable & Serializable> {

  public Example() {
  }

  // Can be an inner type or in its own file, shouldn't matter.
  public static interface Example2<R extends Example> {
    public R getR();
  }
}
```

However, running it through the current master of the Checker Framework produces an `java.lang.IllegalArgumentException`. See the command below (substituting `$CHECKER_JAR` for the path to the `checker.jar`):

```
javac -Xbootclasspath/p:$CHECKER_JAR \
   -cp $CHECKER_JAR \
   -processor org.checkerframework.checker.guieffect.GuiEffectChecker \
   -AprintErrorStack Example.java
```

The (partial) stack trace is as follows: 

```
error: SourceChecker.typeProcess: unexpected Throwable (IllegalArgumentException) while processing Example.java; message: java.lang.Object&java.lang.Runnable&java.io.Serializable
  Compilation unit: Example.java
  Exception: java.lang.IllegalArgumentException: java.lang.Object&java.lang.Runnable&java.io.Serializable; Stack trace: com.sun.tools.javac.model.JavacTypes.getWildcardType(JavacTypes.java:209)
  org.checkerframework.framework.type.BoundsInitializer$InitializerVisitor.getJavaType(BoundsInitializer.java:607)
  org.checkerframework.framework.type.BoundsInitializer$InitializerVisitor.initializeTypeArgs(BoundsInitializer.java:546)
  org.checkerframework.framework.type.BoundsInitializer$InitializerVisitor.visitDeclared(BoundsInitializer.java:316)
  org.checkerframework.framework.type.BoundsInitializer$InitializerVisitor.visitDeclared(BoundsInitializer.java:265)
  org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedDeclaredType.accept(AnnotatedTypeMirror.java:849)
  org.checkerframework.framework.type.BoundsInitializer$InitializerVisitor.visit(BoundsInitializer.java:304)
[...]
```

The underlying cause seems to be that `com.sun.tools.javac.model.JavacTypes.getWildcardType` doesn't support arguments for which `((Type)arg).getKind()` is `TypeKind.INTERSECTION`, as is the case for the `Runnable & Serializable` bound. This method is called when trying to resolve the type of `R` in the `Example2` interface. 

Since `JavacTypes` is not part of the checker framework itself, I am not really sure where is the right place to fix this issue. An idea that occurs to me is to subclass `JavacTypes` and make `org.checkerframework.framework.type.BoundsInitializer.types` an object of that subclass, then override `getWildcardType` with something that can handle `TypeKind.INTERSECTION`. But I suspect you guys might have a better solution.

cc: @msridhar 

Some context for the curious: This crash is not life-or-death urgent for us, but it is keeping us on an old version of the CF (+ cherry-picked patches).

smillst:
Fixes #1690

### DIFF
--- a/framework/src/org/checkerframework/framework/type/BoundsInitializer.java
+++ b/framework/src/org/checkerframework/framework/type/BoundsInitializer.java
@@ -73,31 +73,9 @@ public class BoundsInitializer {
         for (int i = 0; i < typeElement.getTypeParameters().size(); i++) {
             TypeMirror javaTypeArg;
             if (declaredType.wasRaw()) {
-                TypeMirror upperBound =
-                        ((TypeVariable) typeElement.getTypeParameters().get(i).asType())
-                                .getUpperBound();
-                switch (upperBound.getKind()) {
-                    case ARRAY:
-                    case DECLARED:
-                    case TYPEVAR:
-                        javaTypeArg =
-                                declaredType.atypeFactory.types.getWildcardType(upperBound, null);
-                        break;
-                    case INTERSECTION:
-                        // Can't create a wildcard with an intersection as the upper bound, so use
-                        // an unbound wildcard instead.  The extends bound of the
-                        // AnnotatedWildcardType
-                        // be initialized properly below.
-                        javaTypeArg = declaredType.atypeFactory.types.getWildcardType(null, null);
-                        break;
-                    default:
-                        ErrorReporter.errorAbort(
-                                "Unexpected upper bound kind: %s type: %s",
-                                upperBound.getKind(), upperBound);
-                        javaTypeArg = null; // dead code
-                        break;
-                }
-
+                TypeVariable typeVariable =
+                        (TypeVariable) typeElement.getTypeParameters().get(i).asType();
+                javaTypeArg = getUpperBoundAsWildcard(typeVariable, declaredType.atypeFactory);
             } else {
                 javaTypeArg = declaredType.getUnderlyingType().getTypeArguments().get(i);
             }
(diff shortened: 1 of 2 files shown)


framework/src/org/checkerframework/framework/type/BoundsInitializer.java
```diff
@@ -133,6 +111,31 @@ public static void initializeTypeArgs(AnnotatedDeclaredType declaredType) {
...
+        TypeMirror upperBound = typeVariable.getUpperBound();
+        switch (upperBound.getKind()) {
+            case ARRAY:
+            case DECLARED:
+            case TYPEVAR:
+                return factory.types.getWildcardType(upperBound, null);
+            case INTERSECTION:
+                // Can't create a wildcard with an intersection as the upper bound, so use
+                // an unbound wildcard instead.  The extends bound of the
+                // AnnotatedWildcardType be initialized properly elsewhere.
```
wmdietl: `be` -> `will be`

wmdietl: I'm confused by this comment that the upper bound is properly initialized elsewhere.
Why do we need the code in this location at all then? Why not always return an unbounded wildcard, if logic to initialize the bound already is elsewhere?

smillst: The upper bound of the AnnotatedWildcardType is initialized else where, but this method is returning the type mirror that will be used as the underlying type of the AnnotatedWildcardType.  It's best if upper bound of underlying type matches the upper bound of the AnnotatedWildcardType.

wmdietl: It would be useful to make that mismatch explicit: that here you create an unbound wildcard as the underlying type, which will have a different bound than the AnnotatedWildcardType. I had read the comment as saying that the bound for the unbound wilcard will eventually be set later.
Can you reference where the AWT bound is correctly set?
Would it make sense to also set the bound of the underlying type when the bound of the AWT is changed?

PR #1692, opened 2017-11-29
cleared ambiguity in intellij installation instructions

Haaris710:


docs/manual/external-tools.tex
```diff
@@ -767,20 +767,25 @@ \section{IntelliJ IDEA\label{intellij}}
...
-  Add the annotated JDK library.
+  Add the annotated JDK library using the following method.
   Go to the ``Settings'' menu via (File>Settings) or (Ctrl-Alt-S) and
   navigate to the ``Java Compiler'' sub-menu via (Build, Execution,
   Deployment>Compiler>Java Compiler).  Add the following to the field
   ``Additional command line parameters'': \\
   \<-Xbootclasspath/p:\$CHECKERFRAMEWORK/checker/dist/jdk8.jar>
+  Note: It is a good idea to enclose the address in quotation
+  marks to escape special characters and spaces.
+  E.g. \<-Xbootclasspath/p:"\$CHECKERFRAMEWORK/checker/dist/jdk8.jar">
```
mernst: How about making this change on line 775, so you don't have to add lines 776-778?

docs/manual/external-tools.tex
```diff
@@ -767,20 +767,25 @@ \section{IntelliJ IDEA\label{intellij}}
...
 \item
   Create an annotation profile and specify which checkers to run.
   Directly under the ``Java Compiler'' sub-menu is the ``Annotation
   Processors'' sub-menu. Navigate here and click on the green ``+'' to create
-  a new Annotation Processor profile. Select the modules you would like to
-  use checkers with from under the ``Default'' profile and move them to your
-  new profile via the button directly right of the ``-'' button.
+  a new Annotation Processor profile. Give this profile a unique name of your
+  choosing. Avoid using special characters or spaces. E.g. AnnotationProcessor.
+  Select the modules you would like to use checkers with from under the
```
mernst: Small nit: please don't refill paragraphs.  Doing so makes the diffs larger, and makes it harder for others to see what has changed.  Ragged paragraphs in the `.tex` source are not a problem.

docs/manual/external-tools.tex
```diff
@@ -767,20 +767,25 @@ \section{IntelliJ IDEA\label{intellij}}
...
 
 \item
   Create an annotation profile and specify which checkers to run.
   Directly under the ``Java Compiler'' sub-menu is the ``Annotation
   Processors'' sub-menu. Navigate here and click on the green ``+'' to create
-  a new Annotation Processor profile. Select the modules you would like to
-  use checkers with from under the ``Default'' profile and move them to your
-  new profile via the button directly right of the ``-'' button.
+  a new Annotation Processor profile. Give this profile a unique name of your
+  choosing. Avoid using special characters or spaces. E.g. AnnotationProcessor.
```
mernst: The sentence "E.g. AnnotationProcessor." is a ffragmentt.  Please combine these three sentences into one.  That will also clarify that "Select" is talking about a new task.

docs/manual/external-tools.tex
Haaris710: Duly noted. I will try to avoid this in future. Thank you

docs/manual/external-tools.tex
Haaris710: Sure thing, will do.

docs/manual/external-tools.tex
Haaris710: Sure thing. I'd like to leave the note on the special characters in however, I discussed with @wmdietl  and I think it would be valuable information for an inexperienced developer trying to use the tool.

mernst: I don't think that the Checker Framework manual is a place to include general advice for inexperienced programmers.  Let's just give the exact text that the programmer needs to type in -- and let's not give the incorrect information first with the correct information in a note afterward.  

Also, the formatting of the note on the same line with `.../jdk8.jar` is confusing; you would have wanted a line break, but just removing lines 776-778 resolves the problem.

Haaris710: Sure thing. I'll make those changes. Thank you

Haaris710:
@mernst I have made the requested changes. Apologies for the reversions in the commit history, I must have pulled in some other changes accidentally. I hope this is okay now, please let me know if there's anything else I need to do on my end.

mernst:
Don't worry about the commit history; it will be cleaned up when the commit is merged.
The paragraph refilling still needs to be undone.  I will do that and then merge.

PR #1702, opened 2017-12-04
Add index annotations to Arrays and String

panacekcz:
This PR adds annotations to index JDK and value stub file, which will help typecheck Guava.

- Adds `@SameLen` and `@PolyValue` annotations to string methods that return the same value.
- Replaces `@PolyLength` on `String.intern` by `@SameLen` and `@PolyValue` according to kelloggm#196
- Adds `@MinLen(2)` annotations to `Arrays` methods that return string surrounded by square brackets or `"null"`. This helps typecheck code that removes the brackets by taking a substring.

#### `@SameLen` not applied to `String` constructors
`@SameLen` is not applied to constructors on `String`. Doing so results in failures in code such as:
```java
public class SameLenNewChar {
	void m() {
		String b = new String(new char[] {'a'});
	}
}
```
Output:
```
SameLenNewChar.java:3: error: [constructor.invocation.invalid] creation of @SameLen("new char[]") String <init>(@SameLenUnknown char @SameLenUnknown [] p0) not allowed with given receiver;
		String b = new String(new char[] {'a'});
		           ^
  found   : @SameLen("[error for expression: new char[]; error: 'new char']") String
  required: @SameLen("new char[]") String
SameLenNewChar.java:3: error: [expression.unparsable.type.invalid] Expression in dependent type annotation invalid: [error for expression: new char[]; error: 'new char']
		String b = new String(new char[] {'a'});
		           ^
2 errors
```
Occurs in plume-lib: https://travis-ci.org/panacekcz/checker-framework/jobs/311294966


panacekcz:
Travis failure in plume-lib should be solved by mernst/plume-lib#35.

smillst:
FYI, we've changed where the jdk.jar is downloaded.  See  https://github.com/typetools/annotated-libraries/blob/master/README.jdk for an explanation.

PR #1716, opened 2017-12-20
Skip Upper Bound Checker warnings on uses of sequence base classes

panacekcz:
As a partial fix for kelloggm#154 and kelloggm#197, this PR skips warnings on uses of base classes of sequences which contain methods that change the length and have both mutable and immutable implementations.

In order to do that, `SourceChecker` had to be modified to allow overriding `shouldSkipUses`, so that we can ignore these classes in addition to classes matching the pattern specified by the user.

The base classes are:
- `java.util.List`
- `java.util.AbstractList`
- ~`java.lang.CharSequence`~

Index jdk chages:
- add upper bound annotations to index parameters in `java.util.List` and `java.util.AbstractList`
- ~disable upper bound annotations outside `java.lang.CharSequence` on parameters that are an index for a `CharSequence` parameter~


panacekcz:
@smillst or @wmdietl, could you review this PR? Your review is required because of the change to `SourceChecker`.

panacekcz:
Travis failure (typechecking plume-lib) should be fixed by mernst/plume-lib#37.

checker/jdk/index/src/java/io/CharArrayWriter.java
```diff
@@ -201,7 +201,7 @@ public CharArrayWriter append(CharSequence csq) {
      *
      * @since  1.5
      */
-    public CharArrayWriter append(CharSequence csq, @IndexOrHigh("#1") int start, @IndexOrHigh("#1") int end) {
+    public CharArrayWriter append(CharSequence csq, @NonNegative /*!IndexOrHigh("#1")*/ int start, @NonNegative /*!IndexOrHigh("#1")*/ int end) {
```
wmdietl: What is the purpose of all these `/*!...*/` comments?

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundChecker.java
```diff
@@ -17,6 +18,23 @@
 @SuppressWarningsKeys({"index", "upperbound"})
 public class UpperBoundChecker extends BaseTypeChecker {
 
+    private HashSet<String> collectionBaseClasses;
+
+    public UpperBoundChecker() {
+        collectionBaseClasses = new HashSet<>(3);
+        collectionBaseClasses.add("java.util.List");
```
wmdietl: How about using the `.class` literals to get the correct name? This code is executed only once, so it shouldn't have a performance impact.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundChecker.java
```diff
@@ -17,6 +18,23 @@
...
+        collectionBaseClasses.add("java.util.AbstractList");
+        collectionBaseClasses.add("java.lang.CharSequence");
+    }
+
+    @Override
+    public boolean shouldSkipUses(String typeName) {
```
wmdietl: Wouldn't it be easier and clearer to simply use the top annotation for the problematic methods in these three classes?
I'm not quite sure what problem this PR is trying to solve.

checker/jdk/index/src/java/io/CharArrayWriter.java
panacekcz: The purpose is to retain the information about the upper bounds of the annotated parameters in the source code, without the upper bounds being actually checked. The `/*!` comments are already used for this purpose elsewhere.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundChecker.java
panacekcz: We want to be able to write these annotations when overriding these methods (to typecheck the body). If we just have top annotations in the base class, we get `override.param.invalid` warnings when writing a more specific annotation in the overriding method.

This PR helps reduce the number of `@SuppressWarnings` that need to be written in Guava when checked by the Index Checker:
- in code overriding `AbstractList` methods
- in code using `CharSequence` (makes consistent the way `CharSequence` and `List` are annotated)

panacekcz:
@wmdietl, I have addressed your comments.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundChecker.java
wmdietl: If this is only useful for Guava, and other projects that implement these collection classes, how about simply providing the corresponding `-AskipUses=xxx` command-line flags to those projects?
Would the suppressed warnings not be useful for projects that are mere users of the collection classes?
Shouldn't there at least be some comment in the manual about what the Index Checker guarantees for these classes?

framework/src/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -1922,7 +1922,7 @@ public final boolean shouldSkipUses(Element element) {
      * @param typeName the fully-qualified name of a type
      * @return true iff the enclosing class of element should be skipped
      */
-    public final boolean shouldSkipUses(String typeName) {
```
wmdietl: If we do this, can you add some comment to the javadoc when this method should be overridden.

checker/jdk/index/src/java/io/CharArrayWriter.java
wmdietl: How about adding a `checker/jdk/index/README` file that documents this syntax?

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundChecker.java
```diff
@@ -17,6 +18,28 @@
 @SuppressWarningsKeys({"index", "upperbound"})
 public class UpperBoundChecker extends BaseTypeChecker {
 
+    private HashSet<String> collectionBaseTypeNames;
+
+    public UpperBoundChecker() {
+        // These classes are bases for both mutable and immutable sequence collections
+        // Upper bound checker warnings are skipped at uses of them
```
smillst: I don't understand why the warnings are being skipped.

smillst: Mike explained this to me.

smillst:
@panacekcz If you would please address, @wmdietl comment, then I'll review it. Thanks!

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundChecker.java
panacekcz: We cannot use command-line arguments,, because we want this to apply only in the Upper Bound checker, not other subcheckers such as the Lower Bound checker (we want the lower-bound annotations checked).

We consider `List` not to be a fixed-length sequence, because it provides methods that change its length (`CharSequence` is now not included in this PR). The index checker is designed to only check fixed length sequences, so we do not want uses of `List` checked, regardless of whether the object is actually fixed-size or not, or whether the code uses the mutating methods.

panacekcz:
@smillst, I have added the requested comment.

PR #1718, opened 2017-12-22
 JDK annotations: collections that don't permit null.  Fixes #1714

### LINKED ISSUE #1714: ArrayDeque.offer(T) cannot accept null
I find that ArrayDeque.offer(T) is modeled to be able to take a nullable argument. However, according to its javadoc, it will throw an NPE is the parameter is null.

(https://github.com/typetools/checker-framework/blob/master/checker/jdk/nullness/src/java/util/ArrayDeque.java#L26)

public boolean offer(E e)
Inserts the specified element at the end of this deque.
This method is equivalent to offerLast(E).

Specified by:
offer in interface Deque<E>
Specified by:
offer in interface Queue<E>
Parameters:
e - the element to add
Returns:
true (as specified by Queue.offer(E))
Throws:
NullPointerException - if the specified element is null

mernst:


msridhar:
@mernst this may be outside the scope of this PR, but in some models like [`LinkedList`](https://github.com/typetools/checker-framework/blob/master/checker/jdk/nullness/src/java/util/LinkedList.java#L10) I see `class LinkedList<E extends @Nullable Object>`, and I *think* the `extends @Nullable Object` part is unnecessary.  Is that correct?  If so, I can file a follow-up issue on cleaning that up

wmdietl:
The annotated JDK classes are compiled with the corresponding checkers and the normal source defaults are applied. So yes, `class List<E>` would have the same effect. However, the comment on the previous line also calls out that null elements are allowed, so this might have been done to make that explicit.
The collection classes I looked at all made the upper bound annotations explicit, but this likely comes from a time when we didn't apply defaults to the JDK sources.

mernst:
You are both right:
 * the annotations are not necessary
 * in some cases they are there for emphasis (but still not necessary)
 * they used to be necessary before we made defaulting more consistent

I would accept a pull request that removed them, but I don't think it belongs in this pull request.

PR #1721, opened 2017-12-27
Index JDK annotations

panacekcz:
Improvements to Index JDK annotations:
- add lower bound annotations to `java.lang.AbstractStringBuilder`, `java.lang.StringBuilder`, `java.lang.Enum`
- add index annotations to `java.lang.reflect.Array`
- add missing lower bound annotations to `java.util.AbstractList`
- remove wrong annotations from `java.lang.String`


checker/jdk/index/src/java/lang/String.java
```diff
@@ -1735,7 +1735,7 @@ private int lastIndexOfSupplementary(int ch, int fromIndex) {
      * @param   target       the characters being searched for.
      * @param   fromIndex    the index to begin searching from.
      */
-    static @LTEqLengthOf("this") @SubstringIndexFor(value="this",offset="#1.length()-1") int indexOf(char[] source, int sourceOffset, int sourceCount,
+    static int indexOf(char[] source, int sourceOffset, int sourceCount,
```
kelloggm: Is there a reason that these annotations are wrong, or are you just removing these because these methods aren't public?

panacekcz: They are wrong - referring to `this` in a static method. Because the methods aren't public, I removed them rather than trying to fix them.

kelloggm: You're right. I missed that these were static.

PR #1722, opened 2017-12-27
Fix flow2 test readme [ci skip]

panacekcz:
The command stated in the README for flow2 tests does not work.


mernst:
Thanks for noticing this typo.

Issue #1606, opened 2017-10-20
@EnsuresQualifierIf annotation that supports annotation arguments


mernst:
The Checker Framework needs a variant of `@EnsuresQualifier` and `@EnsuresQualifierIf` that establishes an annotation with arguments.
Perhaps this could be done by adding a `qualifierArgs` field that defaults to null.

This would eliminate the need for:
 * RegexTransfer special cases for `isRegex(String, int)` and `asRegex(String, int)`.
 * KeyForTransfer special cases for `Map.put` and `Map.containsKey`.
 * 38 `@SuppressWarnings` in Daikon.

mernst:
A general qualifier would be nice to have.  However, its syntax in a program would be ugly.

We decided that, for now, we will hard-code the three cases that have come up:  the Regex, KeyFor, and Index type systems.

**Regex**:
There has not been a need for EnsuresRegexIf in user code yet, so we will skip that for now.

**KeyFor**:
Given these declarations:
```java
  VarInfoAux via;
  String key;
```
If `via.hasValue(key)` returns true, then key should receive the annotation `@KeyFor("via.map").`
This can be expressed by annotating the `hasValue()` method:
```java
  @EnsuresKeyForIf(result=true, expression="#1", map="this.map")
  public boolean hasValue(String key) { ... }
```

**Index**:
The `@SuppressWarnings` in guava that refer to this issue seem to be about non-conditional annotations rather than conditional ones, so they may not be relevant to this issue.

smillst:
A general way to add `@EnsuresSomeQualifier` annotation has been added.  (See PR #1723)
The implementation of `@EnsureLockHeldIf` could be updated to use this.

The Index Checker uses this feature, but the Regex and KeyFor check do not yet.

mernst:
Commit 3290705a41587c7e9974f6b04131d281c7130cc9 makes the Map Key Checker use this feature

PR #1723, opened 2017-12-27
Support qualifiers with arguments in contract meta-annotations

panacekcz:
As a partial fix to #1606, this PR extends the implementation of `PrecondtionAnnotation`, `PostconidtionAnnotation`, and `ConditionalPostconditionAnnotation`, to allow passing arguments from the contract annotation to the qualifier.

Not yet part of this PR:
- The code that standardizes the annotations is duplicated in several places. I suggest extracting the code related to parsing the expression
and standardizing the annotation into a utility method that would be used everywhere (doing roughly what `org.checkerframework.common.basetype.BaseTypeVisitor.resolveContracts` currently does) and also replacing `Pair<Receiver, AnnotationMirror>` by a class.

This PR extends a feature that is not completely implemented and documented. In particular, this PR does not fix:
- `ContractUtils` does not have proper error handling when the contract meta-annotations are used incorrectly (see TODO at the top of the file). This PR adds more ways in which that can happen.
- contract meta-annotations are not described in the manual (see docs/manual/todo.txt)

To be able to standardize the annotations, this PR moves the method `DependentTypesHelper.standardizeAnnotation` and makes it public.


panacekcz:
When #1726 is merged, the commented-out part of the test can be enabled.

panacekcz:
A nicer way of specifying the mapping of arguments would be a meta-annotation placed on the elements of the contract annotation.

For example:
```java
@PostconditionAnnotation(LTLengthOf.class)
public @interface EnsuresLTLengthOf {
  ...
  @QualifierArgument("value")
  String[] targetValue();
  @QualifierArgument
  String[] offset();
}
```

Would mean that `EnsuresLTLengthOf.targetValue` specifies values for `LTLengthOf.value` and  `EnsuresLTLengthOf.offset` specifies values for `LTLengthOf.offset`.

javacutil/src/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -180,6 +180,31 @@ public void copyElementValuesFromAnnotation(
...
+     * @param mapping mapping from source to target names
+     */
+    public void copyRenameElementValuesFromAnnotation(
+            AnnotationMirror valueHolder, Map<String, String> mapping) {
+
+        for (Map.Entry<? extends ExecutableElement, ? extends AnnotationValue> eltValToCopy :
+                valueHolder.getElementValues().entrySet()) {
+
+            String sourceName = eltValToCopy.getKey().getSimpleName().toString();
+            String targetName = mapping.get(sourceName.toString());
```
wmdietl: `sourceName` is already a String, so the extra `.toString()` call seems superfluous.

framework/src/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -452,6 +452,61 @@ protected String standardizeString(
     public boolean hasDependentAnnotations() {
         return !annoToElements.isEmpty();
     }
+    /**
+     * Standardizes an annotation. If the annotation is not an expression annotation, returns the
+     * same annotation unhanged.
```
wmdietl: unhanged -> unchanged.

framework/src/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -452,6 +452,61 @@ protected String standardizeString(
     public boolean hasDependentAnnotations() {
         return !annoToElements.isEmpty();
     }
+    /**
+     * Standardizes an annotation. If the annotation is not an expression annotation, returns the
```
wmdietl: Can you add an explanation of what `expression annotation` means, e.g. to the class doc? Isn't `dependent type annotation` the term you mean? What's the difference between the two?

checker/tests/index/CustomContractWithArgs.java
```diff
@@ -0,0 +1,170 @@
...
+        )
+        boolean ltlPost(int[] a, int c) {
+            if (b < a.length - c - 1 && b < a.length - 10) {
+                return true;
+            } else {
+                return false;
+            }
+        }
+
+        // TODO: this should be an error, because targetOffset refers to non-existent parameter
```
kelloggm: I agree that this should be an error. If you add an `@JavaExpression` annotation to the definition of `targetOffset`, does this still fail? So, the definition of `@EnsuresLTLIf` would be:

```java
@interface EnsuresLTLIf {
         public boolean result();
 
         public String[] expression();
 
         @JavaExpression
         public String[] targetValue();
 
         @JavaExpression
         public String[] targetOffset();
     }
```

(see the definition of `LTLengthOf`)

framework/src/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -221,19 +223,87 @@ public int hashCode() {
...
         }
     }
+    /**
+     * Returns the annotation mirror as specified by the "qualifier" value in {@code contractAnno}.
+     */
+    private AnnotationMirror getAnnotationMirrorOfContractAnnotation(
+            AnnotationMirror contractAnno) {
+        return getAnnotationMirrorOfQualifier(contractAnno, null, null);
+    }
+    /** Makes a map from source argument names to target argument names. */
```
kelloggm: I think the javadoc on this method should also mention how the lists don't need to be the same length, but do need to be correlated.

checker/tests/index/CustomContractWithArgs.java
kelloggm: Unfortunately, that's how errors that occur when parsing arguments to annotations are reported in general. This is a known issue with `FlowExpressionParseUtil`, so the "errors in the annotations" is actually the expected behavior. I'd add the `@JavaExpression` annotation, since it should be present, even if the behavior isn't exactly what you'd prefer.

panacekcz:
We decided that we want to get this feature working without unnecessary changes. The concerns listed in the first comment can be filed as issues and resolved later.

I am not quite sure about implementing https://github.com/typetools/checker-framework/pull/1723#issuecomment-354663345, because although we would prefer a simpler solution right now, changing it to a nicer solution later would break the public API, which we might want to avoid.

panacekcz:
@kelloggm, I have addressed your review comments.

panacekcz:
I want to implement `@QualifierArgument` in this PR.

panacekcz:
@wmdietl, I have addressed the comments and implemented `@QualifierArgument`. This PR is ready for review.

framework/src/org/checkerframework/framework/qual/PostconditionAnnotation.java
```diff
@@ -16,6 +16,12 @@
  * <p>The value {@code qualifier} that is necessary for a postcondition specified with {@link
  * EnsuresQualifier} is hard-coded here with the value {@code qualifier}.
  *
+ * <p>Additionally, the elements of the postcondition annotation (annotated by this meta-annotation)
+ * can be used to specify values of arguments of the qualifier. Each such element must be annotated
+ * by {@link QualifierArgument}, with a value specifying the name of the target qualifier argument.
+ * If no value is not specified, the name of the element is used. The element must have the same
```
mernst: "no value is not specified" looks like a typo.  Same in PreconditionAnnotation.java

framework/src/org/checkerframework/framework/qual/QualifierArgument.java
```diff
@@ -0,0 +1,24 @@
+package org.checkerframework.framework.qual;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * An annotation to use on an element of a contract annotation to specify that the element specifies
```
mernst: Please define "contract annotation".
The text is hard to read, such as "specify that the element specifies".  There are a lot of different parts and it's hard for a reader to keep them clear.

Please also add an example of use.  That will make the intention clearer.

framework/src/org/checkerframework/framework/qual/PostconditionAnnotation.java
```diff
@@ -16,6 +16,12 @@
  * <p>The value {@code qualifier} that is necessary for a postcondition specified with {@link
  * EnsuresQualifier} is hard-coded here with the value {@code qualifier}.
  *
+ * <p>Additionally, the elements of the postcondition annotation (annotated by this meta-annotation)
```
mernst: Please give an example of use to help make the ideas more concrete.
Likewise in PreconditionAnnotation.java.

framework/src/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -221,19 +225,96 @@ public int hashCode() {
...
+     *
+     * @param qualifierAnno annotation specifying the qualifier class
+     * @param argumentAnno annotation containing the argument values
+     * @param argumentMap map of argument names, which maps from names in {@code argumentAnno} to
+     *     names used in the returned annotation
+     */
+    private AnnotationMirror getAnnotationMirrorOfQualifier(
+            AnnotationMirror qualifierAnno,
+            AnnotationMirror argumentAnno,
+            Map<String, String> argumentMap) {
```
mernst: I suggest naming variables not according to their type (such as `argumentMap`), but according to their semantics or their use.  For example, this could be called `argumentRenaming` or something similar.

framework/src/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -221,19 +225,96 @@ public int hashCode() {
         return result;
     }
 
-    /** Returns the annotation mirror as specified by the "qualifier" value in metaAnno. */
-    private AnnotationMirror getAnnotationMirrorOfQualifier(AnnotationMirror metaAnno) {
+    /**
+     * Returns the annotation mirror as specified by the "qualifier" value in {@code qualifierAnno}.
+     * If {@code argumentAnno} is specified, then arguments are copied from {@code argumentAnno} to
```
mernst: Does this mean that `argumentAnno` is allowed to be null?  If so, please state that in the `@param` comment.

framework/src/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -221,19 +225,96 @@ public int hashCode() {
...
     }
+    /**
+     * Returns the annotation mirror as specified by the "qualifier" value in {@code contractAnno}.
+     */
+    private AnnotationMirror getAnnotationMirrorOfContractAnnotation(
+            AnnotationMirror contractAnno) {
+        return getAnnotationMirrorOfQualifier(contractAnno, null, null);
+    }
+    /**
+     * Makes a map from source element names to target element names. Each element of {@code
```
mernst: "source" and "target" are not defined in this comment.

framework/src/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -221,19 +225,96 @@ public int hashCode() {
         return result;
     }
 
-    /** Returns the annotation mirror as specified by the "qualifier" value in metaAnno. */
-    private AnnotationMirror getAnnotationMirrorOfQualifier(AnnotationMirror metaAnno) {
+    /**
+     * Returns the annotation mirror as specified by the "qualifier" value in {@code qualifierAnno}.
```
mernst: Please clarify what is the "qualifier" value.  Is that the value of a field named `qualifier`, or something else?

mernst: Is this a helper function that should only be called in the two places it is currently called?  If so, I would document that.

framework/src/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -452,6 +452,61 @@ protected String standardizeString(
     public boolean hasDependentAnnotations() {
         return !annoToElements.isEmpty();
     }
+    /**
+     * Standardizes an annotation. If the annotation is not a dependent type annotation, returns the
```
mernst: What does it mean to standardize an annotation?

javacutil/src/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -180,6 +180,30 @@ public void copyElementValuesFromAnnotation(
...
+     * Copies the specified element values from the given annotation, using the specified name
+     * mapping. If an element specified as a target in the mapping doesn't exist in the annotation
+     * to be built, an error is raised. If an element from the given annotation is not in the
+     * mapping, it is ignored.
+     *
+     * @param valueHolder the annotation that holds the values to be copied
+     * @param mapping mapping from source to target names
+     */
+    public void copyRenameElementValuesFromAnnotation(
+            AnnotationMirror valueHolder, Map<String, String> mapping) {
```
mernst: Don't use variable names like `mapping`.  Even `nameMapping` would be more informative.

javacutil/src/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -180,6 +180,30 @@ public void copyElementValuesFromAnnotation(
...
     }
 
+    /**
+     * Copies the specified element values from the given annotation, using the specified name
+     * mapping. If an element specified as a target in the mapping doesn't exist in the annotation
+     * to be built, an error is raised. If an element from the given annotation is not in the
+     * mapping, it is ignored.
+     *
+     * @param valueHolder the annotation that holds the values to be copied
+     * @param mapping mapping from source to target names
```
mernst: The documentation doesn't say that these are element names.

javacutil/src/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -180,6 +180,30 @@ public void copyElementValuesFromAnnotation(
         return;
     }
 
+    /**
+     * Copies the specified element values from the given annotation, using the specified name
+     * mapping. If an element specified as a target in the mapping doesn't exist in the annotation
+     * to be built, an error is raised. If an element from the given annotation is not in the
```
mernst: Rather than saying "an error is raised", I would state the requirement:  each value in the mapping must be an element name in the annotation being built.  (But isn't that obvious?)

javacutil/src/org/checkerframework/javacutil/AnnotationBuilder.java
mernst: `valueHolder` is the source, and `this` is the target.  I would indicate this more explicitly in the documentation or naming.

panacekcz:
@mernst, I have addressed your commments.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -637,6 +638,14 @@ private void checkContractsAtMethodDeclaration(
                         FlowExpressionContext.buildContextForMethodDeclaration(
                                 node, getCurrentPath(), checker.getContext());
             }
+
+            DependentTypesHelper dependentTypesHelper = atypeFactory.getDependentTypesHelper();
```
mernst: This block of code (the next 6 lines) happens quite a few times in this and other files.  Consider abstracting it out, so that it's a single method call in all the places it occurs.

panacekcz:
@mernst, I have slightly changed the documentation to say that only values of elements annotated by `@QualifierArgument`  are copied.

mernst:
@panacekcz , thanks for clarifying the part of the specification I did not understand.
I suggest that you still do the refactoring that I mentioned, and that is also mentioned in your initial comment for this code review.  I think the refactoring will be easy to do, the change will make the code review easier to do, and we might as well get the code right in the initial commit.
@panacekcz and @kelloggm , Martin's review still shows as not resolved.  Are there still changes that need to be made?

panacekcz:
I have addressed Martins comments.
I am going to do the refactoring to reduce code duplication.

panacekcz:
@mernst, I see your comment is only about extracting the new code. I have done that now. I originally thought about a bigger change, because the surrounding existing code is also duplicated (parsing the expression to be annotated, etc.). So I am not doing that.

mernst:
@panacekcz , I see.  That other change might be worthwhile, but I agree it's not part of this code review.

mernst:
@kelloggm , Your pull request is marked as "requested changes".  Is there anything remaining?  If not, can you either re-review or approve?  Thanks!

framework/src/org/checkerframework/framework/flow/CFAbstractTransfer.java
```diff
@@ -518,24 +518,38 @@ protected void addInformationFromPreconditions(
...
     }
 
+    /** Standardize a type qualifier annotation obtained from a contract. */
+    private AnnotationMirror standardizeAnnotationFromContract(
+            AnnotationMirror annoFromContract,
+            FlowExpressionContext flowExprContext,
+            TreePath path) {
+        if (analysis.dependentTypesHelper != null) {
+            return analysis.dependentTypesHelper.standardizeAnnotation(
+                    flowExprContext, path, annoFromContract, false);
```
wmdietl: Class `BaseTypeVisitor` also has a `standardizeAnnotationFromContract` method, which looks quite similar.
That other method also calls `dependentTypesHelper.checkAnnotation`.
Is such a call also necessary here? Would a comment help explain why or why not?
Cross-references between the two methods would make finding this similar logic easier.

It would be even nicer if that common logic were only implemented once.

PR #1728, opened 2017-12-31
Fully qualify javadoc links to remove import.

jschaf:
The reason is so I can build a Qualifier only JAR using a regex like "**/qual/*.java".  I'm using Bazel to build from source and Guava requires the checker framework annotations.


wmdietl:
Thanks for the contribution! I'm not quite sure I understand your motivation, but inlining these class names seems good to me.
Note that the formatting isn't quite right:
https://travis-ci.org/typetools/checker-framework/jobs/323522135#L7082

Running `ant reformat` should fix this for you. It's strange that the pre-commit hook didn't prevent you from doing this. Did you force commit?

jschaf:
I tried `ant reformat` and it wanted to reformat the world.  I ran google-java-format on the modified files instead which I think should do the trick.

For my fork, I did a normal git push.

PR #1742, opened 2018-01-04
Fix #1685: don't skip member references and lambda expressions.

### LINKED ISSUE #1685: Consider supporting lambdas in BaseTypeVisitor.commonAssignmentCheck
The method `BaseTypeVisitor.commonAssignmentCheck(AnnotatedTypeMirror, ExpressionTree, String)` currently skips checking the type of lambda expressions. This is used during normal variable/field assignments, as well as when checking the subtyping of method arguments. The underlying reason is that `BaseTypeVisitor.shouldSkipUses(...)` explicitly skips checking of lambdas and method references. Is there still a good reason for this?

In PR #1481 , we implemented our own version of this method `lambdaAssignmentCheck` and didn't run into any significant problems because of that. This methods calls directly into the four parameter version of `commonAssignmentCheck(AnnotatedTypeMirror, AnnotatedTypeMirror, Tree, String)`, skipping the skip check. Note that we might not be calling it in all cases in which the 3 parameter `commonAssignmentCheck` is usually called.

wmdietl:
Skipping them doesn't seem necessary any longer.


wmdietl:
@smillst: I've looked at some of the failures, for example on Issue1027.java and they all seem related to #979 .
Instead of removing skipping member references and lambda trees, how about adding a reference to #979 in that location and once #979 is fixed, we remove that code?

smillst:
I actually think this code is correct.  Here's why:

Because lambdas and member references are poly expressions, their types depend on context in which they appear.  So instead of checking that the assignment is valid, we check if the lambda or method reference is valid in the context in which it appears.  (This happens in BaseTypeVisitor#visitLambdaExpression and BaseTypeVisitor#visitMemberReference.)

So, if that makes sense, we should add that as comment here.

wmdietl:
`shouldSkipUses` seems like the wrong place for this.
Once the type of the poly expression is determined, in the context in which it is used, the resulting type should trivially satisfy the constraints of that context.
So there shouldn't be a need to skip checking uses of lambdas and member references. If anything, that would be a performance optimization, in an awkward place.

Do you understand why there are additional errors raised without skipping lambdas/member refs?

wmdietl:
I've updated this PR with such a comment.

PR #1744, opened 2018-01-04
Fix #1232: count annotations on the attributed tree.

### LINKED ISSUE #1232: AnnotationsCounter crashes on array constructor annotation
Running `AnnotationsCounter` on the following program causes it to crash.

```java
import org.checkerframework.framework.qual.PolyAll;

public final class ArraysMDE2 {
  void subarray(int length) {
    Object[] result = new @PolyAll Object[length];
  }
}
```

The problem is that `anno.annotationType.type` is null, so the line
```java
Name annoName = anno.annotationType.type.tsym.name;
```
throws a NullPointerException

Here is a concrete command line to reproduce the problem.

$ch/bin-devel/javac -processor org.checkerframework.common.util.count.AnnotationsCounter ArraysMDE2.java

More concretely, I was trying to run it on plume-lib.  The command, from plume-lib's `java/` folder, was:

```$ch/bin-devel/javac -proc:only -cp plume.jar:lib/junit-4.12.jar -processor org.checkerframework.common.util.count.AnnotationsCounter src/plume/*.java```

### PR DESCRIPTION
**Pull request by wmdietl**

### CODE REVIEW

File: framework/src/org/checkerframework/common/util/count/AnnotationsCounter.java [hunk 1]
```diff
@@ -26,34 +22,31 @@
...
 @SupportedAnnotationTypes("*")
 @SupportedSourceVersion(SourceVersion.RELEASE_8)
-public class AnnotationsCounter extends AbstractProcessor {
+public class AnnotationsCounter extends AbstractTypeProcessor {
     final Map<Name, Integer> annotationCount = new HashMap<Name, Integer>();
 
     protected void incrementCount(Name annoName) {
         if (!annotationCount.containsKey(annoName)) {
-            annotationCount.put(annoName, 2);
+            annotationCount.put(annoName, 1);
```
wmdietl: I didn't see a reason why counting should start at 2. With the small examples I tried 1 made more sense.

framework/src/org/checkerframework/common/util/count/AnnotationsCounter.java
```diff
@@ -62,7 +55,7 @@ public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment
                 @Override
                 public Void visitAnnotation(AnnotationTree node, Void p) {
                     JCAnnotation anno = (JCAnnotation) node;
-                    Name annoName = anno.annotationType.type.tsym.name;
+                    Name annoName = anno.annotationType.type.tsym.getQualifiedName();
```
wmdietl: This now uses the fully-qualified name, to allow keeping track of e.g. Nullness annotations in different packages.

PR #1750, opened 2018-01-05
Link to NullAway in the dataflow manual

msridhar:
Also some slight grammatical cleanup


mernst:
Thanks for the update, Manu.
If there is someplace the Checker Framework manual should reference NullAway, let us know that too.

PR #1752, opened 2018-01-07
Handle concurrent commits

mernst:


.travis.yml
```diff
@@ -46,16 +47,18 @@ script:
...
 - echo "TRAVIS_REPO_SLUG=${TRAVIS_REPO_SLUG}"
 - >-
-  docker run mdernst/ubuntu-for-cf-$JDKVER /bin/bash -c "true &&
-     export JAVA_HOME=\`which javac|xargs readlink -f|xargs dirname|xargs dirname\` &&
+  docker run mdernst/$OS-for-cf-$JDKVER /bin/bash -c "true &&
      if [ $TRAVIS_EVENT_TYPE = pull_request ] ; then
        git clone --quiet --depth 9 $REMOTE_ORIGIN_URL $THIS_REPO || git clone --quiet --depth 9 $REMOTE_ORIGIN_URL $THIS_REPO
        cd $THIS_REPO
        git fetch origin +refs/pull/$TRAVIS_PULL_REQUEST/merge
+       git checkout -qf $TRAVIS_PULL_REQUEST_SHA
```
wmdietl: I don't understand what the two consecutive `git checkout` do. Doesn't the second one cancel out the first?
Can you add comments what the effect of this is?

mernst: Because of the `||` connective, the second command is attempted only if the first one fails.  Thus, the job gets two tries to perform the checkout, in case of network lossage.

mernst: I added a comment to this effect.

wmdietl: There is no `||` connective for the `git checkout` between this line and the next one. One checks out `$TRAVIS_PULL_REQUEST_SHA` and the next one checks out `FETCH_HEAD`, so this doesn't look like a retry.
I think clarifying what these two checkouts do would be good. The repeated git commands with `||` were clear, but having it documented is definitely good.

mernst: You are right.  That was a stray line.  Thanks.

PR #1757, opened 2018-01-11
Update error-prone version.

wmdietl:


dataflow/src/org/checkerframework/dataflow/analysis/FlowExpressions.java
```diff
@@ -1088,13 +1088,18 @@ public ArrayCreation(TypeMirror type, List<Node> dimensions, List<Node> initiali
             return initializers;
         }
 
+        @SuppressWarnings("EqualsIncompatibleType")
```
wmdietl: @smillst This seems like a real problem: classes `Node` and `Receiver` are in no relationship, so the two loops in the body will never return true.
Do you know what the intended behavior here is?

smillst: dimensions and initializers should be FlowExpressions.Receivers.  I'll fix it in a separate pull request.

smillst: See PR #1758

PR #1764, opened 2018-01-16
Add index annotations to CharSequence

panacekcz:
Previously, `java.lang.CharSequence` did not have upper-bound and samelen annotations on the return type. That prevented geting `@IndexFor` or `@SameLen` for a mutable sequence (`StringBuilder`), but caused false positives in code using `CharSequence`.

This PR adds the annotations.


panacekcz:
Travis failure is fixed by mernst/plume-lib#40. Unfortunately this adds 2 new `@SuppressWarnigns` to plume-lib.

panacekcz:
This should be merged at the same time as or after mernst/plume-lib#40.
If it is merged before, then type-checking plume-lib with the Index Checker will fail before the other one is merged.

smillst:
I merged mernst/plume-lib#40.  So I'll restart these tests then merge this pull request.

mernst:
Weird merge conflict.  I'll look into this shortly (but not immediately).

mernst:
I think Suzanne resolved this before I did, and I was confused why it wasn't getting Suzanne's commit.  Will wait for test to pass.

smillst:
I just merged master into this branch to see if that would fix the problem.  Hopefully it will pas now.

PR #1766, opened 2018-01-16
Warn if declaration annotation is interpreted as an array component type qualifier

smillst:
The warning is issued once per annotation that is used as a component type qualifier.

An example of the new warning:
```
javacheck -processor nullness -cp $CHECKERFRAMEWORK/checker/dist/checker.jar  $CHECKERFRAMEWORK/checker/tests/nullness/Issue1667.java
warning: Interpreting @com.sun.istack.internal.NotNull as a type annotation on an array component type.
warning: Interpreting @javax.annotation.CheckForNull as a type annotation on an array component type.
... (Nullness Checker warnings.) ...
2 warnings
```
This pull request also adds `warn` and `warnOnce` methods to `ErrorReporter`.

### DIFF
--- a/checker/tests/nullness/Issue1667.java
+++ b/checker/tests/nullness/Issue1667.java
@@ -1,0 +1,19 @@
+import javax.annotation.CheckForNull;
+
+class Issue1667 {
+    @com.sun.istack.internal.NotNull Object[] foo(Object[] bar) {
+        // :: error: (return.type.incompatible)
+        return null;
+    }
+
+    @CheckForNull
+    Object[] foo2(@CheckForNull Object[] bar) {
+        // :: error: (return.type.incompatible)
+        return null;
+    }
+
+    void test() {
+        // :: error: (argument.type.incompatible)
+        foo2(null);
+    }
+}
--- a/checker/tests/nullness/Issue1667.java
+++ b/checker/tests/nullness/Issue1667.java
@@ -1,3 +1,5 @@
+// warning: Interpreting @com.sun.istack.internal.NotNull as a type annotation on an array component type.
+// warning: Interpreting @javax.annotation.CheckForNull as a type annotation on an array component type.
 import javax.annotation.CheckForNull;
 
 class Issue1667 {
--- a/framework/src/org/checkerframework/framework/source/SourceChecker.java
+++ b/framework/src/org/checkerframework/framework/source/SourceChecker.java
@@ -698,6 +698,21 @@ public abstract class SourceChecker extends AbstractTypeProcessor
         return swString.split(",");
     }
 
+    @Override
+    public void warn(String format, Object... args) {
+        String msg = String.format(format, args);
+        this.messager.printMessage(Kind.MANDATORY_WARNING, msg);
+    }
+
+    private Set<String> warnedBefore = new HashSet<>();
+
+    @Override
+    public void warnOnce(String format, Object... args) {
+        if (warnedBefore.add(String.format(format, args))) {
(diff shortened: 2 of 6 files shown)


framework/src/org/checkerframework/framework/util/element/ElementAnnotationUtil.java
```diff
@@ -87,7 +87,17 @@ public static void applyAllElementAnnotations(
     static void addAnnotationsFromElement(
             final AnnotatedTypeMirror type, final List<? extends AnnotationMirror> annotations) {
         AnnotatedTypeMirror innerType = AnnotatedTypes.innerMostType(type);
-        innerType.addAnnotations(annotations);
+        if (innerType != type) {
+            for (AnnotationMirror annotation : annotations) {
+                if (innerType.addAnnotation(annotation)) {
+                    ErrorReporter.warnOnce(
```
smillst: @mernst I would appreciate feed back on this warning message.

framework/src/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -698,6 +698,21 @@ private Pattern getOnlyDefsPattern(Map<String, String> options) {
         return swString.split(",");
     }
 
+    @Override
+    public void warn(String format, Object... args) {
+        String msg = String.format(format, args);
+        this.messager.printMessage(Kind.MANDATORY_WARNING, msg);
```
smillst: @wmdietl  These warnings need to be suppresable.  Any ideas on how best to implement that?

framework/src/org/checkerframework/framework/util/element/ElementAnnotationUtil.java
mernst: Looks good to me.

framework/src/org/checkerframework/framework/util/element/ElementAnnotationUtil.java
```diff
@@ -87,7 +87,17 @@ public static void applyAllElementAnnotations(
...
+                            "Interpreting %s as a type annotation on an array component type.",
+                            annotation);
```
wmdietl: You should be able to pass in the `Element` for the corresponding declaration to get a warning on that element, instead of a warning without line number.

checker/tests/nullness/Issue1667.java
```diff
@@ -0,0 +1,21 @@
+// warning: Interpreting @com.sun.istack.internal.NotNull as a type annotation on an array component type.
```
wmdietl: It would be much more user-friendly if these messages were attached to the first occurrence of the warning. That is, on lines 6 and 11 in this file instead of here without line numbers.

framework/src/org/checkerframework/framework/source/SourceChecker.java
wmdietl: I think if you attach the message to an `Element` as source, you can use the normal `@SuppressWarnings` to achieve this.

checker/tests/nullness/Issue1667.java
smillst: The "first" occurrence might change depending on the order in which files are compiled or which files are compiled at once.  If the warning is suppressed relative to code location, the warning would not always be suppressed.

I'm thinking about making it suppressible using `-AsuppressWarnings`.

mernst:
This pull request should be changed to reinterpret the location of the declaration annotation, rather than just issuing a warning.

PR #1767, opened 2018-01-17
Remove org.checkerframework.framework.qual.Dependent

smillst:
Only the Nullness Checker supported this annotation.  It's not well documented or well implemented.  (One of the two tests using this annotation is skipped.)

It is implemented using the `GeneralAnnotatedTypeFactory` which supports storing
annotations that are not qualifiers in AnnotatedTypeMirrors.  It's causing the test failures in #1766.  I moved it `org.checkerframework.common.util.debug` because `org.checkerframework.common.util.debug.TypeOutputtingChecker` still uses it.

~~The checker-framework-inference tests fail because the Nullness Inference Checker still uses the GeneralAnnotatedTypeFactory.  I'll make a pull request to remove it.~~


wmdietl:
Yes, I'm very much for removing this! Last time @mernst objected to removing this functionality.

I think the use of `GeneralAnnotatedTypeFactory` in `TypeOutputtingChecker` should also be removable: the only annotations that the `GeneralATF` supports are explicit in the source, so normal javac mechanisms to output the annotations should be enough.

framework/src/org/checkerframework/common/util/debug/GeneralAnnotatedTypeFactory.java
```diff
@@ -1,4 +1,4 @@
-package org.checkerframework.framework.type;
+package org.checkerframework.common.util.debug;
```
wmdietl: I would try completely removing this class, to make sure no new uses are added.

smillst: I don't want to rewrite `TypeOutputtingChecker` right now,  but I agree that we really don't want any new uses of `GeneralAnnotatedTypeFactory`.  I made `GeneralAnnotatedTypeFactory` a static inner class of `TypeOutputtingChecker`, so that should prevent new uses.

wmdietl: That sounds like a good compromise. I would just add a comment that these two inner classes should be replaced by a normal javac implementation, to make it clear that these are historic artifacts, not interesting logic.

mernst:
I'm in favor.  Thanks!

PR #1772, opened 2018-01-19
Expand support for custom collections.

smillst:
Treat a method annotated with @Length("this") the same as `String#length`.

This fixes kelloggm#194.

### DIFF
--- a/checker/src/org/checkerframework/checker/index/IndexMethodIdentifier.java
+++ b/checker/src/org/checkerframework/checker/index/IndexMethodIdentifier.java
@@ -1,12 +1,18 @@
 package org.checkerframework.checker.index;
 
+import com.sun.source.tree.MethodInvocationTree;
 import com.sun.source.tree.Tree;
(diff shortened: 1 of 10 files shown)


checker/src/org/checkerframework/checker/index/IndexMethodIdentifier.java
```diff
@@ -65,18 +77,49 @@ public boolean isRandomNextInt(Tree tree, ProcessingEnvironment processingEnv) {
...
+            return true;
+        }
+
+        AnnotationMirror len = factory.getDeclAnnotation(ele, LengthOf.class);
+        if (len == null) {
+            return false;
+        }
+        List<String> values =
+                AnnotationUtils.getElementValueArray(len, "value", String.class, false);
+        if (values.size() == 1) {
```
kelloggm: Is there a reason not to just loop over all of the values and check if any of them are `this`? I doubt this will come up much in practice, but it would be nice not to require `this` to be the only argument to the `@LengthOf` annotation.

PR #1779, opened 2018-01-22
Clean up annotation counting annotation processors

### LINKED ISSUE #1769: AnnotationCounter vs Locations; also -AskipDefs and -AonlyDefs
Is AnnotationsCounter.java obsolete because its functionality is subsumed by Locations.java?

Also, it would be nice if the program (or programs) supported -AskipDefs and -AonlyDefs, to enable passing a command line more like that which is issued to perform type-checking.

Count the number of types that are generic. (Including diamonds.)
Count the number of array accesses and constructions. (array[i][0] is two accesses.) 
#

smillst:
* Merge Locations and AnnotationsCounter into a new class is called AnnotationStatistics.
* Add support for -AskipDefs and -AonlyDefs
* Add a new checker to print:
    * The number of types that are generic. (Including diamonds.)
    * The number of array accesses and constructions. (array[i][0] is two accesses.)

Fixes #1769.


framework/src/org/checkerframework/common/util/count/TypeStatistics.java
```diff
@@ -0,0 +1,79 @@
...
+                // recursively call visitors on subtrees; we want to skip the
+                // class entirely.
+                return null;
+            }
+            return super.visitClass(tree, p);
+        }
+
+        @Override
+        public Void visitNewArray(NewArrayTree node, Void aVoid) {
+            arrayAccesses += node.getDimensions().size();
```
smillst: This doesn't count arrays created using initializers.  Is that what we want?

framework/src/org/checkerframework/common/util/count/AnnotationStatistics.java
```diff
@@ -54,22 +57,44 @@
...
+            System.out.println("No annotations found.");
+        } else {
+            System.out.println("Found annotations: ");
+            for (Map.Entry<Name, Integer> entry : annotationCount.entrySet()) {
+                System.out.println(entry.getKey() + "\t" + entry.getValue());
+            }
+        }
+    }
+
+    protected void incrementCount(Name annoName) {
```
mernst: I would write at least a one-line Javadoc comment.

framework/src/org/checkerframework/common/util/count/TypeStatistics.java
mernst: I think that's OK.  We really want to count the number of indexes, and the argument i to `new T[i]` is index-like but there is no index-like thing in `a = {0, 1, 2}`.  We should document what the class counts, though.

framework/src/org/checkerframework/common/util/count/TypeStatistics.java
```diff
@@ -0,0 +1,79 @@
...
+import com.sun.source.tree.NewArrayTree;
+import com.sun.source.tree.ParameterizedTypeTree;
+import javax.annotation.processing.SupportedSourceVersion;
+import javax.lang.model.SourceVersion;
+import org.checkerframework.framework.source.SourceChecker;
+import org.checkerframework.framework.source.SourceVisitor;
+import org.checkerframework.javacutil.AnnotationProvider;
+
+/**
+ * An annotation processor for listing the facts about types and array accesses locations of
```
mernst: Document what this does -- what "the facts" are.
Generic types feel rather different than array accesses.  I would put each into its own class, or jam everything into AnnotationStatistics with different command-line arguments.  (Maybe this class is the kitchen-sink class for everything except annotations, though?  In that case, the "Type" in its name is a bit misleading.  Array access locations aren't really related to types.)

framework/src/org/checkerframework/common/util/count/TypeStatistics.java
```diff
@@ -0,0 +1,79 @@
...
+ * annotations. To invoke it, use
+ *
+ * <pre>
+ * javac -proc:only -processor org.checkerframework.common.util.count.TypeStatistics <em>MyFile.java ...</em>
+ * </pre>
+ */
+@SupportedSourceVersion(SourceVersion.RELEASE_8)
+public class TypeStatistics extends SourceChecker {
+
+    int generics = 0;
```
mernst: Helpful to document what is being counted.  (Less critical, but still useful here, once there is class javadoc.)

PR #1782, opened 2018-01-23
Work around some restrictions imposed by the lgtm build environment

aibaars:
* The parent folder of the source checkout is not writeable in the lgtm build environment
* The lgtm build environment limits network access and permits only HTTP GET operations; as
  a result git/hg clone commands fail.

To workaround these restrictions the dependencies are downloaded in the form of a source zip, and
stored in the $LGTM_WORKSPACE folder

@wmdietl

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -5,7 +5,21 @@
 extraction:
   java:
     before_index:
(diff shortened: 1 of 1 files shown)


wmdietl:
Thanks a lot for these improvements! Let's see what lgtm finds now.

PR #1784, opened 2018-01-23
(WIP) Linear type checker

renatoathaydes:
This is work in progress.

Please let me know if I am going in the correct direction. I found it very difficult to understand how the flow framework and the checker visitor work. This is my attempt so far.

I've been writing/testing this code in a [separate repo](https://github.com/renatoathaydes/linear-type-checker) because the build in this repo is impossible to use for development...

To run just the tests for my checker takes almost 30 seconds! It seems to recompile the whole thing even when I just change one class (or don't change anything at all).

Also, the tests are failing, even though in my own repo the (simialr) tests pass when I enable my checker. I can't see why that would be the case!

Finally, there's a git hook stopping me from committing due to code formatting issues.... I've used the default Java format that seems to be used in all other files in this repo... can you explain how I can re-format my code to pass the git hook? Running `ant reformat` doesn't work.

Thanks.

### DIFF
--- a/checker/src/org/checkerframework/checker/linear/LinearAnnotatedTypeFactory.java
+++ b/checker/src/org/checkerframework/checker/linear/LinearAnnotatedTypeFactory.java
@@ -1,93 +1,85 @@
 package org.checkerframework.checker.linear;
 
-import javax.lang.model.element.AnnotationMirror;
-import javax.lang.model.element.Element;
 import org.checkerframework.checker.linear.qual.Linear;
 import org.checkerframework.checker.linear.qual.Unusable;
+import org.checkerframework.checker.linear.qual.Normal;
+import com.sun.source.tree.Tree;
+import org.checkerframework.checker.nullness.qual.Nullable;
 import org.checkerframework.common.basetype.BaseAnnotatedTypeFactory;
 import org.checkerframework.common.basetype.BaseTypeChecker;
+import org.checkerframework.dataflow.analysis.RegularTransferResult;
+import org.checkerframework.dataflow.analysis.TransferInput;
+import org.checkerframework.dataflow.analysis.TransferResult;
+import org.checkerframework.dataflow.cfg.node.MethodInvocationNode;
+import org.checkerframework.framework.flow.CFAbstractAnalysis;
+import org.checkerframework.framework.flow.CFStore;
+import org.checkerframework.framework.flow.CFTransfer;
+import org.checkerframework.framework.flow.CFValue;
+import org.checkerframework.framework.type.AnnotatedTypeFactory;
 import org.checkerframework.framework.type.AnnotatedTypeMirror;
 import org.checkerframework.javacutil.AnnotationBuilder;
 
+import javax.lang.model.element.AnnotationMirror;
+import java.lang.annotation.Annotation;
+import java.util.Set;
+
 /**
- * Adds {@link Unusable} qualifier to a type if it represents:
- *
+ * Marks a {@link Linear} variable as {@link Unusable} once the variable is:
+ * <p>
  * <ol>
- *   <li value="1">Class declaration tree/element. Such a construct usually requires the top
- *       qualifier.
- *   <li value="2">{@code Linear} reference once it is "used up"
+ * <li value="1">the receiver of a method call</li>
+ * <li value="2">TODO</li>
  * </ol>
  */
 public class LinearAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
 
-    private final AnnotationMirror LINEAR, UNUSABLE;
+    final AnnotationMirror LINEAR, UNUSABLE, NORMAL;
 
     public LinearAnnotatedTypeFactory(BaseTypeChecker checker) {
         super(checker);
 
         LINEAR = AnnotationBuilder.fromClass(elements, Linear.class);
         UNUSABLE = AnnotationBuilder.fromClass(elements, Unusable.class);
+        NORMAL = AnnotationBuilder.fromClass(elements, Normal.class);
 
         this.postInit();
     }
 
-    /** Case 1: type of class declaration */
     @Override
-    public void addComputedTypeAnnotations(Element elt, AnnotatedTypeMirror type) {
-        if (!type.isAnnotatedInHierarchy(LINEAR) && elt.getKind().isClass()) {
-            type.addAnnotation(UNUSABLE);
-        }
-        super.addComputedTypeAnnotations(elt, type);
+    protected Set<Class<? extends Annotation>> createSupportedTypeQualifiers() {
+        return getBundledTypeQualifiersWithPolyAll(
+                Linear.class, Normal.class, Unusable.class);
     }
 
-    // TODO: Re-enable flow with the new org.checkerframework.dataflow framework.
+    @Override
+    public CFTransfer createFlowTransferFunction(CFAbstractAnalysis<CFValue, CFStore, CFTransfer> analysis) {
+        return new LinearFlow(analysis);
+    }
 
-    /**
-     * Performs flow-sensitive analysis to mark reference types {@code Linear} as {@code Unusable}
-     * once they are used up.
-     *
-     * <p>A {code Linear} type is "used up" once the reference is mentioned, as an {@link
-     * IdentifierTree}.
-     */
-    /*
-    private class LinearFlow extends DefaultFlow<DefaultFlowState> {
-        public LinearFlow(BaseTypeChecker checker, CompilationUnitTree root,
-                Set<AnnotationMirror> annotations, AnnotatedTypeFactory factory) {
-            super(checker, root, annotations, factory);
+    private class LinearFlow extends CFTransfer {
+        private final AnnotatedTypeFactory factory;
+
+        LinearFlow(CFAbstractAnalysis<CFValue, CFStore, CFTransfer> analysis) {
+            super(analysis);
+            factory = analysis.getTypeFactory();
         }
 
-        /**
-         * Case 2: add {@code Unusable} to node type, if it is {@code Linear}.
-         *
         @Override
-        public Void visitIdentifier(IdentifierTree node, Void p) {
-            super.visitIdentifier(node, p);
-            markAsUnusableIfLinear(node);
-            return null;
-        }
+        public TransferResult<CFValue, CFStore> visitMethodInvocation(MethodInvocationNode node,
+                                                                      TransferInput<CFValue, CFStore> in) {
+            TransferResult<CFValue, CFStore> result = super.visitMethodInvocation(node, in);
 
-        /**
-         * If the node is of type {@code Linear}, then transit its type
-         * into an {@code Unusable} type.
-         *
-         * The method should be called on every instance of a tree
-         * that causes the reference to be "used up".
-         *
-        private void markAsUnusableIfLinear(ExpressionTree node) {
-            if (!LinearVisitor.isLocalVarOrParam(node)) {
-                return;
-                }
-
-            Element elem = TreeUtils.elementFromUse(node);
-            assert elem != null;
-            if (this.flowState.vars.contains(elem)) {
-                int idx = this.flowState.vars.indexOf(elem);
-                if (this.flowState.annos.get(LINEAR, idx)) {
-                    this.flowState.annos.set(UNUSABLE, idx);
-                    this.flowState.annos.clear(LINEAR, idx);
+            @Nullable Tree receiverTree = node.getTarget().getReceiver().getTree();
+            if (receiverTree != null) {
+                AnnotatedTypeMirror type = factory.getAnnotatedType(receiverTree);
+                if (type.isAnnotatedInHierarchy(LINEAR)) {
+                    return new RegularTransferResult<>(analysis.createSingleAnnotationValue(UNUSABLE,
+                            result.getResultValue().getUnderlyingType()), in.getRegularStore());
                 }
             }
+
+            return result;
         }
     }
-    */
+
 }
--- a/checker/src/org/checkerframework/checker/linear/LinearAnnotatedTypeFactory.java
+++ b/checker/src/org/checkerframework/checker/linear/LinearAnnotatedTypeFactory.java
@@ -1,9 +1,12 @@
 package org.checkerframework.checker.linear;
 
-import org.checkerframework.checker.linear.qual.Linear;
-import org.checkerframework.checker.linear.qual.Unusable;
-import org.checkerframework.checker.linear.qual.Normal;
 import com.sun.source.tree.Tree;
+import java.lang.annotation.Annotation;
+import java.util.Set;
+import javax.lang.model.element.AnnotationMirror;
+import org.checkerframework.checker.linear.qual.Linear;
+import org.checkerframework.checker.linear.qual.Normal;
+import org.checkerframework.checker.linear.qual.Unusable;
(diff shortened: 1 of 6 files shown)


renatoathaydes:
The test I am running in my project can be seen [here](https://github.com/renatoathaydes/linear-type-checker/blob/master/src/test/java/com/athaydes/checkerframework/linear/LinearTypeCheckerTest.java) and it passes.

smillst:
If you merge typetools/master into your branch, the merge problem in the Travis test will go away.  (Also, if you give us permission to push to your branch, we can fix trivial problems like this for you.  See https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/#enabling-repository-maintainer-permissions-on-existing-pull-requests)

mernst:
Can you be more specific than "Running ant reformat doesn't work."?  We'd like to help you with that issue, but we don't have enough information to do so.

renatoathaydes:
@mernst doesn't work as in doesn't exist in the `checker-frameworks/checker` ant build file. Can you give precise instructions on how to make it work to format a single file (that's what I was actually asking)?

@smillst I opted to not give access because the PR is not ready and I want to know exactly what the steps required for things to work are... so I don't depend on someone else to fix simple things.

mernst:
Run `ant reformat` from the top level -- the `checker-framework` directory.

mernst:
Or pull from master -- commit fbfe23f924686477ae5ce3ee315d2941482683d6 makes `ant reformat` work from the `checker` directory too.

renatoathaydes:
@mernst  thanks. Any comments on what my implementation is doing or on why the tests may be failing in the Checker test suite, but not in my separate project?

renatoathaydes:
After rebasing my branch on master and running `ant reformat`, I see this:

```
➜  checker-framework git:(linear-type-checker) ✗ ant reformat
Buildfile: /home/renato/programming/projects/jsr308/checker-framework/build.xml

-run-google-java-format.check:

-get-run-google-java-format:

update-run-google-java-format:
     [exec] error: Your local changes to the following files would be overwritten by merge:
     [exec]     fixup-google-java-format.py
     [exec] Please, commit your changes or stash them before you can merge.
     [exec] Aborting
     [exec] Result: 1

reformat:


BUILD SUCCESSFUL
Total time: 19 seconds
```

The Python file is not modified in my repository... not sure if this worked or not :) it says SUCCESS but at the same time logs an error and says "aborting".

mernst:
Thanks for the details.  That makes it clear what is going on.  What aborted is the `update-run-google-java-format` rule, but the remainder of the rules can run even if it fails.  There is some local change in your repository.  I suggest you delete that entire subrepository and the next time you run `ant reformat`, it will be re-cloned and all should be well.  If the changes in your clone persist, you'll need to figure out what process on your compute is making the changes.

renatoathaydes:
I've found out why my tests were passing but not the tests in the checker framework test suite... the checker test suite applies the `-Anomsgtext` flag when running tests, which causes some extra errors... after adding this flag on my own project's tests, it also fails with the extra errors!

Here's the errors I was not expecting:

```
error: (type.invalid.annotations.on.use)
    String t = s.toUpperCase(); // used up
           ^
Runner.java:7: error: (type.invalid.annotations.on.use)
    System.out.println(t);
                       ^
```

Any ideas why these errors are occurring only when `-Anomsgtext` is used?

checker/src/org/checkerframework/checker/linear/LinearAnnotatedTypeFactory.java
```diff
@@ -1,93 +1,88 @@
...
-            assert elem != null;
-            if (this.flowState.vars.contains(elem)) {
-                int idx = this.flowState.vars.indexOf(elem);
-                if (this.flowState.annos.get(LINEAR, idx)) {
-                    this.flowState.annos.set(UNUSABLE, idx);
-                    this.flowState.annos.clear(LINEAR, idx);
+            @Nullable Tree receiverTree = node.getTarget().getReceiver().getTree();
+            if (receiverTree != null) {
+                AnnotatedTypeMirror type = factory.getAnnotatedType(receiverTree);
+                if (type.isAnnotatedInHierarchy(LINEAR)) {
```
smillst: This will always return true.  (Well, technically if type is an AnnotatedTypeVariable it could return false.) . Did you mean `hasAnnotation`?

checker/src/org/checkerframework/checker/linear/LinearAnnotatedTypeFactory.java
```diff
@@ -1,93 +1,88 @@
...
+                    return new RegularTransferResult<>(
+                            analysis.createSingleAnnotationValue(
+                                    UNUSABLE, result.getResultValue().getUnderlyingType()),
```
smillst: The refined type created must be a subtype of the declared type.  If it is not, the annotation isn't applied.  Because `@Unsable` is top, this line does change any types.

checker/src/org/checkerframework/checker/linear/LinearVisitor.java
```diff
@@ -69,6 +75,21 @@ public Void visitMemberSelect(MemberSelectTree node, Void p) {
...
+    @Override
+    protected void commonAssignmentCheck(
+            AnnotatedTypeMirror varType,
+            AnnotatedTypeMirror valueType,
+            Tree valueTree,
+            String errorKey) {
+        if (varType.hasAnnotation(Linear.class)) {
+            if (valueTree instanceof LiteralTree || valueTree instanceof NewClassTree) {
+                valueType.removeAnnotation(atypeFactory.NORMAL);
+                valueType.addAnnotation(atypeFactory.LINEAR);
```
smillst: If valueType has some other annotation beside `@Normal`, then it will have two annotations.  (An annotated type can only have one annotation per hierarchy.) Use replaceAnnotation instead.

checker/src/org/checkerframework/checker/linear/LinearVisitor.java
```diff
@@ -69,6 +75,21 @@ public Void visitMemberSelect(MemberSelectTree node, Void p) {
...
     }
 
+    @Override
+    protected void commonAssignmentCheck(
+            AnnotatedTypeMirror varType,
+            AnnotatedTypeMirror valueType,
+            Tree valueTree,
+            String errorKey) {
+        if (varType.hasAnnotation(Linear.class)) {
+            if (valueTree instanceof LiteralTree || valueTree instanceof NewClassTree) {
```
smillst: Use the org.checkerframework.framework.qual.ImplicitFor meta-annotation to specify that the type of all literals is `@Linear`.

checker/src/org/checkerframework/checker/linear/LinearAnnotatedTypeFactory.java
renatoathaydes: This is what the old code was doing :) I tried to change as little as possible but looks like that strategy is a bad one.

checker/src/org/checkerframework/checker/linear/LinearAnnotatedTypeFactory.java
renatoathaydes: Ok, so all I need to do is remove the `@Linear` annotation from the type, I guess?!

checker/src/org/checkerframework/checker/linear/LinearVisitor.java
renatoathaydes: Is that better than this? Why? Seems good to be able to  put all cases together... or is there an annotation for `@ImplicitFor(NewClassTree)` as well?

renatoathaydes: Also, not all literals are `@Linear`, only those assigned to a `@Linear` variable.

smillst: Oh, so here you were actually trying to allow literals and NewClassTrees to be assigned to any `@Linear` reference rather than changing the types of those trees.  In that case, it would be better to just return true.

I thought you were trying to change the type of all literals to `@Linear`.

mernst: Renado, if you could add comments to the code, that would help a lot in understanding and reviewing it -- we could give you better advice, more quickly.  Thanks!

checker/src/org/checkerframework/checker/linear/LinearAnnotatedTypeFactory.java
renatoathaydes: I don't get it... the intention here is to make the receiver get the `@Unusable` annotation. That's the top of the hierarchy, but obviously not the default (the default is `@Normal`). So, I don't want to refine, but to widen the type, so as to make the type of the receiver be annotated with `@Unusable`. The concept of sub-typing here seems a little backwards, but I believe my reasoning is correct, isn't it?

checker/src/org/checkerframework/checker/linear/LinearVisitor.java
renatoathaydes: Return true where? I tried overriding `isAssignable` instead and return true in the cases above, but `isAssignable` was not even called?!

renatoathaydes:
I have currently managed to allow literals and new Objects to be assigned to variables annotated with `@Linear`... but I am now stuck because of this error:

```
Runner.java:5: error: [type.invalid.annotations.on.use] invalid type: annotations [@Unusable] conflict with declaration of type java.lang.String
    String t = s.toUpperCase(); // used up
           ^
```

The code in question:

```java
import com.athaydes.checker.linear.qual.Linear;
public class Runner implements Runnable {
  public void run() {
    @Linear String s = \"hello @Linear\";
    String t = s.toUpperCase(); // used up
    System.out.println(t);
  }
}
```

I removed my transfer function, so currently all my code is doing is replacing the annotation of anything being assigned to a `@Linear` variable with the `@Linear` annotation, so it allows compilation of such cases (which it doesn't if I don't replace the annotation)...

Could you help me understand why the framework thinks there's a conflict and where the `@Unsable` annotation is coming from, given I never mentioned it in any code (and I have not modified any of the annotations definitions)?

renatoathaydes:
By debugging, I found out that the defaults seem wrong for my annotations.

This is the value of `QualifierDefaults#toString()` on the `LinearAnnotatedTypeFactory.defaults` variable:

```
Checked code defaults: 
( LOCAL_VARIABLE => @com.athaydes.checker.linear.qual.Unusable )
( RESOURCE_VARIABLE => @com.athaydes.checker.linear.qual.Unusable )
( EXCEPTION_PARAMETER => @com.athaydes.checker.linear.qual.Unusable )
( LOWER_BOUND => @com.athaydes.checker.linear.qual.Linear )
( IMPLICIT_LOWER_BOUND => @com.athaydes.checker.linear.qual.Linear )
( IMPLICIT_UPPER_BOUND => @com.athaydes.checker.linear.qual.Unusable )
( OTHERWISE => @com.athaydes.checker.linear.qual.Normal )
Unchecked code defaults: 
( FIELD => @com.athaydes.checker.linear.qual.Unusable )
( PARAMETER => @com.athaydes.checker.linear.qual.Linear )
( RETURN => @com.athaydes.checker.linear.qual.Unusable )
( LOWER_BOUND => @com.athaydes.checker.linear.qual.Linear )
( UPPER_BOUND => @com.athaydes.checker.linear.qual.Unusable )
useUncheckedCodeDefaultsSource: false
useUncheckedCodeDefaultsBytecode: false
```

So, I added `@DefaultFor({TypeUseLocation.LOCAL_VARIABLE})` on `@Normal`, then the code worked!

The question now is: why was it using `@Unusable` as a default? In `@Normal` there was already the `@DefaultQualifierInHierarchy` annotation! What's the appropriate thing to do? Add several `@DefaultFor` on `@Normal` for all use-locations I think it's appropriate (probably all of them)?

smillst:
`@DefaultQualifierInHierarachy` applies everywhere except [CLIMB](https://checkerframework.org/manual/#climb-to-top)  locations.  The CLIMB rules can be overriden using `@DefaultFor`.

checker/src/org/checkerframework/checker/linear/LinearAnnotatedTypeFactory.java
smillst: Yes, you are trying to widen the type here.  But if you widen the type farther than the declared type, the widened type is ignored.

```
   @Linear String s = "linear";
   String t = s.toUpperCase();
```
After `s.toUpperCase()`, the type of `s` should be `@Unusable`, but the declared type of `s` is `@Linear` which is a subtype of `@Unusable`, so the AnnotatedTypeFactory ignores the type produced by dataflow.

renatoathaydes: So it's impossible to widen the type? That makes it impossible to implement this checker then. At least using the flow framework.

smillst: Right that's what I was trying to say with the main comment on my review https://github.com/typetools/checker-framework/pull/1784#pullrequestreview-91657594.  I think it is possible to design a Linear Checker that can be implemented, but I haven't thought about what the rules/qualifier hierarchy should be.

renatoathaydes: What if I reverse the type hierarchy, so `@Linear` is at the bottom? That would make assigning `@Linear` variables to something else impossible, but that's fine because normally, that's probably what you actually want (that would ensure concurrency safety for the linear variable, similar to how Pony uses capabilities to forbid aliases, which helps make concurrency safe)!
Not sure if I am not seeing some unintended consequence of that, but I will try that if you don't see anything completely wrong with that reasoning :)

smillst: `@Linear` is bottom.

renatoathaydes: The concept of sub-typing is artificially being imposed on these annotations, so it's tricky to reason about it. Is it possible to just not use the sub-typing relationship at all and just implement the logic using transitions from one type to another? If so, what's the best approach to do that?

smillst: You could use just the dataflow framework, that's the classes in `org.checkerframework.dataflow`), write some sort of aliasing prevention analysis.  [NullAway](https://github.com/uber/NullAway) and [Error Prone](http://errorprone.info/) both use the dataflow framework for there analyses.  (Dataflow manual is [here](https://checkerframework.org/manual/checker-framework-dataflow-manual.pdf).
  
But if you want to use a checker, then you need a proper type hierarchy.  I think it is possible to design such a hierarchy for the Linear Checker. I just haven't spent the time to come up with the rules.

renatoathaydes: Thanks @smillst I will see if I can find time to read the full paper on the weekend or some other day. This is fun, but I don't have infinite time to spend on it, unfortunately!

BTW reversing the type hierarchy causes that infamous error again and I have no idea why:

```
[type.invalid.annotations.on.use] invalid type: annotations [@Linear] conflict with declaration of type (@com.athaydes.checker.linear.qual.Linear :: java.lang.Object)
```

But nevermind, it seems clear this approach is not going anywhere.

smillst:
I'm closing this pull request since we decided that the current Linear Checker hierarchy makes it impossible to use the dataflow framework.

PR #1786, opened 2018-01-25
lgtm.yml: fix ant command

aibaars:
A last minute edit to break a long line into two was done wrongly. As a result
the 'ant' command line ended up contain some spurious \ symbols causing the build
to fail.

@wmdietl Apologies for the mistake, I really shouldn't edit  yaml files without validating them.

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -21,5 +21,4 @@ extraction:
     index:
       build_command:
       - cd checker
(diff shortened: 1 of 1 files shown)


aibaars:
Btw you can already see java results on lgtm.com for an older commit to get a feel for the kind of issue it finds: https://lgtm.com/projects/g/typetools/checker-framework/alerts/?mode=list

However, without a working lgtm.yml file the results won't be updated.

PR #1787, opened 2018-01-25
Assume that minimum length of sequences is always at least 0

panacekcz:
The method `ValueAnnotatedTypeFactory.getMinLenValue(AnnotationMirror)` currently returns `null` if there is no information about the minimum length of an array. However, the minimum length is always at least 0. Callers of  this method must usually check for `null` and treat it as 0. However, in the upper bound check of array element access, it treated it differently, causing unnecessary upper bound warning on negative constant indices.

This PR changes the method and several related method to never return `null` and return 0 if the minimum length is not known. This reduces the burden on callers of this method and fixes the upper bound check.
Fixes kelloggm#216.

### DIFF
--- a/framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
+++ b/framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
@@ -202,12 +202,8 @@ public class ValueAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
     @Override
     public AnnotationMirror aliasedAnnotation(AnnotationMirror anno) {
         if (AnnotationUtils.areSameByClass(anno, MinLen.class)) {
-            Integer from = getMinLenValue(anno);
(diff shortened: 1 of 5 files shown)


framework/src/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -2123,14 +2119,14 @@ public Integer getMaxLenValue(AnnotationMirror annotation) {
     }
 
     /**
-     * Used to find the minimum length of an array, which is useful for array bounds checking.
-     * Returns null if there is no minimum length known, or if the passed annotation is null.
+     * Finds a minimum length of an array specified by the provided annotation. Returns 0 if there
+     * is no minimum length known, or if the passed annotation is null.
```
kelloggm: This method still returns `null` if the annotation is `null`.

PR #1789, opened 2018-01-26
Added Awt.astub annotated for nullness checker

ashishrana160796:


mernst:
Please see comments on https://github.com/typetools/annotated-libraries/pull/1.  Annotated libraries and stub files are different things.  The annotated JDK is an annotated library, so adding a stub file to its directory will not have any effect.

ashishrana160796:
Yeah, really my fault for both pull requests. Let me annotate ._java_ files for nullness checker from _awt_ and _swing_ packages and than make a pull requests again and also update java jdk8 in annotated library also. 
Will try to do it as early as possible.

mernst:
That sounds great.  Thanks very much!  This should be less confusing in the future, thanks to the text I added to the manual.

PR #1790, opened 2018-01-26
Preserve upper bound for right shift

panacekcz:
This PR fixes kelloggm#214 (upper bounds for right shift) by copying the upper bound annotation from the left operand, if it is non-negative.

After this is merged, kelloggm#217 can be implemented.

### DIFF
--- a/checker/src/org/checkerframework/checker/index/lowerbound/LowerBoundAnnotatedTypeFactory.java
+++ b/checker/src/org/checkerframework/checker/index/lowerbound/LowerBoundAnnotatedTypeFactory.java
@@ -409,6 +409,7 @@ public class LowerBoundAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
 
     /** Checks if the expression is non-negative, i.e. it has Positive on NonNegative annotation. */
     public boolean isNonNegative(Tree tree) {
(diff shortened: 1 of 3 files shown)


checker/src/org/checkerframework/checker/index/lowerbound/LowerBoundAnnotatedTypeFactory.java
```diff
@@ -406,4 +406,10 @@ private AnnotationMirror checkForMathRandomSpecialCase(Tree randTree, Tree arrLe
         }
         return null;
     }
+
+    /** Checks if the expression is non-negative, i.e. it has Positive on NonNegative annotation. */
```
kelloggm: You should use this method in the modulo PR too.

kelloggm: And, should this replace `isNonNegative` in `LowerBoundTransfer`? At the very least, please leave a TODO there that we can come back to later.

panacekcz: Not exactly - the parameter type is different; perhaps they could two overloads in one class.
I added the TODO.

panacekcz:
Can this be merged?

PR #1791, opened 2018-01-26
Infer upper bounds for modulo operator

panacekcz:
Fixes kelloggm#218 exactly as described there.

### DIFF
--- a/checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
+++ b/checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
@@ -29,10 +29,8 @@ import org.checkerframework.checker.index.qual.LTLengthOf;
 import org.checkerframework.checker.index.qual.LTOMLengthOf;
 import org.checkerframework.checker.index.qual.LengthOf;
 import org.checkerframework.checker.index.qual.NegativeIndexFor;
-import org.checkerframework.checker.index.qual.NonNegative;
 import org.checkerframework.checker.index.qual.PolyIndex;
 import org.checkerframework.checker.index.qual.PolyUpperBound;
-import org.checkerframework.checker.index.qual.Positive;
 import org.checkerframework.checker.index.qual.SameLen;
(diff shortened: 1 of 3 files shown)


checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -470,6 +473,26 @@ private ExpressionTree getLengthSequenceTree(ExpressionTree lengthTree) {
...
 
+        /** Infers upper-bound annotation for {@code numerator % divisor}. */
+        private void addAnnotationForRemainder(
+                ExpressionTree numeratorTree,
+                ExpressionTree divisorTree,
+                AnnotatedTypeMirror resultType) {
+            LowerBoundAnnotatedTypeFactory lowerBoundATF = getLowerBoundAnnotatedTypeFactory();
+            UBQualifier result = UpperBoundUnknownQualifier.UNKNOWN;
+            // if numerator >= 0, then numerator%divisor <= numerator
+            if (lowerBoundATF.getAnnotatedType(numeratorTree).hasAnnotation(NonNegative.class)) {
```
kelloggm: Positive should be permitted here too. There are methods for this in `LowerBoundTransfer`, but they're private. You could make them accessible or just copy them over, since they're one line each. Or just check for Positive directly here.

checker/tests/index/Modulo.java
```diff
@@ -0,0 +1,19 @@
+// Test case for kelloggm 218
+// https://github.com/kelloggm/checker-framework/issues/218
+import org.checkerframework.checker.index.qual.IndexFor;
+import org.checkerframework.checker.index.qual.IndexOrHigh;
+import org.checkerframework.checker.index.qual.NonNegative;
+
+public class Modulo {
+    void m1(Object[] a, @IndexOrHigh("#1") int i, @NonNegative int j) {
```
kelloggm: Add a test like this where `j` is `@Positive`

panacekcz:
@kelloggm, thanks for noticing that. I made a helper method in `LowerBoundAnnotatedTypeFactory` for the check.

PR #1792, opened 2018-01-26
Fold compile-time constants (integer literals and lengths of string literals) in offsets

panacekcz:
This PR makes the Upper Bound Checker recognize integer literals and length of string literals in offset expressions as constants.
Fixes kelloggm#176.

This PR uses a simple name comparison to recognize the string length, not `IndexMethodIdentifier` which would mean passing it as a parameter through all the expression standardization methods.

### DIFF
--- a/checker/src/org/checkerframework/checker/index/upperbound/OffsetEquation.java
+++ b/checker/src/org/checkerframework/checker/index/upperbound/OffsetEquation.java
@@ -257,6 +257,7 @@ public class OffsetEquation {
                 return (Integer) value;
             }
         } else if (termReceiver instanceof FlowExpressions.MethodCall) {
+            // TODO: generalize
             // Length of string literal
             FlowExpressions.MethodCall call = (FlowExpressions.MethodCall) termReceiver;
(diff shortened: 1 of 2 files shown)


checker/src/org/checkerframework/checker/index/upperbound/OffsetEquation.java
```diff
@@ -245,6 +245,65 @@ public boolean isNegativeOrZero() {
...
+     * returns null.
+     */
+    private Integer evalConstantTerm(Receiver termReceiver) {
+        if (termReceiver instanceof FlowExpressions.ValueLiteral) {
+            // Integer literal
+            Object value = ((FlowExpressions.ValueLiteral) termReceiver).getValue();
+            if (value instanceof Integer) {
+                return (Integer) value;
+            }
+        } else if (termReceiver instanceof FlowExpressions.MethodCall) {
```
kelloggm: I don't like that this is specialized to string length, but I'll accept it for now. Can you add a TODO to generalize this here before you merge?

PR #1793, opened 2018-01-26
Fix NPE in Index Checker.

smillst:


mernst:
Tests are failing.

kelloggm:
This is failing due to the plume-lib typecheck job failing with the Lock Checker. It looks like @panacekcz's open PRs are also failing in the same way: see #1792 and #1791. The master build for the Checker Framework is also failing.

My guess is that the changes to the Lock Checker JDK merged earlier today by @mernst in 941d346ebf34b3cd3aabfd3539ede34788d98a54 are responsible, but I don't know why the tests didn't fail for that commit (travis shows master passed for them...).

mernst:
Yes, I broke them.  Sorry!  Like you, I don't understand that.

If only the lock checker is failing, we should merge.  (We'll have to manually check, which is a pain.)
I'll try to fix this, too.

smillst:
I updated the jdk.jar to include the Locker Checker updates.  Everything should pass now.

PR #1795, opened 2018-01-27
Added nullness annotations for java.applet

ashishrana160796:


ashishrana160796:
Closed because contains imports from _java.awt_ package. Therefore, annotations for awt package needed to be done before for proper functionality.

PR #1799, opened 2018-01-28
Refine upper bound of a subtrahend with offset

panacekcz:
This PR adds refinement to the transfer of the Upper Bound Checker, to refine the upper bound of a subtrahend in a condition such as `a.length - i > j`. Currently `j` is refined to `@LTLengthOf(value="a", offset="i")`. This refines `i` to `@LTLengthOf(value="a", offset="j")` as well.

The rules are:
- if `minuend` is `@LTLengthOf(value="a", offset="o")` and `minuend-subtrahend>offset`, then `subtrahend` is `@LTLengthOf(value="a", offset="o+offset+1")` 
- if `minuend` is `@LTLengthOf(value="a", offset="o")` and `minuend-subtrahend>=offset`, then `subtrahend` is `@LTLengthOf(value="a", offset="o+offset")` 

Fixes kelloggm#215

### DIFF
--- a/checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
+++ b/checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
@@ -226,7 +226,7 @@ public class UpperBoundTransfer extends IndexAbstractTransfer {
             propagateToOperands((LessThanLengthOf) largerQualPlus1, smaller, in, store);
         }
 
-        refineSubtrahendWithOffset(larger, smaller, 1, in, store);
+        refineSubtrahendWithOffset(larger, smaller, true, in, store);
 
         Receiver rightRec = FlowExpressions.internalReprOf(analysis.getTypeFactory(), smaller);
         store.insertValue(rightRec, atypeFactory.convertUBQualifierToAnnotation(refinedRight));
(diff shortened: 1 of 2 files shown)


panacekcz:
I needed this to work for `switch`-`case` too. The implementation is not ideal but seems to work.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
```diff
@@ -252,10 +255,49 @@ protected void refineGTE(
...
+     * The type of the subtrahend is refined to the type of the minuend with the offset added.
+     *
+     * <p>This is based on the fact that if {@code (minuend - subtrahend) >= offset}, and {@code
+     * minuend + o < l}, then {@code subtrahend + o + offset < l}.
+     *
+     * <p>If gtNode is not a subtraction, the method does nothing.
+     *
+     * @param gtNode the node that is greater or equal to the offset
+     * @param offsetNode a node part of the offset
+     * @param offsetVal a constant part of the offset
```
kelloggm: I would use a boolean for this, since it only seems to take on the value of 0 or 1 depending on whether the refinement is being applied in a gt/gte node.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundTransfer.java
```diff
@@ -252,10 +255,49 @@ protected void refineGTE(
...
     }
 
+    /**
+     * Refines the subtrahend in a subtraction which is greater than or equal to a certain offset.
+     * The type of the subtrahend is refined to the type of the minuend with the offset added.
+     *
+     * <p>This is based on the fact that if {@code (minuend - subtrahend) >= offset}, and {@code
+     * minuend + o < l}, then {@code subtrahend + o + offset < l}.
+     *
+     * <p>If gtNode is not a subtraction, the method does nothing.
```
kelloggm: Say "if gtNode is not a `NumericalSubtractionNode`" instead

kelloggm: Also do `@code gtNode`

PR #1800, opened 2018-01-29
Fix #1781: Add an independent TypeKind enum to not create a warning.

### LINKED ISSUE #1781: consider avoiding references to `javax.lang.model.type.TypeKind` in annotations for Android compatibility
The checker framework annotations contain some references to `javax.lang.model.type.TypeKind`, which is not available on Android:

https://github.com/typetools/checker-framework/blob/661bd01b19172f42944f1b463af5fa705ee43b04/checker/src/org/checkerframework/checker/nullness/qual/NonNull.java#L40-L43

This results is javac warnings when compiling with the checker annotations on the classpath and an Android bootclasspath, e.g.:

```
  reason: class file for javax.lang.model.type.TypeKind not found
warning: unknown enum constant TypeKind.INT
```

The log spam is usually harmless, but I recently encountered a case where it's tickling a javac bug: [JDK-8195986](https://bugs.openjdk.java.net/browse/JDK-8195986).

I wondered if you'd considered avoiding the references to `javax.lang.model.type.TypeKind` from annotations, and perhaps using a custom enum that could be translated to the `javax.lang.model` types inside the checker framework itself (which unlike the annotations target is on the processorpath, not the classpath, and unaffected by the Android compilation bootclasspath).

This would be similar to the approach taken for `com.sun.source.tree.Tree.Kind` in https://github.com/typetools/checker-framework/commit/6cacf1ed2b1ccba001a2f9ba350da5a93e1e4311.

wmdietl:
... on Android.
Downstream tests depend on
https://github.com/typetools/checker-framework-inference/pull/68

### DIFF
--- a/checker/src/org/checkerframework/checker/interning/qual/Interned.java
+++ b/checker/src/org/checkerframework/checker/interning/qual/Interned.java
@@ -5,10 +5,10 @@ import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
-import javax.lang.model.type.TypeKind;
 import org.checkerframework.framework.qual.ImplicitFor;
 import org.checkerframework.framework.qual.LiteralKind;
 import org.checkerframework.framework.qual.SubtypeOf;
+import org.checkerframework.framework.qual.TypeKind;
 
 /**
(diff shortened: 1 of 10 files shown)


wmdietl:
All tests pass. Only downstream failure in checker-framework-inference which gets fixed by typetools/checker-framework-inference#68

PR #1808, opened 2018-02-01
 Added nullness annotations for javax.xml.parsers 

ashishrana160796:
The specifications are referenced from following links :
https://docs.oracle.com/javase/8/docs/api/javax/xml/parsers/package-summary.html
For remaining SecuritySupport.java and FactoryFinder.java, following link was used  : 
http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/javax/xml/parsers/


checker/jdk/nullness/src/javax/xml/parsers/DocumentBuilder.java
```diff
@@ -0,0 +1,353 @@
...
+     * <p>The reset <code>DocumentBuilder</code> is not guaranteed to have the same {@link EntityResolver} or {@link ErrorHandler}
+     * <code>Object</code>s, e.g. {@link Object#equals(Object obj)}.  It is guaranteed to have a functionally equal
+     * <code>EntityResolver</code> and <code>ErrorHandler</code>.</p>
+     *
+     * @throws UnsupportedOperationException When implementation does not
+     *   override this method.
+     *
+     * @since 1.5
+     */
+    @SuppressWarnings("nullness")
```
mernst: For every `@SuppressWarnings` annotation, please add a brief comment on the same line explaining why the warning is being suppressed.  That is, please justify that the code is safe, even though the Nullness Checker is unable to prove that fact.  For some examples, see https://checkerframework.org/manual/#suppresswarnings-best-practices .

checker/jdk/nullness/src/javax/xml/parsers/SecuritySupport.java
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * This class is duplicated for each JAXP subpackage so keep it in sync.
+ * It is package private and therefore is not exposed as part of the JAXP
+ * API.
+ *
+ * Security related methods that only work on J2SE 1.2 and newer.
+ */
+class SecuritySupport  {
+
+	@SuppressWarnings("nullness")
```
mernst: Please don't use tabs in source files.  Instead, use spaces.  Tabs display differently depending on one's editor -- you can see here that it causes the `@SuppressWarnings` not to line up with the code below it.

checker/jdk/nullness/src/javax/xml/parsers/FactoryConfigurationError.java
```diff
@@ -0,0 +1,137 @@
...
+ * @version $Revision: 1.7 $, $Date: 2010-11-01 04:36:09 $
+ */
+
+public class FactoryConfigurationError extends Error {
+    private static final long serialVersionUID = -827108682472263355L;
+
+    /**
+     *<code>Exception</code> that represents the error.
+     */
+    private Exception exception;
```
mernst: It seems that this needs `@Nullable`.  That would be more accurate, and would also enable you to remove the `@SuppressWarnings("nullness")` on [line 52](https://github.com/typetools/checker-framework/pull/1808/files#diff-b3b560eba2d73854dde827a8969a4e3eR52). and line 64.

checker/jdk/nullness/src/javax/xml/parsers/FactoryConfigurationError.java
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * Create a new <code>FactoryConfigurationError</code> with a
+     * given <code>Exception</code> base cause of the error.
+     *
+     * @param e The exception to be encapsulated in a
+     * FactoryConfigurationError.
+     */
+
+    public FactoryConfigurationError(Exception e) {
```
mernst: Why not permit the formal parameter to be null, by marking it as `@Nullable`?  That's a legal value, and it seems unnecessary to reject clients that pass `null`.

checker/jdk/nullness/src/javax/xml/parsers/SecuritySupport.java
ashishrana160796: Will make changes and keep in mind for next requests.
Thanks.

ashishrana160796:
The main attempt was to stay true to specifications mentioned in the documentation with minimum concern about implementation details  i.e. not writing any extra 'Nullable' annotation in these files as while using these packages only documentation is visible.  As mentioned in Sec. 28.5 of the documentation. 

_When you write library annotations, you should base them on the library documentation, typically the method’s Javadoc.
If a fact is not mentioned in the documentation, then it is an implementation detail. Clients should not depend on such implementation details, which are prone to change without notice._

Instead, I have used this warning suppression approach for annotations as a way around. But this also looks unreasonable as pointed out you in later comments. I'll soon make all the changes as pointed by you in the pull requests.

Thanks.

mernst:
I see; thank you for explaining your reasoning.
Commit 61846fe2f26dcf3c0301a0ef96a7ad10cb062c93 updates the manual to clarify that public annotations should reflect the specification.  Annotations on private fields and methods are useful too, but not to a client.

mernst:
I have revised the manual again to better explain how to annotate a library and when it is acceptable not to suppress warnings:  a21f6fa980954af72c03521351afddff249ab243

checker/jdk/nullness/src/javax/xml/parsers/FactoryConfigurationError.java
ashishrana160796: @mernst  commit : d903ab9 makes all the changes as pointed out.
Thanks.

checker/jdk/nullness/src/javax/xml/parsers/DocumentBuilder.java
```diff
@@ -85,14 +85,14 @@ protected DocumentBuilder () {
      *
      * @since 1.5
      */
-    @SuppressWarnings("nullness")
+    // Commented lines from nullness checker analysis. These dereferences might cause NullPointerExceptions while dereferencing.
```
mernst: Please don't write comments like this.  It is better to suppress warnings than to introduce code changes.
Commenting is only appropriate when the code does not compile with the regular type-checker.
Any time you suppress warnings or comment out code, add an explanation of why.
Your explanation above suggests that this is buggy code in the JDK.  Should you report it?
I'm confused how `this.getClass().getName()` can throw a null pointer exception; can you explain?

ashishrana160796: this.getClass().getName() will not cause any problem, it is wrongly commented out. But this.getClass().getPackage().getSpecificationTitle() & this.getClass().getPackage().getSpecificationVersion() can cause NullPointerExceptions as getPackage() does return null. But, I think I'll go with your suggestion and make changes accordingly.
Thanks.

checker/jdk/nullness/src/javax/xml/parsers/DocumentBuilderFactory.java
```diff
@@ -0,0 +1,613 @@
...
+     * java -Djaxp.debug=1 YourProgram ....
+     * </pre>
+     *
+     * @return New instance of a <code>DocumentBuilderFactory</code>
+     *
+     * @throws FactoryConfigurationError in case of {@linkplain
+     * java.util.ServiceConfigurationError service configuration error} or if
+     * the implementation is not available or cannot be instantiated.
+     */
+    @SuppressWarnings("nullness")    // find method in FactoryFinder accepts @Nullable args whereas here the requirement is @NonNull
```
mernst: I don't understand this warning suppression.  What error is thrown if it is removed?

ashishrana160796: No error will be thrown while running with annotated `FactoryFinder.java`. While annotating. my approach is to annotate files independent of others, keeping in mind the scenario of partially annotated library. 
Will change it.
-Thanks.

checker/jdk/nullness/src/javax/xml/parsers/FactoryFinder.java
```diff
@@ -285,7 +285,6 @@ private static void dPrint(String msg) {
      *
      * @return instance of provider class if found or null
      */
-    @SuppressWarnings("nullness")    // Checker analysis for return type found : @FBCBottom @Nullable  null but it also returns iterator.next() which can be non-null also. In x.getMessage() parameter found : @Initialized @Nullable String but this will not dereference a null here as underlying exception 'e' is not null.
```
ashishrana160796: This removal causes the following error. Therefore, I am adding a suppress warning here.
found   : @FBCBottom @Nullable  null
required: T[ extends @Initialized @Nullable Object super @Initialized @NonNull Void]

checker/jdk/nullness/src/javax/xml/parsers/FactoryFinder.java
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    /*
+     * Try to find provider using the ServiceLoader API
+     *
+     * @param type Base class / Service interface  of the factory to find.
+     *
+     * @return instance of provider class if found or null
+     */
+    @SuppressWarnings("nullness")    // null return is acceptable as return value.
```
mernst: I don't understand this comment.  What warning is being suppressed, and how do we know that cannot possibly cause a problem elsewhere?

ashishrana160796: Without this, while running checker the following mentioned above error comes, which doesn't allow null to be returned in this case. But according to specification null is allowed, so I think this suppression would not be causing any problem elsewhere. 
The comment is about acceptability of null value as return which is producing error. 

---
But I think considering the discussion from pr #1849 will be much better approach here and are valid annotations in this case.

mernst:
Does the code still type-check without warnings?

ashishrana160796:
Yes, with these @nullable annotations in place it does type checks.
If there is a test-case that will get failed, can be pointed out .

ashishrana160796:
Actually I was asking or considering a possibility of test-case that doesn't pass these annotations.
If there exists one, than suppression annotations can be added. Otherwise, there is no need for it.

mernst:
OK.  I think this is acceptable as is.  If such a test case comes up, a user will complain.

Issue #1809, opened 2018-02-02
Member reference caching crash


cushon:
This is a regression since 2.3.1. When I was creating the repro I noticed the crash stopped happening after removing unrelated methods, which appears to be related to `-AatfCacheSize`. Using a lower value makes it possible to reproduce with a minimal example.

```java
import java.util.List;
import java.util.Optional;
import java.util.stream.Stream;

abstract class Test {

  abstract <T> Stream<T> concat(Stream<? extends T>... streams);

  abstract Optional<A> f();

  private static class A {}

  interface B {
    List<C> g();
  }

  interface C {
    List<S> h();
  }

  interface S {}

  private Stream<A> test(B b) {
    return concat(b.g().stream().flatMap(a -> a.h().stream().map(c -> f())))
        .filter(Optional::isPresent)
        .map(Optional::get);
  }
}
```

```
./checker-framework-2.3.2/checker/bin/javac -AatfCacheSize=7 -AprintErrorStack -processor nullness Test.java
...
error: AsSuperVisitor: type is not an erased subtype of supertype.
  type: Object
  superType: A
  Compilation unit: Test.java
  Exception: java.lang.Throwable; Stack trace: org.checkerframework.framework.source.SourceChecker.errorAbort(SourceChecker.java:725)
  org.checkerframework.javacutil.ErrorReporter.errorAbort(ErrorReporter.java:41)
  org.checkerframework.framework.type.AsSuperVisitor.errorTypeNotErasedSubtypeOfSuperType(AsSuperVisitor.java:154)
  org.checkerframework.framework.type.AsSuperVisitor.visitDeclared_Declared(AsSuperVisitor.java:339)
  org.checkerframework.framework.type.AsSuperVisitor.visitDeclared_Declared(AsSuperVisitor.java:28)
  org.checkerframework.framework.util.AtmCombo.accept(AtmCombo.java:308)
  org.checkerframework.framework.type.visitor.AbstractAtmComboVisitor.visit(AbstractAtmComboVisitor.java:59)
  org.checkerframework.framework.type.AsSuperVisitor.visit(AsSuperVisitor.java:85)
  org.checkerframework.framework.type.AsSuperVisitor.asSuper(AsSuperVisitor.java:67)
  org.checkerframework.framework.util.AnnotatedTypes.asSuper(AnnotatedTypes.java:116)
[...]

PR #1813, opened 2018-02-04
Fix Issue #1809 by determining better type for member references.

wmdietl:
1. Correctly type the "qualifier expression" of a member reference, where "qualifier expression" is the type or expression before `::`.
2. Use separate caches for types for Expressions, Members, and Type Trees.  (Using a single cache masked the bug.)

### DIFF
--- a/framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
+++ b/framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
@@ -2504,17 +2504,17 @@ public class BaseTypeVisitor<Factory extends GenericAnnotatedTypeFactory<?, ?, ?
         // That is handled separately in method receiver check.
 
         // The type of the expression or type use, <expression>::method or <type use>::method.
-        ExpressionTree qexpTree = memberReferenceTree.getQualifierExpression();
-        AnnotatedTypeMirror enclosingType = atypeFactory.getAnnotatedTypeFromTypeTree(qexpTree);
-        if (qexpTree.getKind() == Tree.Kind.IDENTIFIER) {
-            // This is a hack around the ambiguity between identifiers that could be
-            // types or expressions. Using TreeUtils.isTypeTree unfortunately doesn't
-            // help, as it isn't precise enough.
-            // TODO: determine a nicer way. Is taking the main qualifiers enough?
-            AnnotatedTypeMirror expType = atypeFactory.getAnnotatedType(qexpTree);
-            if (enclosingType != expType) {
-                enclosingType.replaceAnnotations(expType.getEffectiveAnnotations());
-            }
+        final ExpressionTree qualifierExpression = memberReferenceTree.getQualifierExpression();
+        final ReferenceKind memRefKind = ((JCMemberReference) memberReferenceTree).kind;
+        AnnotatedTypeMirror enclosingType;
+        if (memberReferenceTree.getMode() == ReferenceMode.NEW
+                || memRefKind == ReferenceKind.UNBOUND
+                || memRefKind == ReferenceKind.STATIC) {
+            // The "qualifier expression" is a type tree.
+            enclosingType = atypeFactory.getAnnotatedTypeFromTypeTree(qualifierExpression);
+        } else {
+            // The "qualifier expression" is an expression.
+            enclosingType = atypeFactory.getAnnotatedType(qualifierExpression);
         }
 
         // ========= Overriding Executable =========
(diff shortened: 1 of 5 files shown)


framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -2504,8 +2504,18 @@ protected boolean checkMethodReferenceAsOverride(
...
         // The type of the expression or type use, <expression>::method or <type use>::method.
-        AnnotatedTypeMirror enclosingType =
-                atypeFactory.getAnnotatedType(memberReferenceTree.getQualifierExpression());
+        ExpressionTree qexpTree = memberReferenceTree.getQualifierExpression();
+        AnnotatedTypeMirror enclosingType = atypeFactory.getAnnotatedTypeFromTypeTree(qexpTree);
+        if (qexpTree.getKind() == Tree.Kind.IDENTIFIER) {
+            // This is a hack around the ambiguity between identifiers that could be
+            // types or expressions. Using TreeUtils.isTypeTree unfortunately doesn't
+            // help, as it isn't precise enough.
+            // TODO: determine a nicer way. Is taking the main qualifiers enough?
```
smillst: I think the correct way is to check the `kind` field of the member reference.  I tried this here:
https://github.com/typetools/checker-framework/compare/master...smillst:smillst1809?expand=1  And it seems to work with out changing the ATF caches.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -306,10 +306,25 @@
     private final Map<Tree, AnnotatedTypeMirror> classAndMethodTreeCache;
 
     /**
-     * Mapping from a Tree to its annotated type; before implicits are applied, just what the
-     * programmer wrote.
+     * Mapping from an expression tree to its annotated type; before implicits are applied, just
```
smillst: Why all the changes to caching?

wmdietl: This issue is only triggered with a small cache. The reason why this worked with a large cache is that getting types for Expressions, Members, and Types used the same cache. An earlier request for a type tree was then found when an expression query happened. With a small cache that earlier value gets evicted.
To prevent this kind of problem completely, this changes it into three separate caches. We should never again run into a situation where something gets cached as an expression and then returned for a TypeTree query.
That would seem like a big win.

framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
wmdietl: Ah, I only looked in MemberReferenceTree not in JCMemberReference. Thanks, I'll change this.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
smillst: Makes sense.  Thanks for the explanation!

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -306,10 +306,25 @@
...
+     * what the programmer wrote. Might alias the same object as fromMemberTreeCache and
```
smillst: The maps might alias each other?  I'm not sure what this last sentence means.

framework/src/org/checkerframework/framework/type/TypeFromTypeTreeVisitor.java
```diff
@@ -40,7 +40,7 @@
  *
  * @see org.checkerframework.framework.type.TypeFromTree
  */
-class TypeFromTypeTreeVisitor extends TypeFromTreeVisitor {
+class TypeFromTypeTreeVisitor extends TypeFromExpressionVisitor {
```
smillst: Why this change?

wmdietl: This was necessary because `getAnnotatedTypeFromTypeTree` was possibly called on an expression tree.
This is no longer necessary and I undid this.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
wmdietl: The previous situation was that we had one field and one cache.
This changes it to three fields and three caches. This comment was supposed to say that we could also have three fields that all use the same cache.
I think we should just stick with the three separate caches and I'll remove the comment.

Issue #293, opened 2015-07-03
Incorrect handling of 'finally'?


GoogleCodeExporter:
```
Not sure if this is a bug or a misunderstanding on my end.

    public abstract class Init
    {
        public void f()
        {
            String s;
            try {
                s = read();
            }
            catch (Exception e) {
                return;
            }
            finally {
                write("finished");
            }

            write(s);   // <-- error reported here
        }

        public abstract String read() throws Exception;
        public abstract void write(String s);
    }

I expected the nullness checker to report no errors.

# java -jar binary/checkers.jar -processor checkers.nullness.NullnessChecker 
Init.java
Init.java:16: error: incompatible types in argument.
        write(s);
              ^
  found   : @UnknownInitialization @Nullable String
  required: @Initialized @NonNull String
1 error

Removing the 'finally' block entirely makes the error go away.
```

Original issue reported on code.google.com by `kan...@cakoose.com` on 18 Dec 2013 at 1:10

mernst:
I have added another test case to file checker-framework/checker/tests/initialization/fbc/TryFinally.java.  Thanks to Ed Price for the test case.

The problem is that our analysis considers the finally block only once.  On some paths to the finally block, the field has already been initialized, and on other paths to the finally block, the field has not yet been initialized.  Therefore, it conservatively assumes the field has not yet been initialized, resulting in a false positive warning.

As background, each program is represented by an AST (abstract syntax tree) and a control flow graph (CFG).  The dataflow analysis, including initialization analysis is performed over the CFG.  Type-checking is done over the AST (but it should really be done over the CFG).

The problem is that there is only one copy of the finally block in the constructed control flow graph.  The best solution is to duplicate the finally block in the control flow graph.

The challenge with this solution is that the rest of the framework currently assumes that CFG nodes are unique.  Most importantly, when checking the type of an AST expression, we know that it only maps to one CFG node.  If it could map to more than one, then we'd have to do more work to compute the type, perhaps by taking the LUB of all of the copies.  As a more minor point, we would also have to be careful not to produce two different error messages for different CFG nodes that correspond to a single AST expression.

Eventually, we would like to do all type-checking over the CFG, just as the dataflow analysis is currently done over the CFG.  That would make this bug easier to fix.  However, that would be a more substantial refactoring -- it is not the way to get this bug fixed quickly.  The quicker thing to do is to keep typechecking over the AST, but duplicate the finally block in the CFG that is used for dataflow and initialization analysis.

[...]

PR #1814, opened 2018-02-05
Add CFG edges into finally blocks

### LINKED ISSUE #1797: missing CFG edges from return statement to enclosing finally blocks
I notice from [this comment](https://github.com/typetools/checker-framework/blob/4c7fa25c57f251d473d86ff74985d6d240e94e33/dataflow/src/org/checkerframework/dataflow/cfg/CFGBuilder.java#L4151) that there seem to be missing control-flow edges from a `return` statement to the beginning of enclosing finally blocks.  We noticed some unsoundness in NullAway that I think is due to this issue.  E.g., this example gets through NullAway with no warning:
```java
public void foo(@Nullable Object o) {
  try {
    return;
  } finally {
    o.toString();
  }
}
```
Looking under the hood, it seems the dataflow analysis does not even visit the `finally` block, I'm guessing because it starts from the entry and there's no path to the finally block from the entry block in the CFG?

I notice that the above example and other similar examples seem to be checked correctly by the Checker Framework's nullness checker.  Is there some special workaround in that checker that we can borrow?  Is the underlying issue hard to fix?

wmdietl:
As described in #1797 return statements didn't have CFG edges to their enclosing finally blocks.
This actually was a bigger problem, as it also applied to break and continue statements.
This PR makes sure that there are CFG edges *into* finally blocks from all control flow statements.
Some CFG edges *out of* finally blocks still seem incorrect. This will be addressed after #293 in separate PRs. (This has no effect on type checking, but is important for other uses of dataflow.)

Fixes #1797

### DIFF
--- a/dataflow/src/org/checkerframework/dataflow/cfg/CFGBuilder.java
+++ b/dataflow/src/org/checkerframework/dataflow/cfg/CFGBuilder.java
@@ -802,10 +802,10 @@ public class CFGBuilder {
      * when control flows outside of the try block.
      */
     @SuppressWarnings("serial")
-    protected static class FakeLabelsMap extends HashMap<Name, Label> implements Map<Name, Label> {
+    protected static class TryFinallyScopeMap extends HashMap<Name, Label> {
         private final Label finallyLabel;
 
-        protected FakeLabelsMap(Label finallyLabel) {
+        protected TryFinallyScopeMap(Label finallyLabel) {
             this.finallyLabel = finallyLabel;
         }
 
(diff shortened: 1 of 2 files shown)


smillst:
Could you please change the pull request description to explain how this corrects finally blocks?  I'm not sure what the change is suppose to do, so I can't effectually review it.

wmdietl:
I've added a description to properly explain the PR. I'm not sure how to put that into the title...

wmdietl:
Cross-ref to comment on issue: https://github.com/typetools/checker-framework/issues/1797#issuecomment-363216415

dataflow/src/org/checkerframework/dataflow/cfg/CFGBuilder.java
```diff
@@ -757,6 +781,47 @@ public void popFrame() {
...
+        }
+    }
+
+    /**
+     * A map that keeps track of new labels added within a try block. For names that are outside of
+     * the try block, the finally label is returned. This ensures that a finally block is executed
+     * when control flows outside of the try block.
+     */
+    @SuppressWarnings("serial")
+    protected static class FakeLabelsMap extends HashMap<Name, Label> implements Map<Name, Label> {
```
wmdietl: `FakeLabelsMap` is probably not the best name.
Is `FinallyScopeLabelMap` better? Other suggestions?

smillst: `FinallyScopeLabelMap ` is better.  What about `TryFinallyLablesMap`?   I don't have a preference.

PR #1824, opened 2018-02-07
Unbundle annotations that are only used for addAliasedAnnotation

### LINKED ISSUE #1810: Consider unbundling annotations that are only used for addAliasedAnnotation
The checker framework currently redistributes copies of some common annotations for use with `addAliasedAnnotation`, e.g.:

https://github.com/typetools/checker-framework/blob/d66a363de20c7ec4642dc71aed184f082894eddb/checker/src/lombok/NonNull.java#L15

https://github.com/typetools/checker-framework/blob/62b93aebe90775d28efde01a759a6a5abb16f225/checker/src/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java#L104

Please consider using `addAliasedAnnotation(String, AnnotationMirror)` instead of redeclaring the annotations. The documentation for that overload mentions "it is nicer and less error-prone to use `addAliasedAnnotation(Class, AnnotationMirror)`", but since the annotations are being re-declared this doesn't prevent errors where the checker framework's version diverges from the canonical definitions.

Also, similar to #1739, the duplicate definitions could clash with other copies of those classes on the classpath. Having duplicate definitions of annotations is less problematic than other kinds of version conflicts, but it's still good to avoid it.

cushon:
Fixes #1810


mernst:
@cushon Liam, the changes you made look good -- thanks!  Some other changes are needed -- tests are failing because of failing imports, but we do want to test the alias feature.  Here are some examples (but not all of them):

```
Issue308.java:1: error: package javax.validation.constraints does not exist
FindBugs.java:1: error: package edu.umd.cs.findbugs.annotations does not exist
AnonymousSkipDefs.java:1: error: package org.eclipse.jdt.annotation does not exist
AliasedAnnotations.java:37: error: package edu.umd.cs.findbugs.annotations does not exist
AliasedAnnotations.java:62: error: package org.eclipse.jdt.annotation does not exist
AliasedAnnotations.java:67: error: package org.jetbrains.annotations does not exist
AliasedAnnotations.java:77: error: package org.netbeans.api.annotations.common does not exist
```

smillst:
One way to fix the tests is to move the third-party annotations to tests/src. So the annotations that were in checker/src should go in checker/tests/src and  framework/src should go in framework/tests/src.

cushon:
I was also wondering about adding dependencies on the libraries that provide those annotations to the tests, but if your preference is to move the existing stubs to `/tests/` that sounds good to me.

smillst:
Yes, the right way is to add them as test dependencies; you could do that.  But, we don't currently have any test-only jar dependencies, so that would have to be added.

cushon:
Thanks - I moved the aliases to `{checker,framework}/tests/src/testannotations`, and `ant tests` passes locally without any build changes.

wmdietl:
https://github.com/typetools/checker-framework-inference/commit/c5e301f4df10f2858fac3f068da4e291d42d4cc0 should fix the downstream failure. I've restarted the build.

Issue #341, opened 2015-07-03
Control flow graph construction is not cognizant of Checker Framework types


GoogleCodeExporter:
```
In the following example, shouldn't 'result' definitely be initialized to a 
non-null Object before it is returned?

Perhaps the flow analysis is allowing for 'p' to be null? In that case p.get() 
would throw a NPE and 'result' wouldn't get initialized.

# Steps to reproduce

=== Test.java ===
class Test {
  interface Provider {
    Object get();
  }
  Object execute(Provider p) {
    final Object result;
    try {
      result = p.get();
    } finally {
    }
    return result;
  }
}
===

$ ~/jsr308/checker-framework-1.8.3/checker/bin/javac -processor 
org.checkerframework.checker.nullness.NullnessChecker -cp 
~/jsr308/checker-framework-1.8.3/checker/dist/checker-qual.jar Test.java
Test.java:13: error: [return.type.incompatible] incompatible types in return.
    return result;
           ^
  found   : @UnknownInitialization @Nullable Object
  required: @Initialized @NonNull Object
1 error
```

Original issue reported on code.google.com by `cus...@google.com` on 15 Jul 2014 at 11:37

mernst:
I added a test case in file checker-framework/checker/tests/nullness/flow/Issue341.java .

wmdietl:
Fixed by https://github.com/typetools/checker-framework/commit/04be2b98f5d19eab4b5288f55b0a3c600beffd90

PR #1826, opened 2018-02-08
Fix #293, #341, #548 by duplicating finally trees

### LINKED ISSUE #293: Incorrect handling of 'finally'?

### LINKED ISSUE #1504: Dataflow graph for code with try-finally failing on JDK 9 javac due to a moved field
We updated NullAway to use Checker Framework 2.2.0 (to pick up #1483), but in later testing ran into uber/NullAway#6, a crash on code with try-finally with this stack trace:

```
Caused by: java.lang.NoSuchFieldError: operator
  	at org.checkerframework.javacutil.trees.FullyTreeCopier.visitBinary(FullyTreeCopier.java:67)
  	at org.checkerframework.javacutil.trees.FullyTreeCopier.visitBinary(FullyTreeCopier.java:40)
  	at com.sun.tools.javac.tree.JCTree$JCBinary.accept(JCTree.java:1993)
  	at com.sun.tools.javac.tree.TreeCopier.copy(TreeCopier.java:61)
  	at org.checkerframework.javacutil.trees.FullyTreeCopier.copy(FullyTreeCopier.java:48)
  	at org.checkerframework.javacutil.trees.FullyTreeCopier.copy(FullyTreeCopier.java:40)
  	at com.sun.tools.javac.tree.TreeCopier.visitParenthesized(TreeCopier.java:333)
  	at com.sun.tools.javac.tree.TreeCopier.visitParenthesized(TreeCopier.java:45)
  	at com.sun.tools.javac.tree.JCTree$JCParens.accept(JCTree.java:1849)
```

I think the issue is that in JDK 8 javac, the `JCBinary` class declares the `operator` field [directly](https://github.com/loverdos/javac-openjdk7/blob/master/src/main/java/openjdk7/com/sun/tools/javac/tree/JCTree.java#L1554), whereas [in JDK 9 javac](https://github.com/google/error-prone-javac/blob/upstream/src/jdk.compiler/share/classes/com/sun/tools/javac/tree/JCTree.java#L1970), the field has moved to [a new `JCOperatorExpression` class](https://github.com/google/error-prone-javac/blob/upstream/src/jdk.compiler/share/classes/com/sun/tools/javac/tree/JCTree.java#L1886).  We need to run on JDK 9 javac since NullAway is an Error Prone plugin check.

Not sure of the best way to fix this, but given that JDK 9 is coming out soon, it'd be great if this code could be written in a way that works on both JDK 8 and JDK 9 javac.

wmdietl:
Fix #293, #341, and #548 by duplicating finally blocks.
Also, fixes #1504.
Key changes are in `dataflow/src/org/checkerframework/dataflow/cfg/CFGBuilder.java`.
Other dataflow changes adapt the API to return `Set<Node>` for `Tree`s.
Small changes to framework and checkers, to adapt to this new API.

Return, break, continue, break & continue with label CFGs are now correct.


checker/tests/nullness/Issue1797.java
```diff
@@ -3,6 +3,7 @@
 
 import org.checkerframework.checker.nullness.qual.Nullable;
 
+// @skip-test
```
wmdietl: The `return` cases work, but the `break` and `continue` unfortunately don't. Their CFGs look good to me, but something in the lookup doesn't work.

wmdietl:
@lazaroclapp @msridhar Can you have a look at this and let me know how it works for you?

lazaroclapp:
Hey @wmdietl,

The good news is that this patch does seem to fix the main error we care about. See e.g. the test at line 227 here: https://github.com/uber/NullAway/blob/master/nullaway/src/test/resources/com/uber/nullaway/testdata/NullAwayTryFinallyCases.java (it also fixes a few other cases there, making try/finally behavior strictly more correct without a code change on our part so 👍 )

Unfortunately, it also appears to break something related to lambda expressions with expression bodies. For NullAway, it causes false positives on these lines:
```
    NonNullParamFunction n2 = (@Nullable Object x) -> (x == null) ? "null" : x.toString();
    NullableParamFunction n3 = (@Nullable Object x) -> (x == null) ? "null" : x.toString();
    NullableParamFunction n4 = (x) -> (x == null) ? "null" : x.toString();
```
Here: https://github.com/uber/NullAway/blob/master/nullaway/src/test/resources/com/uber/nullaway/testdata/NullAwayJava8NegativeCases.java

It says all of these can return null somehow, even though we assume `toString()` returns non-null. I am still digging into it a bit more. One possibility is that, seeing a body without a `return`, we are inserting/assuming `return null` somehow?

Neither `2.3.1` (the version NullAway is known to work with), nor `master` have this problem, only this branch does. Will let you know when I understand this a bit more.

lazaroclapp:
I looked a bit more into this, it seems like `treeLookup` is mapping those lambda's bodies to two separate nodes (despite no `try` being involved) and one looks sort of strange.

I added this code for debugging:

```
--- a/dataflow/src/org/checkerframework/dataflow/analysis/AnalysisResult.java
+++ b/dataflow/src/org/checkerframework/dataflow/analysis/AnalysisResult.java
@@ -114,6 +114,11 @@ public class AnalysisResult<A extends AbstractValue<A>, S extends Store<S>> {
      */
     public /*@Nullable*/ A getValue(Tree t) {
         Set<Node> nodes = treeLookup.get(t);
+        if(t.getKind().equals(Tree.Kind.CONDITIONAL_EXPRESSION)) {
+            System.err.println("Tree: " + t);
+            System.err.println("Nodes: " + nodes);
+            return getValue(nodes.iterator().next());
+        }
```

This makes those particular examples pass (of course, it is not general), but the most interesting part is that it prints something like this:

```
Tree: (x == null) ? "null" : x.toString()
Nodes: [((x == null) ? "null" : x.toString()), -> ((x == null) ? "null" : x.toString())]
```

No idea what that second `-> ((x == null) ? "null" : x.toString())` node is (note the arrow), but our analysis thinks it is nullable. Am I missing something about how the new API is supposed to work?

wmdietl:
I'm not running into issues with lambdas and the tests you show work correctly with the Nullness Checker.

I think this code generates this: https://github.com/typetools/checker-framework/blob/master/dataflow/src/org/checkerframework/dataflow/cfg/CFGBuilder.java#L1628

```java
            // If we are building the CFG for a lambda with a single expression as the body, then add an extra node
            // for the result of that lambda
            if (underlyingAST.getKind() == UnderlyingAST.Kind.LAMBDA) {
                LambdaExpressionTree lambdaTree =
                        ((UnderlyingAST.CFGLambda) underlyingAST).getLambdaTree();
                if (lambdaTree.getBodyKind() == LambdaExpressionTree.BodyKind.EXPRESSION) {
                    Node resultNode =
                            new LambdaResultExpressionNode(
                                    (ExpressionTree) lambdaTree.getBody(),
                                    finalNode,
                                    env.getTypeUtils());
                    extendWithNode(resultNode);
                }
            }
```

The `(ExpressionTree) lambdaTree.getBody()` tree is the `CONDITIONAL_EXPRESSION` tree, so you would have one Node from the original analysis and another Node from here.
Previously, I guess only the `LambdaResultExpressionNode` remained in `treeLookup`.

Maybe you can slightly tweak your analysis that it gives the same value to these two Nodes?
But maybe in this particular case we don't really want the two Nodes?

lazaroclapp:
Hey Werner, a few things (note that point 2 is the important one):

1. Thanks for that explanation! I think what was happening was that somehow only the underlying node ended up inside `treeLookup` at the end (rather than the `LambdaResultExpressionNode`, which was the one with the spurious `NULLABLE` value). But we were still seeing the later during actual dataflow were we cared about it, just not while checking the final results. Either way, we just fixed the nullness propagation for `LambdaResultExpressionNode` and this now works for us. All NullAway unit tests pass with this change and try/finally behavior is actually improved, which is good news!
2. However, once actually testing this CF branch on our full codebase, we are getting an NPE inside `CFGBuilder` here:
```
Caused by: java.lang.NullPointerException
  	at org.checkerframework.dataflow.cfg.CFGBuilder$CFGTranslationPhaseTwo.process(CFGBuilder.java:1309)
  	at org.checkerframework.dataflow.cfg.CFGBuilder.run(CFGBuilder.java:340
```

A minimized test case that triggers the crash for us:
```
  private boolean nestedCFGConstructionTest(Object o) {
    boolean result = true;
    java.io.BufferedWriter out = null;
    try {
      try {
      } finally {
        out = new java.io.BufferedWriter(new java.io.OutputStreamWriter(System.err));
      }
      if (o != null) {
        out.write(' ');
      }
    } catch (Exception e) {
    } finally {
    }
    return result;
  }
```
3. I don't think for this case it will make a difference to test it with the CF nullness checker, as opposed to NullAway. In general, however, I am not sure how to build `checkers.jar` from this branch, `ant dist` in the corresponding directory fails because of tests from the various broken checkers. Is there a better way to test things with the built in checkers in this branch?

wmdietl:
Thanks for the new test case. I'll make sure it works. I'll try to get all tests working again soon.

You can run `ant dist-nobuildjdk` in the main directory to just build the jar files. That should be good enough to run `checker/bin/javac` on examples. `ant dist` doesn't run any tests.

wmdietl:
@lazaroclapp @msridhar I think this works nicely now.
Would you mind checking whether I missed any cases? Is there any performance/memory impact?
I'll make another pass through the Javadoc, but otherwise I think this is ready.

lazaroclapp:
@wmdietl Thanks so much! Unfortunately, in checking our full (instrumented) codebase, we are running into at least one other case that throws an NPE during CFG construction. Same stack trace as the last one, but here the minimized test case is as follows:

```
private boolean nestedCFGConstructionTest2() throws IOException {
    java.io.BufferedWriter out = new java.io.BufferedWriter(new java.io.OutputStreamWriter(System.err));
    try {
      try {
        return true;
      } finally {
      }
    } finally {
      out.write(' ');
      out.close();
    }
  }
```

It seems all the following are necessary to trigger the crash: the two nested try statements, the return nested inside the second try, and the method call with a checked exception.

wmdietl:
@lazaroclapp Another attempt :-) Your test works, previous tests for #548 work, so hopefully things are good now.
The Checker Framework build currently fails because of other issues.
Could you give it another swing and give me feedback?

lazaroclapp:
@wmdietl It seems to work now, with everything I can think of throwing at it (our own unit tests, un-instrumented code, instrumented code, etc)! Performance seems fine too. Whenever you are ready, feel free to merge. I think we will backport this into our branch until 2.3.3 is out.  Thanks so much!

wmdietl:
@smillst I finished the javadoc updates. This is ready for review.

wmdietl:
Add a `getFirstNonLambdaExpressionNodeForTree` method to ATF.

wmdietl:
@smillst The changes look nicer with the helper method. Let's just try to not use that new method.

PR #1829, opened 2018-02-08
Allow subtyping between wildcards with different underlying types

### LINKED ISSUE #1818: change in lambda parameter type inference in 2.3.2
The following example is accepted by 2.3.1, and rejected by 2.3.2. Is the change deliberate?

```java
import java.util.List;
import java.util.function.Consumer;

class Test {
  void f() {
    Consumer<List<?>> c = values -> values.forEach(value -> g(value));
  }

  void g(@Nullable Object o) {}
}
```

```
$ ./checker-framework-2.3.1/checker/bin/javac -AprintErrorStack -processor nullness -AsuppressWarnings=keyfor Test.java
... OK
$ ./checker-framework-2.3.2/checker/bin/javac -AprintErrorStack -processor nullness -AsuppressWarnings=keyfor Test.java
Test.java:7: error: [lambda.param.type.incompatible] incompatible parameter types in lambda expression.
    Consumer<List<?>> c = values -> values.forEach(value -> g(value));
                                                   ^
  found   : ? extends @Initialized @Nullable Object
  required: capture#16 of ? extends @Initialized @Nullable Object
1 error
```

smillst:
Fixes #1818.


framework/src/org/checkerframework/framework/type/DefaultTypeHierarchy.java
```diff
@@ -1084,7 +1084,7 @@ protected boolean visitWildcardSubtype(
             }
         }
 
-        if (subtype.getUnderlyingType() == superTypeMirror) {
+        if (superTypeMirror.getKind() == TypeKind.WILDCARD) {
```
wmdietl: Can you update the documentation in the branch to reflect this change?
What changed in 2.3.2 that makes this the right fix?
Some such comments might make the logic better to follow.

PR #1830, opened 2018-02-09
Issue type.invalid.conflicting.anno for errors related to conflicting annos

### LINKED ISSUE #194: Validity check for AnnotatedTypeMirrors incomplete
```
The method AnnotatedTypes.isValidType is supposed to check the validity of an 
AnnotatedTypeMirror (e.g. that the number of annotations is correct, etc.).  
For certain types it is necessary to apply these checks recursively (e.g. for 
arrays).

At the moment, the recursion for type arguments is not done, because it causes 
various tests to fail.  The failures are not related to type validity, but 
rather a variety of other type-checking errors occur (such as 
argument.type.incompatible or fields.uninitialized) that did not occur before.  
Without having spent too much time debugging this issue, it seems that there 
might be an issue with lazy initialization, such that the types change because 
they are "looked at" in a different order with this new validity check.

For now, the code performing this recursive check for type arguments is 
commented out (see the end of method AnnotatedTypes.isValidType).  However, it 
would be nice to investigate the cause, and enable the checks.
```

Original issue reported on code.google.com by `stefanheule@gmail.com` on 11 Jan 2013 at 2:25

smillst:
Always issue a `type.invalid.conflicting.anno` error when there are too many or two few annotations.

Fixes #194.


framework/src/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -3506,7 +3506,11 @@ public boolean validateType(Tree tree, AnnotatedTypeMirror type) {
         // basic consistency checks
         if (!AnnotatedTypes.isValidType(atypeFactory.getQualifierHierarchy(), type)) {
             checker.report(
-                    Result.failure("type.invalid", type.getAnnotations(), type.toString()), tree);
```
wmdietl: Is this changed error key really an improvement? This error gets raised, e.g. when too few annotations are present. Is the "conflicting" message really clearer than the "type.invalid" message?
How about keeping `type.invalid` here and adding an issue to give a more precise error message, like 'type.invalid.too.few.annotations`, 'type.invalid.too.many.annotations`, etc.
The logic from `checkConflictingPrimaryAnnos` could then be used to raise `type.invalid.conflicting.annos`.
(`too.many` is probably the same as `conflicting`?)

framework/src/org/checkerframework/framework/util/AnnotatedTypes.java
```diff
@@ -884,26 +885,34 @@ public static boolean hasTypeQualifierElementTypes(ElementType[] elements, Class
      *       as bounds of type variables and wildcards.
      * </ol>
      */
-    public static boolean isValidType(
+    public static Result isValidType(
```
wmdietl: How about keeping the return type here `boolean` and doing:

```java
Result isValid = scanner.visit(type);
return isValid.isSuccess();
```

That would hide this implementation detail and wouldn't require the checker-framework-inference change.

framework/src/org/checkerframework/common/basetype/messages.properties
```diff
@@ -18,6 +18,8 @@ bound.type.incompatible=incompatible bounds in %s\ntype: %s\nupper bound: %s\nlo
 monotonic.type.incompatible=cannot assign %s to %s (monotonic type).\nfull type found: %s
 type.invalid=invalid type: annotations %s in type "%s"
 type.invalid.conflicting.annos=invalid type: conflicting annotations %s in type "%s"
+type.invalid.too.many.annotations=invalid type: too many annotations %s in type "%s"
```
wmdietl: This key doesn't seem to be used. Remove it? Or document that it's for future extension?

framework/src/org/checkerframework/common/basetype/messages.properties
```diff
@@ -18,6 +18,8 @@ bound.type.incompatible=incompatible bounds in %s\ntype: %s\nupper bound: %s\nlo
...
+type.invalid.too.few.annotations=invalid type: missing annotations %s in type "%s"
```
wmdietl: At the moment the given annotations are output. So this would be clearer if either the message is changed or the provided value is the top annotations from the missing hierarchies.

framework/src/org/checkerframework/framework/util/AnnotatedTypes.java
wmdietl: Ah, sorry. I ignored that the caller uses the error code in `checker.report(isValid, tree);`.
Would it be ugly to report the error here?

wmdietl: The `BaseTypeValidator` also directly reports errors, so I think this would be OK. Maybe add a note to the javadoc of this method.

smillst: If the error is reported here, then checker-framework-inference would report the error.

wmdietl: Good point.
Could we just move all of `AnnotatedTypes.isValidType` into the `TypeValidator`/`BaseTypeValidator`?
Logically, they belong together so putting one validity check into a static method seems wrong.
Then the `BaseTypeVisitor` code above would simply become:

```java
     public boolean validateType(Tree tree, AnnotatedTypeMirror type) {
         return typeValidator.isValid(type, tree);
     }
```

And inference already has it's own `TypeValidator`.

PR #1833, opened 2018-02-12
Classpath command added in annotating-libraries.tex

ashishrana160796:


mernst:
What is "clearance"?
Also, the build is failing.

ashishrana160796:
I think it will be better to have _CLASSPATH_ command just before `astub command`  for more ease so as to build stub files. Let me see the build fail issue.

mernst:
Does the command as given fail without the CLASSPATH environment variable set?

ashishrana160796:
Here is the following that came with me while running that command:
```
java org.checkerframework.framework.stub.StubGenerator javax.net.ssl > Ssl.astub
Error: Could not find or load main class org.checkerframework.framework.stub.StubGenerator
```
And I have followed the simple installation steps mentioned section 30.1 not builded checker from source.

mernst:
OK, if the command issued an error, that's a bug in the manual -- not just "for more ease of use".  It should be fixed.

ashishrana160796:
Can you please point out for possible causes for build failure in this case as I have just added single line in `.tex` file.

docs/manual/annotating-libraries.tex
```diff
@@ -572,6 +572,7 @@ \subsubsection{If you do not have access to the Java source code\label{stub-crea
   must be on your classpath.)
 
 \begin{Verbatim}
+  export CLASSPATH=/home/user_name/checker-framework-2.3.2/checker/dist/checker.jar:javac.jar
```
wmdietl: You should use `$JSR308` to refer to the installation path:

`export CLASSPATH=$JSR308/checker-framework/checker/dist/checker.jar`

The `javac.jar` shouldn't be necessary, as javac is contained in `checker.jar`. It would also need a full path to work.

docs/manual/annotating-libraries.tex
```diff
@@ -568,10 +568,10 @@ \subsubsection{If you do not have access to the Java source code\label{stub-crea
...
 
 \item
-  Create a stub file by running the stub class generator.  (\<checker.jar> and \<javac.jar>
-  must be on your classpath.)
+  Create a stub file by running the stub class generator.  (\<checker.jar> must be on your classpath.)
 
 \begin{Verbatim}
+  export CLASSPATH=export CLASSPATH=$JSR308/checker-framework/checker/dist/checker.jar
   cd nullness-stub
   java org.checkerframework.framework.stub.StubGenerator java.lang.String > String.astub
```
wmdietl: Can you remove setting the `CLASSPATH` variable (which incorrectly includes two `exports`) and change this to:

```
cd nullness-stub
java -cp $JSR308/checker-framework/checker/dist/checker.jar org.checkerframework.framework.stub.StubGenerator java.lang.String > String.astub
```

Using `-cp` instead of setting `CLASSPATH` is a more local change and achieves the same thing.

ashishrana160796: Thanks for correcting it.

PR #1836, opened 2018-02-14
Add postcondition annotations for index and value qualifiers with arguments

panacekcz:
As a follow-up to #1723, this PR adds new postcondition annotations for qualifiers with arguments:
- `@EnsuresMinLenIf`
- `@EnsuresLTLengthOf`
- `@EnsuresLTLengthOfIf`


panacekcz:
@kelloggm, I added a test and descriptions to the manual. I didn't link to a section on postconditions in general because I didn't find one. The closest would be "Side effects, determinism, purity, and flow-sensitive analysis" but that's not particularly relevant to the annotaitons.

kelloggm:
@panacekcz The lack of a section on postcondition annotations seems to be an oversight, but if there is nothing to link to, then we can't link to it. Can you file an issue that requests one be written?

Issue #1838, opened 2018-02-15
Undetected NPE in method with parameter type "List<? extends List<Object>>"


sebright:
This is a simplified example of an issue that appeared in [opencensus-java](https://github.com/census-instrumentation/opencensus-java) (https://github.com/census-instrumentation/opencensus-java/pull/998#issuecomment-366015863) with Checker Framework version 2.3.2.

The Nullness Checker passes, but the program throws a NPE when it calls `element.toString()` on the null element that was inserted in `f()`.

```java
import java.util.ArrayList;
import java.util.List;
import org.checkerframework.checker.nullness.qual.Nullable;

public class Example {
  public static void main(String[] args) {
    f();
  }

  public static void f() {
    List<@Nullable Object> list = new ArrayList<@Nullable Object>();
    list.add(null);
    List<List<@Nullable Object>> listList = new ArrayList<List<@Nullable Object>>();
    listList.add(list);
    processElements(listList);
  }

  private static void processElements(List<? extends List<Object>> listList) {
    for (List<Object> list : listList) {
      for (Object element : list) {
        element.toString();
      }
    }
  }
}
```

This also passes in the live demo ([link](http://eisop.uwaterloo.ca/live/#mode=display&code=import+java.util.ArrayList%3B%0Aimport+java.util.List%3B%0Aimport+org.checkerframework.checker.nullness.qual.Nullable%3B%0A%0Apublic+class+Example+%7B%0A++public+static+void+main(String%5B%5D+args)+%7B%0A++++f()%3B%0A++%7D%0A%0A++public+static+void+f()+%7B%0A++++List%3C%40Nullable+Object%3E+list+%3D+new+ArrayList%3C%40Nullable+Object%3E()%3B%0A++++list.add(null)%3B%0A++++List%3CList%3C%40Nullable+Object%3E%3E+listList+%3D+new+ArrayList%3CList%3C%40Nullable+Object%3E%3E()%3B%0A++++listList.add(list)%3B%0A++++processElements(listList)%3B%0A++%7D%0A%0A++private+static+void+processElements(List%3C%3F+extends+List%3CObject%3E%3E+listList)+%7B%0A++++for+(List%3CObject%3E+list+%3A+listList)+%7B%0A++++++for+(Object+element+%3A+list)+%7B%0A++++++++element.toString()%3B%0A++++++%7D%0A++++%7D%0A++%7D%0A%7D&typeSystem=nullness)).

[...]

PR #1841, opened 2018-02-19
Work on fixing #1838

### LINKED ISSUE #849: Generic types in wildcard bounds
Generic types in wildcard bounds are not checked during common assignment check. For example, neither of the expected errors are issued below:

```
class Gen<G> { }
void polyAll(Gen<Gen<@H1S2 Object>> genGenNonNull) {
    Gen<@H1Top ? extends @H1Top Gen<@H1Top Object>> a = genGenNonNull;
}

void interning(Gen<Gen<@Interned Object>> genGenNonNull) {
    Gen<@UnknownInterned ? extends @UnknownInterned Gen<@UnknownInterned Object>> a = genGenNonNull;
}
```

However, the Nullness Checker does correctly issue the error:

```
 void nullness(Gen<Gen<@NonNull Object>> genGenNonNull) {
    Gen<@Nullable ? extends @Nullable Gen<@Nullable Object>> a = genGenNonNull;
}
```

wmdietl:
Fix #849 and #1838 by correctly handling generics in subtype tests.
Refactors interactions between the type hierarchies and the visitor history.


wmdietl:
@smillst Could you have a look at this? Some tests now fail with the KeyFor hierarchy. I think many of those were masked previously, because as soon as one hierarchy was OK, all following hierarchies were ignored. I think the tests that I skipped are all caused by problems with KeyFor.
Maybe there is something about the KeyFor hierarchy or defaults that needs adapting?
We can go through the other changes in the meeting tomorrow.

checker/tests/nullness/generics/MyMap.java
```diff
@@ -1,6 +1,8 @@
 import java.util.Map;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
+// @skip-test issue with @KeyFor
```
smillst: checker/tests/nullness/generics/MyMap.java:16: error: [enhancedfor.type.incompatible] incompatible types in enhanced for loop.
```
  found   : @UnknownKeyFor Entry<@KeyFor("map") ?[ extends K[ extends @KeyFor("map") Object super @KeyFor("map") Void] super @KeyFor("map") Void], ?[ extends V[ extends @UnknownKeyFor Object super @KeyForBottom Void] super @KeyForBottom Void]>
  required: @UnknownKeyFor Entry<?[ extends K[ extends @UnknownKeyFor Object super @KeyForBottom Void] super @KeyForBottom Void], ?[ extends V[ extends @UnknownKeyFor Object super @KeyForBottom Void] super @KeyForBottom Void]>
```

framework/tests/all-systems/Issue1587b.java
```diff
@@ -25,6 +25,7 @@
         Iterable<R> q();
     }
 
+    @SuppressWarnings("keyfor:type.argument.type.incompatible")
```
smillst: This is only only remaining KeyFor regression.

smillst:
> Could you have a look at this? Some tests now fail with the KeyFor hierarchy. I think many of those were masked previously, because as soon as one hierarchy was OK, all following hierarchies were ignored.

The KeyFor Checker is a separate checker, so the failures weren't masked because of this.  They were masked because of KeyFor's DefaultTypeHierarchy overrides isSubtype(ATM, ATM).  I added the code to isSubtype(..., AnnotationMirror top) and that fixes all but one of the regressions.

We still need to figure out that last regression though.

framework/tests/all-systems/Issue1587b.java
```diff
@@ -25,6 +25,7 @@
         Iterable<R> q();
     }
 
+    @SuppressWarnings("type.argument.type.incompatible") // TODO: FIX
```
smillst: This is the only remaining regression.

framework/src/org/checkerframework/framework/type/DefaultRawnessComparer.java
```diff
@@ -253,50 +269,55 @@ public Boolean visitWildcardSupertype(
...
-        return false;
+        // Needed to prevent infinite recursion e.g. in
+        // tests/all-systems/java8inference/Issue1815.java
+        visited.add(subtype, supertype, currentTop, true);
+
+        // TODO: tests/all-systems/InferTypeArgsCondtionalExpression.java
+        // breaks with this:
+        //Boolean result = this.visit(subtypeUb, supertype, visited);
+        // it doesn't make sense to me to compare subtype and subtypeUB
+        Boolean result = this.visit(subtype, subtypeUb, visited);
```
wmdietl: Highlighting this comment and the test that fails when using a more meaningful check.

checker/tests/nullness/java8/Issue579.java
```diff
@@ -1,6 +1,8 @@
 // Test case for Issue579
 // https://github.com/typetools/checker-framework/issues/579
 
+// @skip-test caused by DefaultRawnessComparer.visitNull_Declared
```
wmdietl: Note the strange failure when using a meaningful implementation of DefaultRawnessComparer.visitNull_Declared.

wmdietl:
Thanks for your cleanups! I've done a few more things and highlighted two strange things in comments.

smillst:
I fixed the two things you commented one.  We still have the framework/tests/all-systems/Issue1587b.java regression to figure out.

framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java
```diff
@@ -1143,7 +1096,13 @@ protected boolean visitWildcardSubtype(
                 final AnnotatedTypeMirror sourceTypeArg = asSuperAdt.getTypeArguments().get(0);
                 final AnnotatedTypeMirror resultTypeArg = resultAtd.getTypeArguments().get(0);
                 resultTypeArg.clearAnnotations();
-                resultTypeArg.addAnnotations(sourceTypeArg.getAnnotations());
+                if (resultTypeArg.getKind() == TypeKind.TYPEVAR) {
```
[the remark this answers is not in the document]
smillst: This makes sense to me, too.  (Although, I don't really understand why Enums are special cased here in the first place.

PR #1843, opened 2018-02-20
Developer documentation

mernst:
This moves developer documentation together in `docs/developer`.
I also put the release scripts there -- let me know if that is not a good idea.


mernst:
@wmdietl My rationale was to keep the top-level directory less cluttered.  The release scripts are never of interest to anyone but developers.  They aren't documentation, but this put them near the documentation of them.

PR #1844, opened 2018-02-21
Avoid using dataflow nodes in checkers

smillst:
Change instances where nodes are used when TreeUtil methods or other means could be used instead.  (This clean up will make #1826 change less code.)

### DIFF
--- a/checker/src/org/checkerframework/checker/i18nformatter/I18nFormatterTreeUtil.java
+++ b/checker/src/org/checkerframework/checker/i18nformatter/I18nFormatterTreeUtil.java
@@ -200,15 +200,13 @@ public class I18nFormatterTreeUtil {
 
     /** Returns an I18nFormatCall instance, only if FormatFor is called. Otherwise, returns null. */
     public I18nFormatCall createFormatForCall(
-            MethodInvocationTree tree,
-            MethodInvocationNode node,
-            I18nFormatterAnnotatedTypeFactory atypeFactory) {
+            MethodInvocationTree tree, I18nFormatterAnnotatedTypeFactory atypeFactory) {
         ExecutableElement method = TreeUtils.elementFromUse(tree);
         AnnotatedExecutableType methodAnno = atypeFactory.getAnnotatedType(method);
         for (AnnotatedTypeMirror paramType : methodAnno.getParameterTypes()) {
             // find @FormatFor
             if (paramType.getAnnotation(I18nFormatFor.class) != null) {
-                return atypeFactory.treeUtil.new I18nFormatCall(tree, node, atypeFactory);
+                return atypeFactory.treeUtil.new I18nFormatCall(tree, atypeFactory);
             }
         }
         return null;
@@ -229,17 +227,14 @@ public class I18nFormatterTreeUtil {
 
         private AnnotatedTypeMirror formatAnno;
 
-        public I18nFormatCall(
-                MethodInvocationTree tree,
-                MethodInvocationNode node,
-                AnnotatedTypeFactory atypeFactory) {
+        public I18nFormatCall(MethodInvocationTree tree, AnnotatedTypeFactory atypeFactory) {
             this.tree = tree;
             this.atypeFactory = atypeFactory;
             List<? extends ExpressionTree> theargs = (tree).getArguments();
             this.args = null;
             ExecutableElement method = TreeUtils.elementFromUse(tree);
             AnnotatedExecutableType methodAnno = atypeFactory.getAnnotatedType(method);
-            initialCheck(theargs, method, node, methodAnno);
+            initialCheck(theargs, method, tree, methodAnno);
         }
 
         @Override
@@ -254,7 +249,7 @@ public class I18nFormatterTreeUtil {
         private void initialCheck(
                 List<? extends ExpressionTree> theargs,
                 ExecutableElement method,
-                MethodInvocationNode node,
+                MethodInvocationTree node,
                 AnnotatedExecutableType methodAnno) {
             int paramIndex = -1;
             Receiver paramArg = null;
--- a/checker/src/org/checkerframework/checker/i18nformatter/I18nFormatterTreeUtil.java
+++ b/checker/src/org/checkerframework/checker/i18nformatter/I18nFormatterTreeUtil.java
@@ -255,43 +255,41 @@ public class I18nFormatterTreeUtil {
             Receiver paramArg = null;
             int i = 0;
             for (AnnotatedTypeMirror paramType : methodAnno.getParameterTypes()) {
-                if (paramType.getAnnotation(I18nFormatFor.class) != null) {
-                    this.formatArg = theargs.get(i);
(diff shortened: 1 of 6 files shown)


checker/src/org/checkerframework/checker/lock/LockVisitor.java
```diff
@@ -1091,10 +1089,12 @@ private boolean isTreeSymbolEffectivelyFinalOrUnmodifiable(Tree tree) {
...
     @Override
     public Void visitIdentifier(IdentifierTree tree, Void p) {
-        Node node = atypeFactory.getNodeForTree(tree);
-        if (node instanceof FieldAccessNode) {
-            Node receiverNode = ((FieldAccessNode) node).getReceiver();
-            if (receiverNode instanceof ImplicitThisLiteralNode) {
+        if (TreeUtils.isFieldAccess(tree)) {
+            Tree parent = getCurrentPath().getParentPath().getLeaf();
+            if (parent.getKind() != Kind.MEMBER_SELECT
+                    || ((MemberSelectTree) parent).getExpression() == tree) {
```
wmdietl: Instead of `getExpression` shouldn't this be `getIdentifier()` and then some call on `tree` to get its `Name`?
I'm not quite understanding the logic otherwise and a comment would be helpful.

framework/src/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -1202,6 +1204,36 @@ public static FlowExpressionContext buildContextForMethodUse(
             return flowExprContext;
         }
 
+        /**
+         * @return a {@link FlowExpressionContext} for the method {@code methodInvocation}
+         *     (represented as a {@link MethodInvocationTree} as seen at the method use (i.e., at a
+         *     method call site).
```
wmdietl: Would cross-references between the Node and Tree versions of this method be useful?

checker/src/org/checkerframework/checker/i18nformatter/I18nFormatterTreeUtil.java
```diff
@@ -254,49 +249,47 @@ public String toString() {
         private void initialCheck(
                 List<? extends ExpressionTree> theargs,
                 ExecutableElement method,
-                MethodInvocationNode node,
+                MethodInvocationTree node,
```
wmdietl: Can you also rename the parameter to `tree`.

checker/src/org/checkerframework/checker/i18nformatter/I18nFormatterTreeUtil.java
```diff
@@ -254,49 +249,47 @@ public String toString() {
...
+                    // report errors here
+                    checker.report(
+                            org.checkerframework.framework.source.Result.failure(
+                                    "i18nformat.invalid.formatfor"),
+                            tree);
+                    return;
+                }
+                if (formatforArg.startsWith("#")) {
+                    // formatArg should be #Number, with 1-indexing.
+                    paramIndex = Integer.valueOf(formatforArg.substring(1)) - 1;
```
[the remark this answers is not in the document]
smillst: No, even if it's a valid int, it might be greater than the number of parameters.

framework/src/org/checkerframework/framework/util/FlowExpressionParseUtil.java
smillst: I don't think so.

checker/src/org/checkerframework/checker/i18nformatter/I18nFormatterTreeUtil.java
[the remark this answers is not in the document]
wmdietl: I don't understand. What is the advantage of this over the old version?
Why not assign the result of `parse` to `paramArg` and then calculate `paramIndex` as was done before?
At least some documentation would be needed to highlight this.

framework/src/org/checkerframework/framework/util/FlowExpressionParseUtil.java
wmdietl: Why is the implementation of `buildContextForMethodUse(MethodInvocationNode methodInvocation, BaseContext checkerContext)` above not simply

```
    return buildContextForMethodUse(methodInvocation.getTree(), checkerContext);
```
?

Are the two methods calculating slightly different things? Then that should be documented.
Is this duplication more efficient? Then please document it.
If the duplication isn't intended, let's remove it.

checker/src/org/checkerframework/checker/i18nformatter/I18nFormatterTreeUtil.java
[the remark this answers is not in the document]
smillst: The old way was incorrect because two arguments might be equal, such as the literal one.

framework/src/org/checkerframework/framework/util/FlowExpressionParseUtil.java
smillst: methodInvocation.getTree() might return null.   Every single method that creates a `FlowExpressionContext` has one version for Trees and one for Nodes.

checker/src/org/checkerframework/checker/i18nformatter/I18nFormatterTreeUtil.java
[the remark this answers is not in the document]
smillst: I really don't understand what you want to documentation to be.

framework/src/org/checkerframework/framework/util/FlowExpressionParseUtil.java
wmdietl: What methods do you mean? For example for the similar `buildContextForNewClassUse` I only see a version that takes a Node.
Wouldn't it be nice to remove that duplication?

checker/src/org/checkerframework/checker/i18nformatter/I18nFormatterTreeUtil.java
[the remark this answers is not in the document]
smillst: The current code allows any FlowExpression.  Your example code does not.  I don't know the code well enough to know if restricting to just parameter references is ok.  I'll move this to a different pr.

framework/src/org/checkerframework/framework/util/FlowExpressionParseUtil.java
smillst: Again, we can't remove the duplication because node.getTree() might return null.  So, we need both methods.

wmdietl: Ok, but which methods do you mean when you say `Every single method that creates a FlowExpressionContext has one version for Trees and one for Nodes.`?
I'm trying to understand whether this duplication is already happening or not.

smillst: I see the other methods all use trees, except for `buildContextForNewClassUse`.   I was thinking of the `internalReprOf` methods in FlowExpressions.

PR #1848, opened 2018-02-23
[WIP] Convert the build scripts from Ant to Gradle.

smillst:
I tried to make the Gradle tasks do the same thing as the Ant tasks did.  However, since we are ending support for JSR308, I removed all references to it in the build scripts except for the jtreg tests, which is explained in a comment in that task.  CheckerMain still uses it though because plume-lib still uses annotations in comments and because we haven't removed mentions of annotations in comments from the manual.

Also, framework now depends on checker-qual, so all annotations in comments have been removed.

I also haven't documented the new build system in the manual which must be done before merging. 

You should use the Gradle wrapper rather than installing gradle locally.  Use `gradlew` in the checker-framework directory.

If you run a task from the main directory then, it will run that task in all subprojects with a tasks by that name.  So, if run `./gradlew reformat` that reformats all Java files in everywhere.  But `(cd framework && ../gradlew reformat` only reformats Java files in framework.  Alternatively, calling `:framework:reformat` from the main directory or any subproject has the same effect.

Here are some tasks that I use often. (I think this should go in the manual, but maybe it's too many?)
* `assemble`: builds all jars except jdk8.jar,  including javadoc jars and jars contain all dependencies. 
* `build`: everything `assemble` does plus runs all JUnit tests.
* `buildJdk`: builds jdk8.jar
* `allTests`: runs all tests.
* `:framework:allTests`: runs all framework tests.  (Or `(cd framework && ../gradlew allTests)`)
* `reformat`: reformats Java files.  If it is run in a subproject, then it will only reformat the java files in that project.
* `NameOfJUnitTest`: runs the JUnit test with that name `NullnessFbcTest` for example.
* `installCheckerJar`: Puts all jars in checker/dist required by `CheckerMain`.  So if you are making changes to the Checker Framework then testing them using `checker/bin/java` this task will update all required jars.  (Better name for this?)


mernst:
Should annotations in comments be removed from the manual as part of this pull request, or a different pull request early next week?

smillst:
> Should annotations in comments be removed from the manual as part of this pull request, or a different pull request early next week?

This pull request does not end support for annotations in comments.  So, it should be in a different pull request.

smillst:
We decided to break this into two pull requests.  One that removes annotations in comments from Checker Framework source and one that adds the Gradle build files.

PR #1849, opened 2018-02-24
java.security.PrivilegedAction.java added 

ashishrana160796:
`java.security.PrivilegedAction.java` is added to redefine behavior of method `<T> run()`  which has to class dependent according to specification but is behaving as @NonNull type and causing unwanted return type errors :
```
[override.return.invalid] Incompatible return type.
                            public @Nullable Field run() {
```
Here is example code that is using this method and return @Nullable value:

 ```
new java.security.PrivilegedAction<Field>() {
                            public @Nullable Field run() {
                                Field field = null;
                                try {
                                    field = InputEvent.class.
                                        getDeclaredField("canAccessSystemClipboard");
                                    field.setAccessible(true);
                                    return field;
                                } catch (SecurityException e) {
                                    // if (log.isLoggable(PlatformLogger.Level.FINE)) {
                                        // log.fine("AWTEvent.get_InputEvent_CanAccessSystemClipboard() got SecurityException ", e);
                                    // }
                                } catch (NoSuchFieldException e) {
                                    // if (log.isLoggable(PlatformLogger.Level.FINE)) {
                                        // log.fine("AWTEvent.get_InputEvent_CanAccessSystemClipboard() got NoSuchFieldException ", e);
                                    // }
                                }
                                return null;
                            }
                        }
```
Update on Code example having both problems, pull request #1849 and #1850 in one code :
```
new PrivilegedAction<Constructor>() {
            public Constructor run() {
                try {
                    Constructor ctor = clazz.getDeclaredConstructor((Class[]) null);
                    if (ctor != null) {
                        ctor.setAccessible(true);
                    }
                    return ctor;
                } catch (SecurityException e) {
                } catch (NoSuchMethodException e) {
                }
                return null;
            }
        }
``` 
---
See : Last Commit e7502a4

### CODE REVIEW

**Comment by mernst** [comment]
I'm not sure why you used `@PolyNull` instead of `@Nullable`.
This section of the manual explains that it is usually undesirable to write a single `@PolyNull` annotation:
https://checkerframework.org/manual/#qualifier-polymorphism

**Comment by wmdietl** [comment]
Why do we need this annotation at all?
Can't your example be simply written as:

```java
new java.security.PrivilegedAction<@Nullable Field>() {
                           public @Nullable Field run() {
                               ...
                               return null;
                           }
                       }
```

That is, simply use the correct type argument to the class and the return type follows.
You should only annotate a type variable if the type must vary from the type parameter declaration. Is that the case for this method?

ashishrana160796:
@mernst annotations are added to type parameter and return type and commit has been made.

wmdietl:
Can you give a test case that works with these new annotations that doesn't work with the original code?
I think neither `@Nullable` is necessary.

ashishrana160796:
It looks like code will work on without these annotations just fine. With proper annotations of code in  your comment annotating these `@Nullable` annotations does become redundant. I can't possibly think of a test case now, where these annotations comes into play.

But both of these annotations can be added to highlight the standalone specification behavior only. Otherwise, it doesn't look like a necessary addition.

wmdietl:
The `@Nullable` on the type parameter declaration forbids subtypes to use a `@NonNull` instantiation.
Look at the following example:

```java

class Foo<@Nullable T> {}
class Bar extends Foo<String> {}
```

The supertype of Bar is forbidden.

The annotation on the return type would also prevent anyone from having a `@NonNull` return.
So I'll close this PR. Please reopen if you find a compelling use case for these annotations.

PR #1854, opened 2018-02-26
Don't use nodes in I18nFormatterTreeUtil

smillst:
Also, fix small bug when computing index for the formatting args.

### DIFF
--- a/checker/src/org/checkerframework/checker/i18nformatter/I18nFormatterTreeUtil.java
+++ b/checker/src/org/checkerframework/checker/i18nformatter/I18nFormatterTreeUtil.java
@@ -31,7 +31,6 @@ import org.checkerframework.checker.i18nformatter.qual.I18nInvalidFormat;
 import org.checkerframework.checker.i18nformatter.qual.I18nMakeFormat;
 import org.checkerframework.checker.i18nformatter.qual.I18nValidFormat;
 import org.checkerframework.common.basetype.BaseTypeChecker;
-import org.checkerframework.dataflow.analysis.FlowExpressions.Receiver;
 import org.checkerframework.dataflow.cfg.node.ArrayCreationNode;
 import org.checkerframework.dataflow.cfg.node.FieldAccessNode;
 import org.checkerframework.dataflow.cfg.node.MethodInvocationNode;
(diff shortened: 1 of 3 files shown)


checker/src/org/checkerframework/checker/i18nformatter/I18nFormatterTreeUtil.java
```diff
@@ -254,49 +248,46 @@ public String toString() {
...
+                            formatforArg, flowExprContext, atypeFactory.getPath(tree), true);
+                } catch (FlowExpressionParseException e) {
+                    // report errors here
+                    checker.report(
+                            org.checkerframework.framework.source.Result.failure(
+                                    "i18nformat.invalid.formatfor"),
+                            tree);
+                    return;
+                }
+                if (formatforArg.startsWith("#")) {
```
wmdietl: In https://github.com/typetools/checker-framework/pull/1844#discussion_r170652526 you said `The current code allows any FlowExpression. Your example code does not. I don't know the code well enough to know if restricting to just parameter references is ok.`
This check enforces that the string is of the format `#integer`, so it equally restricts what flow expressions are accepted.
So why not replace lines 275 - 289 by

```java
if (formatforArg.startsWith("#")) {
    // formatArg should be #Number, with 1-indexing.
    try {
        paramIndex = Integer.valueOf(formatforArg.substring(1)) - 1;
    } catch (NumberFormatException nfe) {
        checker.report(Invalid format);
    }
    if (paramIndex < 0 || paramIndex > method.getParameters().size()) {
        checker.report(Invalid index);
    }
}
```

Alternatively, could the old code be used and the ambiguity between 1 and #1 be resolved by checking for #, without re-parsing the integer.

smillst: I don't know if other code relies on this code to issue parse error of any flow expression.  I would prefer to keep the code as close to the old code as possible.

>Alternatively, could the old code be used and the ambiguity between 1 and #1 be resolved by checking for #, without re-parsing the integer.

The problem is not the difference between `1` and `#1`, it's between to arguments that parse to equal Receivers.

PR #1856, opened 2018-02-26
Remove annotations from comments in all source files

smillst:
Removes all annotations in comments from all source files.  Also, adds dependency on checker-qual for all projects.


framework/src/org/checkerframework/framework/test/TestConfigurationBuilder.java
```diff
@@ -404,15 +404,7 @@ public static String getDefaultClassPath() {
...
-        return "build"
-                + File.pathSeparator
-                + "junit-4.12.jar"
-                + File.pathSeparator
-                + "hamcrest-core-1.3.jar"
-                + File.pathSeparator
-                + classpath
-                + File.pathSeparator
-                + globalclasspath;
+        return classpath + File.pathSeparator + globalclasspath;
```
mernst: Is this related to the annotations-in-comments change?

smillst:
@wmdietl Do you want to take a look or should I merge this?

PR #1868, opened 2018-03-04
Fixes #1866 by outputting the currently visited Tree.

### LINKED ISSUE #1866: Report source position for crashes
This is a feature request to detect which AST node was being visited when fatal exceptions occur and include a source position when reporting crashes like #1865. This would be helpful when minimizing repros for crashes. Knowing the file that caused the error is a good start, but it still takes some time to bisect to the part of the file. Since a crash can often be reproduced with a single expression or statement, having a source position would cut out some steps.

We do something similar in Error Prone: the main AST visitor handles fatal exceptions when visiting each node: https://github.com/google/error-prone/blob/66c0e96b90a8fb28a2d3febde5bef3b339b9f6cc/check_api/src/main/java/com/google/errorprone/scanner/ErrorProneScanner.java#L1149. There's nothing particularly great about that implementation, but I've found it helpful when triaging bugs.

wmdietl:


framework/src/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -763,6 +763,10 @@ private void logCheckerError(CheckerError ce) {
             if (this.currentRoot != null && this.currentRoot.getSourceFile() != null) {
                 msg.append("\nCompilation unit: " + this.currentRoot.getSourceFile().getName());
             }
+            if (this.visitor != null) {
+                Tree lv = this.visitor.lastVisited;
+                msg.append("\nLast visited tree: " + lv);
```
cushon: Consider including the line and column position in the diagnostic, and printing the source line corresponding to the tree's diagnostic position ([example](https://github.com/google/error-prone/blob/66c0e96b90a8fb28a2d3febde5bef3b339b9f6cc/check_api/src/main/java/com/google/errorprone/ErrorProneError.java#L61)).

Prettying-printing the entire tree will result in a lot of diagnostic output if the last tree was e.g. a class or compilation unit, and it might not uniquely the position of the crash for small subexpressions.

wmdietl: I agree and changed to line/col information. Thanks.

smillst:
FYI, if the crash happens in during dataflow analysis, then the tree output will always be a class tree rather than the expression/statement that really caused it.  But that is still more helpful than just the java file when a java file contains more than one class.

PR #1871, opened 2018-03-05
Convert the build scripts from Ant to Gradle.

smillst:
I tried to make the Gradle tasks do the same thing as the Ant tasks did.  However, since we are ending support for JSR308, I removed all references to it in the build scripts except for the jtreg tests, which is explained in a comment in that task.  CheckerMain still uses it though because plume-lib still uses annotations in comments and because we haven't removed mentions of annotations in comments from the manual.

See [developer-manual.html](https://github.com/typetools/checker-framework/pull/1871/files#diff-521ff72506c5df28427c8b58da6cb293) for overview of the Gradle tasks.

The release scripts need to be updated for this change, but that will happen after this is merged.

After this is merged, we will remove the ant build scripts.  After that, we should relocate source and tests files so that each source set is in its own directory.


.travis-build.sh
```diff
@@ -88,14 +72,15 @@ if [[ "${GROUP}" == "downstream" || "${GROUP}" == "all" ]]; then
   else
     CFISLUGOWNER=${SLUGOWNER}
   fi
+  CFISLUGOWNER=smillst
   set -e
   echo "Running:  (cd .. && git clone --depth 1 https://github.com/${CFISLUGOWNER}/checker-framework-inference.git)"
-  (cd .. && git clone --depth 1 https://github.com/${CFISLUGOWNER}/checker-framework-inference.git) || (cd .. && git clone --depth 1 https://github.com/${CFISLUGOWNER}/checker-framework-inference.git)
+  (cd .. && git clone --depth 1 -b gradle https://github.com/${CFISLUGOWNER}/checker-framework-inference.git) || (cd .. && git clone -b gradle --depth 1 https://github.com/${CFISLUGOWNER}/checker-framework-inference.git)
```
smillst: Note: This needs to be reverted before merge.  (See PR https://github.com/typetools/checker-framework-inference/pull/71)

framework/tests/src/tests/RangeTest.java
```diff
@@ -183,6 +183,7 @@ public void remove() {
 
         public ValuesInRangeIterator(Range range) {
             this.range = range;
+            Range.IGNORE_OVERFLOW = false;
```
mernst: How is this change related to the rest of the pull request?

.travis-build.sh
```diff
@@ -88,14 +72,15 @@ if [[ "${GROUP}" == "downstream" || "${GROUP}" == "all" ]]; then
   else
     CFISLUGOWNER=${SLUGOWNER}
   fi
+  CFISLUGOWNER=smillst
```
mernst: This should be changed before being merged.

smillst: Yes, it's related to the change below.

smillst:
> There are still occurrences of -target 7:

This pull request doesn't disable annotations in comments, so these are fine.  (A separate pull request will do that along with updating the manual.)

framework/tests/src/tests/RangeTest.java
smillst: This pull request uncovered a bug in the Value Checker.  I'll move the fix to another branch.

framework/tests/annotationclassloader/Makefile
```diff
@@ -47,34 +45,35 @@ demo1:
 demo2:
 	@echo "***** This command is expected to produce an error on line 7:"
 	$(JAVAC) \
-	  -J-Xbootclasspath/p:$(FRAMEWORKJAR):$(STUBPARSERJAR):$(JAVACJAR) \
-	  -classpath :$(PROJECTDIR) \
+	  -processorpath $(FRAMEWORKJAR) \
```
wmdietl: Are these "fat" jars or do they just contain the individual project files? It seems a bit odd that the different demos in this file use different `processorpath`s, even though they probably all have the same dependencies. It would probably be best if all of them depend on the `build` versions and you add a new variable for the union of all of them.

framework/tests/src/tests/RangeTest.java
wmdietl: Is this something that should be pulled into this PR?

framework/tests/src/tests/WholeProgramInferenceValidationTest.java
```diff
@@ -19,6 +19,13 @@ public WholeProgramInferenceValidationTest(List<File> testFiles) {
         super(testFiles, WholeProgramInferenceTestChecker.class, "value", "-Anomsgtext");
     }
 
+    @Override
+    public void run() {
+        if (new File("tests/whole-program-inference/annotated/").exists()) {
+            super.run();
+        }
```
wmdietl: Should this raise a user-friendly error if the directory is not found?

.travis-build.sh
wmdietl: I merged that PR, so the change above can be undone.

framework/tests/annotationclassloader/Makefile
smillst: They are fat jars.  The tests are suppose to use different processorpaths to test loading annotations from a jar file and loading annotations from classes.

PR #1872, opened 2018-03-06
For a member select that refers to a class use the type of the whole tree

wmdietl:
Previously, for a fully-qualified class name like p.q.C it used p.q as ClassName, which doesn't make sense. But maybe I misunderstood and this needs documentation.

### DIFF
--- a/dataflow/src/org/checkerframework/dataflow/analysis/FlowExpressions.java
+++ b/dataflow/src/org/checkerframework/dataflow/analysis/FlowExpressions.java
@@ -304,7 +304,7 @@ public class FlowExpressions {
                 }
                 break;
             case MEMBER_SELECT:
-                receiver = internalRepOfMemberSelect(provider, (MemberSelectTree) receiverTree);
+                receiver = internalReprOfMemberSelect(provider, (MemberSelectTree) receiverTree);
                 break;
             case IDENTIFIER:
                 IdentifierTree identifierTree = (IdentifierTree) receiverTree;
@@ -392,7 +392,7 @@ public class FlowExpressions {
         }
     }
 
-    private static Receiver internalRepOfMemberSelect(
+    private static Receiver internalReprOfMemberSelect(
             AnnotationProvider provider, MemberSelectTree memberSelectTree) {
         TypeMirror expressionType = TreeUtils.typeOf(memberSelectTree.getExpression());
         if (TreeUtils.isClassLiteral(memberSelectTree)) {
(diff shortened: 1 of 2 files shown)


dataflow/src/org/checkerframework/dataflow/analysis/FlowExpressions.java
```diff
@@ -304,7 +304,7 @@ public static Receiver internalReprOf(
                 }
                 break;
             case MEMBER_SELECT:
-                receiver = internalRepOfMemberSelect(provider, (MemberSelectTree) receiverTree);
+                receiver = internalReprOfMemberSelect(provider, (MemberSelectTree) receiverTree);
```
wmdietl: There are other methods in this class that are named `internalRepOf` instead of the more consistent `internalReprOf`.
Those methods were public so I didn't want to change them. It would be good to make this more uniform, either as part of this PR or separately.

dataflow/src/org/checkerframework/dataflow/analysis/FlowExpressions.java
```diff
@@ -407,7 +407,8 @@ private static Receiver internalRepOfMemberSelect(
             case ENUM:
             case INTERFACE: // o instanceof MyClass.InnerInterface
             case ANNOTATION_TYPE:
-                return new ClassName(expressionType);
+                TypeMirror selectType = TreeUtils.typeOf(memberSelectTree);
```
wmdietl: The motivation for this change is that previously, for a fully-qualified class name like `p.q.C` it used `p.q` as `ClassName`, which doesn't make sense. But maybe I misunderstood and this needs documentation.

dataflow/src/org/checkerframework/dataflow/analysis/FlowExpressions.java
```diff
@@ -407,7 +407,8 @@ private static Receiver internalRepOfMemberSelect(
...
+                return new ClassName(selectType);
             case ENUM_CONSTANT:
             case FIELD:
                 Receiver r = internalReprOf(provider, memberSelectTree.getExpression());
```
wmdietl: The line below this one (on which I can't comment) uses `ElementUtils.getType(ele)`.
This seems wrong or at least not as precise as possible. I would use `TreeUtils.typeOf(memberSelectTree)`.
The difference is that the type of the field from the Element is the type from the declaration of the field, which could e.g. be a type parameter. The actual type from the `memberSelectTree` should be the fully-substituted type.

mernst:
This pull request resolves the Lock Checker failure on Daikon.

PR #1877, opened 2018-03-07
Treat declaration annotations on array elements as annotations on the array type

smillst:
Stop treating declaration annotations on array elements as if they were type annotations written in the same position.


docs/manual/nullness-checker.tex
```diff
@@ -856,10 +856,16 @@ \section{Other tools for nullness checking\label{nullness-related-work}}
 
 If your code is already annotated with a different nullness
```
smillst: Should we add something like this:
This support is mainly for third-party libraries.  If you use the Checker Framework to check your code, you should use the Checker Framework nullness annotations.

docs/manual/nullness-checker.tex
```diff
@@ -856,10 +856,16 @@ \section{Other tools for nullness checking\label{nullness-related-work}}
...
 annotation, the Checker Framework can type-check your code.
-It treats annotations from other tools
```
smillst: The Interning, Lock, and Value checkers have the exact same wording as this.  Should we move the new text some and just refer it in those sections?

framework/src/org/checkerframework/framework/util/element/ElementAnnotationUtil.java
```diff
@@ -78,7 +78,8 @@ public static void applyAllElementAnnotations(
      * understand them as type annotations. In particular, this allows the transition from Java 5
      * declaration annotations to Java 8 type annotations.
      *
-     * <p>There are some caveats to this: the interpretation for declaration and type annotations
+     * <p>If the annotation is one of the Checker Framework compatibility annotations, for example
+     * {@link org.checkerframework.checker.nullness.compatqual.NonNullDecl}, the interpretation
```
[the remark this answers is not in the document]
mernst: Do we need `@NonNullDecl` any more?  Since the Checker Framework does not support Java 7 any more, I would think that we can eliminate the `@NonNullDecl` and all related annotations.  Then we don't need to talk about them, anywhere in the manual.

docs/manual/nullness-checker.tex
mernst: This is true, but I don't want to repeat this 4 times in the manual.  I think it's OK to omit.

framework/src/org/checkerframework/framework/util/element/ElementAnnotationUtil.java
```diff
@@ -78,16 +78,27 @@ public static void applyAllElementAnnotations(
      * understand them as type annotations. In particular, this allows the transition from Java 5
      * declaration annotations to Java 8 type annotations.
      *
-     * <p>There are some caveats to this: the interpretation for declaration and type annotations
-     * differs, in particular for arrays and inner types. See the manual for a discussion.
+     * <p>If the annotation is one of the Checker Framework compatibility annotations, for example
```
wmdietl: I think it would make sense to highlight this in the `declaration-annotations-moved` section. It might be useful to have this option of using a type annotation to annotate an array component type.

framework/src/org/checkerframework/framework/util/element/ElementAnnotationUtil.java
```diff
@@ -78,16 +78,27 @@ public static void applyAllElementAnnotations(
      * understand them as type annotations. In particular, this allows the transition from Java 5
```
wmdietl: This comment should be updated.

PR #1878, opened 2018-03-08
closes #1776

### LINKED ISSUE #1776: Duplicate Nullness Initialization tests
Both directories `checker/tests/nullness` and `checker/tests/nullness/init` contain the following files:

```
Uninit11.java
Uninit12.java
Uninit2.java
Uninit3.java
Uninit4.java
Uninit5.java
Uninit6.java
Uninit7.java
Uninit8.java
Uninit9.java
Uninit.java
```

Most of the files are identical, though some have differences. Also note that `init` contains a few extra `Uninit` tests.

As far as I can tell by looking at the `checker/tests/src/tests/NullnessXXXTest.java` files, the `init` directory uses the same configuration as `nullness`, so these tests are redundant.

Anyone know a reason why we should keep both?
I would be for removing the duplicates in `nullness` and keeping only the files in `nullness/init`.

atuldada:
closes #1776  Duplicate Nullness Initialization tests


wmdietl:
Hi @atuldada 
Thanks for the PR! Did you look through all the removed files and make sure this only removes tests we have in other files?
cu, WMD.

atuldada:
Yes! I have checked it for all of them.

PR #1885, opened 2018-03-11
 Put all side effect annotations together.

mernst:
The main purpose of this pull request is to eliminate `flow.astub`, merging its contents into the annotated JDK.  The Nullness annotated JDK currently contains purity annotations as well, so that is where the annotations currently in flow.astub need to go.  Eventually, all annotations will be in a single copy of the JDK.

This pull request makes some other needed or desirable changes -- such changes are collateral damage that increase the size of the pull request.
 * Some Java 8 source code is now uncommented, which was commented before when we used the same source code to create jdk7.jar and jdk8.jar, and the code had to compile with Java 7.  This relates to anything that uses Java 8 features such as Optional, Supplier, etc.
 * For some JDK classes, the source code is updated to use a newer version of the JDK.  This is sometimes necessary because JDK classes reference implementation details of other classes, and sometimes because APIs have changed (especially package-protected ones).  The JDK currently contains .java files from multiple versions of the JDK.  Sorting classes are the main culprits here.
 * More consistent code style.  Add message "skeleton method" to exceptions thrown by skeleton methods.
 * When merging different versions of the JDK, prefer full sources to skeleton methods.  This will make merging into the real JDK easier, later on.
 * Fix line breaks, such as putting declaration annotations `@SideEffectFree` and `@Pure` on their own line.
 * Change framework tests that depended JDK annotations (previously in flow.astub, now in the annotated JDK which is part of the checker/ package, which framework/ does not depend upon.
 * Tweaks in PurityChecker and BaseTypeVisitor that don't really belong in this pull request but were noticed while checking the purity annotations and tests. :-/


mernst:
Unlike the checker/ tests, the framework/ tests do not seem to use the annotated JDK.  Yet, a number of them depend on side effect annotations (see the diffs in this pull request).  I think we should make the annotated JDK available to the tests, or move the affected tests into the checker/ directory, or make the tests use a locally-defined (and annotated) class rather than calling JDK methods.

mernst:
@smillst This is ready for your review now.  Thanks!

checker/tests/nullness/flow/PureAndFlow.java
```diff
@@ -50,11 +50,11 @@ void withPure() {
 
     class Cons {
         @org.checkerframework.dataflow.qual.Pure
-        // :: warning: (purity.deterministic.constructor)
+        // :: warning: (purity.deterministic.constructor) :: error: (purity.not.deterministic.not.sideeffectfree.call)
```
smillst: This is error is happening be the constructor for Object is no longer marked as pure.  Is that intentional?

framework/tests/purity-suggestions/Test.java
```diff
@@ -7,7 +7,6 @@
 // copied from Purity.java)
 
 // This warning is for the implicit constructor of class Test
```
smillst: This should be deleted.

checker/tests/nullness/flow/PureAndFlow.java
mernst: I don't think it makes sense for a constructor to be `@Deterministic`.  In a separate pull request, we should prohibit `@Deterministic` and `@Pure` on constructors.

framework/tests/purity-suggestions/Test.java
mernst: I found the comment helpful because the warning is about implicit code.  If you think it's obvious, we could delete the comment.

smillst: The warning doesn't occur anymore.

mernst:
That is unrealistic, because the changes took hours to merge.  Furthermore, the revision history will be lost/reset when we move the JDK, so it doesn't seem important to carefully maintain it now.  It also seems unlikely that you were going to review every one of the hundreds of changes. :-)

wmdietl:
The PR description should at least mention these large changes to the annotated JDK and what they are.
At the moment it is impossible to know what parts to look at and what parts are just updates from upstream (or whatever the other changes are, there is no indication).
The goal shouldn't be to make it impossible to review a pull request, under the assumption that nobody will look at it.

mernst:
> "The goal shouldn't be to make it impossible to review a pull request".

I have not stated this as a goal.

mernst:
I updated the pull request description.

wmdietl:
Thanks for the details!
Once the tests pass I'm fine with merging.

PR #1886, opened 2018-03-11
Annotated the squash library with Nullness checker

uemarica:
I annotated the squash library with the nullness checker and the README.md file, the pom.xml in the readme.md refers to the squash library's pom.xml file.


wmdietl:
This is not the right place for these changes.
You should work on annotating the library in a fork of the respective library and then send us a link to that.
There is no point in making this a PR against the Checker Framework.

Issue #1876, opened 2018-03-07
Manual update required for Section 30.7.1 example definitions


ashishrana160796:
Definition of `@NonNull` annotation class stated in this section doesn't match with current one and needs a update in manual.
Earlier manual section pointed:
```
For example, consider the definitions of the @Nullable and `@NonNull` type qualifiers:
...
@SubtypeOf( { Nullable.class } )
...
      literals = { LiteralKind.BOOLEAN, LiteralKind.CHAR, LiteralKind.DOUBLE,
                   LiteralKind.FLOAT, LiteralKind.INT, LiteralKind.LONG,
                   LiteralKind.STRING })
..
```
Updated manual section suggested and also addition of `@MonotonicNonNull` definition : 
```
For example, consider the definitions of the @Nullable and `@NonNull` type qualifiers:
...
@SubtypeOf(MonotonicNonNull.class)
...
literals = {LiteralKind.STRING}
...
```
But if this example is for illustration purpose than close this issue, still I'll suggest not reference `@NonNull` and `@Nullable` from actual link of JavaDoc which is quite misleading.

mernst:
Closed by  7390552

PR #1887, opened 2018-03-11
1876 issue resolved

ashishrana160796:
Update for` @NonNull `'s definition in manual example in reference to issue #1876 .


mernst:
Thanks.  It's good to use consistent code in the manual and in the implementation.
I fixed an incorrect comment and now will merge.

PR #1888, opened 2018-03-12
Require Java 8, and always require the annotated JDK

mernst:


framework/src/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -427,6 +428,15 @@ public final synchronized void init(ProcessingEnvironment env) {
...
         // This is used to trigger AggregateChecker's setProcessingEnvironment.
         setProcessingEnvironment(env);
+
+        double jreVersion = PluginUtil.getJreVersion();
+        if (jreVersion != 1.8) {
+            report(
+                    org.checkerframework.framework.source.Result.failure(
+                            "The Checker Framework must be run under JDK 1.8.  You are using version %f.",
+                            jreVersion),
+                    null);
```
mernst: This uses `null` as the location at which to issue the warning.
Is there a better way to issue the warning, such as by immediately terminating compilation?

smillst: Yes, use ErrorReporter.errorAbort().  That method throws an exception.

mernst:
@smillst This pull request is still failing because many tests seem not to use the annotated JDK.
I think probably those tests should use the annotated JDK, but an alternative would be to give a way to override the Checker Framework's requirement to use the annotated JDK.

smillst:
Framework tests need 'nocheckjdk' arguments.

mernst:
@smillst Framework tests need to pass some "don't check for JDK" flag.

checker/src/org/checkerframework/checker/nullness/KeyForAnnotatedTypeFactory.java
```diff
@@ -93,7 +93,7 @@ protected TypeHierarchy createTypeHierarchy() {
         return new KeyForTypeHierarchy(
                 checker,
                 getQualifierHierarchy(),
-                checker.getOption("ignoreRawTypeArguments", "true").equals("true"),
+                checker.getBooleanOption("ignoreRawTypeArguments"),
```
smillst: I think this pull request contains changes that are also in #1891.

mernst:
@smillst I think I addressed your comments. Can you see whether this is ready to be merged?  Thanks!

PR #1891, opened 2018-03-12
 Fail, don't continue silently, if a boolean value is misspelled

mernst:


framework/src/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -1729,6 +1744,17 @@ public final String getOption(String name) {
...
 
+    /**
+     * Determines the boolean value of the option with the given name. Returns false if the option
+     * is not set.
+     *
+     * @see SourceChecker#getLintOption(String,boolean)
+     */
+    @Override
+    public final boolean getBooleanOption(String name) {
+        return Boolean.valueOf(getOption(name, "true"));
```
wmdietl: The documentation says that `Returns false if the option is not set.`
However, here `true` is provided as the default value. So doesn't this return `true` if the option is not set?

mernst:
I fixed the typo in the default.
The error comes from `Boolean.valueOf`.  This is better than silently proceeding when the user has a typo in a command line, such as `-Demit.test.debug=tru`.

framework/src/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -1729,6 +1744,18 @@ public final String getOption(String name) {
...
+    /**
+     * Determines the boolean value of the option with the given name. Returns false if the option
+     * is not set.
+     *
+     * @param name the name of the option to check
+     * @see SourceChecker#getLintOption(String,boolean)
+     */
+    @Override
+    public final boolean getBooleanOption(String name) {
+        return Boolean.valueOf(getOption(name, "false"));
```
wmdietl: `Boolean.valueOf` returns false if the input is null or a string other than `"true"`. It does not raise an exception.
Even if it did raise an exception, it would be nicer to raise a user-friendly error message that mentions the option name.
You do need to check for valid values here and in the other use of `Boolean.valueOf`.

mernst: Thanks.  I had no idea that was the behavior.  (It seems wrong to me!)
I will update this to accommodate the actual behavior rather than what I thought it was.

framework/src/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -680,7 +680,7 @@ protected TypeHierarchy createTypeHierarchy() {
         return new DefaultTypeHierarchy(
                 checker,
                 getQualifierHierarchy(),
-                checker.getOption("ignoreRawTypeArguments", "true").equals("true"),
+                checker.getBooleanOption("ignoreRawTypeArguments"),
```
smillst: This should be `checker.getBooleanOption("ignoreRawTypeArguments", true)`

framework/src/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -1730,42 +1745,67 @@ public final String getOption(String name) {
...
+    public final boolean getBooleanOption(String name, boolean defaultValue) {
+        String value = getOption(name);
+        if (value == null) {
+            return defaultValue;
+        }
+        if (value.equals("false")) {
+            return false;
+        }
+        if (value.equals("true")) {
+            return false;
```
smillst: Should this be true?

mernst: Yes.  Thanks!

mernst:
Thanks for the comments -- they made it better.

mernst:
Diffs in GitHub are messed up.  Closing and re-opening.

PR #1893, opened 2018-03-13
I too found the same Uninit files repeated so deleted the redundant files from checker/tests/nullness

uemarica:


wmdietl:
#1878 already did this.

PR #1895, opened 2018-03-14
Documentation spelling error

uemarica:


mernst:
Thanks for pointing out the spelling error -- we appreciate it, because we want the documentation to be as high-quality as possible.

I fixed this in 1843fc7bb, because your pull request also makes changes to many other irrelevant files.

I suggest that you make the master branch of your fork be an unmodified copy of the master of upstream, and only make changes on new branches.  There is some advice about making a GitHub pull request at https://homes.cs.washington.edu/~mernst/advice/github-pull-request.html.  Thanks again!

PR #1904, opened 2018-03-19
Fail, don't continue silently, if a boolean value is misspelled

mernst:


mernst:
@smillst You gave some comments that I addressed.  Then Werner gave comments, I addressed them, and Werner approved this pull request.  Do you want to review it again, or not?

PR #1918, opened 2018-03-27
Remove mention of annotations in comments from manual.

smillst:


lazaroclapp:
Hi @smillst, sorry to jump in on this out of nowhere, but: Is support for annotations in comments in CF going away? 

We are currently using it heavily as a workaround for an extremely nasty JDK bug (https://bugs.openjdk.java.net/browse/JDK-8059531 ), which doesn't look like is going to be fixed before JDK 11. The bug silently generates broken bytecode when an anonymous inner class is annotated with a type use annotation. I was wondering if you guys were aware of that problem.

cc: @msridhar

mernst:
@lazaroclapp We plan to remove all documentation of annotations in comments and to stop supporting it.  The goal is to simplify the manual for users (shorter text, fewer choices, simpler instructions) and to simplify maintenance for the Checker Framework developers.  We made this decision because on February 5, I sent a [survey](https://groups.google.com/forum/#!topic/checker-framework-discuss/uftr0CLrzQE) to the mailing list and no one indicated that they depended on annotations in comments.

The feature will still work in the same way it did before, but it won't be documented.

msridhar:
@mernst thanks for the info.  I'm now actually getting notifications for the email list (before it was disabled for some reason).  Understood re: the maintenance burden.  However, we may want to warn Checker Framework users about putting type use annotations on anonymous classes.  [This duplicate javac bug](https://bugs.openjdk.java.net/browse/JDK-8177316) has a better example showing the issue.  The major concern here is that it's a wrong code bug, so the failure happens at runtime, not compile time.  IMO type use annotations on anonymous classes are quite dangerous due to this bug and should be avoided (unless there is a workaround we are missing).

smillst:
@lazaroclapp and @msridhar thanks for bring that bug to our attention.  The Checker Framework should probably issue a warning.  I've opened issue #1928 to do so.

PR #1921, opened 2018-03-28
Use a one-element array instead of javax.xml.ws.Holder

cushon:
In JDK 9, javax.xml.ws is not included in the default set of root
modules, and is scheduled for removal from a future JDK version:
http://openjdk.java.net/jeps/320


wmdietl:
Thanks for the contribution!

PR #1931, opened 2018-04-02
Use pos for comparing local variable flow expressions

### LINKED ISSUE #1847: False negative on nullness check due to overlapping variable names in method
Complete example is below to reproduce with the nullness checker and annotated JDK in 2.3.2.  You will find that as written, it doesn't give nullness errors on the two lambdas passed to the withLookup calls, even though it should (map.get(myVar) can be null in the lambdas).  My guess is that the `@KeyFor("map")` annotation inferred for the myVar loop variable gets stored by local var name, and thus the checker thinks the two lambda variables of the same name (but which are totally different variables) must also be `@KeyFor("map")` and doesn't give a nullness error.   This seems to only affect lambda vars, I briefly tried with a second local var called myVar later in the method, and I couldn't seem to reproduce the same issue.

```

import java.util.HashMap;
import java.util.Iterator;
import java.util.Map;
import java.util.Map.Entry;

public class MapBug
{
    final Map<String, String> map = new HashMap<>();
    
    public void test()
    {
        // Should give null error here:
        withLookup(myVar -> map.get(myVar).length());
        for (Iterator<Entry<@KeyFor("map") String, String>> iterator = map.entrySet().iterator(); iterator.hasNext(); )
        {
            Entry<@KeyFor("map") String, String> entry = iterator.next();
            // Problem is that myVar gets inferred as @KeyFor("map") here,
            // and this variable is not distinguished from the lambda variables of the same name,
            // even though their scopes do not overlap and they are different variables.
            // Change this variable name to myVar2 and you will see the null errors on the lambdas:
            String myVar = entry.getKey();
        }

        // Should also give null error here:
        withLookup(myVar -> map.get(myVar).length());
    }
    
    public void withLookup(Function<String, Integer> getFromMap)
    {
    }
}
```

smillst:
Fixes #1847.

### DIFF
--- a/checker/tests/nullness/Issue1847B.java
+++ b/checker/tests/nullness/Issue1847B.java
@@ -1,0 +1,33 @@
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.function.Function;
+import org.checkerframework.checker.nullness.qual.KeyFor;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public class Issue1847B {
+    final Map<String, String> map = new HashMap<>();
+
+    public void test() {
+        // Should give null error here:
+        // :: error: (dereference.of.nullable)
+        withLookup((@Nullable String myVar) -> myVar.toString());
+        for (Iterator<Entry<@KeyFor("map") String, String>> iterator = map.entrySet().iterator();
+                iterator.hasNext();
+                ) {
+            Entry<@KeyFor("map") String, String> entry = iterator.next();
+            // Problem is that myVar gets inferred as @KeyFor("map") here,
+            // and this variable is not distinguished from the lambda variables of the same name,
+            // even though their scopes do not overlap and they are different variables.
+            // Change this variable name to myVar2 and you will see the null errors on the lambdas:
+            String myVar = entry.getKey();
+        }
+
+        // Should also give null error here:
+        // :: error: (dereference.of.nullable)
+        withLookup(myVar -> map.get(myVar).toString());
+    }
+
+    public void withLookup(Function<String, String> getFromMap) {}
+}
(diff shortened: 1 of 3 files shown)


dataflow/src/org/checkerframework/dataflow/analysis/FlowExpressions.java
```diff
@@ -728,18 +728,23 @@ public boolean equals(Object obj) {
...
+            // The code below isn't just return vs.equals(vsother) because an element might be different
+            // between subcheckers.
+
             // Use TypeAnnotationUtils.unannotatedType(type).toString().equals(...) instead of
             // Types.isSameType(...) because Types requires a processing environment, and
             // FlowExpressions is designed to be independent of processing environment.  See also
             // calls to getType().toString() in FlowExpressions.
             return vsother.name.contentEquals(vs.name)
+                    && vsother.owner.toString().equals(vs.owner.toString())
+                    && vs.pos == vsother.pos
```
wmdietl: I would move the `pos` comparison first, as it is the cheapest one to perform and most likely to fail.
Do we still need the type comparison? Can two symbols have the same position, name, and owner but have different types?
An update to the comment above this condition would also help in understanding why we need to consider the position.
Symbols from bytecode have no `pos`, I assume; I guess this isn't ever a problem for this.

checker/tests/nullness/Issue1847B.java
```diff
@@ -0,0 +1,33 @@
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.function.Function;
+import org.checkerframework.checker.nullness.qual.KeyFor;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public class Issue1847B {
```
wmdietl: The two tests look very similar. Maybe a comment could highlight what the difference is?

PR #1932, opened 2018-04-02
Cleaner separation of CFGBuilder classes.

wmdietl:


dataflow/src/org/checkerframework/dataflow/cfg/CFGBuilder.java
```diff
@@ -256,106 +237,53 @@ public static ControlFlowGraph build(
...
+        ControlFlowGraph phase3result = CFGTranslationPhaseThree.process(phase2result);
+        return phase3result;
     }
 
-    /** Build the control flow graph of a method. */
+    // NullAway uses this entry point.
+    @Deprecated
     public static ControlFlowGraph build(
-            CompilationUnitTree root,
+            TreePath bodyPath,
```
smillst: This method is marked `@Deprecated`, so should the signature stay the same?

wmdietl: I think the alignment in the diff is off.
The call is here: https://github.com/uber/NullAway/blob/master/nullaway/src/main/java/com/uber/nullaway/dataflow/DataFlow.java#L110
and the parameters here should match that call.

wmdietl:
@lazaroclapp Should we keep the deprecated `build` method for NullAway or is it easy enough for you to change that?
Any other comments on this PR?

dataflow/src/org/checkerframework/dataflow/cfg/CFGBuilder.java
smillst: Yes, the diff if weird.  Thanks!

lazaroclapp:
@wmdietl I see nothing preventing us to change `CFGBuilder.build(...)` to `new CFGBuilder(...).run(...)` with the same values. We use it in exactly one place: https://github.com/uber/NullAway/blob/master/nullaway/src/main/java/com/uber/nullaway/dataflow/DataFlow.java#L110 

I think given how we pin the CF version for building NullAway, you two changing this API wouldn't affect us, and we can switch to the new way of doing things whenever we bump the CF dependency (which will be soon). Not sure what other CF users do, though.

cc: @msridhar just in case and to keep this on everyone's radar, but he is away for the week.

wmdietl:
Thanks! It would continue to be `CFGBuilder.build(...)`, just the order of arguments changed. But as you'll get a compiler error if you don't change the invocation, this seems low risk.
I'll remove the deprecated version.

PR #1938, opened 2018-04-04
Use a single Analysis/TransferFunction instance for all CFG operations.

wmdietl:
This makes the interaction between the different components clearer and reduces instantiations (which might happen reflectively otherwise.)


wmdietl:
I am running into NPEs with the Index Checker. All other tests pass.
A minimal example:

```java

class Min {
    void dummy(String[] a, @IndexFor("#1") int i) {
        @IndexFor("a") int j = i - 1;
        @LTLengthOf(value = "a", offset = "1") int k = i;
    }

    void extraMethodNeededForCrash() {}
}
```

It is very strange that the empty method `extraMethodNeededForCrash` is needed for the NPE.
I tried adding a few more null tests, but always in the end ran against a wall...

@smillst  Could you have a look to see what the Index Checker is doing? Thanks!

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -737,16 +737,14 @@ private UBQualifier fromLessThanOrEqual(
                 continue;
             }
             CFStore store = getStoreBefore(tree);
-            if (store != null) {
```
smillst: Why can't the store be null here?

wmdietl: When would the Store for `tree` not be available? It looked like this is always done for parts of the code for which we have a Store. I don't think a NPE is raised here, is it?

Why do we need to look up the Store within this loop? Could this be hoisted out of the loop and checked for once?

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -419,16 +419,14 @@ protected void setSelfTypeInInitializationCode(
         // there might still be subclasses that need initialization.
         if (areAllFieldsCommittedOnly(enclosingClass)) {
             Store store = getStoreBefore(tree);
-            if (store != null) {
```
wmdietl: Sorry, maybe this change should be undone, b/c it now adds the null check on line 497 below.
What I find strange about the null check below is that the nullness tests don't fail with an NPE, but with a completely unrelated assertion error.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
smillst: org.checkerframework.framework.type.GenericAnnotatedTypeFactory#getStoreBefore(org.checkerframework.dataflow.cfg.node.Node) might return null, so getStoreBefore(Tree) might return null, too.  But I was mostly wondering if something changed in this pull request that would make this redundant.

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
wmdietl: Oh, and I only run into this assertion violation when running the JUnit tests, not when manually compiling all the same files at once. So something is odd about this null check, too.

checker/src/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
wmdietl: But the Store is only null if one uses the wrong tree. I rather have a NPE than silently doing nothing.

dataflow/src/org/checkerframework/dataflow/analysis/Analysis.java
```diff
@@ -422,15 +426,18 @@ protected boolean updateNodeValues(Node node, TransferResult<A, S> transferResul
...
     /** Initialize the analysis with a new control flow graph. */
     protected void init(ControlFlowGraph cfg) {
+        thenStores.clear();
+        elseStores.clear();
+        if (blockCount != null) {
+            blockCount.clear();
+        }
+        inputs.clear();
+        storesAtReturnStatements.clear();
+        //        nodeValues.clear();
```
smillst: This is what "fixed" the crash.   Here's why:

The Value Checker runs dataflow on each block in the current class tree, clearing out the nodeValues before each block.  After the Value Checker checks the current class tree, the Upper Bound Checker requests a type from the Value Checker.  When the Value Checker re-runs dataflow to get that type, the nodeValues for that block have already been cleared.

I don't think this is the right "fix", but maybe with this information you can decide how to proceed.

wmdietl: Thanks for figuring this out. I agree that this isn't the right fix :-) All results that need to survive should already have been added to `flowResults` and then should be available for other analyses.

Could you point me to where the Upper Bound Checker interacts with the Value Checker?

smillst: Here's one example:
UpperBoundVisitor.relaxedCommonAssignmentCheck(UpperBoundVisitor.java:262)

But, the Index Checker is full of such interactions.

wmdietl:
@smillst All tests passed for me locally. I'm now threading the `nodeValues` to the right places when re-running an `Analysis`. Overall, this should create many fewer Analysis & Transfer instances and make their interaction a bit clearer.
I'll file a few issues for possible further improvements.

checker/src/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
wmdietl: The store was null only in a first pass through a constructor. I don't understand why the NPE would only show up in JUnit tests, not when run on an individual file. Something might still remain to be cleaned up here...

wmdietl:
CF-inference needs a PR go go along https://github.com/typetools/checker-framework-inference/pull/77

dataflow/src/main/java/org/checkerframework/dataflow/analysis/Analysis.java
```diff
@@ -53,9 +52,11 @@
     protected boolean isRunning = false;
 
     /** The transfer function for regular nodes. */
+    // TODO: make final. Currently, the transferFunction has a reference to the analysis, so it
```
smillst: I add more info to this TODO, which I think is the problem.  But please have a look.

wmdietl: Thanks a lot for cherry-picking the unrelated changes! This expanded explanation is good.

PR #1943, opened 2018-04-09
Add mapping from type argument (Fixes #1942)

### LINKED ISSUE #1942: Error when using recursive generics
I'm not 100% sure if this is a bug, but it seems like the framework is issuing an error where I don't think there is an error.  The issue comes with recursive generics.  I've boiled it down to a simple test case.  I have a simple annotation which I've used here with the subtying checker for simplicity:

```
import java.lang.annotation.ElementType;
import org.checkerframework.framework.qual.Unqualified;

@SubtypeOf(Unqualified.class)
@Target({ElementType.TYPE_USE, ElementType.TYPE_PARAMETER})
public @interface Recorded {}
```

Then I have this code which pulls in that annotation on the classpath (and runs the subtyping checker with the appropriate -Aquals argument to check the Recorded annotation):

```
import java.util.List;

public class RecursiveTypes
{
    public interface LoadableExpression<EXPRESSION>
    {
    }

    static abstract class OperatorSection<EXPRESSION extends LoadableExpression<EXPRESSION>>
    {
        abstract EXPRESSION makeExpression(List<@Recorded EXPRESSION> expressions);
    }

    static class BinaryOperatorSection<EXPRESSION extends LoadableExpression<EXPRESSION>> extends OperatorSection<EXPRESSION>
    {
        @Override
        EXPRESSION makeExpression(List<@Recorded EXPRESSION> expressions) // <-- Error on this line
        {
            throw new RuntimeException("");
        }
    }
```

(The example doesn't make much sense now that I've cut it down, admittedly, but at least the code is short.)  Here's the error I get:

```
[override.param.invalid] Incompatible parameter type.
Method
EXPRESSION extends @Unqualified LoadableExpression<EXPRESSION extends @Unqualified LoadableExpression<EXPRESSION>> makeExpression(@Unqualified BinaryOperatorSection<EXPRESSION extends @Unqualified LoadableExpression<EXPRESSION>> this, @Unqualified List<EXPRESSION extends @Recorded LoadableExpression<EXPRESSION extends @Unqualified LoadableExpression<EXPRESSION>>> p0) in RecursiveTypes.BinaryOperatorSection
cannot override
EXPRESSION extends @Unqualified LoadableExpression<EXPRESSION> makeExpression(@Unqualified OperatorSection<EXPRESSION extends @Unqualified LoadableExpression<EXPRESSION>> this, @Unqualified List<EXPRESSION extends @Recorded LoadableExpression<EXPRESSION>> p0) in RecursiveTypes.OperatorSection
found   : @Unqualified List<EXPRESSION extends @Recorded LoadableExpression<EXPRESSION extends @Unqualified LoadableExpression<EXPRESSION>>>
required: @Unqualified List<EXPRESSION extends @Recorded LoadableExpression<EXPRESSION>>
```

That last part strikes me as weird.  The required type seems sensible, but the found type has unrolled the recursive generic by one extra iteration to complain.  But I can't see a way for me to fix the type, and it seems odd that the "outer" EXPRESSION effectively has a different type/bound to the "inner" EXPRESSION which should be the same type?  I am not doing anything complicated in the code with the error, just trying to override a method using the exact same signature as the parent.  The problem goes away if you make the generic type non-recursive.  Tested with 2.5.0, although I also had the problem in 2.4.0.

smillst:


checker/tests/tainting/Issue1942.java
```diff
@@ -0,0 +1,18 @@
...
+public class Issue1942 {
+    public interface LoadableExpression<EXPRESSION> {}
+
+    abstract static class OperatorSection<A extends LoadableExpression<A>> {
+        abstract A makeExpression(List<@Untainted A> expressions);
+    }
+
+    static class BinaryOperatorSection<B extends LoadableExpression<B>> extends OperatorSection<B> {
+        @Override
+        B makeExpression(List<@Untainted B> expressions) // <-- Error on this line
```
wmdietl: I would clarify the comment and say that previously there was a false positive on that particular line.
Putting the comment on the line before would make formatting a bit nicer.

PR #1947, opened 2018-04-13
Check for uninitialized static fields once on the class tree.

wmdietl:
Checking on the last static initializer block can lead to false
positives.


smillst:
Looks good.  Thanks!  (Merge when tests pass.)

PR #1950, opened 2018-04-20
Don't modify original AnnotatedTypeMirror 

### LINKED ISSUE #1949: Failed to override method
```
public class CFTest {
    public interface Base<R>{}
    public interface Child<R> extends Base<@Nullable R>{}
    public static abstract class BaseClass<R> implements Child<R>{ // 1
//    public static abstract class BaseClass<R { // 2
        abstract List<Child<R>> foo();
    }
    public static class ChildClass extends BaseClass<String> {

        @Override
        public List<Child<String>> foo() {
            return new ArrayList<>();
        }
    }
}
```
Failed to compile:
```
[ERROR] CFTest.java: [override.return.invalid] Incompatible return type.
  Method
    @Initialized @NonNull List<@Initialized @NonNull Child<@Initialized @NonNull String>> foo(@Initialized @NonNull ChildClass this) in CFTest.ChildClass
  cannot override
    @Initialized @NonNull List<@Initialized @NonNull Child<@Initialized @Nullable String>> foo(@Initialized @NonNull BaseClass<@Initialized @Nullable String> this) in CFTest.BaseClass
  found   : @Initialized @NonNull List<@Initialized @NonNull Child<@Initialized @NonNull String>>
  required: @Initialized @NonNull List<@Initialized @NonNull Child<@Initialized @Nullable String>>

```
But when comment line //1 and uncomment //2 compilation successes.

smillst:
Fixes #1949.


checker/tests/nullness/Issue1949.java
```diff
@@ -0,0 +1,22 @@
+import java.util.ArrayList;
+import java.util.List;
+import org.checkerframework.checker.nullness.qual.*;
+
+public class Issue1949 {
+    public interface Base<R> {}
+    //public interface Child<R> extends Base<@Nullable R>{}
```
wmdietl: Instead of the commented out lines I think it would be better to explicitly have all the intended combinations.
At the moment the code doesn't correspond to the original issue, so it is hard to see what the comments mean.

PR #1957, opened 2018-04-24
Reset Analysis in case the checker crashed on a previous class.

smillst:
Otherwise, isRunning might still be true, which causes an assertion failure in `org.checkerframework.dataflow.analysis.Analysis#performAnalysis`.

### DIFF
--- a/dataflow/src/main/java/org/checkerframework/dataflow/analysis/Analysis.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/analysis/Analysis.java
@@ -176,141 +176,139 @@ public class Analysis<
 
     /** Perform the actual analysis. */
     public void performAnalysis(ControlFlowGraph cfg) {
+        try {
+            preformAnalysisHelper(cfg);
+        } finally {
+            // In case preformatAnalysisHelper crashed, reset isRunning to false.
+            isRunning = false;
+        }
+    }
+
+    /** Perform the actual analysis. */
+    private void preformAnalysisHelper(ControlFlowGraph cfg) {
         assert !isRunning;
         isRunning = true;
 
-        try {
-            init(cfg);
+        init(cfg);
 
-            while (!worklist.isEmpty()) {
-                Block b = worklist.poll();
+        while (!worklist.isEmpty()) {
+            Block b = worklist.poll();
 
-                switch (b.getType()) {
-                    case REGULAR_BLOCK:
-                        {
-                            RegularBlock rb = (RegularBlock) b;
+            switch (b.getType()) {
+                case REGULAR_BLOCK:
+                    {
+                        RegularBlock rb = (RegularBlock) b;
 
-                            // apply transfer function to contents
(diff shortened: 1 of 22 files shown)


framework/src/main/java/org/checkerframework/framework/flow/CFAbstractAnalysis.java
```diff
@@ -88,6 +88,13 @@ public CFAbstractAnalysis(
                 factory.getQualifierHierarchy().numberOfIterationsBeforeWidening());
     }
 
+    /** Reset the analysis so it can be reused. */
```
wmdietl: Why provide a reset mechanism only for `CFAbstractAnalysis` instead of for all kinds of `Analysis`?
Wouldn't it be better to put the body of `org.checkerframework.dataflow.analysis.Analysis.performAnalysis(ControlFlowGraph)` into a helper method and have a try-catch block around the invocation of that helper method and set/reset `isRunning` there?

smillst: >Wouldn't it be better to put the body of org.checkerframework.dataflow.analysis.Analysis.performAnalysis(ControlFlowGraph) into a helper method and have a try-catch block around the invocation of that helper method and set/reset isRunning there?

I can do that, but this problem happens if any of the code crashes GenericAnnotatedTypeFactory#performFlowAnalysis, we still need to sit isRunning to false in that method.

smillst: The other alternative is to remove the `assert !isRunning` from Analysis#performAnalysis.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/Analysis.java
```diff
@@ -174,134 +174,143 @@ public ProcessingEnvironment getEnv() {
...
-                                succ, lastNode, currentInput, rb.getFlowRule(), addToWorklistAgain);
-                        break;
-                    }
+        try {
+            init(cfg);
+
+            while (!worklist.isEmpty()) {
+                Block b = worklist.poll();
+
+                switch (b.getType()) {
```
wmdietl: How about putting this whole switch block into a protected helper method, something like `analyseBlock`? That makes the overall control flow much easier to read.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/Analysis.java
```diff
@@ -174,134 +174,143 @@ public ProcessingEnvironment getEnv() {
...
+                    default:
+                        assert false;
+                        break;
+                }
             }
-        }
 
-        assert isRunning;
-        isRunning = false;
+            assert isRunning;
```
wmdietl: I would keep the assert and the update to `isRunning` together, as they are a logical unit.

PR #1960, opened 2018-04-25
Work around JDK bug

### LINKED ISSUE #1954: NPE in checked code
This test is compilable with CF enabled and produce NPE 
```
public class NPETest {
    public interface Getter<R>{
        R get();
    }

    public interface NullStringGetter extends Getter<@Nullable String>{}

    public <T, R> Getter<R> transform(Function<Stream<T>, R> fn, Getter<T> getter){
        return () -> fn.apply(Stream.of(getter.get()));
    }

    public static <T> @Nullable T fn (Stream<T> arg){
        return arg.findFirst().orElse(null);
    }

    @Test
    public void doo(){
        NullStringGetter nullStringGetter = () -> null;
        transform(NPETest::fn, nullStringGetter).get();
    }
}

```

### LINKED ISSUE #1958: Generic arg annotations are not passed across modules
```
public class SupplierDefs {
    public abstract static class Supplier<R>{
        public abstract R get();
    }
    public static class NullSupplier extends Supplier<@Nullable String> {

        @Override
        public @Nullable String get() {
            return null;
        }
    }
}
```

```
public class NPE2Test {
    @Test
    public void testNPE() {
        SupplierDefs.Supplier<String> s = new SupplierDefs.NullSupplier();
        boolean b = s.get().equals("");
    }
}
```

If this two classes are in the very same java module and very same source root then CF found nullness issue:

```
Error:(7, 43) java: [assignment.type.incompatible] incompatible types in assignment.
  found   : @Initialized @NonNull NullSupplier
  required: @UnknownInitialization @Nullable Supplier<@Initialized @NonNull String>
```

But if put SupplierDefs to source root and NPE2Test to test root, or to two different modules, there are no compilation errors and NPE occures in testNPE().

smillst:
Fix #1958


justmin:
Will it fix 1954 and 1958 both?

framework/src/main/java/org/checkerframework/framework/util/element/SuperTypeApplier.java
```diff
@@ -73,10 +73,10 @@ public int getElementIndex() {
         return index;
     }
 
-    /** @return the type_index of anno's TypeAnnotationPosition */
+    /** @return the pos of anno's TypeAnnotationPosition */
     @Override
     public int getTypeCompoundIndex(Attribute.TypeCompound anno) {
-        return anno.getPosition().type_index;
+        return anno.getPosition().pos;
```
wmdietl: I don't think this is the right change. `pos` is the source tree position of the annotation; for an annotation from bytecode this is just -1, the default value.
The result of this method is compared against the result of `getElementIndex()` (in IndexedElementAnnotationApplier).
As `pos` and `index` count completely different things, the values only match in corner cases.

Can you add test cases where `pos` and `index` are not -1 and ensure they work?
For example, what happens when you use an interface instead of `String`?

smillst:
@justmin No, I original wrote the wrong issue number.

smillst:
We reverted the original change because the part that stored annotations in extends bounds caused a crash on Daikon.   Then you put back the changes to SuperTypeApplier.java in this pull request:
https://github.com/typetools/checker-framework/pull/1253/files
The change in this pull request should have been added back, too.

PR #1968, opened 2018-05-01
Change the default for null to @UnknownKeyFor.

### LINKED ISSUE #1922: NullPointerException when using KeyFor and a generic method
I'm not 100% sure if this count as a bug in the framework or just an invalid usage, but I have a NullPointerException being thrown without using any SuppressWarnings annotations, which seems like it may count as a bug.  Below is a minimal working example of where I encountered the issue (compiled with version 2.4.0 and the nullness checker on).  I am managing to "launder" a non-key null value via the findKey method and have it return as if it was a key.  In the code as-is, no error is issued (but I do get a NullPointerException if I run it).  However, assigning null directly to myKey (the commented out line) correctly gives an error:

```

import java.util.Collection;
import java.util.HashMap;
import java.util.Map;

public class TestKeyFor
{
    // A method to find a K in the collection and return it, or return null.
    public static <K> @Nullable K findKey(Collection<@NonNull K> keys, Object target)
    {
        for (K key : keys)
        {
            if (target.equals(key))
                return key;
        }
        return null;
    }
    
    // Find a key in a map and return String version of its value.
    public static String findKeyAndFetchString(Map<String, Object> someMap)
    {
        @Nullable @KeyFor("someMap") String myKey = TestKeyFor.<@KeyFor("someMap") String>findKey(someMap.keySet(), "Foo");
        // If I change above line to this line, I do get an error:
        //@Nullable @KeyFor("someMap") String myKey = null;
        
        Object value = someMap.get(myKey);
        // value will be null here, but checker gives no error: 
        return value.toString();
    }
    
    public static void main(String[] args)
    {
        findKeyAndFetchString(new HashMap<>());
    }
}
```

It's a slightly weird example now it's been cut down, but it's at least short.  I think the issue is the way that I can attach NonNull/Nullable extras to the K type of findKey, but that interacts with the KeyFor annotation in a way that the checker is not realising breaks the type safety.

smillst:
Also, change the default for lower bounds to `@UnkownKeyFor`.

For Java packages in the annotated jdk, switch the default for lower bounds to `@KeyForBottom`.

Fixes #1922.


checker/tests/nullness/TestInfer.java
```diff
@@ -11,7 +11,7 @@
     void bar(Object o) {}
 
     void foo() {
-        List<? extends Object> ls = new ArrayList<>();
+        List<@UnknownKeyFor ? extends Object> ls = new ArrayList<>();
```
smillst: This is required because the defaults on a wildcard are set to the same as the type parameter.

checker/jdk/nullness/src/java/package-info.java
```diff
@@ -0,0 +1,5 @@
+@DefaultQualifier(value = KeyForBottom.class, locations = TypeUseLocation.LOWER_BOUND)
+package java;
+
+import  org.checkerframework.framework.qual.*;
+import  org.checkerframework.checker.nullness.qual.*;
```
wmdietl: Newline at end of file is preferred.

checker/jdk/nullness/src/java/package-info.java
```diff
@@ -0,0 +1,5 @@
+@DefaultQualifier(value = KeyForBottom.class, locations = TypeUseLocation.LOWER_BOUND)
+package java;
+
+import  org.checkerframework.framework.qual.*;
```
wmdietl: I would replace the wildcard imports with the three specific imports.

checker/tests/nullness/Issue1922.java
```diff
@@ -0,0 +1,32 @@
...
+            if (target.equals(key)) return key;
+        }
+        return null;
+    }
+
+    // Find a key in a map and return String version of its value.
+    public static String findKeyAndFetchString(Map<String, Object> someMap) {
+        // :: error: (type.argument.type.incompatible)
+        @Nullable @KeyFor("someMap") String myKey = Issue1922.<@KeyFor("someMap") String>findKey(someMap.keySet(), "Foo");
+        // If I change above line to this line, I do get an error:
```
wmdietl: This comment is misleading, as the above line does give an error.
I would either make the next line a separate test or remove it.

checker/tests/nullness/Issue1922.java
```diff
@@ -0,0 +1,32 @@
...
+        //@Nullable @KeyFor("someMap") String myKey = null;
+
+        Object value = someMap.get(myKey);
+        // value will be null here, but checker gives no error:
```
wmdietl: Comment also no longer true.

checker/tests/nullness/TestPolyAll2.java
```diff
@@ -12,7 +12,7 @@ public static boolean noDuplicates1(@PolyAll @NonNull @UnknownKeyFor String[] a)
         return true;
     }
 
-    public static boolean noDuplicates2(@PolyAll @Nullable String[] a) {
+    public static boolean noDuplicates2(@PolyAll @UnknownKeyFor @Nullable String[] a) {
```
wmdietl: I don't understand why this change is required. Isn't `@UnknownKeyFor` the default?

smillst: Without `@UnknownKeyFor`, the `@PolyAll` is that annotation in the Key For hierarchy.  Line 17, `a[0] = null` issue an error without this annotation.

wmdietl: Ah, true. Let's hope not many people use PolyAll, otherwise they will get many unexpected warnings.

smillst:
Add change log entry?

docs/manual/map-key-checker.tex
```diff
@@ -49,6 +49,19 @@ \section{Map key annotations\label{map-key-annotations}}
...
 \end{figure}
 
+\section{Default annotations\label{map-key-defaults}}
+
+The null literal is implicitly annotated with \code{@UnknownKeyFor}, because
+\code{null} might be a key for any map.
+
+Lower bounds are defaulted to \code{@KeyForBottom}.
+This reduces the number of false
+positives, but it does require you to take action if you expect a
```
mernst: What does it mean to "expect"?

docs/manual/map-key-checker.tex
```diff
@@ -49,6 +49,19 @@ \section{Map key annotations\label{map-key-annotations}}
...
+type variable to be instantiated with \code{@KeyFor} for a particular map.
+In that case,
+you must explicitly annotate the lower bound of that type variable with
```
mernst: Is this at its declaration site or use site?

docs/manual/map-key-checker.tex
```diff
@@ -49,6 +49,19 @@ \section{Map key annotations\label{map-key-annotations}}
...
+\code{@KeyForBottom}.  For example, \code{List<@KeyForBottom T>}.
```
mernst: Can you expand this into a complete example?  This doesn't show a full use or clarify the question about whether to write at a declaration or use site.  I think an example will help to clarify the issue.

smillst:
Add note to manual about collections that don't contain null.  (There lower bounds can be changed to `@KeyForBottom`.)

PR #1977, opened 2018-05-10
[GuiEffect] Add support for UI effect inference for anonymous inner classes.

lazaroclapp:
We use the method being overriden and its direct invocations to determine when to trigger inference.

If an anonymous inner class overrides a `@PolyUIEffect` method of a `@PolyUIType` class, and then calls an `@UIEffect` method from it, we infer that the anonymous inner class declaration is `@UI`.

This is similar to the `@UI` polymorphic effect inference added for lambdas in #1481. 

Test cases added to `AnonInnerDefaults.java`.

Note that we found this support to be critical for using the GuiEffect checker past release 2.5.0. After Checker Framework 2.5.0, use of annotations in comments is deprecated (e.g. `new /*@UI*/ MyClass {...}`) and the CF javac fork stopped being updated. Using type use annotations directly for this checker is very dangerous, due to an existing javac bug that will silently generate broken code when type use annotations are used on Anonymous Inner classes (see https://bugs.openjdk.java.net/browse/JDK-8059531).


lazaroclapp:
cc: @msridhar @csgordon @wmdietl

csgordon:
That is a pretty unfortunate Java bug to have been open for 3+ years. I'll try to take a look at this over the weekend. Have you been using the CF javac fork all this time to work around this JDK bug?

msridhar:
> That is a pretty unfortunate Java bug to have been open for 3+ years. I'll try to take a look at this over the weekend. Have you been using the CF javac fork all this time to work around this JDK bug?

We only transitioned to Java 8 for our Android codebase a few months ago.  But yes, since that time, we have had to use the CF javac fork + annotations in comments because of this JDK bug.

mernst:
@csgordon Have you had a chance to look at this?

checker/tests/guieffect/TestProgram.java
```diff
@@ -53,11 +53,13 @@ public void doGenericStuff() { // Should be inst. w/ @UI
...
                 });
         safecons.runAsync(
+                // :: error: (argument.type.incompatible)
                 new IGenericTask() {
                     @Override
                     public void doGenericStuff() {
                         // Test that the package annotation works
-                        // :: error: (call.invalid.ui)
+                        // Safe here due to anonymous inner class effect inference, but will trigger an error above
+                        // due to safecons.runAsync not taking an @UI IGenericTask
```
csgordon: This is a perfectly reasonable thing to test, but this particular change effectively *removes a test that should still work.*  This test used to check that unannotated methods of a ```@UIPackage``` did in fact have the UI effect.  I'd prefer that you preserve the existence of such a test, though I'm agnostic as to whether that occurs by reverting this test and creating a new one for what this now checks, or by keeping this change and (re)creating a new version of the old test.

checker/tests/guieffect/AnonInnerDefaults.java
```diff
@@ -57,13 +106,108 @@ public void doStuff() {
...
+                    public void doStuff() {
+                        // Safe due to anonymous inner class effect inference
+                        e.dangerous(); // should be okay
+                    }
+                });
+        callSafePolyIface(getSafePolyIface(e));
+        callUIPolyIface(getUIPolyIface(e));
+
+        // ToDo: To decide, this case and whether or not we base on @PolyUIType for super type or @PolyUIEffect for
+        // overriden method.
```
csgordon: Was this decided, or is this TODO stale?

checker/tests/guieffect/TestProgram.java
lazaroclapp: Sure, my bad! The best way to test that now, is to copy these two calls outside of any anonymous inner class (or into one where the FI is not `@PolyUIType` with an `@PolyUIEffect` method), so that the inference becomes irrelevant to the enclosing method. I'll do the "fully outside of any class option", unless you prefer that I create a subinterface of `IGenericTask` that is forced to be safe? :)

checker/tests/guieffect/AnonInnerDefaults.java
lazaroclapp: No clue why that TODO is here, of all places. It has nothing to do with the next test case. Removing it now. But, for the record, this was decided to be based on overriden method according to whatever `AnnotatedTypes.overriddenMethods` returns 😃 (but the declaring class must also be annotated `@PolyUIType`, we are fairly conservative about turning on inference)

lazaroclapp:
@csgordon Hey, sorry for the delay. I just had time to look at the comments now. Will update this addressing those two issues, and then take a look at the documentation. I can definitely add some info about effect inference there and add it to this PR so it lands atomically.

docs/manual/guieffect-checker.tex
```diff
@@ -66,7 +66,7 @@ \chapter{GUI Effect Checker\label{guieffect-checker}}
 @SafeEffect
 public void calledFromBackgroundThreads(JLabel l) \{
     l.setText("Foo");       // Error: calling a @UIEffect method from a @SafeEffect method
-    Display.syncExec(new @UI Runnable \{
+    Display.syncExec(new Runnable \{
```
lazaroclapp: There is actually not a lot about `@UI` as a type use annotation. Unless you mean some sort of more general Checker Framework instruction elsewhere in the docs? I just had to change this example, you still need `@UI` if storing the class somewhere as described below.

docs/manual/guieffect-checker.tex
```diff
@@ -277,6 +277,10 @@ \subsection{Using an effect-polymorphic type\label{guieffect-polymorphism-using}
 
 It is an error to apply an effect instantiation qualifier to a type that is not effect-polymorphic.
 
+Note that no annotation is required on the anonymous class declaration itself (e.g. \code{new Runnable()\{...\}} 
+does not require a type use annotation, although the variable, field, or argument it ends up being assigned to might).
+Instead, the GUI Effect Checker will infer the effect qualifier based on the method being called from within the 
+members of that specific anonymous class.
```
lazaroclapp: To make things explicit, here is a note on the type use annotation being no longer needed. Let me know if there is anything else missing, @csgordon? (Also, might need to validate the LaTeX still, what is the target for building the manual, again? 😅 )

lazaroclapp:
@csgordon Sorry for the spam, but I believe I also addressed all your comments here, let me know if something is not quite right, though, and I'll work on that!

csgordon:
Not spam at all; I lost track of this and appreciate the ping.

csgordon:
@smillst @wmdietl Any reason to hold off on merging this?

mernst:
It looks OK to me.

checker/src/main/java/org/checkerframework/checker/guieffect/GuiEffectTypeFactory.java
```diff
@@ -55,6 +55,8 @@
      */
     protected final Set<LambdaExpressionTree> uiLambdas = new HashSet<>();
 
+    private final Set<TypeElement> uiAnonClasses = new HashSet<>();
```
wmdietl: Some javadoc similar to that of `uiLambdas` above would probably be useful.
Should this field also be `protected` for consistency with `uiLambdas`?

checker/src/main/java/org/checkerframework/checker/guieffect/GuiEffectTypeFactory.java
```diff
@@ -550,6 +562,11 @@ public void constrainLambdaToUI(LambdaExpressionTree lambdaExpressionTree) {
         uiLambdas.add(lambdaExpressionTree);
     }
 
+    public void constrainAnonymousClassToUI(TypeElement callerReceiverElt) {
```
wmdietl: Maybe also adapt a copy of the `constrainLambdaToUI` javadoc?

checker/src/main/java/org/checkerframework/checker/guieffect/GuiEffectVisitor.java
```diff
@@ -222,20 +225,33 @@ private void lambdaAssignmentCheck(
...
         Void result = super.visitVariable(node, p);
-        if (node.getInitializer() != null
-                && node.getInitializer().getKind() == Tree.Kind.LAMBDA_EXPRESSION) {
-            lambdaAssignmentCheck(
-                    atypeFactory.getAnnotatedType(node),
-                    (LambdaExpressionTree) node.getInitializer(),
-                    "assignment.type.incompatible");
+        if (node.getInitializer() != null) {
+            if (node.getInitializer().getKind() == Tree.Kind.NEW_CLASS) {
+                commonAssignmentCheck(
```
wmdietl: Why is this special logic needed? Isn't super already checking that an initializer has a suitable type for the variable?
Why only if the top-level expression is NEW_CLASS and not if e.g. there is a ternary expression that contains a NEW_CLASS?

checker/src/main/java/org/checkerframework/checker/guieffect/GuiEffectVisitor.java
```diff
@@ -222,20 +225,33 @@ private void lambdaAssignmentCheck(
...
+                        atypeFactory.getAnnotatedType(node),
+                        atypeFactory.getAnnotatedType(node.getInitializer()),
+                        node.getInitializer(),
+                        "assignment.type.incompatible");
+            } else if (node.getInitializer().getKind() == Tree.Kind.LAMBDA_EXPRESSION) {
+                lambdaAssignmentCheck(
```
wmdietl: I also don't understand why that previous additional check is needed. Maybe some javadoc could explain this?

checker/src/main/java/org/checkerframework/checker/guieffect/GuiEffectVisitor.java
```diff
@@ -322,6 +349,51 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
...
+                                    != null
+                            && atypeFactory.getDeclAnnotation(
+                                            overriddenType.getUnderlyingType().asElement(),
+                                            PolyUIType.class)
+                                    != null) {
+                        overridesPolymorphic = true;
+                        break;
+                    }
+                }
+                // Perform anonymous class polymorphic effect inference:
```
wmdietl: Why is effect inference in the Visitor instead of in the AnnotatedTypeFactory?

wmdietl:
@lazaroclapp Did you get a chance to look at my comments?

lazaroclapp:
@wmdietl Hey, sorry for the wait, I haven't yet had a chance to context-switch to this. Will try to take a look at it again this week. Thanks for both looking and this and for the reminder, though!

checker/src/main/java/org/checkerframework/checker/guieffect/GuiEffectTypeFactory.java
lazaroclapp: Done in next commit.

checker/src/main/java/org/checkerframework/checker/guieffect/GuiEffectTypeFactory.java
lazaroclapp: Done in next commit.

checker/src/main/java/org/checkerframework/checker/guieffect/GuiEffectVisitor.java
lazaroclapp: @wmdietl There are two parts to this:

1) The reason this special logic is needed is because inner class and lambda effect inference happen during the visit. Basically, I only know if a lambda/class should be inferred to be `@UI` after I have visited its methods and seen if they call in turn a method which has `@UIEffect` (either directly or based on e.g. polymorphism). Because of the order of the traversal in `BaseTypeVisitor`, I believe the lambda/class won't have been marked `@UI` by the time the variable assignment is evaluated inside the call to `super.visitVariable(...)`.

2) You are absolutely right, this breaks down for ternary expressions, and I suspect other expressions which can contain a new class as well. I am currently trying to figure this out, since even just unconditionally calling `commonAssignmentCheck` for all nodes is not enough for a ternary expression that includes one instance with inferred `@UI` type and one instance that is `@AlwaysSafe`

checker/src/main/java/org/checkerframework/checker/guieffect/GuiEffectVisitor.java
lazaroclapp: See comment above. Can definitely add a javadoc or an inline comment block if the explanation above makes sense to you.

checker/src/main/java/org/checkerframework/checker/guieffect/GuiEffectVisitor.java
lazaroclapp: @wmdietl The reason for this is that effect inference depends on visiting the method invocations inside the lambda body or the methods of the anonymous inner class for which inference is being performed. As far as I know, this information is only present during the Visitor traversal.

There was a lot of talk in the original lambda inference PR about using `GuiEffectTreeAnnotator`, but I couldn't get it to work, either because some necessary info was missing (`visitorState`, see [this discussion](https://github.com/typetools/checker-framework/pull/1481#discussion_r135813364)), or because certain nodes weren't being visited at all (e.g. `visitMethodInvocation` see [this discussion](https://github.com/typetools/checker-framework/pull/1481#issuecomment-326118260)). If there is a better way to do this, I would love to know of it, though.

lazaroclapp:
@wmdietl Once more, sorry for the delay, I haven't had the time to context switch into this for a while. That said, I just had a chance to look at your comments, and I replied inline. Thanks for looking into this!

checker/src/main/java/org/checkerframework/checker/guieffect/GuiEffectTypeFactory.java
wmdietl: @lazaroclapp Sorry for the delay on this. Do you have a local commit with more javadoc and forgot to push?
Thanks for commenting on the other issues and finding the earlier discussions. I would suggest shelving those cleanups for a future PR.
Please add the javadoc for the new functionality and maybe add some comments for the other parts; then merge master and we can merge this.

lazaroclapp:
Edit: I see that Travis is failing, but I don't understand the error. Looks unrelated, as it shows a bunch of compile errors inside `/annotation-tools/asmx`. Is there something I am missing there? (`./gradlew build` succeeds)

@wmdietl Hey Werner, I ended up having to make a more significant change to this PR than I thought, but now it handles ternary expressions, and should be easy to patch for other expressions that contain anonymous classes or lambdas. The fix is that, rather than checking twice when visiting assignments, method invocations, etc, instead, whenever it finishes visiting an anonymous inner class or lambda, it traverses the `TreePath` up from there, re-doing any `commonAssignmentCheck` type check that might involve that class/lambda. See the `scanUp` method.  Let me know if this approach makes sense. Because it is only done for newly inferred classes/lambdas, and it stops at the containing method / initializer boundary, it seems fairly innocuous from a performance standpoint.

I added the test case, and I believe all methods either override an existing one or have a javadoc comment.

Also, heads up, we are trying to get our Rx Thread/Effect Checker code open-sourced by September 3rd to match our talk on ASE 2018 that week. Thus, I would really like to be able to get this PR and  (for treating Android's `@MainThread` annotation as implying `@UIEffect`) into the next Checker Framework release. If that happens, we can have our checker depend on that version, rather than requiring a manually patched Checker Framework or a fork. This is all pending internal approval at Uber too, of course.

cc: @msridhar

csgordon:
@wmdietl @mernst @smillst any reason not to merge now?

lazaroclapp:
@wmdietl @csgordon @mernst Thanks so much for all the help, everyone! We are all super happy over here to have this one upstreamed 👍

mernst:
Thanks for the contribution.  We appreciate your improvements to the Checker Framework!

PR #1980, opened 2018-05-16
Fixes asking credentials for non-existent repository

Ravi-Roshan:
The `git ls-remote https://github.com/USERNAME/REPONAME.git` in the .travis-build-without-test.sh asks for username and password when looking for a non-existent repository. Changing it to `wget https://github.com/USERNAME/REPONAME.git` solves the issue. Here the purpose of `git ls-remote` is to check the existence of the repository. This purpose is fulfilled by `wget` too.


.travis-build-without-test.sh
```diff
@@ -31,8 +31,8 @@ if [ -d ../annotation-tools ] ; then
     echo "... done: (cd ../annotation-tools && git pull)"
 else
     set +e
-    echo "Running: git ls-remote https://github.com/${SLUGOWNER}/annotation-tools.git &>-"
-    git ls-remote https://github.com/${SLUGOWNER}/annotation-tools.git &>-
+    echo "Running: wget https://github.com/${SLUGOWNER}/annotation-tools.git &>-"
```
wmdietl: Please change all calls of `wget` to `wget  -q --spider`. Put that in a variable to not copy many times.
This will only check for the file, not actually download it.

Note that the PR is currently failing a Travis test, b/c the downloaded files contain trailing whitespaces.
In general we don't want these extra files lying around, so let's change the command to not do that.

.travis-build-without-test.sh
```diff
@@ -24,15 +24,16 @@ if [[ "$SLUGOWNER" == "" ]]; then
 fi
 
 ## Build annotation-tools (Annotation File Utilities)
+WGETCMD="wget -q --spider"
```
wmdietl: How about moving this before the annotations-tools comment, b/c this isn't just used for AFU.
How about naming it 'GITEXISTS' as a more descriptive name of what we want to use this for and a short comment.
ls-remote is also used in `.travis-build.sh`, so please also update that file.

.travis-build.sh
```diff
@@ -48,11 +48,14 @@ fi
 # subsequent command to build it except to test building it.
 
 set -e
+# Checks for the existence of a repository
+GITEXISTS="wget -q --spider"
```
wmdietl: Instead of duplicating this definition here, is the one from `.travis-build-without-test.sh` visible?
Maybe `export` the variable there?

wmdietl: Ping.

PR #1985, opened 2018-05-18
[Index Checker] Add offsets to SameLen

kelloggm:
This pull request adds offsets to `SameLen`, using the machinery of `OffsetEquation` in a manner similar to `UBQualifier` in the Upper Bound Checker.

Though offsets are never inferred for SameLen annotations (for now), they can be written and are taken into account when typechecking array accesses in the Upper Bound Checker. This will allow us to write annotations such as `@SameLen("this", "start")` on the arrays backing Guava data structures, allowing us to typecheck the actual implementations of various methods correctly with only a single suppressed warning per collection. This *should* reduce the number of false positives in Guava by ~40, though I haven't yet run those tests (I will on Monday).

Inference of these offsets should be possible in the future, allowing us to handle all the cases in kelloggm#158. This pull request lays the foundation for that.

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/index/samelen/SameLenQualifier.java
+++ b/checker/src/main/java/org/checkerframework/checker/index/samelen/SameLenQualifier.java
@@ -14,13 +14,13 @@ import javax.annotation.processing.ProcessingEnvironment;
 import javax.lang.model.element.AnnotationMirror;
 import org.checkerframework.checker.index.qual.SameLen;
 import org.checkerframework.checker.index.upperbound.OffsetEquation;
+import org.checkerframework.checker.nullness.qual.Nullable;
 import org.checkerframework.javacutil.AnnotationBuilder;
 import org.checkerframework.javacutil.AnnotationUtils;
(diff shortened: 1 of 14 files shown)


checker/src/main/java/org/checkerframework/checker/index/samelen/SameLenQualifier.java
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    /**
+     * If the sequences are disjoint, returns bottom. Otherwise, returns the a combined SameLen
+     * qualifier.
+     */
+    public AnnotationMirror glb(
+            SameLenQualifier other, AnnotationMirror bottom, ProcessingEnvironment env) {
+        if (!Collections.disjoint(sequences, other.sequences)) {
+            return this.combine(other).convertToAnnotation(env);
```
smillst: Will this trigger the assert in combine?  `@SameLen("a", offset = "b")` `@SameLen("a")`

mernst:
This won't be merged.

PR #1989, opened 2018-05-22
Improve count-suppressions [ci skip]

panacekcz:
- use `bash` to run the script, because `[[` may be unavailable in `sh`
- ignore lines ending with a semicolon


mernst:
Vlastimil, have you confirmed that this script produces the same results for all our case studies:
 * currently
 * if you make this change and also removing the two undesired comments
?

mernst:
@panacekcz , a ping on my question above.  I'd like to get this merged.
Thanks!

panacekcz:
Trying it on case studies, this would exclude asserts such as
`assert con1.val != null : "@AssumeAssertion(nullness)";`.
I will try to get these back.

panacekcz:
Warning suppression ending with a semicolon for unknown reason:
https://github.com/mernst/daikon/blob/ec4f7b13a54d55d6ed7259820f7f149b5d9633c5/java/daikon/tools/compare/LogicalCompare.java#L281

panacekcz:
I have changed the regular expression so that:
- lines with `AssumeAssertion` are not affected
- otherwise, a line is included if after `"keyword"` or `"keyword:`, there is `//` or end of line following  immediately or occuring preceded by a character other than `;`

Examples:

- `"index"` included
- `"index" // something with ;` included
- `f("index");` excluded
- `f("index"); // call f` included, although it shouldn't be

panacekcz:
@mernst, I have compared the output of the new and old script using `count-suppressions nullness` and `count-suppressions index` on these case studies:
- typetools/guava, master
- kelloggm/jfreechart, master
- typetools/java-getopt, master
- typetools/commons-bcel, typecheck-nullness
- typetools/commons-io, master
- mernst/plume-lib, master
- mersnt/daikon, master, check-index

In guava, jfreechart and daikon, the new script omits a few lines of code where the string literal `"index"`, `"lowerbound"`, or `"upperbound"` is used in a method call, so the output is improved. In the other projects, the output is the same.

PR #1990, opened 2018-05-22
Move some methods of SignednessUtil to new class

### LINKED ISSUE #1978: checker-qual is not Android lint compatible
When using the checker-qual library in Andorid libraries, the Android lint tool issues this error even though we just use the annotations:
```
Error: Invalid package reference in library; not included in Android: java.awt. Referenced from org.checkerframework.checker.signedness.SignednessUtil. [InvalidPackage]
```

This causes problems when shipping our library, as we either need to require every user to add a lint suppression for this, or we need to remove the dependency before shipping the code. See for example, [here](https://github.com/google/ExoPlayer/issues/4234) for the issue reported for our library.

I was wondering if
 - The checker-qual library can be made Android compatible by only using Java classes also available on Android (that is: not java.awt).
 - Or by providing an annotations only version without the helper classes as already requested in #1510. 

Thanks!

smillst:
Android doesn't include java.awt packages, so including references to those packages cause the Android lint tool to fail.  The only references to those packages are in SignednessUtil, so this pull request moves them to a new class that isn't included in checker-qual.jar.

Fixes #1978.

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/signedness/SignednessUtil.java
+++ b/checker/src/main/java/org/checkerframework/checker/signedness/SignednessUtil.java
@@ -10,7 +10,9 @@ import org.checkerframework.checker.signedness.qual.Unsigned;
 /**
  * Provides static utility methods for unsigned values. Some re-implement functionality in JDK 8,
  * making it available in earlier versions of Java. Others provide new functionality. {@link
- * SignednessUtilExtra} has more methods that reference
+ * SignednessUtilExtra} has more methods that reference packages that Android does not provide.
+ *
+ * @checker_framework.manual #signedness-utilities Utility routines for manipulating unsigned values
  */
 public final class SignednessUtil {
 
--- a/checker/src/main/java/org/checkerframework/checker/signedness/SignednessUtil.java
+++ b/checker/src/main/java/org/checkerframework/checker/signedness/SignednessUtil.java
@@ -1,7 +1,5 @@
 package org.checkerframework.checker.signedness;
 
-import java.awt.Dimension;
-import java.awt.image.BufferedImage;
 import java.io.IOException;
 import java.io.RandomAccessFile;
 import java.math.BigInteger;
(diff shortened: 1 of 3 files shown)


checker/src/main/java/org/checkerframework/checker/signedness/SignednessUtil.java
```diff
@@ -11,26 +9,15 @@
 
 /**
  * Provides static utility methods for unsigned values. Some re-implement functionality in JDK 8,
- * making it available in earlier versions of Java. Others provide new functionality.
+ * making it available in earlier versions of Java. Others provide new functionality. {@link
+ * SignednessUtilExtra} has more methods that reference
```
wmdietl: That sentence ends abruptly...

docs/manual/signedness-checker.tex
```diff
@@ -153,3 +153,7 @@ \section{Utility routines for manipulating unsigned values\label{signedness-util
 properly annotated with \refqualclass{checker/signedness/qual}{Unsigned}
 where appropriate, so using them may reduce the number of annotations that
 you need to write.
+
+Class \refclass{checker/signedness}{SignednessUtilExtra} contains more utility
+methods that reference packages not included in Android.  This class is not
+included in checker-qual.jar, so you may want to copy the methods to your code.
```
wmdietl: Use `\code{}` for file name?

checker/src/main/java/org/checkerframework/checker/signedness/SignednessUtilExtra.java
```diff
@@ -0,0 +1,63 @@
+package org.checkerframework.checker.signedness;
+
+import java.awt.Dimension;
+import java.awt.image.BufferedImage;
+import org.checkerframework.checker.signedness.qual.Unsigned;
+
+/**
+ * Provides more static utility methods for unsigned values. These methods use Java packages not
+ * included in Android. {@link SignednessUtil} has more methods.
```
wmdietl: I would include more of the text from the manual, to make the motivation for this class clear.

mernst: Or, reference the manual.  That way avoids text duplication and the possibility that the two versions get out of sync.

Issue #1991, opened 2018-05-24
AsSuperVisitor: type is not an erased subtype of supertype 16


cushon:
This reproduces at head (82702ae4498f25afd21a351cd48b68f1fa7c6a4f).

```java
import java.io.Serializable;

abstract class Z {

  abstract void g(A obj);

  static class A {
    A(C<?, ?> c) {}
  }

  interface B extends C<D, E> {}

  interface C<X extends Comparable<? super X>, Y extends Serializable> {}

  public class E implements Serializable {}

  static abstract class D implements Comparable<D>, Serializable {}

  void f(B b) {
    g(new A(b));
  }
}
```

```
./checker-framework/checker/bin/javac -processor Nullness -AprintErrorStack Test.java
error: AsSuperVisitor: type is not an erased subtype of supertype.
  type: Comparable<? extends Object super X extends Comparable<?>>
  superType: D
  Compilation unit: X.java
  Last visited tree at line 20 column 7:
      g(new A(b));
  Exception: java.lang.Throwable; Stack trace: org.checkerframework.framework.source.SourceChecker.errorAbort(SourceChecker.java:739)
  org.checkerframework.javacutil.ErrorReporter.errorAbort(ErrorReporter.java:41)
  org.checkerframework.framework.type.AsSuperVisitor.errorTypeNotErasedSubtypeOfSuperType(AsSuperVisitor.java:154)
  org.checkerframework.framework.type.AsSuperVisitor.visitDeclared_Declared(AsSuperVisitor.java:339)
  org.checkerframework.framework.type.AsSuperVisitor.visitDeclared_Declared(AsSuperVisitor.java:28)
  org.checkerframework.framework.util.AtmCombo.accept(AtmCombo.java:308)
  org.checkerframework.framework.type.visitor.AbstractAtmComboVisitor.visit(AbstractAtmComboVisitor.java:59)
  org.checkerframework.framework.type.AsSuperVisitor.visit(AsSuperVisitor.java:85)
  org.checkerframework.framework.type.AsSuperVisitor.asSuper(AsSuperVisitor.java:67)
...
error: AsSuperVisitor: type is not an erased subtype of supertype.
  type: @Initialized @NonNull Comparable<? extends @Initialized @NonNull Object super X extends @Initialized @NonNull Comparable<?>>
  superType: @Initialized @NonNull D
  Compilation unit: X.java
  Last visited tree at line 20 column 7:
      g(new A(b));
[...]

PR #1994, opened 2018-05-25
Fix Issue #1991 by ignoring type variables in lower bounds of wildcards.

wmdietl:


checker/tests/nullness/java8/Issue579.java
```diff
@@ -18,7 +18,7 @@ public int compare(T a, T b) {
 
     @Override
     public Comparator<T> thenComparing(Comparator<? super T> other) {
-        // :: error: (argument.type.incompatible) :: warning: (known.nonnull)
+        // :: warning: (known.nonnull)
```
wmdietl: The error that I get here originally:

```
checker/tests/nullness/java8/Issue579.java:22: error: [argument.type.incompatible] incompatible types in argument.
        return new Issue570<>(real == null ? other : real.thenComparing(other));
                                           ^
  found   : @Initialized @NonNull Comparator<capture#196 of ? super T extends @Initialized @Nullable Object super T extends @Initialized @Nullable Object>
  required: @Initialized @NonNull Comparator<? extends @Initialized @Nullable Object super T extends @Initialized @Nullable Object>
1 error
```

Was this really an expected error or was this a false positive?

framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java
```diff
@@ -365,7 +365,12 @@ protected boolean isContainedBy(
                 return false;
             }
 
-            return canBeCovariant || checkAndSubtype(outsideWc.getSuperBound(), inside);
+            AnnotatedTypeMirror lowerbound = outsideWc.getSuperBound();
+            if (lowerbound.getKind() == TypeKind.TYPEVAR) {
+                // TODO: determine correct check for (captured) type variables.
```
wmdietl: This fixes the crash, but is an ugly work-around.
Anyone have a better suggestion?

PR #2001, opened 2018-05-29
 Don't issue Lower Bound Checker Warnings on List and other classes.

smillst:
This is a follow up to #1716.

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/index/lowerbound/LowerBoundChecker.java
+++ b/checker/src/main/java/org/checkerframework/checker/index/lowerbound/LowerBoundChecker.java
@@ -1,5 +1,6 @@
 package org.checkerframework.checker.index.lowerbound;
 
+import java.util.HashSet;
 import java.util.LinkedHashSet;
 import org.checkerframework.checker.index.inequality.LessThanChecker;
 import org.checkerframework.checker.index.searchindex.SearchIndexChecker;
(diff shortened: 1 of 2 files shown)


kelloggm:
I have two concerns with this:

1. There's nothing in this pull request that shows _why_ we need to disable these warnings for `List` and `AbstractList`. At the very least can you add a test showing a false positive derived from them?
2. We ought to be type-checking that the argument to `List#get` is non-negative, anyway. We have the ability to do so, so we should use it. We've found some bugs this way (I can think of at least one in JFreeChart), so disabling it at this point to reduce false positives seems a little sketchy.

smillst:
Here's one example of a false positive.
```
import java.util.ArrayList;
import java.util.List;

public class FalsePositive {
    List<String> list = new ArrayList<>();
    String getLast() {
        if (!list.isEmpty()) {
            return list.get(list.size() - 1);
        }
        return "";
    }
}
```

Issues:
```
: error: [argument.type.incompatible] incompatible types in argument.
            return list.get(list.size() - 1);
                                        ^
  found   : @GTENegativeOne int
  required: @NonNegative int
```

There are 8 false positives in JFreeChart because of this. (None in plume-lib or guava) One of the false positives in JFreeChart is probably the bug to which you refer.  If you think it is better to have the 1 bug and 7 false positives, we don't have to merge this right now.   However, for real users, it's very confusing for the Index Checker to issue warning on Lists.  Neither you nor I remembered why List had `@IndexFor` annotations, so we really can't expect others to understand.

panacekcz:
There are more classes in the JDK where the same reasoning would apply, for example `ArrayList`, `LinkedList`, `Vector`, `Bitset`.

PR #2002, opened 2018-05-30
Warn about unneeded suppressions

### LINKED ISSUE #1701: warn about unnecessary suppressions
When upgrading a large code base to use the latest version of the checker it's sometimes necessary to add `@SuppressWarnings` for newly reported errors and investigate them later, instead of coordinating a large number of fixes with the upgrade. This makes it easy to accumulate suppressions, some of which were work-arounds for false positives that are unnecessary in the latest version of the checker. Leaving the annotations in place clutters up code and makes the checker less effective.

It would be helpful to have a mode where the checker could report unnecessary `@SuppressWarnings` annotations.

smillst:
Fixes #1701.


framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
```diff
@@ -497,14 +505,99 @@ public void typeProcess(TypeElement element, TreePath tree) {
...
+                }
+                if (keyFromAnno.contains(":")) {
+                    // The key starts with a checker name, if that is this checker, then the warning
+                    // was issued above.  For example, if this is the Nullness Checker and the
+                    // keyForAnno is "index:override.return.invalid", then don't issue a warning.
+                    continue;
+                }
+
+                for (String errorKey : errorKeys) {
+                    if (keyFromAnno.equals(errorKey)) {
```
wmdietl: Some documentation would be helpful to understand what these three nested for loops each do.
Why can the `checkerKeys` and `errorKeys` lists not be merged? Why is one using `contains` and the other `equals`?

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
```diff
@@ -497,14 +505,99 @@ public void typeProcess(TypeElement element, TreePath tree) {
...
+            subChecker.elementsWithSuppressedWarnings.clear();
+            checkerKeys.addAll(subChecker.getSuppressWarningsKeys());
+            errorKeys.addAll(subChecker.messages.stringPropertyNames());
+            subChecker.getVisitor().treesWithSuppressWarnings.clear();
+        }
+        // It's not clear for which checker this suppression is intended,
+        // so never report it as unused.
+        checkerKeys.remove(SourceChecker.SUPPRESS_ALL_KEY);
+
+        for (Tree tree : getVisitor().treesWithSuppressWarnings) {
```
wmdietl: I can write `@SuppressWarnings("purity")` to suppress all purity warnings.
However, `"purity"` isn't a message key. Would this logic notice if it is unnecessary?

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
```diff
@@ -497,14 +505,99 @@ public void typeProcess(TypeElement element, TreePath tree) {
...
+                if (keyFromAnno.contains(":")) {
+                    // The key starts with a checker name, if that is this checker, then the warning
+                    // was issued above.  For example, if this is the Nullness Checker and the
+                    // keyForAnno is "index:override.return.invalid", then don't issue a warning.
+                    continue;
+                }
+
+                for (String errorKey : errorKeys) {
+                    if (keyFromAnno.equals(errorKey)) {
+                        Tree swTree = findSWTree(tree);
```
wmdietl: Can this be rewritten to share the common code with the the error report above?

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
```diff
@@ -497,14 +505,99 @@ public void typeProcess(TypeElement element, TreePath tree) {
...
+                                        "\"" + keyFromAnno + "\""),
+                                swTree);
+                    }
+                }
+            }
+        }
+        getVisitor().treesWithSuppressWarnings.clear();
+    }
+
+    private Tree findSWTree(Tree tree) {
```
wmdietl: Some documentation would be helpful to understand what this is doing. Sometime it returns annotation trees, sometime just the input parameter.
Wouldn't it be nicer to use a more precise return type and raise an error if the unexpected `return tree` happens?
But maybe I misunderstand the purpose.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
```diff
@@ -497,14 +505,99 @@ public void typeProcess(TypeElement element, TreePath tree) {
...
+        if (TreeUtils.isClassTree(tree)) {
+            annotations = ((ClassTree) tree).getModifiers().getAnnotations();
+        } else if (tree.getKind() == Kind.METHOD) {
+            annotations = ((MethodTree) tree).getModifiers().getAnnotations();
+        } else {
+            annotations = ((VariableTree) tree).getModifiers().getAnnotations();
+        }
+
+        for (AnnotationTree annotationTree : annotations) {
+            if (annotationTree.getAnnotationType().toString().equals("SuppressWarnings")) {
```
wmdietl: Why not use the more precise and faster class literal for the comparison?

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -173,6 +173,10 @@
     /** An instance of the {@link ContractsUtils} helper class. */
     protected final ContractsUtils contractsUtils;
 
+    /**
+     * A set of trees that are annotated with {@code @SuppressWarnings}/
+     *//*package-private*/ final List<Tree> treesWithSuppressWarnings;
```
wmdietl: The formatting here looks a bit strange... not sure if there is something you can do to fix.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -1449,6 +1458,12 @@ public boolean useUncheckedCodeDefault(String kindOfCode) {
         return false;
     }
 
+    /**
+     * Elements with a {@code @SuppressWarnings} that actually suppressed a warning for this
+     * checker.
+     */
+    public final Set<Element> elementsWithSuppressedWarnings = new HashSet<>();
```
wmdietl: It seems very strange that some parts of this feature are implemented on the SourceChecker level and some on the BaseTypeChecker/BaseTypeVisitor level.
Wouldn't it be more consistent to pick the right level and implement the feature there?
You can have a SourceChecker subtype that doesn't use a BaseTypeVisitor subtype, for example.

smillst: I wanted to implement it at the BaseTypeVisitor level, but I couldn't, because collecting elements with a `@SuppressWarnings` that actually suppressed a warning has to happen in the `Source Checker#shouldSuppressWarnings(javax.lang.model.element.Element, java.lang.String)`.  If the BaseTypeVisitor overrides that method and adds an element to `elementsWithSuppressedWarnings` if the super call returns true, then an element will be added for each recursive call to  `shouldSuppressWarnings`.   (Unless I've missed something here.)

I could implement at the Source Checker level, too, but the BaseType Checker would have to override that implementation to correctly implement it for compound checkers.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
smillst: Good point.  It would not have, but I fixed it.  (I also add a test case.)

checker/jtreg/index/UnneededSuppressionsTest.out
```diff
@@ -0,0 +1,6 @@
+UnneededSuppressionsTest.java:9:1: compiler.warn.proc.messager: (unneeded.suppression)
+UnneededSuppressionsTest.java:22:5: compiler.warn.proc.messager: (unneeded.suppression)
```
wmdietl: Is this for the `tainting` or the `lowerbound` suppression?
Can you separate the two to make sure that the `tainting` doesn't raise a warning when running the Index Checker.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -320,6 +326,21 @@ public final Void visitClass(ClassTree classTree, Void p) {
         return null;
     }
 
+    /**
+     * If {@code tree} has a {@code @SuppressWarnings} add it to treesWithSuppressWarnings.
+     *
+     * @param tree a declaration on which a {@code @SuppressWarnings} annotation may be placed.
+     */
+    private void checkForSuppressWarningsAnno(Tree tree) {
```
wmdietl: `checkForXXX` is rather confusing for a void function.
`storeXXX` maybe? Or `storeAnyXXX`?

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
wmdietl: Most warning management happens on the SourceChecker level, so that would seem like the right place to put all this logic.
Maybe you can add a method that can be overridden in BaseTypeChecker to handle compound checkers?

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
```diff
@@ -505,6 +506,31 @@ public void typeProcess(TypeElement element, TreePath tree) {
...
+     */
+    @Override
+    protected void warnUnneededSuppressions() {
+        if (parentChecker != null || !hasOption("warnUnneededSuppressions")) {
+            return;
+        }
+        Set<Element> elementsSuppress = new HashSet<>(this.elementsWithSuppressedWarnings);
+        this.elementsWithSuppressedWarnings.clear();
+        Set<String> checkerKeys = new HashSet<>(getSuppressWarningsKeys());
+        Set<String> errorKeys = new HashSet<>(messages.stringPropertyNames());
```
wmdietl: It would be nice if there were less duplication between the two methods.
It seems important that this method needs to call `this.elementsWithSuppressedWarnings.clear();`. Slightly restructuring to ensure this would be best, either by adding another function or calling `super`. Alternatively, document this requirement.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -192,6 +193,12 @@
     // suppress that warning.
     "showSuppressWarningKeys",
 
+    // Warn about @SuppressWarnings annotations that do not suppress any warnings.
+    // org.checkerframework.common.basetype.BaseTypeChecker.warnUnneededSuppressions
+    // org.checkerframework.framework.source.SourceChecker.shouldSuppressWarnings(javax.lang.model.element.Element, java.lang.String)
+    // org.checkerframework.common.basetype.BaseTypeVisitor.checkForSuppressWarningsAnno
```
wmdietl: This is now in `SourceVisitor`.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -1449,6 +1569,12 @@ public boolean useUncheckedCodeDefault(String kindOfCode) {
         return false;
     }
 
+    /**
+     * Elements with a {@code @SuppressWarnings} that actually suppressed a warning for this
+     * checker.
+     */
+    public final Set<Element> elementsWithSuppressedWarnings = new HashSet<>();
```
wmdietl: Can this field be `protected` now? Should the documentation make it more explicit that usually nobody but the framework needs to modify it?

framework/src/main/java/org/checkerframework/framework/source/SourceVisitor.java
```diff
@@ -28,6 +35,12 @@
     /** The root of the AST that this {@link SourceVisitor} will scan. */
     protected CompilationUnitTree root;
 
+    /** A set of trees that are annotated with {@code @SuppressWarnings}. */
```
wmdietl: Maybe also stronger documentation to not mess with this.

PR #2003, opened 2018-05-30
Use ArrayRange and ArrayLen in LessThan Checker.

smillst:
Fixes https://github.com/kelloggm/checker-framework/issues/225.

The test case was failing because `bar.length` isn't in the Value Checker's store until after `0`.  So the Value Checker doesn't have a value for `bar.length`.  But is does have an ArrayRange for `bar`, so I changed the code to use that.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
@@ -764,11 +764,9 @@ public abstract class GenericAnnotatedTypeFactory<
      * @param tree current tree
      * @param clazz the Class of the annotation
      * @return the annotation on expression or null if one does not exist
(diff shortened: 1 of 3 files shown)


framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -764,11 +764,9 @@ public AnnotationMirror getAnnotationFromJavaExpressionString(
      * @param tree current tree
      * @param clazz the Class of the annotation
      * @return the annotation on expression or null if one does not exist
-     * @throws FlowExpressionParseException thrown if the expression cannot be parsed
      */
     public AnnotationMirror getAnnotationFromReceiver(
-            FlowExpressions.Receiver receiver, Tree tree, Class<? extends Annotation> clazz)
```
kelloggm: Why the change to `GenericAnnotatedTypeFactory`? This seems unrelated to the rest of the change.

PR #2005, opened 2018-05-30
[Index Checker] HasSubsequence annotation

kelloggm:
`@HasSubsequence(a, from, to)` implemented as a declaration annotation. Useful for typechecking Guava with the Index Checker, where it will reduce false positive counts by ~50. For a description of the annotation, see the changes to `index-checker.tex` in this PR.

There are ten main parts of this pull request:

1. The new annotation, `@HasSubsequence`, in `HasSubsequence.java`.
2. Changes to the LowerBound, LessThan, and UpperBound visitor classes to enforce the assignment rules for sequences with `@HasSubsequence`. These are implemented as part of the common assignment check.
3. Changes to the UpperBound visitor's visitAccess method so that it uses the same code as the common assignment check.
4. Changes to the UpperBound visitor's common assignment check to permit `from + x` to be assigned to `LTLengthOf(a)`  when `x` is `IndexFor(b)`, which has the appropriate `HasSubsequence` annotation. Similarly for `x - from`.
5. A new method in UBQualifier that was needed for number 4, which permits removing a sequence, offset pair from a UBQualifier. Used in the common assignment check to partially relax required qualifiers.
6. Changes to UpperBound transfer so that `LTEqLengthOf(a)` is inferred for `to - from`, given the appropriate `HasSubsequence` annotations.
7. A change to `GenericAnnotatedTypeFactory` that adds a method that provides the annotation on a Java expression String, regardless of what it is, in the factory's type system. I'm not sure why this method wasn't provided before - all that was available before was a method that took an annotation class and only would return annotations of that class.
8. Test cases for `HasSubsequence`.
9. A model of a Guava class, useful for testing.
10. Documentation for `HasSubsequence` in the manual, which appears in the UBC section.


smillst:
Building the JDK is crashing:

```
error: SourceChecker.typeProcess: unexpected Throwable (NullPointerException) while processing src/java/io/PushbackInputStream.java
  Compilation unit: src/java/io/PushbackInputStream.java
  Last visited tree at line 52 column 1:
  public
  Exception: java.lang.NullPointerException; Stack trace: org.checkerframework.javacutil.TreeUtils.typeOf(TreeUtils.java:1265)
  org.checkerframework.framework.type.GenericAnnotatedTypeFactory.getReceiverFromJavaExpressionString(GenericAnnotatedTypeFactory.java:801)
  org.checkerframework.checker.index.upperbound.UpperBoundVisitor.getReceiverFromJavaExpressionString(UpperBoundVisitor.java:294)
  org.checkerframework.checker.index.upperbound.UpperBoundTransfer.visitNumericalSubtraction(UpperBoundTransfer.java:593)
  org.checkerframework.checker.index.upperbound.UpperBoundTransfer.visitNumericalSubtraction(UpperBoundTransfer.java:103)
  org.checkerframework.dataflow.cfg.node.NumericalSubtractionNode.accept(NumericalSubtractionNode.java:23)
  org.checkerframework.dataflow.analysis.Analysis.callTransferFunction(Analysis.java:408)
  org.checkerframework.dataflow.analysis.Analysis.performAnalysisBlock(Analysis.java:210)
  org.checkerframework.dataflow.analysis.Analysis.performAnalysis(Analysis.java:187)
  org.checkerframework.framework.flow.CFAbstractAnalysis.performAnalysis(CFAbstractAnalysis.java:94)
  org.checkerframework.framework.type.GenericAnnotatedTypeFactory.analyze(GenericAnnotatedTypeFactory.java:1235)
  org.checkerframework.framework.type.GenericAnnotatedTypeFactory.analyze(GenericAnnotatedTypeFactory.java:1194)
  org.checkerframework.framework.type.GenericAnnotatedTypeFactory.performFlowAnalysis(GenericAnnotatedTypeFactory.java:1130)
  org.checkerframework.framework.type.GenericAnnotatedTypeFactory.checkAndPerformFlowAnalysis(GenericAnnotatedTypeFactory.java:1495)
  org.checkerframework.framework.type.GenericAnnotatedTypeFactory.preProcessClassTree(GenericAnnotatedTypeFactory.java:258)
```

checker/src/main/java/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -245,6 +267,53 @@ private boolean relaxedCommonAssignment(AnnotatedTypeMirror varType, ExpressionT
                 && relaxedCommonAssignmentCheck((LessThanLengthOf) qualifier, valueExp);
     }
 
+    static FlowExpressions.Receiver getReceiverFromJavaExpressionString(
```
smillst: Could you please add documentation for the new methods in this class?

checker/src/main/java/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -170,6 +138,60 @@ private void visitAccess(ExpressionTree indexTree, ExpressionTree arrTree) {
         }
     }
 
+    @Override
+    protected void commonAssignmentCheck(
+            Tree varTree, ExpressionTree valueTree, @CompilerMessageKey String errorKey) {
+
+        // check that when an assignment to a variable b declared as @HasSubsequence(a, from, to)
```
smillst: Are you checking that a, from, and to are all valid Expressions anywhere?

checker/src/main/java/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -308,6 +400,78 @@ private boolean relaxedCommonAssignmentCheck(
...
+            Subsequence subSeq = Subsequence.getSubsequenceFromReceiver(rec, atypeFactory);
+            if (subSeq != null) {
+                // viewpoint adapt strings from HasSubsequence expression
+                FlowExpressionContext context = getContextFromReceiver(rec, checker);
+                String from = standardizeAndViewpointAdapt(subSeq.from, getCurrentPath(), context);
+                String a = standardizeAndViewpointAdapt(subSeq.array, getCurrentPath(), context);
+
+                if (expQual.hasSequenceWithOffset(a, negateString(from))) {
+                    // this cast is safe because LTLs cannot contain duplicates
+                    newLHS = ((LessThanLengthOf) newLHS).removeOffset(lhsSeq, 0);
```
smillst: If varLtlQual has more than one lhsSeq with a subsequence, then is only updates the qualifier for one of them.  (Previous updates to newLHS are overwritten.) It's probably fine to assume that only one lhsSeq has a subsequence, but it should be clear.  Also, you could just return the newLHS here.

checker/tests/index/BasicSubsequence.java
```diff
@@ -0,0 +1,44 @@
+import org.checkerframework.checker.index.qual.*;
+
+class BasicSubsequence {
```
smillst: Need tests for `@HasSubsequence` with invalid flow expressions.

checker/src/main/java/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -269,6 +338,27 @@ private boolean relaxedCommonAssignmentCheck(
             return true;
         }
 
+        // Take advantage of information available on a HasSubsequence(a, from, to) annotation on either:
+        // 1. the lhs qualifier (varLtlQual): this allows us to show that iff varLtlQual includes LTL(b), b has HSS, and expQual includes LTL(a, -from), then the LTL(b) can be removed from varLtlQual
+        // 2. the rhs qualifier (expQual): this allows us to show that iff expQual includes LTL(b, from) for some b that has HSS, and varLtlQual includes LTL(a), then the LTL(a) can be removed from varLtlQual
```
smillst: This rule seems wrong:
```
@HasSub(a, form, to) String[] b = new String[10];
int from = 2;
int to = 4;
@LTL(b, form) int rhs = 6;
@LTL(a) int lhs = rhs; // this assignment shouldn't be legal, but according to this rule it is.
```

checker/src/main/java/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -269,6 +338,27 @@ private boolean relaxedCommonAssignmentCheck(
             return true;
         }
 
+        // Take advantage of information available on a HasSubsequence(a, from, to) annotation on either:
```
smillst: Can you clarify this comment? I don't understand what it is trying to say.  Also, please add some line breaks, so the line is closer to 100 characters long.  (It doesn't have to be exacted, but it's currently 200 chars.)

checker/src/main/java/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
kelloggm: It is wrong. There's supposed to be one more check: that `rhs` is `@LessThan("to")`. I'm working on a fix.

checker/src/main/java/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -308,6 +398,80 @@ private boolean relaxedCommonAssignmentCheck(
...
+                    getContextFromReceiver(rec, checker);
+
+            if (subSeq != null) {
+                // viewpoint adapt strings from HasSubsequence expression
+                String from = standardizeAndViewpointAdapt(subSeq.from, getCurrentPath(), context);
+                String a = standardizeAndViewpointAdapt(subSeq.array, getCurrentPath(), context);
+
+                if (rhsQual.hasSequenceWithOffset(rhsSeq, from)) {
+                    if (varLtlQual.hasSequenceWithOffset(a, "0")) {
+                        newLHS = ((LessThanLengthOf) newLHS).removeOffset(a, 0);
```
smillst: If varLtlQual has more than one lhsSeq with a subsequence, then is only updates the qualifier for one of them. (Previous updates to newLHS are overwritten.) It's probably fine to assume that only one lhsSeq has a subsequence, but it should be clear. Also, you could just return the newLHS here.

docs/manual/index-checker.tex
```diff
@@ -338,6 +338,26 @@ \section{Upper bounds\label{index-upperbound}}
...
+can be written on sequence declarations to indicate that a sequence has a named subsequence with the given name,
+which ranges from \code{from} to \code{to}. So, for example:
+
+\begin{Verbatim}
+  @HasSubsequence(value="this", from="this.start", to="this.end") int [] array;
+\end{Verbatim}
+
+The array declaration above indicates that the \code{this} object is a subsequence of \code{array},
+and that valid indexes to \code{this} range from \code{start} to \code{end}. The length of \code{this}
+is \code{end - start}. When an array is assigned to \code{array}, three facts are checked: that \code{start}
```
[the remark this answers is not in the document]
smillst: Mike and I discussed it.  If the name of the subsequence is `this` then a warning isn't needed, because the definition of the sequence `this` is that it is a subsequence from start to end of the array with `@HasSubsequence`.  If more than one field is annotated with `@HasSubsequence("this", ...)` then a warning should be issued.

If the name isn't `this`, then a warning should be issued at every assignment.

checker/src/main/java/org/checkerframework/checker/index/qual/HasSubsequence.java
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * A warning will still be issued at this assignment, because the Index Checker cannot prove that
+ * <code>"this"</code> is the name of the subsequence, only that such a subsequence exists. You
+ * should manually verify that the named sequence is, in fact, the subsequence described in the
+ * annotation and then suppress the warning.
+ *
+ * <p>For an example of how this annotation is used in practice, see the test GuavaPrimitives.java
+ * in /checker/tests/index/.
+ *
+ * <p>Though this annotation technically permits its user to write multiple values in each of its
```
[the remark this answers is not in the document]
kelloggm: I tried to when I was first defining the annotation, and it's forbidden by `@JavaExpression` for some reason. I don't know why.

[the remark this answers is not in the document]
kelloggm: The documentation for `@JavaExpression` is quite clear though that it can only be applied to arrays of Strings.

checker/tests/index/BasicSubsequence.java
kelloggm: This was a good thought, because what I have definitely doesn't handle invalid flow expressions correctly. I'm not sure what the right way to handle it is, though - I'd like to be able to test whether a String is a valid FlowExpression, but I don't know how to do that other than trying to parse it. Maybe that's the right answer?

checker/src/main/java/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
kelloggm: `removeOffset` returns a new qualifier without the given offset, so this doesn't actually make such an assumption. This code can probably be clearer, though.

kelloggm:
@smillst This is in much better shape than when you last reviewed. I think I've responded to all of your comments, and fixed most of them.

You're right that I'm not testing that all the expressions in HSS annotations are valid flow expressions. I don't know how to test that, though.

checker/src/main/java/org/checkerframework/checker/index/Subsequence.java
```diff
@@ -0,0 +1,182 @@
...
+        FlowExpressionParseUtil.FlowExpressionContext context = null;
+        if (rec instanceof FlowExpressions.FieldAccess) {
+            FieldAccess fa = (FlowExpressions.FieldAccess) rec;
+            context =
+                    new FlowExpressionParseUtil.FlowExpressionContext(
+                            fa.getReceiver(), null, checker);
+
+        } else if (rec instanceof FlowExpressions.LocalVariable) {
+            LocalVariable lv = (FlowExpressions.LocalVariable) rec;
+            context = new FlowExpressionParseUtil.FlowExpressionContext(lv, null, checker);
```
smillst: This isn't right for local variables.  The second parameter needs to be a list for formal parameters of the method in which the local variable is declared.  (You may just want to restrict `@HasSubsequence` to fields to make this simpler.)

checker/src/main/java/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -308,6 +410,37 @@ private boolean relaxedCommonAssignmentCheck(
...
+                        Subsequence.getSubsequenceFromReceiver(
+                                rec, atypeFactory, getCurrentPath(), context);
+
+                if (subSeq != null) {
+                    String from = subSeq.from;
+                    String a = subSeq.array;
+
+                    if (expQual.hasSequenceWithOffset(a, negateString(from, context))) {
+                        // this cast is safe because LTLs cannot contain duplicates
+                        newLHS = ((LessThanLengthOf) newLHS).removeOffset(lhsSeq, 0);
```
smillst: `newLHS` might be re-assigned twice, if `varLtlQual` has two sequences with `@HasSubseq`.  I'm not really worried about that happening, but you should note that here and possible just return newLHS here.

kelloggm: That shouldn't be a problem, since it's updating from `newLHS` each time. I'll add a note here, though.

smillst: Oh, yes, I see that now.

checker/src/main/java/org/checkerframework/checker/index/qual/HasSubsequence.java
```diff
@@ -0,0 +1,68 @@
...
+ * fields, writing more than one String to any of them is an error, and the behavior of the Index
+ * Checker is undefined.
+ *
+ * @checker_framework.manual #index-checker Index Checker
+ */
+@Target({ElementType.FIELD, ElementType.PARAMETER, ElementType.LOCAL_VARIABLE})
```
smillst: Should PARAMETER and LOCAL_VARIABLE be removed?

checker/src/main/java/org/checkerframework/checker/index/upperbound/UpperBoundVisitor.java
```diff
@@ -74,75 +91,71 @@ public Void visitAnnotation(AnnotationTree node, Void p) {
...
+            }
+            String seq = sequences.get(0);
+            String from = froms.get(0);
+            String to = tos.get(0);
+            // check that each expression is parseable in this context
+            ClassTree enclosingClass = TreeUtils.enclosingClass(getCurrentPath());
+            FlowExpressionContext context =
+                    FlowExpressionContext.buildContextForClassDeclaration(enclosingClass, checker);
+            try {
+                FlowExpressionParseUtil.parse(seq, context, getCurrentPath(), false);
```
smillst: You should check for effectively final here rather than at every assignment.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -753,40 +754,58 @@ public AnnotationMirror getAnnotationFromJavaExpressionString(
...
-        AnnotationMirror annotationMirror = null;
+    /**
+     * Returns the set of annotations on a receiver.
+     *
+     * @param receiver the receiver whose annotation is returned
+     * @param tree current tree
+     * @return a set containing the annotations on the receiver, or the empty set if none are found
+     */
+    public Set<AnnotationMirror> getAnnotationsFromReceiver(Receiver receiver, Tree tree) {
         if (CFAbstractStore.canInsertReceiver(receiver)) {
```
smillst:
```
import java.util.Arrays;
import org.checkerframework.common.value.qual.MinLen;

public class PlumeFail {
    void method() {
        @SuppressWarnings({"index", "value"})
        String @MinLen(1)[] args = getArray();
        String[] argArray = Arrays.copyOfRange(args, 1, args.length);
    }
    String[] getArray() {
        return null;
    }
}
```
The above code is failing on in plume lib.
```
PlumeFail.java:9: error: [argument.type.incompatible] incompatible types in argument.
        String[] argArray = Arrays.copyOfRange(args, 1, args.length);
                                                     ^
  required: @LTEqLengthOf("args") int
1 error
```
Before these changes, if the exact annotation requested didn't exist, then the value from the factory was used.  I tried switching this code so that the factory annotation was always used for fields and local variables.  But that caused a crash.  (I don't see a quick solution to fix the crash.)  We can work around the crash by casting:
```
        String[] args = (String @MinLen(1)[]) getArray();
```
That seems ok to me, but I don't know if this will cause a lot of new errors in the case studies. You can revert my last commit and your changes to this code.  (Though you will need to add `CFAbstractStore.canInsertReceiver(receiver)` to your original code.)

kelloggm:
The right choice is probably reverting and then adding the call as you
suggested. That seems kind of unpleasant to me, though, because it will
still leave us with two implementations of nearly the same thing.

smillst:
>The right choice is probably reverting and then adding the call as you
suggested. That seems kind of unpleasant to me, though, because it will
still leave us with two implementations of nearly the same thing.

Yes, in the long run, I'll figure out how to fix the crash.  Please add a todo comment explaining the problem.

smillst:
@kelloggm Plume-lib type checks now.  The Index Test aren't passing because they haven't been adapted since the expressions in `@HasSubsequence` must be final.

Issue #2008, opened 2018-06-02
Faulty implemetation of InternalForm annotation


gsaga:
The current implementation of signature checker will recognize `java/lang/Object[]` as the internal form of the signature of an array of `Object`s when it should be `[Ljava/lang/Object;`.

The internal form signature *is* exactly equal to the output of Class.getName with '.'s replaced with '/'s. The  JVM specs say that the Internal Form of a signature of a type is equal to the **binary name** with '.'s replaced with '/'s" (https://docs.oracle.com/javase/specs/jvms/se10/html/jvms-4.html#jvms-4.2.1). But that's only true for binary names of non array types. As Remi Forax from ASM mailing list [states](http://asm.objectweb.narkive.com/SEQPgPWW/what-s-the-precise-definition-of-an-internal-name):

    because historically, prior to 1.5 i think, you were not able to use an array as an internal name

This is backed by the following sections of jvm specs:

https://docs.oracle.com/javase/specs/jvms/se10/html/jvms-4.html#jvms-4.3.2 (near the end of section 4.3.2)

https://docs.oracle.com/javase/specs/jvms/se10/html/jvms-4.html#jvms-4.3.3 ( In the second box)

I'm working on fixing this in the following branch: https://github.com/gsaga/checker-framework/tree/fixInternalName

mernst:
Thanks for the details.  I'm asking the JVMS authors for clarification of the inconsistency between section 4.2 and section 4.3.

In the meanwhile, I will remark that Remi Forax's [message](http://narkive.com/SEQPgPWW:2.621.144) is a bit confusing.

He says
"I've always though that the section 4.1 was confusing",
but section 4.2 (not 4.1) defines internal form.

He also says,
"The section 5.1 is more precise of what is an internal name of an array https://docs.oracle.com/javase/specs/jvms/se10/html/jvms-5.html#jvms-5.1".
but neither the term "internal form" or "internal name" appears anywhere in JVMS chapter 5, much less in section 5.1 which is about the constant pool representation, not the internal form.

mernst:
Internal form does not have a representation for arrays, because it is a variant of binary names.
Binary names do not have a representation for arrays, because binary names are used for class loading, and arrays and primitives are irrelevant to class loading.  I have opened #2098 to correct this.

You mentioned references in the JVM spec:
https://docs.oracle.com/javase/specs/jvms/se10/html/jvms-4.html#jvms-4.3.2 (near the end of section 4.3.2)
[...]

PR #2009, opened 2018-06-03
Fix issue #2008

gsaga:
This pull reques fixes issue #2008.
These are the changes proposed in this pull request:

* Modified the `stringPattern`s of sub annotations of `Identifier` so that they *do not match* field descriptors of primitives. They should have an implicit type of `FieldDecriptor`.

* Modified `InternalForm` annotation and some of it's sub annotations so that they represent an actual internal form as they're used in a class file.

* Added a new annotation `FieldDescriptorForArrayInUnnamesPackage`.

Old hierarchy:     https://drive.google.com/open?id=1pv1kCsBfXsARdWelTxWkQoFLwrXVr-BA

New hierarchy:    https://drive.google.com/open?id=1ua0hKKOjcElsBGTUoEnJWMJOVRA1jjgN

* Corrected tests

* Implemented correct `string.replace` rules


docs/manual/signature-checker.tex
```diff
@@ -45,7 +45,7 @@ \section{Signature annotations\label{signature-annotations}}
   A \emph{field descriptor} (\href{https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.3.2}{JVMS \S 4.3.2}), such as
   \<Lpackage/Outer\$Inner;>, is used in a \<.class> file's constant pool,
   for example to refer to other types.  It abbreviates primitives and
-  arrays.  It uses internal form (binary names, but with \</> instead of
+        arrays.  It uses internal form (\emph{ClassGetName}s, but with \</> instead of
```
mernst: Why does this state that internal form is a variant of ClassGetName rather than a variant of binary name?
[JVMS section 4.2.1](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.2.1) defines internal form, and says it is a variant of binary names.
Is JVMS section 4.2.1 incorrect, or am I misinterpreting it?

gsaga: I asked about it on the asm mailing list and this is the reply I received:

>I've always though that the section 4.1 was confusing, i believe it's because historically, prior to 1.5 i think, you were not able to use an array as an internal name, now you can do a ldc on an array or an invokevirtual clone().
The section 5.1 is more precise of what is an internal name of an array
  https://docs.oracle.com/javase/specs/jvms/se10/html/jvms-5.html#jvms-5.1

So it looks like internal form in a variant on class get name only for non array types. Upon further inquiry I was told that the internal form of `Object` class is `java/lang/Object` and that of an array of `Object`s is `[Ljava/lang/Object;` and that the form `java/lang/Object[]` isn't used in a class file.
The links mentioned in [this](https://github.com/typetools/checker-framework/issues/2008) issue seem to verify these findings.

mernst:
Can you please pull master into your fork and resolve the conflicts?

mernst:
Issue #2008 was wrong about the definition of internal form (you were misled by Remi Forax).

The change to `@Identifier` is also wrong, because a programmer is allowed to use the identifier `B` in a program.

However, perhaps some of the changes in your branch are still worth making.

Issue #266, opened 2015-07-03
Flow refinement is ignored when initializing the store for inner classes


GoogleCodeExporter:
List of test cases:
checker/tests/nullness/FinalVar2.java
checker/tests/nullness/Issue411.java
checker/tests/nullness/Issue811java
checker/tests/nullness/Issue266.java
checker/tests/nullness/Issue266a.java

```
# What steps will reproduce the problem?

Run the nullness checker on the attached file.

# What is the expected output? What do you see instead?

I expected the nullness checker to recognize that the deference of 'tmp' in the 
inner class will always succeed. Instead, I got the error:

FinalNullable.java:16: error: dereference of possibly-null reference tmp
        return tmp.toString();
               ^

#What version of the product are you using? On what operating system?

javac 1.8.0-jsr308-1.6.7 (7102:18502dce6db5)
```

Original issue reported on code.google.com by `cus...@google.com` on 16 Oct 2013 at 4:45

Attachments:
- [FinalNullable.java](https://storage.googleapis.com/google-code-attachments/checker-framework/issue-266/comment-0/FinalNullable.java)

scottkennedy:
While the workaround seems reasonable, it would be nice to not require to change valid code to meet the requirements of a static analysis tool.

smillst:
This issue was reported by another user in #811.

smillst:
I added a test case checker/tests/nullness/Issue811java.

wmdietl:
`org.checkerframework.dataflow.cfg.CFGBuilder.CFGTranslationPhaseOne.visitNewClass(NewClassTree, Void)` contains the comment "We ignore any class body because its methods should be visited separately.".
This is likely the location where the current store should be propagated to that anonymous class.

Handling of inner/anonymous classes probably needs a solution similar to how lambda expressions are handled. Note how rewriting an anonymous class to an equivalent lambda expression gives the correct refinement behavior.
See `org.checkerframework.dataflow.cfg.CFGBuilder.declaredLambdas` and `org.checkerframework.framework.type.GenericAnnotatedTypeFactory.analyze(Queue<ClassTree>, Queue<Pair<LambdaExpressionTree, Store>>, UnderlyingAST, List<Pair<VariableElement, Value>>, ClassTree, boolean, boolean, boolean, Store)`

mihai-dumitresq:
[...]

Issue #1248, opened 2017-04-21
Refined local variables in lambda bodies


mernst:
Local variable type refinement is ignored within lambda bodies.

Below is a test case, which is checked in as `checker/tests/nullness/java8/lambda/RefinedLocalInLambda.java`.
The problem is that `printIntegersGreaterThan` fails nullness type-checking with this output:

```
% checker/bin/javac -g RefinedLocalInLambda.java -processor nullness
RefinedLocalInLambda.java:18: error: [unboxing.of.nullable] unboxing a possibly-null reference limit
        printIntegersWithPredicate(i -> i > limit);
                                            ^
1 error
```

```java
// Test case for issue #1248:
// https://github.com/typetools/checker-framework/issues/1248
// @skip-test until the issue is fixed

import java.util.function.Predicate;
import org.checkerframework.checker.nullness.qual.Nullable;

public class RefinedLocalInLambda {

    public static void main(String[] args) {
        printIntegersGreaterThan(10);
    }

    public static void printIntegersGreaterThan(@Nullable Integer limit) {
        if (limit == null) {
            return;
        }
        printIntegersWithPredicate(i -> i > limit);
    }

    public static void printIntegersWithPredicate(Predicate<Integer> tester) {
        for (int i = 0; i < 100; i++) {
            if (tester.test(i)) {
                System.out.println(i);
            }
        }
    }
}
```

wmdietl:
Related to #266 and in https://github.com/typetools/checker-framework/issues/266#issuecomment-242923644 I had observed that lambdas are handled better than inner classes. It will be good to understand why these cases don't work.

smillst:
See an additional test case reported here: https://github.com/typetools/checker-framework/issues/1305#issue-231673972

kelloggm:
Additional related test cases:

This test fails:
```java
import java.util.concurrent.*;

import org.checkerframework.checker.nullness.qual.*;

class NullnessLambdaFail {
    void test(@Nullable Object obj, ExecutorService myExecutor) {
	if (obj == null) {
	    return;
	}

	Future myFuture = myExecutor.submit(() -> do_things(obj));
    }

    void do_things(Object obj) { }
}
```
with the following error:
```java
NullnessLambda.java:11: error: [argument.type.incompatible] incompatible types in argument.
	Future myFuture = myExecutor.submit(() -> do_things(obj));
	                                                    ^
  found   : @Initialized @Nullable Object
  required: @Initialized @NonNull Object
[...]

PR #2014, opened 2018-06-06
Fix issues #266 and #1248 by keeping captured stores around

wmdietl:


dataflow/src/main/java/org/checkerframework/dataflow/cfg/node/ClassDeclarationNode.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import org.checkerframework.javacutil.TreeUtils;
+
+/**
+ * A node representing a class declaration that occurs within a method, for example, an anonymous
+ * class declaration. In contrast to a top-level class declaration, such a declaration has an
+ * initialization store that contains captured variables.
+ */
+public class ClassDeclarationNode extends Node {
```
wmdietl: @lazaroclapp @cushon Just a heads-up that we added this new CFG node, which is also now used by ObjectCreationNode. We use this to keep a Store for the location where a nested/anonymous class is declared, fixing #266 
All comments welcome.

PR #2015, opened 2018-06-06
Add missing annotation in class java.lang.reflect.Constructor for Nullness Checker

dilraj45:
Add missing annotations for method 'newInstance' in class java.lang.reflect.Constructor for Nullness Checker.

Following is an excerpt from the JavaDocs for method `newInstance` for class `Constructor`
> ```
>     * <p>If the number of formal parameters required by the underlying constructor
>     * is 0, the supplied {@code initargs} array may be of length 0 or null.
>```

It clearly states that constructor should expect `null` as an argument, in case the underlying constructor receives no formal arguments. Further, it is possible for the argument itself to be valued `null`. So newInstance method should be annotated as  `@Nullable Object @Nullable ... args` instead `Object ... args`


ashishrana160796:
Hi, 
I think similar discussion exist before. Refer #1365 and #1388.

dilraj45:
Thanks! @ashishrana160796. Closing this pull request in reference to  [#1365](https://github.com/typetools/checker-framework/issues/1365) and [#1388](https://github.com/typetools/checker-framework/issues/1365).

PR #2017, opened 2018-06-07
Fix an issue with #2003

kelloggm:
#2003 added use of the Value Checker to the Less Than Checker, so that constants are considered less than array lengths with appropriate MinLen annotations. This PR fixes an oversight in that PR, which assumed that arrays with no MinLen annotation have length `Long.MIN_VALUE`. This changes that default to `0` for array lengths, which removes some warnings in Guava. I also removed an expected warning in the test suite that should have been a suppressed warning - that warning was a false positive (it's identical to the case I'm trying to fix).

I'd normally wait for Suzanne for this, but since it's an easy fix and removes 3 false positives in Guava I figured I'd do it now. Whoever gets to it first can review it and remove the other.


mernst:
@kelloggm Tests are failing.

kelloggm:
As far as I can tell, the failure is from Maven. It looks like the
stubparser isn't able to correctly download its dependencies due to a
timeout. I saw this earlier and restarted the build, thinking it was flaky.

To test my theory, I just restarted that particular build on
typetools/master for the Checker Framework (which was passing yesterday).
If I'm correct, that will also fail.

Martin

PR #2019, opened 2018-06-10
Modified StubParser to allow the stub file to accept negative values

txiang61:


framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1418,6 +1419,11 @@ private Object getValueOfExpressionInAnnotation(
             return convert(((IntegerLiteralExpr) expr).asInt(), valueKind);
         } else if (expr instanceof LongLiteralExpr) {
             return convert(((LongLiteralExpr) expr).asLong(), valueKind);
+        } else if (expr instanceof UnaryExpr) {
+            if (((UnaryExpr) expr).getOperator() == UnaryExpr.Operator.MINUS) {
+                int value = Integer.parseInt(expr.toString());
```
mernst: What if the negated integer is a long rather than an int?

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1434,10 +1440,9 @@ private Object getValueOfExpressionInAnnotation(
         } else if (expr instanceof NullLiteralExpr) {
             stubWarn("Null found as value for %s. Null isn't allowed as an annotation value", name);
             return null;
-        } else {
-            stubWarn("Unexpected annotation expression: " + expr);
-            return null;
         }
+        stubWarn("Unexpected annotation expression: " + expr);
```
mernst: Please don't make unrelated changes such as this; keep the diffs small.
If you want to make this change, please propose it in a different pull request.

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
txiang61: Addressed

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
txiang61: It's related

mernst: Can you please explain how?  It would be helpful to also give a test case that fails without this change but succeeds with this change.

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1418,6 +1419,17 @@ private Object getValueOfExpressionInAnnotation(
             return convert(((IntegerLiteralExpr) expr).asInt(), valueKind);
         } else if (expr instanceof LongLiteralExpr) {
             return convert(((LongLiteralExpr) expr).asLong(), valueKind);
+        } else if (expr instanceof UnaryExpr) {
+            if (((UnaryExpr) expr).getOperator() == UnaryExpr.Operator.MINUS) {
+                Expression subexpr = ((UnaryExpr) expr).getExpression();
+                if (subexpr instanceof IntegerLiteralExpr) {
```
mernst: Rather than handling just literals here, how about making a recursive call?
That will handle more cases, and will also avoid duplicating code.

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
mernst: Oh, I see now that you introduced fallthroughs above.  I think it's less confusing to handle them at the point.  Though in that particular case, I think you can replace parts of the code with a recursive call.  Thanks.

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
wmdietl: This method returns an Object, so we would need a case distinction for the type of the result of the recursive call to then multiply by -1 and convert it to the correct result object.
This would give us the additional case for `expr instanceof FieldAccessExpr || expr instanceof NameExpr` case.
So the code wouldn't be much shorter, but handling the additional case would be nice.

txiang61: addressed

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
txiang61: addressed

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1504,6 +1516,36 @@ private Object convert(Number number, TypeKind expectedKind) {
         }
     }
 
+    /**
+     * Converts {@code number} to {@code expectedKind} with * -1.
+     * <p>
+     * {@code @interface Anno { long value();})
```
mernst: This comment confused me.  It doesn't start with "for example" or indicate to the reader what to examine.  Also, ithe example uses a positive number, which is confusing in a method that is about handling negative numbers.  I would drop the comment from here forward, since it doesn't seem relevant to the method.  Instead, please write `@param` and `@returns` Javadoc tags.

I was going to remark that the method name is very confusing too.
Oh, I see you just cut-and-pasted the method name and documentation from a different method.  Could you please change the name and documentation, to document the method you have written?

An alternative would be to add an optional boolean `negate` argument to `convert`.  That would make the body of `convert` slightly more complex (every expression would be multiplied by the factor), but would avoid creating an entirely new method.

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1504,6 +1516,36 @@ private Object convert(Number number, TypeKind expectedKind) {
         }
     }
 
+    /**
+     * Converts {@code number} to {@code expectedKind} with * -1.
```
mernst: I would delete the word "with", which is confusing.

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
txiang61: addressed. I added a boolean `negate` argument and overloaded the `convert` method. I think this could minimise the changes and I don't have to modify the `convert` method.

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1504,6 +1516,41 @@ private Object convert(Number number, TypeKind expectedKind) {
         }
     }
 
+    /**
+     * Converts {@code number} to {@code expectedKind} * -1 if {@code negate}. Converts {@code
+     * number} to {@code expectedKind} if not {@code negate}.
+     *
+     * @param number Number value to be converted
+     * @param expectedKind converts the Number to {byte, short, int, long, char, float, double}
+     * @param negate converted value
```
wmdietl: It seems odd to add a parameter, if the method is only ever invoked with `true`.
How about simply naming the method `convertNegated` and dropping the parameter?

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1504,6 +1516,41 @@ private Object convert(Number number, TypeKind expectedKind) {
...
+        switch (expectedKind) {
+            case BYTE:
+                return -1 * number.byteValue();
+            case SHORT:
+                return -1 * number.shortValue();
+            case INT:
+                return -1 * number.intValue();
+            case LONG:
+                return -1 * number.longValue();
+            case CHAR:
```
wmdietl: Char is unsigned, so I'm not sure this case is needed. I just tried an annotation with a `char` value and using a negation results in 'Error: incompatible types: possible lossy conversion from int to char'.

txiang61: it's possible for the value returned from the field to be negative value and we negate the negative .

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1504,6 +1516,33 @@ private Object convert(Number number, TypeKind expectedKind) {
         }
     }
 
+    /**
+     * Converts {@code number} to {@code expectedKind} * -1
+     *
+     * @param number Number value to be converted
+     * @param expectedKind converts the Number to {byte, short, int, long, float, double}
```
mernst: This isn't a comment about the formal parameter.
Maybe you are trying to say,
"the type of the result.  Must be one of {...}."

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1504,6 +1516,33 @@ private Object convert(Number number, TypeKind expectedKind) {
         }
     }
 
+    /**
+     * Converts {@code number} to {@code expectedKind} * -1
```
mernst: This multiplication suffers a type error.  A type cannot be multiplied by -1.  I would write this more clearly.

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1504,6 +1516,33 @@ private Object convert(Number number, TypeKind expectedKind) {
...
+     *
+     * @param number Number value to be converted
+     * @param expectedKind converts the Number to {byte, short, int, long, float, double}
+     * @return the converted Number Object
+     */
+    private Object convertNegate(Number number, TypeKind expectedKind) {
```
mernst: It's unfortunate to have all this code duplication.
I would have created a version with a boolean `negate`.  Then the first line could set
```
scalefactor = negate ? -1 : 1;
```
and the body `convert` could be a call to that method with the boolean false.
The extra code duplication, while unfortunate, isn't a huge deal, though.

mernst: Why is the return type `Object` instead of `Number`?

txiang61: I could definitely add another parameter to the `convert` method but I don't think Java handles default parameter values as nicely as other languages like C++. I could add another parameter and change the places where `convert` is called or I can add a vararg but that will look weird with wanting only one parameter. Which one do you prefer?

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
txiang61: addressed

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
txiang61: addressed

wmdietl:
One other comment from our meeting: `-1 * short` results in an `int`, so you will need another cast in the branches.

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1483,21 +1495,38 @@ private Object getValueOfExpressionInAnnotation(
...
     private Object convert(Number number, TypeKind expectedKind) {
+        return convert(number, expectedKind, false);
+    }
+
+    /**
+     * Converts {@code number} to {@code expectedKind}. The value converted is multiplied by -1 if
+     * {@code negate} is true
+     *
+     * @param number Number value to be converted
+     * @param expectedKind converts the Number object to one of type {byte, short, int, long, char,
```
mernst: This is not documentation of the parameter.  It doesn't say anything about `expectedKind`.
Do you want to delete the first 5 words of the parameter documentation, or do something else?

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1483,21 +1495,38 @@ private Object getValueOfExpressionInAnnotation(
...
+     *     float, double}
+     * @param negate the value of the Number Object while converting
+     * @return the converted Object
+     */
+    private Object convert(Number number, TypeKind expectedKind, boolean negate) {
+        int scalefactor = negate ? -1 : 1;
```
mernst: Could this be `byte` to avoid the need for the casts?

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1483,21 +1495,38 @@ private Object getValueOfExpressionInAnnotation(
...
+    }
+
+    /**
+     * Converts {@code number} to {@code expectedKind}. The value converted is multiplied by -1 if
+     * {@code negate} is true
+     *
+     * @param number Number value to be converted
+     * @param expectedKind converts the Number object to one of type {byte, short, int, long, char,
+     *     float, double}
+     * @param negate the value of the Number Object while converting
```
mernst: This is not documentation of `negate`.  Read the generated Javadoc to see how it will get formatted.  Maybe add two words "whether to" at the beginning of the parameter documentation.

mernst:
@txiang61 Please pull master into your branch.

Issue #2020, opened 2018-06-11
AnnotationClassLoader throws StringIndexOutOfBoundsException when dealing with Windows paths in development environment


patrickmeiring:
**Issue**
When developing a Checker in Checker Framework on Windows, running tests for that checker (e.g. in a Unit test class derived from CheckerFrameworkPerDirectoryTests) results in the error:

`1 unexpected diagnostic was found:
-1: other: error: InvocationTargetException when invoking constructor for class org.mychecker.checker.MyCheckerVisitor; Underlying cause: java.lang.StringIndexOutOfBoundsException: String index out of range: -2`

Where 'MyChecker' is the name of my checker. 

**Expected result**
No such error occurs.

**Analysis**
Debugging to identify the root cause of the problem, I find that the org.checkerframework.framework.type.AnnotationClassLoader is the root cause of the exception, when invoked from getBundledTypeQualifiersWithoutPolyAll() in the createSupportedTypeQualifiers() method of my AnnotatedTypeFactory.

The root cause appears to be that the annotation class loader does not deal well with Windows File paths. In loadBundledAnnotationClasses() of AnnotationClassLoader.java it determines if the checker is located in a JAR file or on disk. If it is on disk (which is true in my development environment), it calls into getAnnotationNamesFromDirectory(packageName, rootDirectory, subDirectory). This method then performs various path manipulation operations.

On Mac, the rootDirectory passed in is of the form "/Users/somedir". On Windows, the rootDirectory passed in is of the form "file:/C:/Users/somedir". This is then compared with a file path from file.getAbsolutePath(), which gives a path of the form "C:/Users/" on Windows and "/User/somedir" on Mac.

Because the format of the paths are different (i.e. on Windows rootDirectory contains an additional "file:/" at the start which filePath does not have), we get an error on line 632: 
`qualPackageName = filePath.substring(rootDirectory.length() + 1, filePath.length()).replace(SLASH, DOT)`

A pull request with a bug fix to this problem will be submitted shortly.

patrickmeiring:
Ignore the first commit; this contained a code styling issue. The second commit contains the change with correct code styling. This is in pull request #2021. 

My Travis CI build is failing on the "framework:downloadJtreg" step due to a connection timeout. On my local version, I have only one test failing (nullness/Issue1438.java) failing, but this is due to a timeout, and it was already failing prior to making the changes.

jyluo:
[...]

PR #2021, opened 2018-06-11
Fix issue #2020. AnnotationClassLoader handling of Windows file paths.

patrickmeiring:
Fix AnnotationClassLoader's handling of windows file paths. Normalise rootDirectory using getAbsolutePath() before comparison with filePath.


wmdietl:
Thanks!

PR #2028, opened 2018-06-15
Improve building the jdk8.jar and using the built jdk8.jar.

smillst:
1. Merge jdk8/build.gradle into checker/build.gradle.
2. Add a property, -PlocalJdk=, to use a jdk8.jar that was built locally (and building if needed) rather than downloading one.
3. Change the travis jdk.jar test to use the built jdk.

(I also updated the instructions in the annotated-libraries README: https://github.com/typetools/annotated-libraries/pull/4)


checker/build.gradle
```diff
@@ -28,11 +28,52 @@ jar {
     }
 }
 
+task downloadJdk() {
+    description 'Downloads jdk8.jar and copies it to checker/dist'
+    // See instructions for updating the jdk.jar at https://github.com/typetools/annotated-libraries/blob/master/README.md
+    String shaHash = '3e57e020050ee8b679622e7f864e7dff17aa8acc'
+    File jdk = new File(buildDir, "jdk8-${shaHash}.jar");
+    onlyIf {
+        !rootProject.hasProperty("builtJdk") && (!jdk.exists() || !(new File('"dist/jdk8.jar').exists()))
```
mernst: Somewhere, please document the meaning of `builtJdk`:
If set (regardless of what value it is set to), then use `dist/jdk8.jar` instead of downloading.

checker/build.gradle
```diff
@@ -28,11 +28,52 @@ jar {
     }
 }
 
+task downloadJdk() {
+    description 'Downloads jdk8.jar and copies it to checker/dist'
+    // See instructions for updating the jdk.jar at https://github.com/typetools/annotated-libraries/blob/master/README.md
+    String shaHash = '3e57e020050ee8b679622e7f864e7dff17aa8acc'
+    File jdk = new File(buildDir, "jdk8-${shaHash}.jar");
```
mernst: I would name this variable "downloadedJdk" or the like.

smillst:
> Try to download a remote jdk8, preserving timestamps.
> Use whichever is newer of the locally-built and the downloaded JDK. (Either one might not exist.)
What do you think of this

This won't work if you build the jdk with changes on one branch, then switch to a second branch and run tests on the second branch.  This also won't work someone else uploads a new jdk to master and then you merge master into your branch where you want to use a local version of the jdk.

Since `builtJdk` is a Gradle property, you can add it to a `gradle.properties` file `checker-framework` or you can modify `checker-framework/build.gradle` to contain the property.

mernst:
> This won't work if you build the jdk with changes on one branch, then switch to a second branch and run tests on the second branch.

When switching branches, any built binary might become out of date.  The buildfile can handle this by rebuilding the out-of-date artifact.  This is why git gives every changed file a new timestamp rather than keeping the old timestamps, when switching branches.

> This also won't work someone else uploads a new jdk to master and then you merge master into your branch where you want to use a local version of the jdk.

If someone else changes the JDK, then after merging you need to rebuild your JDK anyway.


Maybe rather than "builtJdk", there should be a "localJdk" property that says to use a locally-built version of the Jdk, and builds it if necessary (if the .jar file is older than some source file that contributes to it).  This requires getting the dependences right in the buildfiles, but that seems like something we would want anyway, and it seems to handle both of the above situations.

I think these changes will remain relevant even when the JDK is in its own repository, but maybe I am missing something there.

.travis-build.sh
```diff
@@ -122,10 +122,10 @@ if [[ "${GROUP}" == "downstream" || "${GROUP}" == "all" ]]; then
 fi
 
 if [[ "${GROUP}" == "jdk.jar" || "${GROUP}" == "all" ]]; then
-  ./gradlew :jdk8:buildJdk
+  ./gradlew buildJdk
 
   ## Run the tests for the type systems that use the annotated JDK
-  ./gradlew IndexTest LockTest NullnessFbcTest OptionalTest
+  ./gradlew IndexTest LockTest NullnessFbcTest OptionalTest -PlocakJdk=
```
mernst: change `locak` to `local`

checker/build.gradle
```diff
@@ -28,11 +28,10 @@ jar {
     }
 }
 
-task copyJarsToDist(dependsOn: shadowJar, group: 'Build') {
-    description 'Builds or downloads jars required by CheckerMain and puts them in checker/dist.'
+task copyJarsOtherThanJdk8ToDist(dependsOn: shadowJar, group: 'Build') {
+    description 'Builds jars required by CheckerMain, except for jdk8.jar and puts them in checker/dist.'
```
mernst: minor: add comma after "jdk8.jar"

smillst:
TODO: Add task to re-build jdk8.jar even if the files haven't changed.  Should be used if defaulting code has changed in your checker.

PR #2033, opened 2018-06-16
 Wrap long lines, as per google-java-format

mernst:


wmdietl:
@mernst Question related to this PR: will comments now be automatically reformatted? Or do we manually need to make sure we follow these line breaks?

mernst:
@wmdietl Brief answer:  No, and no.

This pull request is from an experiment I did with a newer version of google-java-format.

Long `//` comments are automatically broken across lines by google-java-format, as of version 1.4 (released in April 2017).  We are still using google-java-format 1.3.  I believe (but am not certain) that is because we are waiting for opprop/checker-framework-inference to be merged into typetools/cecker-framework-inference -- we don't want to reformat while big merges are outstanding.  Do you know when that merge might occur?  (I could upgrade the gjf version for some projects but not CFI, but I think it's less confusing to be consistent across projects.)

mernst:
Thanks for the clarification.  My notes must have been wrong; sorry about that.  That's a reason not to start using gjf on CFI, but not a reason not to upgrade from gjf 1.3.  There is another issue with upgrading gjf, which is dealing with long `// ::` lines that get wrapped, but that is orthogonal to CFI.  (It will be good to have CFI upgraded regardless.)

PR #2034, opened 2018-06-18
Remove extraneous space from output

mernst:


mernst:
@wmdietl Do you want to review this kind of change?

framework/src/main/java/org/checkerframework/framework/util/DefaultAnnotationFormatter.java
```diff
@@ -28,8 +28,8 @@ public static boolean isInvisibleQualified(AnnotationMirror anno) {
     }
 
     /**
-     * Creates a space String of each annotation in annos separated by a single space character,
-     * obeying the printInvisible parameter.
+     * Creates a space String of each annotation in annos separated by a single space character and
```
wmdietl: Can the first `space` be removed? I'm not sure what a `space String` would be.

mernst: Yes, that's a pre-existing typo.  I'll fix it.

PR #2036, opened 2018-06-18
Add EnumMap annotated for Nullness Checker

dilraj45:
Annotate java.util.EnumMap for Nullness Checker

Source: [Open-JDK-8](http://hg.openjdk.java.net/jdk8/) repository


liach:
the annotation on unmasknull is wrong

dilraj45:
I have marked parameter for `unmaskNull` as `@Nullable` for there are instances in `EnumMap.java` itself when a null value can be passed to this method. Plese consider the following code snippet
>```
>public V put(K key, V value) {
>    typeCheck(key);
>
>    int index = key.ordinal();
>    Object oldValue = vals[index];
>    vals[index] = maskNull(value);
>    if (oldValue == null)
>      size++;
>    return unmaskNull(oldValue);
>  }
>```
Further, unmaskNull does not throw any `NullPointerException` when a null is passed.

The return type is marked is`V` for the type of value returned will depend upon the type of `V` specified while instantiating `EnumMap`. Please let me know If I'm missing something.

mernst:
Thanks for the pull request.
You changed the indentation on every line of the file.
This makes it very difficult to diff it against upstream, or to incorporate upstream changes.
Could you please restore the original indentation?
Thanks!

checker/jdk/nullness/src/java/util/EnumMap.java
```diff
@@ -0,0 +1,839 @@
...
+     * identically to {@link #EnumMap(EnumMap)}.  Otherwise, the specified map
+     * must contain at least one mapping (in order to determine the new
+     * enum map's key type).
+     *
+     * @param m the map from which to initialize this enum map
+     * @throws IllegalArgumentException if <tt>m</tt> is not an
+     *     <tt>EnumMap</tt> instance and contains no mappings
+     * @throws NullPointerException if <tt>m</tt> is null
+     */
+    @SuppressWarnings("nullness:method.invocation.invalid") // All the fields required by putAll method
```
mernst: In this case, I suggest that you annotate `putAll` to express its contract:
 * the receiver is `@UnknownInitialization`
 * `@RequiresNonNull` for some expressions

It's better to express (and check!) the contract than to write `@SuppressWarnings`.  In this case, I don't think that you need `@SuppressWarnings`.

checker/jdk/nullness/src/java/util/EnumMap.java
```diff
@@ -0,0 +1,839 @@
...
+        }
+        return false;
+    }
+
+    /**
+     * Returns true if key is of the proper type to be a key in this
+     * enum map.
+     */
+    @EnsuresNonNullIf(expression="#1", result=true)
+    private boolean isValidKey(Object key) {
```
mernst: Should `key` also be annotated as `@Nullable`?  Otherwise a call with a possibly-null value is not permitted, and the `@EnsuresNonNullIf` annotation is moot.

checker/jdk/nullness/src/java/util/EnumMap.java
```diff
@@ -0,0 +1,839 @@
...
+     * The returned set obeys the general contract outlined in
+     * {@link Map#keySet()}.  The set's iterator will return the keys
+     * in their natural order (the order in which the enum constants
+     * are declared).
+     *
+     * @return a set view of the keys contained in this enum map
+     */
+    public Set<K> keySet() {
+        throw new RuntimeException("skeleton method");
+        // Does not compile with actual version, java.util.AbstractMap is reduced to skeleton methods
```
mernst: Could you please expand `AbstractMap.java` to full source code, at the same time as doing this?  Thanks.

checker/jdk/nullness/src/java/util/EnumMap.java
```diff
@@ -0,0 +1,839 @@
...
+        //     return ks;
+        // else
+        //     return keySet = new KeySet();
+    }
+
+    private class KeySet extends AbstractSet<K> {
+        public Iterator<K> iterator() {
+            return new KeyIterator();
+        }
+        @Pure public int size() {
```
mernst: `@Pure` is a method annotation, so it belongs on its own line.
The same comment applies elsewhere in this file.

checker/jdk/nullness/src/java/util/EnumMap.java
```diff
@@ -0,0 +1,839 @@
...
+        public void clear() {
+            EnumMap.this.clear();
+        }
+        public Object[] toArray() {
+            return fillEntryArray(new Object[size]);
+        }
+        @SuppressWarnings({
+            "unchecked",
+            "nullness:argument.type.incompatible", // 'a' is known to be of array class type
+            "nullness:override.param.invalid" // Annotation for to Array are technically incorrect. Refer
```
mernst: Does the checker use these incorrect annotations, or a more precise version as mentioned in the note in Collection.java?

Minor: change "to Array" to "toArray", for clarity.

checker/jdk/nullness/src/java/util/EnumMap.java
```diff
@@ -0,0 +1,839 @@
...
+        private @Nullable Entry lastReturnedEntry = null;
+
+        public Map.Entry<K,V> next() {
+            if (!hasNext())
+                throw new NoSuchElementException();
+            lastReturnedEntry = new Entry(index++);
+            return lastReturnedEntry;
+        }
+
+        @SuppressWarnings("nullness:dereference.of.nullable") // super.remove throws a NullPointerException
```
mernst: I would use `@AssumeAssertion` to put this comment on the line after the `super.remove` call.  I think it will be easier to read and understand there.

checker/jdk/nullness/src/java/util/EnumMap.java
dilraj45: As per the note for the toArray method in Collection.java, the checker uses some hard-coded rules for defining its behaviour. I'm assuming that Checker will use the same hard-coded rules for this method as well as. (Although I have made an effort to express as much as of its behaviour as possible using annotations). Please correct me if I'm wrong.

Thanks!!

mernst: 1. Please check rather than assuming.
2. Please explicitly say that the checker uses hard-coded rules. That will be clearer than just calling them "technically incorrect", which confused me.  My request is really for clearer documentation.
Thanks!

dilraj45:
@mernst I have addressed your comments. Please take another look. Thanks!!

PR #2037, opened 2018-06-19
Don't include Emacs autosave files (`.#Filename.java`) in SOURCES

mernst:


smillst:
@mernst I can't delete this branch.

PR #2042, opened 2018-06-19
 Handle wrapped diagnostic lines

mernst:


util/src/main/java/org/checkerframework/framework/test/diagnostics/TestDiagnosticUtils.java
```diff
@@ -214,6 +214,29 @@ protected static TestDiagnostic fromPatternMatching(
...
+        return trimmedLine.startsWith("// ::") || trimmedLine.startsWith("// warning:");
+    }
+
+    /** Return true if this line in a Java file continues an expected diagnostic. */
+    public static boolean isJavaDiagnosticLineContinuation(String originalLine) {
+        if (originalLine == null) {
+            return false;
+        }
+        final String trimmedLine = originalLine.trim();
+        return trimmedLine.startsWith("// ") && !trimmedLine.startsWith("// warning:");
```
wmdietl: I would find a comment helpful to understand why `"// warning:"` must not be the beginning of a continuation line.
If two warnings are supposed to be issued on one line, why can't the two lines start with `"// warning:"`?

PR #2044, opened 2018-06-20
Correct build jdk.jar and using it in Gradle tasks.

smillst:
Also, updated the README in https://github.com/typetools/annotated-libraries/pull/5/files.


checker/build.gradle
```diff
@@ -51,22 +60,38 @@ task copyJarsOtherThanJdk8ToDist(dependsOn: shadowJar, group: 'Build') {
         }
     }
 }
+boolean shouldForceBuildJdk = false
```
mernst: Who would use this target?  (I'm wondering whether it is necessary.)
Also, a cleaner way to achieve the goal is to just delete jdk.jar and then run buildJdk.  That doesn't require any `shouldForceBuildJdk` or any change to the `buildJdk` task.

PR #2050, opened 2018-06-22
Make cross-module Javadoc references work

mernst:
A comment in `build.gradle` says:

> Add the fat checker.jar to the classpath of every Javadoc task. This allows Javadoc in any module to reference classes in any other module.

Nonetheless, cross-module Javadoc references cause warnings when I run `./gradlew` Javadoc, and many Javadoc cross-references are missing.

This pull request resolves these issues, and seems to work.

I'm not sure whether this is the right way to do it, though.
 * One issue is that the URLs are external, absolute URLs rather than relative ones.
    * It does not refer to the current Javadoc, but that of the last release.
    * It is not usable offline.
   I don't see how to resolve this without postprocessing, given that the Javadoc is generated separately and only later copied together by the `allJavadoc` task.
 * I wonder whether there will be problems when the API changes.  If a pull request changes the API, will Javadoc fail because it is getting Javadoc links from https://checkerframework.org/api/ rather than from a local source?  Or does the use of the current fat jar handle this case, and the only problem will be symbols in the Javadoc that are not linked to anywhere (as is already the case, and as this pull request fixes)?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -89,11 +89,13 @@ allprojects {
                 classpath += files(project(':checker').tasks.getByName('shadowJar').archivePath)
                 options.taglets 'org.checkerframework.taglet.ManualTaglet'
                 options.tagletPath(project(':util').sourceSets.taglet.output as File[])
-                options.links = ['https://docs.oracle.com/javase/8/docs/api/', 'https://docs.oracle.com/javase/8/docs/jdk/api/javac/tree/']
+                options.links = ['https://docs.oracle.com/javase/8/docs/api/', 'https://docs.oracle.com/javase/8/docs/jdk/api/javac/tree/', 'https://checkerframework.org/api/']
                 // This file is looked for by Javadoc.
                 file("${destinationDir}/resources/fonts/").mkdirs()
                 ant.touch(file: "${destinationDir}/resources/fonts/dejavu.css")
             }
+            // Turn javadoc warnings into errors.
+            options.addStringOption('Xwerror', '-quiet')
         }
 
         // Add standard javac options


build.gradle
```diff
@@ -89,11 +89,13 @@ allprojects {
...
                 options.taglets 'org.checkerframework.taglet.ManualTaglet'
                 options.tagletPath(project(':util').sourceSets.taglet.output as File[])
-                options.links = ['https://docs.oracle.com/javase/8/docs/api/', 'https://docs.oracle.com/javase/8/docs/jdk/api/javac/tree/']
+                options.links = ['https://docs.oracle.com/javase/8/docs/api/', 'https://docs.oracle.com/javase/8/docs/jdk/api/javac/tree/', 'https://checkerframework.org/api/']
                 // This file is looked for by Javadoc.
                 file("${destinationDir}/resources/fonts/").mkdirs()
                 ant.touch(file: "${destinationDir}/resources/fonts/dejavu.css")
             }
+            // Turn javadoc warnings into errors.
+            options.addStringOption('Xwerror', '-quiet')
```
smillst: This causes the `javadoc` tasks to fails.  (It does not cause `allJavadoc` to fail, which Travis uses, because that task does not use the `javadoc` tasks.)

mernst:
[Here is the output](https://github.com/typetools/checker-framework/files/2129628/gradle-clean-javadoc-output.txt) I get from `./gradlew clean javadoc` on on master, running under Ubuntu 18.04.

On the other hand, I do not get these errors on CentOS.  Do you have any ideas about how I can investigate what is different about my environment?

wmdietl:
I did `./gradlew clean javadoc` in a clean master and get only warnings about param and return documentation. No warnings about links.
Ubuntu 18.04.

mernst:
I was just able to reproduce the output above with a clean clone on Centos, then `./gradlew clean javadoc` as the first command in that clone.

And, here is a Travis job that shows the Javadoc warnings:
 https://travis-ci.org/mernst/checker-framework/jobs/395857341#L1405
Maybe there is a dependency problem in the Gradle scripts.

mernst:
@wmdietl That is not the behavior I am seeing on Ubuntu 18.04.
Could you paste the output of
`git clone --depth=1 git@github.com:typetools/checker-framework.git checker-framework && cd checker-framework && ./gradlew clean javadoc`
?

wmdietl:
I first did `(git clone --depth=1 git@github.com:typetools/checker-framework.git checker-framework && cd checker-framework && ./gradlew clean javadoc) > log-emptydir.txt 2>&1` in an empty directory.
[log-emptydir.txt](https://github.com/typetools/checker-framework/files/2263831/log-emptydir.txt)
However, those errors are caused by the missing related projects.

So in a directory that contains the built related projects, I did `(git clone --depth=1 git@github.com:typetools/checker-framework.git checker-framework-for-javadoc && cd checker-framework-for-javadoc && ./gradlew clean javadoc) > log.txt 2>&1` and get a successful build:
[log.txt](https://github.com/typetools/checker-framework/files/2263851/log.txt)

Please let me know if I should change something else about the setup.

PR #2057, opened 2018-06-28
Add unit test for added IntRange annotation

NhatDinh:
+ Move LongAndIntegerBitsMethods.java to checker/tests/index


NhatDinh:
@panacekcz @kelloggm Please review and let me know if any changes are needed.

framework/tests/value/BitsMethodsIntRange.java
```diff
@@ -0,0 +1,15 @@
+public class BitsMethodsIntRange {
+    void caseInteger(int index, char val1, char val2) {
+        char[] arr1 = new char[33];
+        char[] arr2 = new char[33];
+        arr1[Integer.numberOfLeadingZeros(index)] = val1;
+        arr2[Integer.numberOfTrailingZeros(index)] = val2;
+    }
+
+    void caseLong(int index, char val1, char val2) {
```
panacekcz: I'd prefer `long index` for completeness.

framework/tests/value/BitsMethodsIntRange.java
```diff
@@ -0,0 +1,15 @@
+public class BitsMethodsIntRange {
+    void caseInteger(int index, char val1, char val2) {
+        char[] arr1 = new char[33];
+        char[] arr2 = new char[33];
+        arr1[Integer.numberOfLeadingZeros(index)] = val1;
```
panacekcz: I thought that the point of splitting this test was so that the value checker test does not use array indexing.
You can test that an expression is compatible with a qualified type by assigning to an annotate variable like here for example: https://github.com/typetools/checker-framework/blob/master/framework/tests/value/BitwiseAnd.java#L14

NhatDinh: @panacekcz I see now, the direct approach is to use `@IntRange()` directly, not using another part of the checker or array indexing, I was too hung up on the definition of `unit test` and `integration test`.

NhatDinh:
@panacekcz @kelloggm  Please give me another round of review on this. Thank you.

Issue #1614, opened 2017-10-24
Replace the flow expression parser with a real parser


smillst:
The current implementation attempts to parse Java expressions by applying a set of regular expressions.  The regex approach could perhaps be made to work on our test cases, but cannot be made to work in general.  Use of regexes was perhaps defensible when flow expressions were only allowed to be field accesses, but not beyond that.

Furthermore, those regular expressions are buggy (every time I fix one problem, I find another), such an approach is inefficient, and the implementation is ill-documented.

The current flow expression parser needs to be replaced by a real parser, such as [JavaParser](https://javaparser.org/), which is already used by the Checker Framework.  To parse an expression, wrap it in a dummy class and extract it.  (There is code to do that somewhere; I'm not sure if it is in the Checker Framework.)

A complication is that [the flow expression syntax](https://checkerframework.org/manual/#java-expressions-as-arguments) permits non-standard Java expressions such as `"#1"`.  (`"<self>"` is pre-processed and wouldn't need to be handled.)  Maybe pre-process to replace occurrences of `#n` by the correct identifier (which is already known:  it is passed to the flow expression parser already, maybe in the context).

(The above was written by @mernst in #1609, but was moved to here as it is a separate issue.)

KankshaZ:
Hi, I have attempted to tackle the above problem of replacing the Flow Expression Parser with calls to the JavaParser (it is a very crude and limited implementation - just wanted to make sure that my approach is correct). Please follow this link: [https://github.com/KankshaZ/ReplaceFlowExpressionParser]( https://github.com/KankshaZ/ReplaceFlowExpressionParser)

PR #2062, opened 2018-07-03
Replacement of the flow expression parser with calls to Java Parser

KankshaZ:
# Replacement of the flow expression parser with calls to Java Parser [#1614](https://github.com/typetools/checker-framework/issues/1614)
The CheckerFramework uses Regex to parse expressions/arguments given to Annotations. To make it more generalizable and robust, the current flow expression parser is replaced by calls to the JavaParser. The JavaParser's quick API is used where static methods parse fragments of source code which are Strings. 

### Changes
In the current [FlowExpressionParseUtil.java](https://github.com/KankshaZ/checker-framework/blob/master/framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java#L483) file of the Checker Framework, the following patterns have been created from regular expressions
- unanchoredParameterPattern
- parameterPattern
- identifierPattern
- intPattern
- longPattern
- stringPattern
- parenthesisPattern
- memberSelectOfStringPattern
- pattern for indentParser

These patterns are used to identify the type of Java Expression. Patterns also help to group to extract other information like receivers or names.

In the modified [FlowExpressionParseUtil.java](https://github.com/KankshaZ/checker-framework/blob/ReplaceFlowExpressionParser/framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java) file, indentification of the type of expression is done by extracting the class obtained after parsing expressions using the JavaParser. Information like name, arguments, and scope, is now extracted from the expression parsed by the JavaParser as well. This is how regular expressions used for matching have been replaced with calls to the JavaParser. 

Replacement of unanchoredParameterPattern and parameterPattern is not done since they are non-standard Java Expressions. Occurences of parameters are replaced by arguments, and then replaced back before the expression is finally returned.


mernst:
Please pull master into your branch and resolve the conflicts.  Thanks!

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -1,5 +1,9 @@
 package org.checkerframework.framework.util;
 
+import static com.github.javaparser.JavaParser.*;
+
+import com.github.javaparser.ParseProblemException;
+import com.github.javaparser.ast.expr.*;
```
mernst: Our coding style is not to use wildcard imports; so please expand this.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -158,7 +145,73 @@ private static Pattern anchored(String regex) {
     }
 
     private static boolean isMemberSelect(String s, FlowExpressionContext context) {
-        return parseMemberSelect(s) != null;
+        return parseMemberSelect(s, context) != null;
+    }
+
+    /**
+     * Matches a field access which contains a method call either in receiver or in remaining. First
```
mernst: The method is named `parseMethodCall`, but it's documented to match a field access.
Furthermore, its return type is similar to `parseMemberSelect` and dissimilar to other occurrences of `parseMethodCall`.  Please rename this to indicate what does, which is to parse a field access (a MemberSelect).

mernst: What is "remaining"?  How is it related to "field" in the line below?

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -158,7 +145,73 @@ private static Pattern anchored(String regex) {
...
+     * of returned pair is object and second is field. The object is not a field access expression
```
mernst: What does "object" refer to here?  If it's an expression, don't call it an object.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
mernst: Can you give an example of what the expression is?  I don't understand the method's specification.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -158,7 +145,73 @@ private static Pattern anchored(String regex) {
...
+     * itself.
+     *
+     * @param s expression string with replaced formal parameters
+     * @param origString expression string
```
mernst: Variables `s` and `origString` are related.  They should have related names too.  The current names don't suggest any relationship between the variables.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -158,7 +145,73 @@ private static Pattern anchored(String regex) {
...
-        return parseMemberSelect(s) != null;
+        return parseMemberSelect(s, context) != null;
+    }
+
+    /**
+     * Matches a field access which contains a method call either in receiver or in remaining. First
+     * of returned pair is object and second is field. The object is not a field access expression
+     * itself.
+     *
+     * @param s expression string with replaced formal parameters
```
mernst: What is its format?  Are arguments replaced, or are formal parameters replaced?

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -158,7 +145,73 @@ private static Pattern anchored(String regex) {
...
+     * @param origString expression string
+     * @param argumentList list of arguments corresponding to formal parameters
```
mernst: Formal parameters of what?  The method call that is in s?  The method call that is annotated by an annotation that contains s?  Something else?

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -167,12 +220,26 @@ private static boolean isMemberSelect(String s, FlowExpressionContext context) {
      * @param s expression string
      * @return pair of object and field
      */
-    private static Pair<String, String> parseMemberSelect(String s) {
-        Pair<Pair<String, String>, String> method = parseMethodCall(s);
-        if (method != null && method.second.startsWith(".")) {
-            return Pair.of(
-                    method.first.first + "(" + method.first.second + ")",
-                    method.second.substring(1));
+    private static Pair<String, String> parseMemberSelect(String s, FlowExpressionContext context) {
```
mernst: Please add an `@param` for each new formal parameter (in this case, `context`).

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -167,12 +220,26 @@ private static boolean isMemberSelect(String s, FlowExpressionContext context) {
...
+
+        // replace occurence of formal parameters, such as "#2", with corresponding arguments
+        String sCopy = s;
```
mernst: `sCopy` is a poor variable name.  It does not indicate what the variable represents or how it is used or how it differs from s.  Even a name containing `replaced` would be better, or one containing `withArgs`.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -167,12 +220,26 @@ private static boolean isMemberSelect(String s, FlowExpressionContext context) {
...
+        if (context.arguments != null) {
+            for (Receiver argument : context.arguments) {
+                String replaceParam = "#" + Integer.toString(k + 1);
+                sCopy = sCopy.replace(replaceParam, argument.toString());
+                k++;
+                argumentList.add(argument.toString());
+            }
+        }
+
+        // parses field accesses and method calls which contain a method call
```
mernst: I don't understand this comment, but that's probably because I don't understand the new `parseMethodCall` routine.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -181,12 +248,23 @@ private static boolean isMemberSelect(String s, FlowExpressionContext context) {
                     array.first.first + "[" + array.first.second + "]", array.second.substring(1));
         }
 
-        Pattern memberSelectOfStringPattern = anchored("(" + STRING_REGEX + ")" + "\\.(.*)");
-        Matcher m = memberSelectOfStringPattern.matcher(s);
-        if (m.matches()) {
-            return Pair.of(m.group(1), m.group(2));
+        // parses member selects which are strings
```
mernst: I think you mean that the receiver is a string literal.  If so, say that.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -181,12 +248,23 @@ private static boolean isMemberSelect(String s, FlowExpressionContext context) {
...
+                MethodCallExpr methodCallExpr = (MethodCallExpr) e;
+                String mceString = methodCallExpr.toString();
+                String receiver = methodCallExpr.getScope().get().toString();
+                String remaining = mceString.substring(receiver.length(), mceString.length());
+                if (receiver.charAt(0) == '\"') {
+                    return Pair.of(receiver, remaining.substring(1));
+                }
+            }
+        } catch (ParseProblemException e) {
+
```
mernst: Never catch an exception without doing anything.  At the very least, add a comment saying why this is acceptable.
Keep the scope of the `try` statement as small as possible.
If the only statement that can throw `ParseProblemException` is `parseExpression`, then make the `try` body only that statement.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -266,8 +344,16 @@ private static boolean isIntLiteral(String s, FlowExpressionContext context) {
         if (context.parsingMember) {
             return false;
         }
-        Matcher intMatcher = INT_PATTERN.matcher(s);
-        return intMatcher.matches();
+        try {
+            Expression e = parseExpression(s);
```
mernst: The next 6 lines can be replaced by
`return parseExpression(s).getClass().equals(IntegerLiteralExpr.class)`

In general, never do
```
if (p) {
  return true;
} else {
  return false;
}
```
Instead, just do `return p;`

The same comment applies to several other blocks of code.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -471,42 +577,113 @@ private static Receiver parseParameter(String s, FlowExpressionContext context)
...
+        } else {
+            return s;
+        }
+    }
+
     /**
      * Parse a method call. First of returned pair is a pair of method name and arguments. Second of
      * returned pair is a remaining string.
      *
      * @param s expression string
```
mernst: Is `s` required to be a method call?

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -471,42 +577,113 @@ private static Receiver parseParameter(String s, FlowExpressionContext context)
...
+     * @param argumentList list of arguments corresponding to formal parameters
```
mernst: Of what method?
Also, arguments always correspond to formal parameters, so you can cut "corresponding to formal parameters".

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -648,31 +824,31 @@ private static int matchingCloseParen(String s, int openPos, char open, char clo
...
-        while (i < s.length()) {
-            char ch = s.charAt(i++);
-            if (ch == '"') {
-                i--;
-                // TODO: inefficient to re-compute this on every iteration, should be static
-                Pattern stringPattern = anchored("(" + STRING_REGEX + ").*");
-                Matcher m = stringPattern.matcher(s.substring(i));
-                if (!m.matches()) {
-                    break;
+        int count = -1;
```
mernst: Count is not a descriptive name.  Maybe `bracketDepth` would be better.
However, I wonder if this is necessary.  It seems that `openBracket` is a stack.  So why not use `Stack` and `push` and `pop` and `isEmpty` rather than introducing a new integer variable?

mernst:
You committed an empty file named "-".  Please remove it from your pull request.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -158,21 +153,102 @@ private static Pattern anchored(String regex) {
...
     private static boolean isMemberSelect(String s, FlowExpressionContext context) {
-        return parseMemberSelect(s) != null;
+        return parseMemberSelect(s, context) != null;
+    }
+
+    /**
+     * Matches field accesses whose receiver is a method call expression. First of returned pair is
+     * a method call expression and the second is a field or method call expression.
+     *
+     * @param sWithFormalParamNames expression string with replaced occurences of one-based
```
mernst: Choose argument names based on semantics, not on types.
When an argument is a string, `expr` is a better variable name than `s` (for "string").

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -158,21 +153,102 @@ private static Pattern anchored(String regex) {
...
+     *     parameter index of with formal parameter names
```
mernst: I don't understand this comment.  That may be because there is no information in the Javadoc for `s` about what it is.
Are there any constrains on `s`?  I think it may have numbered formal parameters, such as `#1.foo()` or `foo(#2)`.  The documentation needs to be specific about the format of strings.

Maybe the two formal parameters should be named `exprWithNamedParameters` and `exprWithNumberedParameters`.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -158,21 +153,102 @@ private static Pattern anchored(String regex) {
...
+     * @param s expression string
+     * @param argumentList list of formal parameter names corresponding to index of formal
+     *     parameters in the method call that is in s
```
mernst: This implies that `s` is, or contains, a method call.  That needs to be indicated in the documentation for `s`, not obliquely referenced here.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -158,21 +153,102 @@ private static Pattern anchored(String regex) {
...
+        return parseMemberSelect(s, context) != null;
+    }
+
+    /**
+     * Matches field accesses whose receiver is a method call expression. First of returned pair is
+     * a method call expression and the second is a field or method call expression.
+     *
+     * @param sWithFormalParamNames expression string with replaced occurences of one-based
+     *     parameter index of with formal parameter names
+     * @param s expression string
```
mernst: Does `s` have to be an expression containing a dot (`.`) at the top level?

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -158,21 +153,102 @@ private static Pattern anchored(String regex) {
...
+     * @param argumentList list of formal parameter names corresponding to index of formal
+     *     parameters in the method call that is in s
+     * @return pair of method call expression and field, or two method call expressions
```
mernst: Giving an example of input and output would be helpful.

mernst: The method body returns `null` sometimes, but the documentation does not indicate that is possible nor state when it occurs.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -158,21 +153,102 @@ private static Pattern anchored(String regex) {
...
+     */
+    private static Pair<String, String> parseMethodCall(
```
mernst: I still have the concern, which I mentioned in my previous review, that the signature of this method is different than that of other `parseMethodCall` methods,  This seems to be parsing not a method call, but an expression whose receiver is a method call.  Please clarify the documentation.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -158,21 +153,102 @@ private static Pattern anchored(String regex) {
...
+            }
+        }
+        return null;
     }
 
     /**
      * Matches a field access. First of returned pair is object and second is field.
      *
      * @param s expression string
+     * @param context information about any receiver and arguments
```
mernst: I don't understand this.  The method is parsing a field access (a "member select").  What are the arguments?  Please make the documentation descriptive.  Vague text like "any receiver and arguments" doesn't provide any information beyond what readers can determine from the type `FlowExpressionContext`.
Make the same fix elsewhere.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -181,12 +257,23 @@ private static boolean isMemberSelect(String s, FlowExpressionContext context) {
...
+                MethodCallExpr methodCallExpr = (MethodCallExpr) e;
+                String mceString = methodCallExpr.toString();
+                String receiver = methodCallExpr.getScope().get().toString();
+                String remaining = mceString.substring(receiver.length(), mceString.length());
+                if (receiver.charAt(0) == '\"') {
+                    return Pair.of(receiver, remaining.substring(1));
+                }
+            }
+        } catch (ParseProblemException pe) {
+            // if format of the expression is such that it cannot be parsed by JavaParser
```
mernst: This is not grammatical (it starts with "if" but doesn't have a consequent).  It also doesn't explain why the exception is caught but ignored.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -181,12 +257,23 @@ private static boolean isMemberSelect(String s, FlowExpressionContext context) {
...
         }
 
+        // parses member selects within brackets
```
mernst: By "brackets" do you mean parentheses?  This comment is inconsistent with the next line.
It would be helpful to give an example of the form of the expression (especially if you are having trouble describing it in words).

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -266,8 +353,12 @@ private static boolean isIntLiteral(String s, FlowExpressionContext context) {
         if (context.parsingMember) {
             return false;
         }
-        Matcher intMatcher = INT_PATTERN.matcher(s);
-        return intMatcher.matches();
+        try {
+            Expression e = parseExpression(s);
```
mernst: Don't evaluate `parseExpression` twice.  It is inefficient and it will confuse readers as to why you need to do it twice.

You can write:

```
Expression e;
try {
  e = parseExpression(s);
} catch (ParseProblemException pe) {
  return false;
}
return e.getClass().equals(IntegerLiteralExpr.class);
```

but in this case I think it's even better to write 

```
try {
  return parseExpression(s).getClass().equals(IntegerLiteralExpr.class);
} catch (ParseProblemException pe) {
  return false;
}
```

because it is obvious that only `parseExpression` can throw `ParseProblemException`.

This is now short and clear.  You can make a similar change elsewhere.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -472,41 +575,118 @@ private static Receiver parseParameter(String s, FlowExpressionContext context)
...
         }
-
-        String arguments = s.substring(i + 1, rparenPos);
-        String remaining = s.substring(rparenPos + 1);
-        return Pair.of(Pair.of(ident, arguments), remaining);
     }
 
-    private static boolean isMethodCall(String s, FlowExpressionContext contex) {
-        Pair<Pair<String, String>, String> result = parseMethodCall(s);
+    private static boolean isMethodCall(String s, FlowExpressionContext context) {
```
mernst: Please add Javadoc comments for all routines that you added or changed.

mernst:
By the way, the expression `foo(x).y` is not a method call.  It is a field select, whose receiver expression happens to be a method call.  By contrast, `foo(x).bar(y)` is a method call of `bar`.
I can't tell what your code is supposed to do, but maybe there is a confusion around that terminology.

KankshaZ:
I am working on making the documentation and comments more clear.

> By the way, the expression foo(x).y is not a method call. It is a field select, whose receiver expression happens to be a method call. By contrast, foo(x).bar(y) is a method call of bar.
> I can't tell what your code is supposed to do, but maybe there is a confusion around that terminology.

I understand this. But the current version of the FlowExpression parser does not follow that terminology. 
I had inserted print statements in FlowExpressionParseUtil.java's current version [here](https://github.com/typetools/checker-framework/blob/526c5ff7852e5fd68d6b7d156f81587edbe5d990/framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java#L217) to confirm that the parser does not follow terminology. 

On observing the code snippet (given below) on running all tests, it can be seen that the `isMemberSelect()` function returns true for all field selects and certain method calls. Further, the original `isMethod()` function returns false for some methods. 

More clearly, `isMethod()` returns false for `myMethod(#1).get(0)` and considers `myMethod(#1)` and `get(0)` method calls only individually. 

This confused me as well. But I did not want to make any major change without discussion, and hence completed the work of replacing regex with the Java Parser without addressing this issue.

Here are the inserted print statements:
```
System.out.println("s: " + s);
System.out.println("Is member select.");
System.out.println("select: " + select);
```
This is the snippet of the result I got on running all tests:
```
tests.NullnessFbcTest > run[nullness] STANDARD_OUT
    s: myMethod(#1).get(0)
    Is member select.
    select: Pair(myMethod(#1), get(0))
    s: myMethod2().get(0)
    Is member select.
    select: Pair(myMethod2(), get(0))
    s: myMethod(#1).get(0)
    Is member select.
    select: Pair(myMethod(#1), get(0))
    s: myMethod2().get(0)
    Is member select.
    select: Pair(myMethod2(), get(0))
    s: this.menuBar
    Is member select.
    select: Pair(this, menuBar)
    s: FlowExpressionParsingBug.i
    Is member select.
    select: Pair(FlowExpressionParsingBug, i)

```

KankshaZ:
I have documented what my code does. I hope it clears up the confusion.

Thanks!

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -148,31 +143,134 @@ private static Pattern anchored(String regex) {
...
-        return parseMemberSelect(s) != null;
+    /** Return true iff s contains a dot (.) and a method call. */
+    private static boolean isMemberSelectOrMethodChain(String s, FlowExpressionContext context) {
+        return parseMemberSelectOrChainedMethodCalls(s, context) != null;
     }
 
     /**
-     * Matches a field access. First of returned pair is object and second is field.
+     * Matches expressions containing a dot (.) and a method call. If expression contains a method
+     * call whose caller is a method call as well (method chain), first of the returned pair is the
```
mernst: Should "caller" be "receiver" here?  I don't see how this has access to the caller.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -148,31 +143,134 @@ private static Pattern anchored(String regex) {
...
+     * caller, and second of the pair is the method invoked. For example.
```
mernst: Same comment as above about "caller".

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -148,31 +143,134 @@ private static Pattern anchored(String regex) {
...
+     * myMethod().myMethod2.get(0) returns Pair(myMethod().myMethod2(), get(0)). If expression is a
```
mernst: On this line `myMethod2` is once a field (which is confusing) and once a method.  Are there parentheses missing?

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -148,31 +143,134 @@ private static Pattern anchored(String regex) {
...
-    private static boolean isMemberSelect(String s, FlowExpressionContext context) {
-        return parseMemberSelect(s) != null;
+    /** Return true iff s contains a dot (.) and a method call. */
+    private static boolean isMemberSelectOrMethodChain(String s, FlowExpressionContext context) {
+        return parseMemberSelectOrChainedMethodCalls(s, context) != null;
     }
 
     /**
-     * Matches a field access. First of returned pair is object and second is field.
+     * Matches expressions containing a dot (.) and a method call. If expression contains a method
```
mernst: Should this "contains" be "is"?  If you mean "contains" then the expression might contain many different method calls that fit this description.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
mernst: If the expression contains are multiple method calls, which one is chosen?  This comment does not make it clear.
Is the last top-level method call in the expression the one chosen, or something else?

KankshaZ:
1. I made the requested changes and added punctutation where it was missing.
2. I am working on changing the design. Will give you an update by tonight.

The problem statement described here: [https://github.com/typetools/checker-framework/issues/1614](https://github.com/typetools/checker-framework/issues/1614) or on the [Project-Ideas page](https://rawgit.com/typetools/checker-framework/master/docs/developer/gsoc-ideas.html#flow-expression-parser) does not clearly specify that the goal was to redesign the FlowExpression Parser. From the problem statement, I understood that the problem was with buggy regular expressions, and the aim was to replace regular expressions with calls to the Java Parser.

mernst:
The link you gave says, "The current flow expression parser needs to be replaced by a real parser, such as JavaParser,"  That's a replacement of the parser, not of small parts of it.  I'm sorry that wasn't clear, but thanks for working on this.

mernst:
This pull request is the wrong approach.  We would welcome a pull request with a more complete solution.

PR #2064, opened 2018-07-05
Minor Documentation update for Option 3 of Section 31.6

ashishrana160796:
Complete command example will be more meaning & easy from user point of view while running any checker. As, `processor` as argument is specified for` javac `but for `java` it is not mentioned that clearly.

```
java [ options ] -jar file.jar [ arguments ]
arguments
The arguments passed to the main function.
```

reference links :
https://docs.oracle.com/javase/7/docs/technotes/tools/windows/javac.html
https://docs.oracle.com/javase/7/docs/technotes/tools/windows/java.html


mernst:
Thanks for the improvement!

PR #2077, opened 2018-07-22
Add class BigInteger in package.math 

NhatDinh:


NhatDinh:
Please review and let me know if any changes are needed
@kelloggm @panacekcz

framework/src/main/java/org/checkerframework/common/value/statically-executable.astub
```diff
@@ -334,6 +334,57 @@ class String{
...
 
+package java.math;
+
+class BigInteger {
+    @StaticallyExecutable BigInteger(@IntRange(from = -1, to = 1) int signum, byte[] magnitude);
+    @StaticallyExecutable BigInteger(byte[] val);
+    @StaticallyExecutable BigInteger(int bitLength, int certainty, Random rnd);
+    @StaticallyExecutable BigInteger(int numBits, Random rnd);
+    @StaticallyExecutable BigInteger(String val);
+    @StaticallyExecutable BigInteger(String val, @IntRange(from = 2, to = 36) int radix);
```
kelloggm: None of these constructors are `@StaticallyExecutable`, because they are not guaranteed to return the exact same value given the same arguments. After all, constructors usually allocate new objects! The other annotations here seem fine, but those should all be removed. As a general rule, I'd suggest only writing annotations you're really confident of in the JDK - they will be trusted by the compiler!

framework/src/main/java/org/checkerframework/common/value/statically-executable.astub
```diff
@@ -334,6 +334,57 @@ class String{
...
+
+    @StaticallyExecutable BigInteger abs();
+    @StaticallyExecutable BigInteger add(BigInteger val);
+    @StaticallyExecutable BigInteger and(BigInteger val);
+    @StaticallyExecutable BigInteger andNot(BigInteger val);
```
kelloggm: Similar to the constructors, these methods cannot be `@StaticallyExecutable`: they allocate new objects.

kelloggm: In fact, no methods that return a non-primitive type in this class are statically executable. You should remove `@StaticallyExecutable` from them all.

NhatDinh:
I made the requested changes. I'm updating the jdk8.jar file right now. Can you have a quick review and let me know if changes made are right? Thank you.
@kelloggm

PR #2078, opened 2018-07-23
ConcurrentHashMap: annotate nullness of compute*

awturner:
The (re)mapping parameters and return types of compute, computeIfAbsent and computeIfPresent can be annotated with @PolyNull to indicate the relationship between their nullness.


checker/jdk/nullness/src/java/util/concurrent/ConcurrentHashMap.java
```diff
@@ -1755,7 +1758,9 @@ else if ((val = mappingFunction.apply(key)) != null) {
      * @throws RuntimeException or Error if the remappingFunction does so,
      *         in which case the mapping is unchanged
      */
-    public V computeIfPresent(K key, BiFunction<? super K, ? super V, ? extends V> remappingFunction) {
+    public @PolyNull V computeIfPresent(
+            K key,
+            BiFunction<? super K, ? super V, ? extends @PolyNull V> remappingFunction) {
```
awturner: Missing @Nullable here.

mernst: Can you explain this comment?  Does this mean you need to update your pull request?

awturner:
Sorry, I don't use GitHub regularly enough to know how to do it right.

Yes, I needed to update my PR. I thought that I had...?

checker/jdk/nullness/src/java/util/concurrent/ConcurrentHashMap.java
mernst: Please check the pull request (on GitHub).  If it needs changes, please make them.  If it's correct, please verify that.  Thanks!

PS: It's generally better to answer on the GitHub website.  Even though it's convenient to make a reply by email, that tends to put the response in the wrong location.

Welcome to GitHub; you'll be using it like a pro in no time.

awturner:
I thought I had updated it, with
https://github.com/typetools/checker-framework/pull/2078/commits/e83e00de17a07e62a9623333a5382bda8d850b7f.
What do I need to do to make this part of the PR?

checker/jdk/nullness/src/java/util/concurrent/ConcurrentHashMap.java
wmdietl: @awturner 
https://github.com/typetools/checker-framework/pull/2078/commits/e83e00de17a07e62a9623333a5382bda8d850b7f is part of the pull request, but it added `@Nullable` on line 1855 below, not here.
Did you also intend to have an annotation here?

mernst: @awturner Werner answered your question; could you take action or explain https://github.com/typetools/checker-framework/pull/2078/files#r204313117 ?
Thanks!

awturner: I have placed the @Nullable where I intended to.

wmdietl: So what did you mean with the `Missing @nullable here.` comment that started this thread?
Why the different signatures for the two methods?

awturner: I simply put the comment in the wrong place. It was intended as a note to self that I needed to fix it.

They are different signatures because the methods do different things: ConcurrentHashMap does not allow null values, so the current value passed to computeIfPresent is not null; the current value passed to compute is null if there is no mapping present.

mernst:
Thanks for the improved library annotations!

PR #2085, opened 2018-08-01
 Expand wildcards on classpath, as javac does

mernst:
The [daikon-typecheck](https://travis-ci.org/typetests/daikon-typecheck/builds/410510237) Travis job is currently failing.  This fixes it.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/util/CheckerMain.java
+++ b/framework/src/main/java/org/checkerframework/framework/util/CheckerMain.java
@@ -156,20 +156,20 @@ public class CheckerMain {
     protected List<String> createCpOpts(final List<String> argsList) {
         final List<String> extractedOpts = extractCpOpts(argsList);
         extractedOpts.add(0, this.checkerQualJar.getAbsolutePath());
-        return extractedOps;
+        return extractedOpts;
     }
 
     // Assumes that createCpOpts has already been run.
     protected List<String> createPpOpts(final List<String> argsList) {
-        final List<String> extractedOps = extractPpOpts(argsList);
-        if (extractedOps.isEmpty()) {
+        final List<String> extractedOpts = extractPpOpts(argsList);
+        if (extractedOpts.isEmpty()) {
             // If processorpath is not provided, then javac uses the classpath.
             // CheckerMain always supplies a processorpath, so if the user
             // didn't specify a processorpath, then use the classpath.
-            extractedOps.addAll(this.cpOpts);
+            extractedOpts.addAll(this.cpOpts);
         }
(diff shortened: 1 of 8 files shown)


framework/src/main/java/org/checkerframework/framework/util/CheckerMain.java
```diff
@@ -407,17 +409,65 @@ protected void addMainToArgs(final List<String> args) {
...
+        } else {
+            return Collections.singletonList(pathElement);
+        }
+    }
+
+    /** Return all the .jar and .JAR files in the given directory. */
+    private List<String> jarFiles(String directory) {
+        File dir = new File(directory);
+        File[] jarFiles =
+                dir.listFiles(
```
wmdietl: Using a lambda would make this much shorter:
`File[] jarFiles = dir.listFiles((d, name) -> name.endsWith(".jar") || name.endsWith(".JAR"));`
(I think, didn't test it...)

framework/src/main/java/org/checkerframework/framework/util/CheckerMain.java
```diff
@@ -407,17 +409,65 @@ protected void addMainToArgs(final List<String> args) {
...
+     * Given a path element that might be a wildcard, return a list of the elements it expands to.
+     * If the element isn't a wildcard, return a singleton list containing the argument.
+     */
+    private List<String> expandWildcards(String pathElement) {
+        if (pathElement.equals("*")) {
+            return jarFiles(".");
+        } else if (pathElement.endsWith(FILESEP_STAR)) {
+            return jarFiles(pathElement.substring(0, pathElement.length() - 1));
+        } else if (pathElement.equals("")) {
+            return Collections.<String>emptyList();
```
wmdietl: The explicit type argument isn't needed.

mernst: I tried without the explicit type argument, and I got a compiler error.

framework/src/main/java/org/checkerframework/framework/util/CheckerMain.java
mernst: You are right.  I forgot we were using Java 8. :-)

framework/src/main/java/org/checkerframework/framework/util/CheckerMain.java
wmdietl: That is strange. #2086 removes the type argument and builds successfully.

mernst: Weird.  I have no explanation for that.  I only added the type argument because it was needed (locally).  Feel free to remove it.  Thanks for investigating.

Issue #2023, opened 2018-06-12
Add an Automatic-Module-Name


lukaseder:
In order to maintain a dependency on the checker-framework, it would be great if you folks exported an `Automatic-Module-Name` for forwards compatibility, in case you ever do decide to modularise the checker framework. See:
http://branchandbound.net/blog/java/2017/12/automatic-module-name

This can be done without much effort or backwards compatibility concerns, even in a patch release.

mernst:
We have modularized the Checker Framework distribution into jar files, according to what we believe that clients may need.

The module names should reflect this modularization.  That is, every jar file should get its own module name, to permit importing at that level of granularity.

We will use the same names as Maven artifacts.  However, module names cannot use hyphens, so checker-qual will become checker.jar.  Here are the module names to use:
```
org.checkerframework.checker
org.checkerframework.checker.qual
org.checkerframework.dataflow
org.checkerframework.framework
org.checkerframework.javacutil
org.checkerframework.testlib
```

We will make these changes, but note that the Checker Framework does not yet support Java 9, so the utility of these will be limited until #1224 is fixed.  We are actively working on #1224, but don't know when that work will be complete.

lukaseder:
Ok, cool, thank you very much, Mike! Note, just in case you decide to be able to publish these new maven modules prior to supporting Java 9, the `Automatic-Module-Name` would definitely help creating a dependency on the relevant modules already ahead of time...

PR #2093, opened 2018-08-05
 Rename the testlib Maven artifact to framework-test

mernst:
There is a mismatch in the names "util", "test", and "testlib" as names for directory, package, and jar file.

util: the Checker Framework subprobject, in directory checker-framework/util
test: the Java package, in checker-framework/util/src/main/java/org/checkerframework/framework/test
testlib: the name of the jar file checker-framework/util/build/libs/testlib-2.5.4.jar

This is confusing; other subprojects use consistent names where this one uses three different names.
Furthermore, the chosen names are confusing because of other uses of those names:
The "util" subproject and directory is not related to other "util" directories:
    checker-framework/dataflow/src/main/java/org/checkerframework/dataflow/util
    checker-framework/framework/src/test/java/testlib/util
    checker-framework/framework/src/main/java/org/checkerframework/common/value/util
    checker-framework/framework/src/main/java/org/checkerframework/common/util
    checker-framework/framework/src/main/java/org/checkerframework/framework/util
The "testlib" jar file is not related to
    checker-framework/checker/src/test/java/testlib
    checker-framework/framework/src/test/java/testlib

This pull request makes the names consistent, to reduce confusion and to ease accommodating Java 9 modules (initially, issue #2023).
I used the name "framework-test" throughout.
Its name is similar to checker-qual which currently exists.

This pull request leaves the Javadoc taglet, ManualTaglet.java, in framework-test.  The taglet doesn't belong there, but this pull request doesn't make it any worse.  Moving it to a more appropriate place is necessary for Java 9 modularization, but can be deferred to another pull request, to keep this one small and focused.


mernst:
Merge https://github.com/typetools/checker-framework-inference/pull/85 after this.

PR #2105, opened 2018-08-09
Modify the try-catch block to give errors during wrong reflective call

Ravi-Roshan:
Modifies the try-catch block in DefaultReflectionResolver.java to always issue an error when 'Resolve.class.getDeclaredMethod' and its invocation doesn't work


wmdietl:
@Ravi-Roshan The code isn't formatted correctly and therefore the Travis CI build is failing.
Please run `./gradlew reformat`.

framework/src/main/java/org/checkerframework/common/reflection/DefaultReflectionResolver.java
```diff
@@ -520,11 +531,7 @@ private boolean isUnknownMethod(MethodInvocationTree tree) {
             if (result.isEmpty()) {
                 debugReflection("Unable to resolve method: " + className + "@" + methodName);
             }
-        } catch (SecurityException
-                | NoSuchMethodException
-                | IllegalAccessException
-                | IllegalArgumentException
-                | InvocationTargetException e) {
+        } catch (SecurityException e) {
```
wmdietl: Which of the methods in this large block can throw the `SecurityException`? Make the scope as small as possible.

framework/src/main/java/org/checkerframework/common/reflection/DefaultReflectionResolver.java
```diff
@@ -487,10 +488,20 @@ private boolean isUnknownMethod(MethodInvocationTree tree) {
...
 
         List<Symbol> result = new ArrayList<>();
+        Method loadClass = null;
+        Symbol sym = null;
         try {
-            Method loadClass = Resolve.class.getDeclaredMethod("loadClass", Env.class, Name.class);
+            loadClass = Resolve.class.getDeclaredMethod("loadClass", Env.class, Name.class);
             loadClass.setAccessible(true);
-            Symbol sym = (Symbol) loadClass.invoke(resolve, env, names.fromString(className));
+            sym = (Symbol) loadClass.invoke(resolve, env, names.fromString(className));
```
wmdietl: Shouldn't these two calls be split up?
One depends on the javac API and should always work.
The other is reflectively looking for user code and might legitimately fail and only give a warning.

framework/src/main/java/org/checkerframework/common/reflection/DefaultReflectionResolver.java
Ravi-Roshan: No specific method was throwing `SecurityException`. So, I have removed it in the latest commit.

framework/src/main/java/org/checkerframework/common/reflection/DefaultReflectionResolver.java
```diff
@@ -487,47 +488,51 @@ private boolean isUnknownMethod(MethodInvocationTree tree) {
...
+        try {
+            sym = (Symbol) loadClass.invoke(resolve, env, names.fromString(className));
         } catch (SecurityException
-                | NoSuchMethodException
                 | IllegalAccessException
                 | IllegalArgumentException
                 | InvocationTargetException e) {
-            debugReflection("Exception during resolution of reflective method: " + e.getMessage());
+            ErrorReporter.errorAbort("Error in invoking reflective method.");
+        }
```
jyluo: the loading logic up to this point for setting `sym` seems to be the same as below for setting `symClass`, can you extract the reflective loading logic and the error throwing into a common private helper method which returns a `Symbol`?

framework/src/main/java/org/checkerframework/common/reflection/DefaultReflectionResolver.java
```diff
@@ -487,47 +488,51 @@ private boolean isUnknownMethod(MethodInvocationTree tree) {
         Names names = Names.instance(context);
 
         List<Symbol> result = new ArrayList<>();
+        Method loadClass = null;
```
wmdietl: Don't initialize local variables to dummy `null` values. Like this, the definite assignment checks by the compiler will not warn you if you missed a branch.

framework/src/main/java/org/checkerframework/common/reflection/DefaultReflectionResolver.java
```diff
@@ -544,41 +534,50 @@ private boolean isUnknownMethod(MethodInvocationTree tree) {
...
         List<Symbol> result = new ArrayList<>();
-        try {
-            Method loadClass = Resolve.class.getDeclaredMethod("loadClass", Env.class, Name.class);
-            loadClass.setAccessible(true);
-            Symbol symClass = (Symbol) loadClass.invoke(resolve, env, names.fromString(className));
-            if (!symClass.exists()) {
-                debugReflection("Unable to resolve class: " + className);
-                return Collections.emptyList();
-            }
+        Method loadClass = null;
```
jyluo: is this variable still used? If not, please remove.

framework/src/main/java/org/checkerframework/common/reflection/DefaultReflectionResolver.java
```diff
@@ -544,41 +534,50 @@ private boolean isUnknownMethod(MethodInvocationTree tree) {
...
+            debugReflection("Unable to resolve constructor!");
         }
         return result;
     }
 
+    private Symbol getSymbol(String className, Env<AttrContext> env, Names names, Resolve resolve) {
+        try {
+            Method loadClass = Resolve.class.getDeclaredMethod("loadClass", Env.class, Name.class);
+            loadClass.setAccessible(true);
+            try {
```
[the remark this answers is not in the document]
wmdietl: @Ravi-Roshan Could you please update this PR? It would be good to get into the next release.

PR #2108, opened 2018-08-12
Eliminate ErrorReporter class

mernst:


framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1037,7 +1037,7 @@ protected void checkVarargs(AnnotatedExecutableType invokedMethod, Tree tree) {
                 args = ((NewClassTree) tree).getArguments();
                 break;
             default:
-                throw new AssertionError("Unexpected kind of tree: " + tree);
+                throw new AssertionError("Unexpected kind of tree: " + tree); // unreachable
```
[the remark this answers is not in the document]
wmdietl: For this particular kind of error, default cases that should never happen, shouldn't we use `ErrorReporter.errorAbort` so that in a future update we get a better message?

mernst:
@wmdietl This is ready for your review again.  The implementation approach has completely changed.

I propose that we merge this despite the failure of the requireJavadoc task, because I don't understand MISC_COMPILER well enough to document it.  You could add the missing documentation if you prefer Travis to pass fully before merging.

mernst:
@wmdietl Thanks to a change upstream, the pull request now fully passes on Travis.

checker/src/main/java/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -118,8 +118,7 @@ private boolean maskIgnoresMSB(Kind maskKind, LiteralTree shiftAmountLit, Litera
             // Check that the shiftAmount most significant bits of the mask were 1.
             return mask == (1 << shiftAmount) - 1;
         } else {
-            // This shouldn't be possible.
-            throw new RuntimeException("Invalid Masking Operation");
+            throw new RuntimeException("Invalid Masking Operation"); // unreachable
```
wmdietl: For some of these the previous comment about `shouldn't happen` was more correct than `unreachable`.
If for whatever reason Java 12 adds a new maskKind, this branch will get executed.
I would use `unreachable` only if its really an imprecision in the analysis.

wmdietl: Also, here and elsewhere, why use a `RuntimeException` or `IllegalArgumentException` instead of `BugInCF`?

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
```diff
@@ -262,25 +262,21 @@ public boolean shouldResolveReflection() {
...
                 String msg;
-                if (err instanceof CheckerError) {
-                    CheckerError ce = (CheckerError) err;
-                    if (ce.userError) {
-                        // Don't add another stack frame, just show the message.
-                        throw ce;
-                    } else {
-                        msg = err.getMessage();
-                    }
+                // TODO: Handle if UserError?
```
wmdietl: Yes, we should handle `UserError` here to provide the user message without being wrapped in another exception.

PR #2109, opened 2018-08-13
Use ParameterizedMethodType instead of Pair

mernst:


mernst:
@wmdietl The Travis job will pass when merged as you can see from [the mfu-type branch here](https://travis-ci.org/mernst/checker-framework/branches).  It's failing in this pull request because it's looking for a branch of checker-framework-inference named `mfu-type`, and that is in the `mernst` organization rather than the `typetools` organization.  (This indicates that in the future, these coordinated pull requests should be from branches in the `typetools` organization when possible.)

TL;DR this is ready for review despite the (spurious) Travis failure.

checker/src/main/java/org/checkerframework/checker/guieffect/GuiEffectVisitor.java
```diff
@@ -1,5 +1,7 @@
 package org.checkerframework.checker.guieffect;
 
+import static org.checkerframework.framework.type.AnnotatedTypeFactory.ParameterizedMethodType;
```
smillst: It's a bit strange that this is statically imported

mernst:
@wmdietl I merged this because it got one approval, you are traveling, and other pull requests require it.  We can revisit the class and field names later if you have better suggestions.

checker/src/main/java/org/checkerframework/checker/guieffect/GuiEffectVisitor.java
wmdietl: Overall, replacing the `Pair` with a dedicated class is definitely good.
I also don't like the static imports. Are they needed because `ParameterizedMethodType` is an inner class?

mernst: > Are they needed because ParameterizedMethodType is an inner class?

Yes.  The alternative is writing `AnnotatedTypeFactory.ParameterizedMethodType` everywhere that `ParameterizedMethodType` is currently used.

smillst: I thought static imports where for importing methods and fields -- not classes.  The code compiles even when static is removed.

wmdietl: I agree with @smillst . The `static` isn't needed and the code would be nicer without.

Issue #2107, opened 2018-08-09
TreeUtils#isThisCall returns true when it should return false


smillst:
```
import org.checkerframework.checker.tainting.qual.PolyTainted;

public abstract class PolyReceiverTainted {

    abstract @PolyTainted int method(@PolyTainted PolyReceiverTainted this);
    @PolyTainted int method2(@PolyTainted PolyReceiverTainted this) {
        // Error
        return this.method();
    }

    @PolyTainted int method3(@PolyTainted PolyReceiverTainted this) {
        // No error
        return method();
    }
}
```

PR #2112, opened 2018-08-14
Reflection Resolution:

smillst:
Verify that the first argument to an invoke call is a subtype of the possible class.

The fix for https://github.com/typetools/checker-framework/issues/2107 uncovered this bug.  The test cases with `this` were crashing that fix.  This fixes the crash.  With the crash fixed, the test case with `this` correctly issued errors.  But, since the point of those tests were to check the reflection resolution was improving the precision of the checker, I had to change them to something that would pass.

### DIFF
--- a/framework/tests/reflection/MethodTest.java
+++ b/framework/tests/reflection/MethodTest.java
@@ -8,6 +8,7 @@ public class MethodTest {
 
     @Sibling1 int sibling1;
     @Sibling2 int sibling2;
+    @ReflectBottom SuperClass superClass;
 
     public void real_class() {
         try {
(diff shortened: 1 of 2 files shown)


framework/tests/reflection/MethodTest.java
```diff
@@ -21,35 +22,35 @@ public void real_class() {
         }
     }
 
-    public void pass1(@ReflectBottom MethodTest this) {
+    public void pass1() {
         try {
             Class<?> c = Class.forName("MethodTest$SuperClass");
             Method m = c.getMethod("getA", new Class[] {});
-            @Sibling1 Object a = m.invoke(this, (@ReflectBottom Object[]) null);
+            @Sibling1 Object a = m.invoke(superClass, (@ReflectBottom Object[]) null);
```
wmdietl: There no longer is a test case that has `m.invoke(this, ....)` in a passing situation. Would having such a test continue to be useful?

smillst:
The fix for https://github.com/typetools/checker-framework/issues/2107 uncovered this bug.  The test cases with `this` were crashing that fix.  This fixes the crash.  With the crash fixed, the test case with `this` correctly issued errors.  But, since the point of those tests were to check the reflection resolution was improving the precision of the checker, I had to change them to something that would pass.

PR #2113, opened 2018-08-14
Treat android.support.annotation.MainThread as an alias for UIEffect

lazaroclapp:
This makes the annotation `@MainThread` imply `@UIEffect` for the `guieffect` checker. This is expected by the Rx Thread/Effect Checker at Uber and getting this plus #1977 landed would allow us to use the main-line Checker Framework as our base again.


lazaroclapp:
We also have a small performance optimization we could upstream, a change to stub-parser warnings, and we internally change a few annotations from `RUNTIME` to `CLASS` retention. I'd be happy to submit PRs for those as well, which would allow us to deprecate our internal fork entirely, but the two PRs already up are enough for us to base the open-source version of our checker on the normal Checker Framework releases.

mernst:
Thanks for the contribution.
Feel free to submit the other changes (as separate pull requests).  We would be happy to have the enhancements, and it will be good to let you deprecate your fork.

PR #2117, opened 2018-08-15
Reuse the treePathCache across multiple type factories.

lazaroclapp:
This is only true when the Compilation Unit also remains the same.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
@@ -295,7 +295,7 @@ public class AnnotatedTypeFactory implements AnnotationProvider {
      */
     public boolean shouldCache;
 
-    /** Keep track of current compilation unit across multiple type factories that share caches */
+    /** Keep track of current compilation unit across multiple type factories that share caches. */
     private static int currentCompilationUnitHash = -1;
 
     /** Size of LRU cache if one isn't specified using the atfCacheSize option. */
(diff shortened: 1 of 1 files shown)


wmdietl:
@lazaroclapp Sorry for ignoring this PR for so long. We discussed it when you initially made it, but didn't understand the logic.

I now had another look and still don't understand the logic. :-)
Each ATF has their own independent TreePathCacher instance.
Once setRoot is called, it's called on all instances and will reset all of them.
Do you ever observe a call of setRoot with the same CompilationUnitTree?

In #2296 I tried sharing the TreePathCacher between multiple sub-checkers and AnnotatedTypeFactories, which I would have expected would benefit especially the Index Checker, which has many separate ATFs.
However, the index tests are about as fast as before.

Could you try #2296 and see whether it improves performance?

wmdietl:
Some timing from daikon-typecheck, for this branch:
https://travis-ci.org/typetests/daikon-typecheck/builds/492742723

The branch in #2296:
https://travis-ci.org/typetests/daikon-typecheck/builds/492734820

No change:
https://travis-ci.org/typetests/daikon-typecheck/builds/492737362

#2296 is the fastest, but it's probably all just Travis noise.
In particular, the Index Checker test, which should see the largest benefits from these changes, still times out.

I still like #2296, as it conceptually makes sense and cleaned up that code a bit.
Do you see bigger changes?

lazaroclapp:
Hey @wmdietl, sorry for the delayed response. Unfortunately, we are no longer set up to test this change internally anymore. I do remember this having a performance impact, but without being able to test it, I can't really tell you for certain that it did, or if/when/how we saw the else case of this check being triggered:

```
if (currentCompilationUnitHash != root.hashCode()) {
 treePathCache.clear();
 currentCompilationUnitHash = root.hashCode();
}
```

I am going to close this PR and will re-open if it if we are ever in a position to test this change in the original scenario in which we came up with it. Sorry about that.

wmdietl:
@lazaroclapp Thanks for checking and please do let us know if you see something that helps improve performance!

PR #2118, opened 2018-08-15
Make guieffect qualifiers CLASS retention (not RUNTIME).

lazaroclapp:
This is one of the small internal patches mentioned in #2113 and is **not** needed for us to open source the RxThreadEffectChecker, so it is not high priority to us. The idea is to avoid these annotations being preserved into the APKs of analyzed Android apps, since we never check them after building. It probably doesn't make much of a difference for plain Java apps, but at least in the way our build system works, `CLASS` annotations and not retained after dexing. Chances are other type annotations could be switched as well, but these are the only ones we use.

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/guieffect/qual/AlwaysSafe.java
+++ b/checker/src/main/java/org/checkerframework/checker/guieffect/qual/AlwaysSafe.java
@@ -20,6 +20,6 @@ import org.checkerframework.framework.qual.SubtypeOf;
 @DefaultQualifierInHierarchy
 @ImplicitFor(literals = LiteralKind.NULL)
 @Documented
-@Retention(RetentionPolicy.RUNTIME)
+@Retention(RetentionPolicy.CLASS)
 @Target({ElementType.TYPE_USE, ElementType.TYPE_PARAMETER})
 public @interface AlwaysSafe {}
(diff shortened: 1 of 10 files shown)


csgordon:
Looks good to me.

mernst:
@lazaroclapp There is a FAQ about why type annotations are declared with `@Retention(RetentionPolicy.RUNTIME)`:
https://checkerframework.org/manual/#faq-runtime-retention
Can you configure the compiler to strip these out (that is, not preserve them into the APK) if you don't care about them?
If your concern is size and it is impossible to strip them out, what are the APK sizes with and without the annotations?

lazaroclapp:
@mernst I see your point there. The original issue wasn't about APK sizes, although that's important too, but about certain annotations not being compatible with the version of Java that Android tooling can deal with, see below.

@kageiit @msridhar I remember that the issue here was that type use annotations in particular (e.g. `@UI`) produced bytecode incompatible with Java 6, which meant we couldn't convert them to dex: a) Is this still true in our current pipeline? b) If yes, how easy is it to just drop these annotations right before dexing? (They should still be on intermediate jars as long as we are building stuff where the RxThreadEffectChecker might need to look for them)

kageiit:
The biggest concern with runtime annotations is that types annotated with such annotations are always kept in the main dex by the default android gradle plugin which is used by 99% of android users in the world. This means that using these annotations can effectively lead to main dex overflows and any users of this library effectively unable to build their android apps.

`CLASS` retention annotations are stripped out before packaging by all build systems as a convention. Another option would be to provide two separate annotation artifacts `checker-qual-annotations` (runtime retention and can be used as usual for jvm based applications) and `checker-qual-annotations-android` which can have class retention. The classes can be exactly the same in both artifacts with the same package names etc.

PR #2119, opened 2018-08-15
StubParser warning changes (AstubWarnIfNotFoundIgnoresClasses and Wildcards)

lazaroclapp:
* Add `-AstubWarnIfNotFoundIgnoresClasses`, which allows us to use `-AstubWarnIfNotFound` in a way that ignores missing classes due to packages that are distributed across more than one JAR
* Report a better error, with type and file info, when the code has a wildcard type but the stub has something else, rather than giving a generic cast error. There is probably more generic code-type to stub-type missmatch checking that could be done here, but this is the case that has bitten us before.

### DIFF
--- a/docs/manual/annotating-libraries.tex
+++ b/docs/manual/annotating-libraries.tex
@@ -675,6 +675,12 @@ overrides the \<-AstubWarnIfNotFound> command-line option, and no warning
 will be issued.
 
 Use command-line option
+\<-AstubWarnIfNotFoundIgnoresClasses in conjuntion with the previous option
+to report only missing methods/fields, but ignore missing classes, even if
+other classes from the same package are present.
+Useful if a package spans more than one jar.
+
+Use command-line option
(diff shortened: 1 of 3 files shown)


lazaroclapp:
This is the last of the small internal patches mentioned in #2113 . It has proved pretty useful at finding issues with our stubs, since `-AstubWarnIfNotFound` on its own causes problems with our build (same package can span multiple jars, not all of which are used to build every CU). Took a bit longer to get this one ready, as our internal version was just changing the behavior of `-AstubWarnIfNotFound` . Instead, here we made this an option and added a line to the docs.

lazaroclapp:
Np. Thanks for looking at all of this so quickly. Doesn't look like I can merge them, but they are all complete, I believe, unless there is something you think needs to be changed. 

Also, I almost forgot to mention: A lot of the credit for these last three improvements actually goes to @bennostein who originally added them to our fork 😃

mernst:
I'm waiting for the Travis tests to pass before merging.

PR #2120, opened 2018-08-16
 Remove references to Java 7

mernst:
Should 
  `checker-framework/docs/examples/GradleExamples/`
be removed?  Its README says it is old.
Should
  `checker-framework/checker/src/main/java/org/checkerframework/checker/nullness/compatqual/`
be removed?  It's only for the purpose of Java 7 compatibility.


smillst:
>Should
checker-framework/docs/examples/GradleExamples/
be removed? Its README says it is old.

Yes.

>Should
checker-framework/checker/src/main/java/org/checkerframework/checker/nullness/compatqual/
be removed? It's only for the purpose of Java 7 compatibility.

Yes, we already discussed doing so, but I have had a chance to do so.  (It needs to be removed from the release scripts.)

PR #2141, opened 2018-08-27
Adds support for Generic signatures

gsaga:
This pull request:

1. Fixes `signature-types.svg` so that it appears correctly in pdf manual.

2. Adds annotations for generic signatures to signature checker.

I couldn't add `stringPattern`s for the new annotations since `java.util.regex` doesn't support recursive regex. Maybe they could be added if the checker framework used this regex library instead of standard library(https://github.com/florianingerl/com.florianingerl.util.regex).

This pull request just adds the annotations without `stringPattern`s.


mernst:
Sorry for the delay in responding.  The type introduction feature for string literals is handy.  (That is what `stringPattern` enables.)  It would be convenient to have either `stringPattern` or to write code to obtain the same effect.  I don't have a problem taking a dependency on the more capable regex library.

checker/src/main/java/org/checkerframework/checker/signature/qual/ArrayTypeSignature.java
```diff
@@ -0,0 +1,20 @@
+package org.checkerframework.checker.signature.qual;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Target;
+import org.checkerframework.framework.qual.SubtypeOf;
+
+/**
+ * Represents the signature of an array reference type as defined in <a
+ * href="https://docs.oracle.com/javase/specs/jvms/se9/html/jvms-4.html#jvms-4.7.9.1">Java Virtual
+ * Machine Specification, section 4.3.2</a>.
```
mernst: This section number is inconsistent with the URL.

checker/src/main/java/org/checkerframework/checker/signature/qual/ClassSignature.java
```diff
@@ -0,0 +1,17 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Target;
+import org.checkerframework.framework.qual.SubtypeOf;
+
+/**
+ * A class signature encodes type information about a (possibly generic) class declaration. It
+ * describes any type parameters of the class, and lists its (possibly parameterized) direct
+ * superclass and direct superinterfaces, if any. A type parameter is described by its name,
+ * followed by any class bound and interface bounds.
+ *
```
mernst: Can you please add an example of a class signature?

docs/manual/signature-checker.tex
```diff
@@ -14,7 +14,10 @@ \chapter{Signature String Checker for string representations of types\label{sign
 
 \section{Signature annotations\label{signature-annotations}}
 
-Java defines six formats for the string representation of a type.
+\subsection{String representation of types\label{signature-annotations-types}}
+
+Java defines seven formats for the string representation of a type.
```
mernst: I adjusted the text to reflect my understanding of the number of formats.

Should `@JavaTypeSignature` be listed as a seventh representation for types, both in the list in section 15.1 "Signature annotations" and also in the table in that section?  (I think so.)
I think that will clarify the relationship between Java type signatures and the other representations of Java types.

checker/src/main/java/org/checkerframework/checker/signature/qual/GenericSignature.java
```diff
@@ -0,0 +1,31 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Target;
+import org.checkerframework.framework.qual.SubtypeOf;
+
+/**
+ * Represents a generic type signature as defined in <a
+ * href=https://docs.oracle.com/javase/specs/jvms/se9/html/jvms-4.html#jvms-4.7.9.1>JVM specs</a>.
+ * Can be one of the following:
+ *
+ * <p>1. JavaTypeSignature
```
mernst: Here, I think it will lead to less documentation duplication to move the documentation to the subtypes, and link from here to the subtypes (there aren't currently links).

checker/src/main/java/org/checkerframework/checker/signature/qual/ArrayTypeSignature.java
```diff
@@ -0,0 +1,20 @@
+package org.checkerframework.checker.signature.qual;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Target;
+import org.checkerframework.framework.qual.SubtypeOf;
+
+/**
+ * Represents the signature of an array reference type as defined in <a
+ * href="https://docs.oracle.com/javase/specs/jvms/se10/html/jvms-4.html#jvms-4.7.9.1">Java Virtual
+ * Machine Specification, section 4.3.2</a>.
```
mernst: This section number is inconsistent with the URL.

checker/src/main/java/org/checkerframework/checker/signature/qual/TypeVariableSignature.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Target;
+import org.checkerframework.framework.qual.ImplicitFor;
+import org.checkerframework.framework.qual.SubtypeOf;
+
+/**
+ * Represents the signature of a Type Variable as defined in <a
+ * href="https://docs.oracle.com/javase/specs/jvms/se10/html/jvms-4.html#jvms-4.7.9.1">Java Virtual
+ * Machine Specification, section 4.3.2</a>.
```
mernst: The section number is inconsistent with the URL.

checker/src/main/java/org/checkerframework/checker/signature/qual/GenericSignatureWithTypeParameters.java
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * A signature that represents a class or method or declaration.
+ *
+ * @see GenericSignature
+ * @see MethodSignature
+ * @see ClassSignature
+ */
+@SubtypeOf(GenericSignature.class)
+@Target({ElementType.TYPE_USE, ElementType.TYPE_PARAMETER})
+public @interface GenericSignatureWithTypeParameters {}
```
mernst: Since this represents the signature for a declaration, I think it would be simpler and clearer to give it a name that reflects that fact, such as `GenericSignatureForDeclaration` or even `SignatureForDeclaration`.

docs/manual/signature-checker.tex
gsaga: It might not be possible to add another column for `JavaTypeSignature`s since they're the same as `FieldDescriptor`s for non generic types. I've added brief descriptions of these types to section 15.1.2.

gsaga:
I've updated the pull request with changes proposed in the review.

PR #2144, opened 2018-08-28
Add some methods to AnnotationMirrorMap and AnnotationMirrorSet.

smillst:
Used by PR #2143.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/util/AnnotationMirrorMap.java
+++ b/framework/src/main/java/org/checkerframework/framework/util/AnnotationMirrorMap.java
@@ -123,9 +123,4 @@ public class AnnotationMirrorMap<V> implements Map<AnnotationMirror, V> {
     public Set<Entry<AnnotationMirror, V>> entrySet() {
         return shadowMap.entrySet();
     }
-
-    /** @return an empty annotation mirror map */
-    public static <V> AnnotationMirrorMap<V> emptyMap() {
-        return new AnnotationMirrorMap<>();
-    }
 }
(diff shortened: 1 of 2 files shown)


framework/src/main/java/org/checkerframework/framework/util/AnnotationMirrorMap.java
```diff
@@ -116,4 +125,9 @@ public void clear() {
     public Set<Entry<AnnotationMirror, V>> entrySet() {
         return shadowMap.entrySet();
     }
+
+    /** @return an empty annotation mirror map */
+    public static <V> AnnotationMirrorMap<V> emptyMap() {
```
smillst: Why is this needed?

framework/src/main/java/org/checkerframework/framework/util/AnnotationMirrorSet.java
```diff
@@ -133,4 +133,16 @@ public boolean removeAll(Collection<?> c) {
...
         shadowSet.clear();
     }
+
+    /**
+     * Returns a new {@link AnnotationMirrorSet} that contains {@code value}.
+     *
+     * @param value AnnotationMirror to put in the set
+     * @return a new {@link AnnotationMirrorSet} that contains {@code value}.
+     */
+    public static AnnotationMirrorSet singleton(AnnotationMirror value) {
```
smillst: Wrong name.  Look at Set implementation.

smillst: This method has the same name as [Collections#singleton](
https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#singleton(T)).  So, I think it is the right name.  We could also call it `signletonSet`.

wmdietl: That method states: `Returns an immutable set containing only the specified object.`
So should this method also return something immutable?
Or the javadoc should highlight this difference to the standard API behavior.

smillst: Good point.  I changed it to `singleElement`.

wmdietl:
Thanks, these extensions look good now.

PR #2145, opened 2018-08-28
Change the AnnotationMirror Comparator to allow null.

smillst:
This is mainly to allow debugging of AnnotationMirror sets that erroneously contain null.


wmdietl:
I don't understand the motivation.
This now hides the fact that the sets erroneously contain null and produce some result.
I prefer having the NPE and fixing the illegal use, instead of just incorrectly comparing null.

PR #2149, opened 2018-08-31
fix eclipse project files to not depend on jsr308-langtools

jyluo:
The eclipse projects now depend on `tools.jar` and `src.zip` given by the JRE, which is symlinked in `AFU/lib`. Note that these two files are also required by the `asmx`, `scene-lib`, and `AFU` projects, hence its symlink placement in `AFU/lib`.

PR https://github.com/typetools/annotation-tools/pull/180 updates the eclipse project paths for `asmx`, `scene-lib`, and `AFU`, and should be merged at the same time as this PR.

The updated manual instructions for Configure Eclipse to edit the Checker Framework requires the user to download `src.zip` and create the symlinks to the two files, so that the eclipse projects remain OS independent and easier to maintain.

Also change "util" project name to "framework-test".

### DIFF
--- a/docs/manual/troubleshooting.tex
+++ b/docs/manual/troubleshooting.tex
@@ -947,18 +947,41 @@ These instructions are relevant if you use Eclipse as your IDE\@.
 \begin{enumerate}
 \item Follow the instructions earlier in Section~\ref{build-source} to
 clone and build all projects from their sources.
+\item Download the source code zip for your JDK:
+On Linux, run `sudo apt-get -qqy install openjdk-8-source`.
+On Window, re-run the JDK install wizard and in the custom setup panel, ensure
+``Source Code'' is selected for installation.
+If this step is successful, there should be a \<src.zip> file located in the
+directory pointed to by \<JAVA\_HOME>.
+\item Create symbolic links to \<tools.jar> and \<src.zip>:
+In a terminal (Linux / Mac) or command prompt with administrative
+priviliges (Windows), navigate to the
+\<\$CHECKERFRAMEWORK/../annotation-tools/annotation-file-utilities/lib>
+directory. Create symbolic links to \<\$JAVA\_HOME/lib/tools.jar> and to
+\<\$JAVA\_HOME/src.zip>. Ensure the symbolic links are named \<tools.jar> and
+\<src.zip>, respectively.
 \item Download Eclipse from
 the \href{https://www.eclipse.org/downloads/}{official Eclipse website}
 and install it.
-\item Run Eclipse.
-\item Enter the main Eclipse working screen and in the ``File'' menu,
-select ``Import'' $\rightarrow$ ``General'' $\rightarrow$ ``Existing
-Projects into Workspace''.
+\item Run Eclipse and enter the main Eclipse working screen.
+\item Checker Framework's sources depend on classes packaged in \<rt.jar> from
+the JRE. Attaching the source code to \<rt.jar> makes it easier to understand
+and work with these classes:
+In the ``Window'' menu, select ``Preferences'' $\rightarrow$ ``Java''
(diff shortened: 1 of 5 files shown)


docs/manual/troubleshooting.tex
```diff
@@ -947,18 +947,32 @@ \subsection{Configure Eclipse to edit the Checker Framework\label{building-eclip
...
+\<tools.jar> and \<src.zip>.
 \item Download Eclipse from
 the \href{https://www.eclipse.org/downloads/}{official Eclipse website}
 and install it.
-\item Run Eclipse.
-\item Enter the main Eclipse working screen and in the ``File'' menu,
-select ``Import'' $\rightarrow$ ``General'' $\rightarrow$ ``Existing
-Projects into Workspace''.
+\item Run Eclipse and enter the main Eclipse working screen.
+\item Checker Framework's sources depend on classes packaged in \<rt.jar> from
```
wmdietl: Is this really for rt.jar? If a user doesn't do anything, Eclipse wouldn't even show me javadoc for `java.util.List`?

wmdietl: Everything depends on `rt.jar`, so this sentence reads strange. Do you mean `tools.jar`?

docs/manual/troubleshooting.tex
```diff
@@ -947,18 +947,32 @@ \subsection{Configure Eclipse to edit the Checker Framework\label{building-eclip
...
+``Apply and Close''.
+\item In the ``File'' menu, select ``Import'' $\rightarrow$ ``General''
+$\rightarrow$ ``Existing Projects into Workspace''.
 \item After the ``Import Projects'' window appears, select ``Select
 Root Directory'', and select the parent of the \code{\$CHECKERFRAMEWORK} directory.
 \item Ensure ``Search for nested projects'' is selected in the Options panel.
-\item Select all the projects in the folder except for ``personalblog-demo'',
-then select ``Finish'' to import the projects.
+\item Select all the projects in the list except for ``personalblog-demo''
+and ``jsr308-langtools'', then select ``Finish'' to import the projects.
```
wmdietl: We no longer clone jsr308-langtools. Why would it now show up in this list of projects?

jyluo: `./gradlew cloneAndBuildDependencies` runs './.travis-build-without-test.sh', and in that script we download and extract jsr308-langtools: https://github.com/typetools/checker-framework/blob/master/.travis-build-without-test.sh#L65

@smillst please remove the .project and .classpath files in the zip we distribute for jsr308-langtools, then it won't show up in the import here

docs/manual/troubleshooting.tex
jyluo: It would show javadoc for `java.util.List` without the source attachment, but it won't show the source code for that class.

jyluo: Agreed that it reads strange. Will fix that part.

As to dependencies, do we want to have sources for classes that are in `rt.jar` or only for classes in `tools.jar`?

docs/manual/troubleshooting.tex
wmdietl: If we already have a `jsr308-langtools` version cloned by `./gradlew cloneAndBuildDependencies` why not use it?
Why add the steps with `tools.jar` and `src.zip` if we have a clone of all of langtools?

jyluo: My understanding is that we don't want to clone `jsr308-langtools` anymore. Is this still true?

jyluo:
Per discussions offline, the eclipse projects will depend on the eclipse jsr308-langtools project downloaded by https://github.com/typetools/checker-framework/blob/master/.travis-build-without-test.sh#L65

I'm closing this PR to create another one that does depend on the jsr308 project

Issue #1863, opened 2018-03-02
Stub for javax.xml.parsers.FactoryFinder breaks annotation tools


cushon:
When compiled with a JDK 9 or newer javac this type annotation:

https://github.com/typetools/checker-framework/blob/bff92a84531d54efb74356df59bcd7e1f592e46c/checker/jdk/nullness/src/javax/xml/parsers/FactoryFinder.java#L290

results in a bogus `CLASS_EXTENDS` `RuntimeVisibleTypeAnnotations` attribute on the enclosing class:

```
  private static <T extends java.lang.Object> T findServiceProvider(java.lang.Class<T>);
    descriptor: (Ljava/lang/Class;)Ljava/lang/Object;
    flags: ACC_PRIVATE, ACC_STATIC
...
    RuntimeVisibleTypeAnnotations:
      0: #110(): METHOD_RETURN
      1: #110(): CLASS_EXTENDS, type_index=0, location=[TYPE_ARGUMENT(0)]
```

which breaks the annotation tools:

```
reading: javax/xml/parsers/FactoryFinder.class
Unknown error trying to extract annotations from: javax/xml/parsers/FactoryFinder
scenelib.annotations.el.AMethod cannot be cast to scenelib.annotations.el.AClass
Please submit a bug report at
  https://github.com/typetools/annotation-tools/issues
```

This is probably a javac bug: [JDK-8198945](https://bugs.openjdk.java.net/browse/JDK-8198945).

As a work-around, perhaps that type annotation could be removed? It's on an anonymous class, so it's not part of an API that's visible outside of the JDK, and doesn't seem necessary to include in the stub API.

mernst:
Sorry for the problem.  I have commented out the annotation you suggested.
We have not done a lot of testing with a Java 9 compiler, but that is on our roadmap.
Thanks for the report!

PR #2160, opened 2018-09-11
Comment out an annotation

cushon:
to work around another instance of
https://github.com/typetools/checker-framework/issues/1863.


mernst:
Thanks for noticing that.

PR #2165, opened 2018-09-15
Char int range value checker

txiang61:
Added a integral range (o, 65535)  in the Range file for char types. 
For all expression casting to char, the annotation on the target type is @intRange( from = 0, to = 65535) in value checker.


framework/src/test/java/tests/RangeTest.java
```diff
@@ -265,6 +276,28 @@ public void testShortRange() {
...
 
+    @Test
+    public void testCharRange() {
+        Range.ignoreOverflow = false;
+        for (Range range : ranges) {
+            Range result = range.charRange();
+            for (long value : values) {
+                if (value < range.from + CHAR_WIDTH
+                        && value > range.to - CHAR_WIDTH
+                        && !result.isCharEverything()) {
```
mernst: This test is inconsistent with the corresponding tests for byte and short.  Should this one be changed, or those?

txiang61: Addressed. Changed to be consistent with other tests.

mernst:
@txiang61 The Travis build is failing because:
```
Trailing whitespace: /checker-framework/framework/src/main/java/org/checkerframework/common/value/util/Range.java
```

Could you please fix that?

wmdietl:
@mernst Thanks, https://github.com/typetools/checker-framework/pull/2165/commits/0df98a369a8a3edb4c357072da08f6ed441a1857 fixed the failure.

.travis.yml
```diff
@@ -2,14 +2,15 @@ sudo: false
...
 
-before_cache:
-  - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
-  - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
-cache:
-  directories:
-    - $HOME/.gradle/caches/
-    - $HOME/.gradle/wrapper/
-    - $HOME/.m2/
+## Sometimes, the cache gets corrupted.  So don't cache.
```
wmdietl: @mernst These changes still show up here, even though that changeset was pulled in from master.
Is there something to do to un-confuse github?

wmdietl:
Thanks for fixing this! You can delete your local branch now.
In the future, for each pull request start from a fresh branch off of master.

PR #2175, opened 2018-09-29
update eclipse project classpaths and names

jyluo:
sort entries by src / con / lib / output, and by names within each kind

update output paths to correspond with gradle 4.5 outputs

update eclipse project name of `framework-test`


wmdietl:
Thanks for the improvements!

PR #2176, opened 2018-10-01
When expanding varargs, copy parameters rather than using the same object. Fixes #2171

### LINKED ISSUE #2171: Polymorphic varargs methods fail
When a variable argument formal parameter is annotated with the polymorphic type, it should be possible to pass arguments annotated with any type in the type hierarchy in any order. It looks like the checker framework incorrectly resolves the polymorphic type to the type of the first argument in the variable argument list, instead of taking their least upper bound.
Consider the following example to be checked with the nullness checker:

```

public class Issue24PolyNull {
    static void varArgsMethod(@PolyNull Object... args) {
    }
    static void callToVarArgsObject(@PolyNull Object a, @NonNull Object b) {
        varArgsMethod(b, a);
        varArgsMethod(a, b);
    }
}

```

We expect this code to type check without errors. But, executing the command `javac -processor nullness Issue24PolyNull.java` results in the following error:

```
Issue24PolyNull.java:9: error: [argument.type.incompatible] incompatible types in argument.
        varArgsMethod(b, a);
                         ^
  found   : @Initialized @PolyNull Object
```
Note that the call `varArgsMethod(a, b)` type checks correctly. It appears that for variable arguments, the type of the first argument is used to resolve the polymorphic type. For `varArgsMethod(b, a);`, `@PolyNull` resolves to `@NonNull` causing the error. For `varArgsMethod(a, b);`, `@PolyNull` resolves to `@PolyNull` correctly. So the solution would be for the checker framework to take the least upper bound of all arguments in the variable argument list while resolving such polymorphic annotations.

smillst:
Fixes #2171

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/util/AnnotatedTypes.java
+++ b/framework/src/main/java/org/checkerframework/framework/util/AnnotatedTypes.java
@@ -629,7 +629,7 @@ public class AnnotatedTypes {
 
         parameters = new ArrayList<>(parameters.subList(0, parameters.size() - 1));
         for (int i = args.size() - parameters.size(); i > 0; --i) {
-            parameters.add(varargs.getComponentType());
+            parameters.add(varargs.getComponentType().shallowCopy());
         }
 
         return parameters;
--- a/framework/src/main/java/org/checkerframework/framework/util/AnnotatedTypes.java
+++ b/framework/src/main/java/org/checkerframework/framework/util/AnnotatedTypes.java
@@ -629,7 +629,7 @@ public class AnnotatedTypes {
 
         parameters = new ArrayList<>(parameters.subList(0, parameters.size() - 1));
         for (int i = args.size() - parameters.size(); i > 0; --i) {
-            parameters.add(varargs.getComponentType().shallowCopy());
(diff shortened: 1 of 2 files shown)


framework/src/main/java/org/checkerframework/framework/util/AnnotatedTypes.java
```diff
@@ -629,7 +629,7 @@ public static AnnotatedTypeMirror leastUpperBound(
 
         parameters = new ArrayList<>(parameters.subList(0, parameters.size() - 1));
         for (int i = args.size() - parameters.size(); i > 0; --i) {
-            parameters.add(varargs.getComponentType());
+            parameters.add(varargs.getComponentType().shallowCopy());
```
wmdietl: Why is a shallow copy enough?
What happens with a poly annotation that is on a type argument?
Can you add a test case to ensure this doesn't cause a problem?

wmdietl: I expanded the test and verified that shallowCopy doesn't do the right thing. I changed this to deepCopy.

PR #2191, opened 2018-10-30
Rename an odd constructor to unparse()

mernst:


wmdietl:
The documentation test is failing.

mernst:
@wmdietl Thanks for pointing that out.  (It's weird, since Travis passed in my fork.)
I have added documentation and now it's passing.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesError.java
```diff
@@ -47,11 +47,13 @@ private boolean isFormalParamNameError(String expression) {
     /** An error message about that expression. */
     public final String error;
 
+    /** Create a DependentTypesError for the given expression and error message. */
```
wmdietl: Are these other constructors less odd than the one you renamed to `unparse`?
Would it be more consistent to only have one private constructors and three `unparse` methods?

mernst: The constructor you noted does no unparsing, so I feel naming it `unparse` would be surprising and misleading.  It's just an ordinary constructor that sets fields from its arguments.  I don't think it is surprising to users.

wmdietl: Ok, maybe better documentation for the parameters would be helpful. Both this constructor and `unparse` take a `String error` parameter, so I assumed that similar things happen with them.
The second constructor below does something with its parameter, so should that one be turned into a method?
Instead of splitting out the `unparse` method I would find clearer descriptions for the parameters more useful. With this change one has to decide whether to use a constructor or the `unparse` method, whereas before one had to pick the right constructor.

mernst:
`unparse` is a standard method name for a routine that parses a formatted output.  I find it much clearer than overloading the constructor.  It is used in rare circumstances, so most programmers won't need to find it (and they will find it easier to find with a descriptive name).  It can be removed when the ugly DependentTypesError design is changed (as suggested by TODOs elsewhere in the code).

PR #2194, opened 2018-10-31
Improve documentation for QualifierPolymorphism

mernst:


mernst:
@wmdietl 
In QualifierPolymorphism, why is the word "Top" in the name getPolymorphicQualifierTop?
Also, what is the purpose of
  "value", true
near the end of its implementation?  These details seem important, but they are missing from its specification.

There are three methods named "annotate" with documentation "Resolves polymorphism annotations for the given type.".  What does "resolve" mean in this context?  Does it mean to replace each polymorphic qualifier with a concrete one, based on a particular call site?

I am also confused about PolyCollector, which "instantiates", but I'm not sure what that means.  It would be helpful to define it or to refer to the standard definition.

framework/src/main/java/org/checkerframework/framework/type/poly/QualifierPolymorphism.java
```diff
@@ -73,24 +75,32 @@ static boolean isPolymorphicQualified(AnnotationMirror qual) {
     /**
      * Resolves polymorphism annotations for the given type.
      *
+     * <p>On entry, {@code type} is the type of a method declaration.
```
wmdietl: These comments about declaration vs. use types aren't correct.
On entry, these are already fully substituted method signatures in a particular calling context. They are not from a method declaration. These parts should be undone.

framework/src/main/java/org/checkerframework/framework/type/poly/AbstractQualifierPolymorphism.java
```diff
@@ -32,10 +32,15 @@
 import org.checkerframework.javacutil.TreeUtils;
 
 /**
- * Implements framework support for qualifier polymorphism.
+ * Implements framework support for qualifier polymorphism. Works for a particular type system,
```
wmdietl: I don't think the extra parts add anything useful.

mernst:
@wmdietl Could you please answer the questions in the comment from October 31?

framework/src/main/java/org/checkerframework/framework/type/poly/AbstractQualifierPolymorphism.java
mernst: Unless you object, I'm going to leave it there because it is helpful to people who aren't you. :-)

framework/src/main/java/org/checkerframework/framework/type/poly/QualifierPolymorphism.java
mernst: I'm not sure exactly what "these parts" refers to in your comment.  Can you please make the edits?

framework/src/main/java/org/checkerframework/framework/type/poly/AbstractQualifierPolymorphism.java
```diff
@@ -94,7 +100,7 @@ public AbstractQualifierPolymorphism(ProcessingEnvironment env, AnnotatedTypeFac
 
     /**
      * Reset to allow reuse of the same instance. Subclasses should override this method to clear
-     * their additional state; they must call the super implementation.
+     * their additional state and to the super implementation.
```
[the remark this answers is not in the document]
mernst: The previous version used different wording for "clear additional state" and "call the super implementation", as if those would be done in two different ways (and it wasn't clear how the second would be done).  The current wording makes clear that both are dove via overriding this method.  Feel free to change the wording, but please use consistent terminology when referring to the same concept.

PR #2204, opened 2018-11-04
Improve error messages about purity

mernst:
Make them more specific.

Merge after https://github.com/typetools/checker-framework/pull/2201 (at which point the tests will pass)


dataflow/src/main/java/org/checkerframework/dataflow/util/PurityChecker.java
```diff
@@ -209,12 +209,11 @@ public Void visitNewClass(NewClassTree node, Void ignore) {
...
-            if (!sideEffectFree && !deterministic) {
-                purityResult.addNotBothReason(node, "object.creation");
-            } else if (!deterministic) {
+            if (!deterministic) {
                 purityResult.addNotDetReason(node, "object.creation");
-            } else if (!sideEffectFree) {
-                purityResult.addNotSEFreeReason(node, "object.creation");
+            }
+            if (!sideEffectFree) {
+                purityResult.addNotSEFreeReason(node, "call.constructor");
```
wmdietl: The distinction between `object.creation` and the new `call.constructor` isn't clear.
It's also not clear why the same distinction isn't made for method calls.
Also, for method calls there is still the `addNotBothReason` case, whereas here it is gone.
Some comments would help understand this.

mernst: They handle different rules.
From `messages.properties`:
```
purity.not.deterministic.object.creation=object creation not allowed in deterministic method
purity.not.sideeffectfree.call.constructor=call to non-side-effect-free constructor not allowed in side-effect-free method
```

One is about determinism and the other is about side effects.  One forbids all object creation, whereas the other forbids some calls.  I felt that changing the names (rather than overloading, using the same messages for different underlying problems) made the code easier to understand.

The distinction doesn't need to be made for method calls

I agree it would be better to eliminate all uses of the ugly and complicated `addNotBothReason` throughout, but didn't want to make this PR even bigger nor delay it based on that refactoring, which I am not confident I can make quickly and correctly.

wmdietl: 1. The `node` on which the error is raised is a `NewClassTree` in all cases. It does not make sense to call that `object creation` in one error message and `constructor invocation` in another. At the moment the error messages are consistent. The difference in the two rules is whether it is about determinism or side-effects, not about the kind of tree.

2. Using `addNotBothReason` allows us to give one single error message for a single violation, instead of two separate error messages. This makes sense for method calls and makes equal sense for object creations.

Can I undo these two changes and merge the rest?

framework/src/main/java/org/checkerframework/common/basetype/messages.properties
```diff
@@ -46,24 +46,24 @@ type.arguments.not.inferred=Could not infer type arguments for %s.
 unallowed.access=access of the field (%s) is not permitted on receiver of type (%s)
 cast.redundant=Redundant cast;\ntype   : %s
 
-purity.deterministic.void.method=a deterministic method without return value is redundant
-purity.deterministic.constructor=a deterministic constructor is redundant
-purity.invalid.overriding=Incompatible purity declaration\nMethod\n  %s in %s\n  cannot override\n  %s in %s\nfound   : %s\nrequired: %s
+# TODO: The call.{constructor,method} messages should take an argument indicating which method or constructor is the problem
```
wmdietl: These errors are raised on the method invocation that is problematic. Repeating that seems redundant.

dataflow/src/main/java/org/checkerframework/dataflow/util/PurityChecker.java
mernst: 1. The naming is consistent.  The rules about side effects and purity are worded in terms of invocations and creations.  (`NewClassTree` is an instance of both.)  I believe it's better for the names to be consistent with the specification rather than consistent with an implementation detail about the name of the node.  This would be like requiring two messages to use the same key if one is about serialization and the other about security, just because both apply to `RSAPrivateKey`.

2. The benefits of one error message are minor, and the ugly code to achieve it has caused many maintenance problems over the years.  We should eliminate it entirely. These changes only make partial progress toward that goal (and are orthogonal to the rest of the pull request), but I don't see the benefit in the work of separating this into a separate pull request.

framework/src/main/java/org/checkerframework/common/basetype/messages.properties
mernst: I added this because the information in the current messages is not sufficient.  For example, it is helpful to inform the user override is being invoked.

PR #2211, opened 2018-11-15
Initialization documentation tweaks

mernst:


checker/src/main/java/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -518,7 +522,7 @@ protected AnnotationMirror getFreeOrRawAnnotationOfSuperType(TypeMirror type) {
                 annotation = createUnclassifiedAnnotation(superClass);
             }
         } else {
-            // Use Object as a valid super-class
+            // Use Object as a valid super-class, using overloaded version of constructors
```
wmdietl: It's not quite clear what the `constructors` refers to here, as it calls two methods below. Maybe clarify?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1879,6 +1879,8 @@ public AnnotatedDeclaredType getSelfType(Tree tree) {
         if (enclosingClass.getSimpleName().length() != 0 && enclosingMethod != null) {
             AnnotatedDeclaredType methodReceiver;
             if (TreeUtils.isConstructor(enclosingMethod)) {
+                // This seems questionable, since it's the type of `this` after the constructor
```
wmdietl: This seems like a `TODO` comment. Is this a concern for how FbC is implemented? Or also for other type systems?

mernst: It's for all initialization type systems (rawness too).  But probably not for anything else, because other type systems don't change the type of `this` throughout the constructor.

Yes, it is a TODO comment.  I was hoping you could double-check whether my concern is valid.

mernst: (I didn't edit to add "TODO" because that would resolve this conversation.)

PR #2213, opened 2018-11-21
Add `@EnsuresNonNullIf` to `nonNull`

seanmcl:


wmdietl:
Thanks for the contribution!
I've expanded it to also cover the `isNull` method.
I'll merge after the tests pass.

PR #2219, opened 2018-11-30
Fix getPolymorphicAnnotation()

t-rasmud:


framework/src/main/java/org/checkerframework/framework/util/MultiGraphQualifierHierarchy.java
```diff
@@ -325,14 +325,7 @@ public AnnotationMirror getPolymorphicAnnotation(AnnotationMirror start) {
         if (AnnotationUtils.containsSame(polyQualifiers.keySet(), polymorphicQualifier)) {
             return polyQualifiers.get(polymorphicQualifier);
         } else {
-            // No polymorphic qualifier exists for that hierarchy.
```
wmdietl: What code is calling `getPolymorphicAnnotation` and why are you running into an exception now?
If a particular type system doesn't have a poly qualifier, it shouldn't invoke `getPolymorphicAnnotation`. It would be more obvious for the type system designer to get the current BugInCF in that case, instead of returning null - which would then likely cause an NPE.

mernst: There is new framework code under construction that runs for every type system.
In order to avoid making the call, it would need to call some method to determine whether the type system has a polymorphic annotation.  It seems more direct to use this method for both purposes.

For non-framework code that is specific to a single type system, I agree that the type system designer knows whether there is a polymorphic qualifier and can avoid calling it.

wmdietl: `getPolymorphicAnnotation` is called by `QualifierHierarchy.findAnnotationInHierarchy` which is widely called in the framework.
Can you construct an example that triggers this exception? That might be possible for an example with `PolyAll` and one of the type systems that has no polymorphic qualifier.

Otherwise, I would prefer keeping a useful error message until the under-construction framework feature is available to review the change within that use case.

t-rasmud:
The under construction code no longer requires the call to `getPolymorphicAnnotation`.
Closing this PR.

PR #2222, opened 2018-12-04
Documentation

t-rasmud:


docs/manual/generics.tex
```diff
@@ -624,7 +624,7 @@
 
 \subsectionAndLabel{The \code{@PolyAll} qualifier applies to every type system}{polyall}
 
-Each type system has its own polymorphic type qualifier.
+A type system may have its own polymorphic type qualifier.
```
wmdietl: Would it be clearer as `Each type system may have its own...`?

Issue #2147, opened 2018-08-29
Annotations on enum constants in stub files are silently ignored


kelloggm:
All typecheckers silently ignore annotations written on stub files containing enums. For example, consider the following enum, defined in `SampleEnum.java`:

```java
public enum SampleEnum {
	FIRST,
        SECOND;
} 
```

Suppose that this code was available only as bytecode, but must be annotated. So, we write a stub file `SampleEnum.astub`:

```java
import org.checkerframework.checker.tainting.qual.*;

enum SampleEnum {
	@Untainted FIRST,
	@Tainted SECOND;
}
```

When this stub file is supplied to the compiler via `-Astubs=SampleEnum.astub`, I expect to see only one warning in the following client code (`EnumStubTest.java`):

```java
import org.checkerframework.checker.tainting.qual.*;

class EnumStubTest {
    void test() {
	requireEnum(SampleEnum.FIRST);
	// :: error: assignment.type.incompatible
	requireEnum(SampleEnum.SECOND);
    }

    void requireEnum(@Untainted SampleEnum sEnum) { }
}
```

Instead, errors are issued at both calls to `requireEnum`:
```bash
▶ checker/bin/javac -processor tainting EnumStubTest.java -Astubs=SampleEnum.astub
EnumStubTest.java:5: error: [argument.type.incompatible] incompatible types in argument.
	requireEnum(SampleEnum.FIRST);
	                      ^
  found   : @Tainted SampleEnum
  required: @Untainted SampleEnum
EnumStubTest.java:7: error: [argument.type.incompatible] incompatible types in argument.
	requireEnum(SampleEnum.SECOND);
	                      ^
  found   : @Tainted SampleEnum
  required: @Untainted SampleEnum
2 errors
```
If I move the annotations on `SampleEnum` into the `.java` file, the error goes away, so I suspect this is something in the stubparser. [Javaparser#1806](https://github.com/javaparser/javaparser/issues/1806) might be related.

I was able to reproduce the issue using the Value Checker, also, using `@StringVal` annotations instead of tainted/untainted.

mernst:
There is currently a pull request (https://github.com/javaparser/javaparser/pull/1812) open to fix https://github.com/javaparser/javaparser/issues/1806.  I propose that we wait for it to be merged, then pull upstream into stubparser.  If that doesn't resolve the issue, then we should investigate further.

wmdietl:
We do not use the symbol resolution part from JavaParser, so that pull request won't help.

[...]

PR #2242, opened 2019-01-15
Support annotations on enum constants in stub files

jyluo:
Add test suite and fix for https://github.com/typetools/checker-framework/issues/2147


jyluo:
@kelloggm due to the way CF tests work, I can't seem to just define the enum in 2147 in a separate file in the same test dir (or reuse Issue2156.java) and have it load the annotations from stub. As a simple substitution I use 2 enum constants in java.lang.Thread.State in a stub. Is this ok to you?

I'm waiting for a travis failure before pushing the stub to verify the fix.

kelloggm:
Yes, that seems like a reasonable solution.

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1281,6 +1309,21 @@ private VariableElement findElement(TypeElement typeElt, VariableDeclarator vari
...
+    /**
+     * Looks for an enum constant element in the typeElt and returns it if the element has the same
+     * name as provided. In case enum constant element is not found it returns null.
+     *
+     * @param typeElt type element where enum constant element should be looked for
+     * @param enumConstDecl the declaration of the enum constant
+     * @return enum constant element in typeElt with the provided name or null if enum constant
+     *     element is not found
+     */
+    private VariableElement findElement(
```
wmdietl: How about moving this a bit further up to be next to `private Element findElement(TypeElement typeElt, EnumDeclaration enumDecl)` and keeping the related methods together?

framework/src/test/java/tests/StubParserEnumTest.java
```diff
@@ -0,0 +1,24 @@
+package tests;
+
+import java.io.File;
+import java.util.List;
+import org.checkerframework.framework.test.FrameworkPerDirectoryTest;
+import org.junit.runners.Parameterized.Parameters;
+
+public class StubParserEnumTest extends FrameworkPerDirectoryTest {
```
wmdietl: I would also move this test to a jtreg test. Many other stub tests are in `checker/jtreg/stubs/`.

jyluo:
Done. I've also adjusted the test in checker/ to use the original example in the issue since in jtreg tests we can pre-compile the enum.

PR #2246, opened 2019-01-23
Rename areSameIgnoringValues to areSameByName

mernst:
This is clearer, shorter, and consistent with existing areSameByName methods.


kelloggm:
@mernst I just wanted to let you know that this change breaks a custom typechecker I'm working on, preventing it from moving to 2.6.0 without someone renaming its calls. When renaming methods in this way, can we use `@Deprecated` and keep the old version around, at least for a release or two, in the future? Thanks.

mernst:
@kelloggm Yes, that is a good idea.  This renaming was a reason that the version number jumped from 2.5.* to 2.6.*, but I could have kept it around, deprecated, nonetheless.

PR #2250, opened 2019-01-24
Don't call getDeclAnnotationNoAliases in ContractsUtils

mernst:
No tests fail.  Is this an acceptable change to make?


wmdietl:
What is your motivation for the change?
At the moment I think the assumption is that `@QualifierArgument` never has an alias, so it is faster to only look directly for that annotation and not consider possible aliases.
Is that assumption changing? Or is there another motivation?

mernst:
I wasn't aware of that assumption, which wasn't documented.
The code is more general without the assumption, a violation of the assumption might be difficult to track down to this line of code, and I don't think the performance hit will be severe.
getDeclAnnotationNoAliases was introduced to support a hack in the Lock Checker, and if we restrict its use to that hack, then it will be easier to undo the hack and eliminate the method in the future.

PR #2253, opened 2019-01-25
Don't use fromName; preserve annotation elements

mernst:
The new constructor
```
public AnnotationBuilder(Elements elements, AnnotationMirror annotation) {
```
is a bit gross.  Is there a nicer way to do that?


checker/src/main/java/org/checkerframework/checker/units/UnitsRelationsTools.java
```diff
@@ -215,14 +215,12 @@ private static boolean hasNoPrefix(@Nullable final AnnotationValue annotationVal
...
+            // Optimization, though the else case would also work.
             return unitsAnnotation;
         } else {
-            // In the Units Checker, the only annotation value is the prefix value.  Therefore,
-            // fromName (which creates an annotation with no values) is acceptable.
-            // TODO: refine sensitivity of removal for extension units, in case extension
-            // Annotations have more than just Prefix in its values.
-            return AnnotationBuilder.fromName(
-                    elements, unitsAnnotation.getAnnotationType().toString());
+            AnnotationBuilder builder = new AnnotationBuilder(elements, unitsAnnotation);
```
wmdietl: I don't understand the motivation for this whole change. `fromName` returns a new annotation without any elements.
Why change this to a new constructor and then remove the element?

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -86,8 +86,18 @@ public AnnotationBuilder(ProcessingEnvironment env, CharSequence name) {
...
 
     public AnnotationBuilder(ProcessingEnvironment env, AnnotationMirror annotation) {
-        this.elements = env.getElementUtils();
-        this.types = env.getTypeUtils();
+        this(env.getElementUtils(), env.getTypeUtils(), annotation);
+    }
+
+    // TODO: This is a bit gross, is used by UnitsRelationsTools.
+    /** Leaves type field null, should not be used if that field is needed. */
+    public AnnotationBuilder(Elements elements, AnnotationMirror annotation) {
```
wmdietl: I don't find this new constructor appropriate. There is no indication how to choose between this constructor and the one just before, which takes a `ProcessingEnvironment` instead of a `Elements`.
It sets a field to null, although the field should be non-null. So this creates an `AnnotationBuilder` that might result in NPEs when some operations are performed on it.

checker/src/main/java/org/checkerframework/checker/units/UnitsRelationsTools.java
mernst: I am trying to remove all uses of `fromName`, which is an error-prone interface.  If called with the name of an annotation that has elements/fields, then it silently produces an invalid annotation that can create problems (incorrect results, crashes) much later.

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationBuilder.java
mernst: This is used only by UnitsRelationsTools (see comment on line 92).  I didn't see how to obtain a `ProcessingEnvironment` in UnitsRelationsTools.  If you can suggest a way to do that, then we can remove this constructor -- like you, I would prefer that.  Otherwise, this is needed by UnitsRelationsTools.

PR #2257, opened 2019-01-26
Remove unreachable code

seanmcl:
Here depth can never be negative, since it starts positive, and
if it ever reaches 0, we return immediately.


mernst:
This looks right.  Thanks!  I have pulled upstream into your branch in hopes of making the Travis-CI tests work.  When they pass, I will merge.

mernst:
After pulling in two months of commits, the Travis-CI tests pass.
Thanks for cleaning up the logic.

PR #2259, opened 2019-01-26
Remove dead copy-paste error

seanmcl:


mernst:
Thanks for noticing that.

PR #2261, opened 2019-01-27
Fix false negative in Map.get

mernst:


checker/src/main/java/org/checkerframework/checker/nullness/KeyForAnnotatedTypeFactory.java
```diff
@@ -39,8 +41,15 @@
         extends GenericAnnotatedTypeFactory<
                 KeyForValue, KeyForStore, KeyForTransfer, KeyForAnalysis> {
 
+    /** The types in the KeyFor hierarchy. */
     protected final AnnotationMirror UNKNOWNKEYFOR, KEYFOR, KEYFORBOTTOM;
 
+    /** A parameter list consisting of just Object. */
+    public final List<TypeMirror> PARAMS_OBJECT;
```
wmdietl: Can the field be protected, like the other similar fields?

checker/src/main/java/org/checkerframework/checker/nullness/KeyForAnnotatedTypeFactory.java
```diff
@@ -39,8 +41,15 @@
...
+
+    /** A parameter list consisting of [Object, Object]. */
+    public final List<TypeMirror> PARAMS_OBJECT_OBJECT;
```
wmdietl: Can the field be protected, like the other similar fields?

checker/src/main/java/org/checkerframework/checker/nullness/KeyForAnnotatedTypeFactory.java
```diff
@@ -218,19 +234,71 @@ public boolean isSubtype(AnnotationMirror subAnno, AnnotationMirror superAnno) {
...
+        List<Node> args = n.getArguments();
+        if (args.size() != paramTypes.size()) {
+            return false;
+        }
+        for (int i = 0; i < args.size(); i++) {
+            Node arg = args.get(i);
+            TypeMirror erasedArgType = types.erasure(n.getArguments().get(i).getType());
+            TypeMirror paramType = paramTypes.get(i);
+            // This test isn't right, because there could be a more specific overload that is
+            // applied.  Exact equality wouldn't be right either.
```
wmdietl: This comment is a bit confusing.
This test isn't right, but it's right enough? Or are there things that need improvement?
Can you clarify why exact equality isn't right?
At the moment the passed in parameter types are always `j.l.Object`, so the tests are trivially false. Would just passing the number of arguments be enough to disambiguate overloading?

checker/src/main/java/org/checkerframework/checker/nullness/NullnessTransfer.java
```diff
@@ -55,6 +61,9 @@
     /** Annotations of the non-null type system. */
     protected final AnnotationMirror NONNULL, NULLABLE;
 
+    /** Java's Map interface. */
+    AnnotatedDeclaredType MAP_TYPE;
```
wmdietl: Can this be `protected final`?

checker/src/main/java/org/checkerframework/checker/nullness/NullnessTransfer.java
```diff
@@ -55,6 +61,9 @@
     /** Annotations of the non-null type system. */
     protected final AnnotationMirror NONNULL, NULLABLE;
 
+    /** Java's Map interface. */
```
wmdietl: Maybe note that the qualifiers in this type don't matter - it is not used as a fully-annotated ADT, just for asSuper.

mernst:
@wmdietl Thanks for the helpful suggestions.  I have improved the confusing comment and made the changes to visibility modifiers.  I will merge it when the tests pass, if that's OK with you.

mernst:
@wmdietl Can you take a quick look to double-check that I have done this correctly?

Issue #2247, opened 2019-01-23
Nullness Checker: Any instance type of a class must be a subtype of its declared type


t-rasmud:
If the declaration type of a class is `@A`, then every instance type of this class must be a subtype of `@A` and the checker-framework must report an error otherwise. Currently, the checker-framework does not perform this check.

Consider this example:

```
import org.checkerframework.checker.nullness.qual.NonNull;
import org.checkerframework.checker.nullness.qual.Nullable;

@NonNull class DeclaredClass{}

class ValidUseType {
    void test(){
        @Nullable DeclaredClass object;
    }
}
```
Since `DeclaredClass` has the type `@NonNull`, it should be illegal to write `@Nullable DeclaredClass object;`

Running the command `javac -processor nullness ValidUseType.java ` does not issue any warning.

wmdietl:
Related discussion: #1751
The current implementation uses class declaration annotations only as a default for uses of that type, not as a restriction on all possible uses.
For example:

```java
import org.checkerframework.checker.nullness.qual.NonNull;
import org.checkerframework.checker.nullness.qual.Nullable;

@NonNull
class DeclaredClass{}

class validUseType {
    void test(){
         DeclaredClass object = null;
    }
}
```

Gives an incompatible assignment error.

As outlined in the design document, I agree that changing class declaration annotations to be upper bounds would be better.

t-rasmud:
Fixed in 937286d2e404db5b49177c2ba6dd1d511aedb711

wmdietl:
Didn't that commit only add the possibility to declare the default for type declarations?
The issue with e.g. the Nullness Checker doesn't seem fixed.

mernst:
Werner is right: the original issue is not fixed.  I have added a test case at `checker/tests/nullness/Issue2247.java`.

smillst:
This is already implemented in the framework.  The Nullness Checker changes this behavior by overriding isValidUse in NullnessVisitor.

Issue #1751, opened 2018-01-06
local variable type refinement limitation


cushon:
The following example is accepted by the nullness checker if the local variable declaration `Optional<String> y` is explicitly annotated as `@Nullable`. Should the checker be able to infer the local variable type?

```java
import java.util.Optional;
import org.checkerframework.checker.nullness.qual.Nullable;

abstract class T {
  interface I<T> {}

  @Nullable abstract <T> T g(I<T> i);

  void f(I<Optional<String>> x) {
    Optional<String> y = g(x);
    if (y != null) {
      System.err.println(y);
    }
  }
}
```

```
./checker-framework-2.3.1/checker/bin/javac -processor Nullness -cp checker-framework-2.3.1/checker/dist/checker-qual.jar T.java
T.java:10: error: [assignment.type.incompatible] incompatible types in assignment.
    Optional<String> y = g(x);
                          ^
  found   : @Initialized @Nullable Optional<@Initialized @NonNull String>
  required: @UnknownInitialization @NonNull Optional<@Initialized @NonNull String>
1 error
```

wmdietl:
Class `java.util.Optional` is by default never allowed to be `@Nullable`, see the `@NonNull` on the `Optional` class declaration https://github.com/typetools/checker-framework/blob/master/checker/jdk/nullness/src/java/util/Optional.java#L63 .
This enforces "Rule #1: Never, ever, use null for an Optional variable or return value." from https://stuartmarks.files.wordpress.com/2016/09/optionalmotherofallbikesheds3.pdf and seems generally accepted practice.
(Explicitly making the local variable `@Nullable` should also be forbidden, as it currently allows to violate that constraint.)

In your example, if you replace `Optional` with `List` throughout, local variable type inference works as expected.

So I would say this is intended behavior and that explicitly writing `@Nullable Optional<...>` should also be forbidden.

cushon:
Thanks for the explanation!

> Never, ever, use null for an Optional variable or return value

I'm aware of that school of thought, but it is not universally accepted. There's a comment about `null` and guava's `Optional` here that still applies: https://stackoverflow.com/questions/9561295/whats-the-point-of-guavas-optional-class#comment12143090_9561334. In our case I think we'd prefer that `Optional` didn't have the `@NonNull` declaration annotation.

But if it does, I agree that disallowing the explicit annotation would have been less confusing.

[...]

PR #2266, opened 2019-01-30
Default for type declarations

t-rasmud:
This PR adds a new `TypeUseLocation (TYPE_DECLARATION)` for type declarations.
When a type qualifier is annotated with `@DefaultFor(TYPE_DECLARATION)`, that type qualifier is added by default to class and interface declarations. This will be useful for checkers that want their class and interface declarations to have a default that's not the same as the usual default.


t-rasmud:
@mernst A test case for this PR would require changing the current default for class type in one of the checkers. This feature is most relevant to the determinism checker which is not merged into the master branch.

framework/src/main/java/org/checkerframework/framework/qual/TypeUseLocation.java
```diff
@@ -11,6 +11,9 @@
  */
 public enum TypeUseLocation {
 
+    /** Apply default annotations to class and interface declarations. */
```
wmdietl: Why not also to enums and annotation types?
Arguably annotation types aren't ever going to be annotated and contain no code.
However, enums contain code and can be annotated.

framework/src/main/java/org/checkerframework/framework/util/defaults/QualifierDefaults.java
```diff
@@ -818,6 +818,16 @@ public Void scan(AnnotatedTypeMirror t, AnnotationMirror qual) {
                 }
 
                 switch (location) {
+                    case TYPE_DECLARATION:
+                        {
+                            if (scope != null
+                                    && (scope.getKind() == ElementKind.CLASS
+                                            || scope.getKind() == ElementKind.INTERFACE)
```
wmdietl: Add ENUM?

t-rasmud:
> Could you address #2247 and #1751 by changing the Nullness Checker to use this feature and then call super?
> Then you fix those bugs and test this new feature.

@wmdietl 
I tried this by making `@Nullable` as the DefaultFor(TYPE_DECLARATION). But there are test cases failing because the initialization checker has `@Initialized` as the default qualifier and I see that the default use type is `@UnderInitialization`. Would you have ideas on how this should be fixed?

t-rasmud:
Thanks @mernst  and @wmdietl . I created a toy checker which has `@ToyBottom` as the default qualifier in hierarchy and `@ToyTop` as the `DefaultFor(TYPE_DECLARATION)`. So for a class declared as `class X`, writing `@TopType X` should not display any warning. The test case verifies that this is indeed the case.

checker/src/main/java/org/checkerframework/checker/toy/ToyChecker.java
```diff
@@ -0,0 +1,6 @@
+package org.checkerframework.checker.toy;
```
mernst: Defining this in the main Checker Framework means that it will appear in distributions and be visible to users.
I suggest that this be defined under checker/tests.

checker/src/main/java/org/checkerframework/checker/toy/ToyChecker.java
```diff
@@ -0,0 +1,6 @@
...
+
+import org.checkerframework.common.basetype.BaseTypeChecker;
+
+/** A type-checker plug-in for the Toy type system. */
+public class ToyChecker extends BaseTypeChecker {}
```
mernst: Rather than calling this "Toy", please give it a name that is more evocative of its purpose, which is defaults to type declarations.

mernst:
#2247 and #1751 will be good to fix, but they are a distinct issue.  Although related to this issue, I don't think that this pull request has to be broadened to include them.

checker/tests/toy/TestDefaultForTypeDecl.java
```diff
@@ -0,0 +1,5 @@
+import org.checkerframework.checker.toy.qual.*;
+
+public class TestDefaultForTypeDecl {
+    void test(@ToyTop TestDefaultForTypeDecl arg) {}
```
wmdietl: Could you expand the test a bit to show that normal uses of the type work, e.g. unannotated uses and with other qualifiers.

checker/src/main/java/org/checkerframework/checker/toy/ToyChecker.java
wmdietl: https://github.com/typetools/checker-framework/blob/master/framework/src/test/java/lubglb/LubGlbChecker.java is a similar test checker and something relative to it would work well.

t-rasmud:
Thanks for the review @wmdietl and @mernst. I have addressed all the comments in my latest commit.

wmdietl:
Thanks for adding this feature!
It would have been nicer to squash the PR instead of adding 31 commits to the history.
Could we revert the merge and redo with a squash?

mernst:
I agree with Werner:  per https://homes.cs.washington.edu/~mernst/advice/github-pull-request.html#merging, it is always better to squash-and-merge rather than to just merge.  (I guess I should have been clearer about that.)

At this point, it's too late to revert because the merge is in the wild, so the history will always contain them.  Let's be careful in the future.

PR #2271, opened 2019-02-02
Eliminate method getElementValueClass

mernst:
It is preferable to use Names rather than classes.  They are guaranteed to come from the program being compiled, which might differ from the classes on the classpath.  (As a less important side benefit, names are marginally more efficient.)

The existence of methods that return classes is an attractive nuisance:  it tempts programmers to use those APIs.  Likewise, the existence of code that uses those APIs encourages programmers to mimic it.  This pull request reduces both.

This pull request:
 * reduces use of AnnotationBuilder.fromClass.
 * eliminates methods `AnnotationUtils.getElementValueClass` and `AnnotationUtils.getElementValueAnnotationClass`.

This makes the interface smaller and less error-prone.


checker/src/main/java/org/checkerframework/checker/units/UnitsRelations.java
```diff
@@ -34,4 +37,51 @@
      *     is known
      */
     @Nullable AnnotationMirror division(AnnotatedTypeMirror lht, AnnotatedTypeMirror rht);
+
+    /**
```
wmdietl: What is the motivation for moving these methods here?
Similarly-named methods remain in `UnitsRelationsTools`.
It would seem cleaner to me to keep all these related methods in `UnitsRelationsTools` and not add them as static methods to this interface.

mernst: The other version of the methods are the desirable Name ones.
These are the undesirable Class ones, which should not be used by most clients.
These are needed as a convenience for UnitsRelations, but are here so they are not noticeable to other clients.

wmdietl: But the UnitsRelations interface is the more visible place, because one would have to implement that interface for custom units.
How about moving the methods back and adding javadoc when to use them?

mernst: When implementing custom units, using the class versions is acceptable.  Otherwise, it isn't.  So this moves the class version to a place accessible to the people who need it (and no one else).
It keeps them out of UnitsRelationsTools, which is a class that more clients such as the type factory use.  Keeping them less visible to the type factory reduces the likelihood that they will be used incorrectly, compared to Javadoc.
Also note that one of the classes contains 3 methods and the other contains 2 methods, a difference that would also have to be documented if they are all put together.

But, if you feel strongly about this, I can move it back.

PR #2275, opened 2019-02-04
Remove undocumented behavior for null arguments

mernst:
Clients might depend on this behavior.
It's better for clients to not pass null, and none seem to do so now.

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/units/UnitsRelationsTools.java
+++ b/checker/src/main/java/org/checkerframework/checker/units/UnitsRelationsTools.java
@@ -32,10 +32,6 @@ public class UnitsRelationsTools {
      */
     public static @Nullable AnnotationMirror buildAnnoMirrorWithDefaultPrefix(
             final ProcessingEnvironment env, final Class<? extends Annotation> annoClass) {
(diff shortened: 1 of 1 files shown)


checker/src/main/java/org/checkerframework/checker/units/UnitsRelationsTools.java
```diff
@@ -141,10 +125,6 @@
      * @return true if it has no prefix, false otherwise
      */
     public static boolean hasNoPrefix(@Nullable final AnnotatedTypeMirror annoType) {
```
wmdietl: Here, and for several of the changes below, the parameter type is explicitly `@Nullable`, so the null check made sense.
I don't know why the parameters are `@Nullable`, but either the annotation should also be removed or the checks should be added again.

mernst: You are right.  Thanks for catching that!  I hadn't noticed the type annotations.  I fixed them in 90dba3bbb304115fe5cdceef952f024a8ee4c03c.
I do think the contract is cleaner without having to specify behavior for null, and it doesn't seem to be a problem for clients.

PR #2280, opened 2019-02-06
Improve tests for which method is being invoked

mernst:


mernst:
Sounds good.  Thank you for the improvements.

PR #2284, opened 2019-02-07
Make a variable public because it is used by checker-framework-inference

mernst:
Merge with https://github.com/typetools/checker-framework-inference/pull/97


framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -342,7 +342,7 @@
     public final boolean ignoreUninferredTypeArguments;
 
     /** The Object.getClass method. */
-    protected final ExecutableElement objectGetClass;
+    public final ExecutableElement objectGetClass;
```
wmdietl: Is the only use here https://github.com/typetools/checker-framework-inference/pull/97/files ?
That class is a sub-class of this class, so `protected` should be enough, shouldn't it?

PR #2281, opened 2019-02-06
Generate OSGi manifest for checker-qual

travishaagen:
Generate OSGi manifest for `checker-qual`.  Not having this is a big problem for anyone using Guava in an OSGi project.

re: https://github.com/typetools/checker-framework/issues/2065

Shout out to @jlemay who helped figure this out! 

Here's what the `MANIFEST.MF` looks like:

```
Manifest-Version: 1.0
Automatic-Module-Name: org.checkerframework.checker.qual
Bnd-LastModified: 1549483933073
Bundle-ManifestVersion: 2
Bundle-Name: checker-qual
Bundle-SymbolicName: checker-qual
Bundle-Version: 2.5.8
Created-By: 1.8.0_152 (Oracle Corporation)
Export-Package: org.checkerframework.checker.compilermsgs.qual;uses:="
 org.checkerframework.framework.qual";version="2.5.8",org.checkerframe
 work.checker.fenum.qual;uses:="org.checkerframework.framework.qual";v
 ersion="2.5.8",org.checkerframework.checker.formatter;uses:="org.chec
 kerframework.checker.formatter.qual";version="2.5.8",org.checkerframe
 work.checker.formatter.qual;uses:="org.checkerframework.dataflow.qual
 ,org.checkerframework.framework.qual";version="2.5.8",org.checkerfram
 ework.checker.guieffect.qual;uses:="org.checkerframework.framework.qu
 al";version="2.5.8",org.checkerframework.checker.i18n.qual;uses:="org
 .checkerframework.framework.qual";version="2.5.8",org.checkerframewor
 k.checker.i18nformatter;uses:="org.checkerframework.checker.i18nforma
 tter.qual";version="2.5.8",org.checkerframework.checker.i18nformatter
 .qual;uses:="org.checkerframework.framework.qual";version="2.5.8",org
 .checkerframework.checker.index.qual;uses:="org.checkerframework.fram
 ework.qual";version="2.5.8",org.checkerframework.checker.initializati
 on.qual;uses:="org.checkerframework.framework.qual";version="2.5.8",o
 rg.checkerframework.checker.interning.qual;uses:="org.checkerframewor
 k.framework.qual";version="2.5.8",org.checkerframework.checker.lock.q
 ual;uses:="org.checkerframework.framework.qual";version="2.5.8",org.c
 heckerframework.checker.nullness;uses:="org.checkerframework.checker.
 nullness.qual";version="2.5.8",org.checkerframework.checker.nullness.
[...]

PR #2287, opened 2019-02-07
Add OSGI bundle information to all jars.

wmdietl:
Following #2281 to do the same for all jars.
Doesn't seem to work yet. `checker.jar` contains `Export-Package: *` instead of the expanded list.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,9 +1,20 @@
 import de.undercouch.gradle.tasks.download.Download
 import org.gradle.internal.jvm.Jvm
 
+buildscript {
+    repositories {
+        maven {
+            url "https://plugins.gradle.org/m2/"
+        }
+    }
+    dependencies {
+        classpath "biz.aQute.bnd:biz.aQute.bnd.gradle:4.1.0"
+    }
+}
+
 plugins {
     id 'com.github.johnrengelman.shadow' version '4.0.4'
-    id "de.undercouch.download" version "3.4.3"
+    id 'de.undercouch.download' version '3.4.3'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -435,6 +447,7 @@ subprojects {
                 attributes("Implementation-Version": "${version}")
                 attributes("Implementation-URL": "https://checkerframework.org")
                 attributes('Automatic-Module-Name': "org.checkerframework." + project.name.replaceAll('-', '.'))
+                attributes('Export-Package': '*')
```
jlemay: I think the reason this does not work is here https://plugins.gradle.org/plugin/biz.aQute.bnd

I you go to the bottom and expand the "about new plugin" section it says:

> Cannot be used in conjunction with subprojects {}, allprojects {} etc.
> It is not possible to use the familiar pattern of applying a plugin to multiple projects at once using subprojects {} etc. at the moment. There is currently no mechanism for applying a plugin to multiple projects at once. At the moment, each project that requires a plugin must declare so in the plugins {} block in its buildscript.
> 
> Future versions of Gradle will remove this restriction.
> 
> If the current limitations of the new plugin mechanism are problematic for your build, the recommended approach is to use the “traditional” mechanism for the time being.

This seems to suggest to me that each of the individual sub project grade file need to declare this plugin them selves. Kinda unfortunate you cant centralize the config like this though.

wmdietl: Thanks for looking that up. I'll abandon this for now until somebody actually requires this feature.

PR #2290, opened 2019-02-09
Make annotations in extends clause affect valid types; fixes #2243

### LINKED ISSUE #2243: Annotations in extends clause don't affect valid types
Consider the qualifier hierarchy: `@B <: @A`
If a class is declared as `class X extends @B Y { ...}`, then all instances of `X` should be instances of `@B`, thus it should be impossible to have `@A X`.

Consider

```

public class ExtendsSubtypingIssue extends @NonNull Object {
    public static void f(@Nullable ExtendsSubtypingIssue a) {
    }
}
```

This currently type-checks which is incorrect. We expect the checker-framework to report an error at the declaration`@Nullable ExtendsSubtypingIssue a`.

t-rasmud:


framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1063,6 +1064,22 @@ public void postProcessClassTree(ClassTree tree) {
             // do so is here.
             wholeProgramInference.saveResults();
         }
+
+        if (tree.getExtendsClause() == null) {
```
mernst: This test seems suspect to me.  I would think that an implicit extends clause can conflict with an explicit annotation on the class declaration.
If so, then there should be a test for that, and also logic to handle it.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1063,6 +1064,22 @@ public void postProcessClassTree(ClassTree tree) {
...
+
+        // if "@B class Y extends @A X {}", @B must be a subtype of @A
+        Set<? extends AnnotationMirror> topAnnotations = qualHierarchy.getTopAnnotations();
+        for (AnnotationMirror topAnno : topAnnotations) {
+            AnnotationMirror classType = getAnnotatedType(tree).getAnnotationInHierarchy(topAnno);
+            AnnotationMirror extendsType =
+                    getAnnotatedType(tree.getExtendsClause()).getAnnotationInHierarchy(topAnno);
+            if (!qualHierarchy.isSubtype(classType, extendsType)) {
+                @CompilerMessageKey String errorKey = "extends.clause.invalid";
+                checker.report(Result.failure(errorKey, classType, extendsType), tree);
```
wmdietl: Validation rules should be enforced in the Visitor https://github.com/typetools/checker-framework/blob/master/framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java, not in the AnnotatedTypeFactory.
The ATF would add default/implicits for this location. I would expect that similarly to #2266 we will need a new type use location to determine what should apply to this location.

checker/tests/nullness/FlowField.java
```diff
@@ -33,6 +33,7 @@ void test1() {
         protected String field = null;
     }
 
+    // :: error: (extends.clause.invalid)
     class B extends A {
```
wmdietl: Could you explain the motivation for this error? I'm confused why we want it.
Per #2247 we want class declarations to be the top type. What default applies to the extends clause?
This particular example is in a `DefaultQualifier(Nullable.class)` scope, so isn't it
`@Nullable class B extends @Nullable A`?
That would confirm to both the bound from the declaration of A (which is also `@Nullable class A`) and the use as bound for B.

mernst: I agree that this declaration seems like `@Nullable class B extends @Nullable A`.  Good catch!

#2247 doesn't say all declarations are top type (except in a comment that refers to a proposal for a design change).  We could make such a change in the future.  In the meanwhile, it will be good to make the implementation consistent with the current design.

t-rasmud: > I agree that this declaration seems like `@Nullable class B extends @Nullable A`. Good catch!

I did a little debugging and found that the nullness checker replaces the annotation on `class A` with `@NonNull`. This replacement is done in the `NullnessAnnotatedTypeFactory`:
https://github.com/typetools/checker-framework/blob/513dda9710ba76e7d6d742d08b5972fdd7f7605b/checker/src/main/java/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java#L379

mernst: That method is ugly.  Here is a pull request that eliminates it: https://github.com/typetools/checker-framework/pull/2292 .

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
t-rasmud: As we discussed yesterday, we assume that a class implicitly extends `@Top Object` and there is no need for any check in this case.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
t-rasmud: Thanks @wmdietl I moved the rules to BaseTypeVisitor.java.

t-rasmud:
@wmdietl, @mernst I have addressed the comments from your review. Can you please take a look again? Thanks!

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -106,13 +106,7 @@
...
 import org.checkerframework.framework.util.FlowExpressionParseUtil.FlowExpressionParseException;
 import org.checkerframework.framework.util.dependenttypes.DependentTypesHelper;
-import org.checkerframework.javacutil.AnnotationUtils;
-import org.checkerframework.javacutil.BugInCF;
-import org.checkerframework.javacutil.ElementUtils;
-import org.checkerframework.javacutil.Pair;
-import org.checkerframework.javacutil.PluginUtil;
-import org.checkerframework.javacutil.TreeUtils;
-import org.checkerframework.javacutil.TypesUtils;
+import org.checkerframework.javacutil.*;
```
mernst: It is our convention to use explicit import statements.  I fixed this.

framework/src/main/java/org/checkerframework/framework/source/messages.properties
```diff
@@ -1,2 +1,3 @@
 annotation.not.completed=Element %s contains annotation %s which can't be found. Make sure your classpath is set correctly.
 unneeded.suppression=this warning suppression is not used by the %s for key %s
+
```
mernst: Please avoid whitespace changes.  I fixed this.

checker/tests/nullness/JavaCopExplosion.java
```diff
@@ -3,6 +3,7 @@
 
 @org.checkerframework.framework.qual.DefaultQualifier(Nullable.class)
 class Explosion {
+    // :: error: (extends.clause.invalid)
```
wmdietl: Is this error telling us something useful? What is invalid about this declaration?
For the Nullness Checker, when do you expect this kind of error?

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -340,6 +340,29 @@ public final Void visitClass(ClassTree classTree, Void p) {
         try {
             processClassTree(classTree);
             atypeFactory.postProcessClassTree(classTree);
+
+            // If "@B class Y extends @A X {}", then enforce that @B must be a subtype of @A.
+            // classTree.getExtendsClause() is null when there is no explicitly-written extends
+            // clause, as in "class X {}". We assume that this is equivalent to writing
+            // "class X extends @Top Object {}", and there is no need to do any subtype checking.
+            if (classTree.getExtendsClause() != null) {
```
wmdietl: I think this logic should go into org.checkerframework.common.basetype.BaseTypeVisitor.processClassTree(ClassTree) which already validates extends and implements clauses.
This code should also validate implements clauses (without duplicating the check).

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -340,6 +340,29 @@ public final Void visitClass(ClassTree classTree, Void p) {
...
+                    AnnotationMirror classType =
+                            atypeFactory
+                                    .getAnnotatedType(classTree)
+                                    .getAnnotationInHierarchy(topAnno);
+                    AnnotationMirror extendsType =
+                            atypeFactory
+                                    .getAnnotatedType(classTree.getExtendsClause())
+                                    .getAnnotationInHierarchy(topAnno);
+                    if (!atypeFactory.getQualifierHierarchy().isSubtype(classType, extendsType)) {
+                        checker.report(
```
wmdietl: The error name `extends.clause.invalid` is rather confusing, as the problem is with the annotation on the class declaration, not the extends/implements clause.

checker/tests/nullness/JavaCopExplosion.java
```diff
@@ -3,6 +3,7 @@
 
 @org.checkerframework.framework.qual.DefaultQualifier(Nullable.class)
 class Explosion {
+    // :: error: (declaration.inconsistent.with.extends.clause)
```
mernst: This error seems to be about 
```
class @Nullable ExplosiveException extends @Nullable Exception {}
```
What is wrong with that declaration?

mernst: The error message is: ```Class @Nullable cannot extend @NonNull```.
This suggests that the default is not getting applied to extends clauses.

framework/src/main/java/org/checkerframework/framework/util/defaults/QualifierDefaults.java
```diff
@@ -462,6 +462,20 @@ private void applyDefaults(Tree tree, AnnotatedTypeMirror type) {
...
+                    applyDefaultsToExtendsAndImplicits(elt, type);
+                }
+                if (!((ClassTree) tree).getImplementsClause().isEmpty()) {
+                    List<? extends Tree> implicitClauses = ((ClassTree) tree).getImplementsClause();
+                    for (Tree implicitClause : implicitClauses) {
+                        elt = TreeUtils.elementFromTree(implicitClause);
+                        applyDefaultsToExtendsAndImplicits(elt, type);
+                    }
+                }
+                elt = nearestEnclosingExceptLocal(tree);
```
mernst: This unconditionally resets `elt`, which may have been set just above.  I suspect the uses just above should be fresh local variables, to avoid confusion about what they are for and their (non-)relationship to the point of the switch statement, which is to set the `elt` local variable.
To clarify the logic of the `applyDefaults` method, how about abstracting its first half (the switch statement) into a helper method?

framework/src/main/java/org/checkerframework/framework/util/defaults/QualifierDefaults.java
```diff
@@ -484,6 +498,26 @@ private void applyDefaults(Tree tree, AnnotatedTypeMirror type) {
         applyToTypeVar = false;
     }
 
+    /**
+     * Applies default annotations to {@code type}.
```
mernst: This documentation is rather vague and general, but the method name is more evocative.  Which one should be changed?

PR #2294, opened 2019-02-12
Add framework artifacts.

### LINKED ISSUE #2291: Release framework artifact
An artifact for `framework` is missing, which is the dependency one would add when developing a new type system.
At the moment one would have to use `checker`, which contains all distributed type systems and therefore many unnecessary classes.

The Maven artifact poms are here:
https://github.com/typetools/checker-framework/tree/master/maven-artifacts/poms

A `framework-test` artifact exists and can probably be used as basis for a `framework` artifact.

wmdietl:
Fixes #2291.


maven-artifacts/poms/frameworkPom.xml
```diff
@@ -0,0 +1,29 @@
...
+        <repository>
+            <id>checker-releases</id>
+            <url>https://checkerframework.org/dev/m2-repo</url>
+        </repository>
+    </distributionManagement>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.checkerframework</groupId>
+            <artifactId>dataflow</artifactId>
```
wmdietl: This is blocked on having `stubparser` and `annotation-tools` Maven dependencies.

PR #2295, opened 2019-02-12
Added rule: type on constructor declaration must be a supertype of super() invocation; Fixes #2264

### LINKED ISSUE #2264: The type of a constructor declaration must be a supertype of the type of its super() constructor.
Consider the tainting checker where `@Untainted <: @Tainted`. 

```

public class SubtypeConstructorSuper extends SuperClass{
    @Untainted SubtypeConstructorSuper(){
        super();
    }
}

class SuperClass{
    @Tainted SuperClass(){}
}
```

In the example above, `SubtypeConstructorSuper()` is annotated as `@Untainted`, but it calls `super()` which is `@Tainted`. This should be illegal. The checker framework should enforce that the type of `SubtypeConstructorSuper()` be a supertype of the type of ` SuperClass()`.

Currently, the checker framework places no such restriction and running the command `javac -processor tainting SubtypeConstructorSuper.java` type checks without any warning.

t-rasmud:


checker/tests/interning/Arrays.java
```diff
@@ -33,6 +33,7 @@ public static void test(@Interned Integer i, @Interned String s) {
         @Interned String @Interned [][] multi2 = new @Interned String @Interned [2][];
     }
 
+    // :: error: (super.invocation.invalid)
```
wmdietl: Does this error make sense for the Interning Checker?
Won't this always be created for every `@Interned class`, because `java.lang.Object`'s constructor is not `@Interned`?
Would an alternative be to mark the j.l.Object constructor `@Interned`?

A discussion of `@Interned` on a class declaration and constructor would be very helpful in the manual to understand the intended semantics for this system.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1029,6 +1029,42 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
...
+                Set<? extends AnnotationMirror> topAnnoations =
+                        atypeFactory.getQualifierHierarchy().getTopAnnotations();
+                for (AnnotationMirror topAnno : topAnnoations) {
+                    AnnotationMirror superTypeMirror = superType.getAnnotationInHierarchy(topAnno);
+                    AnnotationMirror constructorTypeMirror =
+                            constructorType.getReturnType().getAnnotationInHierarchy(topAnno);
+                    // Checking AnnotaionUtils.AreSameByName for the following case:
+                    // Example from tests/nullness/DaikonTests.java: "class Bug2 extends Bug2Super"
+                    // Here, constructorTypeMirror is
+                    // @UnderInitialization(DaikonTests.Bug2Super.class)
```
wmdietl: I do not understand from this why this uses `areSameByName`. Instead of the giving the specific test case, could you try to explain the general logic?

Couldn't there be an example where the `@UnderInitialization` annotation is used incorrectly and an error should be raised, even though they have the same annotation name?

framework/src/main/java/org/checkerframework/common/basetype/messages.properties
```diff
@@ -39,6 +39,7 @@ contracts.precondition.expression.parameter.name=The precondition %s on the decl
 contracts.postcondition.expression.parameter.name=The postcondition %s on the declaration of method '%s' refers to formal parameter '%s' by name. Please use "#%s" instead of "%s".
 contracts.conditional.postcondition.expression.parameter.name=The conditional postcondition %s on the declaration of method '%s' refers to formal parameter '%s' by name. Please use "#%s" instead of "%s".
 
+super.invocation.invalid=Constructor of type %s cannot call super() of type %s.
```
wmdietl: Couldn't the `super()` call have arguments?
The type you give as third argument is the constructor return type. Maybe instead use the full signature `constructorType`?

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1029,6 +1029,42 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
...
         checkPreconditions(node, contractsUtils.getPreconditions(invokedMethodElement));
 
+        // Checks that the following rule is satisfied:
+        // The type on a constructor declaration must be a supertype of
+        // the return type of "super()" invocation within that constructor.
+        if (TreeUtils.isSuperCall(node)) {
+            TreePath path = atypeFactory.getPath(node);
+            MethodTree enclosingMethod = TreeUtils.enclosingMethod(path);
+            if (TreeUtils.isConstructor(enclosingMethod)) {
+                AnnotatedTypeMirror superType = atypeFactory.getAnnotatedType(node);
```
wmdietl: I think we should extract this logic into a protected `checkSuperConstructorCall` helper method, similar to the other `checkXXX` methods that are called from here.
This will make type-system-dependent changes easier, makes this method shorter, and gives you more space to document the logic.

checker/tests/interning/Arrays.java
mernst: Yes, this would be a problem with every `@Interned class`:  human investigation would be needed.
I was thinking that was acceptable, because using an object as interned needs consideration from a person.

I hadn't thought of the alternative of marking the `java.lang.Object` constructor `@Interned`.  That would certainly eliminate the warnings, with the pros and cons attending to that.  I think I like that suggestion.

framework/src/main/java/org/checkerframework/common/basetype/messages.properties
mernst: Although people often write `()` just to mean "it's a method", I agree with making overloads explicit.  Thanks for the suggestion.

PR #2296, opened 2019-02-12
Share the TreePathCacher between (Sub-)Checkers and AnnotatedTypeFactories.

wmdietl:


framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -335,6 +335,12 @@
     /** Mapping from an Element to the source Tree of the declaration. */
     private final Map<Element, Tree> elementToTreeCache;
 
+    /** Mapping from a Tree to its TreePath. Shared between all instances. */
+    private final TreePathCacher treePathCache;
+
+    /** Mapping from CFG generated trees to their enclosing elements. */
+    private final Map<Tree, Element> pathHack;
```
mernst: Can we choose a better variable name than `pathHack`?

wmdietl:
@smillst Pulling in master fixed the Travis checks. Is this ok with you?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
smillst: What about `generatedTreeToEnclosing`?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -3043,6 +3065,9 @@ public final TreePath getPath(Tree node) {
      *
      * <p>TODO: Find a better way to store information about enclosing Trees.
      *
+     * <p>TODO: Is this actually guaranteed to be a method? What about CFG generated trees in field
```
smillst: `CFCFGTranslationPhaseOne#handleArtificialTree` adds both methods and classes as elements in the `pathHack`, so this method should be named something like `getEnclosingElementForGeneratedTree`.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
wmdietl: I had discussed this renaming with Mike when he made the comment.
This PR just moves that field to a more consistent location.
I'll start a separate PR that renames the field, after this one is merged to avoid conflicts.
How about `generatedTreeToEnclosingElementMap` to make it completely documented?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
wmdietl: Thanks for confirming that my suspicion that this stores more than just methods!
Let me remove this TODO and then address this in the separate PR that renames the `pathHack` field.

smillst:
Sounds good!

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
smillst: >How about generatedTreeToEnclosingElementMap to make it completely documented?

That sounds good.  Thanks!

PR #2297, opened 2019-02-13
Reduce duplication in BaseTypeVisitor constructors

mernst:
Subsequent pull requests add more fields, making the duplication in constructors more annoying.


wmdietl:
I prepared an alternative in #2299 that doesn't require removing the `final` from the fields. Please take a look.

mernst:
That is much better, thanks!

PR #2301, opened 2019-02-13
Defaults the type on a constructor declaration to be the same as the type of the enclosing class declaration; Fixes #2186

### LINKED ISSUE #2186: Constructor annotated with any type other than the default qualifier not allowed
Annotating a class with a non-default type qualifier results in an unexpected error on its constructor.
Example:
```
import org.checkerframework.checker.propkey.qual.*;
@PropertyKey class ConstructorIssue {
    ConstructorIssue() {}
}
```
The above code should type check without errors and the type of the constructor should be `@PropertyKey`. But, running the command: `javac -processor propertykey ConstructorIssue.java ` gives the following error:

```
ConstructorIssue.java:4: error: [type.invalid.annotations.on.use] invalid type: annotations [@UnknownPropertyKey] conflict with declaration of type ConstructorIssue
    ConstructorIssue() {}
    ^
1 error
```

t-rasmud:


framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1626,6 +1626,40 @@ public Void visitExecutable(AnnotatedExecutableType type, AnnotatedTypeFactory p
             // Also skip constructor return types (which somewhat act like
             // the receiver).
             MethodSymbol methodElt = (MethodSymbol) type.getElement();
+
+            // If a constructor declaration is not explicitly annotated,
+            // annotate it with the type on the class declaration.
+            Element elem = type.getElement();
+            if (elem.getKind() == ElementKind.CONSTRUCTOR) {
+                AnnotatedDeclaredType returnType = (AnnotatedDeclaredType) type.getReturnType();
+                if (returnType.getAnnotations().isEmpty()) {
```
mernst: What is the purpose of this check?  Is it just an optimization to avoid a later check, or is it necessary for correctness?
If defaults are already applied, might this test return those defaults and thus incorrectly avoid executing the body?  That is, I could imagine that if defaults are already applied, `getAnnotations()` might be non-empty even though the user did not write any annotations explicitly.  A word of explanation would be useful.

docs/manual/advanced-features.tex
```diff
@@ -285,6 +285,23 @@
 Section~\ref{climb-to-top}.
 
 
+\subsectionAndLabel{Default qualifier on constructor declarations}{constructor-defaults}
+
+If a constructor declaration is unannotated, it defaults to the same type as that of
+its enclosing class (rather than the default qualifier in hierarchy).
+For example: the Tainting type checker (Chapter~\ref{tainting-checker}) has \code{@Tainted}
```
wmdietl: For consistency with that chapter, change `Tainting type checker` to `Tainting Checker`.

checker/tests/lock/LockEffectAnnotations.java
```diff
@@ -141,6 +141,7 @@ void pureMethodWithSynchronizedBlock() {
 
     @GuardedByUnknown class MyClass2 {}
 
+    // :: error: (expression.unparsable.type.invalid)
```
wmdietl: I do not understand why this error appears. Can you explain?

checker/tests/tainting/AnonymousProblem.java
```diff
@@ -0,0 +1,5 @@
+import java.nio.file.SimpleFileVisitor;
+
+public class AnonymousProblem {
```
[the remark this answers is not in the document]
wmdietl: I later saw the implementation comment. Let me look what checker-framework-inference does.

checker/tests/lock/LockEffectAnnotations.java
mernst: There is no variable in scope named "lock".
It was a bug before that the Lock Checker did not issue this warning.
Like you, I was surprised to see this bug corrected when I reviewed the code.  But it's an improvement, so I'll take it.

framework/tests/polyall/Inheritance.java
```diff
@@ -1,7 +1,5 @@
 import polyall.quals.*;
 
-// This is a bug
```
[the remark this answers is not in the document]
mernst: It's a mystery.  But it is a good change...

checker/tests/lock/LockEffectAnnotations.java
wmdietl: The code didn't magically fix a bug. Class declarations don't validate dependent type annotations. I made a proper fix in #2309 and it actually revealed another incorrect test.
This PR only happened to raise an error here because the annotation is copied from the class declaration to the constructor return and constructor return types are validated.
It is much easier to debug these things in the scope of the changes of a PR. Just accepting miraculous improvements leads to much more pain later on.

framework/tests/polyall/Inheritance.java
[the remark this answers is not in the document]
wmdietl: This actually wasn't a bug initially. The error is about the default constructor, which had `@H1Top @H2Top` as return type by default, which violates the `@H1S1` annotation on the class declaration.
This PR copies the `@H1S1` from the class declaration to the constructor return type, so it is a good thing that the error is going away.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
t-rasmud: I have added more explanation in my latest commit.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1626,6 +1626,36 @@ public Void visitExecutable(AnnotatedExecutableType type, AnnotatedTypeFactory p
             // Also skip constructor return types (which somewhat act like
```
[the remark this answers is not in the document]
t-rasmud: Updated.

checker/tests/tainting/AnonymousProblem.java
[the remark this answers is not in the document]
t-rasmud: Thanks @wmdietl . It would be great if you can take a look at what checker-framework-inference is doing.

docs/manual/advanced-features.tex
```diff
@@ -285,6 +285,23 @@
 Section~\ref{climb-to-top}.
 
 
+\subsectionAndLabel{Default qualifier on constructor declarations}{constructor-defaults}
```
mernst: There is already text in the introduction about annotations on constructors (https://checkerframework.org/manual/#annotations-on-constructor-invocations).   I think it would be better to consolidate the information there rather than adding a new section (especially since there are no cross-references between the two sections).

Note that other pull requests have also changed that part of the manual.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1680,12 +1680,42 @@ public Void visitExecutable(AnnotatedExecutableType type, AnnotatedTypeFactory p
...
                 scan(type.getReturnType(), p);
+            } else if (methodElt.isConstructor()) {
+                // If a constructor declaration is not explicitly annotated,
+                // annotate it with the type on the class declaration.
+                AnnotatedDeclaredType returnType = (AnnotatedDeclaredType) type.getReturnType();
+                // At this point defaults have been applied to class declarations but not
+                // to constructor return types. The check "returnType.getAnnotations().isEmpty()"
+                // returns true for constructors whose return types haven't been explicitly
+                // annotated.
+                // NOTE: change to getExplicitAnnotations() when http://tinyurl.com/cfissue/2324 is
```
mernst: Could you abstract this out into a method?  Then the call can be changed to `getExplicitAnnotations()` in the future, and the hack will be better-encapsulated.

PR #2308, opened 2019-02-15
Test case that shows Issue #2263 is already fixed; fixes #2263

### LINKED ISSUE #2263: Receiver type checks for `super` invocations
Take the following simple example:

```java
import org.checkerframework.checker.initialization.qual.UnderInitialization;

class Super {
    void foo(@UnderInitialization(Super.class) Super this) {}
}

class Sub extends Super {
    @Override
    void foo(@UnderInitialization(Object.class) Sub this) {
        super.foo();
    }
}
```

This incorrectly passes the Nullness Checker.
The invocation `super.foo()` has an invalid receiver type and an error should be raised.

t-rasmud:


wmdietl:
@t-rasmud Thanks for confirming this was fixed!
I did try right before I filed the bug and didn't get an error. Do you happen to know which change since then fixed this?

checker/tests/initialization/fbc/ReceiverSuperInvocation.java
```diff
@@ -0,0 +1,16 @@
+// Test case for issue 2263
+// https://github.com/typetools/checker-framework/issues/2263
+
+import org.checkerframework.checker.initialization.qual.UnderInitialization;
+
+class ReceiverSuperInvocation {
+    void foo(@UnderInitialization(ReceiverSuperInvocation.class) ReceiverSuperInvocation this) {}
+}
+
+class Sub extends ReceiverSuperInvocation {
```
wmdietl: One slight nit I just notice: `Sub` is a bad name for this class as it might easily cause a clash with another test case that also uses `Sub`.
In general we should either keep to one top-level class or use unique names. So as a small cleanup we should maybe rename this to `ReceiverSuperInvocationSub` or some such.

PR #2309, opened 2019-02-17
Properly validate dependent type annotations on class declarations.

wmdietl:


framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -254,6 +255,34 @@ public void standardizeReturnType(MethodTree m, AnnotatedTypeMirror atm) {
...
+     * @param node the class declaration
+     * @param type the type of the class declaration
+     * @param ele the element of the class declaration
+     */
+    public void standardizeClass(ClassTree node, AnnotatedTypeMirror type, Element ele) {
+        if (!hasDependentType(type)) {
+            return;
+        }
+        TreePath path = factory.getPath(node);
+        if (path == null) {
```
mernst: When would the path for a `ClassTree` be null?

wmdietl: I don't know why this code checked for null originally.
We could separately try cleaning these null checks up.

mernst: OK, feel free to go ahead.

mernst:
Tests are failing because checker-framework-inference needs to be adapted to the new names.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
mernst: Or, just leave it as is; someone else will have to do the cleanup in the future.

Issue #1307, opened 2017-05-29
Support @Repeatable on appropriate annotations


aabdagic:
Suppose I want to annotate an interface like this:

```java
interface FooOrBar {
  @Nullable getFoo();
  @Nullable getBar();
  
  @EnsuresNonNullIf(expression = "getFoo()", result = true)
  @EnsuresNonNullIf(expression = "getBar()", result = false)
  boolean hasFoo();

  @EnsuresNonNullIf(expression = "getBar()", result = true)
  @EnsuresNonNullIf(expression = "getFoo()", result = false)
  boolean hasBar();
}
```

This isn't possible today because `@EnsuresNonNullIf` doesn't support `@Repeatable`. There are a few more use-cases that unlock if `@Repeatable` is supported. The same applies to several other annotations, e.g. `@EnsuresNonNull`, `@EnsuresQualifierIf`, etc.

wmdietl:
Hi Alvin,
thanks for the suggestion!

I agree that for the declaration annotations that take a set of expressions and a truth value it might be useful to make them `@Repeatable`, to allow specifying expressions for both true and false.

I don't see a benefit for an annotation like `@EnsuresNonNull` that only takes a set of expressions - you can simply concatenate the two sets.

One thing to consider is the implementation of these `hasXXX` methods. They would usually be something like:

```
@EnsuresNonNullIf(expression = "getFoo()", result = true)
@EnsuresNonNullIf(expression = "getBar()", result = false)
boolean hasFoo() {
  return getFoo() != null;
}
```

When we type check that method we can make sure that the first contract is established - we see that `getFoo` is not null.
However, we don't see that `getBar` will return null. So one would have to suppress a warning on this method declaration.

Another thing to consider is future extensions. Let's say a new `getBaz` method is added that is non-null if the other two are null. Now one has to add that specification to all the existing methods and make sure that all existing conditions are adapted.
Sticking with just a single post-condition would force one to write:

```
if (hasFoo()) {
  getFoo() is non-null;
} else if (hasBar()) {
  getBar() is non-null;
} else {
  raise an error!
}
```

This way every has/get pair only needs to specify their own properties and adding a new case for hasBaz/getBaz is easy to track down.

mernst:
@aabdagic In his comment above, Werner hypothesizes that this feature would be hard to implement and would be of little use in practice.  Can you give a realistic use case to support your feature request?

aidanPB:
[...]

PR #2312, opened 2019-02-17
qualifiers is already null

gokaco:


framework/src/main/java/org/checkerframework/framework/util/defaults/QualifierDefaults.java
```diff
@@ -588,9 +588,7 @@ private DefaultSet defaultsAt(final Element elt) {
         {
             AnnotationMirror ds = atypeFactory.getDeclAnnotation(elt, DefaultQualifiers.class);
             if (ds != null) {
-                if (qualifiers == null) {
-                    qualifiers = new DefaultSet();
-                }
+                qualifiers = new DefaultSet();
```
wmdietl: `qualifiers` might already be initialized from line 579 above. This change would simply ignore those results.
However, we should discuss whether it ever makes sense to allow both a single `DefaultQualifier` and a `DefaultQualifiers` at the same time. However, we should then make that explicit and rewrite the code accordingly.

gokaco: I was studying `DefaultQualifier` and `DefaultQualifiers` and I ran the given piece of code:-

import java.io.*;


@DefaultQualifier(value=Nullable.class,locations=TypeUseLocation.FIELD)
@DefaultQualifiers({@DefaultQualifier(value=NonNull.class,locations=TypeUseLocation.ALL)})
class ex
{
		Object nullableField = null;
    	`@NonNull `Object nonNullField = new Object();
    	void perl(Integer a)
    	{
    		a=null;
    	}
}
It was giving the output:-

.java:14: error: [assignment.type.incompatible] incompatible types in assignment.
    		a=null;
    		  ^
  found   : null
  required: @Initialized @NonNull Integer
1 error

But I think that it should give 2 errors.
Here if the proposed changes will be made then `@DefaultQualifiers` will work separately independent with this `@DefaultQualifier` and will give 2 errors:-
error: [assignment.type.incompatible] incompatible types in assignment.
		Object nullableField = null;
		                       ^
  found   : null
/home/kartik/Desktop/gsoc/ex.java:14: error: [assignment.type.incompatible] incompatible types in assignment.
    		a=null;
    		  ^
  found   : null
  required: @Initialized @NonNull Integer
2 errors

May In some cases this change would not be good. It totally depends on you whether you want this change or not. I made this PR to tell you about this thing. I should have created an issue rather than PR.

mernst:
It seems that the underlying problem is a duplicate of #1307.  I think that fixing that might be a better solution.  Do you want to try that instead?

gokaco:
> It seems that the underlying problem is a duplicate of #1307. I think that fixing that might be a better solution. Do you want to try that instead?

Ya, I can try. This will be a valuable feature in the software. You are right about that every declaration annotation that takes arguments should be repeatable. Will give you the updates asap.

gokaco:
Mike, I am not understading the real motive of #1307 . @DefaultQualifier is repeatable using @DefaultQualifiers and @EnsureQualifierIf is also repeatable using @EnsuresQualifiersIf. Do you want that without @DefaultQualifiers and @EnsuresQualifiersIf they should be repeatable?

mernst:
Please write discussion of #1307 at that issue, not here.

wmdietl:
Work on this is continuing in #2369 so I'm closing this.

PR #2313, opened 2019-02-19
Warn if annotation before modifier

wmdietl:
Small improvements to #2298, to be merged into that PR.


wmdietl:
Instead use https://github.com/mernst/checker-framework/pull/2

PR #2319, opened 2019-02-21
Add RxJava nullness annotations as aliases

msridhar:
See https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/annotations/Nullable.java and https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/annotations/NonNull.java.

Also fix a comment typo.


mernst:
Thanks for this enhancement.

PR #2329, opened 2019-02-26
Add a section to the Tainting Checker manual that warns about polymorphism and side effects

kelloggm:
This is a common failure mode for the Tainting Checker (and other information flow checkers). This is only an explanation of the problem, and a warning to make people aware of the unsoundness that can be introduced this way.

The solution is something like `@PolyDet("use")`, which is in the Determinism Checker. That should be referenced from this text once it's merged into master. Alternatively, we could reference the implementation in this 3rd party checker, which is based on that code: https://github.com/awslabs/data-classification-checker. I chose to leave off the reference to the solution until the original is merged into master.


mernst:
@kelloggm Please take a look at my edits

kelloggm:
Your presentation of the problem is much better than my original draft. Thanks!

We might want to mention that the way that this problem usually manifests is via stub files, so any stub files with `@PolyTainted` should be checked very carefully, looking for this issue.

kelloggm:
I think this looks good now. Merge when tests pass.

PR #2340, opened 2019-03-01
Add some JDK annotations for Signedness Checker

gokaco:


gokaco:
The changes are working fine in my machine and ./gradlew allTests also passed then why travis is giving error here?

gokaco:
InputStream in is getting null value and is not getting read.

mernst:
Master was broken for a time today. :-(
I pulled master and I believe this will resolve the problem.

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
```diff
@@ -94,6 +98,11 @@ class Long {
     public static @Unsigned long remainderUnsigned(@Unsigned long dividend, @Unsigned long divisor);
 }
 
+class Object {
+    protected @Unsigned Object clone() throws CloneNotSupportedException;
```
mernst: `@Unsigned` should not be applied to non-numeric types.

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
```diff
@@ -193,6 +202,18 @@ class Arrays {
     public static @PolySigned float[] copyOfRange(@PolySigned float[] original, int from, int to);
     public static @PolySigned double[] copyOfRange(@PolySigned double[] original, int from, int to);
     public static @PolySigned boolean[] copyOfRange(@PolySigned boolean[] original, int from, int to);
+    public static <T> @PolySigned T[] copyOfRange(@PolySigned T[] original, int from, int to);
```
mernst: Why is this necessary?

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
```diff
@@ -193,6 +202,18 @@ class Arrays {
...
+
+    public static @PolySigned byte[] copyOf(@PolySigned byte[] original, int newLength);
+    public static @PolySigned short[] copyOf(@PolySigned short[] original, int newLength);
+    public static @PolySigned int[] copyOf(@PolySigned int[] original, int newLength);
+    public static @PolySigned long[] copyOf(@PolySigned long[] original, int newLength);
+    public static @PolySigned char[] copyOf(@PolySigned char[] original, int newLength);
+    public static @PolySigned float[] copyOf(@PolySigned float[] original, int newLength);
+    public static @PolySigned double[] copyOf(@PolySigned double[] original, int newLength);
+    public static @PolySigned boolean[] copyOf(@PolySigned boolean[] original, int newLength);
+    public static <T> @PolySigned T[] copyOf(@PolySigned T[] original, int newLength);
```
mernst: Why is this necessary?

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
gokaco: Array is an object in java and if we want to clone an integer or byte array that is unsigned then due to this change we will not get error?
for ex.-
import java.util.Arrays; 
import org.checkerframework.checker.signedness.qual.*;
class ex { 

	private static @Unsigned byte[] de(@Unsigned byte[] ci){

		@Unsigned byte b[]=ci.clone();
		return b;
	}
}
This will give error if this change is not made.

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
gokaco: Yes, I think line 215 is not necessary. I just annotated all the copyOf functions. Sorry for the mistake.
And yes float double boolean cant be unsigned so their is no use of them.

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
gokaco: Yes, I think line 205 is not necessary. Sorry for the mistake.

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
mernst: I agree the change to the annotated JDK will fix this use case.
However, it will break others:
 * cloning a signed byte array will result in an unsigned byte array.  This introduces unsoundness.
 * cloning a string will result in an unsigned string.  This introduces illegal annotations.

It is not acceptable to fix one problem while introducing both unsoundness and illegal annotations.

mernst: Can you think of a different way to achieve your goals?

gokaco: Yes, If we use @PolySigned then it should work fine.

mernst: Yes, that is a better choice.  Please write a few test cases to ensure that the annotations behave as you expect.  Thanks!

gokaco: Yes, I checked for ex-
import java.lang.*; 
import org.checkerframework.checker.signedness.qual.*;
class ex2 { 

	private static @Unsigned byte[] de(byte[] ci){

		@Unsigned byte b[]=ci.clone();
		return b;
	}
}
This is giving error which would not occur earlier due to @Unsigned, but now due to @PolySigned it is giving the error. Rest all are working fine.

mernst: Thanks for checking that.  It is usually a good idea to have both positive and negative test cases (that is, both code that typechecks and code that produces a type-checking error), and to try with multiple types (including non-arrays which was the original concern).

It may be that your changes are correct, but it's good to double-check it.
Thanks!

gokaco: Sorry, but I am unable to understand "non arrays" such as? Can you give examples.
Thanks!

mernst: You could call clone on a String, or an Integer, or other types that are not arrays.

gokaco: In order to expose the clone() method to clients, we have to declare it in our subclass with a more liberal access modifier, either public or package private. However, we cannot extend java.lang.Integer because it is a final class. Also note that java.lang.Integer does not implement Cloneable then how can we use clone() method with Integer or string etc?

mernst: Yes, you have identified a serious and interesting limitation.  We are not sure how to deal with it.
I'm not sure it is a problem in your case, though.  Why do you need to change the contract of clone?  Isn't the contract on the Object class sufficient for all subclasses?

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
```diff
@@ -94,6 +98,11 @@ class Long {
     public static @Unsigned long remainderUnsigned(@Unsigned long dividend, @Unsigned long divisor);
 }
 
+class Object {
+    protected @PolySigned Object clone() throws CloneNotSupportedException;
```
mernst: You have used only one `@PolySigned`, and there is no occurrence in a binding site. 
You need to write `@PolySigned` on the receiver as well.

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
gokaco: I am not able to apply clone in Integer or String.

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
gokaco: Yes, you are right. On the other hand, if we have to use clone in Integer or String then we again have to define the function in the program. I think for the time being we should not annotate it.

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
mernst: Why not?  Can you show test code and the results?

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
mernst: > if we have to use clone in Integer or String then we again have to define the function in the program.

Can you explain why?

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
gokaco: If we simply run in java-
import java.lang.*; 
class ex2 { 

	private static Integer de(Integer ci){

		Integer b=ci.clone();
		return b;
	}
}
then it will give error.-
gsoc/ex2.java:6: error: clone() has protected access in Object
		Integer b=ci.clone();
		            ^
gsoc/ex2.java:6: error: incompatible types: Object cannot be converted to Integer
		Integer b=ci.clone();
		                  ^
2 errors

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
gokaco: Like this-
import java.lang.*;
class Integer implements Cloneable{
	public Object clone()throws CloneNotSupportedException{
	return super.clone();  
	}
}
class ex2 {

		void play(){

		try{
		Integer ci=null;
		Integer b=(Integer)ci.clone();
		}
		catch(CloneNotSupportedException c){}
	}
}

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
gokaco: 2nd error can be corectted by-
 Integer b=(Integer)ci.clone();
but how the 1st one.

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
gokaco: If we try to implement clone inn Integer than we can do by this.-
import java.lang.*;
class Integer implements Cloneable{
	public Object clone()throws CloneNotSupportedException{
	return super.clone();  
	}
}
class ex2 {

		void play(){

		try{
		Integer ci=null;
		Integer b=(Integer)ci.clone();
		}
		catch(CloneNotSupportedException c){}
	}
}

but here Integer is a class with a name "Integer". This Integer is not the same Integer which we want. I googled but I am unable to find anything on how to use clone with it.

mernst: Now I understand that the issue is that Integer does not implement Cloneable, so myInteger.clone() could never work, regardless of annotations.  Thanks for the explanation.  Removing the annotation on Object.clone() seems fine, in this case.

mernst:
Thanks for this contribution!

PR #2341, opened 2019-03-02
Changes suggested by Error Prone 2.3.3

mernst:


mernst:
Merging because only Travis CI problems are missing documentation for methods.

PR #2349, opened 2019-03-07
Fix bug in CheckerMessage.equals

mernst:


mernst:
I don't think so.  It's a bug, though, that makes you wonder why it wasn't noticed before.

Issue #599, opened 2016-02-16
Inferring array element type for nested arrays


mernst:
When the component type of an array creation expression is itself an array creation expression, the element type of the outermost array is not correctly inferred, resulting in type-checking errors.

See file checker-framework/checker/tests/nullness/ArrayCreationSubArray.java, where several array creation expressions do not typecheck:  those on the right-hand side of the assignments to o3a, o3b, and o3d.

kelloggm:
I think I've found another test case for this issue:

```java
class RaggedArrays {
    public void doStuff(int r) {
	int BCount = r * 32;

        int[] blockX1 = new int[16];
        int[] blockX2 = new int[16];
        int[] blockY = new int[BCount];

        int[] X = new int[BCount];

	int[][] myNewArray = new int[][]{ X, blockX1, blockX2, blockY };
    }
}
```

Issued errors:
```
~/jsr308/checker-framework> checker/bin-devel/javac -processor org.checkerframework.common.value.ValueChecker RaggedArrays.java 
RaggedArrays.java:11: error: [array.initializer.type.incompatible] incompatible types in array initializer.
	int[][] myNewArray = new int[][]{ X, blockX1, blockX2, blockY };
	                                  ^
  found   : @UnknownVal int @UnknownVal []
  required: @UnknownVal int @ArrayLen(16) []
RaggedArrays.java:11: error: [array.initializer.type.incompatible] incompatible types in array initializer.
	int[][] myNewArray = new int[][]{ X, blockX1, blockX2, blockY };
	                                                       ^
  found   : @UnknownVal int @UnknownVal []
  required: @UnknownVal int @ArrayLen(16) []
2 errors
```

mernst:
@kelloggm The problem you encountered is a bug in the Index Checker.  It is distinct from #599, which is a bug in the Nullness Checker.  I have opened a pull request to fix the bug you encountered:  https://github.com/typetools/checker-framework/pull/2351 .

mernst:
Fixing this problem might involve moving some code from the Index Checker (see code near #2351) into the framework, since the Index Checker infers all levels of the array type in an array creation expression.

PR #2352, opened 2019-03-09
Prefer @Initialized to @FBCBottom for array component types

mernst:
This does not fix issue #599, but it fixes part of it and is worth merging on its own.

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
+++ b/checker/src/main/java/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
@@ -476,7 +476,7 @@ public class NullnessAnnotatedTypeFactory
             AnnotatedArrayType arrayType = (AnnotatedArrayType) type;
             AnnotatedTypeMirror componentType = arrayType.getComponentType();
             if (componentType.hasEffectiveAnnotation(FBCBOTTOM)) {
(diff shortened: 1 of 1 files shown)


checker/src/main/java/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
```diff
@@ -461,6 +468,18 @@ public Void visitNewClass(NewClassTree node, AnnotatedTypeMirror type) {
...
         }
+
+        @Override
+        public Void visitNewArray(NewArrayTree node, AnnotatedTypeMirror type) {
+            // The most precise element type for `new Object[] {null}` is @FBCBottom, but
+            // the most useful element type is @Initialized (which is also accurate).
+            AnnotatedArrayType arrayType = (AnnotatedArrayType) type;
+            AnnotatedTypeMirror componentType = arrayType.getComponentType();
+            if (componentType.hasEffectiveAnnotation(FBCBOTTOM)) {
+                componentType.replaceAnnotation(INITIALIZED);
```
smillst: You should use the field "COMMITTED" declared in InitializationAnnotatedTypeFactory so that this code works for the NullnessRawness Checker, too. 
```suggestion
                componentType.replaceAnnotation(COMMITTED);
```

PR #2355, opened 2019-03-13
 Add option to require checker name in warning suppression. (Fixes 2318)

### LINKED ISSUE #2318: Option to make `nullness:` prefix for suppression strings mandatory
As a follow-up to #1701, it might be nice to have a way to make the Checker only recognize suppressions with a prefix corresponding to a type system, e.g. `nullness:`. The advantage of the prefixes is that it makes it possible to recognize suppressions that are no longer needed, as discussed in the other bug.

smillst:
Fixes #2318.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
+++ b/framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
@@ -1271,7 +1271,7 @@ public abstract class SourceChecker extends AbstractTypeProcessor
 
         } else {
             // The key for the warning/error being printed, in brackets; prefixes the error message.
-            final String =suppressing;
+            final String suppressing;
             if (this.processingEnv.getOptions().containsKey("showSuppressWarningKeys")) {
                 suppressing = String.format("[%s:%s] ", this.getSuppressWarningsKeys(), msgKey);
             } else if (this.processingEnv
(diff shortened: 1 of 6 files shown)


framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -199,6 +199,12 @@
     // org.checkerframework.framework.source.SourceeVisitor.checkForSuppressWarningsAnno
     "warnUnneededSuppressions",
 
+    // Require that warning suppression annotations contain a checker key as a prefix in order for
+    // the warning to be suppressed.
+    // org.checkerframework.framework.source.SourceChecker.checkSuppressWarnings(java.lang.String[],
+    // java.lang.String)
+    "requireCheckerPrefixInWarningSuppressions",
```
smillst: This name is pretty bad. Maybe just `requireCheckerInWarningSuppressions`?  Other ideas?

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -199,6 +199,12 @@
     // org.checkerframework.framework.source.SourceeVisitor.checkForSuppressWarningsAnno
     "warnUnneededSuppressions",
 
+    // Require that warning suppression annotations contain a checker key as a prefix in order for
+    // the warning to be suppressed.
+    // org.checkerframework.framework.source.SourceChecker.checkSuppressWarnings(java.lang.String[],
```
mernst: What is the purpose of these two lines of comment?  Is it a cut-and-paste error?

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
mernst: I think a long name is OK for this rarely-used feature.
I changed it to `requirePrefixInWarningSuppressions`.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
smillst: The comment at the top of the `@SupportedOptions` annotation says:
```
    // When adding a new standard option:
    // 1. Add a brief blurb here about the use case
    //    and a pointer to one prominent use of the option.
```
This is the pointer to the use of the option.

PR #2361, opened 2019-03-14
Abstract out a new method

mernst:


checker/src/main/java/org/checkerframework/checker/nullness/NullnessTransfer.java
```diff
@@ -264,21 +264,25 @@ protected NullnessValue finishValue(NullnessValue value, NullnessStore store) {
...
-                result.setResultValue(newResultValue);
+                refineToNonNull(result, n);
             }
         }
 
         return result;
     }
 
+    /** Refine the given result to @NonNull. */
+    void refineToNonNull(
```
smillst: This method is also refining the MethodInvocationNode `n` to non-null, but that's not mentioned in the Javadoc.  I would move the call to `makeNonNull(result, n)` out of this method.  Then the parameter `n` isn't needed either.

PR #2369, opened 2019-03-18
Make annotations repeatable for which wrappers not exist

gokaco:


gokaco:
Please take a look am I doing correct, for now ```@EnsuresNonNull``` and ```@EnsuresNonNullIf``` has been made repeatable.

checker/src/main/java/org/checkerframework/checker/nullness/qual/EnsuresMultiNonNull.java
```diff
@@ -0,0 +1,28 @@
...
+ * @checker_framework.manual #java-expressions-as-arguments Syntax of Java expressions
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD, ElementType.CONSTRUCTOR})
+@PostconditionAnnotation(qualifier = NonNull.class)
+@InheritedAnnotation
+public @interface EnsuresMultiNonNull {
+
+    EnsuresNonNull[] value();
```
mernst: Please write a Javadoc comment on every annotation element (field).

checker/src/main/java/org/checkerframework/checker/nullness/qual/EnsuresMultiNonNullIf.java
```diff
@@ -0,0 +1,27 @@
...
+ * @see EnsuresNonNullIf
+ * @checker_framework.manual #java-expressions-as-arguments Syntax of Java expressions
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD, ElementType.CONSTRUCTOR})
+@ConditionalPostconditionAnnotation(qualifier = NonNull.class)
+@InheritedAnnotation
+public @interface EnsuresMultiNonNullIf {
+    EnsuresNonNullIf[] value();
```
mernst: Please write a Javadoc comment on every annotation element (field).

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -418,16 +418,35 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+                List<String> expressions =
+                        AnnotationUtils.getElementValueArray(anno, "value", String.class, false);
+                AnnotationMirror postcondAnno = getAnnotationMirrorOfMetaAnnotation(metaAnno, anno);
+                if (postcondAnno == null) {
+                    continue;
+                }
+                for (String expr : expressions) {
+                    result.add(new Postcondition(expr, postcondAnno, anno));
+                }
+            } catch (Exception e) {
```
mernst: It's a bit suspicious to catch *all* exceptions.  What is the specific case that you are checking for?  I hope that you can check it in a normal if statement instead, and simplify the logic.

checker/src/main/java/org/checkerframework/checker/nullness/qual/EnsuresMultiNonNull.java
gokaco: Is this Ok?

checker/src/main/java/org/checkerframework/checker/nullness/qual/EnsuresMultiNonNullIf.java
gokaco: Is this Ok?

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
gokaco: There would be many If statements i.e. for every annotation like 
```
if (type.equals("org.checkerframework.checker.nullness.qual.EnsuresNonNull")
                    || type.equals("org.checkerframework.checker.nullness.qual.EnsuresKeyFor")|| .....
```
It will look messy. Should I do it?

gokaco: Or I can catch a specific exception i.e ```ClassCastException```

mernst: That would be better.  But the logic still seems strange.  It seems possible to iterate through most of the list, cause an exception, and then iterate through more of the list.  How about commenting what the exception means and how the second loop differs from the first one?

checker/src/main/java/org/checkerframework/checker/nullness/qual/EnsuresMultiKeyFor.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * @see EnsuresKeyFor
+ * @checker_framework.manual #java-expressions-as-arguments Syntax of Java expressions
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD, ElementType.CONSTRUCTOR})
+@PostconditionAnnotation(qualifier = KeyFor.class)
+@InheritedAnnotation
+public @interface EnsuresMultiKeyFor {
```
wmdietl: I think it would be nicer to use the base name of the annotation plus some suffix instead of adding `Multi` in the middle. So I would find `EnsuresKeyForContainer` or `EnsuresKeyForMultiple` clearer. Note that usually people won't write these container annotations, so we can pick a long name.
This comment applies to all new annotations this PR adds.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
wmdietl: I agree that this try/catch and the repeated logic is undesirable.
Why do you need it?
Don't these APIs look through the container API automatically and you only need to look for the single version?

checker/src/main/java/org/checkerframework/checker/nullness/qual/EnsuresMultiKeyFor.java
gokaco: Yes, will do the necessary changes.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
gokaco: > I agree that this try/catch and the repeated logic is undesirable.
> Why do you need it?
> Don't these APIs look through the container API automatically and you only need to look for the single version?

Can you please elaborate? I am unable to understand.
If I will not edit this file, then it will give the error and It was originally working for the single version.

gokaco:
The following annotations have been made repeatable:-
```
@EnsuresNonNull
@EnsuresNonNullIf
@EnsuresKeyFor
@EnsuresKeyForIf
@EnsuresLTLengthOf
@EnsuresLTLengthOfIf
@EnsuresLockHeld
@EnsuresLockHeldIf
@EnsuresMinLenIf
```

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOf.java
```diff
@@ -52,6 +53,7 @@
 @Target({ElementType.METHOD, ElementType.CONSTRUCTOR})
 @PostconditionAnnotation(qualifier = LTLengthOf.class)
 @InheritedAnnotation
+@Repeatable(EnsuresLTLengthOfMultiple.class)
```
mernst: By convention, the repeating annotation is given a plural name, which here would be `EnsuresLTLengthOfs`.
I agree that's a bit hard to read.

Another possibility is to use an [inner annotation named `@EnsuresLTLengthOf.List`](https://dzone.com/articles/repeatable-annotations-in-java-8-1).  I like that convention and propose that we adopt it.
@wmdietl @smillst What do you think?

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOfIfMultiple.java
```diff
@@ -0,0 +1,34 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import org.checkerframework.framework.qual.ConditionalPostconditionAnnotation;
+import org.checkerframework.framework.qual.InheritedAnnotation;
+
+/**
+ * An annotation wrapper to allow multiple postcondition annotations.
```
mernst: This comment is too vague.  It does not permit arbitrary postcondition annotations, only `@EnsuresLengthOfIf`.  Please say that instead:
```
A wrapper for multiple @EnsuresLengthofIf annotations.
```
Also, use a link rather than `@code`.

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOfIfMultiple.java
```diff
@@ -0,0 +1,34 @@
...
+ * @checker_framework.manual #index-checker Index Checker
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD, ElementType.CONSTRUCTOR})
+@ConditionalPostconditionAnnotation(qualifier = LTLengthOf.class)
+@InheritedAnnotation
+public @interface EnsuresLTLengthOfIfMultiple {
+    /**
+     * Constitutes the java expression(s) that are less than the length of the given sequences after
```
mernst: Use correct capitalization.  ("Java", not "java".)
You don't need the word "constitutes".

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOfMultiple.java
```diff
@@ -0,0 +1,33 @@
...
+ * @checker_framework.manual #index-checker Index Checker
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD, ElementType.CONSTRUCTOR})
+@PostconditionAnnotation(qualifier = LTLengthOf.class)
+@InheritedAnnotation
+public @interface EnsuresLTLengthOfMultiple {
+    /**
+     * Constitutes The java expressions that are less than the length of the given sequences on
```
mernst: You don't need "Constitutes".  "The" and "java" are both miscapitalized in the current text.

checker/src/main/java/org/checkerframework/checker/lock/qual/EnsuresLockHeldIfMultiple.java
```diff
@@ -0,0 +1,34 @@
...
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD, ElementType.CONSTRUCTOR})
+@ConditionalPostconditionAnnotation(qualifier = LockHeld.class)
+@InheritedAnnotation
+public @interface EnsuresLockHeldIfMultiple {
+    /**
+     * Constitutes the java expressions whose values are held after the method returns the given
+     * result.
+     *
```
mernst: This doesn't have the text about how programmers don't need to write the annotation.  I'm fine with omitting it, but then do so throughout, so that all the new annotations are consistent.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -418,16 +418,37 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
-                    AnnotationUtils.getElementValueArray(anno, "value", String.class, false);
-            AnnotationMirror postcondAnno = getAnnotationMirrorOfMetaAnnotation(metaAnno, anno);
-            if (postcondAnno == null) {
-                continue;
-            }
-            for (String expr : expressions) {
-                result.add(new Postcondition(expr, postcondAnno, anno));
+            try {
+                // If multiple contracts, then this will throw an exception which is catched at Line
+                // 433
```
mernst: Line numbers are never useful in source code, because they will become out of date as the code before them is edited.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -418,16 +418,37 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+                List<String> expressions =
+                        AnnotationUtils.getElementValueArray(anno, "value", String.class, false);
```
mernst: I think I understand the code now.  The `value` field is always either:
 * a list of Strings, or
 * a list of AnnotationMirrors.

Rather than using an "if" test, you attempt an assignment and that determines which case to execute.

Use of thrown and caught exceptions for control flow is confusing.
I suggest that you make an explicit test.
The test can set a variable that is a list of AnnotationMirror:  either
 * a newly-created singleton list, or
 * the list in the annotation

Then, most of the logic can be the same for both cases.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -418,16 +418,37 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+                AnnotationMirror postcondAnno = getAnnotationMirrorOfMetaAnnotation(metaAnno, anno);
```
mernst: Please add tests to ensure that this code works correctly.  (In general, when a pull request changes code without adding a test, that is a code smell.)

The tests should include various cases, including the possibility that both the list and the singleton annotation are present, and that both the list and multiple singleton annoattions are present.

mernst:
@gokaco Did you delete the branch upon which this pull request is based?
The top of this page now says

> 	gokaco wants to merge 48 commits into typetools:master from **unknown repository**

gokaco:
> @gokaco Did you delete the branch upon which this pull request is based?
> The top of this page now says
> 
> > gokaco wants to merge 48 commits into typetools:master from **unknown repository**

Oh, Mike I think so I deleted the whole fork of mine months back. How to recover it?

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOf.java
smillst: I like `@EnsuresLTLenghtOf.List`, too.

smillst:
Apparently GitHub saves changes in a pull request as a patch.  The patch for this pull request is 
https://github.com/typetools/checker-framework/pull/2369.patch 

To apply this patch to a branch, follow step one looking in the "command line instructions" link next to the "Squash and merge" button.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
mernst: Via reflection, [`AnnotatedElement.getAnnotationsByType(Class<T>)`](https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/AnnotatedElement.html#getAnnotationsByType-java.lang.Class-) abstracts away whether repeated annotations or a wrapper were used.  Maybe there is a compile-time interface that is similar that you could use.

gokaco:
> Apparently GitHub saves changes in a pull request as a patch. The patch for this pull request is
> https://github.com/typetools/checker-framework/pull/2369.patch
> 
> To apply this patch to a branch, follow step one looking in the "command line instructions" link next to the "Squash and merge" button.

@smillst I cannot see the "Squash and merge" because I think I am not authorized. It will be helpful if you can tell me the instructions.
Thanks

smillst:
```
git checkout -b gokaco-gokaco-patch-1 master
curl -L https://github.com/typetools/checker-framework/pull/2369.patch | git am -3
```

gokaco:
I have made a new branch and also made a pull request related to that so all future discussions should be on that pull request #2625
Thanks!

mernst:
@gokaco For clarification, have you addressed all the comments on this pull request?  If not, could you please do so, so that they do not get lost?  Then get me to review the new pull request.  Thanks!

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOf.java
gokaco: Done this convention in #2625

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOfIfMultiple.java
gokaco: @mernst I just did appiled the convention like `@EnsuresLTLenghtOf.List` in #2625. I have made some comments onto that. Please review it and give a comment on that.

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOfIfMultiple.java
gokaco: Done in #2625.

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOfMultiple.java
gokaco: Done in #2625.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
gokaco: @mernst I am not able to found the interface which can perform the task but I have modified the logic in a different way in #2625. Please give a look at it and review it!
Thanks.

PR #2372, opened 2019-03-19
Only check the primary annotations for constructor invocations.

### LINKED ISSUE #2330: Invalid receiver on constructor with polymorphic type parameters
Consider this class
```java
import java.util.*;

public class A<T extends @Tainted Object> {
    public @Untainted A(@PolyTainted int i) {
    }

    public @Untainted A() {
    }

    public static void f(@PolyTainted int i) {
        new @Untainted A<@PolyTainted Integer>(i);
        new @Untainted A<@PolyTainted Integer>();
    }
}
```

When running the tainting checker, I get this output
```
A.java:13: error: [constructor.invocation.invalid] creation of @Untainted A<@Tainted Integer> <init>() not allowed with given receiver;
        new @Untainted A<@PolyTainted Integer>();
        ^
  found   : @Untainted A<@PolyTainted Integer>
  required: @Untainted A<@Tainted Integer>
```

I don't think there should be an error for this line, it's just calling an empty constructor. This indicates the receiver for a method is invalid, which doesn't make sense for constructors.

Notice, that the line that causes an error is `new @Untainted A<@PolyTainted Integer>()`, not `new @Untainted A<@PolyTainted Integer>(i)`. The error only appears when there are no `@PolyTainted` parameters. For methods with no `@PolyTainted` parameters, the return type would usually default to `@Tainted`, but this shouldn't apply to type parameters. We might want to write `new List<@PolyTainted String>()`, and then add instances of `@PolyTainted String` later.

Regardless, the error message is confusing and worth fixing. We believe the error occurs because the new operator allocates a new `A<@PolyTainted Integer>` and then passes it to the `<init>` method, which expects `@Tainted Integer`. It probably expects `@Tainted Integer` because of the fact there are no `@PolyTainted` parameters.

smillst:
the type arguments are checked elsewhere. Fixes #2330.

### DIFF
--- a/framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
+++ b/framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
@@ -2642,33 +2642,19 @@ public class BaseTypeVisitor<Factory extends GenericAnnotatedTypeFactory<?, ?, ?
             AnnotatedDeclaredType invocation,
             AnnotatedExecutableType constructor,
             NewClassTree newClassTree) {
-        AnnotatedDeclaredType returnType = (AnnotatedDeclaredType) constructor.getReturnType();
-        // When an interface is used as the identifier in an anonymous class (e.g. new Comparable()
-        // {}) the constructor method will be Object.init() {} which has an Object return type When
-        // TypeHierarchy attempts to convert it to the supertype (e.g. Comparable) it will return
-        // null from asSuper and return false for the check.  Instead, copy the primary annotations
-        // to the declared type and then do a subtyping check.
-        if (invocation.getUnderlyingType().asElement().getKind().isInterface()
-                && TypesUtils.isObject(returnType.getUnderlyingType())) {
-            final AnnotatedDeclaredType retAsDt = invocation.deepCopy();
-            retAsDt.replaceAnnotations(returnType.getAnnotations());
-            returnType = retAsDt;
-        } else if (newClassTree.getClassBody() != null) {
-            // An anonymous class invokes the constructor of its super class, so the underlying
-            // types of invocation and returnType are not the same.  Call asSuper so they are the
-            // same and the is subtype tests below work correctly
-            invocation = AnnotatedTypes.asSuper(atypeFactory, invocation, returnType);
-        }
+        // Only check the primary annotations, the type arguments are checked elsewhere.
+        Set<AnnotationMirror> returnType = constructor.getReturnType().getAnnotations();
+        Set<AnnotationMirror> invocationAnnos = invocation.getAnnotations();
 
         // The return type of the constructor (returnType) must be comparable to the type of the
         // constructor invocation (invocation).
-        if (!(atypeFactory.getTypeHierarchy().isSubtype(invocation, returnType)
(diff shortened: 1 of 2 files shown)


framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -2642,33 +2642,19 @@ protected boolean checkConstructorInvocation(
...
-            retAsDt.replaceAnnotations(returnType.getAnnotations());
-            returnType = retAsDt;
-        } else if (newClassTree.getClassBody() != null) {
-            // An anonymous class invokes the constructor of its super class, so the underlying
-            // types of invocation and returnType are not the same.  Call asSuper so they are the
-            // same and the is subtype tests below work correctly
-            invocation = AnnotatedTypes.asSuper(atypeFactory, invocation, returnType);
-        }
+        // Only check the primary annotations, the type arguments are checked elsewhere.
+        Set<AnnotationMirror> returnType = constructor.getReturnType().getAnnotations();
```
wmdietl: How about renaming this to `returnAnnos` to be consistent with `invocationAnnos` below and to avoid confusion that this is an ATM?

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -2642,33 +2642,19 @@ protected boolean checkConstructorInvocation(
...
-        if (!(atypeFactory.getTypeHierarchy().isSubtype(invocation, returnType)
-                || atypeFactory.getTypeHierarchy().isSubtype(returnType, invocation))) {
+        if (!(atypeFactory.getQualifierHierarchy().isSubtype(invocationAnnos, returnType)
+                || atypeFactory.getQualifierHierarchy().isSubtype(returnType, invocationAnnos))) {
             checker.report(
                     Result.failure(
                             "constructor.invocation.invalid",
                             constructor.toString(),
-                            invocation,
+                            invocationAnnos,
```
wmdietl: How about adapting the error message in `framework/src/main/java/org/checkerframework/common/basetype/messages.properties` from
`constructor.invocation.invalid=creation of %s not allowed with given receiver;\nfound   : %s\nrequired: %s`

to something like
`constructor.invocation.invalid=creation of %s not allowed with given annotations;\nfound annotations "%s" not comparable to "%s"`

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -2677,15 +2663,10 @@ protected boolean checkConstructorInvocation(
...
-        if (!atypeFactory
-                .getQualifierHierarchy()
-                .isSubtype(returnType.getAnnotations(), invocation.getAnnotations())) {
+        if (!atypeFactory.getQualifierHierarchy().isSubtype(returnType, invocationAnnos)) {
             checker.report(
                     Result.warning(
-                            "cast.unsafe.constructor.invocation",
-                            returnType.toString(true),
-                            invocation.toString(true)),
+                            "cast.unsafe.constructor.invocation", returnType, invocationAnnos),
```
wmdietl: I think this message would be easier to read if it had the same arguments as the error message above, i.e. include the constructor that is invoked and order the components consistently.

PR #2379, opened 2019-03-22
Fix value checker bit-wise-and transfer function to take into account of the commutative property

txiang61:
Previous result:
`int & 0xff => @IntRange(0, 255)`
`0xff & int => @UnknownVal`
The bit-wise-and transfer function didn't take into account of the commutative property of `&`. 
Expected behavior:
`int & 0xff => @IntRange(0, 255)`
`0xff & int => @IntRange(0, 255)`


mernst:
There are tests failing.  Could you please ensure that they pass before you open a pull request?  (Or mark it as a draft and indicate what you are looking for feedback on.)

Could you also please add tests for your feature -- especially code that didn't type-check before but does now, but also code that still should not type-check.

Thanks!

txiang61:
> There are tests failing. Could you please ensure that they pass before you open a pull request? (Or mark it as a draft and indicate what you are looking for feedback on.)
> 
> Could you also please add tests for your feature -- especially code that didn't type-check before but does now, but also code that still should not type-check.
> 
> Thanks!

Addressed

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -396,9 +396,10 @@ public long getToValueFromIntRange(AnnotatedTypeMirror atm) {
      * Performs pre-processing on annotations written by users, replacing illegal annotations by
      * legal ones.
      */
-    private class ValueTypeAnnotator extends TypeAnnotator {
+    protected class ValueTypeAnnotator extends TypeAnnotator {
```
mernst: Are the changes in this file related to your bug fix?
If not, I suggest removing them from this pull request and putting them in a different pull request of their own.

mernst:
Thanks for this fix!

PR #2384, opened 2019-03-24
Make `MethodCall.isUnassignableByOtherCode` more precise.  Fixes #2334.

### LINKED ISSUE #2334: Index checker fails to recognize non immediate statements that follow an if statement.
Consider the following small program:
```
import java.util.*;
class indexcheck {
    Scanner sc = new Scanner(System.in);
    String a = sc.nextLine();
    String b = "";
    if (a.indexOf('d') != -1) {
      System.out.println("hey");
      b = a.substring(0, a.indexOf('d')); // valid index as the possibility of -1 is checked
    }
  }
}
```
One would expect no error as in b = a.substring(0,a.indexOf('d')); a.indexOf('d') will be non negative as checked by the if statement.
On the command line, type:
```javac -processor org.checkerframework.checker.index.IndexChecker indexcheck.java```
Output:
```
indexcheck.java:10: error: [argument.type.incompatible] incompatible types in argument.
      b = a.substring(0,a.indexOf('d'));
                                 ^
  found   : @GTENegativeOne int
  required: @NonNegative int
1 error
```

Now modify the code by removing System.out.println("hey"), i.e., the code becomes:

```
import java.util.*;
class indexcheck {
    Scanner sc = new Scanner(System.in);
    String a = sc.nextLine();
    String b = "";
    if (a.indexOf('d') != -1) {
      b = a.substring(0, a.indexOf('d'));
    }
  }
}
```
Now run by
```javac -processor org.checkerframework.checker.index.IndexChecker indexcheck.java```
and the compilation is successful.

### PR DESCRIPTION
**Pull request by mernst**
Also creates an `isUnmodifiableByOtherCode` method.  This has the same name
as one that was recently renamed to isUnassignableByOtherCode, but it has
more sensible semantics.  I have not examined which existing uses of
isUnassignableByOtherCode should be changed to isUnmodifiableByOtherCode.

### DIFF
--- a/checker/tests/index/Issue2334.java
+++ b/checker/tests/index/Issue2334.java
@@ -1,7 +1,6 @@
 // Test case for issue #2334: http://tinyurl.com/cfissue/2334
 
 import org.checkerframework.checker.index.qual.NonNegative;
-import org.checkerframework.dataflow.qual.Pure;
 
 class Issue2334 {
 
@@ -20,23 +19,5 @@ class Issue2334 {
             // :: error: (assignment.type.incompatible)
             @NonNegative int i = stringField.indexOf('d');
         }
-    }
-
-    public @Pure @Nullable Object getSuperclass() {
-        return null;
-    }
-
-    void setSuperclass(@Nullable Object no) {
-        // set the field returned by getSuperclass.
-    }
-
-    static void testInstanceofPositive3(PureTest pt) {
-        if (!(pt.getSuperclass() instanceof Object)) {
-            return;
-        } else {
-            pt.setSuperclass(null);
-        }
-        // :: error: (dereference.of.nullable)
-        pt.getSuperclass().toString();
     }
 }
--- a/checker/tests/index/Issue2334.java
+++ b/checker/tests/index/Issue2334.java
@@ -1,6 +1,7 @@
 // Test case for issue #2334: http://tinyurl.com/cfissue/2334
 
 import org.checkerframework.checker.index.qual.NonNegative;
+import org.checkerframework.dataflow.qual.Pure;
 
 class Issue2334 {
 
@@ -19,5 +20,23 @@ class Issue2334 {
             // :: error: (assignment.type.incompatible)
             @NonNegative int i = stringField.indexOf('d');
         }
+    }
+
+    public @Pure @Nullable Object getSuperclass() {
+        return null;
+    }
+
+    void setSuperclass(@Nullable Object no) {
+        // set the field returned by getSuperclass.
+    }
+
+    static void testInstanceofPositive3(PureTest pt) {
+        if (!(pt.getSuperclass() instanceof Object)) {
+            return;
+        } else {
+            pt.setSuperclass(null);
+        }
+        // :: error: (dereference.of.nullable)
+        pt.getSuperclass().toString();
     }
 }
--- a/checker/tests/index/Issue2334.java
+++ b/checker/tests/index/Issue2334.java
@@ -4,16 +4,33 @@ import org.checkerframework.checker.index.qual.NonNegative;
 
 class Issue2334 {
 
+    void hasSideEffect() {}
+
+    String stringField;
+
     void m1(String stringFormal) {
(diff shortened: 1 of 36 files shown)

### CODE REVIEW

File: javacutil/build.gradle [hunk 1]
```diff
@@ -3,6 +3,9 @@ repositories {
 }
 
 dependencies {
+    // https://mvnrepository.com/artifact/org.plumelib/plume-util
+    compile 'org.plumelib:plume-util:1.0.6'
```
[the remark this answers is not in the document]
mernst: @smillst Do [these changes](https://github.com/typetools/checker-framework/pull/2384/commits/d8aa16d194c5e28cd61be0c26f4fba2efa4b79c5) look right?

dataflow/src/main/java/org/checkerframework/dataflow/util/PurityUtils.java
```diff
@@ -38,12 +39,39 @@ public static boolean isDeterministic(AnnotationProvider provider, MethodTree tr
         return isDeterministic(provider, methodElement);
     }
 
+    /**
+     * Is the method {@code tree} deterministic?
+     *
+     * <p>This implementation does not require an AnnotationProvider and does not handle aliases.
+     * There are currently no aliases for @Deterministic, and the only alias for @Pure is
```
wmdietl: I think it would be much nicer to not add these additional overloaded versions.
The single use of these methods in `FlowExpressions.java` can be changed to use the available `AnnotationProvider`.
The issue isn't just aliases that this version would miss, it is also stub file support. Without an AnnotationProvider, this logic will not support stub files. This will lead to hard to debug problems and strange behavior.

mernst: > The single use of these methods in `FlowExpressions.java` can be changed to use the available `AnnotationProvider`.

What is the available `AnnotationProvider`?  I don't see one that is available at the call to `isDeterministic`, but maybe I missed something.

wmdietl: I think this is actually even easier: note that we only create a `MethodCall` if the method is deterministic, as checked here: https://github.com/typetools/checker-framework/blob/master/dataflow/src/main/java/org/checkerframework/dataflow/analysis/FlowExpressions.java#L176

https://github.com/typetools/checker-framework/pull/2384/commits/165101757681cd655fd424defbef4d9d0ecebe10 undoes these changes and locally tests passed.
Does this look good?

mernst: Clever!

This subtle property definitely needs to be documented, both at the call site that you removed and in the documentation for MethodCall.

checker/tests/index/Issue2334.java
```diff
@@ -0,0 +1,23 @@
+// Test case for issue #2334: http://tinyurl.com/cfissue/2334
```
wmdietl: The reference to #2334 was rather confusing. This PR seems to solve a different problem: information about immutable types can be maintained across modifications.
How about modifying this test case and using a general `sideeffecting()` method call instead of the `System.out.println` calls and adding a separate test case that uses `System.out.println`, but uses a mutable type instead of the immutable `String`. That second test is what #2334 is actually about.

dataflow/src/main/java/org/checkerframework/dataflow/util/PurityUtils.java
mernst: I added the documentation for the change.

checker/tests/index/Issue2334.java
mernst: This solves #2334.  There are other ways to do so, but I don't think we can say which way #2334 is actually about.  I expanded the test case in the way you suggested, and we could create a new issue to focus on other improvements.

javacutil/build.gradle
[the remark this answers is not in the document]
mernst: @smillst a ping before this gets merged.

mernst:
Thanks for the typo fixes. :-)

PR #2386, opened 2019-03-25
Clean up constructor/new class tree/type declaration semantics

### LINKED ISSUE #2230: Type annotations in javac type of object creation
I am trying to get the explicit annotations at constructor invocation sites which requires me to make a call to `TreeUtils.typeOf()` as shown [here](https://github.com/t-rasmud/checker-framework/commit/9fdde94fb19f9b6fd58de0c3e361d582daef52d2#diff-b997fe0effda8ef8635e7396f1861e64R1018).
However, printing out the results of this call produces ambiguous results.

Consider the test case below:
```

class ExtendsAndAnnotation {}

class IssueATFAmbiguity {
    void testTainted() {
        new @Tainted Object();
    }
}
```
Running the tainting checker on this example (`javac -processor tainting IssueATFAmbiguity.java`) along with the changes [here](https://github.com/t-rasmud/checker-framework/commit/9fdde94fb19f9b6fd58de0c3e361d582daef52d2#diff-b997fe0effda8ef8635e7396f1861e64R1018) prints the following line:
`... java.lang.Object`
But, if I comment out the line `class ExtendsAndAnnotation {}` in the test case, the same code prints out the following line:
`... (@org.checkerframework.checker.tainting.qual.Tainted :: java.lang.Object)`

### LINKED ISSUE #2324: getExplicitAnnotations() method in the AnnotatedTypeMirror class doesn't return expected result.

```

public class ExplicitAnnotationsIssue {
    void test() {
        Object x = new @Untainted Object();
    }
}
```
I want to get the explicit annotations on the `newClassTree` type which is `new @Untainted Object()` in the above example. I call the method `getExplicitAnnotations()` and print its size. See the diff [here](https://github.com/typetools/checker-framework/compare/master...t-rasmud:getExplicitAnnotationsIssue#diff-b997fe0effda8ef8635e7396f1861e64R2319).

I ran the tainting checker with the command: `javac -processor tainting ExplicitAnnotationsIssue.java`
I expected that the print statement would print `Explicit Annotations for new @Untainted() Object() : 1`. Instead, I get the following output:
```
Explicit Annotations for new @Untainted() Object() : 0
Explicit Annotations for new @Untainted() Object() : 0
Explicit Annotations for new @Untainted() Object() : 0
```

### LINKED ISSUE #2330: Invalid receiver on constructor with polymorphic type parameters

### LINKED ISSUE #2353: Value Checker unsafe cast warning for primitive wrappers applied to constants
Warnings are occurring while running the given piece of code:-
command-
`javac -processor index ex2.java`
code-
```
public class Issue2353 {

        Integer a = new Integer(1);
        Integer b = new Integer("2");
    }
}
```

Output:
```
Issue2353.java:5: warning: [cast.unsafe.constructor.invocation] constructor invocation cast from "@UnknownVal Integer" to "@IntVal(1) Integer" cannot be statically verified
        Integer b = new Integer(1);
                    ^
Issue2353.java:6: warning: [cast.unsafe.constructor.invocation] constructor invocation cast from "@UnknownVal Integer" to "@IntVal(2) Integer" cannot be statically verified
        Integer c = new Integer("2");
                    ^
2 warnings
```
How to make it verify statically?

### LINKED ISSUE #2385: Regression: Unable to type check code with Qualifiers parameterised by a dot (".")
The fix to #2186 has broken my custom checker which uses qualifiers with string arguments containing dots. For example, the following code fails to type-check due to an exception:
```
new @MyAnno(".") Object()
```

The stack trace is:

```
Exception: java.lang.StringIndexOutOfBoundsException: String index out of range: -2; Stack trace: java.lang.String.substring(String.java:1967) 
org.checkerframework.framework.type.AnnotatedTypeFactory.getExplicitAnnotationsOnNewClassTree(AnnotatedTypeFactory.java:2398)
 org.checkerframework.framework.type.AnnotatedTypeFactory.fromNewClass(AnnotatedTypeFactory.java:2367) 
 org.checkerframework.framework.type.GenericAnnotatedTypeFactory.fromNewClass(GenericAnnotatedTypeFactory.java:489) 
 my.namespace.XYZAnnotatedTypeFactory.fromNewClass(MyAnnotatedTypeFactory.java:329)  org.checkerframework.framework.type.TypeFromExpressionVisitor.visitNewClass(TypeFromExpressionVisitor.java:250) 
 org.checkerframework.framework.type.TypeFromExpressionVisitor.visitNewClass(TypeFromExpressionVisitor.java:44)
  com.sun.tools.javac.tree.JCTree$JCNewClass.accept(JCTree.java:1532)
  com.sun.source.util.SimpleTreeVisitor.visit(SimpleTreeVisitor.java:53)
  org.checkerframework.framework.type.TypeFromTree.fromExpression(TypeFromTree.java:36)
  ...
```

The root cause appears to be the logic in org.checkerframework.framework.type.AnnotatedTypeFactory.getExplicitAnnotationsOnNewClassTree(), which was introduced by #2186:

```Java
    // TODO: This is a hack.  Remove this method when http://tinyurl.com/cfissue/2324 is fixed.
    Set<AnnotationMirror> getExplicitAnnotationsOnNewClassTree(
            NewClassTree newClassTree, Set<? extends AnnotationMirror> allAnnotations) {
        // The following code extracts explicit annotations from "newClassTree" using string
        // manipulations.
        Set<AnnotationMirror> explicitAnnotations = new HashSet<>();
        String newClassTreeString = newClassTree.toString();
        for (AnnotationMirror anno : allAnnotations) {
            String annoString = anno.toString();
            String annoStringName =
                    annoString.substring(annoString.lastIndexOf('.') + 1, annoString.length() - 1);
            if (annoString.contains("(")) {
                annoStringName = annoStringName.substring(0, annoStringName.indexOf('(') - 1);
            }
            if (newClassTreeString.contains(annoStringName)) {
                explicitAnnotations.add(anno);
            }
        }
        return explicitAnnotations;
    }
```

Stepping through the code for the example above for the qualifier my.namespace.qual.MyAnno("."):
annoString = `my.namespace.qual.MyAnno(".")`
annoStringName = `")`
`annoString.contains("(")` returns true
`annoStringName.indexOf('(') - 1` is -2, resulting in an exception being thrown on the substring() call

I note that the code is also likely to misbehave for strings containing brackets ().
Since qualifiers with string parameters with dots (.) are common in my type checker (I am representing paths), I am unable to upgrade to checker 2.7.0 until this issue is fixed.

smillst:
1. Cleans up constructor/new class tree/type declaration semantics and implementation. (In particular, the changes in the following pull requests are mostly reverted and reimplemented:
https://github.com/typetools/checker-framework/pull/2347, https://github.com/typetools/checker-framework/pull/2338, https://github.com/typetools/checker-framework/pull/2322, https://github.com/typetools/checker-framework/pull/2295, and https://github.com/typetools/checker-framework/pull/2228.)

2. Adds methods where subclasses can override `inconsistent.constructor.type`,
`super.invocation.invalid`, `constructor.invocation.invalid`, and `cast.unsafe.constructor.invocation` errors.

3. Adds CONSTRUCTOR_RESULT to TypeUseLocation enum.

4. Fixes #2385, fixes #2330, fixes #2353, fixes #2324, fixes #2230.

Note: the manual needs to be update to reflect the new semantics, but I wanted to get the code up for review now.


framework/src/main/java/org/checkerframework/framework/util/defaults/QualifierDefaults.java
```diff
@@ -452,6 +452,10 @@ private void applyDefaults(Tree tree, AnnotatedTypeMirror type) {
 
             case IDENTIFIER:
                 elt = TreeUtils.elementFromUse((IdentifierTree) tree);
+                if (ElementUtils.isTypeDeclaration(elt)) {
```
smillst: I'm finding that the defaulting for TypeUseLocation.TYPE_DECLARATION isn't implemented correctly. This change is a work around to fix one problem I found, but I think I'm going to have to re-write that code in a separate PR.

checker/tests/tainting/Issue2243.java
```diff
@@ -20,3 +20,8 @@
 
 // :: error: (declaration.inconsistent.with.implements.clause)
 @Tainted class Z implements SuperClass {}
+
+class Issue224Test {
```
wmdietl: Should this be `Issue2243Test`?

checker/tests/tainting/Issue2330.java
```diff
@@ -0,0 +1,15 @@
+import java.util.*;
+import org.checkerframework.checker.tainting.qual.*;
+
+public class Issue2330<T extends @Tainted Object> {
+    @SuppressWarnings("tainting")
```
wmdietl: Could you add a comment why these suppressions are necessary? Would an annotation on the class declaration be an alternative?

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -381,42 +381,59 @@ public void processClassTree(ClassTree classTree) {
...
         // as in "class X {}". This is equivalent to writing "class X extends @Top Object {}", so
         // there is no need to do any subtype checking.
         if (classTree.getExtendsClause() != null) {
-            AnnotatedTypeMirror extendsClauseType =
-                    atypeFactory.getAnnotatedType(classTree.getExtendsClause());
-            checkExtendsOrImplementsClause(
-                    classTree, extendsClauseType, "declaration.inconsistent.with.extends.clause");
+            Set<AnnotationMirror> extendsAnnos =
+                    atypeFactory.getAnnotatedType(classTree.getExtendsClause()).getAnnotations();
+            for (AnnotationMirror classAnno : classType.getAnnotations()) {
```
wmdietl: How about extracting the `for` loop into a helper method, as the same code is used below? Similar to the previous `checkExtendsOrImplementsClause` helper method.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -381,42 +381,59 @@ public void processClassTree(ClassTree classTree) {
                 validateTypeOf(im);
             }
         }
+        checkExtendsImplements(classTree);
```
wmdietl: The code starting at line 369 is also about extends/implements clauses. Should it be moved into `checkExtendsImplements`?

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -381,42 +381,59 @@ public void processClassTree(ClassTree classTree) {
...
+        super.visitClass(classTree, null);
+    }
+
+    /**
+     * If "@B class Y extends @A X {}", then enforce that @B must be a subtype of @A.
+     *
+     * @param classTree class tree to check
+     */
+    protected void checkExtendsImplements(ClassTree classTree) {
+        if (classTree.getSimpleName().contentEquals("")) {
```
wmdietl: I thought there was a helper for this... I only found https://github.com/typetools/checker-framework/blob/master/javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java#L1206 which checks for a flag on the constructor. Maybe extracting this in a helper would be simpler.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1228,7 +1250,9 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
         // check precondition annotations
         checkPreconditions(node, contractsUtils.getPreconditions(invokedMethodElement));
 
-        checkSuperConstructorCall(node);
+        if (TreeUtils.isSuperCall(node)) {
```
wmdietl: Should we rename `TreeUtils.isSuperCall` and `TreeUtils.isThisCall` to something like `isSuperConstructorCall`?
I needed to look around to see that they only apply to constructor invocations. Or am I missing something?

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -705,6 +698,35 @@ public Void visitMethod(MethodTree node, Void p) {
...
+        Set<AnnotationMirror> classAnnotations =
+                atypeFactory
+                        .getAnnotatedType(constructorElement.getEnclosingElement())
+                        .getAnnotations();
+        QualifierHierarchy qualifierHierarchy = atypeFactory.getQualifierHierarchy();
+        for (AnnotationMirror classAnno : classAnnotations) {
+            AnnotationMirror constructorAnno =
+                    qualifierHierarchy.findAnnotationInSameHierarchy(
+                            constructorAnnotations, classAnno);
+            if (!qualifierHierarchy.isSubtype(classAnno, constructorAnno)) {
```
wmdietl: The javadoc says `Issue a warning if the result type of the constructor is a subtype of the declaring class.`.
Wouldn't that be more directly implemented as `if (qH.isSubtype(constuctorAnno, classAnno)`?
Ah, but we want to allow that they are the same. So maybe adjust the javadoc?

In the end, don't these rules enforce that the annotation on the constructor result is exactly the same as the one on the class declaration?
I thought the rule would be that a specific constructor could use a more specific type. Otherwise, why do we need annotations on constructor results at all?

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1238,43 +1262,30 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
...
-            Set<? extends AnnotationMirror> topAnnoations =
-                    atypeFactory.getQualifierHierarchy().getTopAnnotations();
-            for (AnnotationMirror topAnno : topAnnoations) {
-                AnnotationMirror superTypeMirror = superType.getAnnotationInHierarchy(topAnno);
-                AnnotationMirror constructorTypeMirror =
-                        constructorType.getReturnType().getAnnotationInHierarchy(topAnno);
-
-                if (!AnnotationUtils.areSameByName(
-                                constructorTypeMirror,
-                                "org.checkerframework.checker.initialization.qual.UnderInitialization")
```
wmdietl: Great that you could remove these hard-coded annotations!

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -2632,56 +2643,56 @@ protected void checkMethodInvocability(
...
+     *
+     * <p>Issue a warning if the annotations on the constructor invocation is a subtype of the
+     * constructor result type. This is equivalent to down-casting.
+     *
+     * @param invocation
+     * @param constructor
+     * @param newClassTree
+     * @return
+     */
     protected boolean checkConstructorInvocation(
```
wmdietl: Should we just make this method consistent with the others and make it return `void`? The single caller doesn't check the returned value and most other `checkXXX` methods are void.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
wmdietl: Oh, I also found https://github.com/typetools/checker-framework/blob/master/javacutil/src/main/java/org/checkerframework/javacutil/TypesUtils.java#L138

framework/src/main/java/org/checkerframework/framework/util/element/TypeDeclarationApplier.java
```diff
@@ -83,10 +83,15 @@ public static boolean accepts(final AnnotatedTypeMirror type, final Element elem
      */
     @Override
     protected void handleTargeted(List<Attribute.TypeCompound> extendsAndImplementsAnnos) {
-        for (final Attribute.TypeCompound anno : extendsAndImplementsAnnos) {
-
-            if (anno.position.type_index >= SUPERCLASS_INDEX && anno.position.location.isEmpty()) {
-                type.addAnnotation(anno);
+        if (typeSymbol.fullname.isEmpty()) {
```
wmdietl: Note how you check for an anonymous class differently in this part.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
wmdietl: Maybe this method should be called `checkExtendsAndImplementsClauses` and the helper `checkExtendsOrImplementsClause`. But there are probably better names...

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
smillst: >So maybe adjust the javadoc?

Yes, I'll do that. 

>why do we need annotations on constructor results at all?

We are going to add something to the manual about this.  I should probably add something here or a pointer to the manual.  The manual section will say something like this:

Any annotation on a constructor result, other than that on the declaration of the class, will result in a warning.  For an arbitrary checker, this is the desired behavior because the Checker Framework can't verify that the constructor result annotation is correct.  Specific checkers may be able to verify these annotations and thus not issue the warning.

For example, constructor results are always non-null, so it is legal to annotation the constructor result as `@NonNull` even if the annotation on the class declaration is `@Nullable`.  In the Aliasing Checker, a constructor result can be annotated with `@Unique` because the checker issues a warning if the constructed object is leaked.

checker/tests/tainting/Issue2330.java
smillst: I added a comment.   If we add `@Untainted` to the class declaration, a `super.invocation.invalid` error will be issued.  Also, the problem in Issue 2330 was when the annotation for the class declaration and the constructor result are not the same.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
smillst: I tried that, but I found it more confusing.  The helper method isn't really about extends/implements, its more like `ifNotSubtypeIssueError`. I also requires 4 arguments.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
smillst: Yes, they only apply to constructor invocations, so that rename would make sense.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
wmdietl: I agree with this somewhat; however, we also have no way to verify that the constructor produces an object of the type on the class declaration, either. So in some situations this should also warn if there is no annotation on the constructor.

Let's take a qualifier hierarchy `@A <: @B <: @C:`
For a class `@B class Foo { Foo() {} }` there is no way for us to veryify that the constructor establishes that the object actually fulfills the property `@B`. Maybe the default only establishes `@C` and the `@B` on the class is too strong.
Nothing special gets enforced about the annotation on the class declaration, so I don't see how we can guarantee that the constructor establishes that property.
Does that make sense?

smillst: You're saying that if a checker issues an `inconsistent.constructor.type` warning for `@A Foo(){}`, then it should also issue a warning for`@B Foo(){}`, right?  That makes sense to me.  Let's discuss with Mike tomorrow.

Note, a checker will issue a `super.invocation.invalid` warning for both constructors--unless Object was annotated with `@B`.

mernst: Can you write the example `@B class Foo { Foo() {} }` without defaults?  Defaults don't affect the answer, so leaving them off is confusing.

I'm assuming the constructor result type is `@B`.  In that case,  then there is no need for a warning, because every `Foo` (including the newly-constructed one) is a `@B` due to the class annotation.

PS: A small request: please use the hierarchy `@C <: @B <: @A`.  It is a long-standing tradition to assign letters from the top down, so it is confusing to reverse it.

wmdietl: Ok, switching to `@C <: @B <: @A`.
If I write `@A class Foo { @B Foo() {} }` I get a warning about the `inconsistent.constructor.type`.
If I write `@B class Foo { @B Foo() {} }` I do not get that warning.
However, the `@B` on the class declaration does not enforce anything about the constructed objects.
So why would the framework issue a warning for the first code and not the second? For neither code can the framework verify that the object is actually `@B`.

wmdietl: We discussed this some more: if a constructor result annotation is anything other than the top qualifier in the hierarchy, we issue the `inconsistent.constructor.type` error.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1238,43 +1263,30 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
...
-                            Result.failure(
-                                    "super.invocation.invalid",
-                                    constructorTypeMirror,
-                                    node,
-                                    superTypeMirror),
-                            enclosingMethod);
-                }
+                                superCall,
+                                superTypeMirror),
+                        enclosingMethod);
```
baoruiz: Currently the error is issued on the constructor declaration. Should we use the superCall MethodInvocationTree instead?

smillst: I did that. Thanks for pointing this out!

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -382,41 +397,44 @@ public void processClassTree(ClassTree classTree) {
...
         // If "@B class Y extends @A X {}", then enforce that @B must be a subtype of @A.
         // classTree.getExtendsClause() is null when there is no explicitly-written extends clause,
         // as in "class X {}". This is equivalent to writing "class X extends @Top Object {}", so
         // there is no need to do any subtype checking.
         if (classTree.getExtendsClause() != null) {
-            AnnotatedTypeMirror extendsClauseType =
-                    atypeFactory.getAnnotatedType(classTree.getExtendsClause());
-            checkExtendsOrImplementsClause(
-                    classTree, extendsClauseType, "declaration.inconsistent.with.extends.clause");
+            Set<AnnotationMirror> extendsAnnos =
```
wmdietl: Is there a reason to separate this validation from the one above? This now iterates twice over the interfaces.
Maybe we need two methods, one for the basic validation done above and a second one for this `declaration.inconsistent.with.extends.clause` check.
See my comment on the GuiEffectVisitor why this might be necessary.

checker/src/main/java/org/checkerframework/checker/guieffect/GuiEffectVisitor.java
```diff
@@ -229,6 +230,17 @@ public Void visitLambdaExpression(LambdaExpressionTree node, Void p) {
         return v;
     }
 
+    @Override
+    protected void checkExtendsImplements(ClassTree classTree) {
+        // Skip this check
```
wmdietl: Shouldn't at least the basic `validateTypeOf(im)` checks still be enforced? Like this, one could e.g. write multiple qualifiers on extends/implements.
Having separate methods for the basic and extended validation would allow selective overriding.

checker/src/main/java/org/checkerframework/checker/lock/LockVisitor.java
```diff
@@ -329,6 +329,12 @@ protected boolean skipReceiverSubtypeCheck(
         return annotationSet;
     }
 
+    @Override
+    protected void checkConstructorResult(
+            AnnotatedExecutableType constructorType, ExecutableElement constructorElement) {
+        // Skip this check
```
wmdietl: Why does this make sense for the Lock Checker? Can you make the comment more concrete?

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1238,43 +1260,30 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
...
-     * supertype of the return type of "super()" invocation within that constructor (except for the
-     * Initialization and Rawness Checkers).
+     * supertype of the return type of "super()" invocation within that constructor.
      */
-    void checkSuperConstructorCall(MethodInvocationTree node) {
-        if (!TreeUtils.isSuperCall(node)) {
-            return;
-        }
-        TreePath path = atypeFactory.getPath(node);
+    protected void checkSuperConstructorCall(MethodInvocationTree superCall) {
```
wmdietl: Do we need something similar for a `this` constructor call? The result types could also be inconsistent there, right?

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
wmdietl: Ping on this.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
wmdietl: Filed #2391 to remember this.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
smillst: I think these should just stay in processClassTree.  Otherwise it's too confusing on which method one should override.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
smillst: Sure, but that's a breaking change, so we would need to increase the minor number for this release.  We are only at 2.7.0, so I think we should wait.

wmdietl: This PR makes many breaking changes, so increasing the minor version seems fine to me.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
wmdietl: That's also not very nice, but let's do that for now.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
smillst: This PR does not change any public or protect methods, so all external checker will still compile.  Though I did changed the semantics of AnnotatedTypeFactory#newClassTree, so that counts as breaking.  Either way, this is low priority.

wmdietl: It adds a few protected methods and if type systems don't override them they will get unexpected warnings. So at least calling out these changes in the changelog seems important.

smillst: Yes, I absolutely need to add a change log for this PR.  I think the only new warning is the warning for any constructor result that isn't top.  The other warnings were in the last release.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
smillst: Yes! I implemented it.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -382,41 +379,62 @@ public void processClassTree(ClassTree classTree) {
...
+     *
+     * @param classTree class tree to check
+     */
+    protected void checkExtendsImplements(ClassTree classTree) {
+        if (TypesUtils.isAnonymous(TreeUtils.typeOf(classTree))) {
+            // Don't check extends clause on anonymous classes.
+            return;
+        }
+
+        AnnotatedTypeMirror classType = atypeFactory.getAnnotatedType(classTree);
```
wmdietl: Note that this was already calculated on line 364. If this computation isn't cached, it might be worth passing as a second parameter.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1238,43 +1262,43 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
...
+     */
+    protected void checkSuperConstructorCall(MethodInvocationTree superCall) {
+        checkThisOrSuperConstructorCall(superCall, "super.invocation.invalid");
+    }
+
+    /**
+     * Checks that the following rule is satisfied: The type on a constructor declaration must be a
+     * supertype of the return type of "this()" or "super()" invocation within that constructor.
+     */
+    private void checkThisOrSuperConstructorCall(
```
wmdietl: This method should be `protected`, so that subclasses only need to override one method.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1238,43 +1262,43 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
...
-     * supertype of the return type of "super()" invocation within that constructor (except for the
-     * Initialization and Rawness Checkers).
+     * supertype of the return type of "this()" invocation within that constructor.
      */
-    void checkSuperConstructorCall(MethodInvocationTree node) {
-        if (!TreeUtils.isSuperCall(node)) {
-            return;
-        }
-        TreePath path = atypeFactory.getPath(node);
+    protected void checkThisConstructorCall(MethodInvocationTree thisCall) {
```
wmdietl: We could make this method `final`, to not allow overrides in subclasses.
At least, there should be a reference to `checkThisOrSuperConstructorCall` to allow easy adaptation of behavior.

checker/src/main/java/org/checkerframework/checker/initialization/InitializationVisitor.java
```diff
@@ -94,6 +96,17 @@ protected boolean checkConstructorInvocation(
...
     }
 
+    @Override
+    protected void checkConstructorResult(
+            AnnotatedExecutableType constructorType, ExecutableElement constructorElement) {
+        // Nothing to check
+    }
+
+    @Override
+    protected void checkSuperConstructorCall(MethodInvocationTree superCall) {
```
wmdietl: This should be `checkThisOrSuperConstructorCall`.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
smillst: Fixed in https://github.com/typetools/checker-framework/pull/2403

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
smillst: FYI, there are breaking changes in other pull requests, so I changed this to a void method.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
smillst: It's cached.  But thanks for pointing this out.

checker/src/main/java/org/checkerframework/checker/initialization/InitializationVisitor.java
smillst: Good catch.  I fixed this:
https://github.com/typetools/checker-framework/commit/f5979aff85ec8f52c96674bcbe2bf58e973b3bec

PR #2392, opened 2019-03-29
Make value type annotator protected for future checker extensions

txiang61:


framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -396,9 +396,10 @@ public long getToValueFromIntRange(AnnotatedTypeMirror atm) {
      * Performs pre-processing on annotations written by users, replacing illegal annotations by
      * legal ones.
      */
-    private class ValueTypeAnnotator extends TypeAnnotator {
+    protected class ValueTypeAnnotator extends TypeAnnotator {
 
-        private ValueTypeAnnotator(AnnotatedTypeFactory atypeFactory) {
+        /** Changed to protected for future checker extensions */
```
wmdietl: The documentation shouldn't refer to the change, that's not useful to the reader.
Simply use something like `Construct a new ValueTypeAnnotator`.

wmdietl:
Thanks!

PR #2393, opened 2019-03-31
Restructure and document TypeFromExpressionVisitor

mernst:


framework/src/main/java/org/checkerframework/framework/type/TypeFromExpressionVisitor.java
```diff
@@ -83,8 +83,11 @@ public AnnotatedTypeMirror visitCompoundAssignment(
 
         // Recurse on the type of the variable.
         AnnotatedTypeMirror res = visit(node.getVariable(), f);
-        // TODO: why do we need to clear the type?
-        res.clearAnnotations();
+
+        if (f instanceof GenericAnnotatedTypeFactory
+                && ((GenericAnnotatedTypeFactory) f).getUseFlow()) {
+            res.clearAnnotations();
```
wmdietl: Could you add a comment that explains why this is only necessary when flow-sensitive type refinement is enabled?

mernst: The problem seems to be that getAnnotatedTypeLhs does not treat compound assignments specially.
It should do so, and then the call to cleanAnnotations can be removed entirely.

mernst:
Accidentally closed

framework/src/main/java/org/checkerframework/framework/type/TypeFromExpressionVisitor.java
```diff
@@ -80,11 +80,10 @@ public AnnotatedTypeMirror visitBinary(BinaryTree node, AnnotatedTypeFactory f)
...
             CompoundAssignmentTree node, AnnotatedTypeFactory f) {
-
-        // Recurse on the type of the variable.
-        AnnotatedTypeMirror res = visit(node.getVariable(), f);
-        // TODO: why do we need to clear the type?
-        res.clearAnnotations();
+        // The annotated type of the compound assignment depends on the checker,
+        // so let org.checkerframework.framework.type.treeannotator.PropagationTreeAnnotator
+        // add annotations.  See visitBinary above.
+        AnnotatedTypeMirror res = f.type(node);
```
smillst: This is basically the same as the previous behavior, but with an explanation of why it's the right thing.

mernst:
@smillst You are a reviewer.  Do you want to approve this pull request?
@wmdietl Do you want to take a quick look?  (It's a 3-line change.)

framework/src/main/java/org/checkerframework/framework/type/TypeFromExpressionVisitor.java
wmdietl: But the explanation seems incorrect to me.
https://github.com/typetools/checker-framework/blob/master/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java#L2729
says: "Determines an empty annotated type of the given tree." and the method is final, so there is no checker-specific behavior.
How does this end up calling the PropagationTreeAnnotator?
The change itself is probably fine - it still creates an unannotated type - just the comment above doesn't seem right to me.

smillst: The TreeAnnotators are called after this method.  I updated the comment.

wmdietl: How about moving the comment to the class and explaining there why some methods use `f.type(node);` and thereby create an un-annotated type?
The reference to `visitBinary above` doesn't help me, because there is no documentation in `visitBinary`.
Also note that there are other methods further down that return un-annotated types.
For example, the `// TODO: why not visit(node.getExpression(), f)` on line 300 is equally addressed with this explanation.

Documenting this once at the class level seems like a nicer solution.

PR #2394, opened 2019-03-31
Add missing super call

mernst:


mernst:
This should not be merged -- it's wrong.
We need to add documentation to indicate that a TreeAnnotator should not call super unless you have done something with the default method.

smillst:
Closed in favor of https://github.com/typetools/checker-framework/pull/2411.

PR #2400, opened 2019-04-02
Add signedness annotations for RandomAccessFile

gokaco:
By this change, most of the errors will get omitted which came in the case study-
https://github.com/gokaco/Signedness-Checker-2

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
+++ b/checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
@@ -3279,7 +3279,6 @@ class RandomAccessFile {
     private native int readBytes(@PolySigned byte b[], int off, int len) throws IOException;
     public int read(@PolySigned byte[] b) throws IOException;
     public int read(@PolySigned byte b[], int off, int len) throws IOException;
-    public void seek(@Unsigned long pos) throws IOException;
     public final void readFully(@PolySigned byte[] b) throws IOException;
     public final void readFully(@PolySigned byte[] b, int off, int len) throws IOException;
     public final @Unsigned int readUnsignedByte() throws IOException;
--- a/checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
+++ b/checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
@@ -3269,7 +3269,7 @@ class RandomAccessFile {
     public void write(@PolySigned byte b[]);
     public void write(@PolySigned byte b[], int off, int len);
     public int read(@PolySigned byte[] b) throws IOException;
-    public int read(@PolySigned byte b[], int off, int len) throws IOException;  
+    public int read(@PolySigned byte b[], int off, int len) throws IOException;
     public native @Constant long length() throws IOException;
     public void seek(@PolySigned long pos) throws IOException;
     public final void readFully(@PolySigned byte[] b) throws IOException;
--- a/checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
+++ b/checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
@@ -3270,7 +3270,7 @@ class RandomAccessFile {
     public void write(@PolySigned byte b[], int off, int len);
     public int read(@PolySigned byte[] b) throws IOException;
     public int read(@PolySigned byte b[], int off, int len) throws IOException;
-    public native @Constant long length() throws IOException;
+    public native @Unsigned long length() throws IOException;
     public void seek(@Unsigned long pos) throws IOException;
     public final void readFully(@PolySigned byte[] b) throws IOException;
     public final void readFully(@PolySigned byte[] b, int off, int len) throws IOException;
--- a/checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
+++ b/checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
@@ -3273,7 +3273,7 @@ class RandomAccessFile {
     public native @Constant long length() throws IOException;
     public void seek(@PolySigned long pos) throws IOException;
     public final void readFully(@PolySigned byte[] b) throws IOException;
-    public final void readFully(@PolySigned byte[] b) throws IOException;
+    public final void readFully(@PolySigned byte[] b, int off, int len) throws IOException;
 }
 
 package java.net;
--- a/checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
+++ b/checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
@@ -3268,12 +3268,34 @@ class PrintStream {
 class RandomAccessFile {
     public void write(@PolySigned byte b[]);
     public void write(@PolySigned byte b[], int off, int len);
+    public void write(@PolySigned int b) throws IOException;
+    public final void writeByte(@PolySigned int v) throws IOException;
+    public final void writeShort(@PolySigned int v) throws IOException;
+    public final void writeChar(@PolySigned int v) throws IOException;
+    public final void writeInt(@PolySigned int v) throws IOException;
+    public final void writeLong(@PolySigned long v) throws IOException;
+    private native void writeBytes(@PolySigned byte b[], int off, int len) throws IOException;
+    public final void writeChars(String s) throws IOException {
+        int clen = s.length();
+        int blen = 2*clen;
+        @Unsigned byte[] b = new byte[blen];
+        char[] c = new char[clen];
+        s.getChars(0, clen, c, 0);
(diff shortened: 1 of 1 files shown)


checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
```diff
@@ -3268,6 +3268,12 @@ class PrintStream {
 class RandomAccessFile {
     public void write(@PolySigned byte b[]);
     public void write(@PolySigned byte b[], int off, int len);
+    public int read(@PolySigned byte[] b) throws IOException;
+    public int read(@PolySigned byte b[], int off, int len) throws IOException;
+    public native @Constant long length() throws IOException;
+    public void seek(@PolySigned long pos) throws IOException;
```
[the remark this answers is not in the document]
gokaco: I think you are talking about the most significant bit that it should not be set as the argument should always be positive. Yes, you are right. Sorry for being careless. I have corrected it by annotating it as  ```@Unsigned```.

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
```diff
@@ -3268,6 +3268,12 @@ class PrintStream {
 class RandomAccessFile {
     public void write(@PolySigned byte b[]);
     public void write(@PolySigned byte b[], int off, int len);
+    public int read(@PolySigned byte[] b) throws IOException;
+    public int read(@PolySigned byte b[], int off, int len) throws IOException;
+    public native @Constant long length() throws IOException;
```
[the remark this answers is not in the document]
gokaco: I wanted it to be interpreted both as unsigned or signed that's why I did that. But signedness checker has a feature that it can interpret it as both signed and unsigned if it is within signed positive range so it would be good to annotate it with ```@Unsigned```. I have corrected it.

gokaco:
> There are other methods in RandomAccessFile that take or return bytes, namely `readUnsignedByte` and two overloaded `write` methods. Why didn't you annotate them? As described at https://checkerframework.org/manual/#annotating-libraries-create, "When you annotate a file, annotate the whole thing, not just a few of its methods." Otherwise, the next developer has to look over the entire file again.

Sorry, I thought only the functions which I did in my case study I should annotate that only as I would not be allowed to annotate others. Will annotate all of them as soon as possible.

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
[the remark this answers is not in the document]
mernst: Why do you want to interpret it both as signed or unsigned?
Presumably because it is definitely non-negative (the MSB is zero).
It would be helpful to write such a comment for clarification.
I would not write it as `@Unsigned`, because that would restrict its uses.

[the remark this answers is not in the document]
gokaco: I wanted it to interpreted as both because as you said as ```@Unsigned```, it would restrict its uses.
And in my case study, it was giving an error but by annotating the method with the annotation it resolved the issue.
I think for now we should not annotate this method. A proper study is needed or maybe a new annotation to solve this issue.

gokaco:
I have annotated most of the methods. Ready for review.

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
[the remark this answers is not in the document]
mernst: You are right that we can interpret signed positive numbers as either signed or unsigned.
The Signedness Checker [now](https://github.com/typetools/checker-framework/commit/e412a5f9a55937edeea04ebfddd0984786a0f1e9?w=1) behaves that way for all values -- before it only had that behavior for compile-time constants.

With the new feature, no Signedness annotation is needed -- the `@NonNegative` annotation on `length()`, which ought to already appear in the JDK, is sufficient, and I think is more clear.

I also opened [an issue](https://github.com/typetools/checker-framework/issues/2457) to rename `@Constant` to `@SignednessEither`.  It is misleading to call the annotation `@Constant` when it applies more broadly than that.

Thanks for bringing up this point!

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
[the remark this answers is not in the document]
mernst: You are right -- I mixed up "least" and "most" in my comment.  Thanks for figuring it out anyway!

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
```diff
@@ -3268,6 +3268,34 @@ class PrintStream {
...
     public void write(@PolySigned byte b[], int off, int len);
+    public void write(@PolySigned int b) throws IOException;
+    public final void writeByte(@PolySigned int v) throws IOException;
+    public final void writeShort(@PolySigned int v) throws IOException;
+    public final void writeChar(@PolySigned int v) throws IOException;
+    public final void writeInt(@PolySigned int v) throws IOException;
+    public final void writeLong(@PolySigned long v) throws IOException;
+    private native void writeBytes(@PolySigned byte b[], int off, int len) throws IOException;
+    public final void writeChars(String s) throws IOException {
+        int clen = s.length();
```
mernst: Why does your stub file contain the implementation of `writeChars` but not of other methods?  I think it would be better to be consistent -- either the full file, or only method signatures for each annotated method.

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
```diff
@@ -3268,6 +3268,34 @@ class PrintStream {
...
+            b[j++] = (byte)(c[i] >>> 8);
+            b[j++] = (byte)(c[i] >>> 0);
+        }
+        writeBytes(b, 0, blen);
+    }
+    public native @PolySigned int read() throws IOException;
+    private native int readBytes(@PolySigned byte b[], int off, int len) throws IOException;
+    public int read(@PolySigned byte[] b) throws IOException;
+    public int read(@PolySigned byte b[], int off, int len) throws IOException;
+    public void seek(@Unsigned long pos) throws IOException;
```
mernst: This annotation looks wrong to me.  The argument is interpreted as signed.  In particular, passing
9223372036854775808 is illegal because it is interpreted as a negative number.

The value does need to be positive, but that is orthogonal to how the argument is interpreted, which is what `@Unsigned` is about.

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
gokaco: There is no need to implement other methods. Only method signatures for each annotated method is sufficient.
Only in ```writeChars``` method, the array was storing unsigned values, therefore, it was needed to be implemented.

mernst: Can you tell me what goes wrong if you leave the method body out of this file?

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
gokaco: Yes, I got your point. It looks wrong to me too, but it in the case study done by me the parameter which was getting passed was ```@Unsigned```, so will we always have to suppress this kind of error whenever it occurs.

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
gokaco: Nothing goes wrong if we use this method in a program. As signedness checker only runs on the method signature and not on its body. So I think it is not necessary to annotate this.

gokaco: Nothing goes wrong if we use this method in a program. As it works normally. So I think it is not necessary to annotate this.

checker/src/main/java/org/checkerframework/checker/signedness/jdk.astub
mernst: In the case study code, the value should be in the positive integer range.  Not every unsigned value is permitted.  So it sounds like the case study code might be wrong (it could pass wrong values), or there are properties you need to add to it, such as an [`@IntRange`](https://checkerframework.org/api/org/checkerframework/common/value/qual/IntRange.html) annotation.

You should never write an incorrect annotation in a library just to suppress a warning in some particular client.  Incorrect library annotations make the checker unsound for every client program, which is a bad tradeoff.

gokaco: Thanks for the prompt reply. I got it.

mernst:
Thanks for contributing these annotations!

PR #2413, opened 2019-04-08
Diamond inference improvements

smillst:
This reverts some of the changes in https://github.com/typetools/checker-framework/pull/2193 and instead fixes the AsSuper crashes by finding the correct type for type arguments to diamond trees.


wmdietl:
Thanks!

PR #2414, opened 2019-04-08
Use `implementation` rather than `compile` dependency type

mernst:


mernst:
> Some of these should maybe be ‘api’. 

That is only possible if you use the ['java-library' plugin], which our buildfiles do not do.

> Why not change ‘checker/build.gradle’, too?

Because if I change it, tests fail:  at run time, plume-util is needed, but cannot be found.  I'm not sure why that happens, but these others can be changed without test failures.

mernst:
We should instead use the `java-library` plugin, make some instances use `api` instead of `implementation`, and wait until the shadow plugin handles `implementation` (the shadow plugin fails if we make the same changes in the `checker/` directory, which is why it isn't in this pull request).

PR #2416, opened 2019-04-10
Remove broken compilerPom installation.

wmdietl:
There is no `maven-artifacts/poms/compilerPom.xml` file.

### DIFF
--- a/maven-artifacts/poms/localInstall.sh
+++ b/maven-artifacts/poms/localInstall.sh
@@ -13,7 +13,6 @@ dataflowDist="${mydir}/../../dataflow/dist"
 
 eval "mvn install:install-file -DpomFile=${mydir}/checkerQualPom.xml -Dfile=${checkerDist}/checker-qual.jar"
 eval "mvn install:install-file -DpomFile=${mydir}/checkerPom.xml -Dfile=${checkerDist}/checker.jar"
-eval "mvn install:install-file -DpomFile=${mydir}/compilerPom.xml -Dfile=${checkerDist}/javac.jar"
 eval "mvn install:install-file -DpomFile=${mydir}/jdk8Pom.xml -Dfile=${checkerDist}/jdk8.jar"
 eval "mvn install:install-file -DpomFile=${mydir}/javacutilPom.xml -Dfile=${javacutilDist}/javacutil.jar"
 eval "mvn install:install-file -DpomFile=${mydir}/dataflowPom.xml -Dfile=${dataflowDist}/dataflow.jar"


wmdietl:
Should each sub-project create a separate `dist`, e.g. should there be a `framework/dist/framework.jar`?
Or does that only make sense for `checker/dist`, which contains the fat jar and dependencies, mostly for the release.
Or is this file just out-of-date and should be completely removed?

wmdietl:
Also note that shadowjar, in particular `./gradlew framework:shadowjar` creates a fat `framework/dist/framework.jar` file.

smillst:
>Should each sub-project create a separate dist, e.g. should there be a framework/dist/framework.jar?
Or does that only make sense for checker/dist, which contains the fat jar and dependencies, mostly for the release.

Only checker has a `dist` because that's where the checker fat jar is and dependencies needed for checker/bin/javac.  The other projects follow the Gradle convention of putting jars in .

>Or is this file just out-of-date and should be completely removed?

Yes.  (I mean if you find it useful we could update it, but it's not currently used.)

smillst:
>Also note that shadowjar, in particular ./gradlew framework:shadowjar creates a fat framework/dist/framework.jar file.

I'd forgotten that.  The other shadowjars are in `build/libs/dataflow-2.8.0-all.jar`

wmdietl:
A somewhat related question: do we need the pom files in `https://github.com/typetools/checker-framework/tree/master/maven-artifacts/poms` at all?
Don't the files in `maven-artifacts/release` simply contain more information?
Could we remove the files in `poms` and only keep the versions in `release`, moved to a proper location one level up?

smillst:
The poms in maven-artifacts/poms are used in the release to test the maven artifacts. I don’t know if the release poms could be used for these tests or not.

smillst:
>Do we need the pom files in https://github.com/typetools/checker-framework/tree/master/maven-artifacts/poms at all?

Yes, because these poms are for releasing to https://checkerframework.org/dev/m2-repo, which we do during the release.

mernst:
> > Do we need the pom files in https://github.com/typetools/checker-framework/tree/master/maven-artifacts/poms at all?
> 
> Yes, because these poms are for releasing to https://checkerframework.org/dev/m2-repo, which we do during the release.

It might be good to document that purpose in a README in that directory.

wmdietl:
I'll test whether we can use the same poms for both purposes. The duplication is hard to maintain. Maybe a simple script could filter out parts that are not allowed for the local installation.

PR #2417, opened 2019-04-10
Delete README-jdk

wmdietl:
The information is incorrect and does not seem useful in this location.


maven-artifacts/release/README-jdk
```diff
@@ -1,10 +0,0 @@
-This .jar file is a standin for either source/Javadoc for one of the
```
smillst: This file is used in the release process here:
https://github.com/typetools/checker-framework/blob/master/docs/developer/release/release.xml#L304

wmdietl: Aha, I see. I'll close this and document this in `maven-artifacts/release/README`.

PR #2425, opened 2019-04-11
Inline constants that are used just once and aren't documented

mernst:


mernst:
Not a bad pull request, but not needed to prevent documentation warnings.

PR #2428, opened 2019-04-11
Don't create LessThan annotations with double value

### LINKED ISSUE #2030: Index Checker creates `@LessThan(doubleValue)`
Consider:
```java
package math;

import java.math.RoundingMode;

public class UnparsableBug {
    double roundIntermediate(double x, RoundingMode mode){
        if (x <= 0.0) {
            return x;
        } else {
            return (long) x + 1;
        }

    }

    }
}
```
Warning:
```java
Error:(15, 22) java: [expression.unparsable.type.invalid] Expression in dependent type annotation invalid: [error for expression: 0.0; error: '0']
```

smillst:
Fixes #2030.

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/index/inequality/LessThanTransfer.java
+++ b/checker/src/main/java/org/checkerframework/checker/index/inequality/LessThanTransfer.java
@@ -60,7 +60,7 @@ public class LessThanTransfer extends IndexAbstractTransfer {
                 // right is already bottom, nothing to refine.
                 return;
             }
-            if (isDoubleOrFloatLiteral(leftRec)) {
+            if (!isDoubleOrFloatLiteral(leftRec)) {
                 lessThanExpressions.add(leftRec.toString());
             }
             Receiver rightRec = FlowExpressions.internalReprOf(analysis.getTypeFactory(), right);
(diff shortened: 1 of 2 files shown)


checker/src/main/java/org/checkerframework/checker/index/inequality/LessThanTransfer.java
```diff
@@ -57,12 +60,26 @@ protected void refineGT(
...
+            if (isParsable(leftRec)) {
+                lessThanExpressions.add(leftRec.toString());
+            }
             Receiver rightRec = FlowExpressions.internalReprOf(analysis.getTypeFactory(), right);
             store.insertValue(rightRec, factory.createLessThanQualifier(lessThanExpressions));
         }
     }
 
+    /**
+     * Return true if {@code receiver}'s toString can be parsed by {@link FlowExpressionParseUtil}.
```
mernst: I think it would be helpful for the comment to indicate that this is not a test of whether the expression is valid or well-formed (which was my first guess), but merely whether it is of a permitted type.  Also, returning true is not a promise that the expression is parseable -- but returning false means that it is not.  (At least, that's what I think.  I might be wrong, which is a reason to improve the documentation.)

Also:  It seems that this routine belongs in `FlowExpressionParseUtil` rather than in this class.

checker/src/main/java/org/checkerframework/checker/index/inequality/LessThanTransfer.java
```diff
@@ -131,4 +140,17 @@ protected void refineGTE(
             return Collections.emptyList();
         }
     }
+
+    /**
+     * Return true if {@code receiver} is a double or float literal, which can't be parsed by {@link
+     * FlowExpressionParseUtil}.
+     */
+    private boolean isDoubleOrFloatLiteral(Receiver receiver) {
```
mernst: Contrary to its documentation and name, this returns *false* if the argument is a double or float literal.  I think the logic needs to be changed here, and all uses need to be changed to `if (!isDoubleOrFloatLiteral(...))`.

PR #2431, opened 2019-04-16
Re-implement defaulting for TYPE_DECLARATION.

smillst:
It's not possible to determine if an element is a type declaration or a type use, so that information has to be passed to the defaulting code.  It is possible to tell if a tree is a type declaration, but extends and implements clause should be defaulted as type declarations.  Also, `getAnnotatedType(Element)` converts elements to trees, so even if a tree is used, it might be from a type use.

Also, adds Nullable as the default for TYPE_DECLARATION.

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/nullness/qual/Nullable.java
+++ b/checker/src/main/java/org/checkerframework/checker/nullness/qual/Nullable.java
@@ -5,6 +5,7 @@ import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
+import org.checkerframework.framework.qual.DefaultFor;
 import org.checkerframework.framework.qual.DefaultInUncheckedCodeFor;
 import org.checkerframework.framework.qual.ImplicitFor;
 import org.checkerframework.framework.qual.LiteralKind;
@@ -29,5 +30,6 @@ import org.checkerframework.framework.qual.TypeUseLocation;
 @DefaultInUncheckedCodeFor({TypeUseLocation.RETURN, TypeUseLocation.UPPER_BOUND})
 @Documented
 @Retention(RetentionPolicy.RUNTIME)
+@DefaultFor(TypeUseLocation.TYPE_DECLARATION)
 @Target({ElementType.TYPE_USE, ElementType.TYPE_PARAMETER})
 public @interface Nullable {}
--- a/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
@@ -1023,6 +1023,37 @@ public class AnnotatedTypeFactory implements AnnotationProvider {
     }
 
     /**
+     * Returns an AnnotatedTypeMirror representing the annotated type of {@code elt} as a type
+     * declaration.
+     *
+     * <p>It is not possible to tell whether an element is a use of a type or a declaration, so this
+     * method differentiates the two.
+     *
+     * @param elt element whose type is requested
+     * @return the annotated type of {@code elt} as a type declaration
+     */
+    public AnnotatedTypeMirror getAnnotatedTypeAsTypeDecl(Element elt) {
+        return getAnnotatedType(elt);
+    }
+
+    /**
+     * Returns an AnnotatedTypeMirror representing the annotated type of {@code tree} as a type
+     * declaration.
+     *
+     * <p>It is possible to tell whether a tree is a type declaration, but extends and implements
+     * clauses should also be treated as type declarations so defaults are applied correctly. (Also,
+     * {@link #getAnnotatedType(Element)} converts elements to ClassTrees if one is available, so
+     * even if the tree is a class declaration, that doesn't mean that it should be treated as a
+     * type declaration.)
+     *
+     * @param tree element whose type is requested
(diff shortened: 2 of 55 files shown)


framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -161,6 +161,9 @@
      */
     private boolean shouldDefaultTypeVarLocals;
 
+    /** Should the type be calculated as a type declaration? */
```
wmdietl: Some more documentation here would be helpful to understand what `the type` this is referring to.

checker/src/main/java/org/checkerframework/checker/nullness/qual/Nullable.java
```diff
@@ -29,5 +30,6 @@
 @DefaultInUncheckedCodeFor({TypeUseLocation.RETURN, TypeUseLocation.UPPER_BOUND})
 @Documented
 @Retention(RetentionPolicy.RUNTIME)
+@DefaultFor(TypeUseLocation.TYPE_DECLARATION)
```
wmdietl: I think improving the documentation for `TypeUseLocation.TYPE_DECLARATION` would be important.
Using this default is different from other defaults - the behavior between a class explicitly annotated as `@Nullable` is different from a class that gets this default `@Nullable`.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1022,6 +1022,37 @@ protected int getCacheSize() {
         }
     }
 
+    /**
+     * Returns an AnnotatedTypeMirror representing the annotated type of {@code elt} as a type
+     * declaration.
+     *
+     * <p>It is not possible to tell whether an element is a use of a type or a declaration, so this
```
wmdietl: Technically, an `Element` is always a representation of a declaration, so this description reads a bit strange.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1022,6 +1022,37 @@ protected int getCacheSize() {
...
+     */
+    public AnnotatedTypeMirror getAnnotatedTypeAsTypeDecl(Element elt) {
+        return getAnnotatedType(elt);
+    }
+
+    /**
+     * Returns an AnnotatedTypeMirror representing the annotated type of {@code tree} as a type
+     * declaration.
+     *
+     * <p>It is possible to tell whether a tree is a type declaration, but extends and implements
```
wmdietl: Why should extends/implements clauses also get the same default? This sentence is a bit confusing.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1022,6 +1022,37 @@ protected int getCacheSize() {
...
+     * declaration.
+     *
+     * <p>It is not possible to tell whether an element is a use of a type or a declaration, so this
+     * method differentiates the two.
+     *
+     * @param elt element whose type is requested
+     * @return the annotated type of {@code elt} as a type declaration
+     */
+    public AnnotatedTypeMirror getAnnotatedTypeAsTypeDecl(Element elt) {
+        return getAnnotatedType(elt);
```
wmdietl: Can you add a comment why this is the right implementation?
Why not move the `isTypeDecl` field here and document that subclasses should use that field to implement special logic?

framework/src/main/java/org/checkerframework/framework/util/defaults/QualifierDefaults.java
```diff
@@ -352,7 +352,18 @@ private boolean conflictsWithExistingDefaults(
      * @param type the type to annotate
      */
     public void annotate(Element elt, AnnotatedTypeMirror type) {
```
wmdietl: I'm wondering whether it would be nicer to directly expose the boolean in the public API.
This would be a breaking change, but would force people to adapt to this changed logic. At the moments existing clients would silently not override all the right methods and get strange behavior.
It would also keep this class a bit more manageable.

framework/src/main/java/org/checkerframework/framework/util/defaults/QualifierDefaults.java
```diff
@@ -465,20 +487,6 @@ private void applyDefaults(Tree tree, AnnotatedTypeMirror type) {
                 // TODO cases for array access, etc. -- every expression tree
                 // (The above probably means that we should use defaults in the
                 // scope of the declaration of the array.  Is that right?  -MDE)
-
-            case CLASS:
```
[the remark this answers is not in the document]
smillst: Unfortunately yes,  when getting the type of an element, which could be the declaration or a use, if the tree is available it is used. See [AnnotatedTypeFactory#fromElement](https://github.com/typetools/checker-framework/blob/master/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java#L1187).

checker/src/main/java/org/checkerframework/checker/nullness/qual/Nullable.java
smillst: I'll update [this section](https://checkerframework.org/manual/#annotations-on-class-declarations) of the manual.
Do you this I should add documentation here and/or at the declaration of TypeUseLocation, too?

smillst:
Closed in favor of a new design.

PR #2440, opened 2019-04-21
Fixed offset annotations

Maxi17:
In input classes from java.io package, some methods that take an offset parameter were't annotated according to the documentation. I tried fix those, along with other minor improvements to some already existing annotations. Please refer to [this](https://groups.google.com/forum/#!topic/checker-framework-gsoc/-SmMpAxG8V0) conversation for more observations that can improve the annotations.


checker/jdk/index/src/java/io/PipedInputStream.java
```diff
@@ -155,7 +155,7 @@ public PipedInputStream(@NonNegative int pipeSize) {
         initPipe(pipeSize);
     }
 
-    private void initPipe(@NonNegative int pipeSize) {
+    private void initPipe(@Positive int pipeSize) {
```
kelloggm: Since this method is private to the JDK, we should probably just remove the annotation from it altogether.

checker/jdk/index/src/java/io/PipedReader.java
```diff
@@ -128,7 +128,7 @@ public PipedReader(@NonNegative int pipeSize) {
         initPipe(pipeSize);
     }
 
-    private void initPipe(int pipeSize) {
+    private void initPipe(@Positive int pipeSize) {
```
kelloggm: Since this method is private to the JDK, it doesn't need an annotation.

checker/jdk/index/src/java/io/PipedInputStream.java
mernst: I agree private methods don't *need* annotations.
However, I see no harm in writing them if they are accurate.
Eventually, we would like to type-check the JDK rather than trusting the annotations, and at that time the annotations will come in useful.

Maxi17:
> Would you be willing to write a few test cases that use one of these classes which issue a spurious warning using the old JDK annotations, but succeed with this set of annotations?

I created an initial test case on my latest commit, but I'm not sure I understood correctly what you asked me to do. The problem with the old annotations was that the checker didn't produce any warning at all in situations similar to my test case, so I didn't know how to produce a spurious warning. The checker issues a warning with the new annotations in my example, which is the intended behavior. The problem was that the old annotations were too permissive, not too restrictive.

mernst:
You should create test cases for which the checker produces a warning.
The [README file in directory `checker/tests`](https://github.com/typetools/checker-framework/blob/master/checker/tests/README) tells you how to specify expected warnings, which is to write a
```// ::```
comment on the previous line.  Please ask if you have questions about this.

**Comment by kelloggm** [comment]
@Maxi17, I looked into the build failures we discussed earlier. They're concerning, because they seem to indicate that the Index Checker is unsound: certainly the errors you expected should have been issued.

I wrote a test case that isolates the offset check, which is below. On the latest version of `typetools/checker-framework`'s `master` branch, this test passes (i.e. only the expected errors are issued). Can you check if it does on your branch, as well?

```java
import org.checkerframework.checker.index.qual.IndexOrHigh;
import org.checkerframework.checker.index.qual.LTLengthOf;

public class OffsetsAndConstants {
    static int read(char[] a,
                    @IndexOrHigh("#1") int off,
                    @NonNegative @LTLengthOf(value = "#1", offset = "#2 - 1") int len) {
        int sum = 0;
        for (int i = 0; i < len; i++) {
            sum += a[i + off];
        }
        return sum;
    }

        char[] a = new char[10];

        read(a, 5, 4);

        read(a, 5, 5);

        read(a, 5, 6);

        read(a, 5, 7);
    }
}
```

Maxi17:
Indeed, the build fails on my branch. This is the error message:
```
tests.IndexTest > run[index] FAILED
    java.lang.AssertionError: 389 out of 393 expected diagnostics were found.
    OffsetsAndConstants.java:25: error: (argument.type.incompatible)
    OffsetsAndConstants.java:28: error: (argument.type.incompatible)
    4 expected diagnostics were not found:
    OffsetAnnotations.java:8: error: (argument.type.incompatible)
    OffsetAnnotations.java:15: error: (argument.type.incompatible)
    OffsetsAndConstants.java:25: error: assignment.type.incompatible
    OffsetsAndConstants.java:28: error: assignment.type.incompatible
```

I will change the expected error messages in OffsetsAndConstants.java from

`assignment.type.incompatible`

to

`(argument.type.incompatible)`

as suggested in the assertion error above. It is expected to work, but the initial problem, related to
```
OffsetAnnotations.java:8: error: (argument.type.incompatible)
OffsetAnnotations.java:15: error: (argument.type.incompatible)
```
will likely persist.

kelloggm:
That was my mistake with the test case - was running it manually, not through the framework, and inspecting the output by hand. `argument.type.incompatible` is the right error message.

There's clearly something different between the case in `OffsetAnnotations.java` and the case in that new test case. The annotations are identical, though, so I don't know what's different.

Maxi17:
This is a build that succeeded:
https://travis-ci.org/Maxi17/checker-framework/builds/527438363
The changes I made are:
https://github.com/Maxi17/checker-framework/commit/d1924d2a29e25ed010503861d4cde5c3563af3d6

One explanation could be that the tests are executed with the old JDK annotations in `java.io`. Can this be the case?

kelloggm:
Good catch! That's almost certainly what is happening. @smillst can you help @Maxi17 merge these JDK annotations properly? It's been a long time since I've done that and I don't want to break the build for everyone.

smillst:
The jdk.jar job builds the JDK then runs tests for some of the checkers, including the Index Checker.  That job passed: https://travis-ci.org/typetools/checker-framework/jobs/530451847.  So, to make all-tests pass, you need to change the pre-built jdk used by Travis.  Instructions are here: https://github.com/typetools/annotated-libraries/blob/master/README.md.  (Since you don't have permission to push to that repository, just make a pull request and either Martin or I can merge it.)

Maxi17:
The build of the updated JDK fails on my machine. I submitted a bug report in that repository: [link](https://github.com/typetools/annotated-libraries/issues/14).

PR #2443, opened 2019-04-21
Rename two methods

mernst:
isPolymorphicQualified => hasPolymorphicQualifier
Previously, the code had isPolymorphicQualified and getPolymorphicQualifier.
I feel that parallelism between hasPolymorphicQualifier and getPolymorphicQualifier is nicer.

 getPolymorphicQualifierTop => getPolymorphicQualifierElement
The new name better reflects the actual semantics.
We could instead require that the argument to @PolymorphicQualifier must be exactly
top, and in that case this renaming would not be necessary.


smillst:
>getPolymorphicQualifierElement => getPolymorphicQualifierTop

This is backward.  You renamed getPolymorphicQualifierTop => getPolymorphicQualifierElement.

mernst:
> > getPolymorphicQualifierElement => getPolymorphicQualifierTop
> 
> This is backward. You renamed getPolymorphicQualifierTop => getPolymorphicQualifierElement.

Thanks.  I fixed the comment.

PR #2454, opened 2019-04-25
Set fields in declaration initializers

mernst:


checker/src/main/java/org/checkerframework/checker/signedness/SignednessAnnotatedTypeFactory.java
```diff
@@ -29,21 +29,19 @@
...
      * Provides a way to query the Constant Value Checker, which computes the values of expressions
      * known at compile time (constant propagation and folding).
      */
-    private ValueAnnotatedTypeFactory getValueAnnotatedTypeFactory() {
-        if (valueAtypefactory == null) {
-            valueAtypefactory = getTypeFactoryOfSubchecker(ValueChecker.class);
-        }
-        return valueAtypefactory;
-    }
+    private ValueAnnotatedTypeFactory valueAtypefactory =
```
smillst: `getTypeFactoryOfSubchecker` must be called each time the factory is used.  See the [Javadoc](https://checkerframework.org/api/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.html#getTypeFactoryOfSubchecker-java.lang.Class-) for an explanation of why.

mernst: Thanks.
I saw that Javadoc comment, but this particular factory is used in a very specific way, and the current code already stores the factory in a field.
If storing the factory is a bad practice, that is the current code that [stores the factory in a field](https://github.com/typetools/checker-framework/pull/2454/files#diff-a0cb83447f10ece39c1992a3bca590abL43) buggy, and should it be changed?

smillst: It's possible that no bugs will be caused by this particular use of a sub-factory; however, it could and it would be hard to figure out that this is the problem.  (Also, I made this comment on the original pull request Signedness Checker pull request (https://github.com/typetools/checker-framework/pull/1364#discussion_r124578528))

mernst: OK, I have undone this.  Sorry that Chris and I missed that comment on the original pull request.

Issue #2406, opened 2019-04-04
Build failed in Intellij


aagarwal1012:
The following error I got after building the project for `Nullness` checker in `Intellij IDEA, Windows 10 64-bit`
```
Error:java: InvocationTargetException when invoking constructor for class org.checkerframework.checker.nullness.KeyForAnnotatedTypeFactory; Underlying cause: Bad Alias: null false null null
  Exception: java.lang.reflect.InvocationTargetException; Stack trace: sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
  sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
  sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
  java.lang.reflect.Constructor.newInstance(Constructor.java:423)
  org.checkerframework.common.basetype.BaseTypeChecker.invokeConstructorFor(BaseTypeChecker.java:258)
  org.checkerframework.common.basetype.BaseTypeVisitor.createTypeFactory(BaseTypeVisitor.java:261)
  org.checkerframework.common.basetype.BaseTypeVisitor.<init>(BaseTypeVisitor.java:219)
  org.checkerframework.common.basetype.BaseTypeVisitor.<init>(BaseTypeVisitor.java:208)
  org.checkerframework.common.basetype.BaseTypeChecker.createSourceVisitor(BaseTypeChecker.java:207)
  org.checkerframework.common.basetype.BaseTypeChecker.createSourceVisitor(BaseTypeChecker.java:83)
  org.checkerframework.framework.source.SourceChecker.initChecker(SourceChecker.java:858)
  org.checkerframework.common.basetype.BaseTypeChecker.initChecker(BaseTypeChecker.java:98)
  org.checkerframework.common.basetype.BaseTypeChecker.initChecker(BaseTypeChecker.java:89)
  org.checkerframework.framework.source.SourceChecker.typeProcessingStart(SourceChecker.java:814)
  org.checkerframework.javacutil.AbstractTypeProcessor$AttributionTaskListener.finished(AbstractTypeProcessor.java:157)
  com.sun.tools.javac.api.ClientCodeWrapper$WrappedTaskListener.finished(ClientCodeWrapper.java:681)
  com.sun.tools.javac.api.MultiTaskListener.finished(MultiTaskListener.java:111)
[...]

PR #2471, opened 2019-05-02
Update the IntelliJ instructions.

smillst:
Fixes https://github.com/typetools/checker-framework/issues/2406.

### DIFF
--- a/docs/manual/external-tools.tex
+++ b/docs/manual/external-tools.tex
@@ -395,8 +395,8 @@ then always use \<./gradlew> where you would have used \<gradle>.
 
 This section tells you how to make IntelliJ IDEA automatically run a
 checker on every compile for Java projects and/or modules.  If your project
-uses a build tool such as Gradle, follow the instructions
-for that build tool.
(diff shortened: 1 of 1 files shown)


docs/manual/external-tools.tex
```diff
@@ -391,63 +391,50 @@
...
 \sectionAndLabel{IntelliJ IDEA}{intellij}
 
-% The following method has been tested with IntelliJ IDEA 15.0.3.
+% The following method has been tested with IntelliJ IDEA 2019.1.1.
 
 This section tells you how to make IntelliJ IDEA automatically run a
-checker on every compile.  If you want to run the checker from a terminal
-within IntelliJ IDEA, follow the command-line instructions in
-Section~\ref{javac-installation}.
+checker on every compile for Java projects and/or modules.  If your project
```
mernst: Is this "if" telling the user something they should do instead, or something they should do in addition?
Does it enable type-checking on every compile?

smillst: These instructions do not work if the project uses a build tool.  I updated the text.

PR #2472, opened 2019-05-02
Move code out of InheritedFromClassAnnotator to correct location.

smillst:


framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
```diff
@@ -72,10 +75,30 @@ public AnnotatedTypeMirror visitMethod(MethodTree node, AnnotatedTypeFactory f)
         result.setElement(elt);
 
         ElementAnnotationApplier.apply(result, elt, f);
+        new MethodTypeVarAnnotator().visit(result, f);
 
         return result;
     }
 
+    /**
+     * Annotates uses of method type variables with annotation written explicity on the type
```
wmdietl: Isn't an annotation explicitly on a type parameter declaration an annotation on the lower bound?
Why is this needed only for method type parameters?

smillst: >Isn't an annotation explicitly on a type parameter declaration an annotation on the lower bound?

I meant in both the upper and lower bound locations.  I'll clarify the comment.

>Why is this needed only for method type parameters?

This actually apply to class type parameters, too.  I renamed the class and added a test case.

checker/tests/nullness/ExplictTypeVarAnnos.java
```diff
@@ -0,0 +1,22 @@
...
+        throw new RuntimeException();
+    }
+
+    public static <C extends @Nullable Object> Consumer<C> getConsumer() {
+        Consumer<@Nullable Object> nullConsumer = null;
+        Consumer<C> result = ExplictTypeVarAnnos.<C>cast(nullConsumer);
+        return result;
+    }
+
+    public Consumer<E> getConsumer2() {
```
wmdietl: From the class name and bug fix description I expected an explicit annotation on a type variable declaration - but I don't see one, neither on a declaration nor on a use. Am I missing something?

smillst: The explicit annotation is on the declaration of `E`.

wmdietl: So I thought this was talking about `class C<@Anno T>` - an explicit annotation on the declaration of the type parameter.
Here `E` only has an upper bound. So why call this an "explicit type var annotation"?
To test the lower bound logic, wouldn't you also want an annotation on the type parameter declaration?

smillst: I meant explicit annotation on the upper or lower bound of a type parameter declaration.  I added tests cases with explicit annotations on the lower bound.

framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
```diff
@@ -72,10 +75,30 @@ public AnnotatedTypeMirror visitMethod(MethodTree node, AnnotatedTypeFactory f)
         result.setElement(elt);
 
         ElementAnnotationApplier.apply(result, elt, f);
+        new TypeVarAnnotator().visit(result, f);
```
wmdietl: It seems a bit strange that this is only used in `visitMethod`.
Could you add a test case outside of a method, e.g. in a field initializer or in a superclass instantiation?
In the original code, all type variables were checked. Now only ones in method declarations are checked. Or am I missing where else this is used?

wmdietl: Instead of creating a new `TypeVarAnnotator` every time a method is visited, couldn't you create a single instance and then call `reset()` before every usage?

smillst: All uses of type parameters must be in the class that declared them (including in an enclosing class).  For all uses, except those in a method declaration, the type of the type variable is computed using `TypeFromTree#fromTypeTree`.  `org.checkerframework.framework.type.TypeFromTypeTreeVisitor#forTypeVariable` is called for all uses of type variables and that method looks up the annotations on type parameters and type parameter upper bounds from the tree.  Types in method signatures aren't created using `TypeFromTree#fromTypeTree`, so it needs special handling.  (I tried calling `TypeFromTree#fromTypeTree` on all the type trees in a method signature, but it caused the framework to crash because recursive type variables were not initialized as expected.)   (I did add a test case for fields.)

My main goal with this pull request is to move this code out of `InheritedFromClassAnnotator`, so that I can remove and replace that class in a later pull request.  Since the code in `InheritedFromClassAnnotator#visitTypeVariable` has no documentation and was added in the initial Checker Framework commit, I'm just guessing.   If you don't find the explanation in the above paragraph convincing, then I'm ok with moving this code so that it applies to all type variables again.  (Though, I'm not really sure where that would be.)

smillst: Yes, I'll do that once the above discussion is settled.

PR #2473, opened 2019-05-02
TypeAnnotator: don't skip the receiver

smillst:


checker/src/main/java/org/checkerframework/checker/interning/InterningAnnotatedTypeFactory.java
```diff
@@ -151,6 +152,16 @@ public Void visitDeclared(AnnotatedDeclaredType t, Void p) {
 
             return super.visitDeclared(t, p);
         }
+
+        @Override
+        public Void visitExecutable(AnnotatedExecutableType type, Void p) {
+            scan(type.getReturnType(), p);
+            // TODO: don't skip the receiver
```
smillst: This will be fixed when I implement the new semantics for the Interning Checker.

wmdietl: Shouldn't this be `TODO: skip the receiver`? Or am I missing the receiver here?

wmdietl: D'oh. Misread this.

PR #2475, opened 2019-05-03
Add text to the manual about where to put a custom checker

kelloggm:
With an example! The example is here: https://github.com/typetools/templatefora-checker


mernst:
This is blocked on https://github.com/typetools/templatefora-checker/pull/1

PR #2476, opened 2019-05-03
 Sets the null literal and uses of Void to bottom if no other implicit is given

smillst:
This is so that `@ImplicitFor(NULL)` need not be written on every bottom qualifier.

I added some comments pointing out the most important changes.

Note, this is not a request to merge into `master`, but rather a development branch `defaultyChanges`.  I intend to pull request several changes into that branch and merge them all into master at once.  This way all the "defaulty" changes will be in the same release, but they won't need to be reviewed as one massive change.  (The changes to upper bounds on type uses should be in the same release, too.)


framework/src/main/java/org/checkerframework/framework/type/treeannotator/ImplicitsTreeAnnotator.java
```diff
@@ -114,6 +114,33 @@ public ImplicitsTreeAnnotator(AnnotatedTypeFactory atypeFactory) {
         }
     }
 
+    /**
```
smillst: This code is new!

docs/manual/creating-a-checker.tex
```diff
@@ -444,7 +444,6 @@
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
-import org.checkerframework.framework.qual.ImplicitFor;
```
smillst: Here are some changes in the manual.

docs/manual/creating-a-checker.tex
```diff
@@ -713,12 +709,11 @@
 --- a qualifier that is a (direct or indirect) subtype of every other
 qualifier.
 
-Your type system must give \<null> the bottom type.
+\<null> is the bottom type.
```
wmdietl: A note about `java.lang.Void` would also be useful. In many of the places where you removed `@ImplicitFor` it was `@ImplicitFor(literals = {LiteralKind.NULL}, typeNames = {Void.class})`. Is this change implicitly applying bottom to both?

docs/manual/subtyping-checker.tex
```diff
@@ -8,9 +8,7 @@
 
 The Subtyping Checker can accommodate all of the type system enhancements that
 can be declaratively specified (see Chapter~\ref{creating-a-checker}).
-This includes type introduction rules (implicit
-annotations, e.g., literals are implicitly considered \refqualclass{checker/nullness/qual}{NonNull}) via
-the \refqualclass{framework/qual}{ImplicitFor} meta-annotation, and other features such as
+This includes type introduction rules meta-annotation, and other features such as
```
wmdietl: The first part of this sentence sounds a bit strange. Should it be `type-introduction-rule meta-annotations`?

framework/src/main/java/org/checkerframework/framework/type/treeannotator/ImplicitsTreeAnnotator.java
```diff
@@ -114,6 +114,33 @@ public ImplicitsTreeAnnotator(AnnotatedTypeFactory atypeFactory) {
...
+     * Adds standard implicit. Currently sets the null literal to bottom if no other implicit is set
```
wmdietl: Same for `java.lang.Void`?

wmdietl: First sentence seems a bit chopped off.

framework/tests/all-systems/GenericNull.java
```diff
@@ -1,6 +1,6 @@
 // The test cases GenericNull, FieldAccess, and InferTypeArgs often fail together.
 // Here are some things that might be wrong if they are failing.
-//  * Did you write @ImplicitFor(literals = LiteralKind.NULL) on the bottom type in your type
+//  * Did you write  on the bottom type in your type
```
wmdietl: This sentence is missing something.

docs/manual/creating-a-checker.tex
smillst: I only made `null` bottom.  I think `Void` used to be the type of type variable lower bounds so it needed to be implicitly bottom.  But that's not true anymore.  

Is there some other reason for Void to be bottom? If so, I can add that reason here and make it bottom.

wmdietl: The only value of the `Void` type is `null`. It is therefore most precise for both of them to have the same type. This has nothing to do with lower bounds. `Void` can also be used elsewhere.

For the Nullness Checker:

```java
  void foo(Void p) {
    p.toString();
  }
```

By making `Void` `@Nullable` one get's a useful error on the dereference.

smillst: Thanks for the explanation.  I made this change.

smillst:
@wmdietl You approved this pull request, but then I added the "standard" implicit for Void.  Do you want to take another look?

Issue #16, opened 2015-07-03
Handling Intersection types


GoogleCodeExporter:
```
The Checker Framework doesn't parse intersection bound type annotations (e.g. ? 
extends @A 
Number & @B Cloneable).
```

Original issue reported on code.google.com by `mahm...@notnoop.com` on 17 Nov 2009 at 3:16

PR #2481, opened 2019-05-10
Add nullness annotations to java.util.logging.LogRecord

the1derer:
This pull request is intended to add nullness annotations to java .util.logging.LogRecord in annotated JDK.
I have already added the Test case.
I encountered several problems including this while doing my Case Study of slf4j (there I had to use stub files)
Relies on [#16](https://github.com/typetools/annotated-libraries/pull/16)


checker/jdk/nullness/src/java/util/logging/LogRecord.java
```diff
@@ -0,0 +1,592 @@
...
+     * our promise to keep threadIDs unique by avoiding collisions due
+     * to 32-bit wraparound.  Unfortunately, LogRecord.getThreadID()
+     * returns int, while Thread.getId() returns long.
+     */
+    private static final int MIN_SEQUENTIAL_THREAD_ID = Integer.MAX_VALUE / 2;
+
+    private static final AtomicInteger nextThreadId
+        = new AtomicInteger(MIN_SEQUENTIAL_THREAD_ID);
+
+    @SuppressWarnings("type.argument.type.incompatible")
```
mernst: 1. Please always write an explanation along with each `@SuppressWarnings` annotation.
2. This suppression is incorrect.  The value may be null, so the field should be declared as `ThreadLocal<@Nullable Integer>`.  On [line 160](https://github.com/typetools/checker-framework/pull/2481/files#diff-3f0e5215c54358445cecd19a24665a8bR160) you can see the code check whether the Integer is null.

checker/jdk/nullness/src/java/util/logging/LogRecord.java
```diff
@@ -0,0 +1,592 @@
...
+     * @serialData Default fields, followed by a two byte version number
+     * (major byte, followed by minor byte), followed by information on
+     * the log record parameter array.  If there is no parameter array,
+     * then -1 is written.  If there is a parameter array (possible of zero
+     * length) then the array length is written as an integer, followed
+     * by String values for each parameter.  If a parameter is null, then
+     * a null String is written.  Otherwise the output of Object.toString()
+     * is written.
+     */
+    @SuppressWarnings("dereference.of.nullable") // parameters and parameters[i] is always ensured to be non-null at every dereference
```
mernst: 1. Anytime you suppress a `dereference.of.nullable` warning, you are saying that the value being dereferenced is definitely null.  Therefore, this comment does not provide any information, except to indicate that the concern is with `parameters` and `parameters[i]`.  Please be more specific.
2. The comment is misleading in that there is no warning about `parameters[i]` possibly being null.  The warning is only about `parameters`.

I would write the comment more like this:
```
 // out.writeInt and out.writeObject do not affect parameters field. http://tinyurl.com/cfissue/984
```

checker/jdk/nullness/src/java/util/logging/LogRecord.java
```diff
@@ -0,0 +1,592 @@
...
+        for (int i = 0; i < parameters.length; i++) {
+            if (parameters[i] == null) {
+                out.writeObject(null);
+            } else {
+                out.writeObject(parameters[i].toString());
+            }
+        }
+    }
+
+    @SuppressWarnings({"dereference.of.nullable" , "argument.type.incompatible"}) // method first initializes parameters before its dereference and ClassLoader.getSystemClassLoader() can't be @Nullable here
```
mernst: 1. The comment is misleading.  The Nullness Checker handles the initialization correctly and no warning suppression is needed.  The issue is that the Nullness Checker can't determine that `parameters` *stays* non-null.
Therefore, I would write the comment more like
``` // in.readObject does not affect parameters field.  http://tinyurl.com/cfissue/984```

2. How do you know that `ClassLoader.getSystemClassLoader()` can't be null?  Please explain that.

3. Minor:  "can't be `@Nullable`" is incorrect terminology, since the expression does have `@Nullable` type.  You want to claim that it also has `@NonNull` type, which is a more specific fact.

File: checker/tests/nullness/LogRecordTest.java [hunk 4]
```diff
@@ -0,0 +1,31 @@
+// Test Case for adding nullness annotations to java.util.logging.LogRecord
+
+import java.util.logging.Level;
+import java.util.logging.LogRecord;
+
+public class LogRecordTest {
+
+    @SuppressWarnings(
```
mernst: You can avoid the need for this warning suppression by making `level` a parameter to test, as in:

```
void test(Level level) {
```

the1derer: Corrected.
Thanks for all your suggestions.
I will remember to follow them in the future.

checker/jdk/nullness/src/java/util/logging/LogRecord.java
the1derer: > 2. How do you know that `ClassLoader.getSystemClassLoader()` can't be null?  Please explain that.

My reasoning is that all the logging framework will definitely have a ClassLoader which will also act as SystemClassLoader. I now have mentioned this in comments too.
I could be wrong.
Please let me know if my reasoning is wrong because without `ClassLoader.getSystemLoader()` being `@NonNull` it would straight up produce `NullPointerException`.

checker/jdk/nullness/src/java/util/logging/LogRecord.java
the1derer: Corrected.
Thanks for the suggestion. I will try to be more precise in the future.

checker/jdk/nullness/src/java/util/logging/LogRecord.java
the1derer: Apologies for the negligence.
I was frustrated because of tests not passing because of another error and `@SuppresseWarnings` in a hurry.
I will definitely take care of this in the future.

the1derer:
@mernst I have made the changes according to your suggestions.
Could you have a look at the changes?
Thanks.

checker/build.gradle
```diff
@@ -2,7 +2,7 @@ ext {
     // See instructions for updating jdk8.jar at
     // https://github.com/typetools/annotated-libraries/blob/master/README.md .
     // Set jdkShaHash to 'local' to use a locally-built version of jdk8.jar .
-    jdkShaHash = 'c47e5da9b2b1c8627981c97b75ec0d4cbd7ee052'
+    jdkShaHash = 'cf9185c03d6a63ad28f7dd51615308beeff7a64a'
```
mernst: A [change](https://github.com/typetools/annotated-libraries/commit/c703b5750e37538731189bdce45ca80b67986f22) was made to https://github.com/typetools/annotated-libraries since you made your [pull request](https://github.com/typetools/annotated-libraries/pull/16) against it, and also I [changed](https://github.com/typetools/checker-framework/commit/0b77004b1b2f999d40e64d7414378d3e1796585f) a library annotation (see explanation below).  Could you please update that pull request and this SHA hash?

checker/jdk/nullness/src/java/util/logging/LogRecord.java
```diff
@@ -0,0 +1,593 @@
...
+            if (parameters[i] == null) {
+                out.writeObject(null);
+            } else {
+                out.writeObject(parameters[i].toString());
+            }
+        }
+    }
+
+    @SuppressWarnings({"dereference.of.nullable", "argument.type.incompatible"}) // in.readObject does not affect parameters field. http://tinyurl.com/cfissue/984
+    // and ClassLoader.getSystemClassLoader() is @NonNull because  there will always be a SystemClassLoader present to load the classes of logging framework
```
mernst: The comment about `ClassLoader.getSystemClassLoader()` should appear on [line 536](https://github.com/typetools/checker-framework/pull/2481/files#diff-3f0e5215c54358445cecd19a24665a8bR536), not here.
Please always write `@SuppressWarnings` in the [smallest scope possible](https://checkerframework.org/manual/#suppresswarnings-annotation-locations).
The `readObject` issue is more pervasive in the method, and its warning suppression can be on the entire method.

mernst: I agree that there will always be a system class loader.  That does not imply that getSystemClassLoader will not return null in Java 8.

However, Java 9 has changed the contract of getSystemClassLoader so that it does not return null (it throws an exception instead; compare [JDK 8 Javadoc](https://docs.oracle.com/javase/8/docs/api/java/lang/ClassLoader.html#getSystemClassLoader--) with [JDK 9 Javadoc](https://docs.oracle.com/javase/9/docs/api/java/lang/ClassLoader.html#getSystemClassLoader--)).  

Therefore, I have [changed](https://github.com/typetools/checker-framework/commit/0b77004b1b2f999d40e64d7414378d3e1796585f) the annotation on its return type.  It isn't strictly true for the JDK that is being annotated, but we are switching to Java 9 soon, so I think it's OK.  This means you can remove the `@SuppressWarnings` related to `getSystemClassLoader()` here.

It's interesting that investigating a warning can lead to understanding of the JDK and its changes, and can lead to changed JDK annotations elsewhere.

checker/build.gradle
the1derer: I have updated the pull request and SHA hash.

checker/jdk/nullness/src/java/util/logging/LogRecord.java
the1derer: > I agree that there will always be a system class loader. That does not imply that getSystemClassLoader will not return null in Java 8.

I agree that it doesn't imply getSystemClassLoader will not return null in Java8, but since this class will be mostly used by libraries like slf4j and they will definitely have a ClassLoader(if they don't they will run right into NullPointerExceptiion).

>It's interesting that investigating a warning can lead to understanding of the JDK and its changes, and can lead to changed JDK annotations elsewhere.

Indeed, I didn't think that this small PR will lead to this interesting discovery. I  think Oracle just thought of the same thing and changed the contract in Java9.

@mernst Thanks for your help and all your valuable inputs.

the1derer:
@mernst I have implemented the changes according to your suggestions.
Could you have a look at the changes?
Thanks.

checker/build.gradle
mernst: The [pull request](https://github.com/typetools/annotated-libraries/pull/16) still shows "This branch has conflicts that must be resolved".  Can you please pull master into your branch?  That is the way to update the pull request.  Thanks.

mernst:
> Why does [Third Travis test](https://travis-ci.org/typetools/checker-framework/jobs/534212062) keeps failing?

There was an error in master.  I fixed it and have pulled master into your pull request, so this should work again now.

checker/build.gradle
the1derer: @mernst I have removed the conflict. Please review.
Thanks.

the1derer:
@mernst I have removed the merge conflict from "annotated-libraries" PR. Also, I have pulled master into this PR and Updated the jdkShaHash
Can you please review again.
Thanks.

mernst:
Thanks for the JDK annotations!

the1derer:
> Thanks for the JDK annotations!

Thanks for your time, patience, and valuable suggestions.

PR #2486, opened 2019-05-19
Add method Store.toStringVerbose

mernst:
Sometimes it is helpful to be able to print the contents of arbitrary stores, not just those in the CFG visualization.  This is a slightly hacky way to do that, but it works.


wmdietl:
@mernst I recently also ran into such an issue and discussed a more broad refactoring with @xingweitian.
He prepared a first pull request at: https://github.com/opprop/checker-framework/pull/89
I still want to do a few iterations on that PR (in person with Weitian), but hope to have this ready by next week.
So if this is urgent, I'm fine with merging this PR and then later replacing the hackiness with the (hopefully) nicer alternative.
If this isn't urgent, we could wait until the other PR is ready to review here.

mernst:
I can wait; this is not urgent.
Thanks for making the more extensive refactoring.  If I had known about that, I wouldn't have reinvented the wheel.

wmdietl:
Closed in favor of #2551.

PR #2490, opened 2019-05-20
Handle (some) offsets in `@LessThan` annotations

mernst:


framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -3951,4 +3953,64 @@ public ProcessingEnvironment getProcessingEnv() {
...
+     * "a" => <"a", "0">
+     * "a + 5" => <"a", "5">
+     * "a + -5" => <"a", "-5">
+     * "a - 5" => <"a", "-5">
+     * }</pre>
+     *
+     * There are methods that can only take as input an expression that represents a Receiver. The
+     * purpose of this is to pre-process expressions to make those methods more likely to succeed.
+     *
+     * @param expression and expression to remove a constant offset from
```
kelloggm: "an expression"?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -3951,4 +3953,64 @@ public ProcessingEnvironment getProcessingEnv() {
...
+     * @return a sub-expression and a constant offset
+     */
+    // TODO: generalize.  There is no reason this couldn't handle arbitrary addition and subtraction
+    // expressions, given the Index Checker's support for OffsetEquation.  That might even make its
+    // implementation simpler.
+    public static Pair<String, String> getExpressionAndOffset(String expression) {
+        String expr = expression;
+        // Should the default value be null instead of "0"?
+        String offset = "0";
```
kelloggm: The default value should be `"0"`. There are two cases where this method is called and the default value is returned:
* `expression` is unparseable, and should be defaulted to top, or
* `expression` is parseable but does not contain a constant offset

In the first case, the return value of this method doesn't matter - it will be ignored anyway. The only reason to prefer `null` would be to throw a runtime exception closer to the problem if that value is used.

In the second case, `"0"` matches the current default behavior - which is to parse the expression and apply an offset of zero to it. Adding null-checking to that behavior seems unnecessary, and will make the code much more complex.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -3951,4 +3953,64 @@ public ProcessingEnvironment getProcessingEnv() {
...
+
+        Matcher mPlus = plusConstant.matcher(expr);
+        Matcher mMinus = minusConstant.matcher(expr);
+        if (mPlus.find()) {
+            expr = expr.substring(0, mPlus.start());
+            offset = mPlus.group(1);
+        } else if (mMinus.find()) {
+            expr = expr.substring(0, mMinus.start());
+            offset = mMinus.group(1);
+            if (offset.startsWith("-")) {
```
kelloggm: Could this share code with the `negate()` function you wrote above?

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -816,6 +816,22 @@ public Receiver getReceiverFromJavaExpressionString(String expression, TreePath
         return FlowExpressionParseUtil.parse(expression, context, currentPath, true);
     }
 
+    /**
+     * Produces the receiver and offset associated with an expression. For instance, "n+1" has no
+     * associated Receiver, but it this produces a pair of a Receiver (for n) and an offset ("1").
```
kelloggm: "it this"

mernst:
@kelloggm Sorry the pull request was unclear.
I thought that the title and the [tests](https://github.com/typetools/checker-framework/pull/2490/files#diff-9cb4e62c85d6191b2fced113d16857fcR1) (all of which currently fail) would explain it, but I was wrong.

There is no `@LessThanOrEqualTo` annotation, leading to the need to write `@LessThan("othervar + 1")`.  Currently, the Index Checker does not handle that.  This pull request makes the Index Checker handle it.

This comes up in Guava, for example.

PR #2491, opened 2019-05-20
Use value from store, not from tree

mernst:
This is necessary because a pre-increment operation might have changed the value in the store.

Merge this or #2492, not both.


smillst:
Closed in favor of #2492.

PR #2492, opened 2019-05-20
Use value from store, not from tree

mernst:
This is necessary because a pre-increment operation might have changed the value in the store.

This is an alternative to #2491:  only one of them should be merged.


framework/src/main/java/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -1215,11 +1215,18 @@ private void createAnnotationFromRangeAndAddToStore(CFStore store, Range range,
...
-                    atypefactory
-                            .getAnnotatedType(internal.getTree())
-                            .getAnnotationInHierarchy(atypefactory.BOTTOMVAL);
             Receiver rec = FlowExpressions.internalReprOf(analysis.getTypeFactory(), internal);
+            CFValue currentValueFromStore;
+            if (store.canInsertReceiver(rec)) {
+                currentValueFromStore = store.getValue(rec);
+            } else {
+                // Don't just `continue;` which would skip the calls to refine{Array,String}...
+                currentValueFromStore = null;
```
smillst: This seems like the right choice just in case there isn't a value in the store.

Issue #2215, opened 2018-11-26
Permit assignment to @MonotonicNonNull in constructor


mernst:
Every possibly-nullable assignment to a `@MonotonicNonNull` field is prohibited.
However, such an assignment should be permitted, as the first assignment in a constructor.
For example, the Nullness Checker currently rejects the below code, but it should be permitted.

```java
import org.checkerframework.checker.nullness.qual.Nullable;
import org.checkerframework.checker.nullness.qual.MonotonicNonNull;

public class InitializeMonotonicNonNull {

  @MonotonicNonNull Object f;

  InitializeMonotonicNonNull(@Nullable Object f) {
    this.f = f;
  }

}
```

the1derer:
Hey @mernst I would like to work on this.I came across similar problem during my Case Study of slf4j in following code-type:
```java
import org.checkerframework.checker.nullness.qual.EnsuresNonNull;
import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
import org.checkerframework.checker.nullness.qual.Nullable;

class A {
  @MonotonicNonNull String field1;
  
  A(String formalArgument) {
    this.field1 = null;
  }

  @EnsuresNonNull("field1")
  void initialize(String formalArgument) {
    field1 = formalArgument;
  }

}
```
Can you help me in getting started, like which files to look at ?
Thanks.

PR #2495, opened 2019-05-20
Test Case for #2215

the1derer:
Related to #2215  
Test Case is Ready.
@mernst  Does the Test Case looks good?


mernst:
I overlooked this because it's marked as work in progress rather than ready for review.

To answer your question about the test case:
It's good that the test case is in the right directory.
The test case itself is inadequate, however.  It doesn't have nearly as much detail as the [issue](https://github.com/typetools/checker-framework/issues/2215).  Therefore, a patch that passes the given test case might still be incorrect.
A good test case should also include tests that involve fields in a superclass.

the1derer:
@mernst 
Ready for review

checker/tests/nullness/Issue2215.java
```diff
@@ -0,0 +1,34 @@
+// Test Case for Issue2215
+// https://github.com/typetools/checker-framework/issues/2215
+// @skip-test
```
mernst: It is good to say why the test is skipped, such as
```@skip-test until the bug is fixed```

File: checker/tests/nullness/Issue2215.java [hunk 2]
```diff
@@ -0,0 +1,34 @@
...
+
+import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+class ParentClass {
+
+    @MonotonicNonNull Object parentClassField;
+
+    ParentClass(@Nullable Object o) {
+        this.parentClassField = o; // should be allowed
```
mernst: You don't need "should be allowed" text.  Anything that is not preceded by a `// ::` comment is allowed, and this text clutters the test case.

checker/tests/nullness/Issue2215.java
```diff
@@ -0,0 +1,34 @@
...
+
+    Issue2215(@Nullable Object o) {
+
+        super(o); // should be allowed
+        this.parentClassField = o; // should be allowed
+        this.f = o; // should be allowed
+        this.f = null; // should be allowed
+        this.g = o; // should be allowed
+        this.g = null; // should be allowed
+        this.h = o; // should not be allowed
```
mernst: This is not how an error is indicated in a Checker Framework test case.
See any existing test case, or see https://raw.githubusercontent.com/typetools/checker-framework/master/checker/tests/README .

the1derer:
Thanks for correcting me. I will keep all feedback in mind in the future.
I have made suggested changes. Please have a look.

Thanks.

checker/tests/nullness/Issue2215.java
```diff
@@ -0,0 +1,36 @@
...
+    @MonotonicNonNull Object f;
+    @MonotonicNonNull Object g = null;
+    @MonotonicNonNull Object h = new Object();
+
+    Issue2215(@Nullable Object o) {
+
+        super(o);
+        this.parentClassField = o;
+        this.f = o;
+        this.f = null;
```
mernst: I think this should be an error.  It might set a field from a non-null value to a null value.
The issue in the issue tracker already gets this case right.
I think we should close this pull request, because we have spent a lot of time iterating, mostly just trying to get to what is already in the example in the pull request.  (I just added the supertype to it.)

the1derer: > I think this should be an error. It might set a field from a non-null value to a null value.
> The issue in the issue tracker already gets this case right.

I agree with you.

Thank you very much for correcting the Example in the Issue tracker.

PR #2496, opened 2019-05-20
Add @QualifierForLiterals meta-annotation

smillst:
(This is the next part of the default changes.)
1. Adds `@QualifierForLiterals` meta-annotation
2. Moves `@ImplicitFor(literals, stringpatterns)` to the `@QualifierForLiterals`
2. Renames `ImplicitsTreeAnnotator` to `LiteralTreeAnnotator`

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/i18n/qual/Localized.java
+++ b/checker/src/main/java/org/checkerframework/checker/i18n/qual/Localized.java
@@ -5,8 +5,8 @@ import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
-import org.checkerframework.framework.qual.ImplicitFor;
 import org.checkerframework.framework.qual.LiteralKind;
+import org.checkerframework.framework.qual.QualifierForLiterals;
 import org.checkerframework.framework.qual.SubtypeOf;
 
 /**
@@ -16,21 +16,20 @@ import org.checkerframework.framework.qual.SubtypeOf;
  * @checker_framework.manual #i18n-checker Internationalization Checker
  */
 @SubtypeOf(UnknownLocalized.class)
-@ImplicitFor(
-        literals = {
-            /* All integer literals */
-            LiteralKind.INT,
-            LiteralKind.LONG,
-            LiteralKind.FLOAT,
-            LiteralKind.DOUBLE,
-            LiteralKind.BOOLEAN,
+@QualifierForLiterals({
+    /* All integer literals */
+    LiteralKind.INT,
+    LiteralKind.LONG,
+    LiteralKind.FLOAT,
+    LiteralKind.DOUBLE,
+    LiteralKind.BOOLEAN,
 
-            /* null should be the bottom type */
-            LiteralKind.NULL
+    /* null should be the bottom type */
+    LiteralKind.NULL
 
-            // CHAR_LITERAL,
-            // STRING_LITERAL,
-        })
+    // CHAR_LITERAL,
+    // STRING_LITERAL,
+})
 @Documented
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.TYPE_USE, ElementType.TYPE_PARAMETER})
--- a/checker/src/main/java/org/checkerframework/checker/i18n/qual/Localized.java
+++ b/checker/src/main/java/org/checkerframework/checker/i18n/qual/Localized.java
@@ -17,18 +17,13 @@ import org.checkerframework.framework.qual.SubtypeOf;
  */
 @SubtypeOf(UnknownLocalized.class)
 @QualifierForLiterals({
-    /* All integer literals */
+    // All literals except chars and strings, which may need to be localized.
+    // (null is bottom by default.)
     LiteralKind.INT,
     LiteralKind.LONG,
     LiteralKind.FLOAT,
     LiteralKind.DOUBLE,
-    LiteralKind.BOOLEAN,
-
-    /* null should be the bottom type */
-    LiteralKind.NULL
-
-    // CHAR_LITERAL,
-    // STRING_LITERAL,
+    LiteralKind.BOOLEAN
 })
 @Documented
 @Retention(RetentionPolicy.RUNTIME)
--- a/checker/src/main/java/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
+++ b/checker/src/main/java/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
@@ -470,6 +470,7 @@ public class NullnessAnnotatedTypeFactory
 
         @Override
         public Void visitNewArray(NewArrayTree node, AnnotatedTypeMirror type) {
+            // The result of newly allocated structures is always non-null.
             type.replaceAnnotation(NONNULL);
 
             // The most precise element type for `new Object[] {null}` is @FBCBottom, but
--- a/checker/src/main/java/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
+++ b/checker/src/main/java/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
@@ -471,7 +471,9 @@ public class NullnessAnnotatedTypeFactory
         @Override
         public Void visitNewArray(NewArrayTree node, AnnotatedTypeMirror type) {
             // The result of newly allocated structures is always non-null.
-            type.replaceAnnotation(NONNULL);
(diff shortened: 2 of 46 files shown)


checker/src/main/java/org/checkerframework/checker/i18n/qual/Localized.java
```diff
@@ -16,21 +16,20 @@
...
+    LiteralKind.INT,
+    LiteralKind.LONG,
+    LiteralKind.FLOAT,
+    LiteralKind.DOUBLE,
+    LiteralKind.BOOLEAN,
 
-            /* null should be the bottom type */
-            LiteralKind.NULL
+    /* null should be the bottom type */
+    LiteralKind.NULL
```
wmdietl: Isn't this the default behavior now, even without writing anything?

checker/src/main/java/org/checkerframework/checker/i18n/qual/Localized.java
```diff
@@ -16,21 +16,20 @@
...
 
-            // CHAR_LITERAL,
-            // STRING_LITERAL,
-        })
+    // CHAR_LITERAL,
```
wmdietl: Should we just remove these comments? It's not clear to my why they are here. If you understand, can you add a comment? Later realized what this means and added comment to `All integer literals` above.

checker/src/main/java/org/checkerframework/checker/i18n/qual/Localized.java
```diff
@@ -16,21 +16,20 @@
...
-@ImplicitFor(
-        literals = {
-            /* All integer literals */
-            LiteralKind.INT,
-            LiteralKind.LONG,
-            LiteralKind.FLOAT,
-            LiteralKind.DOUBLE,
-            LiteralKind.BOOLEAN,
+@QualifierForLiterals({
+    /* All integer literals */
```
wmdietl: This comment is confusing, as below it includes floats and boolean.
I see that this is from before, but let's also clean this up.
I think it would be better to explicitly say that these are all primitives other than char and string, because the latter two might require localization.

docs/manual/creating-a-checker.tex
```diff
@@ -709,10 +709,10 @@
 --- a qualifier that is a (direct or indirect) subtype of every other
 qualifier.
 
-\<null> is the bottom type. Because the only value with type \<Void> is \<null>, uses of the type \<Void> are also bottom.
+\<null> is the bottom type.
```
wmdietl: Why did this part go away? Why should `Void` no longer be bottom?

docs/manual/subtyping-checker.tex
```diff
@@ -8,9 +8,7 @@
 
 The Subtyping Checker can accommodate all of the type system enhancements that
 can be declaratively specified (see Chapter~\ref{creating-a-checker}).
-This includes type introduction rules (implicit
-annotations, e.g., literals are implicitly considered \refqualclass{checker/nullness/qual}{NonNull}) via
-the \refqualclass{framework/qual}{ImplicitFor} meta-annotation, and other features such as
+This includes type introduction rules meta-annotation, and other features such as
```
wmdietl: I think it sounds better as `rule meta-annotations`, doesn't it?
Wouldn't a reference to `QualifierForLiterals` still be useful here?

framework/src/main/java/org/checkerframework/framework/qual/QualifierForLiterals.java
```diff
@@ -0,0 +1,28 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/** A meta-annotation that indicates what qualifier should be given to literals. */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.ANNOTATION_TYPE)
+public @interface QualifierForLiterals {
+    /**
+     * @return {@link LiteralKind}s for which an annotation should be implicitly added. For example,
```
wmdietl: Can we replace `implicitly` (here and below) with something else, e.g. `automatically`?

framework/src/main/java/org/checkerframework/framework/type/treeannotator/LiteralTreeAnnotator.java
```diff
@@ -13,39 +13,29 @@
...
- * annotations specified by {@link org.checkerframework.framework.qual.ImplicitFor} for any tree
- * whose visitor is not overridden or does not call {@code super}; it is designed to be invoked from
- * {@link
- * org.checkerframework.framework.type.AnnotatedTypeFactory#addComputedTypeAnnotations(javax.lang.model.element.Element,
- * org.checkerframework.framework.type.AnnotatedTypeMirror)} and {@link
- * org.checkerframework.framework.type.AnnotatedTypeFactory#addComputedTypeAnnotations(com.sun.source.tree.Tree,
- * org.checkerframework.framework.type.AnnotatedTypeMirror)}.
+ * Adds annotations to a type based on the contents of a tree. This class applies annotations
+ * specified by {@link org.checkerframework.framework.qual.QualifierForLiterals}; it is designed to
+ * be add to a {@link ListTreeAnnotator} via {@link
```
wmdietl: `add` -> `added`

framework/src/main/java/org/checkerframework/framework/type/treeannotator/LiteralTreeAnnotator.java
```diff
@@ -13,39 +13,29 @@
...
- *
- * <p>TODO: we currently don't check that any attribute is set, that is, a qualifier could be
- * annotated as @ImplicitFor(), which might be misleading.
- *
- * @see org.checkerframework.framework.type.typeannotator.ImplicitsTypeAnnotator
  * @see TreeAnnotator
  */
-public class ImplicitsTreeAnnotator extends TreeAnnotator {
+//  TODO: we currently don't check that any attribute is set, that is, a qualifier could be
+// annotated as @QualifierForLiterals(), which might be misleading.
```
wmdietl: Should `@QualifierForLiterals()` simply mean `@QualfierForLiterals(LiteralKind.ALL)`?

framework/src/main/java/org/checkerframework/framework/type/treeannotator/LiteralTreeAnnotator.java
```diff
@@ -98,30 +88,30 @@ public ImplicitsTreeAnnotator(AnnotatedTypeFactory atypeFactory) {
...
+            for (String pattern : forLiterals.stringPatterns()) {
                 addStringPattern(pattern, theQual);
             }
         }
     }
 
     /**
-     * Adds standard implicit rules. Currently sets the null literal to bottom if no other implicit
-     * is set for null literals. Also, see {@link ImplicitsTypeAnnotator#addStandardImplicits()}.
+     * Adds standard defaults. Currently sets the null literal to bottom if no other default is set
```
wmdietl: I'm not sure I like the `Defaults` in the name. A literal can never specify alternative annotations, so the qualifiers given here are the effective annotations on those literals.
How about `addStandardLiteralQualifiers`?

checker/src/main/java/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
```diff
@@ -474,6 +470,9 @@ public Void visitNewClass(NewClassTree node, AnnotatedTypeMirror type) {
 
         @Override
         public Void visitNewArray(NewArrayTree node, AnnotatedTypeMirror type) {
+            // The result of newly allocated structures is always non-null.
+            type.replaceAnnotation(NONNULL);
```
wmdietl: On master I tried

```java
  void foo(Object nn) {
    nn = new int @Nullable [5];
  }
```

and this gives an assignment error.
With this change it looks like the code would be quietly accepted.

I see two options:
- only add NONNULL if nothing is present yet, which is the old behavior; this isn't ideal, because a new array always is non-null. So as a follow-up issue we can validate those type better.
- raise an error here if there already is a qualifier other than NONNULL; this isn't ideal, as we don't like raising errors in the type factory.

docs/manual/subtyping-checker.tex
```diff
@@ -8,9 +8,8 @@
 
 The Subtyping Checker can accommodate all of the type system enhancements that
 can be declaratively specified (see Chapter~\ref{creating-a-checker}).
-This includes type introduction rules (implicit
-annotations, e.g., literals are implicitly considered \refqualclass{checker/nullness/qual}{NonNull}) via
-the \refqualclass{framework/qual}{ImplicitFor} meta-annotation, and other features such as
+This includes type introduction rules meta-annotation via the
```
wmdietl: Should the `meta-annotation` come after the `QualifierForLiterals`, as before?

PR #2498, opened 2019-05-21
 Polymorphic constructor results

smillst:
If a constructor result is annotated with a polymorphic qualifier, then treat explicit annotations
on new class trees as a constraint for the instantiation for the polymorphic qualifier.  For example:

```
   @PolyTainted PolyConstructor() {}

    static void uses(@Tainted Object tainted, @Untainted Object untainted) {
        @Untainted PolyConstructor o1 = new @Untainted PolyConstructor(); // legal
        @Tainted PolyConstructor o2 = new @Tainted PolyConstructor(); // legal
  }
```

### DIFF
--- a/docs/manual/advanced-features.tex
+++ b/docs/manual/advanced-features.tex
@@ -646,6 +646,9 @@ for the cast.  For either syntax, the Checker Framework will issue a
 warning that it cannot verify that the cast is correct.  The programmer may
 suppress the warning if the code is correct.
 
+One exception to this is if the annotation on the result type is polymorphic.  In this case, any
+explicit annotation between \<new> and a class name is treated as a constraint for the instantiation
+for the polymorphic qualifier.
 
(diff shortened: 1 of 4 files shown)


docs/manual/advanced-features.tex
```diff
@@ -646,6 +646,9 @@
 warning that it cannot verify that the cast is correct.  The programmer may
 suppress the warning if the code is correct.
 
+One exception to this is if the annotation on the result type is polymorphic.  In this case, any
```
smillst: @mernst Could you please review this?

docs/manual/advanced-features.tex
```diff
@@ -646,6 +646,9 @@
...
+explicit annotation between \<new> and a class name is treated as a constraint for the instantiation
```
wmdietl: I do not understand what constraint this generates. Can you expand a bit, e.g. by expanding the example in the description with an illegal use?

smillst:
Closed in favor of #2504.

PR #2501, opened 2019-05-22
Explicit annotations on type parameter upper bounds.

smillst:
The changes in #2472 caused daikon-type-check to fail. 
Here's the error:
```
daikon/diff/Node.java:22: error: [assignment.type.incompatible] incompatible types in assignment.
    this.userObject = userObject;
                      ^
  found   : @NonNull @NonRaw Pair<CONTENT[ extends @NonNull @NonRaw Object super @NonNull @NonRaw Void], CONTENT[ extends @NonNull @NonRaw Object super @NonNull @NonRaw Void]>
  required: @NonNull @NonRaw Pair<CONTENT[ extends @NonRaw @Nullable Object super @NonNull @NonRaw Void], CONTENT[ extends @NonRaw @Nullable Object super @NonNull @NonRaw Void]>
```
Here is the relevant snippet of the Node.java file:
```
public abstract class Node<CONTENT extends @Nullable Object, CHILD> {

  private List<CHILD> children = new ArrayList<CHILD>();
  // Nonsensical for RootNode
  private Pair<CONTENT, CONTENT> userObject;

  public Node(Pair<CONTENT, CONTENT> userObject) {
    this.userObject = userObject;
  }
```
The error only seems to happen when the tree for the element isn't available, so I'm not sure how to create a test case for this.   However, I have verified locally that this pull request fixes the bug.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
@@ -47,6 +47,7 @@ import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.Name;
 import javax.lang.model.element.TypeElement;
+import javax.lang.model.element.TypeParameterElement;
 import javax.lang.model.element.VariableElement;
 import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.IntersectionType;
@@ -397,6 +398,12 @@ public class AnnotatedTypeFactory implements AnnotationProvider {
     protected final ExecutableElement objectGetClass;
 
     /**
+     * Adds explicit annotations on type parameter declarations an their upper bounds to uses of
+     * type variables.
+     */
+    protected final TypeVarAnnotator typeVarAnnotator;
+
+    /**
      * Constructs a factory from the given {@link ProcessingEnvironment} instance and syntax tree
      * root. (These parameters are required so that the factory may conduct the appropriate
      * annotation-gathering analyses on certain tree types.)
@@ -462,6 +469,7 @@ public class AnnotatedTypeFactory implements AnnotationProvider {
         ignoreUninferredTypeArguments = !checker.hasOption("conservativeUninferredTypeArguments");
 
         objectGetClass = TreeUtils.getMethod("java.lang.Object", "getClass", 0, processingEnv);
+        this.typeVarAnnotator = new TypeVarAnnotator();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1180,6 +1188,7 @@ public AnnotatedTypeMirror fromElement(Element elt) {
                 && (typesFromStubFiles == null || !typesFromStubFiles.containsKey(elt))) {
             type = toAnnotatedType(elt.asType(), ElementUtils.isTypeDeclaration(elt));
             ElementAnnotationApplier.apply(type, elt, this);
+            typeVarAnnotator.visit(type);
```
wmdietl: Both places that use the `typeVarAnnotator` call `ElementAnnotationApplier.apply(type, elt, this);` right before. Conceptually, shouldn't this action already happen there?
Did you look why the `TypeVarUseApplier` doesn't already do this or whether you can move this additional logic there?

smillst: I've done this.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -396,6 +397,12 @@ void checkRep(String aliasName) {
     /** The Object.getClass method. */
     protected final ExecutableElement objectGetClass;
 
+    /**
```
wmdietl: Can't all the changes here now be moved to `ElementAnnotationApplier`?

framework/src/main/java/org/checkerframework/framework/type/ElementAnnotationApplier.java
```diff
@@ -69,6 +69,17 @@ public static void apply(
             final AnnotatedTypeMirror type,
             final Element element,
             final AnnotatedTypeFactory typeFactory) {
+        applyInternal(type, element, typeFactory);
+        // Also copy annotations from type parameters to their uses.
+        typeFactory.typeVarAnnotator.visit(type);
```
wmdietl: Couldn't the conditions from `TypeVarAnnotator` be merged into `applyInternal`?
It would be nice to understand and document when `TypeVarAnnotator` should be used and when `TypeVarUseApplier`.

wmdietl: Further to our in-person discussion: e.g. `TypeVarUseApplier` internally also creates a new visitor instance. So doing this for the `TypeVarAnnotator` doesn't seem too bad.
Eventually however we should try to merge those two.

PR #2504, opened 2019-05-24
Poly return types

smillst:
If a method return or constructor result is annotated with a polymorphic qualifier and the method
or constructor does not take any arguments, then that method or constructor should be treated as bottom.
This permits clients to treat the invocation as having any qualifier.


checker/tests/tainting/PolyReturn.java
```diff
@@ -0,0 +1,24 @@
+import org.checkerframework.checker.tainting.qual.PolyTainted;
+import org.checkerframework.checker.tainting.qual.Tainted;
+import org.checkerframework.checker.tainting.qual.Untainted;
+
+@SuppressWarnings({"inconsistent.constructor.type", "super.invocation.invalid"})
```
wmdietl: Could this be moved to the constructor, to limit the scope?
(A comment why these are suppressed would also be useful.)

framework/src/main/java/org/checkerframework/framework/type/poly/AbstractQualifierPolymorphism.java
```diff
@@ -285,11 +285,11 @@ protected Void scan(AnnotatedTypeMirror type, Void p) {
                     type.removeAnnotation(poly);
                     if (top == null) {
                         // poly is PolyAll -> add all tops not explicitly given
```
wmdietl: The comment should also be updated.

framework/src/main/java/org/checkerframework/framework/type/poly/AbstractQualifierPolymorphism.java
```diff
@@ -285,11 +285,11 @@ protected Void scan(AnnotatedTypeMirror type, Void p) {
...
-                        type.addMissingAnnotations(topQuals);
+                        type.addMissingAnnotations(qualHierarchy.getBottomAnnotations());
                     } else if (type.getKind() != TypeKind.TYPEVAR
                             && type.getKind() != TypeKind.WILDCARD) {
                         // Do not add the top qualifiers to type variables and wildcards
```
wmdietl: Comment needs update.

PR #2511, opened 2019-05-26
Rename skipParens to withoutParens

mernst:
Merge with https://github.com/typetools/checker-framework-inference/pull/110


mernst:
@smillst I had left out a changelog entry because it's just a deprecation rather than a removal.  I have now added the changelog entry.

PR #2513, opened 2019-05-26
Simplify primitiveTypeCast()

mernst:


checker/src/main/java/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -47,10 +47,6 @@ private PrimitiveTypeTree primitiveTypeCast(Tree node) {
 
         TypeCastTree cast = (TypeCastTree) node;
         Tree castType = cast.getType();
-        // All types should be wrapped in some annotation
-        if (castType.getKind() != Kind.ANNOTATED_TYPE) {
```
smillst: This is checking that castType is an AnnotatedTypeTree.   I've added a test case that crashes without this check.

mernst: @smillst Thanks!  I think I should undo the code change and merge the additional test cases.
Is that correct?

smillst: Yes.

PR #2517, opened 2019-05-27
Remove duplicate code from ControlFlowGraph.java.

xingweitian:
Hello, this PR removes the duplicate code from ControlFlowGraph.java. The removed code is the same as the content of getSuccessors(), so we can use this method directly.


xingweitian:
Travis failed. I think it is because that it tried to use `xingweitian/checker-framework-inference` which is forked from `opprop/checker-framework-inference`. 

It can be found at [here](https://travis-ci.org/typetools/checker-framework/jobs/537991138#L1498).

wmdietl:
Thanks for the simplification!

PR #2521, opened 2019-05-29
Remove ImplicitFor

smillst:
(This is a pull request into the defaultyChanges branch)
1.  Rename ImplicitsTypeAnnotator to DefaultForTypeAnnotator.
2. Move typeNames and types from ImplicitFor to DefaultFor.
3. Remove ImplicitFor

Note, I have not yet update the manual for this change.  That's because the concept of "implicit" qualifiers no longer exists, so I need to update https://checkerframework.org/manual/#creating-type-introduction and several other sections.  It make more sense to do this once all the other changes are implemented.  (Or in parallel to those changes.)


framework/src/main/java/org/checkerframework/framework/qual/DefaultFor.java
```diff
@@ -7,21 +7,33 @@
...
+
+    /** @return {@link TypeKind}s of types for which an annotation should be implicitly added */
+    TypeKind[] typeKinds() default {};
+
+    /**
+     * @return {@link Class}es for which an annotation should be applied. For example, if
+     *     {@code @MyAnno} is meta-annotated with {@code @DefaultFor(classes=String.class)}, then
+     *     every occurrence of {@code String} is actually {@code @MyAnno String}.
+     */
+    Class<?>[] classes() default {};
```
wmdietl: I agree that `typeNames` was confusing, given that it isn't a string attribute. `classes` describes the type of the attribute, but isn't very clear about the intention.
Maybe just `types`?

checker/src/main/java/org/checkerframework/checker/lock/qual/GuardedBy.java
```diff
@@ -38,10 +37,10 @@
 @SubtypeOf(GuardedByUnknown.class)
 @Documented
 @DefaultQualifierInHierarchy
-@DefaultFor({TypeUseLocation.EXCEPTION_PARAMETER, TypeUseLocation.UPPER_BOUND})
 @DefaultInUncheckedCodeFor({TypeUseLocation.PARAMETER})
-@ImplicitFor(
-        types = {
+@DefaultFor(
+        value = {TypeUseLocation.EXCEPTION_PARAMETER, TypeUseLocation.UPPER_BOUND},
+        typeKinds = {
```
wmdietl: Same comment as for `NonNull.java` below: these `typeKinds` no longer serve a purpose, as this is already the default qualifier.

smillst: Tests fail without these, I'm looking into it.

framework/src/main/java/org/checkerframework/framework/qual/DefaultFor.java
smillst: I like `types`.  Here's an example annotation with all three attributes:
```
@DefaultFor(
        value = TypeUseLocation.RETURN, 
        typeKinds = TypeKind.INT,
        types = String.class
)
```
Should we change `value` to `locations`?
```
@DefaultFor(
        locations = TypeUseLocation.RETURN, 
        typeKinds = TypeKind.INT,
        types = String.class
)
```
I kept it as value because most uses of `@DefaultFor` only specify locations; if we change `value` to `locations` it makes those uses more verbose.  (Though it also makes the specification clearer.)

Here's the current annotation for `Initialized`:
```
@DefaultFor({
    TypeUseLocation.IMPLICIT_UPPER_BOUND,
    TypeUseLocation.IMPLICIT_LOWER_BOUND,
    TypeUseLocation.EXCEPTION_PARAMETER
})
```
```
@DefaultFor(
        locations = {
            TypeUseLocation.IMPLICIT_UPPER_BOUND,
            TypeUseLocation.IMPLICIT_LOWER_BOUND,
            TypeUseLocation.EXCEPTION_PARAMETER
        })
```

@mernst, do you have any opinion on the names?

mernst: @smillst I agree with your preference for `types`.
I am ambivalent about `values` versus `locations`; either is fine with me.

checker/src/main/java/org/checkerframework/checker/lock/qual/GuardedBy.java
smillst: These are required because the default for local variables is `@GuardedByUnknown`, but if the local variable is one of these type kinds, the default should be `@GuardedByUnknown`.

framework/src/main/java/org/checkerframework/framework/qual/DefaultFor.java
smillst: I'm going to leave it as `values`.

checker/src/main/java/org/checkerframework/checker/signedness/qual/Signed.java
```diff
@@ -34,7 +34,7 @@
         // correctly.
 
         /*,
-        typeNames = {
+        classes = {
```
wmdietl: For consistency with the other changes, this should now be `types`.

framework/src/main/java/org/checkerframework/framework/type/typeannotator/DefaultForTypeAnnotator.java
```diff
@@ -116,15 +119,16 @@ public void addTypeClass(
         }
     }
 
-    public void addTypeName(Class<?> typeName, AnnotationMirror theQual) {
-        String typeNameString = typeName.getCanonicalName();
-        boolean res = qualHierarchy.updateMappingToMutableSet(typeNames, typeNameString, theQual);
+    /** Add default qualifier, {@code theQual}, for the given class. */
+    public void addClasses(Class<?> clazz, AnnotationMirror theQual) {
```
wmdietl: In a few places, in particular field and method names, I would change the `Classes` to `Types`, to be consistent with the name of the annotation attribute.

checker/src/main/java/org/checkerframework/checker/lock/qual/GuardedBy.java
wmdietl: Maybe add a comment? If one is used to CLIMB defaults, this is surprising.

framework/src/main/java/org/checkerframework/framework/qual/DefaultFor.java
wmdietl: If most locations only use the locations, I'm fine with continuing to use `value`. When there are multiple attributes, using `locations` would be much nicer.

PR #2522, opened 2019-05-30
Removed @NonNegative annotation to the argument of write(int)

Kakarot-SSJ4:
The argument of write(int) need not be `@NonNegative`.
Consider the following code:

import java.io.*;
class tryit{
public static void main(String[] args) throws IOException {
	Writer a = new PrintWriter(System.out);
	a.write(-1);
}
}

Compiling it by 'javac tryit.java' and executing it by 'java tryit'.
It gives no error and is run safely.

### DIFF
--- a/checker/jdk/index/src/java/io/Writer.java
+++ b/checker/jdk/index/src/java/io/Writer.java
@@ -105,7 +105,7 @@ public abstract class Writer implements Appendable, Closeable, Flushable {
      * @throws  IOException
      *          If an I/O error occurs
      */
-    public void write() throws IOException {
+    public void write(int c) throws IOException {
         synchronized (lock) {
             if (writeBuffer == null){
                 writeBuffer = new char[writeBufferSize];
--- a/checker/jdk/index/src/java/io/Writer.java
+++ b/checker/jdk/index/src/java/io/Writer.java
@@ -105,7 +105,7 @@ public abstract class Writer implements Appendable, Closeable, Flushable {
      * @throws  IOException
      *          If an I/O error occurs
      */
-    public void write(@NonNegative int c) throws IOException {
+    public void write() throws IOException {
         synchronized (lock) {
             if (writeBuffer == null){
                 writeBuffer = new char[writeBufferSize];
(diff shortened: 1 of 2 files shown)


checker/src/main/java/org/checkerframework/checker/index/upperbound/messages.properties
```diff
@@ -1,6 +1,6 @@
 ### Error messages for the Upper Bound Checker
 array.access.unsafe.high=Potentially unsafe array access: the index could be larger than the array's bound\nfound   : %s\nrequired: @IndexFor("%s") or @LTLengthOf("%s") -- an integer less than %s's length
-array.access.unsafe.high.constant=Potentially unsafe array access: the constant index %s could be larger than the array's bound\nfound   : %s\nrequired: @MinLen("%s") -- an array guaranteed to have at least %s elements
+array.access.unsafe.high.constant=Potentially unsafe array access: the constant index %s could be larger than the array's bound\nfound   : %s\nrequired: @MinLen(%s) -- an array guaranteed to have at least %s elements
```
kelloggm: This change seems unrelated to this pull request, though I agree we should make it. Can you separate it into a separate PR?

checker/jdk/index/src/java/io/Writer.java
```diff
@@ -105,7 +105,7 @@ protected Writer(Object lock) {
      * @throws  IOException
      *          If an I/O error occurs
      */
-    public void write(@NonNegative int c) throws IOException {
+    public void write() throws IOException {
```
kelloggm: `write` still takes an `int`, so this should be `public void write(int c) throws IOException {`

kelloggm:
The Travis build is also failing. Can you follow the instructions at https://github.com/typetools/annotated-libraries to update the JDK version being used in the PR?

When you get to the point where you need to upload a new JDK copy, just make a PR there and tag me, and I'll approve it.

kelloggm:
This PR still needs to modify the annotated JDK. We'd like to avoid keeping PRs open, so it would be great if this PR can move forward.

PR #2523, opened 2019-05-30
Convert signedness jdk.astub to annotated JDK

gokaco:


the1derer:
@gokaco 
Can you give me write access so I can push changes to remove the merge conflicts?
I can also create a separate pull-request and close this one if you want.

Thanks.

checker/jdk/signedness/src/java/lang/Integer.java
```diff
@@ -0,0 +1,1598 @@
...
+    /**
+     * Adds two integers together as per the + operator.
+     *
+     * @param a the first operand
+     * @param b the second operand
+     * @return the sum of {@code a} and {@code b}
+     * @see java.util.function.BinaryOperator
+     * @since 1.8
+     */
+    public static @SignednessEither int sum(@PolySigned int a, @PolySigned int b) {
```
mernst: This annotation is incorrect.  The sum of two arbitrary values is not guaranteed to be in the signed positive range.

You can verify this by type-checking the following program:

```
import org.checkerframework.checker.signedness.qual.SignednessEither;
import org.checkerframework.checker.signedness.qual.PolySigned;

public class Sum {

  public static @SignednessEither int sum(@PolySigned int a, @PolySigned int b) {
         return a + b;
    }
}
```

which yields the following output:

```
javac -processor signedness Sum.java
Sum.java:7: error: [return.type.incompatible] incompatible types in return.
         return a + b;
                  ^
  type of expression: @PolySigned int
  method return type: @SignednessEither int
1 error
```

Could you please:

1. Determine why the Signedness Checker issued no warning when compiling the annotated JDK, and ensure that it does so.
2. Only after doing that, fix this error (and any others that the type-checker issues while compiling the Signedness JDK).

checker/jdk/signedness/src/java/lang/Integer.java
```diff
@@ -0,0 +1,1598 @@
...
+     * distance) == rotateRight(val, distance & 0x1F)}.
+     *
+     * @param i the value whose bits are to be rotated right
+     * @param distance the number of bit positions to rotate right
+     * @return the value obtained by rotating the two's complement binary
+     *     representation of the specified {@code int} value right by the
+     *     specified number of bits.
+     * @since 1.5
+     */
+    public static @PolySigned int rotateRight(@PolySigned int i, int distance) {
```
mernst: This annotation looks suspicious to me.
I opened issue https://github.com/typetools/checker-framework/issues/2543 regarding this.

checker/jdk/signedness/src/java/lang/Integer.java
gokaco: You are right that sum of two arbitrary values is not guaranteed to be in the signed positive range. That's why I used @SignednessEither which means that value may be interpreted as @Signed or @Unsigned.

gokaco: The Signedness Checker issued no warning because-
I have seen previously also that the Signedness Checker only runs on methods signatures and not on its body of the annotated JDK. This same type of thing happened  earlier too [here](https://github.com/typetools/checker-framework/pull/2400#discussion_r279038468) where I annotated the body but nothing was happening due to that change.

mernst: 1. Can you explain why the result of every call to `sum` can be interpreted as either `@Signed` or `@Unsigned`?
2. Other type systems issue a warning regarding annotations in method bodies.  Please confirm that fact, and determine why the Signedness Checker is not.

Thanks!

gokaco: According to Signedness Checker definition-

1. If a number is @Signed and within a positive signed range then its annotation is replaced by 
    @SignednessEither else @Signed.
2. If a number is @Unsigned and within a positive signed range then its annotation is replaced by 
    @SignednessEither else @Unsigned.

Therefore,
1. If the sum is of two unsigned numbers, it can be in signed positive range or maybe not, 
    therefore, @SignednessEither, which should control both.
2. If the sum is of two signed numbers, it can be in signed positive range or maybe not, 
    therefore, @SignednessEither, which should control both.

mernst: I think you may have a mistaken understanding of `@SignednessEither`.

`@SignednessEither` is for expressions for which both signed and unsigned interpretations are definitely legal.  That is not the case for the result of `sum`.  Not every call to `sum` is definitely sensible to interpret as unsigned, and not every call is definitely sensible to interpret as unsigned.

checker/jdk/signedness/src/java/lang/Integer.java
```diff
@@ -0,0 +1,1602 @@
...
+     * </blockquote>
+     *
+     * @param   i   an integer to be converted to a string.
+     * @return  the string representation of the unsigned integer value
+     *          represented by the argument in hexadecimal (base&nbsp;16).
+     * @see #parseUnsignedInt(String, int)
+     * @see #toUnsignedString(int, int)
+     * @since   JDK1.0.2
+     */
+    public static String toHexString(@UnknownSignedness int i) {
```
mernst: toUnsignedString0 takes an unsigned string (and passes it to a method documented as taking an unsigned string), so I think toHexString should too.
The same goes for to toOctalString and toBinaryString.

checker/jdk/signedness/src/java/lang/Integer.java
```diff
@@ -0,0 +1,1602 @@
...
+     * Returns the value obtained by reversing the order of the bits in the
+     * two's complement binary representation of the specified {@code int}
+     * value.
+     *
+     * @param i the value to be reversed
+     * @return the value obtained by reversing order of the bits in the
+     *     specified {@code int} value.
+     * @since 1.5
+     */
+    public static @SignednessGlb int reverse(@UnknownSignedness int i) {
```
mernst: Why isn't this polymorphic?  Similar question for reverseBytes.

checker/jdk/signedness/src/java/lang/Integer.java
```diff
@@ -0,0 +1,1602 @@
...
+     * one-bits in its two's complement binary representation, that is, if it
+     * is equal to zero.
+     *
+     * @param i the value whose highest one bit is to be computed
+     * @return an {@code int} value with a single one-bit, in the position
+     *     of the highest-order one-bit in the specified value, or zero if
+     *     the specified value is itself equal to zero.
+     * @since 1.5
+     */
+    public static int highestOneBit(@UnknownSignedness int i) {
```
mernst: This method should be polymorphic in its argument, I think.
The same goes for lowestOneBit.

checker/jdk/signedness/src/java/util/Arrays.java
```diff
@@ -0,0 +1,5118 @@
...
+     *
+     * <p>Implementation note: The sorting algorithm is a Dual-Pivot Quicksort
+     * by Vladimir Yaroslavskiy, Jon Bentley, and Joshua Bloch. This algorithm
+     * offers O(n log(n)) performance on many data sets that cause other
+     * quicksorts to degrade to quadratic performance, and is typically
+     * faster than traditional (one-pivot) Quicksort implementations.
+     *
+     * @param a the array to be sorted
+     */
+    public static void sort(@PolySigned int[] a) {
```
mernst: I don't think this can be polymorphic.
When called on an array of unsigned values, it does not sort -- very large unsigned values are treated as negative.
The same comment applies to other sort routines, to parallelSort, and binarySearch.

gokaco: Why it cannot sort negative numbers?

gokaco:
> Thanks for these annotations. I left comments about annotations that look questionable to me. Please explain or change them. All comments on integers also apply to longs.

@mernst Thanks for the review! I have updated the annotations. Travis may fail as I have not updated the jdkHash. I will update it when this pull request will get approved. I have left a comment.

checker/jdk/signedness/src/java/lang/Integer.java
```diff
@@ -0,0 +1,1602 @@
...
+     * </ul>
+     *
+     * @param i the value whose number of leading zeros is to be computed
+     * @return the number of zero bits preceding the highest-order
+     *     ("leftmost") one-bit in the two's complement binary representation
+     *     of the specified {@code int} value, or 32 if the value
+     *     is equal to zero.
+     * @since 1.5
+     */
+    public static @NonNegative int numberOfLeadingZeros(@UnknownSignedness int i) {
```
mernst: There is an inconsistency between numberOfLeadingZeros and numberOfLeadingZeros.  Both can be called on any argument, but one uses `@UnknownSignedness` whereas the other uses `@PolySigned`.  If the argument's type isn't used elsewhere in the signature, I suggest using `@UnknownSignedness` consistently throughout.  (An exception is for argument element types, such as in `String` constructors; there you need to use `@PolySigned` even though no other type in the signature is polymorphic.)

checker/jdk/signedness/src/java/lang/Integer.java
```diff
@@ -0,0 +1,1602 @@
...
+     * A constant holding the minimum value an {@code int} can
+     * have, -2<sup>31</sup>.
+     */
+    @Native public static final int   MIN_VALUE = 0x80000000;
+
+    /**
+     * A constant holding the maximum value an {@code int} can
+     * have, 2<sup>31</sup>-1.
+     */
+    @Native public static final @Positive int   MAX_VALUE = 0x7fffffff;
```
mernst: Throughout, I suggest using `@SignedPositive` rather than `@Positive` and `@NonNegative`.
Using `@Positive` makes it look like the code is annotated for the Index Checker, which it is not.

checker/jdk/signedness/src/java/lang/Integer.java
```diff
@@ -0,0 +1,1602 @@
...
+     * distance) == rotateRight(val, distance & 0x1F)}.
+     *
+     * @param i the value whose bits are to be rotated right
+     * @param distance the number of bit positions to rotate right
+     * @return the value obtained by rotating the two's complement binary
+     *     representation of the specified {@code int} value right by the
+     *     specified number of bits.
+     * @since 1.5
+     */
+    public static @Unsigned int rotateRight(@Unsigned int i, int distance) {
```
mernst: Why is rotateRight unsigned->unsigned whereas rotateLeft is polymorphic?  I think both should be polymorphic.

checker/jdk/signedness/src/java/util/Arrays.java
mernst: The issue is not negative numbers.  The issue is incorrect interpretation of unsigned values.
Suppose that a client passed an array of unsigned values:
 [ 0, 2^31-1, 2^31, 2^32-1 ]
Note that all of these are values greater than or equal to zero.

The client would expect these values not to be reordered by the sort routine.  But this method will interpret them as
[0, 2^31-1, -2^31, -1]
and will reorder them.  Thus, clients must not pass an array of unsigned integers to this routine.
The same comment applies to other sort routines, and binary search for the same reason.

checker/jdk/signedness/src/java/util/Arrays.java
```diff
@@ -0,0 +1,5118 @@
...
+     */
+    public static void sort(Object[] a, int fromIndex, int toIndex) {
+        rangeCheck(a.length, fromIndex, toIndex);
+        if (LegacyMergeSort.userRequested)
+            legacyMergeSort(a, fromIndex, toIndex);
+        else
+            ComparableTimSort.sort(a, fromIndex, toIndex, null, 0, 0);
+    }
+
+    /** To
```
mernst: This annotation (and many others) can more usefully be `@PolyAll`.  I suggest that you do that in a later pull request after this one is merged.

checker/build.gradle
```diff
@@ -4,7 +4,7 @@ ext {
     // See instructions for updating jdk8.jar at
     // https://github.com/typetools/annotated-libraries/blob/master/README.md .
     // Set jdkShaHash to 'local' to use a locally-built version of jdk8.jar .
-    jdkShaHash = '59a70f00fadc6abf888e8b3e98430792e4db617a'
+    jdkShaHash = '07a52880879073ad3caa5e39f21b0ba2d48c5eab'
```
mernst: To make the continuous integration tests pass, could you please update the annotated JDK?  Thanks.

checker/jdk/signedness/src/java/lang/Integer.java
gokaco: That might look suspicious.
Check [this](https://github.com/typetools/checker-framework/pull/2523#discussion_r291809482).

checker/jdk/signedness/src/java/lang/Integer.java
gokaco: It may give error because earlier when these fields were not annotated then error of `conflicting annotations` were occurring then I annotated them like this.
Check [this](https://groups.google.com/d/msg/checker-framework-gsoc/qeDMhtFW_uc/AxFyOaXtCgAJ)

checker/jdk/signedness/src/java/lang/Integer.java
gokaco: @mernst I am unable to understand this statement. Please, can you explain!

mernst: This is not a helpful question, because it is too vague.
Which statement do you not understand?
All 4 sentences that I wrote, or just some of them?
What *do* you understand, and what *don't* you understand?

checker/jdk/signedness/src/java/lang/Integer.java
mernst: Exactly what error may it give?
Is there a bug report in the issue tracker?  If not, please add one.  If there is, you could point me at it.

checker/jdk/signedness/src/java/lang/Integer.java
gokaco: > There is an inconsistency between numberOfLeadingZeros and numberOfLeadingZeros. 

I did not understood this statement.

checker/jdk/signedness/src/java/lang/Integer.java
gokaco: > Exactly what error may it give?

It will give the error `type.invalid.conflicting.annos` when files like LessThanValue.java and some guava files will get compiled.
For more details check this travis [link](https://travis-ci.org/typetools/checker-framework/builds/561164283?utm_source=github_status&utm_medium=notification).
> Is there a bug report in the issue tracker? If not, please add one. If there is, you could point me at it.

Is this a bug? It is looking to me more like a property. The thing which is happening is that when the JDK gets compiled it gets two different annotations (One from annotated JDK of signedness checker and one from Index Checker), which are getting conflicted.

checker/build.gradle
gokaco: Done!

checker/jdk/signedness/src/java/util/Arrays.java
gokaco: Fixed.

checker/jdk/signedness/src/java/lang/Integer.java
mernst: Your response is not relevant to the comment, which is about the discrepancy between the annotations on rotateLeft and rotateRight.  I believe that they should be annotated similarly.  Do you agree with that or disagree with it?
Your response might be relevant to deciding how to annotate both rotateLeft and rotateRight.  What annotation do *you* believe is correct for the methods, and why?

checker/jdk/signedness/src/java/lang/Integer.java
mernst: I think you are saying that the user can write two reasonable annotations that are not in conflict with one another, and the Checker Framework issues an undesirable when both are present.  That is a bug; could you please report it on the issue tracker?

(But, I'm not sure because you did not give a concrete example.)

checker/jdk/signedness/src/java/lang/Integer.java
mernst: > Both can be called on any argument, but one uses @UnknownSignedness whereas the other uses @PolySigned

Why don't you annotate both methods with the same annotation?  Is there a reason to prefer different annotations for the two methods?

mernst:
Please don't force-push.  It causes trouble (merge conflicts, different history) for other people who have cloned the branch.

checker/jdk/signedness/src/java/lang/Integer.java
gokaco: >  I believe that they should be annotated similarly. Do you agree with that or disagree with it?

I agree
> Your response might be relevant to deciding how to annotate both rotateLeft and rotateRight. What annotation do _you_ believe is correct for the methods, and why?

I believe that both should be annotated with `@PolySigned` as both rotate the bits which can be unsigned or signed. There was an issue related to this, therefore, I was confused. I will correct it.

checker/jdk/signedness/src/java/lang/Integer.java
mernst: @gokaco Have you addressed this comment?

gokaco: > I think you are saying that the user can write two reasonable annotations that are not in conflict with one another, and the Checker Framework issues an undesirable when both are present. That is a bug; could you please report it on the issue tracker?

I have issued an issue related to this in the issue tracker. Please check it! #2641

checker/jdk/signedness/src/java/lang/Integer.java
gokaco: @mernst About which two methods you are talking about? You have written numberOfLeadingZeros twice. Please rectify that.
>There is an inconsistency between numberOfLeadingZeros and numberOfLeadingZeros.

mernst: highestOneBit

gokaco:
@mernst It is ready for review.

checker/jdk/signedness/src/java/lang/Integer.java
gokaco: Fixed this!

checker/jdk/signedness/src/java/lang/Integer.java
gokaco: Fixed this!

gokaco:
> @mernst It is ready for review.

@mernst Reminder! It is ready for review.

checker/jdk/signedness/src/java/lang/Integer.java
```diff
@@ -1274,7 +1274,7 @@ public static int compareUnsigned(@Unsigned int x, @Unsigned int y) {
      *         conversion
      * @since 1.8
      */
-    public static @Unsigned long toUnsignedLong(@UnknownSignedness int x) {
+    public static @SignedPositive long toUnsignedLong(int x) {
```
gokaco: @mernst Why we have annotated this as @SignedPositive?

mernst: Because it is always a positive value.  More specifically, it is a long value in the range 0..2^32-1.
When you ask a question, please give context.  If you believe that the `@SignedPositive` annotation is wrong, then say that *explicitly* so that we can have a conversation, rather than making me guess about why you are asking the question.  Please clarify.

gokaco: Because it was converting the integer to an unsigned as the method is `toUnsignedLong` thats why I thought that it should be annotated as `@Unsigned`. But now I got your point.

gokaco: @mernst Due to this change `UtilsJava8.java` test was failing. I have updated the test case and the jdkHash.
Now you can continue with your review.

PR #2525, opened 2019-06-01
Improved JDK annotations

Maxi17:
Some improvements to offsets in java.io and java.lang. Some new files are also included.


checker/jdk/index/src/java/nio/channels/ByteChannel.java
```diff
@@ -0,0 +1,45 @@
+/*
```
kelloggm: What is the purpose of including files that have no annotations in them?

kelloggm:
However, there is a merge conflict, which needs to be resolved before this PR can be merged.

checker/jdk/index/src/java/nio/channels/ByteChannel.java
Maxi17: When I added one class, I thought it is a good idea for completeness to add its superclasses/subclasses, or other related classes.

kelloggm: Okay, that is reasonable.

checker/build.gradle
```diff
@@ -11,7 +11,7 @@ sourceSets {
     main {
         resources {
             // Stub files, message.properties, etc.
-            srcDirs += ['src/main/java']
+            srcDirs = ['src/main/java']
```
kelloggm: This is an odd change that doesn't seem to have anything to do with the rest of this PR. What's the reasoning behind it? I don't think I can merge this in good conscience without knowing why this change is needed.

PR #2535, opened 2019-06-05
Change names of some variables that were confusing

kelloggm:
At Mike's request


mernst:
@kelloggm The Travis build is failing because of missing documentation.
It's not your fault, but the person who encounters the problem typically fixes it.
Could you add documentation for the three fields and the three methods?
Thanks.

PR #2542, opened 2019-06-07
Remove unnecessary download of jsr308-langtools

wmdietl:
This was added in #2124 but doesn't seem necessary.
Is there any reason to keep this around?

### DIFF
--- a/.travis-build-without-test.sh
+++ b/.travis-build-without-test.sh
@@ -62,13 +62,6 @@ echo "... done: (cd ../stubparser/ && ./.travis-build-without-test.sh)"
 
 ## Compile
 
-# Download jsr308-langtools replacement for javac.jar that fixes some bugs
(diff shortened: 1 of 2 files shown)


smillst:
This was required for some Jtreg tests:
https://github.com/typetools/checker-framework/blob/e9c4635866e35f61abd80b5036a5a1be87a3e4f2/build.gradle#L485-L491

But, since the tests now pass, I guess it isn't required any longer.

wmdietl:
> This was required for some Jtreg tests:
> https://github.com/typetools/checker-framework/blob/e9c4635866e35f61abd80b5036a5a1be87a3e4f2/build.gradle#L485-L491
> 
> But, since the tests now pass, I guess it isn't required any longer.

Thanks for noticing this! I've now also removed the mention of `javap.jar` there and locally the jtreg tests continued to pass.

PR #2545, opened 2019-06-09
Make get and orElseThrow consistent.

wmdietl:
Simplify signature.
Related to 8c21ebc983f60df3f0c83550a2258fe61f165472
The change to Opt.get should also be added to the changelog.


mernst:
Thanks for making those methods consistent.

PR #2546, opened 2019-06-11
Separate type declaration qualifier bounds from defaults 

### LINKED ISSUE #2247: Nullness Checker: Any instance type of a class must be a subtype of its declared type

smillst:
(This is a PR for defaulty changes.  It should be merged into defautlyChanges after #2521, but it is a pull request into removeImplicitFor.)
 
* Add @DefaultQualifierOnUse and @NoDefaultQualifierOnUse type declaration annotations
* Enforce upper bound on use, add @UpperBoundFor meta-annotation
* Remove TypeUseLocation.TYPE_DECLARATION
* Remove InheritedFromClassAnnotator, replace with DefaultQualifierForUseTypeAnnotator
* Adapt the Nullness Checker and fixes #2247
* Does not adapt the Interning Checker (this is done in the next PR)

Documentation for this change is [in this GoogleDoc](https://docs.google.com/document/d/1z5MWlc3URE6mB1tdq2buRV0AINsvJkv1Gc5xwqtS43Y/).  Parts of this document will be added to the manual in another PR.


framework/src/main/java/org/checkerframework/common/basetype/BaseTypeValidator.java
```diff
@@ -223,16 +242,20 @@ public Void visitDeclared(AnnotatedDeclaredType type, Tree tree) {
 
         final boolean skipChecks = checker.shouldSkipUses(type.getUnderlyingType().asElement());
 
-        if (!skipChecks) {
+        if (!checkTopLevelDeclaredType && !skipChecks) {
```
wmdietl: Related to the "logic wrong way around" comment above: I would not expect a negation here, given the variable name.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -197,6 +196,9 @@
      */
     private final AnnotationFormatter annotationFormatter;
 
+    /** Holds the qualifier upper bounds for type uses. */
+    protected QualifierUpperBounds qualifierUpperBounds;
```
wmdietl: Looks like the field could be final.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1146,6 +1152,41 @@ public AnnotatedTypeMirror getAnnotatedTypeFromTypeTree(Tree tree) {
...
+    /** Returns the set of qualifiers that are the upper bounds for a use of the type. */
+    public Set<AnnotationMirror> getTypeDeclarationBounds(AnnotatedTypeMirror type) {
+        return qualifierUpperBounds.getBoundQualifiers(type);
+    }
+
+    /**
+     * Returns the set of qualifiers that are the upper bound for a type use if no other bound is
+     * specified for the type.
+     *
+     * <p>Super implementation returns the top qualifiers by default. Subclass may override to
```
wmdietl: The `Super` made me look for a super class implementation. Wouldn't `This implementation...` work better?

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1690,4 +1701,30 @@ private String getCheckerName() {
     public CFGVisualizer<Value, Store, TransferFunction> getCFGVisualizer() {
         return cfgVisualizer;
     }
+
+    /**
+     * Adds default qualifiers for uses types of to {@code type}. If {@code element} is a local
```
wmdietl: The word ordering in the first sentence seems strange.

framework/src/main/java/org/checkerframework/framework/qual/NoDefaultQualifierForUse.java
```diff
@@ -0,0 +1,15 @@
+package org.checkerframework.framework.qual;
+
+import java.lang.annotation.Annotation;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Target;
+
+/**
+ * Declaration annotation applied to type declarations to specify that the upper bound qualifier
+ * should not be applied to unannotated uses of the type.
```
wmdietl: It would be helpful to say what other default is then applied instead.

framework/src/main/java/org/checkerframework/framework/qual/NoDefaultQualifierForUse.java
```diff
@@ -0,0 +1,15 @@
+package org.checkerframework.framework.qual;
+
+import java.lang.annotation.Annotation;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Target;
+
+/**
+ * Declaration annotation applied to type declarations to specify that the upper bound qualifier
```
wmdietl: It's not clear to me what qualifier the "upper bound qualifier" refers to here.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
smillst: It's initialized in postInit(), so it can't be final.

smillst:
@wmdietl, I've addressed your first round of feedback.  Because removeImplicitFor is now merged, I'm going to close this pull request and open a new against the defaultyChanges branch.

PR #2547, opened 2019-06-11
Indicate annotation sources, for debugging

mernst:
Merge together with https://github.com/typetools/annotation-tools/pull/217


framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1064,6 +1064,12 @@ public AnnotationMirror getAnnotationMirror(Tree tree, Class<? extends Annotatio
      * @return the annotated type of {@code tree}
      */
     public AnnotatedTypeMirror getAnnotatedType(Tree tree) {
+
+        String treeString = tree.toString();
```
wmdietl: This seems to be left-over from some debugging. The `treeString` variable isn't used and computing it can be costly.

framework/src/main/java/org/checkerframework/framework/util/FieldInvariants.java
```diff
@@ -83,6 +83,9 @@ public FieldInvariants(
             throw new BugInCF("malformed FieldInvariants");
         }
         String fieldString = field.toString();
+        if (!isWellFormed()) {
```
wmdietl: This repeats the check from a few lines above and nothing of the state seems to have changed in between. Is this redundant?

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -147,6 +150,62 @@ public static AnnotationMirror fromClass(
...
+     * Creates an {@link AnnotationMirror} given by a particular annotation class. Each element is
+     * set to its default value. Errs if any element does not have a default value.
+     *
+     * @param elements the element utilities to use
+     * @param aClass the annotation class
+     * @return an {@link AnnotationMirror} of type given type
+     */
+    public static AnnotationMirror fromClassWithDefaults(
+            Elements elements, Class<? extends Annotation> aClass) {
+        AnnotationMirror result = fromName(elements, aClass.getCanonicalName());
```
wmdietl: This should either use `fromClass`, which checks for `null`, or check for `null` here.

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -147,6 +150,62 @@ public static AnnotationMirror fromClass(
...
+        setElementsToDefaultValues(result, aClass);
+        return result;
+    }
+
+    /**
+     * Changes the annotation value for the given key of the given annotation to newValue and
+     * returns the previous value.
+     */
+    @SuppressWarnings("unchecked")
+    public static Object changeAnnotationValue(Annotation annotation, String key, Object newValue) {
```
wmdietl: This method isn't used, in particular not in `setElementsToDefaultValues`.

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -147,6 +150,62 @@ public static AnnotationMirror fromClass(
...
+    }
+
+    /**
+     * Sets the values of the elements (fields) of am to their defaults. Errs if any element does
+     * not have a default value.
+     */
+    public static void setElementsToDefaultValues(
+            AnnotationMirror am, Class<? extends Annotation> aClass) {
+        for (Method m : aClass.getDeclaredMethods()) {
+            System.out.println("  " + m);
```
wmdietl: This implementation only contains debug output and doesn't implement the actual functionality.
The changes to this file seem to be work in progress and not related to this PR. Maybe they were added in error and should be in a separate PR?

PR #2548, opened 2019-06-13
annotations not required in the parameters of regionMatches() in java.lang.String

Kakarot-SSJ4:
The `@IndexOrHigh` annotations on the parameters `toffset`, `ooffset` and `len`  of `regionMatches()` functions is not required as it is checked in the if statements inside the function and a false value is returned if the above annotation is not satisfied by the parameters.

Consider the following code:
```
class tryit{
	String a = "abcd";
	System.out.println(a.regionMatches(-1,"abc",-2,-1));
}
}
```
Compile it by javac tryit.java
Run it by java tryit
It prints false and throws no error.

### DIFF
--- a/checker/jdk/index/src/java/lang/String.java
+++ b/checker/jdk/index/src/java/lang/String.java
@@ -1275,8 +1275,8 @@ public final class String
      *          exactly matches the specified subregion of the string argument;
      *          {@code false} otherwise.
      */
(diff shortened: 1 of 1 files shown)


kelloggm:
Akash,

Thanks for this pull request. Can you follow the instructions in the README at https://github.com/typetools/annotated-libraries to update the JDK binary, as well, and then modify this PR so that the hash of the JDK that is downloaded by default is set to the new binary that you uploaded?

Once that is done, we should be able to merge this.

Martin

PR #2551, opened 2019-06-13
Refactor DOTCFGVisualizer

xingweitian:
Hello, this PR adds a String implementation of `CFGVisualizer` and refactors `DOTCFGVisualizer`. We now have an abstract class `AbstractCFGVisualizer` which minimises the duplicate code between `DOTCFGVisualizer` and `StringCFGVisualizer`. In the future, it may be easier to build custom `CFGVisualizer` based on `AbstractCFGVisualizer`.

We can use `StringCFGVisualizer.visualize()` to see the String version of the control flow graph. I have posted a simple result at [here](https://paste.ubuntu.com/p/9vHjHYYyt7/).

In addition, the value of `Store` can be visualized by calling `StringCFGVisualizer.visualizeStore()`.


framework/src/main/java/org/checkerframework/framework/flow/CFAbstractStore.java
```diff
@@ -940,41 +940,47 @@ public int hashCode() {
     @SideEffectFree
     @Override
     public String toString() {
-        return "Use a CFGVisualizer to see the Store: " + this.hashCode();
+        return "Use visualize(new StringCFGVisualizer<>()) to see the Store: " + this.hashCode();
```
wmdietl: Please replace this with
`return visualize(new StringCFGVisualizer<>(...))`
if possible.

wmdietl:
I performed a few iterations here: https://github.com/opprop/checker-framework/pull/90
This looks good to me, modulo one comment.

wmdietl:
This is the more general alternative to #2486 and that PR can then be closed.

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractStore.java
xingweitian: It works properly. Sorry I got it wrong at first, I have already changed it.

xingweitian: If I understand correctly, now when we call `toString()`, we can get the String version of the value of this Store directly.

mernst:
Colud you please:
 * pull master into your branch, which currently is 3 months behind master.
 * when you make a pull request, please give the maintainers permission to modify it.  There is a checkbox for that when you create the pull request.  You might want to close and re-open this pull request in order to enable that feature.  (Or give me permissions to your fork.)

Thanks!

xingweitian:
> Colud you please:
> 
> * pull master into your branch, which currently is 3 months behind master.
> * when you make a pull request, please give the maintainers permission to modify it.  There is a checkbox for that when you create the pull request.  You might want to close and re-open this pull request in order to enable that feature.  (Or give me permissions to your fork.)
> 
> Thanks!

Thanks for your reply! I just update my branch and I will close and reopen this PR.

xingweitian:
It seems like I can't select the checkbox by re-opening it. So I think maybe giving the permission to my fork would be a good idea.

mernst:
I still don't have permission to make changes.  Could you please do that?

You can give me permission to your fork.
Another approach is to create a new branch, pull your changes into that branch, and make new pull request.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizer.java
```diff
@@ -47,69 +48,77 @@
      * Delegate the visualization responsibility to the passed {@link Store} instance, which will
      * call back to this visualizer instance for sub-components.
      *
-     * @param store the store to visualize
+     * @param store the Store to visualize
+     * @return the String presentation of the value of {@link Store}
```
mernst: This comment has a few issues.
 * you don't need "the value of".  just "the String representation of" is fine.
 * the use of `Store` instead of `store` is incorrect.  This is not a representation of a type, but of a value, and that value is `store`, not `Store`.

Could you make changes like this throughout?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizer.java
```diff
@@ -34,11 +34,12 @@
      *
      * @param cfg the CFG to visualize
      * @param entry the entry node of the control flow graph to be represented
-     * @param analysis an analysis containing information about the program represented by the CFG.
+     * @param analysis the analysis containing information about the program represented by the CFG.
      *     The information includes {@link Store}s that are valid at the beginning of basic blocks
      *     reachable from {@code entry} and per-node information for value producing {@link Node}s.
      *     Can also be {@code null} to indicate that this information should not be output.
-     * @return possible analysis results, e.g. generated file names.
+     * @return possible analysis results, e.g. generated file names({@link DOTCFGVisualizer}) and
```
mernst: Missing spaces

dataflow/src/main/java/org/checkerframework/dataflow/analysis/Store.java
```diff
@@ -85,6 +85,7 @@
      * Delegate visualization responsibility to a visualizer.
      *
      * @param viz the visualizer to visualize this store
+     * @return the String of the value of this store
```
mernst: Missing word "representation"

xingweitian:
> I still don't have permission to make changes. Could you please do that?
> 
> You can give me permission to your fork.
> Another approach is to create a new branch, pull your changes into that branch, and make new pull request.

Could you please click [here](https://github.com/xingweitian/checker-framework/invitations) and accept the invitation so that you can have the permission to my repository.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizer.java
xingweitian: I have already made changes in this file. Could you please have a look if the changes are suitable? Then I can improve the comments in other files of this PR.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizer.java
```diff
@@ -48,68 +49,76 @@
      * call back to this visualizer instance for sub-components.
      *
      * @param store the store to visualize
+     * @return the String presentation of store
```
mernst: 1. "presentation" should be "representation"

2. I would use
```the String representation of {@code store}```
or
```the String representation of the store```
because just "store" is a bit confusing and not good English.

File: dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizer.java [hunk 6]
```diff
@@ -48,68 +49,76 @@
...
      */
-    void visualizeStore(S store);
+    String visualizeStore(S store);
 
     /**
      * Called by a {@code CFAbstractStore} to visualize the class name before calling the {@code
      * CFAbstractStore#internalVisualize()} method.
      *
      * @param classCanonicalName the canonical name of the class
+     * @return the String representation of the header of store
```
mernst: When I read this comment, I was unclear what store this is referring to.  This method doesn't take a store as an argument.
I think the answer is that this class contains a "current store".  `visualizeStore` sets it, and other methods implicitly use it.
I suggest that this should be documented in the class documentation and in `visualizeStore`, and this comment should refer to the "current store".
It's a bit confusing that the class documentation says that the class contains a CFG but doesn't say anything about it maintaining a current store.

I realize this confusion about the design isn't your fault and I don't wish to block the pull request on a lot of extra work.  However, you probably understand the design now and can make some clarifying comments along with this pull request.

Thanks!

wmdietl: There is no current store that is maintained by the CFGVisualizers. This method only returns the "header" for a store, which is only based on the passed class name.
The method description explains this correctly, I think.
So how about changing the return description to:
`@return the header to use for a store with the given canonical name`?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizer.java
xingweitian: I have been keep thinking about it but still a bit confused. I was just wondering if the way I am using to improve the comments is correct.

For example:
```java
    /**
     * Remove the String "Node" from the simple name of the node.
     *
     * @param t a {@link Node}
     * @return the String representation of the {@link Node}'s simple name
     */
```
At that time I am trying to express that we need a variable which type is "Node" as the parameter.

But it seems that I should use "node" instead of "Node".

So I should change it to something like this:

```java
    /**
     * Remove the String "Node" from the simple name of the node.
     *
     * @param t a node
     * @return the String representation of the node's simple name
     */
```

Is it correct? If so, I will do it in this way.

mernst: Usually, you do not need to mention the type of a variable, because each variable's type is already shown in the source code and in the generated API documentation.
Also, it is confusing to refer to a variable by its type rather than by its name.  So, the first line of your comment should be
```Remove the String {@code t} ...```
or just
```Remove {@code t} ...```
I'm confused by the example you showed, because the first line says it's a String and the `@param` say it's a node.

Comment (xingweitian) [reply dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizer.java]: Thanks for your explanation, I get it.

This example actually executes `t.getClass().getSimpleName()` to get the String representation of this node's simple name and replaces String "Node" with "".

File: dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizer.java (hunk 6 above)
Comment (xingweitian) [reply dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizer.java]: > There is no current store that is maintained by the CFGVisualizers. This method only returns the "header" for a store, which is only based on the passed class name.
> The method description explains this correctly, I think.
> So how about changing the return description to:
> `@return the header to use for a store with the given canonical name`?

That sounds good to me.

File: dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizer.java (hunk 2 above)
Comment (xingweitian) [reply dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizer.java]: I think this maybe easier to understand.
```java
    /**
     * Get the simple name of a node.
     *
     * @param t a node
     * @return the String representation of the node's simple name
     */
    protected String getNodeSimpleName(Node t) {
        String name = t.getClass().getSimpleName();
        return name.replace("Node", "");
    }

```

xingweitian:
I just updated some of the comments following what we discussed above, please let me know if I need to change something.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizer.java
```diff
@@ -48,116 +49,135 @@
...
      * CFAbstractStore#internalVisualize()}.
+     *
+     * @return the String representation of the footer to use for a store
      */
-    void visualizeStoreFooter();
+    String visualizeStoreFooter();
 
     /**
-     * Visualize a block based on the analysis.
+     * Visualize a special block based on the analysis.
```
mernst: Why did you add "special" here?  I expected it in `visualizeSpecialBlock`, but I am surprised to see it here in `visualizeBlock`.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizer.java
```diff
@@ -48,116 +49,135 @@
...
      */
-    void visualizeSpecialBlock(SpecialBlock sbb);
+    String visualizeSpecialBlock(SpecialBlock sbb);
 
     /**
      * Visualize the transferInput of a Block based on the analysis.
      *
      * @param bb the block
      * @param analysis the current analysis
+     * @return the String representation of the transferInput of this block
```
mernst: Is it "this block" (which implies that the receiver is or has a block), or "the given block" (which then refers to `bb`)?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizer.java
```diff
@@ -48,116 +49,135 @@
...
      */
-    void visualizeBlockTransferInput(Block bb, Analysis<A, S, T> analysis);
+    String visualizeBlockTransferInput(Block bb, Analysis<A, S, T> analysis);
 
     /**
      * Visualize a Node based on the analysis.
      *
      * @param t the node
      * @param analysis the current analysis
+     * @return the String representation of this node
```
mernst: The word "this" refers to the receiver.  You could say "node `t`" or "the given node".
Could you make similar changes throughout, please?

mernst:
This is looking better, thanks.  I gave a few comments that can be applied throughout, and then I will take another look.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizer.java
xingweitian: It's my fault, I will fix it.

xingweitian:
@mernst Could you please have another look?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizer.java
mernst: If you are replacing the text "Node" by "", then saying that is clear and useful.  I think that would be a good comment.  The other uses of types where variables were intended confused me.

You don't have to say `replaces the String "Node" with ""`; just "replaces "Node" with ""` is better because it does not give the impression that there is a String object that contains "Node" involved.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizer.java
```diff
@@ -48,116 +49,135 @@
...
+     * @return the String representation of the header to use for a store with the given canonical
+     *     name
      */
-    void visualizeStoreHeader(String classCanonicalName);
+    String visualizeStoreHeader(String classCanonicalName);
 
     /**
-     * Called by {@code CFAbstractStore#internalVisualize()} to visualize a local variable.
+     * Called by {@code CFAbstractStore#internalVisualize()} to visualize a local variable in the
+     * given store.
```
mernst: You should use "the given store" when a store is a formal parameter.  That does not seem to be the case here.  What store are you referring to?
The same comment applies elsewhere.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizer.java
```diff
@@ -48,116 +49,135 @@
...
+     * @return the String representation of the specific information collected by the given store
      */
-    void visualizeStoreKeyVal(String keyName, Object value);
+    String visualizeStoreKeyVal(String keyName, Object value);
 
     /**
      * Called by {@code CFAbstractStore} to visualize any information after the invocation of {@code
      * CFAbstractStore#internalVisualize()}.
+     *
+     * @return the String representation of the footer to use for a store
```
mernst: Which store is this for?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizer.java
xingweitian: Thanks for your comments. Here is my thinking. This method is one of the methods called by `internalVisualize()` to generate a part of the visualisation of the store which is the parameter of `visualizeStore(Store store)`. "the given store" does not appear as a parameter at here, but this method is used to support the visualisation of "the given store".

For example, if we call
```java
   viz.visualizeStore(store)
```
Then this method is called to visualize "the given store":
```java
    @Override
    public String visualize(CFGVisualizer<?, S, ?> viz) {
        /* This cast is guaranteed to be safe, as long as the CFGVisualizer is created by
         * CFGVisualizer<Value, Store, TransferFunction> createCFGVisualizer() of GenericAnnotatedTypeFactory */
        @SuppressWarnings("unchecked")
        CFGVisualizer<V, S, ?> castedViz = (CFGVisualizer<V, S, ?>) viz;
        StringBuilder sbVisualize = new StringBuilder();
        sbVisualize.append(castedViz.visualizeStoreHeader(this.getClass().getCanonicalName()));
        sbVisualize.append(internalVisualize(castedViz));
        sbVisualize.append(castedViz.visualizeStoreFooter());
        return sbVisualize.toString();
    }
```
We can find that `internalVisualize(castedViz)` is called at here.
And here is what `internalVisualize(castedViz)` does:
```java
    /**
     * Adds a representation of the internal information of this Store to visualizer {@code viz}.
     *
     * @return a String of the internal information of this {@link Store}
     */
    protected String internalVisualize(CFGVisualizer<V, S, ?> viz) {
        StringBuilder res = new StringBuilder();
        for (Entry<FlowExpressions.LocalVariable, V> entry : localVariableValues.entrySet()) {
            res.append(viz.visualizeStoreLocalVar(entry.getKey(), entry.getValue()));
        }
        if (thisValue != null) {
            res.append(viz.visualizeStoreThisVal(thisValue));
        }
        for (Entry<FlowExpressions.FieldAccess, V> entry : fieldValues.entrySet()) {
            res.append(viz.visualizeStoreFieldVals(entry.getKey(), entry.getValue()));
        }
        for (Entry<FlowExpressions.ArrayAccess, V> entry : arrayValues.entrySet()) {
            res.append(viz.visualizeStoreArrayVal(entry.getKey(), entry.getValue()));
        }
        for (Entry<MethodCall, V> entry : methodValues.entrySet()) {
            res.append(viz.visualizeStoreMethodVals(entry.getKey(), entry.getValue()));
        }
        for (Entry<FlowExpressions.ClassName, V> entry : classValues.entrySet()) {
            res.append(viz.visualizeStoreClassVals(entry.getKey(), entry.getValue()));
        }
        return res.toString();
    }
```
This method calls several `visualizeStoreXXX` methods and combine the return value of them as the String representation of "the given store".

xingweitian: `visualizeStoreHeader` and `visualizeStoreFooter` are called in the same way. They return the String representation of the header and the footer for "the given store" which is the parameter in `visualizeStore(Store store)`.

xingweitian: What about I remove the "store" part, will it be helpful?

Something like:
```java
  /**
     * Called by {@code CFAbstractStore#internalVisualize()} to visualize a local variable.
     *
     * @param localVar the local variable
     * @param value the value of the local variable
     * @return the String representation of the local variable
     */
```

xingweitian:
@mernst I have restored and simplified some comments, maybe it looks better now. Could you please have another look and give me some comments?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
```diff
@@ -0,0 +1,455 @@
...
+        return sbDigraph.toString();
+    }
+
+    /**
+     * Helper method called by {@link #generateGraphHelper(ControlFlowGraph, Block, Analysis)}. It
+     * checks the successors of the current block, and, if possible, adds all the successors to the
+     * work list and the visited blocks list.
+     *
+     * @param cur the current block
+     * @param visited the set of blocks that have already been visited
```
mernst: Or are in the worklist?
If so, please update the comment here and throughout, anywhere else that `@param visited` is documented.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
```diff
@@ -0,0 +1,442 @@
...
+    }
+
+    /**
+     * Helper method to visualize a block.
+     *
+     * @param bb the block
+     * @param analysis the current analysis
+     * @param cbFooter footer for a conditional block
+     * @param osFooter footer for other situations
+     * @param escapeCharacter the escape string to use
```
mernst: To use for what?  Please document.  I think it might be for left-justification or for centering.
Also, it's weird that this is named escape**Character** when its type is `String`.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
```diff
@@ -0,0 +1,442 @@
...
+     * @param bb the block
+     * @param analysis the current analysis
+     * @param separator the separator String to use
+     * @return the String representation of the contents of the block
+     */
+    protected String loopOverBlockContents(
+            Block bb, @Nullable Analysis<A, S, T> analysis, String separator) {
+
+        StringBuilder sbBlockContents = new StringBuilder();
+        boolean notFirst = false;
```
mernst: Replace the `notFirst` logic by use of StringJoiner.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
```diff
@@ -0,0 +1,442 @@
...
+                lastNode = null;
+        }
+        return lastNode;
+    }
+
+    /**
+     * Generate the order of processing blocks.
+     *
+     * @param cfg the current control flow graph
+     * @return the IdentityHashMap which maps from blocks to their orders
```
mernst: Say a bit more about this.  Is the mapped-to value an integer in the range [1..n]?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
```diff
@@ -0,0 +1,442 @@
...
+     */
+    protected IdentityHashMap<Block, List<Integer>> getProcessOrder(ControlFlowGraph cfg) {
```
mernst: Why is the mapped-to value a list?  Is it always a singleton list?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
```diff
@@ -0,0 +1,418 @@
...
+     * @return the String representation of the footer of the control flow graph
+     */
+    protected abstract String visualizeGraphFooter();
+
+    /**
+     * Return the simple String of the process order of a node. Called by {@link #generateNodes(Set,
+     * ControlFlowGraph, Analysis)}.
+     *
+     * @param order the list of the process order to be processed.
+     * @return the String representation of the process order of the node
```
mernst: Document the format of the output.  Since the input is a list, could this be a sequence of numbers?  In that case, what order are nodes processed in?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
```diff
@@ -0,0 +1,418 @@
...
+     *
+     * @return the String representation of the footer of the control flow graph
+     */
+    protected abstract String visualizeGraphFooter();
+
+    /**
+     * Return the simple String of the process order of a node. Called by {@link #generateNodes(Set,
+     * ControlFlowGraph, Analysis)}.
+     *
+     * @param order the list of the process order to be processed.
```
mernst: Javadoc tags should end with a period only if they are a complete sentence.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/DOTCFGVisualizer.java
```diff
@@ -38,442 +27,272 @@
...
-        }
-
-        generateDotNodes(visited, cfg, analysis);
-
-        // footer
-        this.sbDigraph.append("}\n");
+    /**
+     * Visualize the nodes of a control flow graph into a string.
+     *
+     * @param visited the set of all the blocks
```
mernst: If it is *all* the blocks, why is it named `visited`?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/DOTCFGVisualizer.java
```diff
@@ -38,442 +27,272 @@
...
+        return "    " + sId + " -> " + eId + " [label=\"" + flowRule + "\"];" + lineSeparator;
+    }
 
-        generated.put(srcloc.toString(), out);
+    @Override
+    public String visualizeBlock(Block bb, @Nullable Analysis<A, S, T> analysis) {
+        return super.visualizeBlockHelper(
+                bb,
+                analysis,
+                " \",];" + lineSeparator,
```
mernst: It is very confusing that the `"` and `[` of `visualizeNodes` is closed here in a different method, without any documentation about the relationship between them or their contracts.  I suggest that you move that part of these strings into the `visualizeNodes` routine.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
xingweitian: Thanks for the comments!
Here is what I think. We traverse the control flow graph to get all the blocks, it's like breadth first search. So `visited` is to store all the blocks that have been visited, `workList` is to store the blocks that will be processed in the future ( to check the successors ).
So I think maybe for now, it is suitable.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
xingweitian: `cfg.getDepthFirstOrderedBlocks()` will be called in method `getProcessOrder`, it will return the set of basic blocks in the control flow graph. But, blocks may appear more than once in the set (it is documented in `getDepthFirstOrderedBlocks()`), so I think that's why we use a list instead of a integer.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
mernst: My guess from reading the code is that visited is a superset of workList.
For example, addBlock always adds to visited whenever adding to workList.

You are saying that my understanding is incorrect.  Can you point out when visited does *not* contain a value that is on the worklist?
Also, if that is not the case, then workList may contain duplicates and the code is buggy. It may process nodes more than once, because the code that traverses workList does not check for duplicates.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
mernst: Clients won't read the documentation of `getDepthFirstOrderedBlocks()`.  (They won't even know to do so!)  Please make the documentation of the method stand on its own.

For example, if the list is not a singleton, is the block processed multiple times?  Does it get visualized multiple times even though it exists only once in the graph?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
xingweitian: Sorry for the misunderstanding, I think you are right, `visited` is to log all the blocks.
Can we say "the set of blocks that are in the workList" ? Since the number of blocks in visited is always larger than the number of blocks in workList. Also, the entry block never exists in workList. 

Is it suitable to say "the set of blocks that have already logged"?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
xingweitian: If I use Char instead of String, then I cannot use it as an argument to StringJoiner. Also I can only use '\n' instead of `System.lineSeparator()`. So I think maybe it is better to select another name.
If you don't mind, could you please give me an advise?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
xingweitian: Yes, it actually is a sequence of numbers by this version of code.

For example:
`Process order: 23,25`

I think the block with multiple process orders will not be processed multiple times, it is only because that `getDepthFirstOrderedBlocks()` returns it twice. Also, it will be visualized only one time. But, we have two numbers when printing process orders.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
mernst: If there is a clear, documented concept of "logged", that would be OK.
But I think it's clearer to say "the set of blocks that have already been visited or are in the worklist".
Do you have an alternate suggestion?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
mernst: You could rename the variable to `escapeString`.  I'm just asking for the name and the type to be consistent.  You can do that by renaming or by changing the type.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
mernst: Are the multiple numbers in the process order ever used?  From what you said, it sounds like they are not.

xingweitian: If we pass the argument "verbose" to the program, then the program will print the process order.

I tested it with a program, and some blocks actually have two process orders. Although I don't know if two numbers will be useful, but it indeed happens.

Here is an example of output:
```
285:
Process order: 23, 25
Before: then=org.checkerframework.checker.nullness.NullnessStore (
  all > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=boolean}
  i > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=int}
  val > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=java.lang.String}
  c > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=char}
  val.length() > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=int}
  val.charAt(i) > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=char}
  initialized fields = []
  invariant fields = {}
  isPolyNonNull = false
), else=org.checkerframework.checker.nullness.NullnessStore (
  all > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=boolean}
  i > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=int}
  val > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=java.lang.String}
  c > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=char}
  val.length() > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=int}
  val.charAt(i) > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=char}
  initialized fields = []
  invariant fields = {}
  isPolyNonNull = false
)
~~~~~~~~~
((WideningConversion(c, int) == WideningConversion(0, int)) || (WideningConversion(c, int) == WideningConversion(1, int)))   [ ConditionalOr ] > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=boolean}
all = ((WideningConversion(c, int) == WideningConversion(0, int)) || (WideningConversion(c, int) == WideningConversion(1, int)))   [ Assignment ] > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=boolean}
i   [ LocalVariable ] > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=int}
1   [ IntegerLiteral ] > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=int}
(i + 1)   [ NumericalAddition ] > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=int}
i = (i + 1)   [ Assignment ] > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=int}
tempPostfix#num1   [ VariableDeclaration ]
tempPostfix#num1   [ LocalVariable ] > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.UnknownInitialization, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=int}
tempPostfix#num1 = i   [ Assignment ] > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=int}
tempPostfix#num1   [ LocalVariable ] > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=int}
~~~~~~~~~
After: org.checkerframework.checker.nullness.NullnessStore (
  all > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=boolean}
  i > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=int}
  val > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=java.lang.String}
  c > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=char}
  tempPostfix#num1 > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=int}
  val.length() > CFAbstractValue{annotations=[@org.checkerframework.checker.initialization.qual.Initialized, @org.checkerframework.checker.nullness.qual.NonNull], underlyingType=int}
  initialized fields = []
  invariant fields = {}
  isPolyNonNull = false
)
```

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
xingweitian: Thanks for your kind advise! That sounds good, I will do it.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
mernst: Thanks!  Could you add a brief note indicating what you learned?
Then that is more than enough for this pull request.
Thanks a lot for your investigation!

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
xingweitian: That sounds great! I will do it.

xingweitian:
@mernst Thanks for the advise! I just updated this PR, could you please have another look?

mernst:
Thanks!

PR #2552, opened 2019-06-13
Block commit if Javadoc is missing

mernst:


mernst:
@smillst I enabled this only for the master branch.

PR #2554, opened 2019-06-14
Pull from upstream

mernst:
This pull request should not be squash-and-merged, but merged regularly.


mernst:
Opened against wrong repository.  Oops.

PR #2556, opened 2019-06-16
Add @Deprecated to methods no longer used

gokaco:
SignednessUtil.java methods which should no longer be used are deprecated.


checker/src/main/java/org/checkerframework/checker/signedness/SignednessUtil.java
```diff
@@ -372,11 +378,11 @@ public static int compareUnsigned(@Unsigned long x, @Unsigned long y) {
     /**
      * Compares two unsigned ints x and y.
      *
-     * <p>This is a reimplementation of Java 8's {@code Integer.compareUnsigned(int, int)}.
```
mernst: I don't think you need to remove comments like this, which are still helpful.
This pull request can just add the deprecation information.

checker/src/main/java/org/checkerframework/checker/signedness/SignednessUtil.java
```diff
@@ -358,11 +364,11 @@ public static void getUnsigned(ByteBuffer b, @Unsigned byte[] bs) {
     /**
      * Compares two unsigned longs x and y.
      *
-     * <p>This is a reimplementation of Java 8's {@code Long.compareUnsigned(long, long)}.
-     *
      * @return a negative number iff x {@literal <} y, a positive number iff x {@literal >} y, and
      *     zero iff x == y.
+     * @deprecated use Java 8's {@code Long.compareUnsigned(long, long)}.
```
mernst: Elsewhere you used `@link`, which is useful in the Javadoc.  Why did you use `@code` here?

gokaco: It was written earlier like this in the doc. Should I change it to ```@link```?

mernst: I see.  The previous writing isn't your fault, but it would still be nice to make it a link.
Thanks!

mernst:
The Travis job is failing because of some mal-formed javadoc:
https://travis-ci.org/typetools/checker-framework/jobs/546922328
Could you please correct that so this can be merged?

gokaco:
@mernst I think now it should pass all the tests.

PR #2558, opened 2019-06-17
BaseTypeVisitor: scan ParameterizedTypeTrees

smillst:
BaseTypeVisitor should scan ParameterizedTypeTree so that all trees are scanned by the visitor.  This way, subclasses can implement their own logic, such as validating explicit annotations.


wmdietl:
I'm confused by the description of the PR relative to the actual change. BaseTypeVisitor should not visit type trees to validate them, TypeValidators do that part. But we still need to traverse all type trees, to make sure annotations within them get scanned.
With this change, `visitParameterizedType` doesn't call super and doesn't do anything. So `visitAnnotation` won't be called on an annotation that is a type argument of some type.
Could you clarify the PR description to make clear what you want to achieve?

wmdietl:
D'oh. Misread the diffs. This looks good!

PR #2562, opened 2019-06-19
Add @SignedPositive as an alias for @SignednessGlb

mernst:


checker/src/main/java/org/checkerframework/checker/signedness/qual/SignednessBottom.java
```diff
@@ -17,7 +17,7 @@
  * @checker_framework.manual #signedness-checker Signedness Checker
  * @checker_framework.manual #bottom-type the bottom type
  */
-@SubtypeOf({SignednessGlb.class})
+@SubtypeOf({SignednessGlb.class, SignedPositive.class})
```
smillst: `SignednessPosstive` should not be here since it's an alias.

mernst: Thanks for catching that.

PR #2566, opened 2019-06-21
Improve diagnostics for the Subtyping Checker

mernst:


framework/src/main/java/org/checkerframework/common/subtyping/SubtypingAnnotatedTypeFactory.java
```diff
@@ -49,17 +49,30 @@ protected AnnotationClassLoader createAnnotationClassLoader() {
...
+                        loader.loadExternalAnnotationClassesFromDirectory(dirName);
+                if (annos.isEmpty()) {
+                    throw new UserError("No qualifiers found in " + dirName);
+                }
+                qualSet.addAll(annos);
             }
         }
 
+        if (qualSet.isEmpty()) {
+            throw new UserError("SubtypingChecker: no qualifiers specified");
```
wmdietl: The three user errors have different shapes, one starting with a lower case letter, the next with upper case, and this final one includes the checker name.
Always including the checker name seems useful.

mernst: I added more specific command-line argument references, and I think that improves the messages.
I also fixed the capitalization.
Thanks for those suggestions.

The first two messages refer to a command-line argument the user provided incorrectly, so I think that the checker name is not necessary there, and I want to keep error messages short when possible.
The last is about command-line arguments that are missing, which is why I provided the checker name there.
Do you still think it's important to always give the checker name?

PR #2567, opened 2019-06-21
Removed annotations from the argument of write(int) in java.io.Writer and regionMatches() in java.lang.String 

Kakarot-SSJ4:
1.   The argument of write(int) need not be `@NonNegative.`
    Consider the following code:

```
import java.io.*;
class tryit{
    public static void main(String[] args) throws IOException {
        Writer a = new PrintWriter(System.out);
        a.write(-1);
    }
}
```

Compiling it by 'javac tryit.java' and executing it by 'java tryit'.
It gives no error and is run safely.

 2.   The `@IndexOrHigh` annotations on the parameters toffset, ooffset and len of the regionMatches() functions is not required as it is checked in the if statements inside the function and a false value is returned if the above annotation is not satisfied by the parameters.

Consider the following code:

```
class tryit{
	String a = "abcd";
	System.out.println(a.regionMatches(-1,"abc",-2,-1));
  }
}
```

Compile it by javac tryit.java
Run it by java tryit
It prints false and throws no error.


Kakarot-SSJ4:
@kelloggm @panacekcz

panacekcz:
@Kakarot-SSJ4, as you can see in the travis build, tests are failing. Please look into the build log to see which tests are failing and, if the tests are incorrect, fix those tests. You should also be able to run the tests locally.

panacekcz:
Replaces #2555, #2548 and #2522.

panacekcz:
@Kakarot-SSJ4, I think you should leave in the test for `writer.write(-1)`, only remove the comment above it. It tests the property that you stated in point 1 of the first comment of this PR, so it makes sense to have it in the test suite.

kelloggm:
@Kakarot-SSJ4 there is no new JDK hash in this PR now, which is why the tests are failing.

Kakarot-SSJ4:
@kelloggm @panacekcz

PR #2568, opened 2019-06-21
Improve diagnostics for pre/post-increment/decrement; fixes #2485

### LINKED ISSUE #2485: Confusing error message "compound.assignment.type.incompatible" for pre/post increment/decrement
Consider the following code:

```

class PredecrementWarning {
  public static void warningDecrement(@IntRange(from = 1, to = 2) int j) {
    --j;
  }
}
```

When you compile this with

```javac -g PredecrementWarning.java -processor org.checkerframework.common.value.ValueChecker```

the Value Checker correctly issues an error.  Here is the error:

```
PredecrementWarning.java:5: error: [compound.assignment.type.incompatible] expression type incompatible with left-hand side in compound assignment.
    --j;
    ^
  found   : @IntVal({0, 1}) int
  required: @IntRange(from=1, to=2) int
```

This error message is confusing because it mentions the left-hand side of a compound assignment, but there is none in the Java program being compiled.

Error messages about preincrement, postincrement, predecrement, and postdecrement should be changed to mention the actual operator and not to mention a "left-hand side".

This requires creating new error keys, and probably not using different terms than "found:" and "required:".

mernst:
Merge together with https://github.com/typetools/guava/pull/32


smillst:
FYI I cherry-picked commit "Make Travis run a branch and fork of Guava"(f7b84c16dd39161d180286fc084eb67dd64cd8a4) to master.

PR #2571, opened 2019-06-24
IndexChecker treats single-argument `@IntVal` as constant value

### LINKED ISSUE #2484: Index Checker treats `4` differently than `@IntVal(4) int`
Consider the following code:

```java
import org.checkerframework.checker.index.qual.LTLengthOf;
import org.checkerframework.checker.interning.qual.Interned;
import org.checkerframework.common.value.qual.ArrayLen;

public class IndexIntValVsConstant {

  void m() {

    int @ArrayLen(7) [] a1 = new int[] {1, 2, 3, 4, 5, 6, 7};

    @IntVal(2) int i = 2;
    @IntVal(4) int j = 4;

    int[] s0 = internSubsequence(a1, 2, 4);
    int[] s1 = internSubsequence(a1, i, j);
  }

  int @Interned [] internSubsequence(
      int @Interned [] seq,
      @IndexFor("#1") @LessThan("#3") int start,
      @NonNegative @LTLengthOf(value = "#1", offset = "#2 - 1") int end) {
    // dummy implementation
    return new int[0];
  }
}
```

The output of 
`javac -g IndexIntValVsConstant.java -processor index`
is
```
IndexIntValVsConstant.java:19: error: [argument.type.incompatible] incompatible types in argument.
    int[] s1 = internSubsequence(a1, i, j);
                                        ^
  required: @LTLengthOf(value="a1", offset="i - 1") int
1 error
```

The first call to `internSubsequence`, which has constant arguments, is permitted, but the second call, which has `@IntVal(...)` arguments, is not.  Both should be treated the same.

Maxi17:
Fixes #2484.

When an offset is calculated, IndexChecker now treats variables annotated with single-argument `@IntVal` as constant, just like literal values. 

@kelloggm

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/index/upperbound/OffsetEquation.java
+++ b/checker/src/main/java/org/checkerframework/checker/index/upperbound/OffsetEquation.java
@@ -354,10 +354,7 @@ public class OffsetEquation {
             FlowExpressionContext context, TreePath scope, boolean useLocalScope)
             throws FlowExpressionParseException {
 
-        standardizeAndViewpointAdaptExpressions(
-                addedTerms, false, context, scope, useLocalScope, null);
-        standardizeAndViewpointAdaptExpressions(
-                subtractedTerms, true, context, scope, useLocalScope, null);
+        standardizeAndViewpointAdaptExpressions(context, scope, useLocalScope, null);
     }
 
     /**
--- a/checker/src/main/java/org/checkerframework/checker/index/upperbound/OffsetEquation.java
+++ b/checker/src/main/java/org/checkerframework/checker/index/upperbound/OffsetEquation.java
@@ -6,6 +6,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
+import javax.lang.model.element.AnnotationMirror;
 import org.checkerframework.checker.index.IndexUtil;
 import org.checkerframework.common.basetype.BaseAnnotatedTypeFactory;
 import org.checkerframework.common.value.ValueAnnotatedTypeFactory;
@@ -24,9 +25,6 @@ import org.checkerframework.framework.util.FlowExpressionParseUtil.FlowExpressio
 import org.checkerframework.framework.util.dependenttypes.DependentTypesError;
 import org.checkerframework.javacutil.AnnotationProvider;
 import org.checkerframework.javacutil.TreeUtils;
-
-import javax.lang.model.element.AnnotationMirror;
-import javax.lang.model.element.AnnotationValue;
 
 /**
  * An offset equation is 2 sets of Java expression strings, one set of added terms and one set of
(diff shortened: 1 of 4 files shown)


checker/src/main/java/org/checkerframework/checker/index/upperbound/OffsetEquation.java
```diff
@@ -266,6 +271,16 @@ private Integer evalConstantTerm(Receiver termReceiver) {
...
             }
+        } else if (factory != null && termReceiver instanceof FlowExpressions.LocalVariable) {
+            AnnotationMirror am =
+                    ((BaseAnnotatedTypeFactory) factory)
+                            .getTypeFactoryOfSubchecker(ValueChecker.class)
+                            .getAnnotatedType(
+                                    ((FlowExpressions.LocalVariable) termReceiver).getElement())
+                            .getAnnotation(IntVal.class);
+            if (am != null && ValueAnnotatedTypeFactory.getIntValues(am).size() == 1) {
+                return ValueAnnotatedTypeFactory.getIntValues(am).get(0).intValue();
```
smillst: You should create a version of `IndexUtil#getExactValue` that takes an element and call that method.

checker/src/main/java/org/checkerframework/checker/index/upperbound/OffsetEquation.java
```diff
@@ -246,7 +251,7 @@ public boolean isNegativeOrZero() {
      * Evaluates an offset term. If the term is an integer constant, returns its value. Otherwise,
      * returns null.
      */
-    private Integer evalConstantTerm(Receiver termReceiver) {
+    private Integer evalConstantTerm(Receiver termReceiver, AnnotatedTypeFactory factory) {
```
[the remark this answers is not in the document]
smillst: The Javadoc for this method should now say something about how it uses the factory.  (This same comment applies to every method for which you added a factory parameter.

Maxi17:
@smillst Thank you for the review. I updated it.

checker/src/main/java/org/checkerframework/checker/index/upperbound/OffsetEquation.java
```diff
@@ -266,6 +271,16 @@ private Integer evalConstantTerm(Receiver termReceiver) {
                     }
                 }
             }
+        } else if (factory != null && termReceiver instanceof FlowExpressions.LocalVariable) {
+            AnnotationMirror am =
+                    ((BaseAnnotatedTypeFactory) factory)
+                            .getTypeFactoryOfSubchecker(ValueChecker.class)
+                            .getAnnotatedType(
+                                    ((FlowExpressions.LocalVariable) termReceiver).getElement())
+                            .getAnnotation(IntVal.class);
```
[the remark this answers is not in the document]
smillst: Ping

checker/src/main/java/org/checkerframework/checker/index/upperbound/OffsetEquation.java
```diff
@@ -266,6 +274,13 @@ private Integer evalConstantTerm(Receiver termReceiver) {
...
                 }
             }
+        } else if (factory != null && termReceiver instanceof FlowExpressions.LocalVariable) {
+            Element element = ((FlowExpressions.LocalVariable) termReceiver).getElement();
+            AnnotationMirror am =
+                    factory.getTypeFactoryOfSubchecker(ValueChecker.class)
+                            .getAnnotatedType(element)
+                            .getAnnotation(IntVal.class);
+            return IndexUtil.getIntegerFromIntValWithOneArgument(am);
         }
```
smillst: My comment was
>You should create a version of `IndexUtil#getExactValue` that takes an element and call that method.

But that’s not what you did. Was there a reason that didn’t or won’t work?

Maxi17: The method I thought I should create is a static one that takes an `AnnotationMirror` and returns an integer representing its only `IntVal` argument. If the annotation is not `IntVal` or it has more than one argument, then it returns `null`. I think I misunderstood it.

Did you mean I need to create a method that takes an `Element` and returns its `IntVal` annotation?

smillst: I mean a method like:
```
 public static Long getExactValue(Element element, ValueAnnotatedTypeFactory factory) {
        AnnotatedTypeMirror valueType = factory.getAnnotatedType(element);
        Range possibleValues = getPossibleValues(valueType, factory);
        if (possibleValues != null && possibleValues.from == possibleValues.to) {
            return possibleValues.from;
        } else {
            return null;
        }
    }
```

Then call it:
```
Element element = ((FlowExpressions.LocalVariable) termReceiver).getElement();
Long l = getExactValue(element, factory.getTypeFactoryOfSubchecker(ValueChecker.class))
```
`l` might be null, so you'll need to deal with that case.

Maxi17: Thank you for the suggestion. I updated the class accordingly.

smillst:
I tried to push changes to this branch, but didn't have permission.  Please merge https://github.com/vmgama/checker-framework/pull/1.  (I think the problem is that this branch is in a different users fork than the user that created it; otherwise, "Allow edits from maintainers" would be checked by default.) 

Also, please update the title and pull request description to describe the changes, rather than just referencing the issue.  (The description should include "fixes #2484".)  Once that is done @kelloggm  can merge this.  Thanks!

Issue #2493, opened 2019-05-20
Wrong compiler interpretation of length properties


Kakarot-SSJ4:
While annotating Apache Commons Lang, I came across an issue which is reduced in the following code:
```
import org.checkerframework.checker.index.qual.*;
import org.checkerframework.common.value.qual.*;
class tryit{
public static void check_it(int a[], int b @SameLen("#1") []) {
        for(@IndexOrHigh("b") int i = 0; i < a.length ; i++);
}
}
```
Run it by typing in the terminal: javac -processor index tryit.java
Output:
tryit.java:5: error: [compound.assignment.type.incompatible] incompatible result type in compound assignment.
        for(@IndexOrHigh("b") int i = 0; i < a.length ; i++);
                                                                        
  found   : @LTLengthOf(value={"a", "b"}, offset={"-1", "-2"}) int
  required: @LTEqLengthOf("b") int
1 error

Clearly, i <= b.length and a.length, whereas the compiler interprets the properties wrongly.

PR #2572, opened 2019-06-25
Make SameLen annotations on parameters symmetric

### LINKED ISSUE #2434: SameLen property should be inferred on both variables
Consider the following code:
`import org.checkerframework.checker.index.qual.*;`
`class checkerr{`
`public void abc(String x, @SameLen("#1") String y){`
`}`
`public void err(@SameLen("#2") String a, String b){`
`abc(a,b);`
`}`
`}`
Run it by typing in the terminal:
javac -processor org.checkerframework.checker.index.IndexChecker checkerr.java

Output:
checkerr.java:6: error: [argument.type.incompatible] incompatible types in argument.
abc(a,b);
      ^
  found   : @SameLen("b") String
  required: @SameLen("a") String
1 error

Now if we annotate 'a' to be of the same length as of 'b', 'b' should automatically be inferred to have the same length as 'a', which clearly fails during above code.

kelloggm:
This pull request adds defaulting so that `@SameLen` annotations on formal that refer to other parameters are symmetric. For example, the declaration:
```java
void foo(int @SameLen("#2")[] a, int[] b)
```
is defaulted to
```java
void foo(int @SameLen("#2")[] a, int @SameLen("#1")[] b)
```

Fixes #2434.

@Maxi17 contributed significantly to this PR, and I would appreciate any comments he has on it.

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/index/samelen/SameLenTransfer.java
+++ b/checker/src/main/java/org/checkerframework/checker/index/samelen/SameLenTransfer.java
@@ -266,6 +266,7 @@ public class SameLenTransfer extends CFTransfer {
             UnderlyingAST.CFGMethod method,
             MethodTree methodTree,
             ExecutableElement methodElement) {
+        super.addInformationFromPreconditions(info, factory, method, methodTree, methodElement);
         List<? extends VariableTree> paramTrees = methodTree.getParameters();
 
         List<String> paramNames =
--- a/checker/src/main/java/org/checkerframework/checker/index/samelen/SameLenTransfer.java
+++ b/checker/src/main/java/org/checkerframework/checker/index/samelen/SameLenTransfer.java
@@ -6,7 +6,6 @@ import com.sun.source.util.TreePath;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import java.util.stream.Collectors;
 import javax.lang.model.element.AnnotationMirror;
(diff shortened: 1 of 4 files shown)


smillst:
FYI: GitHub allow co-authors: https://help.github.com/en/articles/creating-a-commit-with-multiple-authors.

Maxi17:
It looks great. There is another issue related to this PR whose test case now passes: #2493.

checker/src/main/java/org/checkerframework/checker/index/samelen/SameLenTransfer.java
```diff
@@ -249,4 +256,62 @@ AnnotationMirror getAnno(Node n) {
...
+
+                // the SameLen value is in the list of params, so modify the type of
+                // that param in the store
+                AnnotationMirror newSameLen =
+                        aTypeFactory.createSameLen(
+                                Collections.singletonList(paramNames.get(index)));
+                Receiver otherParamRec = null;
+                try {
+                    otherParamRec =
+                            aTypeFactory.getReceiverFromJavaExpressionString(
```
smillst: You have the tree, so you should be able to get the receiver from that.  Try 
https://checkerframework.org/api/org/checkerframework/framework/util/FlowExpressionParseUtil.html#internalReprOfVariable-org.checkerframework.framework.type.AnnotatedTypeFactory-com.sun.source.tree.VariableTree-

PR #2580, opened 2019-06-27
Fix `@LessThan` false warning on OffsetEquation parameters

### LINKED ISSUE #2541: Index Checker cannot compare `@LessThan("2+1")` against `@LessThan("3")`
The Index Checker issues a false positive warning for the following code.

```
public class LessThanConstantAddition {

    public static void checkedPow(int b) {
        if (b <= 2) {
            int c = (int) b;
        }
    }
}
```
Its output is:
```
javac -processor index LessThanConstantAddition.java
LessThanConstantAddition.java:5: warning: [cast.unsafe] cast from "@LessThan("2 + 1") int" to "@LessThan("3") int" cannot be statically verified
            int c = (int) b;
                    ^
1 warning
```

This comes up in Guava.

Maxi17:
Before checking for valid casting, `@LessThan` arguments are simplified if possible to avoid `[cast.unsafe]` false warnings.

If required, a new method that performs the argument simplification can be written.

Fixes #2541 

@kelloggm


framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1869,7 +1869,7 @@ protected void checkTypecastSafety(TypeCastTree typeCastTree) {
      * @param exprType annotated type of the casted expression
      * @return true if the type cast is safe, false otherwise
      */
-    private boolean isTypeCastSafe(AnnotatedTypeMirror castType, AnnotatedTypeMirror exprType) {
+    protected boolean isTypeCastSafe(AnnotatedTypeMirror castType, AnnotatedTypeMirror exprType) {
```
kelloggm: If you're going to make this method protected, why not do the modifications to the types above in an override of this method, instead.

That would have two advantages:
* you would not need to override `checkTypecastSafety`, and so you wouldn't need to duplicate its logic
* the implementation would be simpler, since you would already have the two `AnnotatedTypeMirror`s. You could then just make the changes and then call super with the modified types

Maxi17: Thanks. I pushed a new commit with the changes.

checker/src/main/java/org/checkerframework/checker/index/inequality/LessThanVisitor.java
```diff
@@ -84,4 +86,27 @@ protected void commonAssignmentCheck(
...
     }
+
+    @Override
+    protected boolean isTypeCastSafe(AnnotatedTypeMirror castType, AnnotatedTypeMirror exprType) {
+
+        AnnotationMirror exprLTAnno =
+                exprType.getEffectiveAnnotationInHierarchy(atypeFactory.UNKNOWN);
+
+        if (exprLTAnno != null) {
+            List<String> initialAnnotations =
```
smillst: This might be `null`, see documentation of LessThanAnnotatedTypeFactory.getLessThanExpressions.

Maxi17:
The commits were messed up because when I pulled the latest changes, AnnotationConverter.java was changed and annotationMirrorToAnnotation(AnnotationMirror am) from that had a bug.

PR #2592, opened 2019-07-01
Fix position of -q option in git pull command

panacekcz:
When running `./gradlew cloneAndBuildDependencies`, git fails to pull recent changes from annotation-tools, due to unkown `-q` option.
This was discovered because due to recent changes, up-to-date version of annotation-tools is required, but running `cloneAndBuildDependencies` does not help.


mernst:
Thanks for noticing and fixing this!

PR #2593, opened 2019-07-01
Update manual to reflect new Interning Checker semantics.

smillst:


docs/manual/interning-checker.tex
```diff
@@ -118,9 +126,11 @@
   for such a class, \<==> and \<equals> are equivalent, and so the
   Interning Checker does not issue errors or warnings for either one.
 
+\item[\refqualclass{checker/interning/qual}{InternMethod}]
+  is a method annotation that indicates that this method returns an interned object and maybe be invoked
```
kelloggm: Does "method annotation" have a well-accepted definition? I would have said "declaration annotation" and then mention elsewhere that it should only be used on methods.

docs/manual/interning-checker.tex
```diff
@@ -148,40 +158,95 @@
...
+type are interned \textit{except for newly created objects}. That means that
+all uses of such types are \<@Interned> by default and the type \<@UnknownInterned
+MyClass> is an invalid type.
+
+An exception is \textit{constructor results}. Constructor results and \<this> within the
+body of the constructor are \<@UnknownInterned> by default, but no ``type.invalid''
+error is issued at the declaration. (It would be issued because
+\<@UnknownInterned InternedClass> is not a legal type, given the
+qualifier \<@Interned> on \<InternedClass>.)  Instead, an ``interned.object.creation''
+error is issued at the invocation of the constructor. The user should inspect
```
kelloggm: This explanation is somewhat confusing, because it leads with what the checker *does not* do, rather than what it does do. I would first explain what the `interned.object.creation` error means, and then explain that no `type.invalid` error will be issued.

docs/manual/interning-checker.tex
```diff
@@ -148,40 +158,95 @@
...
 
 \begin{Verbatim}
-  public @Interned class MyInternedClass { ... }
+@Interned class InternedClass {
+  @UnknownInterned InternedClass() {
+  // error, "this" is @UnknownInterned.
+  @Interned InternedClass that = this;
+  }
+
+  @SuppressWarnings("intern") // ONE is interned for some reason.
```
kelloggm: This is a bit confusing, too, because it's very vague. I'd give an example of a reason, rather than saying "some reason"

docs/manual/interning-checker.tex
```diff
@@ -148,40 +158,95 @@
...
-\end{smaller}
-%END LATEX
+
+
+\subsectionAndLabel{Default qualifiers and qualifiers for literals}{interning-implicit-qualifiers}
+
+The Interning Checker
+adds qualifiers to unannotated types, reducing the number of annotations that must
+appear in your code (see Section~\ref{effective-qualifier}).
+For example, String literals and the \<null> literal are always considered interned.
```
kelloggm: Are there any others besides String literals and null? I'd give the complete list here, since otherwise this feels vague.

PR #2594, opened 2019-07-01
Added NonNegative default annotation to `char`

### LINKED ISSUE #2540: Cast from char to int loses IntRange information
When a char is cast to an int (including via arithmetic promotion, which applies to the arguments of operators such as <<), the Index Checker should keep track of the value's bounds.  That is not currently the case.  (Or is this a problem with the Constant Value Checker?)

Here is a test case, which is commetted to the repository as `checker/tests/index/CharToIntCast.java`.

```

public class CharToIntCast {

    public static void charRange(char c) {
        @IntRange(from = 0, to = Character.MAX_VALUE) int i = c;
    }

    public static void charShift(char c) {
        char c2 = (char) (c >> 4);
    }

    public static void rangeShiftOk(@IntRange(from = 0, to = Character.MAX_VALUE) int i) {
        char c2 = (char) (i >> 4);
    }
}
```

The code should typecheck, but currently the first two methods do not type-check (the last one does).

This comes up in Guava.

Maxi17:
Every `char` method parameter is now `@NonNegative` by default.
In an assignment, this is further improved to `@IntRange(from = 0, to = Character.MAX_VALUE)` if necessary.

Related: #1675, kelloggm#192.

Fixes #2540

Two test cases were written by @panacekcz.

@kelloggm

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/index/lowerbound/LowerBoundTransfer.java
+++ b/checker/src/main/java/org/checkerframework/checker/index/lowerbound/LowerBoundTransfer.java
@@ -4,9 +4,12 @@ import static org.checkerframework.checker.index.IndexUtil.getExactValue;
 
 import com.sun.source.tree.MemberSelectTree;
 import com.sun.source.tree.MethodInvocationTree;
+import com.sun.source.tree.MethodTree;
 import com.sun.source.tree.Tree;
+import com.sun.source.tree.VariableTree;
 import java.util.List;
 import javax.lang.model.element.AnnotationMirror;
+import javax.lang.model.element.ExecutableElement;
 import org.checkerframework.checker.index.IndexAbstractTransfer;
 import org.checkerframework.checker.index.IndexRefinementInfo;
 import org.checkerframework.checker.index.qual.GTENegativeOne;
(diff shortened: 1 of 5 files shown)


checker/src/main/java/org/checkerframework/checker/index/lowerbound/LowerBoundTransfer.java
```diff
@@ -693,6 +699,36 @@ private AnnotationMirror addAnnotationForLiteralRemainder(int val) {
...
+            AnnotatedTypeFactory factory,
+            UnderlyingAST.CFGMethod method,
+            MethodTree methodTree,
+            ExecutableElement methodElement) {
+        super.addInformationFromPreconditions(info, factory, method, methodTree, methodElement);
+
+        List<? extends VariableTree> paramTrees = methodTree.getParameters();
+
+        for (VariableTree variableTree : paramTrees) {
+            if (variableTree.getType().toString().equalsIgnoreCase("char")) {
```
Maxi17: This is a poor way to check that the variable is a `char`, but couldn't figure out another way.

kelloggm: You can use `TreeUtils.typeOf(variableTree).getKind() == TypeKind.CHAR` instead

Maxi17:
Thanks! I made small improvements and added some Javadoc, preparing it for merging.

kelloggm:
Looks good. I'll merge this once the tests pass.

PR #2598, opened 2019-07-02
getClass() should not create an invalid type

smillst:


framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -2147,7 +2147,19 @@ protected void adaptGetClassReturnTypeToReceiver(
         // arguments.  So, we just copy the annotations from the bound of the declared type to the
         // new bound.
         final AnnotatedWildcardType classWildcardArg = (AnnotatedWildcardType) typeArgs.get(0);
-        newBound.replaceAnnotations(classWildcardArg.getExtendsBound().getAnnotations());
+        Set<AnnotationMirror> newAnnos = AnnotationUtils.createAnnotationSet();
+        Set<AnnotationMirror> typeBoundAnnos = getTypeDeclarationBounds(newBound.getUnderlyingType());
+        for (AnnotationMirror typeBoundAnno : typeBoundAnnos) {
+            AnnotationMirror wildcardAnno =
+                    qualHierarchy.findAnnotationInSameHierarchy(
+                            classWildcardArg.getExtendsBound().getAnnotations(), typeBoundAnno);
```
wmdietl: Can you move the `classWildcardArg.getExtendsBound().getAnnotations()` subexpression to outside of the `for` loop? It doesn't change and moving would make the code a bit easier to read.

Issue #1669, opened 2017-11-17
Imprecise refinement in Value Checker in loops


kelloggm:
Consider the following loop:

```java
    void test() {
        for (int i = 0; i < 12; i++) {
            @IntRange(from = 0, to = 12) int j = i;
        }
    }
```
I would expect this code to typecheck. Instead, the following error is issued:

```
error: [assignment.type.incompatible] incompatible types in assignment.
            @IntRange(from = 0, to = 12) int j = i;
                                                 ^
  found   : @IntRange(from=0, to=127) int
  required: @IntRange(from=0, to=12) int
1 error
```

This is very strange behavior - I don't know why the Value Checker might think that 0 to 12 -> 0 to 127.

mernst:
I suspect that this is happening because the Value Checker, after 10 iterations through the loop, widens the bound and chooses 127 as the widened upper bound.

Two possible solutions would be:
 * When analyzing a `for` loop with constant upper and lower bounds, start out the estimate at those bounds, rather than starting out the estimate at the initial value.  This would probably be adequate in most cases, and it would improve performance by eliminating the need to widen the estimate.
 * When widening, if the value has been compared to a constant, use that constant (or a value 1 less than that constant) as the widened bound.

Issue #602, opened 2016-02-19
Refined types in loops should not be cached


smillst:

smillst:
framework/tests/value/rRepo.java issues an error it should not because of this issue.

smillst:
Turing off caching in dataflow with the body of a loop is trick because data flow works on CGF which do not contain information about loops.  So, instead of turning off caching only during loops, we will start by turning off caching during dataflow.

Implementation plan:
1. Cache all expressions in method bodies
2. Turn off caching during data flow.

PR #2599, opened 2019-07-03
Refine IntRange annotations in conditional expressions inside for loops

### LINKED ISSUE #2345: Invalid conditional.type.incompatible error
Consider the following code:
`import java.io.IOException;
import java.io.Reader;

class conderr extends Reader{
public Reader abc;
public conderr(Reader xyz){
	this.abc = xyz;
}
	abc.close();
}
public int read( final char[] cbuf, final @NonNegative int off, final int len ) throws IOException { // offset value has to be non negative
        int c;
        for (@NonNegative int i = 0; i < len; i++ ) {
            c = super.read();
            if ( c == -1 ) {
                return i == 0 ? -1 : i;
            }
        }
        return len;
    }
}`
Type in the terminal: javac -processor org.checkerframework.checker.index.IndexChecker conderr.java
Output:
conderr.java:14: error: [override.return.invalid] Incompatible return type.
public int read( final char[] cbuf, final @NonNegative int off, final int len ) throws IOException { // offset value has to be non negative
       ^
  Method
    @LowerBoundUnknown int read(@LowerBoundUnknown conderr this, @LowerBoundUnknown char @LowerBoundUnknown [] p0, @NonNegative int p1, @LowerBoundUnknown int p2) throws @LowerBoundUnknown IOException in conderr
  cannot override
    @GTENegativeOne int read(@LowerBoundUnknown Reader this, @LowerBoundUnknown char @LowerBoundUnknown [] p0, @NonNegative int p1, @NonNegative int p2) throws @LowerBoundUnknown IOException in java.io.Reader
  found   : @LowerBoundUnknown int
  required: @GTENegativeOne int
conderr.java:14: error: [override.return.invalid] Incompatible return type.
public int read( final char[] cbuf, final @NonNegative int off, final int len ) throws IOException { // offset value has to be non negative
       ^
  Method
    @UpperBoundUnknown int read(@UpperBoundUnknown conderr this, @UpperBoundUnknown char @UpperBoundUnknown [] p0, @UpperBoundUnknown int p1, @UpperBoundUnknown int p2) throws @UpperBoundUnknown IOException in conderr
  cannot override
    @LTEqLengthOf("#1") int read(@UpperBoundUnknown Reader this, @UpperBoundUnknown char @UpperBoundUnknown [] p0, @LTEqLengthOf("#1") int p1, @LTEqLengthOf("#1") int p2) throws @UpperBoundUnknown IOException in java.io.Reader
  required: @LTEqLengthOf("#1") int
**conderr.java:19: error: [conditional.type.incompatible] incompatible types in conditional expression.
                return i == 0 ? -1 : i;
                                     ^
  found   : @IntRangeFromNonNegative int
  required: @IntVal(-1) int**
3 errors

Notice the last error. There is no need for i to be annotated with @IntVal(-1), it can take any non negative integral value. I think it is a bug.
@kelloggm

Maxi17:
In a for loop, `@IntRange` annotations can be unpredictable (#1669). In the test case, `i` had `@IntRange(from = Integer.MIN_VALUE, to = Integer.MAX_VALUE)` annotation, which was wrong to begin with (`i` is greater than 0) and redundant, because it is an `int`.

The actual problem was that the type of the conditional expression was `IntVal(-1)`, which is again wrong. It is supposed to be the LUB of true and false expressions, in this case simply `int`. Here is the bug in CF:

1. The LUB of `@IntVal(-1) int` and `int` is `int` (correct).
2. The LUB of `@IntVal(-1) int` and `@IntRange(from = Integer.MIN_VALUE, to = Integer.MAX_VALUE) int` is `@IntVal(-1) int` (wrong).

In the for loop from the test case, the second case is applied and therefore the bug.

This PR fixes this by changing the ValueChecker annotation of the conditional expression if it is annotated. The updated annotation matches the initial annotation given to the conditional expression in `visitConditionalExpression(ConditionalExpressionTree, AnnotatedTypeFactory)` from `TypeFromExpressionVisitor.java`.

Fixes #2345 

@kelloggm

### DIFF
--- a/framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
@@ -2223,15 +2223,19 @@ public class ValueAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
 
     @Override
     public AnnotatedTypeMirror getAnnotatedType(Tree tree) {
-        if (tree instanceof ConditionalExpressionTree) {
+        AnnotatedTypeMirror atm = super.getAnnotatedType(tree);
+
+        if (atm.isAnnotatedInHierarchy(UNKNOWNVAL) && tree instanceof ConditionalExpressionTree) {
             AnnotatedTypeMirror atmTrue =
                     getAnnotatedType(((ConditionalExpressionTree) tree).getTrueExpression());
             AnnotatedTypeMirror atmFalse =
                     getAnnotatedType(((ConditionalExpressionTree) tree).getFalseExpression());
             TypeMirror alub = TreeUtils.typeOf(tree);
-            return AnnotatedTypes.leastUpperBound(this, atmTrue, atmFalse, alub);
+            atm.replaceAnnotation(
+                    AnnotatedTypes.leastUpperBound(this, atmTrue, atmFalse, alub)
+                            .getEffectiveAnnotationInHierarchy(UNKNOWNVAL));
         }
 
-        return super.getAnnotatedType(tree);
+        return atm;
     }
 }
--- a/framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
@@ -67,7 +67,6 @@ import org.checkerframework.framework.type.treeannotator.PropagationTreeAnnotato
 import org.checkerframework.framework.type.treeannotator.TreeAnnotator;
 import org.checkerframework.framework.type.typeannotator.ListTypeAnnotator;
 import org.checkerframework.framework.type.typeannotator.TypeAnnotator;
-import org.checkerframework.framework.util.AnnotatedTypes;
 import org.checkerframework.framework.util.FieldInvariants;
 import org.checkerframework.framework.util.FlowExpressionParseUtil.FlowExpressionParseException;
 import org.checkerframework.framework.util.MultiGraphQualifierHierarchy;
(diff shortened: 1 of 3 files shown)


framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -2218,4 +2220,22 @@ public int getMinLenFromString(String sequenceExpression, Tree tree, TreePath cu
...
         return getMinLenValue(lengthAnno);
     }
+
+    @Override
+    public AnnotatedTypeMirror getAnnotatedType(Tree tree) {
+        AnnotatedTypeMirror atm = super.getAnnotatedType(tree);
+
+        if (atm.isAnnotatedInHierarchy(UNKNOWNVAL) && tree instanceof ConditionalExpressionTree) {
+            AnnotatedTypeMirror atmTrue =
+                    getAnnotatedType(((ConditionalExpressionTree) tree).getTrueExpression());
```
kelloggm: I would move this cast into a local variable to avoid the repetition. That might also keep these lines from being too long.

smillst:
This is a symptom of https://github.com/typetools/checker-framework/issues/602.  Actually fixing that issue is a big task, so you should add a work around for the Value Checker, like you have done here.  

I would suggest a different work around, though.   I would add the following to ValueTreeAnnotator:

```
        @Override
        public Void visitConditionalExpression(ConditionalExpressionTree node,
                AnnotatedTypeMirror annotatedTypeMirror) {
            // Work around for https://github.com/typetools/checker-framework/issues/602.
            annotatedTypeMirror.replaceAnnotation(UNKNOWNVAL);
            return null;
        }
```

That way the annotated type factory will cache `@UnknownValue` and dataflow can then refine that value each time it analyses the conditional expressions.

Maxi17:
@smillst I hope I understood correctly your suggestion. If the annotation is replaced with `@UnknownVal`, then I see no harm in returning `super.visitConditionalExpression(node, annotatedTypeMirror)`. It will eventually return `null` anyways.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1452,6 +1451,14 @@ private boolean handledByValueChecker(AnnotatedTypeMirror type) {
             */
             return COVERED_CLASS_STRINGS.contains(tm.toString());
         }
+
+        @Override
+        public Void visitConditionalExpression(
+                ConditionalExpressionTree node, AnnotatedTypeMirror annotatedTypeMirror) {
+            // Work around for https://github.com/typetools/checker-framework/issues/602.
+            annotatedTypeMirror.replaceAnnotation(UNKNOWNVAL);
+            return super.visitConditionalExpression(node, annotatedTypeMirror);
```
smillst: TreeAnnotators don't visit their subtrees, so calling super does nothing but return null.  This method should not call super like the others in this class.  If you leave the super call, a reader has to figure out that it does nothing.

PR #2600, opened 2019-07-04
Use TypeAnnotationUtils.unannotatedType() to beautify the output of error [type.invalid.annotations.on.use].

xingweitian:
Hello, this PR improves the output of the error `type.invalid.annotations.on.use`.

If we execute the following command in the directory of check-framework:

```
javacheck -processor nullness checker/tests/nullness/Issue2247.java
```

We will get an output like this:

```
checker/tests/nullness/Issue2247.java:16: error: [type.invalid.annotations.on.use] invalid type:
        annotations [@Initialized, @Nullable] conflict with declaration of type
        (@org.checkerframework.checker.nullness.qual.Nullable :: Issue2247.DeclaredClass)
```

This kind of output can be reproduced by any program which includes `type.invalid.annotations.on.use` error.

This PR uses `TypeAnnotationUtils.unannotatedType()` instead of `type.getUnderlyingType()` to beautify the output.

Now, it looks like:

```
checker/tests/nullness/Issue2247.java:16: error: [type.invalid.annotations.on.use] invalid type:
        annotations [@Initialized, @Nullable] conflict with declaration of type Issue2247.DeclaredClass
```

Please give me some comments, thanks!


framework/src/main/java/org/checkerframework/common/basetype/BaseTypeValidator.java
```diff
@@ -185,8 +185,8 @@ protected void reportValidityResultOnUnannotatedType(
             final Tree p) {
         // TODO: if underlying is a compound type such as List<@Palindrome String>, then it would be
         // nice to print it as "List" instead of as "List<@Palindrome String>".
-        TypeMirror underlying = type.getUnderlyingType();
-        checker.report(Result.failure(errorType, type.getAnnotations(), underlying.toString()), p);
+        TypeMirror unannotated = TypeAnnotationUtils.unannotatedType(type.getUnderlyingType());
```
wmdietl: While you're at it, can you try to address the TODO comment above?
Write two extra tests:
1. the top-level type has invalid annotations: `@Bad List<String>` and we only want to see `List`
2. the type argument has invalid annotations: `List<@Bad Map<...>>` and we only want to see `Map`

To get to the erase type, try changing `type.getUnderlyingType()` to `type.getErased().getUnderlyingType()`.

xingweitian: No problem, I will do it.

wmdietl: Friendly ping :-)

xingweitian: Sorry for the delayed response, I have already updated the PR.
I have kept thinking about how to create these two extra test cases, I tried some test cases but failed to show the differences. I am still confused about it... Could you please give me some advises?

xingweitian:
Update:

For `@Bad List<String>`:
```java
public class Issue2247 {

    static @NonNull class DeclaredClass<e extends Object> {}

    // :: error: (type.invalid.annotations.on.use)
    void test(@Nullable DeclaredClass<Integer> object) {}
}
```
Former Output:
```
error: [type.invalid.annotations.on.use] invalid type: 
annotations [@Initialized, @Nullable] conflict with declaration of type 
Issue2247.DeclaredClass<java.lang.Integer>
```
After using `.getErased()`, it looks like:
```
error: [type.invalid.annotations.on.use] invalid type:
annotations [@Initialized, @Nullable] conflict with declaration of type
Issue2247.DeclaredClass
```

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeValidator.java
```diff
@@ -183,10 +183,9 @@ protected void reportValidityResultOnUnannotatedType(
             final @CompilerMessageKey String errorType,
             final AnnotatedTypeMirror type,
             final Tree p) {
-        // TODO: if underlying is a compound type such as List<@Palindrome String>, then it would be
-        // nice to print it as "List" instead of as "List<@Palindrome String>".
-        TypeMirror underlying = type.getUnderlyingType();
-        checker.report(Result.failure(errorType, type.getAnnotations(), underlying.toString()), p);
+        TypeMirror unannotated =
```
wmdietl: The variable name `unannotated` is now a bit unfitting, as it is also the erased type.
How about keeping `underlying` as before? Or `reportType`?

xingweitian: I will keep it as `underlying`.

xingweitian:
Thanks for your advise! I just updated the PR, one thing I need to mention is that I use `&lt;String&gt;` to represent `<String>` in javadoc.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeValidator.java
```diff
@@ -177,15 +177,16 @@ protected void reportValidityResult(
      * Like {@link #reportValidityResult}, but the type is printed in the error message without
      * annotations. This method would print "annotation @NonNull is not permitted on type int",
      * whereas {@link #reportValidityResult} would print "annotation @NonNull is not permitted on
-     * type @NonNull int".
+     * type @NonNull int". In addition, when underlying is a compound type such as @Bad
```
wmdietl: Can you try whether using `{@code @Bad List<String>}` does the right thing?

wmdietl: `underlying` should be `the underlying type`.

PR #2601, opened 2019-07-04
Print classpath when printing classpath-related problems

mernst:


javacutil/src/main/java/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -140,13 +145,32 @@ public static AnnotationMirror fromClass(
...
         return res;
     }
 
+    /**
+     * Format the classpath as a string.
+     *
+     * @return the classpath, as a string
+     */
+    private static String classpathToString() {
+        URLClassLoader sysLoader = (URLClassLoader) ClassLoader.getSystemClassLoader();
```
wmdietl: Does this really return the compilation classpath that javac will use? It sounds more like the classloader that is used for javac itself.
Maybe using `com.sun.tools.javac.file.Locations.userClassPath()` would return the right results? I looked around a bit in that package and that seems the closest to what we want.
But if the output from the SystemClassloader is already useful, I'm fine with that.

PR #2604, opened 2019-07-05
Annotated ParsePosition.java, CharBuffer.java and ThreadGroup.java

Kakarot-SSJ4:


checker/jdk/index/src/java/text/ParsePosition.java
```diff
@@ -0,0 +1,152 @@
...
+public class ParsePosition {
+
+    /**
+     * Input: the place you start parsing.
+     * <br>Output: position where the parse stopped.
+     * This is designed to be used serially,
+     * with each call setting index up for the next one.
+     */
+    @NonNegative int index = 0;
+    int errorIndex = -1;
```
kelloggm: I would think this field (and it's getters/setters) should be `@GTENegativeOne`, also.

wmdietl:
`checker/bin-devel/.plume-scripts` also shouldn't be added to the repository.

Kakarot-SSJ4:
@kelloggm @panacekcz

checker/jdk/index/src/java/nio/CharBuffer.java
```diff
@@ -0,0 +1,1514 @@
...
+     *
+     * @param  capacity
+     *         The new buffer's capacity, in chars
+     *
+     * @return  The new char buffer
+     *
+     * @throws  IllegalArgumentException
+     *          If the <tt>capacity</tt> is a negative integer
+     */
+    public static CharBuffer allocate(int capacity) {
```
kelloggm: `capacity` should be `@NonNegative`. When annotating a file for the annotated JDK, you should always look through the whole file and annotate every method, not just the ones you need.

checker/jdk/index/src/java/nio/CharBuffer.java
```diff
@@ -0,0 +1,1514 @@
...
+     *         The new buffer's limit will be set to <tt>offset + length</tt>.
+     *
+     * @return  The new char buffer
+     *
+     * @throws  IndexOutOfBoundsException
+     *          If the preconditions on the <tt>offset</tt> and <tt>length</tt>
+     *          parameters do not hold
+     */
+    public static CharBuffer wrap(char[] array,
+                                    int offset, int length)
```
kelloggm: These need annotations

Kakarot-SSJ4:
@kelloggm @panacekcz I have annotated CharBuffer.java completely now

checker/jdk/index/src/java/nio/CharBuffer.java
```diff
@@ -0,0 +1,1523 @@
...
+     *
+     * @param target the buffer to read characters into
+     * @return The number of characters added to the buffer, or
+     *         -1 if this source of characters is at its end
+     * @throws IOException if an I/O error occurs
+     * @throws NullPointerException if target is null
+     * @throws ReadOnlyBufferException if target is a read only buffer
+     * @since 1.5
+     */
+    public int read(CharBuffer target) throws IOException {
```
kelloggm: Based on its documentation, I would expect this to be `@GTENegativeOne`

checker/jdk/index/src/java/nio/CharBuffer.java
```diff
@@ -0,0 +1,1523 @@
...
+     * @param  index
+     *         The index from which the char will be read
+     *
+     * @return  The char at the given index
+     *
+     * @throws  IndexOutOfBoundsException
+     *          If <tt>index</tt> is negative
+     *          or not smaller than the buffer's limit
+     */
+    public abstract char get(@NonNegative @LessThan("this.limit") int index);
```
kelloggm: The use of `this.limit` here smells odd to me. I would expect an `@IndexFor` annotation instead. It will be difficult for clients to establish an invariant against an internal property of the buffer. Perhaps we should change this to `@LessThan("this.limit()")`? Unless there is an appropriate `@IndexFor` annotation. What is the relationship between a buffer's length and its limit?

checker/jdk/index/src/java/nio/CharBuffer.java
```diff
@@ -0,0 +1,1523 @@
...
+     *          If there is insufficient space in this buffer
+     *
+     * @throws  IndexOutOfBoundsException
+     *          If the preconditions on the <tt>start</tt> and <tt>end</tt>
+     *          parameters do not hold
+     *
+     * @throws  ReadOnlyBufferException
+     *          If this buffer is read-only
+     */
+    @SuppressWarnings("index:argument.type.incompatible") // #1: i < src.length() - start, where start is @NonNegative, hence i is @IndexFor("src")
```
kelloggm: Your justification only establishes that `i` is `@LTLengthOf("src")`, not that it is `@IndexFor` - there is no proof that it is non-negative. But I would not expect the checker to warn on the lowerbound side here, so maybe it's just the comment that needs to change.

checker/jdk/index/src/java/nio/CharBuffer.java
```diff
@@ -0,0 +1,1523 @@
...
+     *          <tt>position()&nbsp;+&nbsp;index</tt>
+     *
+     * @throws  IndexOutOfBoundsException
+     *          If the preconditions on <tt>index</tt> do not hold
+     */
+    @SuppressWarnings({"index:override.param.invalid", "index:argument.type.incompatible"}) /*
+    index has to be @LTLengthOf("this.remaining()") as written in the documentation.
+    #1: index being @LTLengthOf("this.remaining()") => position() + checkIndex(index, 1) is @LTLengthOf("this.limit")
+    */
+    public final char charAt(@NonNegative @LTLengthOf("this.remaining()") int index) {
```
kelloggm: I would do either `@LTLenghtOf("this")` or `@LessThan("this.remaining()")`, preferring the former. `@LTLengthOf("remaining()")` doesn't make sense, since `remaining` returns an `int`, not an array

checker/jdk/index/src/java/nio/CharBuffer.java
```diff
@@ -0,0 +1,1523 @@
...
+     *         smaller than <tt>start</tt> and no larger than
+     *         <tt>remaining()</tt>
+     *
+     * @return  The new character buffer
+     *
+     * @throws  IndexOutOfBoundsException
+     *          If the preconditions on <tt>start</tt> and <tt>end</tt>
+     *          do not hold
+     */
+    @SuppressWarnings("index:override.param.invalid") // the annotations are as per the documentation
```
kelloggm: I don't think you should override the annotations on the super class. If they're wrong, we should change them.

However, I think these annotations are wrong: `@LTEqLengthOf` doesn't make sense when applied to `"this.remaining()"`, which is an `int`, not an array.

checker/jdk/index/src/java/nio/CharBuffer.java
```diff
@@ -0,0 +1,1523 @@
...
+     *          If <tt>start</tt> or <tt>end</tt> are negative, <tt>start</tt>
+     *          is greater than <tt>end</tt>, or <tt>end</tt> is greater than
+     *          <tt>csq.length()</tt>
+     *
+     * @throws  ReadOnlyBufferException
+     *          If this buffer is read-only
+     *
+     * @since  1.5
+     */
+    @SuppressWarnings({"index:argument.type.incompatible","index:override.param.invalid"}) // #1: cs.length = csq.length as assigned in #0.1, also, the annotations are as per the documentation
```
kelloggm: Again, the super class's annotations should match these. If they don't then one of them should be changed.

checker/jdk/index/src/java/nio/CharBuffer.java
Kakarot-SSJ4: According to the documentation, it should be `@LessThan("this.remaining()")`. As the documentation says `The index of the character to be read, relative to the position must be non-negative and smaller than <tt>remaining()</tt>`.
This also makes sense because then `position() + checkIndex(index,1)` will also be `@LessThan("this.limit()")` as is required by the `get()` documentation.
The problem is with the `override.param.invalid` error, because the `charAt()` function in `CharSequence.java` cannot be annotated in terms of `limit()` as this class does not extend `Buffer.java`

Kakarot-SSJ4:
@kelloggm @panacekcz

checker/jdk/index/src/java/nio/CharBuffer.java
```diff
@@ -0,0 +1,1524 @@
...
+ *
+ *
+ *
+ *
+ *
+ *
+ *
+ */
+
+// -- This file was mechanically generated: Do not edit! -- //
```
kelloggm: We should probably remove this notice, since we are obviously editing it and there's no reason others shouldn't, either.

checker/jdk/index/src/java/nio/CharBuffer.java
```diff
@@ -0,0 +1,1524 @@
...
+     *
+     * <p> The first character of the resulting string will be the character at
+     * this buffer's position, while the last character will be the character
+     * at index <tt>limit()</tt>&nbsp;-&nbsp;1.  Invoking this method does not
+     * change the buffer's position. </p>
+     *
+     * @return  The specified string
+     */
+    @SuppressWarnings("index:return.type.incompatible") // As in documentation, this function returns a string containing the characters in the buffer
+    public @SameLen("this") String toString() {
```
kelloggm: If we're going to annotate this method as `@SameLen("this")`, we should probably also annotate the version above that returns a char array

checker/jdk/index/src/java/nio/CharBuffer.java
smillst: We want to keep the diffs between the annotated JDK and the JDK as small as possible, so this comment should remain.

Kakarot-SSJ4:
@kelloggm @panacekcz

panacekcz:
Test fails at https://github.com/typetools/checker-framework/blob/5000c1ecb72581aeebd3c10a2851cf003eeb554c/checker/tests/index/CharSequenceTest.java#L69

Kakarot-SSJ4:
@kelloggm @panacekcz The build is failing due to the error at  `NonNegativeCharValue.java:5: error: (argument.type.incompatible)`, i.e.,
```
public static String toString(final @NonNegative Character ch) {
        **return toString(ch.charValue());** // error issued here
    }
}
```
which is related to issue #2644 and #2638

kelloggm:
@Kakarot-SSJ4 #2644 has been merged, and #2638 has been marked as fixed. So no errors should still be issued because of them unless you are not using the latest version of the framework.

If the reason is that they were closed recently and are therefore not in the latest release, you can leave the warning suppression for now, with a todo to remove it once the August release of the Checker Framework is released.

Kakarot-SSJ4:
@kelloggm @panacekcz The Travis build has succeeded.

PR #2605, opened 2019-07-05
Add task to print jdk8.jar manifest.

smillst:
Call this task in Travis when use a prebuilt jdk.  Also adds more information to the manifest:

```
Checker-Framework-Version: 242f33a
Checker-Framework-Commit-Author: Noone <noone@cares.com>
Creation-Date: Fri Jul 5 21:07:03 UTC 2019
Created-By: 1.8.0_131 (Oracle Corporation)
Checker-Framework-Commit-Refs: (HEAD)
```

### DIFF
--- a/checker/build.gradle
+++ b/checker/build.gradle
@@ -1,3 +1,5 @@
+import java.util.zip.ZipFile
+
 ext {
     // See instructions for updating jdk8.jar at
     // https://github.com/typetools/annotated-libraries/blob/master/README.md .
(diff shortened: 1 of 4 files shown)


checker/build.gradle
```diff
@@ -135,6 +137,21 @@ task updateJdk() {
         }
     }
 }
+''
```
wmdietl: Out of curiosity: what does the `''` here do?

smillst: Nothing!  I've removed them.  Thanks.

PR #2614, opened 2019-07-10
Treat positive @Signed values as @SignednessGlb in all contexts; fixes #1430; fixes #2534

### LINKED ISSUE #1430: Signedness Checker only uses range refinement on identifiers
The Signedness Checker now uses the Value Checker to refine the types of identifiers based on their value ranges. This extra refinement should be extended to expressions.

### LINKED ISSUE #2534: Signedness Checker inability to treat positive @Signed values as @Constant on non-identifiers
Consider the following code-
```
import org.checkerframework.checker.signedness.qual.Unsigned;

class Issue2534 {

    @IntRange(from = 0, to = Integer.MAX_VALUE) int field = 3;

    @IntRange(from = 0, to = Integer.MAX_VALUE) int qwe() {
        return 3;
    }

    void m1() {
        @Unsigned int c = qwe();
    }

    void m2() {
        @Unsigned int c = field;
    }

    void m3() {
        @Unsigned int c = this.field;
    }

    void m4(@IntRange(from = 0, to = Integer.MAX_VALUE) int array[]){
        @Unsigned int c=array[0];
    }

}
```

Output:
```
ex.java:13: error: [assignment.type.incompatible] incompatible types in assignment.
        @Unsigned int c = qwe();
                             ^
  found   : @Signed int
  required: @Unsigned int
ex.java:21: error: [assignment.type.incompatible] incompatible types in assignment.
        @Unsigned int c = this.field;
                              ^
  found   : @Signed int
  required: @Unsigned int
ex.java:25: error: [assignment.type.incompatible] incompatible types in assignment.
        @Unsigned int c=array[0];
                             ^
  found   : @Signed int
  required: @Unsigned int
3 errors

```
As the returned value of the method is `@Signed` and within the prescribed range so it should be treated as `@Constant` and error should not occur.
Related to #1430.

gokaco:


checker/src/main/java/org/checkerframework/checker/signedness/SignednessAnnotatedTypeFactory.java
```diff
@@ -74,11 +73,69 @@ protected void addComputedTypeAnnotations(
         // but adding the top type to them, which permits flow-sensitive type refinement.
         // (When it is possible to default types based on their TypeKinds,
         // this whole method will no longer be needed.)
+        addSignednessAnnotation(tree, type);
         addUnknownSignednessToSomeLocals(tree, type);
 
         super.addComputedTypeAnnotations(tree, type, iUseFlow);
     }
 
+    private void addSignednessAnnotation(Tree tree, AnnotatedTypeMirror type) {
```
mernst: Every method must have a Javadoc comment.

checker/src/main/java/org/checkerframework/checker/signedness/SignednessAnnotatedTypeFactory.java
```diff
@@ -74,11 +73,69 @@ protected void addComputedTypeAnnotations(
...
+        TypeMirror javaType = type.getUnderlyingType();
+        TypeKind javaTypeKind = javaType.getKind();
+
+        switch (tree.getKind()) {
+            case IDENTIFIER:
```
mernst: Why is this restricted to just two tree kinds?  Document the reason, or expand it.

mernst: The whole point of this pull request was *not* to special-case identifiers but make the code more general.
This pull request does not seem to achieve that.  It just slightly expands the special case.
I think this logic should apply to all expressions.  Please also add tests for other types of expressions (for example, arithmetic operations.)

mernst:
When a pull request fixes an issue, it should include "fixes #NNNN" in the title.
You need to change the title anyway, because you should write a general fix, not just extend the current hack from identifiers to method calls.

mernst:
The continuous integration tests are failing.  Here is the failure:

```
tests.SignednessTest > run[signedness] FAILED
    java.lang.AssertionError: 284 out of 284 expected diagnostics were found.
    1 unexpected diagnostic was found:
    Issue2534.java:21: error: (assignment.type.incompatible)
    While type-checking /__w/1/s/checker/tests/signedness/PolymorphicReturnType.java, /__w/1/s/checker/tests/signedness/ShiftPropogation.java, /__w/1/s/checker/tests/signedness/LowerUpperBound.java, /__w/1/s/checker/tests/signedness/DefaultsSignedness.java, /__w/1/s/checker/tests/signedness/CastedShifts.java, /__w/1/s/checker/tests/signedness/Issue2534.java, /__w/1/s/checker/tests/signedness/Utils.java, /__w/1/s/checker/tests/signedness/Operations.java, /__w/1/s/checker/tests/signedness/MaskedShifts.java, /__w/1/s/checker/tests/signedness/TestPrintln.java, /__w/1/s/checker/tests/signedness/ValueIntegration.java, /__w/1/s/checker/tests/signedness/UtilsJava8.java, /__w/1/s/checker/tests/signedness/Comparisons.java, /__w/1/s/checker/tests/signedness/AnnoBeforeModifier.java, /__w/1/s/checker/tests/signedness/SignednessManualExample.java, /__w/1/s/checker/tests/signedness/CompoundAssignments.java
Testsuite: tests.SignednessTest
Tests run: 6, Failures: 1, Skipped: 0, Time elapsed: 18.702 sec
```

Please either update the test, or fix a bug in your code, so that the tests pass.

checker/tests/signedness/SignednessGlbIssue.java
```diff
@@ -1,7 +1,7 @@
 import org.checkerframework.checker.signedness.qual.Unsigned;
 import org.checkerframework.common.value.qual.IntRange;
 
-class Issue2534 {
+class SignednessGlbIssue {
```
mernst: Why did you rename this?  I don't see what purpose it serves.  Now, a reader of the code doesn't know what issue this test case is related to.  Also, the name `SignednessGlbIssue` is so vague that it does not convey any intuition.  (The name `Issue2534` also does not convey any intuition, but at least it links to an issue.)

checker/tests/signedness/SignednessGlbIssue.java
```diff
@@ -17,7 +17,7 @@ void m2() {
         @Unsigned int c = field;
     }
 
-    void m3() {
-        @Unsigned int c = this.field;
+    void m3(@IntRange(from = 0, to = Integer.MAX_VALUE) int array[]) {
```
mernst: Why did you replace this rather than adding a test case?  Is the old test wrong?  If it's not wrong, it is usually better to augment rather than delete it.

checker/tests/signedness/SignednessGlbIssue.java
```diff
@@ -17,7 +17,7 @@ void m2() {
...
+        @Unsigned int c=array[0];
```
mernst: The formatting is incorrect.

checker/src/main/java/org/checkerframework/checker/signedness/SignednessAnnotatedTypeFactory.java
```diff
@@ -79,6 +79,10 @@ protected void addComputedTypeAnnotations(
         super.addComputedTypeAnnotations(tree, type, iUseFlow);
     }
 
+    /**
+     * Refines the type of an integer primitive to @SignednessGlb if it is within the signed
+     * positive range (i.e. its MSB is zero).
+     */
```
mernst: This comment doesn't describe the arguments.
Also, it suggests that the name of the method could be improved; even `addSignednessGlbAnnotation` would be better.

checker/tests/signedness/SignednessGlbIssue.java
gokaco: Yes

mernst: You don't need to respond "yes".  Just fix the problem.  Thanks.

checker/src/main/java/org/checkerframework/checker/signedness/SignednessAnnotatedTypeFactory.java
gokaco: @mernst I have updated the comment. Please review it.

checker/src/main/java/org/checkerframework/checker/signedness/SignednessAnnotatedTypeFactory.java
```diff
@@ -74,11 +73,73 @@ protected void addComputedTypeAnnotations(
...
         // (When it is possible to default types based on their TypeKinds,
         // this whole method will no longer be needed.)
+        addSignednessGlbAnnotation(tree, type);
         addUnknownSignednessToSomeLocals(tree, type);
 
         super.addComputedTypeAnnotations(tree, type, iUseFlow);
     }
 
+    /**
+     * Refines the type of an expression such as array, method invocation or an integer primitive
```
mernst: If it's any expression, there is no need to give examples.  It's any expression of primitive type, I think.

checker/src/main/java/org/checkerframework/checker/signedness/SignednessAnnotatedTypeFactory.java
```diff
@@ -74,11 +73,73 @@ protected void addComputedTypeAnnotations(
...
+     * to @SignednessGlb if it is within the signed positive range. (i.e. its MSB is zero).
```
mernst: The grammar is wrong here.  I think you can delete the first period on the line to fix it.

mernst: You didn't do this.  I did so, but please address all comments (either by making a fix or by starting a conversation) rather than ignoring some.

checker/src/main/java/org/checkerframework/checker/signedness/SignednessAnnotatedTypeFactory.java
mernst: You didn't address my feedback, which was that there is no `@param` in the Javadoc.

PR #2615, opened 2019-07-10
@CFComment annotation

the1derer:


framework/src/main/java/org/checkerframework/framework/qual/CFComment.java
```diff
@@ -0,0 +1,41 @@
...
+
+import static java.lang.annotation.ElementType.*;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This annotation is to be used to add comments related to Checker-Framework. This annotation can
```
mernst: The tool is called "the Checker Framework", not "Checker-Framework".

framework/src/main/java/org/checkerframework/framework/qual/CFComment.java
```diff
@@ -0,0 +1,41 @@
...
+ * }
+ * </code></pre>
+ *
+ * <p>As a matter of style, programmers should always use this annotation on the most deeply nested
+ * element where the comment related to Checker-Framework is to be provided.
+ *
+ * <p>These annotations are {@link Documented} and hence will appear in Javadocs.
+ */
+@Documented
+@Target({TYPE, FIELD, METHOD, PARAMETER, CONSTRUCTOR, LOCAL_VARIABLE, PACKAGE})
```
mernst: You can omit this, because omitting it is the same as making it a declaration annotation.
This should not be a type annotation.

framework/src/main/java/org/checkerframework/framework/qual/CFComment.java
```diff
@@ -0,0 +1,41 @@
...
+ *     public String test{
+ *         return "test";
+ *     }
+ * }
+ * </code></pre>
+ *
+ * <p>As a matter of style, programmers should always use this annotation on the most deeply nested
+ * element where the comment related to Checker-Framework is to be provided.
+ *
+ * <p>These annotations are {@link Documented} and hence will appear in Javadocs.
```
mernst: This is also obvious from the meta-annotations.

framework/src/main/java/org/checkerframework/framework/qual/CFComment.java
```diff
@@ -0,0 +1,41 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This annotation is to be used to add comments related to Checker-Framework. This annotation can
+ * be used on packages, classes, interfaces and any elements inside these.
+ *
+ * <p>Each comment should first start with the type-system to which comment is related to followed
```
mernst: Clarify that this is a style guideline or recommendation.

framework/src/main/java/org/checkerframework/framework/qual/CFComment.java
```diff
@@ -0,0 +1,41 @@
...
+ * by ':' and then actual comment. Provide single string if comment is for only one type-system and
```
mernst: Users should be permitted to make their own decisions about this.  In particular, it is sensible for a user to break a long comment across lines, by putting it in multiple strings.

framework/src/main/java/org/checkerframework/framework/qual/CFComment.java
```diff
@@ -0,0 +1,41 @@
...
+import static java.lang.annotation.ElementType.*;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This annotation is to be used to add comments related to Checker-Framework. This annotation can
+ * be used on packages, classes, interfaces and any elements inside these.
```
mernst: You can just say this is a declaration annotation, or leave it out since it is obvious from the

framework/src/main/java/org/checkerframework/framework/qual/CFComment.java
```diff
@@ -0,0 +1,41 @@
...
+ * <p>As a matter of style, programmers should always use this annotation on the most deeply nested
+ * element where the comment related to Checker-Framework is to be provided.
+ *
+ * <p>These annotations are {@link Documented} and hence will appear in Javadocs.
+ */
+@Documented
+@Target({TYPE, FIELD, METHOD, PARAMETER, CONSTRUCTOR, LOCAL_VARIABLE, PACKAGE})
+@Retention(RetentionPolicy.SOURCE)
+public @interface CFComment {
+    /** @return array of {@link String} containing comments */
```
mernst: `@return` is not valid here.
It's obvious from the type that this is an array of strings.  Don't repeat the type in a comment.
Give information about the content.

framework/src/main/java/org/checkerframework/framework/qual/CFComment.java
wmdietl: I agree with omitting this.
`TYPE` is for declarations of types and `CFComment` should be allowed there.
Omitting `@Target` will allow all declaration locations, including `TYPE`, which is good.

mernst:
I agree that the documentation should be in the same pull request.

the1derer:
> I agree that the documentation should be in the same pull request.

@mernst Should I add documentation about `@CFComment` in [annotating-libraries](https://github.com/typetools/checker-framework/blob/master/docs/manual/annotating-libraries.tex) section or any other place?

mernst:
That is a fine place for it.  The documentation doesn't have to be extensive.

mernst:
I moved some documentation to more appropriate places and merged.  Thanks for the annotation!

Issue #2432, opened 2019-04-19
@Poly* in type argument position


wmdietlGC:
Take the following Nullness Checker example that uses `@PolyNull` in a type argument:

```java
import org.checkerframework.checker.nullness.qual.Nullable;
import org.checkerframework.checker.nullness.qual.PolyNull;
import java.util.List;

class PolyNullTest {
  void foo(List<@PolyNull Object> l) {
    l.add(null);
  }
  void test(List<@Nullable Object> n, List<Object> nn) {
      foo(n);
      foo(nn);
  }
}
```

This correctly allows both calls of `foo`, but does not raise an error for `l.add(null)`.
As this could be executed on the non-null list, adding `null` must be forbidden as it could lead to an NPE.

If the signature of `foo` is changed to `List<? extends @Nullable Object>` the addition of `null` is correctly prevented.
Using `@PolyNull` for this is useful when using a wildcard isn't meaningful, e.g. for a bound that is a final class.

smillst:
This is a general problem.  The follow code should issue a warning, but doesn't:
```
import java.util.List;
import org.checkerframework.checker.tainting.qual.PolyTainted;
import org.checkerframework.checker.tainting.qual.Tainted;

class PolyTaintedTest {
    void polyTainted(List<@PolyTainted Object> list, @Tainted String s) {
        list.add(s);
    }
}
```

PR #2616, opened 2019-07-10
Fix on Issue 2432: exclude type variable when replacing poly annotations

abslamp:
Test case and one possible fix on #2432 
Excluding type variables with concrete type arguments ("E" TypeVar) in lub calculation and poly annotation replacement.
Type variables with no assigned arguments ("T" TypeVar) are not affected. (e.g. when visiting void \<T\> fun(T) , T is not excluded)


mernst:
Please use a descriptive title, not just "Issue 2432".  You should also add "fixes #NNNN" which is useful, but does not describe the work you have done.  Thanks!

abslamp:
Logic proven wrong. A new fix: https://github.com/typetools/checker-framework/pull/2643

wmdietl:
Did you push something to this branch in error? Wasn't this replaced by the changes in #2643?

abslamp:
@wmdietl I plan to test some logic on this clean branch to trigger check. After that, I will close this PR again.

PR #2622, opened 2019-07-11
Replaced constants with their value in DependentTypesHelper

### LINKED ISSUE #2613: Checker cannot infer 0 < Integer.MAX_VALUE
Consider the following code:

```
class Issue2613{
    void demo() {
        require_lt(0, Integer.MAX_VALUE);
    }

    void require_lt(@LessThan("#2") int a, int b){ }
}
```

Run it by the following command:
```
javac -processor index Issue2613.java
```

Output:
```
Issue2613.java:4: error: [argument.type.incompatible] incompatible types in argument.
	require_lt(0, Integer.MAX_VALUE);
	       ^
  found   : @LessThanUnknown int
  required: @LessThan("java.lang.Integer.MAX_VALUE") int
1 error
```

However, 0 < Integer.MAX_VALUE, hence this error should not be issued.

Maxi17:
Whenever a constant field is passed as argument, it is replaced by its value when the `String` is standardized.

Fixes #2613


framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -480,6 +480,14 @@ protected String standardizeString(
             if (result == null) {
                 return new DependentTypesError(expression, " ").toString();
             }
+            if (result instanceof FlowExpressions.FieldAccess
+                    && ((FlowExpressions.FieldAccess) result).isFinal()) {
+                Object constant =
+                        ((FlowExpressions.FieldAccess) result).getField().getConstantValue();
```
smillst: Doubles and floats aren't parseable, so if the constant value is a double or float, the name of the field should be used instead.

Maxi17:
I added the test case suggested by @panacekcz. Without adding the check `!(constant instanceof String)`, the test case fails with:

```
Issue2613.java:13: error: [argument.type.incompatible] incompatible types in argument.
        require_lt(0, STRING_CONSTANT);
                   ^
  required: @LTLengthOf(value="Hello") int
1 error
```
I think ignoring `String` constants is fine because it is not important to pass them by value.

About moving the code to `OffsetDependentTypesHelper`, doing so the test case added by @smillst fails because `FlowExp` is only using `DependentTypesHelper`, not `OffsetDependentTypesHelper`. Therefore, the first assignment in the test case fails for the same reason as #2613 (constants passed by name, not value). What do you suggest here?

smillst:
I changed the `FlowExp` tests to use the LessThan Checker. So you should now be able to move this change to the Offset helper and expect the tests to pass.  Thanks!

PR #2625, opened 2019-07-12
Make annotations repeatable

gokaco:
The following annotations have been made repeatable:-
```
@EnsuresNonNull
@EnsuresNonNullIf
@EnsuresKeyFor
@EnsuresKeyForIf
@EnsuresLTLengthOf
@EnsuresLTLengthOfIf
@EnsuresLockHeld
@EnsuresLockHeldIf
@EnsuresMinLenIf
```
Related to Issue #1307


gokaco:
@mernst Please review the changes I have made and do let me know how to put the comments in the files.
Thanks.

wmdietl:
Regarding https://github.com/typetools/checker-framework/pull/2369#discussion_r303263630 see
https://docs.oracle.com/javase/8/docs/api/javax/lang/model/AnnotatedConstruct.html for a discussion "present" vs. "directly present".
Methods like https://docs.oracle.com/javase/8/docs/api/javax/lang/model/AnnotatedConstruct.html#getAnnotationsByType-java.lang.Class- look through container annotations.

Please try using `getAnnotationsByType` to see whether the basic use-case works.
We might need to look into how the StubParser handles repeatable annotations and I suspect some work will be needed there.

gokaco:
> Regarding [#2369 (comment)](https://github.com/typetools/checker-framework/pull/2369#discussion_r303263630) see
> https://docs.oracle.com/javase/8/docs/api/javax/lang/model/AnnotatedConstruct.html for a discussion "present" vs. "directly present".
> Methods like https://docs.oracle.com/javase/8/docs/api/javax/lang/model/AnnotatedConstruct.html#getAnnotationsByType-java.lang.Class- look through container annotations.
> 
> Please try using `getAnnotationsByType` to see whether the basic use-case works.
> We might need to look into how the StubParser handles repeatable annotations and I suspect some work will be needed there.

@wmdietl I tried using `getAnnotationsByType` and found that we will have to write different lines of code for different annotations, i.e will have to use different class object corresponding to the annotation type like-
```
EnsuresNonNull[] ensure = methodElement.getAnnotationsByType(EnsuresNonNull.class);
EnsuresKeyFor[] ensure2 = methodElement.getAnnotationsByType(EnsuresKeyFor.class);
EnsuresLTLengthOf[] ensure3 = methodElement.getAnnotationsByType(EnsuresLTLengthOf[].class);
and so on for more annotations.
```
The above line will work specifically for their corresponding annotation. Is their any method which can extract the class of the annotation getting used at compilation time which should work as a parameter for `getAnnotationsByType` and its return type should get stored in the array of that wrapper class?

wmdietl:
I'm not quite sure what you're asking.

You will probably have to look at `factory.getDeclAnnotationWithMetaAnnotation(methodElement, metaAnnotation);` and `getDeclAnnotation` and decide how they should handle repeatable annotations.
We shouldn't call anything on the Element directly, because that would miss annotations from stub files.

Also, the String comparison against `"List"` can easily break, e.g. if a post-condition happens to contain that String. So it is important to figure out how to use the proper compiler APIs to have this figured out by the compiler correctly.

gokaco:
@mernst @wmdietl Ready for review. I have not added test cases for `@EnsuresLockHeld` and `@EnsuresLockHeldIf` as I do not know much about that checker. Please add that!
Thanks!

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOf.java
```diff
@@ -78,4 +80,20 @@
     @JavaExpression
     @QualifierArgument("offset")
     String[] offset() default {};
+
+    /** An inner annotation which makes {@link EnsuresLTLengthOf} annotation repeatable */
```
mernst: Small comments:
1. End each sentence with punctuation, such as a period.
2. Use "that" instead of "which" in this context.  Here is an explanation from _Strunk and White_:
> That.  Which.  _That_ is the defining, or restrictive pronoun, _which_ the
> nondefining, or nonrestrictive.
>
> 	The lawn mower that is broken is in the garage.  (Tells which one.)
> 	The lawn mover, which is broken, is in the garage.  (Adds a fact
> 		about the only mower in question.)
>
> The use of _which_ for _that_ is common in written and spoken language
> ("Let us now go even unto Bethlehem, and see this thing which is come to
> pass.")  Occasionally _which_ seems preferable to _that_, as in the
> sentence from the Bible.  But it would be a convenience to all if these two
> pronouns were used with precision.  The careful writer, watchful for small
> conveniences, goes _which_-hunting, removes the defining _whiches_, and by
> so doing improves his work.
>
> The Elements of Style
> William Strunk, Jr. and E. B. White

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOf.java
```diff
@@ -78,4 +80,20 @@
...
+    @Documented
+    @Retention(RetentionPolicy.RUNTIME)
+    @Target({ElementType.METHOD, ElementType.CONSTRUCTOR})
+    @PostconditionAnnotation(qualifier = LTLengthOf.class)
```
mernst: Does this work?

The documentation of `@ConditionalPostconditionAnnotation` says:

> The annotation E that is meta-annotated as ConditionalPostconditionAnnotation must
> have an element called `expression` that is an array of Strings of the same format
> and with the same meaning as the value `expression` in `EnsuresQualifierIf`.

and that is not true of this annotation.  So either the definition of this annotation, or the documentation of `@ConditionalPostconditionAnnotation`, must be changed.

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOf.java
```diff
@@ -78,4 +80,20 @@
...
+    @InheritedAnnotation
+    @interface List {
+        /**
+         * Programmers generally do not need to use this; it is created by Java when a programmer
+         * writes more than one {@link EnsuresLTLengthOf} annotation at the same location.
+         *
+         * @checker_framework.manual #java-expressions-as-arguments Syntax of Java expressions
```
mernst: What is the point of this?  The annotation does not use Java expressions.

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOf.java
```diff
@@ -78,4 +80,20 @@
...
+
+    /** An inner annotation which makes {@link EnsuresLTLengthOf} annotation repeatable */
+    @Documented
+    @Retention(RetentionPolicy.RUNTIME)
+    @Target({ElementType.METHOD, ElementType.CONSTRUCTOR})
+    @PostconditionAnnotation(qualifier = LTLengthOf.class)
+    @InheritedAnnotation
+    @interface List {
+        /**
+         * Programmers generally do not need to use this; it is created by Java when a programmer
```
mernst: This comment is out of place.  It refers to the `List` interface, not to the `value` field where you have placed it.

checker/tests/index/RepeatIndexAnno.java
```diff
@@ -0,0 +1,31 @@
...
+        v3 = value3.length() - 4;
+    }
+
+    @EnsuresLTLengthOfIf(expression = "v1", targetValue = "value1", offset = "1", result = false)
+    @EnsuresLTLengthOfIf(expression = "v2", targetValue = "value2", offset = "2", result = false)
+    @EnsuresLTLengthOfIf(expression = "v3", targetValue = "value3", offset = "3", result = true)
+    public boolean func2() {
+        v1 = value1.length() - 2;
+        v2 = value2.length() - 3;
+        v3 = value3.length() + 4;
```
mernst: Why does this line use `+` instead of `-` like the previous two lines do?  What is it testing?

checker/tests/index/RepeatIndexAnno.java
```diff
@@ -0,0 +1,31 @@
+import org.checkerframework.checker.index.qual.EnsuresLTLengthOf;
+import org.checkerframework.checker.index.qual.EnsuresLTLengthOfIf;
+
+class RepeatIndexAnno {
```
mernst: It is best if every test case:
 * includes positive and negative examples -- that is, lines that type-check and lines that cause a type-checking warning.
 * includes a version of the code with no annotations, to show that adding the annotations changes the behavior
 * When there are multiple ways to achieve an effect, includes each of those.

To handle the latter two points, this test case should include three versions of each method:  one with repeated annotations, one with the `List` wrapper, and one with no annotations.  Beyond that, however, it should also have a version that contains both `List` (with two annotations within) and one annotation, to show that its effect is the same as either 3 regular annotations or 1 `List` annotation with 3 annotations within.

Another thing that is missing from this test case is clients of the given methods func1 and func2; you need to show that they behave correctly.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -418,19 +430,41 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
-            List<String> expressions =
-                    AnnotationUtils.getElementValueArray(anno, "value", String.class, false);
-            AnnotationMirror postcondAnno = getAnnotationMirrorOfMetaAnnotation(metaAnno, anno);
-            if (postcondAnno == null) {
-                continue;
-            }
-            for (String expr : expressions) {
-                result.add(new Postcondition(expr, postcondAnno, anno));
+            Annotation p = methodElement.getAnnotation(EnsuresNonNull.List.class);
+            if (p == null) p = methodElement.getAnnotation(EnsuresKeyFor.List.class);
```
mernst: Hard-coded logic is a code smell.  You should avoid that, because it is a bad practice.

As implemented, it is impossible for a type system designer to create their own wrapper annotations, without modifying the code of the Checker Framework and re-compiling it!  We do not want that.  I would expect that you can test for the existence of a meta-annotation, such as whatever you will need to define to replace the uses of `@PostconditionAnnotation`.

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOf.java
gokaco: It is annotated with `@PostconditionAnnotation`, not with `@ConditionalPostconditionAnnotation`.

mernst: Thanks for the correction.  The comment still applies.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
mernst: You could also keep the test you had before (about the type of the value field).  The criticism of your previous approach was primarily about the fact that it used exceptions as control flow (and without any documentation of what it was doing).

gokaco:
@mernst I have updated the test cases as you told me to do and also updated the logic with proper documentation. It is ready for review.

For the issue related to `@PostconditionAnnotation` how should I proceed? Is the documentation incorrect of this annotation?
Thanks.

checker/tests/index/RepeatLTLengthOf.java
```diff
@@ -0,0 +1,73 @@
...
+    protected int v3;
+
+    public void func1() {
+        v1 = value1.length() - 1;
+        v2 = value2.length() - 3;
+        v3 = value3.length() - 4;
+    }
+
+    public boolean func2() {
+        v1 = value1.length() - 3;
```
mernst: Why are these out of order?  That adds extra cognitive load to anyone reading the code.

checker/tests/index/RepeatLTLengthOf.java
```diff
@@ -0,0 +1,73 @@
...
+        return true;
+    }
+
+    @EnsuresLTLengthOf.List({
+        @EnsuresLTLengthOf(value = "v1", targetValue = "value1", offset = "1"),
+        @EnsuresLTLengthOf(value = "v2", targetValue = "value2", offset = "2")
+    })
+    @EnsuresLTLengthOf(value = "v3", targetValue = "value3", offset = "3")
+    public void func3() {
+        v1 = value1.length() - 2;
```
mernst: These values are inconsistent with every other routine.  This makes it hard to understand the relationship between the test cases, and forces a programmer to individually understand them.  I asked you to create the *same* test case, just with different annotations.  Then a programmer can very quickly determine that the behavior is the same in call cases.

checker/tests/nullness/RepeatKeyFor.java
```diff
@@ -0,0 +1,77 @@
...
+        map2.put(b, 3);
+        return true;
+    }
+
+    @EnsuresKeyFor(
+            value = {"#1", "#2"},
+            map = "map")
+    @EnsuresKeyFor(value = "#2", map = "map2")
+    // :: error:  (contracts.postcondition.not.satisfied)
+    public void samefunc1(String a, String b, String c) {
```
mernst: "same" is not very illuminating as a name.  Even "with postconditions" would be clearer.

gokaco:
> You have not written clients of the annotated functions, to test that the postcondition annotations have an effect.

@mernst What is the meaning of clients here?

checker/tests/index/RepeatLTLengthOf.java
gokaco: @mernst What do you mean by out of order?

mernst:
> > You have not written clients of the annotated functions, to test that the postcondition annotations have an effect.
> 
> @mernst What is the meaning of clients here?

A client of a method m1 is a method that calls m1.

checker/tests/index/RepeatLTLengthOf.java
mernst: The integer constants (the subtrahends) are out of order.

gokaco: Fixed!

checker/tests/index/RepeatLTLengthOf.java
gokaco: Fixed!

checker/tests/nullness/RepeatKeyFor.java
gokaco: Fixed!

gokaco:
>you have not written clients of the annotated functions, to test that the postcondition annotations have an effect.

@mernst Does the errors which are occurring are not enough to show that postcondition annotations are working? Would it not be better if I write comments about how they are working. I have made a commit related to it. Please review it and do let me know if it fulfills our mission.
Thanks.

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOf.java
```diff
@@ -78,4 +80,20 @@
     @JavaExpression
     @QualifierArgument("offset")
     String[] offset() default {};
+
+    /**
+     * An inner annotation that makes {@link EnsuresLTLengthOf} annotation repeatable.
```
mernst: You don't need to say it's an *inner* annotation -- that is obvious from the declaration and is not the most important thing (which is implied by making it the second word).  Documentation should focus on what an abstraction means, or sometimes how to use it, rather than duplicating information that is already available.

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOf.java
```diff
@@ -78,4 +80,20 @@
...
+     *
+     * <p>Programmers generally do not need to use this; it is created by Java when a programmer
```
mernst: I would use "write" rather than "use", because it is more specific.  It is also parallel with the later part of the sentence.

As a minor comment, I would also change the semicolon to a period (a sentence break).  The semicolon is perfectly grammatical, but shorter sentences tend to be easier to read and understand.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -418,19 +420,42 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
-            List<String> expressions =
-                    AnnotationUtils.getElementValueArray(anno, "value", String.class, false);
-            AnnotationMirror postcondAnno = getAnnotationMirrorOfMetaAnnotation(metaAnno, anno);
-            if (postcondAnno == null) {
-                continue;
-            }
-            for (String expr : expressions) {
-                result.add(new Postcondition(expr, postcondAnno, anno));
+            /**
+             * If the {@code value} field is a list of Strings, then the try block will work
```
mernst: In a previous code review, I said not to use exceptions as control flow.  You could write an explicit check of the `value` field, but don't make the code throw an exception when an `if` statement is possible.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -473,7 +498,6 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
                 result.addAll(getConditionalPostcondition(a));
             }
         }
-
```
mernst: Why did you delete this line?

checker/tests/index/RepeatLTLengthOf.java
```diff
@@ -0,0 +1,77 @@
...
+    public boolean func2() {
+        v1 = value1.length() - 3;
+        v2 = value2.length() - 3;
+        v3 = value3.length() - 3;
+        return true;
+    }
+
+    // Error occurred because minimum length of "value1" should be greater than "v1+4" as described
+    // in the postcondition.
+    @EnsuresLTLengthOf(value = "v1", targetValue = "value1", offset = "4")
```
mernst: You still have not done as I asked.
The purpose of this test case is to determine whether the behavior of the original and wrapped annotations are the same.
To make the simplest and easiest to understand test of this, you should use the same variables and offsets consistently throughout.
This test case is extremely inconsistent and confusing.
The offsets are:
For withpostconditionsfunc1: 4,2,1
For withcondpostconditionsfunc2: 2,3,1
For withpostconditionfunc1: 0,2,1
For withcondpostconditionfunc2: 2,1,0
What is the rationale for this?  It provides no value, and it means that a programmer cannon easily confirm that the test case is serving its purpose.
I have already pointed this out as a problem several times.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
gokaco: @mernst I tried but was not able to. How to check that an `AnnotationMirror anno` consists of String class or AnnotationMirror class?

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
gokaco: I run `./gradlew reformat`after every change. Maybe that caused this change.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
mernst: Can you tell me what you tried?  Hint: there are only two methods in the [`AnnotationMirror` class](https://docs.oracle.com/javase/8/docs/api/javax/lang/model/element/AnnotationMirror.html).

In general, don't just say "I tried but was not able to".  That is indistinguishable from you doing nothing at all, and it does not enable me to help you.
Rather than asking me to do all the work, it is better for you to say the specific part of it you are having trouble with.

gokaco: @mernst I finally worked out a way on how to remove try-catch. I have made a commit related to it, please do review it. I will push the commit related to the test cases in a nick of time.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -413,24 +416,64 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
-                    AnnotationUtils.getElementValueArray(anno, "value", String.class, false);
-            AnnotationMirror postcondAnno = getAnnotationMirrorOfMetaAnnotation(metaAnno, anno);
-            if (postcondAnno == null) {
-                continue;
+
+            @SuppressWarnings("unchecked")
+            List<AnnotationValue> la =
+                    AnnotationUtils.getElementValue(anno, "value", List.class, false);
+
+            {
```
mernst: This is better in that it does not use exceptions for control flow.  Thanks.

Can you please add some documentation?
For example, I do not understand the distinction between the variables `expressions` and `expressionss`.

Also, please abstract the logic into a new method.
Your new code replaces one or two method calls, and the code will remain clearer if you just change it to call a new method.  That new method may have a complex implementation (40 lines), but it can have a simple and clear specification.

The same comments apply to the code below where you replaced a couple of method calls by a large chunk of code.

gokaco: > When a method has a postcondition specification, the Checker Framework should:
> * establish the postconditions at each call site.  Your tests do not check this.

@mernst I tested the second point and it came out that it is not working on-call site. I am not getting from where this problem is occurring. Please give feedback on it! `@wmdietl` once said that 
>We might need to look into how the StubParser handles repeatable annotations and I suspect some work will be needed there.

Is the problem because of this?

Sample test-
```
import java.util.HashMap;
import java.util.Map;
import org.checkerframework.checker.nullness.qual.EnsuresKeyFor;
import org.checkerframework.checker.nullness.qual.EnsuresKeyForIf;

public class ex {

    Map<String, Integer> map = new HashMap<>();
    Map<String, Integer> map2 = new HashMap<>();

    @EnsuresKeyFor(value = "#1", map = "map")
    @EnsuresKeyFor(value = "#1", map = "map2")
    public void client1(String a, String b, String c) {
        withpostconditionsfunc1(a,b,c);
    }

    @EnsuresKeyFor(value = "#1", map = "map")
    @EnsuresKeyFor(value = "#1", map = "map2")
    public void withpostconditionsfunc1(String a, String b, String c) {
        map.put(a, 1);
        map.put(b, 2);
        map2.put(a, 3);
    }

}
```
Output:-
```
ex.java:13: error: [contracts.postcondition.not.satisfied] the postcondition EnsuresKeyFor about 'a' of this method is not satisfied
    public void client1(String a, String b, String c) {
                ^
1 error
```

checker/tests/index/RepeatLTLengthOf.java
gokaco: > When a method has a postcondition specification, the Checker Framework should:
> * establish the postconditions at each call site.  Your tests do not check this.

@mernst Please review the new test cases!
The `@EnsuresKeyFor` and `@EnsuresKeyForIf` are not working on-call sites when they are repeatable. I am not getting from where the error is occurring. Please give a feedback on it! `@wmdietl` once said that-
>We might need to look into how the StubParser handles repeatable annotations and I suspect some work will be needed there.

Is the problem because of this?

Sample test-
```
import java.util.HashMap;
import java.util.Map;
import org.checkerframework.checker.nullness.qual.EnsuresKeyFor;
import org.checkerframework.checker.nullness.qual.EnsuresKeyForIf;

public class ex {

    Map<String, Integer> map = new HashMap<>();
    Map<String, Integer> map2 = new HashMap<>();

    @EnsuresKeyFor(
            value = {"#1", "#2"},
            map = "map")
    @EnsuresKeyFor(value = "#2", map = "map2")
    public void client1(String a, String b, String c) {
        withpostconditionsfunc1(a,b,c);
    }

    @EnsuresKeyForIf(
            expression = {"#1", "#2"},
            map = "map",
            result = true)
    @EnsuresKeyForIf(expression = "#2", map = "map2", result = true)
    public boolean client3(String a, String b, String c) {
        return withcondpostconditionsfunc2(a,b,c);
    }    

    @EnsuresKeyFor(
            value = {"#1", "#2"},
            map = "map")
    @EnsuresKeyFor(value = "#2", map = "map2")
    public void withpostconditionsfunc1(String a, String b, String c) {
        map.put(a, 1);
        map.put(b, 2);
        map2.put(b, 3);
    }


    @EnsuresKeyForIf(
            expression = {"#1", "#2"},
            map = "map",
            result = true)
    @EnsuresKeyForIf(expression = "#2", map = "map2", result = true)
    public boolean withcondpostconditionsfunc2(String a, String b, String c) {
        map.put(a, 1);
        map.put(b, 2);
        map2.put(b, 3);
        return true;
    }
}
```
Output-
```
ex.java:15: error: [contracts.postcondition.not.satisfied] the postcondition EnsuresKeyFor about 'b' of this method is not satisfied
    public void client1(String a, String b, String c) {
                ^
ex.java:25: error: [contracts.conditional.postcondition.not.satisfied] the conditional postcondition about 'EnsuresKeyForIf' at this return statement is not satisfied
        return withcondpostconditionsfunc2(a,b,c);
        ^
2 errors
```

(The message got deleted earlier by mistake)

mernst: If your test case does not use stub files, then Werner's comment about stub files is not relevant to it.

What have you tried so far?

gokaco: @mernst I broke the logic in the functions and now it is working for `@EnsuresKeyFor and `@EnsuresKeyForIf. I have added the test case.

gokaco:
@mernst The pull request is ready. I have written the documentation but I think it can be improved. You can help me with that. Rest I think is now ready.
Thanks!

gokaco:
@mernst Azure pipeline is failing while the Travis is passing. I am not getting what's the issue. Please check!

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
mernst: That is unlikely.  The formatting routines do not remove whitespace between two lines of code.

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOf.java
mernst: You have not addressed this comment.  Please do not assign a pull request to me until you have completely addressed all the feedback.  It is tedious for me to have to go over all the feedback to see whether you have addressed it.

mernst:
> @mernst Azure pipeline is failing while the Travis is passing. I am not getting what's the issue. Please check!

It looks like a connectivity or caching issue on Azure Pipelines.  I tried re-running the job, but that led to more problems.  As long as Travis is passing, we can consider Continuous Integration to be successful.

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOf.java
gokaco: I have removed the statement.

mernst: I asked you to remove *one word*:  the word "inner".  You instead removed the most important part of the documentation, so that now the Javadoc says *nothing* about what the abstraction means, which I noted is the most important part of the documentation.

gokaco: I have fixed this!

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOf.java
mernst: You still have not addressed this comment, one week later (and after you have multiple times told me that the pull request is ready to review).  Please address all comments rather than making me find the ones you ignored.

gokaco: @mernst The documentation of `@PostconditionAnnotation` says-
>The annotation E that is meta-annotated as PostconditionAnnotation must have an element called  `value` that is an array of Strings of the same format and with the same meaning as the value `expression` in `EnsuresQualifier`

The annotation `@EnsuresLTLengthOf` has an array `value` of Strings which is true for this annotation. Is something else wrong?

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOf.java
gokaco: Fixed that!

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOf.java
gokaco: Fixed this!

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOf.java
gokaco: Fixed this! Now it is correctly placed for `List` interface.

checker/tests/index/RepeatIndexAnno.java
gokaco: I have changed the test case. Now it should be meaningful.

checker/tests/index/RepeatIndexAnno.java
gokaco: I have made 2 test file. One of them show why the errors are occurring and the other test case passes the test.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
gokaco: I have fixed this

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOf.java
gokaco: Changed use to write.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
gokaco: I have added the documentation. There may be grammatical mistakes. You can help me with that.

gokaco:
@mernst Ready for review

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOf.java
mernst: This comment thread is associated with the annotation `EnsuresLTLengthOf.List`.
`EnsuresLTLengthOf.List` is meta-annotated with `@PostconditionAnnotation` (the line of code this comment thread is associated with).  That is in violation of the documentation of `@PostconditionAnnotation`.

This is what I said in the very first comment in this comment thread, so I feel that we are going in circles.  Do you see that `@PostconditionAnnotation`'s documentation says it should only be applied to annotations whose `value` field is a array of strings, but here it is applied to an annotation (namely, `EnsuresLTLengthOf.List`) whose `value` field is not an array of strings?

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -413,13 +414,31 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+        Class<AnnotationMirror> annotationmirrorType = AnnotationMirror.class;
         List<Pair<AnnotationMirror, AnnotationMirror>> declAnnotations =
                 factory.getDeclAnnotationWithMetaAnnotation(methodElement, metaAnnotation);
         for (Pair<AnnotationMirror, AnnotationMirror> r : declAnnotations) {
             AnnotationMirror anno = r.first;
             AnnotationMirror metaAnno = r.second;
-            List<String> expressions =
-                    AnnotationUtils.getElementValueArray(anno, "value", String.class, false);
+
+            @SuppressWarnings("unchecked")
```
mernst: I asked you to abstract lines 425-441 into a method.  Then, the change to this method (`getPostconditions`) is just one or two lines.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -428,6 +447,45 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
             }
         }
+
+        return result;
+    }
+
+    /**
+     * Returns the set of repeatable postconditions.
+     *
+     * @param metaAnno set of post condition meta annotations
```
mernst: This comment is wrong.  The type of `metaAnno` is `AnnotationMirror`, so it is not a set.  Please write accurate comments.  Do you mean that it is a wrapper for repeatable annotations?

Also, "postcondition" is one word (though you may use a hyphen if you prefer), and "meta-annotations" is hyphenated.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -428,6 +447,45 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+     * @param annotationmirrorType the class type used to cast the element value
```
mernst: This comment says how the value is used, but not what it means.  Focus on how a client should use the method, not on how the method uses the formal parameter in its implementation.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -428,6 +447,45 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
                 result.add(new Postcondition(expr, postcondAnno, anno));
             }
         }
+
+        return result;
+    }
+
+    /**
+     * Returns the set of repeatable postconditions.
```
mernst: What set?

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -428,6 +447,45 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+     *
+     * @param metaAnno set of post condition meta annotations
+     * @param annotationmirrorType the class type used to cast the element value
+     * @param elementValue the element value
```
mernst: This comment adds no information.  It just repeats the parameter name.  Write informative documentation.  Don't just add an `@param` to fill up space or to satisfy a code style requirement.

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOf.java
gokaco: @mernst I have edited the documentation of this. Please give a feedback on that.

framework/src/main/java/org/checkerframework/framework/qual/ConditionalPostconditionAnnotation.java
```diff
@@ -13,9 +13,10 @@
  *
  * <p>The annotation E that is meta-annotated as {@link ConditionalPostconditionAnnotation} must
  * have an element called {@code expression} that is an array of {@code String}s of the same format
```
mernst: This is not true.
`EnsuresLTLengthOfIf.List` is meta-annotated with `@ConditionalPostconditionAnnotation`, but it does not have an element called `expression`.

checker/src/main/java/org/checkerframework/checker/index/qual/EnsuresLTLengthOf.java
mernst: Your edits leave the documentation wrong.  Perhaps I should take over this pull request, since you are having trouble completing it.

framework/src/main/java/org/checkerframework/framework/qual/ConditionalPostconditionAnnotation.java
gokaco: @mernst Please read the whole paragraph of the documentation

mernst: > @mernst Please read the whole paragraph of the documentation

I did.  By the way, it is rude to write a short response that says I didn't do my work, but that does not respond to my comment and that does nothing to help resolve the misunderstanding.

Let's read this together,  The sentence is:
"The annotation E ... must have an element called `expression` that is an array of Strings ... or a list of conditional postcondition annotations ...".
In other words, the sentence says that there must be an element called `expression`, which might have one of two different types.  That seems to be in conflict with its use in the code.

Am I reading the sentence wrong, reading the code wrong, or have I made some other mistake?  Please clarify.

gokaco: > I did. By the way, it is rude to write a short response that says I didn't do my work, but that does not respond to my comment and that does nothing to help resolve the misunderstanding.

@mernst Sorry If you felt like that. That was not the intention.
> Let's read this together, The sentence is:
> "The annotation E ... must have an element called `expression` that is an array of Strings ... or a list of conditional postcondition annotations ...".
> In other words, the sentence says that there must be an element called `expression`, which might have one of two different types. That seems to be in conflict with its use in the code.

I meant to say that after before and after `or` they are both different sentences irrespective of each other. I think you are saying to write the documentation like this-
```
The annotation E that is meta-annotated as {@link ConditionalPostconditionAnnotation} must have an element called {@code expression} that is an array of {@code String}s of the same format and with the same meaning as the value {@code expression} in {@link EnsuresQualifierIf} or it must have a list of conditional postcondition annotations in an array{@code value} and with the same meaning as the {@code value} in {@link EnsuresQualifiersIf}. E must also have an element {@code result} with the same meaning as the element {@code result} in {@link EnsuresQualifierIf}.
```
Is this correct?

mernst: That is better -- at least it is not wrong as the previous version was.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
gokaco: I meant to say that it returns a collection of repeatable annotations.
Would the below line would be correct?
```Returns repeatable annotations```

File: framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java (hunk 21 above)
Comment (gokaco) [reply framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java]: I have updated the methods and documentation. Please give a review.

File: framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java (hunk 20 above)
Comment (gokaco) [reply framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java]: I have updated the methods and documentation. Please give a review.

File: framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java (hunk 19 above)
Comment (gokaco) [reply framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java]: @mernst Can you give a hint on how can I write its documentation? I have updated the methods and documentation but I am having trouble with this. Please help!!

File: framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java (hunk 18 above)
Comment (gokaco) [reply framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java]: I have updated the documentation but I am having trouble with this. Please help in this!!

File: framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java (hunk 17 above)
Comment (gokaco) [reply framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java]: I have updated the method. Is it right now?

File: framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java [hunk 23]
```diff
@@ -413,21 +414,83 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+        @SuppressWarnings("unchecked")
+        List<AnnotationValue> la =
+                AnnotationUtils.getElementValue(anno, "value", List.class, false);
+
+        Set<Postcondition> result = new LinkedHashSet<>();
+
+        // Check for multiple contracts.
+        List<AnnotationMirror> annotations = new ArrayList<>(la.size());
+        for (AnnotationValue a : la) {
+            if (a.getClass().getName().equals("javax.lang.model.element.AnnotationMirror")
```
wmdietl: Why the comparison against two Strings? Wouldn't `a instanceof AnnotationMirror` be enough to cover both cases?

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -413,21 +414,83 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+                    || a.getClass()
+                            .getName()
+                            .equals("com.sun.tools.javac.code.Attribute$Compound")) {
+                annotations.add(annotationmirrorType.cast(a.getValue()));
```
wmdietl: I don't see a reason to pass `annotationmirrorType` in as a parameter. Can you explain?

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -413,21 +414,83 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+            AnnotationMirror postcondAnno = getAnnotationMirrorOfMetaAnnotation(metaAnno, a);
             if (postcondAnno == null) {
                 continue;
             }
+            for (String expr : expression) {
+                result.add(new Postcondition(expr, postcondAnno, a));
+            }
+        }
+
+        // Check for a single contract.
```
wmdietl: Can you explain what is happening here? `la` didn't get re-assigned, so why another iteration over it?
Couldn't you merge this loop into the other loop over `la` and cover the two possible cases once?

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -413,21 +414,83 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+        List<String> expressions = new ArrayList<>(la.size());
+
+        for (AnnotationValue a : la) {
+            if (a.getClass().getName().equals("java.lang.String")
+                    || a.getClass()
+                            .getName()
+                            .equals("com.sun.tools.javac.code.Attribute$Constant")) {
+                expressions.add(stringType.cast(a.getValue()));
```
wmdietl: How is this working? The cast will work if it is a `String`. But if it is a `Attribute$Constant` why would a cast to the final class `String` work?

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -413,21 +414,83 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
 
         // Check type-system specific annotations.
         Class<PostconditionAnnotation> metaAnnotation = PostconditionAnnotation.class;
+        Class<String> stringType = String.class;
```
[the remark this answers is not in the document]
gokaco: > Why do you introduce local variables and parameters for `stringType` and `annotationmirrorType`?

The main motive is to make there class object which is used later to cast annotation value.
> In particular the code also contains `String.class` so it it seems rather strange to have parameters for this.

Yes, you are right that it contains `String.class`, but I want to make an Object of it which can be later used for casting. That's why I used it. Is there any other way?

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
gokaco: Yes, It worked.
Thanks

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
gokaco: Yes, we can define it in the function too. I have updated it. Please review it.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
gokaco: > Can you explain what is happening here? 

It is adding all the annotation value to its list if a single annotation is present.
> `la` didn't get re-assigned, so why another iteration over it?
 Couldn't you merge this loop into the other loop over `la` and cover the two possible cases once?

Yes, you are right. I have merged both the cases in a single loop.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
gokaco: > How is this working? The cast will work if it is a `String`. But if it is a `Attribute$Constant` why would a cast to the final class `String` work?

@wmdietl The value for an annotation element of String is `Attribute$Constant`([link](https://static.javadoc.io/org.kohsuke.sorcerer/sorcerer-javac/0.11/com/sun/tools/javac/code/Attribute.Constant.html)).
So if  `a` is `Attribute$Constant` then we should cast to String.
P.S.- We do not need to check that if the value class is `String`. I have updated the code.
Please review it.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -482,16 +527,77 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
-                result.add(new ConditionalPostcondition(expr, annoResult, postcondAnno, anno));
+
+            result.addAll(getConditionalPostConditionAnnotations(anno, metaAnno));
+        }
+
+        return result;
+    }
+
+    /**
+     * Returns conditional postcondition annotations.
```
mernst: Similar comment to above about comment:  initial comment is too vague, `methodElement` doesn't exist.
Make all suggestions from elsewhere to this method (documentation and code).

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -418,16 +419,60 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
-            AnnotationMirror postcondAnno = getAnnotationMirrorOfMetaAnnotation(metaAnno, anno);
+
+            result.addAll(getPostConditionAnnotations(anno, metaAnno));
+        }
+
+        return result;
+    }
+
+    /**
+     * Returns postcondition annotations.
```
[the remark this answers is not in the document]
gokaco: I have updated the documentation. Please review it!

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -418,16 +419,60 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+     *
+     * @param anno the type qualifier that {@code methodElement} should have
```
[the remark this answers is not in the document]
gokaco: I have updated the documentation. Please review it!

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -418,16 +419,60 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+     * @param metaAnno postCondition meta-annotation
```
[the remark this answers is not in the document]
gokaco: I have updated the description. Please review it!

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -418,16 +419,60 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+     */
+    private Set<Postcondition> getPostConditionAnnotations(
+            AnnotationMirror anno, AnnotationMirror metaAnno) {
+
+        Class<String> stringType = String.class;
```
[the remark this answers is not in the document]
gokaco: I have removed the two variables.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -418,16 +419,60 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+        Class<AnnotationMirror> annotationmirrorType = AnnotationMirror.class;
+
+        @SuppressWarnings("unchecked")
+        List<AnnotationValue> la =
```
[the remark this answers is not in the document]
gokaco: I have changed the name to `elementvalue`. Is it correct now?

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -418,16 +419,60 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+                AnnotationUtils.getElementValue(anno, "value", List.class, false);
+
+        Set<Postcondition> result = new LinkedHashSet<>();
+        List<AnnotationMirror> annotations = new ArrayList<>(la.size());
```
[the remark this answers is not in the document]
gokaco: Yes, I agree that it doesn't seem necessary to allocate `la.size()` elements for both of them. I have updated it.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -418,16 +419,60 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+        List<String> expressions = new ArrayList<>(la.size());
+        for (AnnotationValue a : la) {
+            if (a instanceof AnnotationMirror) {
+                annotations.add(annotationmirrorType.cast(a.getValue()));
+            } else if (a.getClass()
+                    .getName()
+                    .equals("com.sun.tools.javac.code.Attribute$Constant")) {
```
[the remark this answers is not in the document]
gokaco: Yes, I was able to use `instanceof`. I have updated it.
Thanks.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -418,16 +419,60 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+                expressions.add(stringType.cast(a.getValue()));
+            }
```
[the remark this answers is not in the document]
gokaco: Fall through is not expected as only two conditions are possible(a list of annotation mirrors or a single annotation).

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -418,16 +419,60 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+                result.add(new Postcondition(expr, postcondAnno, a));
+            }
+        }
+        AnnotationMirror postcondAnno = getAnnotationMirrorOfMetaAnnotation(metaAnno, anno);
+        if (postcondAnno != null) {
             for (String expr : expressions) {
                 result.add(new Postcondition(expr, postcondAnno, anno));
             }
         }
+
```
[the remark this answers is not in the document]
gokaco: Blank line removed.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -418,16 +419,60 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+        Set<Postcondition> result = new LinkedHashSet<>();
+        List<AnnotationMirror> annotations = new ArrayList<>(la.size());
+        List<String> expressions = new ArrayList<>(la.size());
+        for (AnnotationValue a : la) {
+            if (a instanceof AnnotationMirror) {
+                annotations.add(annotationmirrorType.cast(a.getValue()));
+            } else if (a.getClass()
+                    .getName()
+                    .equals("com.sun.tools.javac.code.Attribute$Constant")) {
+                expressions.add(stringType.cast(a.getValue()));
```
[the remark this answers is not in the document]
gokaco: Thanks for the suggestion. Changed to a regular Java cast.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -418,16 +419,60 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+            }
+        }
+
+        for (AnnotationMirror a : annotations) {
+            List<String> expression =
```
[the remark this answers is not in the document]
gokaco: I have moved the statement.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
gokaco: I have updated the documentation. Please review it!

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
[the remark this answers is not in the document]
mernst: In that case, please add an else clause that throws an exception, to give an indication if something has gone wrong.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -417,19 +418,61 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
-            AnnotationMirror postcondAnno = getAnnotationMirrorOfMetaAnnotation(metaAnno, anno);
+
+            result.addAll(getPostConditionAnnotations(anno, metaAnno));
+        }
+
+        return result;
+    }
+
+    /**
+     * Returns a collection of postcondition annotations that can consist of
```
mernst: You still have not addressed the previous code review feedback.
This comment just states the return type, which is obvious from the method signature.
You can remove this redundant text, and instead explain the relationship between the inputs and outputs of the method.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -417,19 +418,61 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+     * {@code @EnsuresNonNull},{@code @EnsuresKeyFor},{@code @EnsuresLTLengthOf},
+     * {@code @EnsuresLockHeld} or any other postcondition annotation developed in the future.
+     *
+     * @param anno wrapper annotation if multiple annotations are present, otherwise single
```
mernst: What kind of single annotation?  This is a place where type information would be useful.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -417,19 +418,61 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+     *     annotation
+     * @param metaAnno postcondition meta-annotation that indicates that an annotation is a
```
mernst: Why is it necessary?  How is it used by the method?

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -417,19 +418,61 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+     *     postcondition annotation
+     */
+    private Set<Postcondition> getPostConditionAnnotations(
+            AnnotationMirror anno, AnnotationMirror metaAnno) {
+
+        @SuppressWarnings("unchecked")
+        List<AnnotationValue> elementvalue =
```
mernst: This name is not very helpful, because it's just the same as the method that returned it.  Since this is the `value` field of the variable `anno`, `annoValue` would be more appropriate.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -417,19 +418,61 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+                AnnotationUtils.getElementValue(anno, "value", List.class, false);
+
+        Set<Postcondition> result = new LinkedHashSet<>();
```
mernst: You have not answered my question about the relationship between these two variables.
Is it the case that one of them is always empty?  (I think so.)

More generally, the flow of control in this method is very confusing.  It would be better to do a single test, at the beginning of the method, and put all the code for handling a wrapper in one branch and all the code for handling a single annotation in the other branch.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -483,23 +526,85 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
-                result.add(new ConditionalPostcondition(expr, annoResult, postcondAnno, anno));
+
+            result.addAll(getConditionalPostConditionAnnotations(anno, metaAnno));
+        }
+
+        return result;
+    }
+
+    /**
+     * Returns a collection of conditional postcondition annotations that can consist of
```
mernst: Similar comment as above about the Javadoc comment.

Also a similar comment about the logic:  How about doing a single test outside the `for` loop rather than a test on each iteration through the `for` loop?

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -483,23 +526,85 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+     *     annotation
+     * @param @param metaAnno conditional postcondition meta-annotation that indicates that an
+     *     annotation is a conditional postcondition annotation
+     */
+    private Set<ConditionalPostcondition> getConditionalPostConditionAnnotations(
+            AnnotationMirror anno, AnnotationMirror metaAnno) {
+
+        Set<ConditionalPostcondition> result = new LinkedHashSet<>();
+
+        Map<? extends ExecutableElement, ? extends AnnotationValue> valmap;
```
mernst: This variable name communicates nothing about what the variable represents.  Use a better name, and perhaps add a comment.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -483,23 +526,85 @@ private AnnotationMirror getAnnotationMirrorOfMetaAnnotation(
...
+        valmap = anno.getElementValues();
```
mernst: Combine this with the declaration statement immediately above.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
gokaco: @mernst What do you mean by `type information` here?
Thanks

mernst: Information about the type of annotation represented by `anno`.  Might that be (a wrapper for) *any* annotation?  Or should clients only pass certain types of arguments?

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
gokaco: Combined.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
gokaco: Channged the name to `annovalue` and added a comment.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
gokaco: Removed for loop.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
gokaco: Yes, you are right. Divided it in two blocks.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
gokaco: Changed `elementvalue` to `annovalue`.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
gokaco: I have updated the documentation that how it is used. Please review it.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
gokaco: I have updated about the type of annotation(i.e. postcondition annotation) and that clients should pass only postcondition annotation. Please review it.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
gokaco: I have updated in the documentation that what is happening inside the code with respect to input and what is the output. Please review it.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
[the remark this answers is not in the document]
gokaco: Added an exception

gokaco:
I have added some documentation where I found it to be useful. I have added some more methods to make the logic more clear.

PR #2628, opened 2019-07-15
Postconditions should refine the refined type of an expression

### LINKED ISSUE #2619: Postconditions should union their arguments
For some postconditions, at a call site the arguments should be unioned into any existing arguments.

After
```
  aux1.hasValue(Aux.MINIMUM_VALUE) && aux2.hasValue(Aux.MINIMUM_VALUE)
```

`Aux.MINIMUM_VALUE` should be a key for both `aux1.map` and `aux2.map`.  But, it is only a key for one of the maps (the first one).
Methods `m01` through `m03` show this behavior.
This is related to #2358.

```
// Compile this file with:
// javac -g PostconditionUnion.java -processor nullness -ArequirePrefixInWarningSuppressions

import java.util.HashMap;
import java.util.Map;
import org.checkerframework.checker.nullness.qual.EnsuresKeyForIf;

public class PostconditionUnion {

  void m01(Aux aux1, Aux aux2) {
    if (aux1.hasValue(Aux.MINIMUM_VALUE) && aux2.hasValue(Aux.MINIMUM_VALUE)) {
      @KeyFor("aux1.map") String s1 = Aux.MINIMUM_VALUE;
    }
  }

  void m02(Aux aux1, Aux aux2) {
    if (aux1.hasValue(Aux.MINIMUM_VALUE) && aux2.hasValue(Aux.MINIMUM_VALUE)) {
      // error: [keyfor:assignment.type.incompatible] incompatible types in assignment.
      @KeyFor("aux2.map") String s2 = Aux.MINIMUM_VALUE;
    }
  }

  void m03(Aux aux1, Aux aux2) {
    if (aux2.hasValue(Aux.MINIMUM_VALUE)) {
      @KeyFor("aux2.map") String s2 = Aux.MINIMUM_VALUE;
    }
  }

}

class Aux {

  public Map<String, String> map = new HashMap<>();

  public static final String MINIMUM_VALUE = "minvalue";

  @Pure
  @EnsuresKeyForIf(result = true, expression = "#1", map = "map")
  public boolean hasValue(String key) {
    return map.containsKey(key);
  }

  @Pure
  public int getInt(@KeyFor("this.map") String key) {
    return 22;
  }
}
```

Here is another example, from the Called Methods Checker.
(The `@EnsuresCalledMethods` annotation does not exist yet.)

```
// Declaration
@EnsuresCalledMethods(receiver="#1", methods={"z"})
callsZ(p) {
  p.z();
}

// Uses

@CalledMethods({"x", "y"}) arg1;
callsZ(arg1);
@CalledMethods({"x", "y", "z"}) after1 = arg1;

@CalledMethods({"v", "w"}) arg2;
callsZ(arg2);
@CalledMethods({"v", "w", "z"}) after2 = arg2;
```

smillst:
Postconditions should refine the *refined* type of an expression so that the type of an expression can become more and more precise.  This is done by storing the glb of the precondition qualifier and the value in the store.  If the glb is bottom, just use the new value.

Fixes #2619.


framework/src/main/java/org/checkerframework/framework/flow/CFAbstractStore.java
```diff
@@ -254,6 +255,46 @@ public void insertValue(FlowExpressions.Receiver r, AnnotationMirror a) {
...
+    /**
+     * Add the annotation {@code a} for the expression {@code r} (correctly deciding where to store
+     * the information depending on the type of the expression {@code r}).
+     *
+     * <p>This method does not take care of removing other information that might be influenced by
+     * changes to certain parts of the state.
+     *
+     * <p>If there is already a value {@code v} present for {@code r}, then the greatest lower bound
+     * of the new and old value is inserted into the store unless it's bottom. If the greates lower
+     * bound is bottom, then the new value is inserted.
```
wmdietl: What is the intuition behind this special treatment for the bottom qualifier? I don't see why this is different from other GLBs.

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractStore.java
```diff
@@ -254,6 +255,46 @@ public void insertValue(FlowExpressions.Receiver r, AnnotationMirror a) {
...
     }
 
+    /**
+     * Add the annotation {@code a} for the expression {@code r} (correctly deciding where to store
+     * the information depending on the type of the expression {@code r}).
+     *
+     * <p>This method does not take care of removing other information that might be influenced by
+     * changes to certain parts of the state.
+     *
+     * <p>If there is already a value {@code v} present for {@code r}, then the greatest lower bound
```
wmdietl: Is this only safe if the method is also pure? If the method is non-pure, it could break any other property and the only thing we can rely on is the postcondition from this method, nothing that held before.

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractStore.java
```diff
@@ -254,6 +255,46 @@ public void insertValue(FlowExpressions.Receiver r, AnnotationMirror a) {
...
+     * of the new and old value is inserted into the store unless it's bottom. If the greates lower
+     * bound is bottom, then the new value is inserted.
+     *
+     * <p>Note that this happens per hierarchy, and if the store already contains information about
+     * a hierarchy other than {@code a}s hierarchy, that information is preserved.
+     */
+    public void insertGlb(FlowExpressions.Receiver r, AnnotationMirror newAnno) {
```
wmdietl: Would `insertOrRefine` be a better name instead of `insertGlb`?

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractStore.java
```diff
@@ -254,6 +255,46 @@ public void insertValue(FlowExpressions.Receiver r, AnnotationMirror a) {
         insertValue(r, analysis.createSingleAnnotationValue(a, r.getType()));
     }
 
+    /**
+     * Add the annotation {@code a} for the expression {@code r} (correctly deciding where to store
```
wmdietl: The second parameter is called `newAnno`. So it would be better to either use that name here instead of `a` or change the parameter name to `a`.

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractStore.java
smillst: For qualifiers that take arguments that don't override `QualifierHierarchy#greatestLowerBound`,  glb is not actually the *greatest* lower bound, but rather a lower bound.  That's why it's not safe to add bottom.  (I'll update the Javadoc.)

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractStore.java
smillst: If the method isn't side-effect-free, then the values that might have been changed by the method call are removed from the store before postcondition values are added.  I'll add a test for this.

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractStore.java
smillst: If bottom, it might be from an incorrect answer and to be conservative about the incorrect value we forbid bottom.

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractStore.java
smillst: I'll add a note, too.

smillst: I added this comment to the callsite in CFAbstractTransfer.

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractStore.java
```diff
@@ -254,6 +255,48 @@ public void insertValue(FlowExpressions.Receiver r, AnnotationMirror a) {
...
+     * changes to certain parts of the state.
+     *
+     * <p>If there is already a value {@code v} present for {@code r}, then the greatest lower bound
+     * of the new and old value is inserted into the store unless it's bottom. Some checkers do not
+     * override {@link QualifierHierarchy#greatestLowerBound(AnnotationMirror, AnnotationMirror)}
+     * and the default implementation will return the bottom qualifier incorrectly. So this method
+     * conservatively does not insert the glb if it is bottom.
+     *
+     * <p>Note that this happens per hierarchy, and if the store already contains information about
+     * a hierarchy other than {@code a}s hierarchy, that information is preserved.
```
wmdietl: The `a` should be `newAnno`, I think.

PR #2631, opened 2019-07-16
Clear typeArg cache in DefaultTypeHierarchy

smillst:
Otherwise, errors for identical subtyping checks are only issued once, even if the check was for a different expressions.

### DIFF
--- a/checker/tests/nullness/WildcardSubtype.java
+++ b/checker/tests/nullness/WildcardSubtype.java
@@ -1,0 +1,54 @@
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+class WildcardSubtype {
+    class MyClass {}
+
+    class Visitor<T> {
+        String visit(T p) {
+            return "";
+        }
+    }
+
+    class MyClassVisitor extends Visitor<@Nullable MyClass> {}
+
+    class NonNullMyClassVisitor extends Visitor<@NonNull MyClass> {}
+
+    void test(MyClassVisitor myClassVisitor, NonNullMyClassVisitor nonNullMyClassVisitor) {
+        // :: error: (argument.type.incompatible)
+        take(new Visitor<@Nullable Object>());
+        // :: error: (argument.type.incompatible)
+        take(new Visitor<@Nullable Object>());
+
+        Visitor<?> visitor1 = myClassVisitor;
+        Visitor<?> visitor2 = nonNullMyClassVisitor;
+
+        // :: error: (assignment.type.incompatible)
+        Visitor<? extends @NonNull Object> visitor3 = myClassVisitor;
+        Visitor<? extends @NonNull Object> visitor4 = nonNullMyClassVisitor;
+
+        // :: error: (assignment.type.incompatible)
+        Visitor<? extends @NonNull Object> visitor5 = new MyClassVisitor();
+        // :: error: (assignment.type.incompatible)
+        Visitor<? extends @NonNull Object> visitor6 = new MyClassVisitor();
+        // :: error: (argument.type.incompatible)
+        take(new MyClassVisitor());
+        // :: error: (argument.type.incompatible)
+        take(new MyClassVisitor());
+    }
+
+    void take(Visitor<@NonNull ? extends @NonNull Object> v) {}
+
+    void bar() {
+        // :: error: (argument.type.incompatible)
+        take(new Visitor<@Nullable Object>());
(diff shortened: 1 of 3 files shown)


checker/tests/nullness/WildcardSubtype.java
```diff
@@ -0,0 +1,54 @@
...
+        Visitor<? extends @NonNull Object> visitor5 = new MyClassVisitor();
+        // :: error: (assignment.type.incompatible)
+        Visitor<? extends @NonNull Object> visitor6 = new MyClassVisitor();
+        // :: error: (argument.type.incompatible)
+        take(new MyClassVisitor());
+        // :: error: (argument.type.incompatible)
+        take(new MyClassVisitor());
+    }
+
+    void take(Visitor<@NonNull ? extends @NonNull Object> v) {}
```
smillst: Add a test case with two wildcards that are the same annotated type.

framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java
```diff
@@ -145,6 +145,7 @@ protected StructuralEqualityComparer createEqualityComparer() {
     @Override
     public boolean isSubtype(
             final AnnotatedTypeMirror subtype, final AnnotatedTypeMirror supertype) {
+        typeargVisitHistory.clear();
```
smillst: This is not just a performance cache, it prevents infinite recursion.

framework/src/main/java/org/checkerframework/framework/type/StructuralEqualityVisitHistory.java
```diff
@@ -0,0 +1,57 @@
+package org.checkerframework.framework.type;
+
+import java.util.Optional;
+import javax.lang.model.element.AnnotationMirror;
+
+/** Stores the result of {@link StructuralEqualityComparer} for type arguments. */
```
wmdietl: It would be useful to get some high-level information about how this differs from the `SubtypeVisitHistory` and what the true/false histories are for.

framework/src/main/java/org/checkerframework/framework/type/StructuralEqualityVisitHistory.java
```diff
@@ -0,0 +1,57 @@
...
+public class StructuralEqualityVisitHistory {
+
+    /** Types in this history are structural equal. */
```
wmdietl: Should `structural` be `structurally`? Same below.

mernst:
> One high-level thought: is the use of `Optional<Boolean>` nicer than using `@Nullable Boolean`?
> In all code so far we've used `@Nullable Boolean` and I don't quite see a benefit to using an Optional here, i.e. there is no streaming going on.

I tend to [agree](https://homes.cs.washington.edu/~mernst/advice/nothing-is-better-than-optional.html) that we should avoid `Optional` when possible.

framework/src/main/java/org/checkerframework/framework/type/StructuralEqualityVisitHistory.java
```diff
@@ -0,0 +1,61 @@
+package org.checkerframework.framework.type;
+
+import javax.lang.model.element.AnnotationMirror;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * Stores the result of {@link StructuralEqualityComparer} for type arguments.
+ *
+ * <p>This is simalar to {@link SubtypeVisitHistory}, but both true and false results are stored.
```
wmdietl: simalar -> similar

wmdietl: `similar` seems a strange word - this class uses two `SubtypeVisitHistory` fields as implementation.

framework/src/main/java/org/checkerframework/framework/type/StructuralEqualityVisitHistory.java
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * Return whether or not the two types are structurally equal for the given hierarchy or {@code
+     * null} if no result exists for the types.
+     */
+    public @Nullable Boolean result(
+            final AnnotatedTypeMirror type1,
+            final AnnotatedTypeMirror type2,
+            AnnotationMirror hierarchy) {
+        if (falseHistory.contains(type1, type2, hierarchy)) {
+            return false;
```
wmdietl: You could directly use `Boolean.FALSE` and `Boolean.TRUE` below.

Issue #1635, opened 2017-11-06
Reference to varargs is wrong at call sites


panacekcz:
When an annotation refers to a varargs parameter such as `String get(@IndexFor("#2") int i, String... varargs)`, at call sites, this reference is resolved to the second argument (first element of the varargs array), not the array itself.

This issue can probably only show when using the index checker, but it seems to be a general problem of the framework (in `FlowExpressionParseUtil.parseParameter`).

Example code:
```java
import org.checkerframework.checker.index.qual.IndexFor;

public class IndexForVarargs {
	String get(@IndexFor("#2") int i, String... varargs) {
		return varargs[i];
	}
	
	void m() {
		get(1, "a", "b"); // line 9
		get(2, "abc"); // line 10
	}
}
```
Expected output:
- no error at line 9
- upper bound error at line 10

Run the checker as:
```
checker/bin/javac -processor org.checkerframework.checker.index.IndexChecker IndexForVarargs.java
```

Output:
```
IndexForVarargs.java:9: error: [argument.type.incompatible] incompatible types in argument.
		get(1, "a", "b");
		    ^
  found   : @UpperBoundUnknown int
  required: @LTLengthOf(value=""a"", offset={}) int
1 error
```

smillst:
At the call `get(1, "a", "b")`, 
`@IndexFor(#1)` should translate to `@LTLengthOf("new String[] {"abc"}")`.  I can implement that.  However, `FlowExpressionParseUtil#parse` can't parse array creations, so `ValueAnnotatedTypeFactory#getMinLenFromString` still returns 0.  

Updating `FlowExpressionParseUtil#parse` is the right way to fix this, but we want to replace it with a real parser (#1614). 

In the mean time, I could probably implement a work around for the Upper Bound Checker.

PR #2634, opened 2019-07-18
FlowExpressionParser replacement

### LINKED ISSUE #1614: Replace the flow expression parser with a real parser

Maxi17:
This version works correctly but it's not complete. Every `parseX` method can be optimized or removed. I added comments to explain the changes.

Double literals can now be parsed.
Fixes #1614


smillst:
Is this a fix for https://github.com/typetools/checker-framework/issues/1614?

Maxi17:
> Is this a fix for #1614?

Yes. Sorry for omitting it. I updated the comment.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -139,54 +129,96 @@ private static Pattern anchored(String regex) {
...
+                return new ValueLiteral(
+                        types.getPrimitiveType(TypeKind.DOUBLE),
+                        expr.asDoubleLiteralExpr().asDouble());
+            } else if (expr.isStringLiteralExpr() && !context.parsingMember) {
+                TypeElement stringTypeElem =
+                        env.getElementUtils().getTypeElement("java.lang.String");
+                return new ValueLiteral(
+                        types.getDeclaredType(stringTypeElem),
+                        expr.asStringLiteralExpr().asString());
+            } else if (expr.isThisExpr() && !context.parsingMember) {
```
panacekcz: `ThisExpr` can also refer to an enclosing instance. After a quick check, I am not sure if the dataflow framework can handle them. For now, I'd at least add a check that the `ThisExpr` has no type name.
(In the dataflow framework, `ThisReference` confusingly seems to be used both for supertypes and outer types, which might be a bug.)

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -487,95 +384,72 @@ private static Receiver getReceiverField(
...
             throw constructParserException(s, "no parameter found");
         }
-        int idx = -1;
-        try {
-            idx = Integer.parseInt(parameterMatcher.group(1));
-        } catch (NumberFormatException e) {
-            // cannot occur by the way the pattern is defined (matches only numbers)
-            assert false;
-        }
+        int idx = Integer.parseInt(s.substring(7));
```
kelloggm: Where does the number 7 come from here?

Maxi17: It's because `"_param_".length()  = 7`, and the substring that start after `"_param_"` is the parameter number. Parameters are now recognized as `"_param_"` instead of `"#"`.

kelloggm: Okay, that makes sense. Can you add a comment that explains the magic number here?

Maxi17:
I also made some small performance improvements that are not related to the JavaParser replacement. Hopefully these won't be a problem.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -119,236 +86,158 @@ private static Pattern anchored(String regex) {
...
-
         Receiver receiver;
         String memberSelected;
 
         Resolver resolver = new Resolver(env);
 
         // Attempt to match a package and class name first.
         Pair<ClassName, String> classAndRemainingString =
-                matchPackageAndClassNameWithinExpression(s, resolver, path);
+                matchPackageAndClassNameWithinExpression(expr.toString(), resolver, path);
```
kelloggm: Do you still need this call? Or would it be sufficient to just use the `else` branch of the `if` statement below and remove this method?

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -119,236 +86,158 @@ private static Pattern anchored(String regex) {
...
-        return floatMatcher.matches();
-    }
+        Pair<ClassName, String> classAndRemainingString =
+                matchPackageAndClassNameWithinExpression(expr.toString(), resolver, path);
 
-    /** Return true iff s is a string literal. */
-    private static boolean isStringLiteral(String s, FlowExpressionContext context) {
-        if (context.parsingMember) {
-            return false;
+        if (classAndRemainingString != null) {
```
kelloggm: Similar question - could we not just avoid this call entirely and use a call to `parseHelper` with `expr.getTypeAsString()` as the first argument?

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -487,95 +361,73 @@ private static Receiver getReceiverField(
...
-        Pair<Pair<String, String>, String> method = parseMethodCall(s);
-        if (method == null) {
-            return null;
+        String s = expr.toString();
+        Resolver resolver = new Resolver(env);
+
+        // methods with scope, like "item.get()", need special treatment.
+        if (expr.getScope().isPresent()) {
+            Pair<ClassName, String> classAndRemainingString =
+                    matchPackageAndClassNameWithinExpression(s, resolver, path);
```
kelloggm: Again, this code is written as if you don't need this call to succeed.

Maxi17:
I tried to remove `matchPackageAndClassNameWithinExpression`. It performed the job of a symbol solver, so the change I made is just a way around it. Anyways, I think this method and other helper methods I deleted at the latest commit are necessary because they throw exceptions with useful and descriptive messages. My 3-line version of it, found in `parseClassExpression`, doesn't.

I will revert to the previous commit if you think this change isn't good enough.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -119,279 +83,161 @@ private static Pattern anchored(String regex) {
...
-            }
-        } else {
-            String receiverString = select.first;
-            memberSelected = select.second;
-            receiver = parseHelper(receiverString, context, path);
+        // A class with its package specified, like "java.lang.String", is treated as field access.
+        // This code checks it and parses it as a class expression if necessary.
+        if (ElementUtils.getType(resolver.findClass(expr.getNameAsString(), path)) != null) {
+            return parseHelper(
+                    StaticJavaParser.parseExpression(expr.toString() + ".class"), context, path);
```
Maxi17: This is bad, but I couldn't find something better.

smillst:
@Maxi17 said:
>I tried to remove `matchPackageAndClassNameWithinExpression`. It performed the job of a symbol solver, so the change I made is just a way around it. Anyways, I think this method and other helper methods I deleted at the latest commit are necessary because they throw exceptions with useful and descriptive messages. My 3-line version of them, found in parseClassExpression, doesn't.

I think you should remove the `matchPackageAndClassNameWithinExpression` method.  The method(s) you replace it with should do something similar , only using JavaParser expressions instead of Strings.  I've added a comment to get you started.

Maxi17:
I made the changes. If the expression is a `ClassExpr`, then it either goes in `parseIdentifier` (in case of `String.class`) or `parseMemberSelect` (in case of `java.lang.String.class`) after the `parseHelper` call. Both `parseIdentifier` and `parseMemberSelect` work correctly. I haven't added the exception handling of the old `matchPackageAndClassNameWithinExpression`.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -119,279 +82,156 @@ private static Pattern anchored(String regex) {
...
         }
 
-        String receiver = s.substring(0, i);
-        String remaining = s.substring(i + 1);
-
-        return Pair.of(receiver, remaining);
+        return updatedExpression;
     }
 
     private static Receiver parseMemberSelect(
```
smillst: Please add a comment that the selected member could be a field, class, or package; but it can't be a field.  Also, many methods are missing Javadoc in this class; please make add documentation to every method.  If it's a method you didn't add and don't understand, let me know and I can document it.  Thanks!

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -119,279 +82,156 @@ private static Pattern anchored(String regex) {
...
+                return parseParameter(expr.asNameExpr().getNameAsString(), context);
+            } else if (expr.isNameExpr()) {
+                return parseIdentifier(expr.asNameExpr().getNameAsString(), env, path, context);
+            } else if (expr.isMethodCallExpr()) {
+                return parseMethodCall(expr.asMethodCallExpr(), context, path, env);
+            } else if (expr.isFieldAccessExpr()) {
+                return parseMemberSelect(expr.asFieldAccessExpr(), env, context, path);
+            } else if (expr.isClassExpr() && !context.parsingMember) {
+                return parseHelper(
+                        StaticJavaParser.parseExpression(expr.asClassExpr().getTypeAsString()),
```
smillst: Please add a comment here explaining why it is necessary to re-parse the type string.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -119,279 +82,156 @@ private static Pattern anchored(String regex) {
...
+                    && expr.asNameExpr().getNameAsString().startsWith("_param_")
+                    && !context.parsingMember) {
+                return parseParameter(expr.asNameExpr().getNameAsString(), context);
+            } else if (expr.isNameExpr()) {
+                return parseIdentifier(expr.asNameExpr().getNameAsString(), env, path, context);
+            } else if (expr.isMethodCallExpr()) {
+                return parseMethodCall(expr.asMethodCallExpr(), context, path, env);
+            } else if (expr.isFieldAccessExpr()) {
+                return parseMemberSelect(expr.asFieldAccessExpr(), env, context, path);
+            } else if (expr.isClassExpr() && !context.parsingMember) {
```
smillst: Please add a comment here that a "ClassExpr" is a class literal.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -119,279 +82,156 @@ private static Pattern anchored(String regex) {
...
         if (result instanceof ClassName && !expression.endsWith("class")) {
             throw constructParserException(
                     expression, "a class name cannot terminate a flow expression string");
         }
         return result;
     }
 
-    private static FlowExpressions.Receiver parseHelper(
-            String expression, FlowExpressionContext context, TreePath path)
+    private static Receiver parseHelper(
```
smillst: This method is no longer "parsing" it's converting JavaParser Expressions to FlowExpression.Receivers. The method name (and those that it calls) should reflect that.

Maxi17:
Some methods still need Javadoc. I will write these. The test I added is superficial, but it only needs to prove that parsing floats/doubles is now permitted.

Maxi17:
@kelloggm @panacekcz @smillst 
I hope I added everything. If the Javadoc or the `ParsingDouble` test case needs improvement, please let me know (or you can update them directly). Otherwise, I think this is the final version of the replacement.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -119,286 +85,180 @@ private static Pattern anchored(String regex) {
...
-        } else if (isMemberSelect(expression)) {
-            return parseMemberSelect(expression, env, context, path);
-        } else if (isParentheses(expression, context)) {
-            return parseParentheses(expression, context, path);
-        } else {
-            String message;
-            if (expression.equals("#0")) {
-                message =
-                        "one should use \"this\" for the receiver or \"#1\" for the first formal parameter";
+        if (expr != null) {
```
smillst: The call to this method issue an error if `expr` is null.  That way this method never returns null.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -119,286 +85,180 @@ private static Pattern anchored(String regex) {
...
+            if (expr.isNullLiteralExpr()) {
+                return new ValueLiteral(types.getNullType(), (Object) null);
+            } else if (expr.isIntegerLiteralExpr()) {
+                return new ValueLiteral(
+                        types.getPrimitiveType(TypeKind.INT), expr.asIntegerLiteralExpr().asInt());
+            } else if (expr.isLongLiteralExpr()) {
+                return new ValueLiteral(
+                        types.getPrimitiveType(TypeKind.LONG), expr.asLongLiteralExpr().asLong());
+            } else if (expr.isDoubleLiteralExpr()) {
+                return new ValueLiteral(
```
smillst: Please update https://checkerframework.org/manual/#java-expressions-as-arguments so that it matches the new implementation.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -119,286 +85,180 @@ private static Pattern anchored(String regex) {
...
-            return parseMemberSelect(expression, env, context, path);
-        } else if (isParentheses(expression, context)) {
-            return parseParentheses(expression, context, path);
-        } else {
-            String message;
-            if (expression.equals("#0")) {
-                message =
-                        "one should use \"this\" for the receiver or \"#1\" for the first formal parameter";
+        if (expr != null) {
+            if (expr.isNullLiteralExpr()) {
```
smillst: This giant if/elseif/else statement can be re-written using a com.github.javaparser.ast.visitor.GenericVisitorWithDefaults.  The class would look something like this:

```
    public static class ExpressionToReceiverVisitor extends GenericVisitorWithDefaults<Receiver, FlowExpressionContext> {
        private final TreePath path;
        public ExpressionToReceiverVisitor(TreePath path) {
            this.path = path;
        }
        @Override
        public Receiver defaultAction(com.github.javaparser.ast.Node n, FlowExpressionContext arg) {
            // throw a parse exception saying something about an unsupported expressions type.
            return null;
        }

        @Override
        public Receiver visit(NullLiteralExpr n, FlowExpressionContext arg) {
              return new ValueLiteral(types.getNullType(), (Object) null);
        }
        // Add a visit method for each kind of Expression that we handle.
    }
```

Then this if/elseif/else can be replaced with   `expr.accept(new ExpressionToReceiverVisitor(path), context);`.
In the Visitor, don't call the private static methods, but rather copy the implementation of those methods to the proper visit method.

smillst: (You'll need to pull and rebuild the stubparser, because I just pushed a fix for GenericVisitorWithDefaults.)

Maxi17: If I replace evert `getXReceiver` method with  `visit` method inside `ExpressionToReceiverVisitor`, then some of them need to throw a `FlowExpressionParseException`. In this case they can't override the methods in `GenericVisitorWithDefaults`. One solution for this would be making `FlowExpressionParseException` an unchecked exception (extend `RuntimeException` instead of `Exception`). Should I do this?

Maxi17: Also, can you please explain what you meant here:

> You'll need to pull and rebuild the stubparser, because I just pushed a fix for GenericVisitorWithDefaults.

I'm not sure what I need to do.

Maxi17:
Supporting `char` literals is simple. Supporting `String` concatenation and mathematical operators can be done using [BinaryExpr](https://static.javadoc.io/com.github.javaparser/javaparser-core/3.4.3/com/github/javaparser/ast/expr/BinaryExpr.html). Something else that can be supported is [ObjectCreationExpr](https://static.javadoc.io/com.github.javaparser/javaparser-core/3.1.2/com/github/javaparser/ast/expr/ObjectCreationExpr.html), for #1635.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
smillst: I mean do a `git pull` in your clone of stubparser and rebuild.  (I think running   `./gradlew cloneAndBuildDependencies` in your Checker Framework directory will work, too.) 

>Update: Does it have something to do with the fact I need to provide an implementation for visit(StubUnit, A)?

Yes, that's what I fixed:
https://github.com/typetools/stubparser/commit/1737a1910b1d26544d8bbb527686d566361f4e3d

smillst: > One solution for this would be making `FlowExpressionParseException` an unchecked exception (extend RuntimeException instead of Exception).

This would work, but we want it to be a checked exception so that clients catch it.  Another option would be to wrap the `FlowExpressionParseException` in a runtime exception.  For example,
```
throw new ParserRuntimeException( constructParserException(...));
```
The the parse method would catch the ParserRuntimeException and then throw the FlowExpressionParseException.

Or the  visitor could return a pair of a Receiver and an FlowExpressionParseException.  
```
class ExpressionToReceiverVisitor extends <Pair<Receiver, ExpressionToReceiverVisitor>, FlowExpressionContext>
```
Then either the receiver is null or the exception is null.  If the visitor returns a non-null exception, then the parse method will throw it.  

I haven't thought about which approach would be better.

smillst:
>Supporting char literals is simple. Supporting String concatenation and mathematical operators can be done using BinaryExpr. Something else that can be supported is ObjectCreationExpr, for #1635.

Great!  Please add support for char literals in this pull request.  Once this pull request is merged, then you can work on add support for the other Expressions.

Maxi17:
@smillst I made the changes. There are still some problems:

1. I had to declare the `ProcessingEnvironment` and `Types` variables in many methods, instead of just once. I think there must be a better way for this, but in the constructor I can't access the `FlowExpressionContext` variable passed as argument.

2. For now, I made `FlowExpressionParseException` an unchecked exception, waiting for something better.

3. Char literals are not permitted, so I will add a test case.

4. I removed unhelpful Javadoc, but haven't added any.

smillst:
>1. I had to declare the `ProcessingEnvironment` and `Types` variables in many methods, instead of just once. I think there must be a better way for this, but in the constructor I can't access the `FlowExpressionContext` variable passed as argument.

You can pass the `ProcessingEnvironment` to the constructor, then uses fields.

>2. For now, I made `FlowExpressionParseException` an unchecked exception, waiting for something better.

Sounds good.  You can figure that out later.  (I added a TODO so I don't merge that change into master.)

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -119,344 +97,390 @@ private static Pattern anchored(String regex) {
...
     }
 
-    private static FlowExpressions.Receiver parseHelper(
-            String expression, FlowExpressionContext context, TreePath path)
+    /**
+     * Converts the passed Javaparser Expression to a {@link Receiver}.
+     *
+     * @return the Receiver of the passed expression
+     */
+    private static Receiver getReceiverFromExpression(
```
smillst: This method should just be removed.  The calls to it in the visitor can be replaced with `expr.accept(this, context)`.

Maxi17:
@smillst I finished it. The checked exception is now wrapped into an unchecked exception inside the class.

framework/tests/flowexpression/ClassLiterals.java
```diff
@@ -0,0 +1,30 @@
...
+        // :: error: (assignment.type.incompatible)
+        @FlowExp("String.class") Object l3 = p3;
+        // :: error: (assignment.type.incompatible)
+        @FlowExp("java.lang.String.class") Object l4 = p1;
+        // :: error: (assignment.type.incompatible)
+        @FlowExp("java.lang.String.class") Object l5 = p2;
+        @FlowExp("java.lang.String.class") Object l6 = p3;
+    }
+
+    @FlowExp("int.class") String s0;
```
smillst: The annotations on fields s0-s3 cause the framework to crash.  It's not critical that these expressions are supported, but they shouldn't cause a crash.  (Also it should be noted in the manual that not all class literals work.)

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -119,935 +102,452 @@ private static Pattern anchored(String regex) {
...
-            if (memberSelected == null) {
-                throw constructParserException(
-                        s, "a class cannot terminate a flow expression string");
-            }
-        } else {
-            String receiverString = select.first;
-            memberSelected = select.second;
-            receiver = parseHelper(receiverString, context, path);
-        }
+    private static class ExpressionToReceiverVisitor
```
smillst: Needs a class comment.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -119,935 +102,452 @@ private static Pattern anchored(String regex) {
...
 
-    private static Receiver parseNullLiteral(Types types) {
-        return new ValueLiteral(types.getNullType(), (Object) null);
-    }
-
-    private static boolean isIntLiteral(String s, FlowExpressionContext context) {
-        if (context.parsingMember) {
-            return false;
+        @Override
+        public Receiver defaultAction(
```
smillst: Needs a comment.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -119,935 +102,452 @@ private static Pattern anchored(String regex) {
...
+                com.github.javaparser.ast.Node n, FlowExpressionContext context) {
+            String message = "is an unrecognized expression";
```
smillst: I think `is not a supported expression` is better than `is an unrecognized expression`.  The expression parsed correctly; it just isn't supported.

Maxi17:
> Also have you tested this branch on any of the Index Checker case studies?

@smillst No, I haven't. But I would appreciate some tips on how to do it.

mernst:
A paper about the Index Checker case studies is here:  https://homes.cs.washington.edu/~mernst/pubs/array-indexing-issta2018.pdf
The three main case studies in that paper are:
 * Guava:  annotated code is at https://github.com/panacekcz/guava/tree/index-checker-suppress but should be moved into https://github.com/typetools/guava
 * JFreeChart: annotated code is at https://github.com/typetools/jfreechart but should be moved into https://github.com/kelloggm/jfreechart/tree/index 
 * plume-lib:  annotated code is in https://github.com/smillst/plume-lib/tree/index-casestudy; however, that repository is now obsolete, and ongoing work should use the Java repositories at https://github.com/plume-lib.

I have given the repositories for the initial work and where the annotations belong.  I have not checked whether the annotations have been moved; if not, could you please move them?

smillst:
@Maxi17  Mike pointed out where the Index Checker case studies are.  (@kelloggm  and @panacekcz can help with building them.)  I would like you to run each one with typetools/master and this branch and then compare time.  I'm only looking for them to take roughly the same amount of time.

Maxi17:
@mernst @smillst I have run the Guava case study and plume-lib case study on both `master` and `issue1614v3` branches multiple times. Running `mvn process-sources` on Guava, the results were roughly the same (on average, 1:02 minutes for both branches, on my machine). For plume-lib, running `make` produced, again, similar results between branches (on average, 38 seconds). Each run of these 2 case studies took either 1-2 seconds longer or 1-2 seconds shorter, but there was no visible evidence to support one implementation.

For JFreeChart, the annotations are still on a PR.

PR #2642, opened 2019-07-23
Character.charValue is non-negative

### LINKED ISSUE #2638: Character.charValue() should be @NonNegative
Consider the following code:
```
class tryit{
        public static String toString(final @NonNegative Character ch) {
        	return toString(ch.charValue());
    }
}
```
Run it by: `javac -processor index tryit.java`
Output:
```
tryit.java:4: error: [argument.type.incompatible] incompatible types in argument.
        	return toString(ch.charValue());
        	                            ^
  found   : @LowerBoundUnknown char
  required: @NonNegative Character
1 error

```
However, ch is annotated to be `@NonNegative`

Maxi17:
Fixes #2638 

Added `@NonNegative` annotation to `Character.charValue()`.
Related: https://github.com/typetools/annotated-libraries/pull/29


Maxi17:
The tests failed because I didn't pull the latest version of the annotated JDK from typetools/annotated-libraries before changing the JdkHash, so the changes in https://github.com/typetools/checker-framework/pull/2567 were not applied. Will fix it. (The test case for this issue passed)

PR #2643, opened 2019-07-24
Fix Issue #2432 (Attempt 2): move poly.resolve before typevar substitution

abslamp:
Test case and another possible fix on #2432

Move poly.resolve() before typevar substitution in asMemberOf(). 3 versions of asMemberOf are spitted to two parts: before and after substitution. The default asMemberOf() is kept as a wrapper of the two methods to avoid alternation of other code.

Note: this is a quick prototype to prove logic correctness (by trigger the tests), still need to remove some code duplications.


framework/src/main/java/org/checkerframework/framework/util/AnnotatedTypes.java
```diff
@@ -353,13 +372,54 @@ public static AnnotatedTypeMirror asMemberOf(
...
+                type = asMemberOfImpl(types, atypeFactory, t, elem, type);
+
                 if (!ElementUtils.isStatic(elem)) {
                     atypeFactory.postAsMemberOf(type, t, elem);
                 }
                 return type;
         }
     }
 
+    public static AnnotatedTypeMirror asMemberOfPreSubstitution(
```
wmdietl: Can you add javadoc to the modified methods to make it clear what they are doing.

wmdietl: `PreSubstitution` sounds wrong - this is doing a substitution. Once there is javadoc of what the method does we can find a better name.

framework/src/main/java/org/checkerframework/framework/util/AnnotatedTypes.java
```diff
@@ -390,8 +450,35 @@ private static AnnotatedTypeMirror asMemberOfImpl(
...
             case UNION:
+            case DECLARED:
+                // return substituteTypeVariables(types, atypeFactory, of, member, memberType);
+                return memberType;
+            default:
+                throw new BugInCF("asMemberOf called on unexpected type.\nt: " + of);
+        }
+    }
+
+    private static AnnotatedTypeMirror asMemberOfImpl(
```
wmdietl: Again, javadoc to make clear what the method does.
Why is this extra method needed? Instead, couldn't you split up the original method into two separate methods?

framework/src/main/java/org/checkerframework/framework/util/AnnotatedTypes.java
abslamp: Maybe the asMemberOfImpl is making the name confusing. Actually it does not do a substitution. So I will make asMemberOfImpl more clear.

abslamp:
The last commit is reversed to make the PR clearer.

checker/tests/nullness/Issue2432.java
```diff
@@ -0,0 +1,112 @@
+// Test case for issue 2432:
```
wmdietl: Above this in the PR I see the addition of two files: `checker/annotation-tools` and `checker-stubparser`.
However, I can't edit nor delete those files.
Can you remove those changes?

checker/tests/nullness/Issue2432.java
```diff
@@ -0,0 +1,112 @@
...
+// https://github.com/typetools/checker-framework/issues/2432
+
+import java.util.List;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.checkerframework.checker.nullness.qual.PolyNull;
+
+class Issue2432 {
```
wmdietl: Can you clean up this test case a bit?
Add a bit of structure and clean naming. Explain a bit (very briefly) what is tested in each section.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1609,6 +1613,7 @@ protected AnnotatedDeclaredType getImplicitReceiverType(ExpressionTree tree) {
                 if (path == null) {
                     // The path is null if the field is in a compilation unit we haven't
                     // processed yet. TODO: is there a better way?
+                    // This only arises in the Nullness Checker when substituting rawness.
```
wmdietl: Why did you add this comment? What do you mean with `rawness`?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1211,7 +1211,11 @@ private AnnotatedTypeMirror fromExpression(ExpressionTree tree) {
 
         AnnotatedTypeMirror result = TypeFromTree.fromExpression(this, tree);
 
-        if (shouldCache) {
+        if (shouldCache && tree.getKind() != Tree.Kind.NEW_CLASS) {
+            // Don't cache types of NEW operator
```
wmdietl: How about:

```
// Don't cache the type of object creations, because incorrect
// annotations would be cached during dataflow analysis.
// See Issue #602.
```

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -223,6 +223,10 @@ public GenericAnnotatedTypeFactory(BaseTypeChecker checker, boolean useFlow) {
             flowResultAnalysisCaches = null;
         }
 
+        // Add common aliases.
```
wmdietl: What is the point of this change?

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1579,10 +1583,19 @@ public ParameterizedExecutableType methodFromUse(MethodInvocationTree tree) {
...
             dependentTypesHelper.viewpointAdaptMethod(tree, method);
         }
-        poly.resolve(tree, method);
+
         return mType;
     }
 
+    @Override
+    public void methodFromUsePreSubstitution(ExpressionTree tree, AnnotatedTypeMirror mirror) {
+        // no call to super needed since super is no-op
```
wmdietl: Well, there is a call to super on the next line... why assume that the behavior in the super method will never change?

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1579,10 +1583,19 @@ public ParameterizedExecutableType methodFromUse(MethodInvocationTree tree) {
...
+        super.methodFromUsePreSubstitution(tree, mirror);
+        if (tree instanceof MethodInvocationTree && mirror instanceof AnnotatedExecutableType) {
```
wmdietl: What other trees/types could be passed in here?
The method name makes it sound like this would only ever make sense for these more specific types.

framework/src/main/java/org/checkerframework/framework/util/AnnotatedTypes.java
```diff
@@ -364,8 +392,9 @@ private static AnnotatedTypeMirror asMemberOfImpl(
             final Types types,
             final AnnotatedTypeFactory atypeFactory,
             final AnnotatedTypeMirror of,
-            final Element member) {
-        AnnotatedTypeMirror memberType = atypeFactory.getAnnotatedType(member);
+            final Element member,
+            AnnotatedTypeMirror memberType) {
```
wmdietl: Can you document the method and additional parameter?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
abslamp: That's odd... It's not added by me, I will check whether there is any merge issues on my branch.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
abslamp: Still not added by me

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
abslamp: I will remove this line

checker/tests/nullness/Issue2432.java
abslamp: I cannot remove either, every time I tried to delete or edit github would return a 404 page.

abslamp:
Pull request closed due to uncorrectable or historical issues.

PR #2646, opened 2019-07-25
Added ValueChecker support for Math.min and Math.max

### LINKED ISSUE #1687: Value Checker support for Math#min and Math#max
The Value Checker has no annotations on `java.util.Math#min` and `java.util.Math#max`, resulting in false positives in computations involving those functions. Consider:

```java
void mathTest(@IntRange(from = 0, to = 20) int x, @IntRange(from = 5, to = 15) int y) {
     @IntRange(from = 0, to = 5) int min = Math.min(x, y);
     @IntRange(from = 15, to = 20) int max = Math.max(x, y);
}
```
Expected: no warnings
Result:
```java
tests/value/MathMinMax.java:5: error: [assignment.type.incompatible] incompatible types in assignment.
        @IntRange(from = 0, to = 5) int min = Math.min(x, y);
                                                      ^
  found   : @UnknownVal int
  required: @IntRange(from=0, to=5) int
tests/value/MathMinMax.java:6: error: [assignment.type.incompatible] incompatible types in assignment.
        @IntRange(from = 15, to = 20) int max = Math.max(x, y);
                                                        ^
  found   : @UnknownVal int
  required: @IntRange(from=15, to=20) int
2 errors
```

Maxi17:
Fixes #1687

`Math.min` and `Math.max` now have a default `@IntRange` annotation if both their arguments have an `@IntRange` annotation.


framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1302,6 +1302,30 @@ private boolean methodIsStaticallyExecutable(Element method) {
 
         @Override
         public Void visitMethodInvocation(MethodInvocationTree tree, AnnotatedTypeMirror type) {
+            if ((getMethodIdentifier().isMathMin(tree, processingEnv)
+                            || getMethodIdentifier().isMathMax(tree, processingEnv))
+                    && type.hasAnnotation(UNKNOWNVAL)) {
+                AnnotatedTypeMirror atm1 = getAnnotatedType(tree.getArguments().get(0));
```
kelloggm: I suggest giving these variables somewhat more descriptive names. I would choose `argType1` and `argType2`, but you might prefer something else.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1302,6 +1302,41 @@ private boolean methodIsStaticallyExecutable(Element method) {
...
+            if (getMethodIdentifier().isMathMin(tree, processingEnv)
+                    && type.hasAnnotation(UNKNOWNVAL)) {
+                AnnotatedTypeMirror argType1 = getAnnotatedType(tree.getArguments().get(0));
+                AnnotatedTypeMirror argType2 = getAnnotatedType(tree.getArguments().get(1));
+                if (argType1.hasAnnotation(IntRange.class)
+                        && argType2.hasAnnotation(IntRange.class)) {
+                    long from =
+                            Math.min(
+                                    getFromValueFromIntRange(argType1),
+                                    getFromValueFromIntRange(argType2));
```
Maxi17: Assignments like this can probably be split because they don't look well.

kelloggm: They do not. I suggest pulling the calls to `getFromValueFromIntRange` and `getToValueFromIntRange` out into locals.

I would also move all of this logic into a private helper method.

Maxi17:
@kelloggm I think it looks better now.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1300,8 +1300,42 @@ private boolean methodIsStaticallyExecutable(Element method) {
...
 
+        /**
+         * @return The Range of the Math.min or Math.max method, or null if the argument is none of
+         *     these methods or their arguments are not annotated with @IntRange.
+         */
+        private Range getRangeForMathMinMax(MethodInvocationTree tree) {
+            if (getMethodIdentifier().isMathMin(tree, processingEnv)) {
+                AnnotatedTypeMirror arg1 = getAnnotatedType(tree.getArguments().get(0));
+                AnnotatedTypeMirror arg2 = getAnnotatedType(tree.getArguments().get(1));
+                if (arg1.hasAnnotation(IntRange.class) && arg2.hasAnnotation(IntRange.class)) {
```
smillst: This logic belongs in the `org.checkerframework.common.value.util.Range` class.  You should add a `Range min(Range other)` and a `Range max(Range other)` methods.

checker/tests/index/MathMinMax.java
```diff
@@ -0,0 +1,8 @@
+import org.checkerframework.common.value.qual.IntRange;
```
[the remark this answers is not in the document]
smillst: Ping.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1300,8 +1300,78 @@ private boolean methodIsStaticallyExecutable(Element method) {
...
```
smillst: Use org.checkerframework.common.value.ValueAnnotatedTypeFactory#getRange instead.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1300,8 +1300,78 @@ private boolean methodIsStaticallyExecutable(Element method) {
...
+                }
+            }
+            return null;
+        }
+
+        /**
+         * @return The constant value of the Math.min or Math.max method, or null if the argument is
+         *     none of these methods or their arguments are not constants.
+         */
+        private Long getMathMinMaxExactValue(MethodInvocationTree tree) {
```
smillst: If you use `org.checkerframework.common.value.ValueAnnotatedTypeFactory#getRange` then this method won't be needed.

PR #2650, opened 2019-07-25
Add rules for @PolySigned; fixes #2543

### LINKED ISSUE #2543: Signedness Checker permits polymorphic type when not all the instantiations are legal
When a method that contains a `@Poly...` annotation type-checks, we expect all instantiations of it to type-check.  That is not the case for the below example.  The `@PolySigned` version of the method type-checks, but the `@Signed` version fails type-checking:

```
javac -processor signedness RotateRight.java
RotateRight.java:13: error: [shift.unsigned] UNSIGNED_RIGHT_SHIFT operation on a signed value
        return (i >>> distance) | (i << -distance);
                ^
RotateRight.java:21: error: [return.type.incompatible] incompatible types in return.
        return (i >>> distance) | (i << -distance);
                                ^
  type of expression: @UnknownSignedness int
  method return type: @Unsigned int
2 errors
```

```
import org.checkerframework.checker.signedness.qual.PolySigned;
import org.checkerframework.checker.signedness.qual.Signed;
import org.checkerframework.checker.signedness.qual.Unsigned;

public class RotateRight {

    public static @PolySigned int rotateRight(@PolySigned int i, int distance) {
        return (i >>> distance) | (i << -distance);
    }

    public static @Signed int rotateRightSigned(@Signed int i, int distance) {
        return (i >>> distance) | (i << -distance);
    }

    public static @Unsigned int rotateRightUnsigned(@Unsigned int i, int distance) {
        return (i >>> distance) | (i << -distance);
    }

    public static @Unsigned int rotateRightUnknownSignedness(@UnknownSignedness int i, int distance) {
        return (i >>> distance) | (i << -distance);
    }
}
```

gokaco:
When this pull request will get merged, Checker Framework will issue an error when `@PolySigned` will perform the following operations-

- Divide or Modulus Operation
- Signed right shift Operation
- Unsigned right shift Operation
- Non-Equality comparisons

Fixes: #2543


gokaco:
@mernst It is ready for review!

checker/src/main/java/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -301,23 +305,27 @@ public Void visitBinary(BinaryTree node, Void p) {
         switch (kind) {
             case DIVIDE:
             case REMAINDER:
-                if (leftOpType.hasAnnotation(Unsigned.class)) {
+                if (leftOpType.hasAnnotation(Unsigned.class)
```
smillst: Instead of changing every use of `atm.hasAnnotation(Unsigned.class)` to  `atm.hasAnnotation(Unsigned.class) || atm.hasAnnotation(PolyUnsigned.class)`, create a method that does that and call it multiple times.

gokaco: Added two methods, one for unsigned and second for signed.

PR #2659, opened 2019-07-26
Better error messages in UpperBoundChecker

### LINKED ISSUE #1902: Poor error message when Index Checker fails to identify required @LTOMLengthOf(String[] names) or @LTLengthOf(value = “”, offset = “” )
Consider:
```java
public class SimpleMethod3 {
   @NonNegative int size;
   int[] vDown = new int[size];
   //simpler version
   void method3(@NonNegative int size,@NonNegative int value){
       this.size = size;
       this.vDown = new int[this.size];
       vDown[1 + value] = 10;
   }
   }
}
```

Received warnings:
```java
Error:(14, 17) java: [array.access.unsafe.high.range] Potentially unsafe array access: the index could be larger than the array's bound
  index type found: @IntRange(from=-2147483648, to=2147483647) int
  array type found: @UnknownVal int @UnknownVal []
  required        : index of type @IndexFor("this.vDown") or @LTLengthOf("this.vDown"), or array of type @MinLen("2147483648")
```
None of the prompted required type would work. On the other hand,  @LTLengthOf(value = "this.vDown", offset = "1") or @LTOMLengthOf(“this.vDown”) will pass the checker. Also, @MinLen("2147483648") don't make much sense to me, why does array need to have minimum of 2147483648 element? I think it should be removed. To be more precise, the checker should prompt:
```java
 required: index of type@LTLengthOf(value = "this.vDown", offset = “”), or @LTOMLengthOf(“this.vDown”)
```
If this is a valid concern, I would like to work on it.

Maxi17:
Fixes #1902 

Issue an `array.access.unsafe.high` error instead of `array.access.unsafe.high.range` when the maximum value of the index is `Integer.MAX_VALUE`.

The error message
```
Error:(14, 17) java: [array.access.unsafe.high.range] Potentially unsafe array access: the index could be larger than the array's bound
  index type found: @IntRange(from=-2147483648, to=2147483647) int
  array type found: @UnknownVal int @UnknownVal []
  required        : index of type @IndexFor("this.vDown") or @LTLengthOf("this.vDown"), or array of type @MinLen("2147483648")
```
becomes 
```
ErrorMessageCheck.java:11: error: [array.access.unsafe.high] Potentially unsafe array access: the index could be larger than the array's bound
        vDown[1 + value] = 10;
                ^
  required: @IndexFor("this.vDown") or @LTLengthOf("this.vDown") -- an integer less than this.vDown's length
1 error
```

I hope this is the desired output.


Maxi17:
I think the downstream tests fail because of some Guava tests.

mernst:
> I think the downstream tests fail because of some Guava tests.

I have improved the CI error reporting.  There are 4 errors while type-checking Guava.
Could you fix them, in a branch of your fork of Guava named `issue1902` (just like this branch)?
Thanks.

PR #2660, opened 2019-07-26
Remove uses of "implicit" that now refer to a default qualifier.

smillst:
Remove uses of the word "implicit" that refer to the old notion of "an implicit qualifier" which now should just be a "default qualifier".


mernst:
@smillst The title is 
> Remove uses of "implicit" that refer to an implicit qualifier.

 Should there be a negation, or a different word, in there?

smillst:
I mean:

Remove uses of the word "implicit" that refer to the old notion of "an implicit qualifier" which now should just be a "default qualifier".  

Does that make sense?

framework/src/main/java/org/checkerframework/framework/qual/LiteralKind.java
```diff
@@ -11,8 +11,7 @@
  * that enum is in the tools.jar which is not on the user's classpath by default. So this enum is
  * used instead.
  *
- * @checker_framework.manual #creating-type-introduction Declaratively specifying implicit
- *     annotations
+ * @checker_framework.manual #creating-type-introduction Declaratively specifying annotations
```
mernst: Should this mention defaults?  The related section is about type introduction, not about specifying annotations per se.

framework/src/main/java/org/checkerframework/framework/qual/UpperBoundFor.java
```diff
@@ -19,7 +19,7 @@
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.ANNOTATION_TYPE)
 public @interface UpperBoundFor {
-    /** @return {@link TypeKind}s of types for which an annotation should be implicitly added */
+    /** @return {@link TypeKind}s of types for which an annotation should be added */
```
mernst: "Added by default?"

PR #2666, opened 2019-07-29
Support .java files as stub files.

wmdietl:


smillst:
A jtreg test is failing:
https://dev.azure.com/typetools/checker-framework/_build/results?buildId=255&view=logs&jobId=1da990f5-78b3-53bf-a5ca-1c432c06bef7&taskId=870b748f-1fb0-5090-e972-5fc3c6e5f396&lineStart=726&lineEnd=727&colStart=1&colEnd=1

checker/jtreg/stubs/Issue1542Driver.java
```diff
@@ -3,7 +3,7 @@
  * @summary Test case for Issue 1542 https://github.com/typetools/checker-framework/issues/1542
  *
  * @compile -XDrawDiagnostics issue1542/NeedsIntRange.java issue1542/Stub.java issue1542/ExampleAnno.java
- * @compile -XDrawDiagnostics -processor org.checkerframework.common.value.ValueChecker issue1542/UsesIntRange.java -Astubs=issue1542/ -AstubWarnIfNotFound -Werror
+ * @compile -XDrawDiagnostics -processor org.checkerframework.common.value.ValueChecker issue1542/UsesIntRange.java -Astubs=issue1542/NeedsIntRange.astub:issue1542/Stub.astub -AstubWarnIfNotFound -Werror
```
smillst: This means that -Astubs can't point to a directory containing Java files, right?  

Also,  The documentation in https://checkerframework.org/manual/#stub-using also needs to be updated:
>specifying a directory or .jar file is equivalent to specifying every file in it whose name ends with .astub. 

Or we may just not want to push this change to master.

wmdietl: It was useful to have this feature for experiments without an annotated JDK - I could simply take the same files as stub files.
I'm not sure how useful this is in general. I'm fine if we decide against this and will undo this change in eisop.

mernst: I believe that Suzanne's proposal is to undo this change in eisop, but retain the `eisop-javastubs` branch for merging at some future date (after the 3.0.0 release, which is more important than this change).

smillst: I've opened a pull request to undo this in eisop:
https://github.com/eisop/checker-framework/pull/41

PR #2669, opened 2019-07-29
@SignednessGlb annotation should not be applied to compound expressions if not annotated explicitly; fixes #2482

### LINKED ISSUE #2482: Compound assignment expression producing error using signedness checker
Consider the following code:-

Command-
```
javac -processor signedness ex.java
```
Code-
```
class ex {

    void regularAssignment(byte[] b, int c) {
        int a = b.length;
        a = a + c;
    }

    void compoundAssignment(byte[] b, int c) {
        int a = b.length;
        a += c; // Error
    }
}
```
Output:-
```
ex.java:10: error: [compound.assignment.type.incompatible] expression type incompatible with left-hand side in compound assignment.
        a += c; // Error
          ^
  found   : @Signed int
  required: @Constant int
1 error
```
Ordinary assignments succeed, whereas compound assignments do not.

gokaco:
Fix: #2482 

The `@SignednessGlb` annotation should not be applied to compound expressions if not annotated explicitly.


smillst:
Please use a descriptive title, not just "Fixes Issue2482 ". You should also add "fixes #NNNN" which is useful, but does not describe the work you have done.

checker/src/main/java/org/checkerframework/checker/signedness/SignednessAnnotatedTypeFactory.java
```diff
@@ -104,32 +106,47 @@ private void addSignednessGlbAnnotation(Tree tree, AnnotatedTypeMirror type) {
...
-                                if (treeRange.isWithin(0, Long.MAX_VALUE)) {
-                                    type.replaceAnnotation(SIGNEDNESS_GLB);
-                                }
-                                break;
-                            default:
-                                // Nothing
+                    if (tree.getKind() == Tree.Kind.PLUS_ASSIGNMENT
+                            || tree.getKind() == Tree.Kind.MULTIPLY_ASSIGNMENT
+                            || tree.getKind() == Tree.Kind.DIVIDE_ASSIGNMENT
+                            || tree.getKind() == Tree.Kind.MINUS_ASSIGNMENT) {
```
mernst: Why doesn't this handle `%=`, `>>=`, `>>>=`, and `<<=`?

More generally, special-casing these operations feels wrong.
Why change refinement here, rather than changing type-checking rules elsewhere?'
Maybe this is the right approach, but can you justify this design?

gokaco: For `%=`, `>>=`, `>>>=`, and `<<=` we get `treeRange` in positive signed range. So LHS and RHS satisfy each other as `@SignednessGlb` applies on both sides and no error occurs.

For `+=`, `-=`, `/=`, and `*=` we get `treeRange` from `type.MIN_VALUE` to `type.MAX_VALUE` , therefore, `@SignednessGlb` only gets applied to LHS, hence error occurs because LHS conflicts with RHS.

Therefore special casing is needed for these expressions so they do not produce error.

P.S.- if an expression is `len %=a;` then `len` is LHS and `len%a` is RHS; `type` can be `Integer`or `Byte` etc......as applicable.

mernst: You said "special casing is needed", but you made no change.  Did you mean "special casing is **not** needed"?

mernst:
Please use descriptive commit messages.  "Update file" does not describe the change, so it is not helpful in understanding what you have done.

checker/src/main/java/org/checkerframework/checker/signedness/SignednessAnnotatedTypeFactory.java
mernst: Can you clarify your comments?

> For `%=`, `>>=`, `>>>=`, and `<<=` we get `treeRange` in positive signed range.

I don't understand this.
`%` can return a negative number, but you seem to be saying that it cannot.

> So LHS and RHS satisfy each other as `@SignednessGlb` applies on both sides and no error occurs.

"No error occurs" is ambiguous.  Do you mean no compile-time warning, no run-time misbehavior, or something else?

As I have pointed out before, when there is a question about your code, it is generally good to improve the code or its comments rather than just responding in the code review.  This is also explained at https://homes.cs.washington.edu/~mernst/advice/github-pull-request.html#respond-to-reviews which you said you have read.

gokaco: > I don't understand this.
> `%` can return a negative number, but you seem to be saying that it cannot.

It can return a negative number but when a user initialise the variable with a **positive value** then the line-
`Range treeRange = IndexUtil.getPossibleValues(valueATM, valueFactory);` 
the value `treeRange` we get is always in positive signed range, that's not the case for `+=`,`-=`,`*=`,`/=`. for ex. when we compile-
```
class ex {

    void assignment(byte[] b, int a,int c) {
    	int len =2; // or b.length i.e. any positive value
        len %= a;
    }
}
```
and observe the `treeRange` in the `SignednessAnnotatedTypeFactory.java` then it shows the value of `len %= a`as `[0..2147483647]`
and if we replace `len %= a` with `len += a` the value `treeRange` of `len +=a` comes to be as
`[-2147483648..2147483647]`

P.S.- I have updated it in the documentation

gokaco: > You said "special casing is needed", but you made no change. Did you mean "special casing is **not** needed"?

I meant to say special conditions for `+=`,`/=',`*=',`-=` which I have committed.

checker/src/main/java/org/checkerframework/checker/signedness/SignednessAnnotatedTypeFactory.java
```diff
@@ -115,12 +115,12 @@ private void addSignednessGlbAnnotation(Tree tree, AnnotatedTypeMirror type) {
...
                         AnnotatedTypeMirror varType = getAnnotatedType(var);
                         AnnotatedTypeMirror exprType = getAnnotatedType(expr);
-                        AnnotatedTypeMirror lht = getAnnotatedType(var);
                         if (!(varType.hasAnnotation(Unsigned.class)
                                         && exprType.hasAnnotation(Signed.class))
                                 && !(varType.hasAnnotation(Signed.class)
-                                        && exprType.hasAnnotation(Unsigned.class)))
-                            type.replaceAnnotations(lht.getAnnotations());
+                                        && exprType.hasAnnotation(Unsigned.class))
+                                && !varType.hasExplicitAnnotation(SignednessGlb.class))
```
gokaco: If the variable is annotated explicitly with `@SigndnessGlb` then the type annotation should not get replaced.

gokaco: > Please use descriptive commit messages. "Update file" does not describe the change, so it is not helpful in understanding what you have done.

I have explained the work of that commit in this comment.

checker/src/main/java/org/checkerframework/checker/signedness/SignednessAnnotatedTypeFactory.java
gokaco: >>"No error occurs" is ambiguous. Do you mean no compile-time warning, no run-time misbehavior, or something else?

Yes, no compile-time warning, no run-time misbehavior.

mernst:
You remarked that in the process of creating this pull request, you discovered multiple bugs in the Signedness Checker.  Can you please make issues for them, so that we can determine which ones need to be fixed before this is merged?

gokaco:
> The correct way to fix this bug is to override `getAnnotatedTypeLhs`, which I've now done. Please let me know if you have any questions.

This looks good. You are right that if `@SignednessGlb` will not get applied while computing LHS then the bug will get fixed.

PR #2670, opened 2019-07-29
Fix .svg to prevent error: `convert-im6.q16: non-conforming drawing p…

wmdietl:
…rimitive definition text-align`.


smillst:
Azure passed, so I'm going to merge this.

PR #2672, opened 2019-07-30
Workaround Java 8 bug (Fixes #2173)

### LINKED ISSUE #2173: ElementAnnotationUtil.getLocationTypeADT: invalid location [WILDCARD] for type: String;
Bug report copied from message by William Shackleford https://groups.google.com/d/msg/checker-framework-dev/jrllYpfH1GQ/nTmnDJ1vAQAJ

I have a maven project that I have integrated checker-framework into that 
has started failing with the message:

```
[ERROR] ElementAnnotationUtil.getLocationTypeADT: invalid location [WILDCARD] for type: String; The Checker Framework crashed.  Please report the crash.
```

(The same message is repeated 10 times.)

Unfortunately I don't see where in my code the issue is and the project is too large to paste/attach to a bug report.

If someone has time they may be able to reproduce the problem by checking out and building the project with:

```
git clone https://github.com/usnistgov/crcl.git
cd crcl
# switch to dev2 branch
git checkout dev2
cd tools/java/crcl4java/crcl4java-ui/
mvn clean compile -Pwith_checkers
```

For me  mvn -version produces:

```
Apache Maven 3.5.2
Maven home: /usr/share/maven
Java version: 1.8.0_181, vendor: Oracle Corporation
Java home: /home/shackle/jdk1.8.0_181/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "4.15.0-34-generic", arch: "amd64", family: "unix"
```

In case I make changes in the future and you need to reproduce the problem
the current latest commit  is  64cfb6dd871a106bb49360afdebd3e5e5de95d3c

smillst:


framework/src/main/java/org/checkerframework/framework/util/element/ElementAnnotationUtil.java
```diff
@@ -463,11 +463,18 @@ private static AnnotatedTypeMirror getLocationTypeADT(
             }
         }
 
-        if (outerToInner.isEmpty() || error) {
+        if (outerToInner.isEmpty()) {
```
wmdietl: Thanks for working on this!
I don't think this fully fixes this bug. Any of the other BugInCF errors in this file could also be raised, for example if the annotation is on an array or wildcard.
So instead of changing all these conditions, how about introducing a new exception that gets thrown here and caught in the top-level method in this class? Then there we can decide whether to raise an error, a warning, or do nothing.

framework/src/main/java/org/checkerframework/framework/util/element/ElementAnnotationUtil.java
```diff
@@ -463,11 +463,18 @@ private static AnnotatedTypeMirror getLocationTypeADT(
...
                     "ElementAnnotationUtil.getLocationTypeADT: invalid location %s for type: %s",
                     location, type);
         }
+
+        // There's a bug in Java 8 compiler that creates bad bytecode such that an annotation
+        // on a lambda parameter is applied to a method parameter.  If this happens, then the
+        // location could refer to a type argument that doesn't exist.  Since Java 8 bytecode might
+        // be on the classpath don't throw an exception, just ignore the error.  (This bug has been
+        // fixed in Java 9)
+        // TODO: Issue an error if this annotation if from Java 9+ bytecode.
```
wmdietl: Second if should be is.

checker/jtreg/nullness/issue2173/ImporterManager.java.tmp
```diff
@@ -0,0 +1,27 @@
+import java.util.ArrayList;
```
wmdietl: A README file documenting why we need to store classfiles here would be good.

framework/src/main/java/org/checkerframework/framework/util/element/ElementAnnotationUtil.java
smillst: Good idea!  I did this.

smillst:
@wmdietl I'm going to go ahead and merge this, so #2173 is fixed in today's release.  Then I'm going to make another pull requests that issues a warning if an `UnexpectedAnnotationLocationException` is thrown and add a command line option to turn off the warning.

PR #2674, opened 2019-08-01
Annotated all `Object clone()` as PolySameLen

Kakarot-SSJ4:


Kakarot-SSJ4:
@kelloggm Fixed the unnecessary diff.

Kakarot-SSJ4:
@kelloggm @panacekcz

kelloggm:
Remind me why we need this PR?

I'm generally hesitant to add annotations to common methods like `clone()`, which should probably be `@PolyAll` anyway in the annotated JDK already.

Kakarot-SSJ4:
@kelloggm Because the checker does not infer an array1 = array2.clone() to have same length as array2

kelloggm:
Closing because we don't think the benefits of this change outweigh the additional annotation burden.

PR #2677, opened 2019-08-02
 Issue warning for invalid annotation locations.

smillst:
Also, adds a command line option `-AignoreInvalidAnnotationLocations` to suppress the warnings.

(Follow up to #2672 )

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
+++ b/framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
@@ -207,6 +207,11 @@ import org.checkerframework.javacutil.UserError;
     // java.lang.String)
     "requirePrefixInWarningSuppressions",
 
+    // Ignore annotations in bytecode that have invalid annotation locations.
+    // See https://github.com/typetools/checker-framework/issues/2173
+    // org.checkerframework.framework.type.ElementAnnotationApplier.apply
+    "ignoreInvalidAnnotationLocations",
+
     ///
     /// Partially-annotated libraries
     ///
--- a/framework/src/main/java/org/checkerframework/framework/type/ElementAnnotationApplier.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/ElementAnnotationApplier.java
@@ -88,10 +88,12 @@ public class ElementAnnotationApplier {
             // location, such as a type argument, that doesn't exist. Since Java 8 bytecode
             // might be on the classpath, catch this exception and ignore the type.
             // TODO: Issue an error if this annotation is from Java 9+ bytecode.
-            typeFactory.checker.report(
-                    Result.warning(
-                            "invalid.annotation.location", ElementUtils.getVerboseName(report)),
-                    element);
+            if (!typeFactory.checker.hasOption("ignoreInvalidAnnotationLocations")) {
+                typeFactory.checker.report(
+                        Result.warning(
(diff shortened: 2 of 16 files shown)


framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -207,6 +207,11 @@
     // java.lang.String)
     "requirePrefixInWarningSuppressions",
 
+    // Ignore annotations in bytecode that have invalid annotation locations.
```
wmdietl: A note in the manual should also be added.

framework/src/main/java/org/checkerframework/framework/type/ElementAnnotationApplier.java
```diff
@@ -132,7 +155,11 @@ private static void applyInternal(
      */
     public static void annotateSupers(
             List<AnnotatedDeclaredType> supertypes, TypeElement subtypeElement) {
-        SuperTypeApplier.annotateSupers(supertypes, subtypeElement);
+        try {
+            SuperTypeApplier.annotateSupers(supertypes, subtypeElement);
+        } catch (UnexpectedAnnotationLocationException e) {
+            throw new BugInCF(e.getMessage());
```
wmdietl: Why is this a bug in the CF? Isn't it the same kind of user error as on line 92 above?

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
smillst: Done.

framework/src/main/java/org/checkerframework/framework/type/ElementAnnotationApplier.java
```diff
@@ -138,7 +167,13 @@ private static void applyInternal(
      */
     public static void annotateSupers(
             List<AnnotatedDeclaredType> supertypes, TypeElement subtypeElement) {
-        SuperTypeApplier.annotateSupers(supertypes, subtypeElement);
+        try {
+            SuperTypeApplier.annotateSupers(supertypes, subtypeElement);
+        } catch (UnexpectedAnnotationLocationException e) {
+            if (!supertypes.isEmpty()) {
```
wmdietl: Could you add a comment why this guard is here? If `supertypes` were empty, how would the exception ever be raised?

framework/src/main/java/org/checkerframework/framework/type/ElementAnnotationApplier.java
```diff
@@ -188,7 +223,11 @@ public Void visitTypeVariable(AnnotatedTypeVariable type, AnnotatedTypeFactory f
             if (type.getAnnotations().isEmpty()
                     && type.getUpperBound().getAnnotations().isEmpty()
                     && tpelt.getEnclosingElement().getKind() != ElementKind.TYPE_PARAMETER) {
-                ElementAnnotationApplier.applyInternal(type, tpelt, factory);
+                try {
+                    ElementAnnotationApplier.applyInternal(type, tpelt, factory);
+                } catch (UnexpectedAnnotationLocationException e) {
+                    // Error was already reported by #apply.
```
wmdietl: This still isn't clear to me... Which invocation of `#apply`? Are we processing the same annotations twice and this is the second iteration and therefore we don't need to issue errors again?
Can you add a short comment when this earlier invocation of `#apply` happened?

PR #2679, opened 2019-08-05
Annotate Scanner, MessageDigest, MessageDigestSpi, and System classes, write() of java.io package and toCharArray() of String.java using Signedness Checker

gokaco:


mernst:
In ObjectOutputStream, method write(int) takes a byte, but its argument is not annotated as `@PolySigned`.  Should it be?  If not, why not?

mernst:
Can you also make a pull request against branch annotations-from-java-files of [typetools/jdk](https://github.com/typetools/jdk)?  That will avoid the work of copying your annotations into that version of the JDK.  Thanks.  Then I can review the two pull requests together.

gokaco:
> In ObjectOutputStream, method write(int) takes a byte, but its argument is not annotated as `@PolySigned`. Should it be? If not, why not?

Yes, it should be annotated. I have fixed that.

gokaco:
> Can you also make a pull request against branch annotations-from-java-files of [typetools/jdk](https://github.com/typetools/jdk)? That will avoid the work of copying your annotations into that version of the JDK. Thanks. Then I can review the two pull requests together.

I have issued a pull request related to it [#30](https://github.com/typetools/jdk/pull/30).

mernst:
This is blocked on https://github.com/typetools/jdk/pull/30 which needs to be complete before both can be reviewed together.  Please don't re-assign me to this pull request until that one is ready.

PR #2682, opened 2019-08-05
Remove the duplicate issues in changelog.txt .

xingweitian:


mernst:
The CI failures cannot be caused by this change, so this change is OK to merge.

PR #2685, opened 2019-08-05
Apply inferred annotation to intersection types

### LINKED ISSUE #2031: Error when using recursive generics and intersection bound
This is a slight variation of the previously fixed issue #1942   This time, the code has an extra ampersand bound (not sure of the exact name in Java) with a second recursive bound, and that seems to resurface the problem.  Tested in 2.5.2.  To reproduce, compile and run this code with the Nullness checker on:

```

import java.util.Map;

public class RecursiveTypes
{
    public interface A<EXPRESSION>
    {
    }

    public interface B<EXPRESSION>
    {
    }

    static abstract class OperatorSection<EXPRESSION extends A<EXPRESSION> & B<EXPRESSION>>
    {
        abstract EXPRESSION makeExpression(Map<String, EXPRESSION> expressions);
    }

    static class BinaryOperatorSection<EXPRESSION extends A<EXPRESSION> & B<EXPRESSION>> extends OperatorSection<EXPRESSION>
    {
        @Override
        EXPRESSION makeExpression(Map<String, EXPRESSION> expressions)
        {
            @Nullable EXPRESSION e = expressions.get("");
            if (e != null)
                return e;
            else
                throw new RuntimeException("");
        }
    }
}
```

The error is:

```
/RecursiveTypes.java:[27,24] [return.type.incompatible] incompatible types in return.
found   : EXPRESSION extends @UnknownInitialization @NonNull A<EXPRESSION extends @Initialized @NonNull A<EXPRESSION extends @Initialized @NonNull A<EXPRESSION> & @Initialized @NonNull B<EXPRESSION>> & @Initialized @NonNull B<EXPRESSION extends @Initialized @NonNull A<EXPRESSION extends @Initialized @NonNull A<EXPRESSION> & @Initialized @NonNull B<EXPRESSION>> & @Initialized @NonNull B<EXPRESSION extends @Initialized @NonNull A<EXPRESSION> & @Initialized @NonNull B<EXPRESSION>>>> & @UnknownInitialization @NonNull B<EXPRESSION extends @Initialized @NonNull A<EXPRESSION extends @Initialized @NonNull A<EXPRESSION extends @Initialized @NonNull A<EXPRESSION> & @Initialized @NonNull B<EXPRESSION>> & @Initialized @NonNull B<EXPRESSION extends @Initialized @NonNull A<EXPRESSION> & @Initialized @NonNull B<EXPRESSION>>> & @Initialized @NonNull B<EXPRESSION extends @Initialized @NonNull A<EXPRESSION> & @Initialized @NonNull B<EXPRESSION>>>
required: EXPRESSION extends @Initialized @NonNull A<EXPRESSION extends @Initialized @NonNull A<EXPRESSION> & @Initialized @NonNull B<EXPRESSION extends @Initialized @NonNull A<EXPRESSION> & @Initialized @NonNull B<EXPRESSION>>> & @Initialized @NonNull B<EXPRESSION extends @Initialized @NonNull A<EXPRESSION extends @Initialized @NonNull A<EXPRESSION> & @Initialized @NonNull B<EXPRESSION>> & @Initialized @NonNull B<EXPRESSION>>
```

If you remove the `& B<EXPRESSION>` bound where it twice occurs, the error goes away, so it seems to be that part of the bound that confuses the checker.

smillst:
Fixes #2031


framework/src/main/java/org/checkerframework/framework/type/DefaultInferredTypesApplier.java
```diff
@@ -90,6 +91,12 @@ private void apply(
                 }
             } else if ((omitSubtypingCheck || hierarchy.isSubtype(inferred, primary))) {
                 type.replaceAnnotation(inferred);
+                if (type.getKind() == TypeKind.INTERSECTION) {
```
mernst: It's not your fault, but the code is confusing when juxtaposed with the code just above it (lines 84-91).
I wonder if that can be refactored to emphasize the commonality, maybe in a separate pull request.
For example, that might clarify why the added code is not necessary when primary is null.

PR #2688, opened 2019-08-06
Fix clean task in checker-qual build

aagarwal1012:
On line 36, qualifiers and other java files are copied into the `src/main/java` folder, so the `clean` task should also be target the same.


wmdietl:
The alternative would be to consistently use `generated-src`.
@smillst which one do you prefer?

smillst:
src is better because it is the standard name.

PR #2689, opened 2019-08-06
Remove Deprecated methods from SignednessUtil

smillst:
I also had to update the uses of the Deprecated methods in SignednessUtil itself.
(A warning is not issued if a deprecated method is used in the same class in which it is declared:
https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.6.4.6)


smillst:
@mernst don't merge this until we are sure the next release with update the major/minor version.  (I can't "request changes" on my own pull request, so you may wish to do that so we don't merge this accidentally.)

PR #2691, opened 2019-08-06
Compile-time char constants are treated as literals

### LINKED ISSUE #1152: @StringVal and compile-time constant characters that are not literals
Methods `getByteArrayStringVal` and `getCharArrayStringVal` handle the case of manifest literals in the source code, but they do not handle the case of other compile-time constants.

Here is a test case:
```java

public class ByteArrayWithNonLiteralConstants {

    byte @StringVal("hello") [] greeting1 = {'h', 'e', 'l', 'l', 'o'};
    @IntVal('e') byte e = 'e';
    byte @StringVal("hello") [] greeting2 = {'h', e, 'l', 'l', 'o'};
  }

}
```

The output of the Constant Value Checker is:
```
$ch/bin/javac -g ByteArrayWithNonLiteralConstants.java -processor index
ByteArrayWithNonLiteralConstants.java:11: error: [assignment.type.incompatible] incompatible types in assignment.
    byte @StringVal("hello") [] greeting2 = {'h', e, 'l', 'l', 'o'};
                                            ^
  found   : @UnknownVal byte @ArrayLen(5) []
  required: @UnknownVal byte @StringVal("hello") []
1 error
```

Maxi17:
When initializing a `char[]`, check if the elements of the array are annotated with single-value `@IntVal`
Fixes #1152

I modified the test case from `byte[]` to `char[]` because of different character encodings.

### DIFF
--- a/framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
@@ -1049,7 +1049,7 @@ public class ValueAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
                 AnnotationMirror newQual;
                 Class<?> clazz = ValueCheckerUtils.getClassFromType(type.getUnderlyingType());
                 String stringVal = null;
-                if (clazz.equals(char[].class) || clazz.equals(byte[].class)) {
+                if (clazz.equals(char[].class)) {
                     stringVal = getCharArrayStringVal(initializers);
                 }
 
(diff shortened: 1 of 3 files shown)


framework/tests/value/ByteArrayWithNonLiteralConstants.java
```diff
@@ -0,0 +1,11 @@
+import org.checkerframework.common.value.qual.IntVal;
+import org.checkerframework.common.value.qual.StringVal;
+
+public class ByteArrayWithNonLiteralConstants {
```
kelloggm: Let's also change the name of this test to something to do with char arrays instead of byte arrays.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1187,6 +1187,13 @@ private String getCharArrayStringVal(List<? extends ExpressionTree> initializers
                 } else if (e.getKind() == Tree.Kind.CHAR_LITERAL) {
                     char charVal = (((Character) ((LiteralTree) e).getValue()));
                     stringVal.append(charVal);
+                } else if (e.getKind() == Tree.Kind.IDENTIFIER) {
```
kelloggm: Is there anything about this code that is specific to `IdentifierTree`? I don't think there is.

What if we made this an `else` block instead?

Further, I would also expect this code to work for `INT_LITERAL`s - so you might be able to replace the code above that handles them with this block, too.

Issue #152, opened 2015-07-03
Dependent Types: no way to differentiate a field that hides a superclass field


GoogleCodeExporter:
```
Attached is a simple program that will exercise this bug.  There are comments 
describing the bug in the code.

Short Description:  If a class and its base class both have a member "lock" 
and fields in both the class and base class are guarded by their respective 
locks then the checker does not report an error when the base class's guarded 
members are synchronized on the subclasses "lock" field even though the 
subclasses "lock" field is a different object from the base classes "lock" 
field.

To Reproduce:
1.  Download the attached file(LockBugExample.java)
2.  Run javac -processor checkers.lock.LockChecker LockBugExample.java

Expected Output:
Unguarded access errors for lines:  45,55,56,58

Actual Output:
A fully compiled program
```

Original issue reported on code.google.com by `Jonathan...@gmail.com` on 28 Aug 2012 at 10:43

Attachments:
- [LockBugExample.java](https://storage.googleapis.com/google-code-attachments/checker-framework/issue-152/comment-0/LockBugExample.java)

smillst:
@jthaine  Could you please add this test case to the repository?

smillst:
Should be fixed once #770 is fixed.

mernst:
Issue #770 is fixed, but this issue remains a problem.  The two tests for issue #152 remain broken (they are currently disabled).

smillst:
framework/tests/flowexpression/ThisSuper.java is another test case.

HuM4NoiD:
I wanted to ask, does it affect other tests, if instead to just using `this`, we can use `MySubClass.this`? That is while converting `ThisReference` to a string using `toString()`, we just return "this" and thus the checker cannot differentiate between `MyBaseClass` and `MySubClass extends MyBaseClass`. If we do this, then the errors are reported in the required places for the class `$CHECKERFRAMEWORK/checker/tests/lock` Let me know it doing this is a hack or it can work.

mernst:
I cannot follow your question.  Can you create a test case that shows the exact code you have in mind (it sounds like there are two variants), and also show the checker's output on each of the variants?  Or if you are talking about a change to the checker, show the exact code changes in the Checker Framework.

HuM4NoiD:
@mernst I have created Pull request in my own fork [here](https://github.com/HuM4NoiD/checker-framework/pull/3), I have explained the fix

PR #2693, opened 2019-08-07
Non-static fields can't be referenced from a static inner class or enum

### LINKED ISSUE #2587: Dependent type annotations: static inner classes should not refer to instance fields of the outer class
```
public abstract class Issue2587 {
    public enum EnumType {
        // :: error: (expression.unparsable.type.invalid) :: error: (assignment.type.incompatible)
        @KeyFor("myMap") MY_KEY,
        // :: error: (assignment.type.incompatible)
        @KeyFor("enumMap") ENUM_KEY;
        private static final Map<String, Integer> enumMap = new HashMap<>();

        void method() {
            @KeyFor("enumMap") EnumType t = ENUM_KEY;
            int x = enumMap.get(ENUM_KEY);
        }
    }

    public static class Inner {
        // :: error: (expression.unparsable.type.invalid)
        @KeyFor("myMap") String MY_KEY = "";
    }

    private final Map<String, Integer> myMap = new HashMap<>();
}
```
Outputs:
```
Issue2564.java:8: error: [assignment.type.incompatible] incompatible types in assignment.
        @KeyFor("myMap") MY_KEY,
                         ^
  found   : EnumType
  required: @KeyFor("this.myMap") EnumType
Issue2564.java:9: error: [assignment.type.incompatible] incompatible types in assignment.
        @KeyFor("enumMap") ENUM_KEY;
                           ^
  found   : EnumType
  required: @KeyFor("Issue2564.EnumType.enumMap") EnumType
Issue2564.java:13: error: [assignment.type.incompatible] incompatible types in assignment.
            @KeyFor("enumMap") EnumType t = ENUM_KEY;
                                            ^
  found   : @KeyFor("enumMap") EnumType
  required: @KeyFor("Issue2564.EnumType.enumMap") EnumType
Issue2564.java:14: error: [unboxing.of.nullable] unboxing a possibly-null reference enumMap.get(ENUM_KEY)
            int x = enumMap.get(ENUM_KEY);
                               ^
Issue2564.java:20: error: [assignment.type.incompatible] incompatible types in assignment.
        @KeyFor("myMap") String MY_KEY = "";
                                         ^
  found   : String
  required: @KeyFor("this.myMap") String
5 errors
```

Maxi17:
Fixes #2587 

If the annotated element is inside a static class or an enum and the field that the annotation refers to is non-static, then throw an exception.

I will try to make the code cleaner.

### DIFF
--- a/checker/tests/nullness/Issue2587.java
+++ b/checker/tests/nullness/Issue2587.java
@@ -1,0 +1,35 @@
+import java.util.HashMap;
+import java.util.Map;
+import org.checkerframework.checker.nullness.qual.KeyFor;
+
+public abstract class Issue2587 {
+    public enum EnumType {
+        // :: error: (expression.unparsable.type.invalid) :: error: (assignment.type.incompatible)
+        @KeyFor("myMap") MY_KEY,
+        // :: error: (assignment.type.incompatible)
+        @KeyFor("enumMap") ENUM_KEY;
+        private static final Map<String, Integer> enumMap = new HashMap<>();
+
+        void method() {
+            @KeyFor("enumMap") EnumType t = ENUM_KEY;
+            int x = enumMap.get(ENUM_KEY);
+        }
+    }
+
+    public static class Inner {
+        // :: error: (expression.unparsable.type.invalid) :: error: (assignment.type.incompatible)
+        @KeyFor("myMap") String MY_KEY = "";
+
+        public static class Inner2 {
+            // :: error: (expression.unparsable.type.invalid) :: error:
+            // (assignment.type.incompatible)
+            @KeyFor("myMap") String MY_KEY2 = "";
+            // :: error: (assignment.type.incompatible)
+            @KeyFor("innerMap") String MY_KEY3 = "";
+        }
+
+        private static final Map<String, Integer> innerMap = new HashMap<>();
+    }
+
+    private final Map<String, Integer> myMap = new HashMap<>();
+}
--- a/checker/tests/nullness/Issue2587.java
+++ b/checker/tests/nullness/Issue2587.java
@@ -2,11 +2,12 @@ import java.util.HashMap;
 import java.util.Map;
 import org.checkerframework.checker.nullness.qual.KeyFor;
 
+@SuppressWarnings("assignment.type.incompatible")
 public abstract class Issue2587 {
     public enum EnumType {
-        // :: error: (expression.unparsable.type.invalid) :: error: (assignment.type.incompatible)
+        // :: error: (expression.unparsable.type.invalid)
         @KeyFor("myMap") MY_KEY,
-        // :: error: (assignment.type.incompatible)
+
         @KeyFor("enumMap") ENUM_KEY;
         private static final Map<String, Integer> enumMap = new HashMap<>();
 
@@ -17,14 +18,13 @@ public abstract class Issue2587 {
     }
 
     public static class Inner {
-        // :: error: (expression.unparsable.type.invalid) :: error: (assignment.type.incompatible)
+        // :: error: (expression.unparsable.type.invalid)
         @KeyFor("myMap") String MY_KEY = "";
 
         public static class Inner2 {
-            // :: error: (expression.unparsable.type.invalid) :: error:
-            // (assignment.type.incompatible)
+            // :: error: (expression.unparsable.type.invalid)
             @KeyFor("myMap") String MY_KEY2 = "";
-            // :: error: (assignment.type.incompatible)
+
             @KeyFor("innerMap") String MY_KEY3 = "";
         }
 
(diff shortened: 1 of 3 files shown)


framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -307,6 +307,19 @@ public Receiver visit(NameExpr expr, FlowExpressionContext context) {
             }
 
             if (fieldElem != null && fieldElem.getKind() == ElementKind.FIELD) {
+                if (!originalReceiver
+                        && TreeUtils.enclosingClass(path) != null
+                        && (TreeUtils.enclosingClass(path)
+                                        .getModifiers()
+                                        .getFlags()
+                                        .contains(Modifier.STATIC)
+                                || TreeUtils.enclosingClass(path).getKind().equals(Tree.Kind.ENUM))
```
smillst: Enums, such as Tree.Kind, should be compared using `==` rather than `.equals`.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -307,6 +307,19 @@ public Receiver visit(NameExpr expr, FlowExpressionContext context) {
             }
 
             if (fieldElem != null && fieldElem.getKind() == ElementKind.FIELD) {
+                if (!originalReceiver
+                        && TreeUtils.enclosingClass(path) != null
```
smillst: `path` is the TreePath to the expression (in source code) whose type has an annotation with the FlowExpression.  So, this code won't work correctly.  I added a failing test case.

framework/tests/flowexpression/InnerClasses.java
```diff
@@ -0,0 +1,38 @@
...
+    static class InnerClass2 {
+        public String outerInstanceField = "";
+
+        @FlowExp("outerInstanceField") Object o = null;
+    }
+
+    class TestUses {
+        void method(InnerClass innerClass, InnerClass2 innerClass2) {
+            // :: error: (expression.unparsable.type.invalid) :: error:
+            // (assignment.type.incompatible)
```
Maxi17: @smillst I'm not sure why an `expression.unparsable.type.invalid` error is expected here, since the inner class `TestUses` and the method `method` are non-static. Can you please explain why you think it should be issued?
I agree with the `assignment.type.incompatible` error, but I think this is issue #152, `framework/tests/flowexpression/ThisSuper.java` test case. In the last part of the test case, something similar is tested and it fails.

smillst: The error should be on the explicitly written annotation `@FlowExp("innerClass.outerInstanceField")`.  
The type of `innerClass` is `InnerClass`. `outerInstanceField` can't be referenced from an instance of `InnerClass`.  Also, if I had written innerClass.outerInstanceField, javac would issue an error.

checker/tests/nullness/Issue2587.java
```diff
@@ -0,0 +1,35 @@
+import java.util.HashMap;
+import java.util.Map;
+import org.checkerframework.checker.nullness.qual.KeyFor;
+
+public abstract class Issue2587 {
+    public enum EnumType {
+        // :: error: (expression.unparsable.type.invalid) :: error: (assignment.type.incompatible)
```
Maxi17: I think that a good idea would be to `@SuppressWarnings("assignment.type.incompatible")` for the entire class. These errors only distract from the actual issue.

smillst: I agree, just make sure you document why you added the `@SuppressWarnings`.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -307,7 +307,21 @@ public Receiver visit(NameExpr expr, FlowExpressionContext context) {
             }
 
             if (fieldElem != null && fieldElem.getKind() == ElementKind.FIELD) {
-                return getReceiverField(s, context, originalReceiver, fieldElem);
+                FieldAccess fieldAccess =
+                        (FieldAccess) getReceiverField(s, context, originalReceiver, fieldElem);
+                String className = fieldAccess.getReceiver().getType().toString();
+                Element scopeClassElement =
```
smillst: Use `org.checkerframework.javacutil.TypesUtils#getTypeElement` to go from a `TypeMirror` to an `TypeElement`.

framework/tests/flowexpression/InnerClasses.java
```diff
@@ -0,0 +1,38 @@
...
+
+    static class InnerClass2 {
+        public String outerInstanceField = "";
+
+        @FlowExp("outerInstanceField") Object o = null;
+    }
+
+    class TestUses {
+        void method(InnerClass innerClass, InnerClass2 innerClass2) {
+            @SuppressWarnings("assignment.type.incompatible") // Issue 152
```
smillst: Because the LHS annotation isn't parsable, this should issue an assignment.type.incompatible error.

PR #2694, opened 2019-08-07
Array creation and primitive type class literals are parsable

Maxi17:
New parsable types:

1. Array creation expressions (`new int[2]`, `new String[] {"a", "b"}`, `new java.lang.String[2]`)
2. Class literals with primitive or void type (`int.class`, `double.class`, `void.class`).

I combined those 2 because I used class literals with primitive type to parse array creation expressions with primitive type.

I will add Javadoc soon.

### DIFF
--- a/docs/manual/advanced-features.tex
+++ b/docs/manual/advanced-features.tex
@@ -1206,8 +1206,7 @@ The set of permitted expressions is a subset of all Java expressions:
 \item
   an array creation. For example: \<new int[10]>, \<new String[] {"a", "b"}>.
 
-\item literals: string, integer, char, long, float, double, null literals.
-   Class literals with class, interface, primitive or void type.
+\item literals: string, integer, char, long, float, double, null, class literals.
 
 \item a method invocation on any expression.
   This even works for overloaded methods and methods with type parameters.
@@ -1250,7 +1249,6 @@ The following Java expressions may not currently be written:
 % The Checker Framework is best at reasoning about Java expressions that
 % are variable references, but these expressions are not.
 \begin{itemize}
-\item Class literals with with array type (String[].class).
 \item String concatenation expressions.
 \item Mathematical operators (plus, minus, division, ...).
 \item Comparisons (equality, less than, etc.).
--- a/docs/manual/advanced-features.tex
+++ b/docs/manual/advanced-features.tex
@@ -1203,8 +1203,11 @@ The set of permitted expressions is a subset of all Java expressions:
 \item
   an array access.  For example:  \<this.myArray[i]>, \<vals[\#1]>.
 
+\item
+  an array creation. For example: \<new int[10]>, \<new String[] {"a", "b"}>.
+
 \item literals: string, integer, char, long, float, double, null literals.
-  Class literals with class or interface type.
+   Class literals with class, interface, primitive or void type.
 
 \item a method invocation on any expression.
   This even works for overloaded methods and methods with type parameters.
@@ -1247,7 +1250,7 @@ The following Java expressions may not currently be written:
 % The Checker Framework is best at reasoning about Java expressions that
 % are variable references, but these expressions are not.
 \begin{itemize}
-\item Class literals with primitive type (int.class) or with array type (String[].class).
+\item Class literals with with array type (String[].class).
 \item String concatenation expressions.
 \item Mathematical operators (plus, minus, division, ...).
 \item Comparisons (equality, less than, etc.).
--- a/framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
+++ b/framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
@@ -571,6 +571,9 @@ public class FlowExpressionParseUtil {
                 }
             } else if (type.isVoidType()) {
                 return types.getNoType(TypeKind.VOID);
+            } else if (type.isArrayType()) {
+                return types.getArrayType(
+                        convertTypeToTypeMirror(type.asArrayType().getComponentType(), context));
             }
             return null;
         }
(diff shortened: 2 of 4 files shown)


framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -492,12 +495,76 @@ public Receiver visit(FieldAccessExpr expr, FlowExpressionContext context) {
...
+
+        @Override
+        public Receiver visit(ArrayCreationExpr expr, FlowExpressionContext context) {
+
+            // Parse the type as a class literal
+            Receiver receiver =
+                    visit(
+                            StaticJavaParser.parseExpression(expr.getElementType() + ".class")
+                                    .asClassExpr(),
+                            context);
```
Maxi17: This is bad, but I couldn't think of something better.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -492,12 +495,76 @@ public Receiver visit(FieldAccessExpr expr, FlowExpressionContext context) {
...
-                        constructParserException(
-                                expr.toString(), "is an unparsable class literal"));
+            if (expr.getType().isClassOrInterfaceType()) {
+                return StaticJavaParser.parseExpression(expr.getTypeAsString())
+                        .accept(this, context);
+            } else if (expr.getType().isPrimitiveType()) {
+                switch (expr.getType().asPrimitiveType().getType()) {
+                    case BOOLEAN:
+                        return new ClassName(
+                                TypesUtils.typeFromClass(
```
smillst: Use `types.getPrimitiveType(kind)`  instead of typeFromClass.

Maxi17:
> What's the relationship between this pull request and issue #1635?

I thought that the first step to fix #1635 was to make array creation expressions parsable. But I'm close to figuring out a fix for it without using array creation expression parsing. I only needed the `ArrayCreation` receiver, which already existed. (if you want to see my progress, link: [GitHub](https://github.com/Maxi17/checker-framework/tree/issue1635))

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -501,52 +501,38 @@ public Receiver visit(ClassExpr expr, FlowExpressionContext context) {
...
         public Receiver visit(ArrayCreationExpr expr, FlowExpressionContext context) {
 
             // Parse the type as a class literal
             Receiver receiver =
-                    visit(
-                            StaticJavaParser.parseExpression(expr.getElementType() + ".class")
-                                    .asClassExpr(),
-                            context);
+                    StaticJavaParser.parseExpression(expr.getElementType() + ".class")
+                            .accept(this, context);
```
Maxi17: What I tried to do here is call `visit(ClassExpr, FlowExpressionContext)` directly, because the `String` would be parsed s a `ClassExpr` anyways. If I don't append `.class` to the end, if `int[]` is parsed, then it would choose the `NameExpr` visitor for `int` (which is `expr.getElementType().asString()`, which throws an exception. Is there a way to improve this?

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -492,12 +496,82 @@ public Receiver visit(FieldAccessExpr expr, FlowExpressionContext context) {
...
+
+            List<Receiver> initializers = new ArrayList<>();
+            if (expr.getInitializer().isPresent()) {
+                for (Expression initializer : expr.getInitializer().get().getValues()) {
+                    initializers.add(initializer.accept(this, context));
+                }
+            }
+            TypeMirror arrayType = convertTypeToTypeMirror(expr.getElementType(), context);
+            if (arrayType == null) {
+                // TODO: issue error
```
smillst: Please add an exception here.

framework/tests/flowexpression/ArrayCreationParsing.java
```diff
@@ -0,0 +1,19 @@
+package flowexpression;
+
+import testlib.flowexpression.qual.FlowExp;
+
+public class ArrayCreationParsing {
+    @FlowExp("new int[2]") Object value1;
```
smillst: Please add tests with initializers and more complex dimension expressions.

framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java
```diff
@@ -525,9 +525,11 @@ public Receiver visit(ArrayCreationExpr expr, FlowExpressionContext context) {
             }
             TypeMirror arrayType = convertTypeToTypeMirror(expr.getElementType(), context);
             if (arrayType == null) {
-                // TODO: issue error
+                throw new ParseRuntimeException(
+                        constructParserException(
+                                expr.getElementType().asString(), "type not parsable"));
             }
-            for (Receiver ignored : dimensions) {
+            for (int i = 0; i < dimensions.size(); i++) {
```
Maxi17: The test failed initially because `ignored` was not used.

docs/manual/advanced-features.tex
```diff
@@ -1247,7 +1250,7 @@
 % The Checker Framework is best at reasoning about Java expressions that
 % are variable references, but these expressions are not.
 \begin{itemize}
-\item Class literals with primitive type (int.class) or with array type (String[].class).
+\item Class literals with with array type (String[].class).
```
smillst: This pull request added support for `String[].class` right?

Maxi17: It does now.

PR #2696, opened 2019-08-08
Add a way for checkers to bias type arg inference toward defaults (Fixes #2686)

### LINKED ISSUE #2686: Type inference is too eager to use Constant Value Checker types
Consider the following code.

```
import org.checkerframework.common.value.qual.UnknownVal;

public class MyTree<Value> {

    public static <V> MyTree<V> newTree(V value) {
        throw new Error("body doesn't matter");
    }

    public MyTree<Value> put(String newKey, Value newValue) {
        throw new Error("body doesn't matter");
    }

    public void client1() {
        // type-checking error
        MyTree<String> root2 = MyTree.newTree("constructorarg").put("key1", "value1");
    }

    public void client2() {
        // OK
        MyTree<String> root2 = MyTree.newTree((@UnknownVal String) "constructorarg").put("key1", "value1");
    }

    public void client3() {
        // OK
        MyTree<String> root2 = MyTree.<String>newTree((@UnknownVal String) "constructorarg").put("key1", "value1");
    }

}
```

You can typecheck it for the Constant Value checker by running

	${CHECKERFRAMEWORK}/checker/bin/javac -processor org.checkerframework.common.value.ValueChecker MyTree.java

Type-checking of client1 fails with the messages:

```
MyTree.java:15: error: [assignment.type.incompatible] incompatible types in assignment.
        MyTree<String> root2 = MyTree.newTree("constructorarg").put("key1", "value1");
                                                                   ^
  found   : @UnknownVal MyTree<@StringVal("constructorarg") String>
  required: @UnknownVal MyTree<@UnknownVal String>
MyTree.java:15: error: [argument.type.incompatible] incompatible types in argument.
        MyTree<String> root2 = MyTree.newTree("constructorarg").put("key1", "value1");
                                                                            ^
  found   : @StringVal("value1") String
  required: @StringVal("constructorarg") String
```

The problem is that in the call `MyTree.newTree("constructorarg")`, the Checker Framework infers a constant-value type for "constructorarg", and then instantiates the constructor with that constant-value type.

Methods `client2` and `client3` show how to solve the problem, but both of them require code changes to the user code.  Is there a way to solve the problem without code changes to the user code?  It would be acceptable to not use constant-value types anywhere that type inference occurs.

smillst:
Currently only the Value Checker uses this.
Fixes #2686.

### DIFF
--- a/framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
@@ -2260,6 +2260,10 @@ public class ValueAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
         return getMinLenValue(lengthAnno);
     }
 
+    /**
+     * Returns the annotation type mirror for the type of {@code expressionTree} with default
+     * annotations applied.
+     */
     @Override
     public AnnotatedTypeMirror getDummyAssignedTo(ExpressionTree expressionTree) {
         TypeMirror type = TreeUtils.typeOf(expressionTree);
--- a/framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
@@ -2259,4 +2259,15 @@ public class ValueAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
 
         return getMinLenValue(lengthAnno);
     }
+
+    @Override
+    public AnnotatedTypeMirror getDummyAssignedTo(ExpressionTree expressionTree) {
+        TypeMirror type = TreeUtils.typeOf(expressionTree);
+        if (type.getKind() != TypeKind.VOID) {
+            AnnotatedTypeMirror atm = type(expressionTree);
+            addDefaultAnnotations(atm);
+            return atm;
+        }
+        return null;
+    }
 }
--- a/framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
@@ -1182,14 +1182,13 @@ public class ValueAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
             boolean allLiterals = true;
             StringBuilder stringVal = new StringBuilder();
             for (ExpressionTree e : initializers) {
-                if (e.getKind() == Tree.Kind.INT_LITERAL) {
-                    char charVal = (char) ((Integer) ((LiteralTree) e).getValue()).intValue();
-                    stringVal.append(charVal);
-                } else if (e.getKind() == Tree.Kind.CHAR_LITERAL) {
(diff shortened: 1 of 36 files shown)


framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -2258,4 +2258,15 @@ public int getMinLenFromString(String sequenceExpression, Tree tree, TreePath cu
 
         return getMinLenValue(lengthAnno);
     }
+
+    @Override
+    public AnnotatedTypeMirror getDummyAssignedTo(ExpressionTree expressionTree) {
```
mernst: A comment would be helpful.  This returns the default annotations.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -3993,4 +3993,21 @@ public static String negateConstant(String constantExpression) {
             return "-" + constantExpression;
         }
     }
+
+    /**
+     * Creates an annotated type mirror that type argument inference should pretend the {@code
+     * expression} is assigned to. This implementation returns null, but subclasses may override
+     * this method to return a type.
+     *
+     * <p>This biases the inference algorithm toward default annotations for type variables in
```
mernst: I think that "this" refers to "override this method" (but not to this implementation).  One choice for overriding is to return default annotations, which biases as this sentence does, but other choices are possible.

The documentation doesn't say what null means or what the effect of returning null is.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -3993,4 +3993,21 @@ public static String negateConstant(String constantExpression) {
             return "-" + constantExpression;
         }
     }
+
+    /**
+     * Creates an annotated type mirror that type argument inference should pretend the {@code
+     * expression} is assigned to. This implementation returns null, but subclasses may override
+     * this method to return a type.
+     *
+     * <p>This biases the inference algorithm toward the annotations for type variables in invariant
```
mernst: I'm still not sure whether "This" refers to "This implementation returns null" or "subclasses may override this method to return a type."  Can you clarify?  I would put "This implementation returns null" and "subclasses may override this method to return a type" in different paragraphs, with "This biases..." in one of those paragraphs.  Thanks!

smillst: I wasn't done addressing your review.  (That's why I didn't reassign you.)  I was just pushing changes to get them to the remote branch.

mernst: OK.  Sorry about that.

mernst: (I just had a moment between other tasks...)

smillst:
@mernst, before we merge this code, could you test that it solves the issue in original code?

mernst:
> @mernst, before we merge this code, could you test that it solves the issue in original code?

I'm inclined to merge, because there is a test case and so this resolves the issue.  If there are still problems in the original code, then they probably deserve a different issue number.

PR #2706, opened 2019-08-11
Revamp Docker setup

mernst:


mernst:
Merging without CI tests, which don't read these files.

PR #2710, opened 2019-08-11
Expand loop to avoid cross-talk between expressions

mernst:


mernst:
CI has succeeded, but shows up as "in progress" here.  So I am merging.

PR #2716, opened 2019-08-14
UpperBoundChecker checks varargs methods correctly

### LINKED ISSUE #1635: Reference to varargs is wrong at call sites

Maxi17:
In a varargs method, the varargs arguments are now placed in an `ArrayCreation` receiver as initializers, such that `UpperBoundChecker` can process them correctly.

Fixes #1635 
(More details in this comment: [link](https://github.com/typetools/checker-framework/issues/1635#issuecomment-344050065))

### DIFF
--- a/checker/tests/index/IndexForVarargs.java
+++ b/checker/tests/index/IndexForVarargs.java
@@ -1,0 +1,28 @@
+import org.checkerframework.checker.index.qual.IndexFor;
+
+public class IndexForVarargs {
+    String get(@IndexFor("#2") int i, String... varargs) {
+        return varargs[i];
+    }
+
+    void method(@IndexFor("#2") int i, String[]... varargs) {}
+
+    void m() {
+        // :: error: (argument.type.incompatible)
+        get(1);
+
+        get(1, "a", "b");
+
+        // :: error: (argument.type.incompatible)
+        get(2, "abc");
+
+        String[] stringArg1 = new String[] {"a", "b"};
+        String[] stringArg2 = new String[] {"c", "d", "e"};
+        String[] stringArg3 = new String[] {"a", "b", "c"};
+
+        method(1, stringArg1, stringArg2);
+
+        // :: error: (argument.type.incompatible)
+        method(2, stringArg3);
+    }
+}
--- a/dataflow/src/main/java/org/checkerframework/dataflow/analysis/FlowExpressions.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/analysis/FlowExpressions.java
@@ -1193,19 +1193,19 @@ public class FlowExpressions {
             sb.append("new " + type);
             if (!dimensions.isEmpty()) {
                 boolean needComma = false;
-                sb.append(" (");
                 for (Receiver dim : dimensions) {
+                    sb.append("[");
                     if (needComma) {
                         sb.append(", ");
                     }
-                    sb.append(dim);
+                    sb.append(dim == null ? "" : dim);
                     needComma = true;
+                    sb.append("]");
                 }
-                sb.append(")");
             }
             if (!initializers.isEmpty()) {
                 boolean needComma = false;
-                sb.append(" = {");
+                sb.append(" {");
                 for (Receiver init : initializers) {
                     if (needComma) {
                         sb.append(", ");
--- a/framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
+++ b/framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
@@ -28,6 +28,7 @@ import java.util.StringJoiner;
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.ElementKind;
+import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
@@ -195,9 +196,37 @@ public class DependentTypesHelper {
             receiver = FlowExpressions.internalReprOf(factory, receiverTree);
         }
 
-        List<FlowExpressions.Receiver> argReceivers = new ArrayList<>(args.size());
-        for (ExpressionTree argTree : args) {
-            argReceivers.add(FlowExpressions.internalReprOf(factory, argTree));
+        List<FlowExpressions.Receiver> argReceivers = new ArrayList<>();
+        boolean varargs = false;
+        if (tree instanceof MethodInvocationTree) {
+            ExecutableElement methodCalled =
+                    ((ExecutableElement)
+                            TreeUtils.elementFromUse(
+                                    ((MethodInvocationTree) tree).getMethodSelect()));
+            if (methodCalled.isVarArgs()) {
+                varargs = true;
+                for (int i = 0; i < methodCalled.getParameters().size() - 1; i++) {
+                    argReceivers.add(FlowExpressions.internalReprOf(factory, args.get(i)));
+                }
+                List<Receiver> initializers = new ArrayList<>();
+                for (int i = methodCalled.getParameters().size() - 1; i < args.size(); i++) {
+                    initializers.add(FlowExpressions.internalReprOf(factory, args.get(i)));
+                }
+                TypeMirror tm =
+                        ElementUtils.getType(
+                                methodCalled
+                                        .getParameters()
+                                        .get(methodCalled.getParameters().size() - 1));
+                argReceivers.add(
+                        new FlowExpressions.ArrayCreation(
+                                tm, Collections.emptyList(), initializers));
+            }
+        }
+
+        if (!varargs) {
+            for (ExpressionTree argTree : args) {
+                argReceivers.add(FlowExpressions.internalReprOf(factory, argTree));
+            }
         }
 
         TreePath currentPath = factory.getPath(tree);
--- a/framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
@@ -2241,6 +2241,11 @@ public class ValueAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
             if (sequenceLiteralValue instanceof String) {
                 return ((String) sequenceLiteralValue).length();
             }
+        } else if (expressionObj instanceof FlowExpressions.ArrayCreation) {
+            FlowExpressions.ArrayCreation arrayCreation =
+                    (FlowExpressions.ArrayCreation) expressionObj;
+            // This is only expected to support array creations in varargs methods
+            return arrayCreation.getInitializers().size();
         }
 
         lengthAnno = getAnnotationFromReceiver(expressionObj, tree, ArrayLenRange.class);


dataflow/src/main/java/org/checkerframework/dataflow/analysis/FlowExpressions.java
```diff
@@ -1193,19 +1193,19 @@ public String toString() {
...
             if (!dimensions.isEmpty()) {
                 boolean needComma = false;
-                sb.append(" (");
                 for (Receiver dim : dimensions) {
+                    sb.append("[");
                     if (needComma) {
                         sb.append(", ");
                     }
-                    sb.append(dim);
+                    sb.append(dim == null ? "" : dim);
```
Maxi17: This change is important because `new String[null] {"a", "b"}` is not valid

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -195,9 +196,37 @@ private void viewpointAdaptExecutable(
...
+                }
+                List<Receiver> initializers = new ArrayList<>();
+                for (int i = methodCalled.getParameters().size() - 1; i < args.size(); i++) {
+                    initializers.add(FlowExpressions.internalReprOf(factory, args.get(i)));
+                }
+                TypeMirror tm =
+                        ElementUtils.getType(
+                                methodCalled
+                                        .getParameters()
+                                        .get(methodCalled.getParameters().size() - 1));
```
kelloggm: Can you split this into one or more local variables so that the line isn't broken up like this? As-is it's completely unreadable.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -195,9 +196,34 @@ private void viewpointAdaptExecutable(
             receiver = FlowExpressions.internalReprOf(factory, receiverTree);
         }
 
-        List<FlowExpressions.Receiver> argReceivers = new ArrayList<>(args.size());
-        for (ExpressionTree argTree : args) {
-            argReceivers.add(FlowExpressions.internalReprOf(factory, argTree));
+        List<FlowExpressions.Receiver> argReceivers = new ArrayList<>();
+        boolean varargs = false;
+        if (tree instanceof MethodInvocationTree) {
```
smillst: Don't use `instanceof`, check the Tree.Kind: https://checkerframework.org/manual/#creating-using-the-apis.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -195,9 +196,34 @@ private void viewpointAdaptExecutable(
...
+            ExpressionTree methodElement = ((MethodInvocationTree) tree).getMethodSelect();
```
smillst: This variable is called methodElement, but it's a tree.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -195,9 +196,34 @@ private void viewpointAdaptExecutable(
...
+            ExecutableElement methodCalled =
+                    (ExecutableElement) TreeUtils.elementFromUse(methodElement);
```
smillst: You should call `elementFormUse` on the method invocation tree, not the method select.  `elementFormUse((MethodInvocationTree) tree)`

checker/tests/index/IndexForVarargs.java
```diff
@@ -0,0 +1,28 @@
...
+public class IndexForVarargs {
+    String get(@IndexFor("#2") int i, String... varargs) {
+        return varargs[i];
+    }
+
+    void method(@IndexFor("#2") int i, String[]... varargs) {}
+
+    void m() {
+        // :: error: (argument.type.incompatible)
+        get(1);
```
smillst: Add a test that calls `get` passing an array of strings.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -195,9 +196,34 @@ private void viewpointAdaptExecutable(
...
+            if (methodCalled.isVarArgs()) {
```
smillst: From the `isVarArgs` Javadoc 
> Returns true if this method or constructor accepts a variable number of arguments 

But that doesn't mean that a particular invocation passes variable arguments.  It could pass an array.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -214,9 +212,14 @@ private void viewpointAdaptExecutable(
                 Element varargsElement =
                         methodCalled.getParameters().get(methodCalled.getParameters().size() - 1);
                 TypeMirror tm = ElementUtils.getType(varargsElement);
-                argReceivers.add(
-                        new FlowExpressions.ArrayCreation(
-                                tm, Collections.emptyList(), initializers));
+                if (initializers.size() == 1 && initializers.get(0).getType().equals(tm)) {
```
Maxi17: This is a bad way to check if the argument passed was an array. It works considering that an exception would have been thrown if it was an invalid type.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -195,9 +196,38 @@ private void viewpointAdaptExecutable(
...
+        List<Receiver> argReceivers = new ArrayList<>();
+        boolean varargs = false;
+        if (tree.getKind() == Kind.METHOD_INVOCATION) {
+            ExecutableElement methodCalled = TreeUtils.elementFromUse((MethodInvocationTree) tree);
+            if (methodCalled != null && methodCalled.isVarArgs()) {
+                varargs = true;
+                for (int i = 0; i < methodCalled.getParameters().size() - 1; i++) {
+                    argReceivers.add(FlowExpressions.internalReprOf(factory, args.get(i)));
+                }
+                List<Receiver> initializers = new ArrayList<>();
```
smillst: `initializers` is a confusing name.  Maybe varargArgs?

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -195,9 +196,38 @@ private void viewpointAdaptExecutable(
...
+                for (int i = methodCalled.getParameters().size() - 1; i < args.size(); i++) {
+                    initializers.add(FlowExpressions.internalReprOf(factory, args.get(i)));
+                }
+                Element varargsElement =
+                        methodCalled.getParameters().get(methodCalled.getParameters().size() - 1);
+                TypeMirror tm = ElementUtils.getType(varargsElement);
+                if (initializers.size() == 1
+                        && initializers.get(0).getType().toString().equals(tm.toString())) {
```
smillst: See org.checkerframework.framework.util.AnnotatedTypes#expandVarArgs:
```
        if (parameters.size() == args.size()) {
            // Check if one sent an element or an array
            AnnotatedTypeMirror lastArg = atypeFactory.getAnnotatedType(args.get(args.size() - 1));
            if (lastArg.getKind() == TypeKind.ARRAY
                    && getArrayDepth(varargs) == getArrayDepth((AnnotatedArrayType) lastArg)) {
                return parameters;
            }
        }
```

You should do something like that here.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/FlowExpressions.java
```diff
@@ -1207,15 +1207,15 @@ public String toString() {
             sb.append("new " + type);
             if (!dimensions.isEmpty()) {
                 boolean needComma = false;
+                sb.append(" [");
```
Maxi17: @smillst Right now, the `ArrayCreation` receivers are printed like this: `new String[1,2,3]`, which is not correct. It should be `new String[1][2][3]`. Therefore, moving the `[` and `]` inside the loop and removing the `needComma` check should do the job. What do you think of this?

smillst: Sounds good, please make that change.

Maxi17: Did it.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -195,9 +198,48 @@ private void viewpointAdaptExecutable(
...
+                    AnnotatedTypeMirror lastArg =
+                            factory.getAnnotatedType(args.get(args.size() - 1));
+                    List<AnnotatedTypeMirror> argTypes = typeFromUse.getParameterTypes();
+                    AnnotatedArrayType varargsArg =
+                            (AnnotatedArrayType) argTypes.get(argTypes.size() - 1);
+                    if (lastArg.getKind() == TypeKind.ARRAY
+                            && AnnotatedTypes.getArrayDepth(varargsArg)
+                                    == AnnotatedTypes.getArrayDepth((AnnotatedArrayType) lastArg)) {
+                        argReceivers.add(varargArgs.get(0));
+                        passedArray = true;
```
panacekcz: It seems that the case of `passedArray == true` has the same result as the case `isVarargs == false`.
I'd recommend moving these two ifs above to the point where `isVarargs = true` is, and perhaps moving all these checks to a helper method `isVarArgsInvocation(methodCalled, typeFromUse, args)` or similar. That would separate the two things this code is doing - (1) checking if it is a varargs invocation and (2) filling the `argReceivers` list - and make it easier to understand.

Maxi17:
The build failed because `DependentTypesHelper.standardizeNewClassTree` is not documented. I haven't changed that method, so I don't know why the error is issued.

mernst:
> The build failed because `DependentTypesHelper.standardizeNewClassTree` is not documented. I haven't changed that method, so I don't know why the error is issued.

You changed the line before that method, and the rule uses a fuzz of one line.
This is necessary to catch some needed edits.  Other times, it just makes us add some documentation that should have been there all along.  This is one of the latter cases.

PR #2732, opened 2019-08-19
Support Java 9 compiler for use with JVM 8

wmdietl:
Changes the Checker Framework to require the Java 9 compiler, but still run on a Java 8 JVM.

Contributors: @wmdietl, @Ray-Light, @jyluo, @zhangjiangqige, @the1derer, @smillst 

co-authored-by: Ravi Roshan <rroshan5647@gmail.com>
co-authored-by: Jeff Luo <j36luo@uwaterloo.ca>
co-authored-by: zhangjiangqige <zhangjiangqige@gmail.com>
co-authored-by: Shubham Raj <shubhamraj1in@gmail.com>
co-authored-by: Suzanne Millstein <smillst@cs.washington.edu>


smillst:
plume-lib tests are failing.  It's failing because of https://github.com/eisop/checker-framework/issues/22.
See test case: checker/tests/index/PlumeFail.java

checker/bin-devel/test-cf-inference.sh
```diff
@@ -26,4 +26,5 @@ BRANCH=`/tmp/plume-scripts/git-find-branch ${REPO} ${CI_BRANCH}`
 
 export AFU=`readlink -f ${AFU:-../annotation-tools/annotation-file-utilities}`
 export PATH=$AFU/scripts:$PATH
+(cd ../checker-framework-inference && ./.travis-build-without-test.sh)
```
wmdietl: Instead of these two lines, why not just execute `.travis-build.sh` here?

checker/bin-devel/javac
```diff
@@ -21,6 +21,12 @@ esac
 
 binaryDir="${mydir}"/../dist
 
+if [ -z "$LT_BIN" ] || [ "$LT_BIN" = "" ]; then
+   ltBinDir="${mydir}"/../../../jsr308-langtools/cfmake/javac.jar:${binaryDir}/javac.jar
```
wmdietl: TODO: this shouldn't refer to jsr308-langtools. Should it refer to the javac from AFU?

checker/bin-devel/test-cf-inference.sh
smillst: Done. Thanks!

checker/bin-devel/javac
smillst: It's not needed.  (The CI tests pass, but they no longer build jsr308-langtools.)  If we remove it, the following assertion fails though:
https://github.com/typetools/checker-framework/blob/0c7cf8b2febb9692d0337dfadff4392ea3c405b3/framework/src/main/java/org/checkerframework/framework/util/CheckerDevelMain.java#L56-L58

wmdietl: How does removing it completely work? Previously there was `tools.jar` on the runtime CP, so I would expect something similar again.
Is the issue maybe that AFU contains a copy of javac and therefore things work, even though we don't have a javac here?
Variable `buildDirs` below uses `annotation-file-utilities-all.jar`. Maybe that should be the slim artifact and here we add javac explicitly?

wmdietl: Put differently, what was the point of splitting up `cp` and `runtime.cp`?

smillst: Thanks!  I'd missed the `${binaryDir}/javac.jar` part.  I've fixed this in 8bfe29c.

smillst: Also the `runtime.cp` is what ends up on the `bootclasspath`, so we just put `javac.jar` there.

PR #2745, opened 2019-08-29
Manual and changelog entries for release 2.11.0

mernst:


changelog.txt
```diff
@@ -1,10 +1,21 @@
+---------------------------------------------------------------------------
+
+Version 2.11.0, August 22, 2019
+
+The Checker Framework uses the Error Prone version of javac.  Compared to
```
wmdietl: How about:

```
The Checker Framework now uses the Java 9 javac API. The Error Prone javac can be used
as a simple dependency that works on a Java 8 JVM.
```

The Checker Framework depends on the normal Java 9 javac API. We don't depend on any Error Prone extensions.
People can simply use the normal Java 9+ javac (modulo the annotated JDK).
However, the Error Prone javac is an easily available Maven dependency that continues to work on a Java 8 JVM.

Error Prone is a version of the Java 9 javac, not of the Java 8 javac.
It does contain a few bugfixes compared to the standard javac, but we don't depend on any of them in particular.

The second sentence from the changelog also appears twice below. Those sentences should similarly be updated.

mernst: Aren't there bug fixes that correct problems in the generated .class files?  I seem to recall problems with those incorrect .class files.

docs/manual/external-tools.tex
```diff
@@ -499,6 +499,14 @@
    the result is to use the Checker
    Framework compiler instead of the regular \<javac>.)
 
+   \item
+   Option 4:
```
wmdietl: Isn't this option missing putting `checker.jar` on the classpath? This option will use the right javac, but it won't have the Checker Framework classes on the classpath and the examples in the manual assume that.

changelog.txt
wmdietl: I couldn't find a list of differences between OpenJDK 9 javac and Error Prone javac.
All framework tests work with normal OpenJDK 9 to 11 javac, so if there are any differences we're not testing for them.

docs/manual/external-tools.tex
mernst: This confuses me too.  It's per instructions in https://docs.google.com/document/d/1mcOJUrYzHDMp2MfPztOmOaaSFjOWZquUN1fO_LCmXuo/edit# .  I'm just going to delete it.

PR #2747, opened 2019-09-02
Enable exampleTests now that 2.11.0 was released.

wmdietl:


docs/examples/lombok/build.gradle
```diff
@@ -4,43 +4,26 @@
...
 plugins {
     id 'java'
-    id 'net.ltgt.apt' version '0.15'
-}
-
-repositories {
-    mavenCentral()
+    id "io.freefair.lombok" version "4.0.1"
+    // Checker Framework pluggable type-checking
+    id 'org.checkerframework' version '0.3.27'
```
wmdietl: @kelloggm This PR is crashing because the Error Prone javac isn't being used.
Is there something wrong with the setup here?

smillst: I fixed this. You have to add `apply false`. See 7e29541.

PR #2751, opened 2019-09-05
Fix the framework jtreg tests on JDK 11

wmdietl:
This fixes the 4 jtreg test failures.


checker/jtreg/multipleexecutions/Main.java
```diff
@@ -24,13 +24,13 @@ public static void main(String[] args) {
         final JavaCompiler javac = ToolProvider.getSystemJavaCompiler();
         final StandardJavaFileManager fileManager = javac.getStandardFileManager(null, null, null);
         if (!doStuff(javac, fileManager)) {
-            System.exit(1);
+            return;
```
wmdietl: The commit message for this change was wrong: this caused a SecurityException in Java 8, I think caused by the move of the classes from the bootclasspath to the normal classpath. The returncode wasn't needed, so a normal return seems fine.

PR #2771, opened 2019-09-13
Fix tests using APIs that are deprecated in JDK11

nhat-nguyen:


checker/tests/lock/ChapterExamples.java
```diff
@@ -453,14 +453,14 @@ void unboxing() {
 
         // Expected, since b and c cannot be @GuardedBy("lock") since they are boxed primitives.
         // :: error: (method.invocation.invalid)
-        c = new Integer(c.intValue() + b.intValue()); // The de-sugared version
+        c = Integer.valueOf(c.intValue() + b.intValue()); // The de-sugared version
```
nhat-nguyen: This change together with the other one in this same file currently produce one unexpected diagnostic `assignment.type.incompatible`. I'm not sure why it is a type compatibilty when I'm simply assigning an `Integer` to an `Integer`.

wmdietl: This is filed as #2777 .
I would suggest undoing this one change so that we can merge all the other fixes.
In #2777 add a note about this update to the test cases that should be done before closing that issue.

PR #2773, opened 2019-09-13
Read `--patch-module` arguments from 'Patch_Modules_argsFile' file

the1derer:
This Enables CheckerMain to read all `--patch-module` arguments in `${CHECKERFRAMEWORK}/checker/dist/annotatedJDK/jdk${JDK_VERSION}/Patch_Modules_argfile`

Merge after successfully integrating AnnotatedJDK-11

This PR was initially made at eisop/checker-framework#36

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-Please see the Checker Framework manual1
+Please see the Checker Framework manual
 ([HTML](https://checkerframework.org/manual/),
 [PDF](https://checkerframework.org/manual/checker-framework-manual.pdf)).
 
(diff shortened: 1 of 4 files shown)


mernst:
This pull request is on hold because we might use a different mechanism for making the JDK available during type-checking.
cc:: @smillst

mernst:
Currently, we cannot create the artifacts (annotated version of JDK jmod files) that are needed in order to do this.
I'm closing this until we can create such modules.
At that time, this will be a good change to make to the Checker Framework.

PR #2774, opened 2019-09-13
Tweak AnnotationBuilder#checkSubtype

smillst:


javacutil/src/main/java/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -545,7 +544,7 @@ private boolean checkSubtype(TypeMirror expected, Object givenValue) {
         }
 
         if (!isSubtype) {
-            if (types.isSameType(found, expected)) {
+            if (found.toString().equals(expected.toString())) {
```
wmdietl: @smillst Could you please add a comment why you changed this from `types.isSameType` to a string comparison?
Is this a problem in other uses of `types.isSameType` as well?

smillst: I changed it to match what the error message says: `given value differs from expected, but same string representation`

>Is this a problem in other uses of types.isSameType as well?

I don't know.

wmdietl: Thanks.
I'm confused why that commit is relevant, as it doesn't seem to be on master.
It's ok here to use `toString`, as it's already on an error path. That wasn't clear from the description. Using `toString` to compare in general is poor style and a performance problem.

PR #2776, opened 2019-09-16
Document how to change the type of parameters.

smillst:


framework/src/main/java/org/checkerframework/framework/type/typeannotator/TypeAnnotator.java
```diff
@@ -1,12 +1,18 @@
...
 import org.checkerframework.framework.type.AnnotatedTypeFactory;
+import org.checkerframework.framework.type.AnnotatedTypeMirror;
+import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedExecutableType;
 import org.checkerframework.framework.type.visitor.AnnotatedTypeScanner;
 
 /**
  * {@link TypeAnnotator} is an abstract AnnotatedTypeScanner to be used with {@link
  * ListTypeAnnotator}.
  *
+ * <p>When implementing this class, each method should return {@code null} instead of calling super,
```
wmdietl: New line 36 calls super, contradicting this advice. Is the advice or the implementation incorrect?

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1554,6 +1554,14 @@ protected void applyInferredAnnotations(AnnotatedTypeMirror type, Value as) {
         applier.applyInferredType(type, as.getAnnotations(), as.getUnderlyingType());
     }
 
+    /**
+     * To add annotations to the type of method or constructor parameters, add a {@link
```
mernst: I would add `{@intheritDoc}` because there is good information in the superclass that is lost in this class.

PR #2778, opened 2019-09-18
Modify instructions to build Annotated-JDK11

the1derer:
Related PRs 
- typetools/jdk#9
- typetools/annotated-libraries#37


the1derer:
@mernst 
One thing I  am confused about is, whether we have to compile `typetools/jdk` or `typetools/jdk11u` to add to Checker Framework's release process?

Thanks.

Issue #2781, opened 2019-09-21
ClassCastException when using lambda with inferred generic destination type


twistedsquare:
With checker framework 2.10.1, I can produce an internal ClassCastException with a certain pattern of lambda plus generics code.  Here's the minimal sample:

```
import java.util.function.Function;
import java.util.stream.Stream;

public class LambaWrapped
{
    class Wrapper<T>
    {
        public Wrapper(T t) { }
    }

    public final Stream<Wrapper<Function<String, String>>> getStreamOfWrappedFunctions()
    {
        // This gives an error (inferred type in new)
        return Stream.<Wrapper<Function<String, String>>>of(new Wrapper<>(e -> e));
        // This is fine (explicit type in new)
        //return Stream.<Wrapper<Function<String, String>>>of(new Wrapper<Function<String, String>>(e -> e));
    }
}
```

I run with the NullnessChecker and I get the error below.  As noted in the example, the problem is the inferred type for the `new Wrapper<>` part.  The exception goes away if I specify the type in the new.  Glancing at the checker framework source, it looks like the code is assuming the lambda destination of a constructor parameter is a known declared type, but here that's not the case.   The Stream wrapper is not a distraction; this seems to be necessary to produce the error (because I guess the type inference becomes harder?).

I don't mind if I have to provide an annotation now for this case, but a compiler error with the real location would be better than the internal ClassCastException where I had to hunt to figure out the cause.

Error:

```
[ERROR] SourceChecker.typeProcess: unexpected Throwable (ClassCastException) while processingambaWrapped.java; message: org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedWildcardType cannot be cast to org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedDeclaredType
  Compilation unit: LambaWrapped.java
  Last visited tree at line 5 column 1:
  public class LambaWrapped
  Exception: java.lang.ClassCastException: org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedWildcardType cannot be cast to org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedDeclaredType; Stack trace: org.checkerframework.framework.type.AnnotatedTypeFactory.getFunctionalInterfaceType(AnnotatedTypeFactory.java:3688)
  org.checkerframework.framework.type.AnnotatedTypeFactory.getFnInterfaceFromTree(AnnotatedTypeFactory.java:3624)
[...]

PR #2784, opened 2019-09-23
WIP on #2781

wmdietl:


wmdietl:
Replaced by #2795, which includes the test case from here.

PR #2785, opened 2019-09-23
Allow annotations from java.lang in stubfiles.

smillst:


checker/jtreg/stubs/general/Driver.out
```diff
@@ -0,0 +1,4 @@
+- compiler.warn.proc.messager: StubParser: No supported annotations found! This likely means stub file /Users/smillst/jsr308/checker-framework/checker/jtreg/stubs/general/MyStub.astub doesn't import them correctly.
```
smillst: Before this change, 
`warning: StubParser: Unknown annotation: @SuppressWarnings("something")`
would have also been issued.

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -415,6 +415,8 @@ private void parseStubUnit(InputStream inputStream) {
                             "No supported annotations found! This likely means stub file %s doesn't import them correctly.",
                             filename));
         }
+        findPackage("java.lang");
```
wmdietl: @smillst Is this call necessary? It appears again as an argument on the line below.
If it's needed, a comment would prevent confusion.

wmdietl: https://github.com/typetools/checker-framework/commit/a37fdbe51c87f48ede8c119f1c9c23db508e7628

PR #2788, opened 2019-09-23
AnnotationUtils#getElementKindsForElementType: Ignore ElementKind.MODULE

smillst:


framework/tests/polyall/polyall.astub
```diff
@@ -4,6 +4,8 @@ import polyall.quals.*;
 
 public final class Object {
     public Class<@H1S1 ? extends @H1S1 Object> getClass();
+    @SuppressWarnings("")
```
wmdietl: How is this change relevant?

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -447,6 +447,9 @@ public static boolean hasInheritedMeta(AnnotationMirror anno) {
             case TYPE_USE:
                 return EnumSet.noneOf(ElementKind.class);
             default:
+                if (elementType.name().contentEquals("MODULE")) {
```
wmdietl: It would be good to document this in the method javadoc.
Is this a temporary workaround or should this method never support `MODULE`?

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -445,8 +445,12 @@ public static boolean hasInheritedMeta(AnnotationMirror anno) {
             case TYPE_PARAMETER:
                 return EnumSet.of(ElementKind.TYPE_PARAMETER);
             case TYPE_USE:
+            case MODULE:
```
smillst: This doesn't compile with Java 8.

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
smillst: This temporary, which I've now noted above.

wmdietl: This is also caused by source level 8. It is a bit surprising that the source level has an impact on this, but when I change the build to not pass `source 8` I can directly refer to `ElementKind.MODULE` in the switch.
As we still want to stay on Java 8, I think this is fine, but should be documented here.

wmdietl: It's already appropriately documented.

PR #2789, opened 2019-09-24
Convert PolyAll to specific polymorphic annotations.

smillst:
Also, add more tests to check that `@PolyAll` is treated identically to a type system specific polymorphic qualifier.  (Some of these test cases failed before these changes.)


checker/tests/nullness/PolyAllTest.java
```diff
@@ -1,12 +1,18 @@
...
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.checkerframework.checker.nullness.qual.PolyNull;
 import org.checkerframework.dataflow.qual.Pure;
 import org.checkerframework.framework.qual.PolyAll;
 
 public class PolyAllTest<T extends Comparable<T>> {
 
     @Pure
-    public int compare(@PolyAll T[] a1, @PolyAll T[] a2) {
+    public int comparePolyAll(@PolyAll T[] a1, @PolyAll T[] a2) {
```
smillst: If `@PolyAll` is written on a use of a type variable, then it is ignored.  Ordinarily, a type system can specify a different annotation than `@PolyAll` by with an explicit qualifier, but in the case of type variables, there is no way to specify a different qualifier.

checker/tests/interning/TestPolyAll.java
```diff
@@ -17,4 +17,12 @@ void fooRef(@Interned String p) {
     void fooArr(@Interned String[] p) {
         p = eqArr(p);
     }
+
+    @Interned @PolyAll String intern(@PolyAll TestPolyAll this) {
+        throw new RuntimeException("Doesn't matter");
+    }
+
+    void use(TestPolyAll test) {
+        @Interned String s = test.intern();
```
smillst: This test case used to fail and is the reason for this pull request.

checker/tests/nullness/PolyAllTest.java
wmdietl: I don't understand this statement and find the difference in error messages between `comparePolyAll` and `compare` very confusing.
I would expect that both methods behave the same, but they don't.
Why are `PolyNull` and `PolyAll` different on type variables?
Is this described in the manual?

What do you mean with `there is no way to specify a different qualifier`? What qualifier would you want to specify?

smillst: I've made the behavior consistent between `@PolyAll` and `@PolyNull`.  (Note, this pull request is not ready for review yet.)

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeMirror.java
```diff
@@ -1458,8 +1459,10 @@ private void fixupBoundAnnotations() {
             // annotations for all hierarchies, so we don't want to avoid placing bottom on the
             // lower bound for those hierarchies that don't have a qualifier in primaryAnnotations.
             if (!this.getAnnotationsField().isEmpty()) {
+                Set<AnnotationMirror> newAnnos =
```
wmdietl: It seems rather confusing that this replacement happens here only for type parameters.
Have you tried replacing all `PolyAlls` in `AnnotatedTypeMirror`, so that we never create an ATM with `PolyAll` in the first place?
Then we would not need the new `PolyAllReplacer` and would handle all types consistently.
At the least, I think adding an explanation to the code where which substitutions happen would be good.

framework/src/main/java/org/checkerframework/framework/type/poly/AbstractQualifierPolymorphism.java
```diff
@@ -55,6 +54,8 @@
  *       qualifier
  * </ul>
  */
+// TODO: All AnnotationMirrorMap<AnnotationMirrorSet> in this class change be changed to
```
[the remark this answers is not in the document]
smillst: I've already done it.  I'll post it for review after this branch is merged.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeMirror.java
smillst: `@PolyAll` can only be replace with type system specific poly qualifiers after all other explicit qualifiers have been added to the type. Otherwise, a poly qualifier might be added for a type system that has a different explicit annotation.  So, we can't replace all `@PolyAll` in `AnnotatedTypeMirror`.  In this case, we can because if another annotation is added the poly qualifier is replaced.  (I'll this explanation as a comment.)

smillst:
Closed in favor of #2841.

PR #2790, opened 2019-09-25
Only check for jdk8.jar not jdk11.jar.

smillst:
Also, don't check for javac.jar when using 9+.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/util/CheckerMain.java
+++ b/framework/src/main/java/org/checkerframework/framework/util/CheckerMain.java
@@ -131,7 +131,7 @@ public class CheckerMain {
             assertFilesExist(Arrays.asList(javacJar, jdkJar, checkerJar, checkerQualJar));
         } else {
             // TODO: once the jdk11 jars exist, check for them.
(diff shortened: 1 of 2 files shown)


framework/src/main/java/org/checkerframework/framework/util/CheckerMain.java
```diff
@@ -131,7 +131,7 @@ protected void assertValidState() {
             assertFilesExist(Arrays.asList(javacJar, jdkJar, checkerJar, checkerQualJar));
         } else {
             // TODO: once the jdk11 jars exist, check for them.
-            assertFilesExist(Arrays.asList(javacJar, checkerJar, checkerQualJar));
```
the1derer: @smillst  Why are we not checking for `javacJar`?

smillst: Because it's not required when using Java 9+.

the1derer: ok, Thanks for the explanation. I have to update my PR too once this is merged.

PR #2792, opened 2019-09-26
Remove --add-exports and most --add-opens.

smillst:
`--add-exports` are only required when source is set to 9 or above.  

Also, target Java 8 bytecode.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -138,19 +138,7 @@ allprojects {
                     "-Xlint:-options",
                     "-Xlint",
             ]
-            if (!isJava8) {
-                options.compilerArgs += [
-                    "--add-exports", "jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED",
-                    "--add-exports", "jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED",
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -138,19 +138,7 @@ allprojects {
                     "-Xlint:-options",
                     "-Xlint",
             ]
-            if (!isJava8) {
-                options.compilerArgs += [
```
wmdietl: At the moment we use source compatibility 8, always.
If you remove all `source 8` settings from the build files, without this you get errors like:

```
.../checker-framework-smillst/javacutil/src/main/java/org/checkerframework/javacutil/AbstractTypeProcessor.java:9: error: package com.sun.tools.javac.comp is not visible
import com.sun.tools.javac.comp.CompileStates.CompileState;
```

So how about instead of removing these we document why we have them. Once we build only for Java 11+ we will need these.

smillst: Thanks for figuring this out.  Since we want to keep source 8 for now, I added the `--add-exports` back, but commented them out.  This also means we can target 8, too.

smillst: Users can still use the Checker Framework with source levels greater than 8.  For example, the framework test pass  on this branch using Java 11.  These tests compile `checker-framework/framework/tests/all-systems/VarKeyword.java` which requires a source level of at least 9.

PR #2796, opened 2019-09-28
Making checker-framework compatible with ASM7

kushg1:
There were only a few places in the checker-framework which used some internals of annotation-tools, specifically InnerTypeLocation. These are some changes to the checker-framework which makes it compatible with the now-ASM7-supported annotation-tools.

See https://github.com/typetools/annotation-tools/pull/255

Both pull requests need to be approved together.


framework/build.gradle
```diff
@@ -44,7 +44,7 @@ task checkDependencies() {
 compileJava.dependsOn(checkDependencies)
 
 task allSourcesJar(type: Jar) {
-    description 'Creates a sources jar that includes sources for all Checker Framework classes in framework.jar'
+    description 'Creates a sources jar that includes sources for all Checker Framework classes in framewFork.jar'
```
smillst: Please revert this change.

wmdietl:
Note that ASM 8.0 was recently released: https://asm.ow2.io/versions.html
From the changelog it looks like it is hopefully easy to directly go to that version.

markro49:
This pull request was superseded by https://github.com/typetools/checker-framework/pull/4155.

PR #2797, opened 2019-09-29
Null references in String concatenations

### LINKED ISSUE #1169: Constant Value Checker: StringVal incorrect handling of null values
When myInteger is null,
```java
  "myInteger=" + myInteger
```
may evaluate to
```java
  "myInteger=null"
```
Also, when myInteger is declared as
```java
  @StringVal(22) String myInteger;
```
then
```java
  "myInteger=" + myInteger
```
may evaluate to two values:
```java
  "myInteger=22"
  "myInteger=null"
```
The Value Checker does not handle this soundly.

The below test case illustrates this problem.

It is necessary for the programmer to specify whether null is possible, and to change the Constant Value Checker to take account of those specifications.

Here are two ways to solve the problem.  Other solutions are welcome.

 1. Have the Value Checker read and trust Nullness Checker annotations if they are present.  At a conversion to String, the nullness annotations are looked up and utilized, adding "null" to the set as needed.  (The Value Checker needs to trust the Nullness Checker because we don't want to make a checker in the `framework` package depend on a checker in the `checker` package.  Careful users would want to run the Nullness Checker to verify the nullness annotations.  Maybe the Constant Value Checker would issue a warning if it depended on nullness annotations and the Nullness Checker was not currently being run.)

 2. Change the @StringVal annotation's documentation so that users are required to include "null" in its argument.  The Constant Value Checker must perform a subset of the work done by the Nullness Checker, so that it can create appropriate StringVal annotations.  Unfortunately, Java doesn't permit null to appear in the argument along with Literal strings, so the annotation needs to use "null" instead.  Furthermore, this proposal doesn't enable @IntVal to specify when an Integer might be null.  So, this proposal seems bad.

Other suggestions?

```java
// Type-checking this program with the Nullness and Constant Value checkers
// yields no errors or warnings:
//   $ch/bin/javac -g StringValNull.java -processor index,nullness
// but should yield the two marked return.type.incompatible errors.

public class StringValNull {

    @StringVal("itsValue") String nbleString = null;
    @StringVal("itsValue") String nnString = "itsValue";

    System.out.println(toString1(nbleString));
    System.out.println(toString2(nbleString));

    System.out.println(toString1(nnString));
    System.out.println(toString2(nnString));
    System.out.println(toString3(nnString));

    @IntVal(22) Integer nbleInteger = null;
    @IntVal(22) Integer nnInteger = 22;

    System.out.println(toString4(nbleInteger));
    System.out.println(toString5(nbleInteger));

    System.out.println(toString4(nnInteger));
    System.out.println(toString5(nnInteger));
    System.out.println(toString6(nnInteger));
  }

  static @StringVal("arg=itsValue") String toString1(@Nullable @StringVal("itsValue") String arg) {
    return "arg=" + arg;
  }

  static @StringVal({"arg=itsValue", "arg=null"}) String toString2(@Nullable @StringVal("itsValue") String arg) {
    return "arg=" + arg;
  }

  static @StringVal("arg=itsValue") String toString3(@StringVal("itsValue") String arg) {
    return "arg=" + arg;
  }

  static @StringVal("arg=22") String toString4(@Nullable @IntVal(22) Integer arg) {
    return "arg=" + arg;
  }

  static @StringVal({"arg=22", "arg=null"}) String toString5(@Nullable @IntVal(22) Integer arg) {
    return "arg=" + arg;
  }

  static @StringVal("arg=22") String toString6(@IntVal(22) Integer arg) {
    return "arg=" + arg;
  }

}
```

Maxi17:
Fixes #1169

In this fix attempt, every object reference is considered `@Nullable`, without checking nullness annotations. I couldn't find a way to check nullness annotations from the Value Checker or vice versa.

All objects can now be `"null"` in String concatenations. I fixed older test cases that had a similar issue.

@kelloggm @panacekcz


framework/src/main/java/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -538,6 +544,29 @@ private Range calculateLengthRangeAddition(Range leftLengths, Range rightLengths
         return leftLengths.plus(rightLengths).intersect(Range.INT_EVERYTHING);
     }
 
+    /**
+     * This method is only used to check for nullable nodes in string concatenations.
+     *
+     * @return true if the node can't be null or false otherwise.
```
kelloggm: I don't think this comment is quite correct, because this method returns "true" for String concatenation nodes, which are definitely nullable in general. Can you be more precise in the description (and the name of the method)?

framework/src/main/java/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -549,12 +578,14 @@ private AnnotationMirror createAnnotationForStringConcatenation(
         if (leftValues != null && rightValues != null) {
             // Both operands have known string values, compute set of results
             List<String> concatValues = new ArrayList<>();
-            if (leftValues.isEmpty()) {
-                leftValues = Collections.singletonList("null");
+
+            if (!isNotNullable(leftOperand)) {
```
kelloggm: That every time this is called there is a `!` first suggests that the method should be called "isNullable"

Maxi17:
@kelloggm @panacekcz 
It is ready for review again.

docs/manual/constant-value-checker.tex
```diff
@@ -334,6 +334,30 @@
...
+  // The possible value "nullb" is omitted
+  @StringVal("ab") String string2 = string1 + "b";
+\end{Verbatim}
+\caption{With the \code{-AnullStringsConcatenation} command-line option, the
+Constant Value Checker considers that the first String can be null.}
+\label{fig-null-strings-concat}
+\end{figure}
+
+It is recommended to use the Nullness Checker before using the Constant Value Checker
+with this command-line option enabled for a better analysis on nullable elements.
```
kelloggm: This recommendation doesn't make sense, because using the nullness checker doesn't help. I would suggest either:
* use the nullness checker, and do NOT use this option, or
* use this option

kelloggm:
@smillst, can you review this PR when you get a chance? Thanks!

framework/src/main/java/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -546,15 +577,29 @@ private AnnotationMirror createAnnotationForStringConcatenation(
...
+                    leftValues.add("null");
+                }
+                if (isNullable(rightOperand)) {
+                    rightValues.add("null");
+                }
+            }
+
             if (leftValues.isEmpty()) {
-                leftValues = Collections.singletonList("null");
+                leftValues.add("null");
```
panacekcz: Why is this needed? Just to preserve the old behavior?
Seems to be an exception from "The Constant Value Checker considers all elements of a String concatenation non-null". If so, and we want to keep this behavior, could it at least be guarded by `isNullable` to reduce false positives?

docs/manual/constant-value-checker.tex
Maxi17: i removed this altogether. Hopefully this is an improvement.

docs/manual/constant-value-checker.tex
```diff
@@ -334,6 +334,27 @@
 that possibility, or verify the lack of overflow using some other tool or
 manual analysis.
 
+\sectionAndLabel{Strings can be null in concatenations}{null-strings-concats}
+
+The Constant Value Checker considers all elements of a String concatenation
+non-null by default. The \code{-AnullStringsConcatenation} command-line option
+makes the checker more sound by analyzing the possibility of each element to
```
smillst: `analyzing the possibility of each element to be null` should be something like `assuming that all object could be null`.

docs/manual/constant-value-checker.tex
```diff
@@ -334,6 +334,27 @@
...
+be null. Only String literals, primitive types and effectively final variables
```
smillst: Please add a comma before `and`.

docs/manual/constant-value-checker.tex
```diff
@@ -334,6 +334,27 @@
...
+are considered non-null. Figure~\ref{fig-null-strings-concat} gives an example
```
smillst: Why are effectively final variables considered non-null?  Couldn't type be null, too?

docs/manual/constant-value-checker.tex
```diff
@@ -334,6 +334,27 @@
...
+of behavior with and without the \code{-AnullStringsConcatenation} option.
+
+\begin{figure}
+\begin{Verbatim}
+  ...
+  // This String can be null
+  @StringVal("a") String string1;
```
smillst: `string1` needs to be initialized to something; otherwise, the compiler will reject the next line.  Maybe just `getString()`?

docs/manual/constant-value-checker.tex
```diff
@@ -334,6 +334,27 @@
...
+  // The possible value "nullb" is omitted
```
smillst: The possible is omitted if `-AnullStringsConcatenation` is not used.  What happens if it is use?  Is there an error on the next line?

framework/src/main/java/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -538,6 +544,31 @@ private Range calculateLengthRangeAddition(Range leftLengths, Range rightLengths
         return leftLengths.plus(rightLengths).intersect(Range.INT_EVERYTHING);
     }
 
+    /**
+     * This method is only used to check for nullable nodes in string concatenations.
```
smillst: The comment on a method should be about what the method does, not how it is used.

framework/src/main/java/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -538,6 +544,31 @@ private Range calculateLengthRangeAddition(Range leftLengths, Range rightLengths
...
+     *
+     * @return false if the node can't take null value in a String concatenation or true otherwise.
```
smillst: I don't understand what `take null value` means.  I think you mean `false if the node can't be null in a String concatentation`.  There's no documentation for node.  Is it always an operand in a string concatenation.

framework/src/main/java/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -538,6 +544,31 @@ private Range calculateLengthRangeAddition(Range leftLengths, Range rightLengths
...
+     */
+    private boolean isNullable(Node node) {
+        if (node instanceof StringLiteralNode) {
+            return false;
+        } else if (node instanceof StringConversionNode) {
+            if (((StringConversionNode) node).getOperand().getType().getKind().isPrimitive()) {
+                return false;
+            }
+        } else if (node instanceof StringConcatenateNode) {
+            // no need to recurse because this method is called for every operand.
```
smillst: I don't understand this comment.  There's no recursion.

framework/src/main/java/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -546,15 +577,35 @@ private AnnotationMirror createAnnotationForStringConcatenation(
         List<String> leftValues = getStringValues(leftOperand, p);
         List<String> rightValues = getStringValues(rightOperand, p);
 
+        boolean allowNullStringConcat =
```
smillst: This isn't allowing `null`, it's assuming that the operand of a string concatenation might be null.

framework/src/main/java/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -546,15 +577,35 @@ private AnnotationMirror createAnnotationForStringConcatenation(
...
+                }
+                if (isNullable(rightOperand)) {
+                    rightValues.add("null");
+                }
             }
-            if (rightValues.isEmpty()) {
-                rightValues = Collections.singletonList("null");
+
+            if (leftOperand instanceof StringConversionNode) {
+                if (((StringConversionNode) leftOperand).getOperand().getType().getKind()
```
smillst: Shouldn't this be in `isNullable`?

docs/manual/constant-value-checker.tex
Maxi17: In a String concatenation, operands that are effectively final variables and are not parameters are considered non-null. Can you please give one example in which the contrary holds?

If I got it wrong, then my initial intention was to consider constant operands of String concatenation non-null, like in `"hello" + Integer.MAX_VALUE`. Then I extended it to every effectively final variable (initialized and unchanged).

framework/src/main/java/org/checkerframework/common/value/ValueTransfer.java
Maxi17: Sorry for being imprecise here. What I meant is:

In String concatenation `"hello" + 12 + Integer.MAX_VALUE`, the first operand is `"hello" + 12 and the second operand is `Integer.MAX_VALUE`. I wanted to say that if a `StringConcatenationNode` enters the method `isNullable` (like the first operand in my example is), then it is not necessary to check the operands of this `StringConcatenationNode` because the method `is Nullable` will eventually be called for the operands of String concatenation `"hello" + 12` too, since it is a String concatenation like any other.

For these simple examples it is not a problem if `isNullable` is called recursively for the operands of a `StringConcatenationNode`, but in some test cases there are some long String concatenations (see `framework/tests/StringConcats.java` line 7) in which these checks would be performed many times for no reason.

Hopefully this explains it. I deleted the comment while thinking of something more descriptive to write there.

framework/src/main/java/org/checkerframework/common/value/ValueTransfer.java
Maxi17: The reason why I haven't included this in `isNullable` is because `isNullable` performs a more complex analysis about null nodes. Here it just checks that the operand is not the `null` reference. Please refer to the comment I added to Vlastimil's change request ([link](https://github.com/typetools/checker-framework/pull/2797#discussion_r335646600)).

docs/manual/constant-value-checker.tex
smillst: final String s = null;

smillst: The fact that a variable is final (or effectively-final) has nothing to with whether or not a variable is null.

smillst: See test case in 56daab506.

framework/src/main/java/org/checkerframework/common/value/ValueTransfer.java
smillst: Thanks for the explanation.  There's no need to call isNullable recursively because a StringConcatenateNode can never be `null`.

docs/manual/constant-value-checker.tex
Maxi17: Added  commit in which I changed the `isNullable` method to reflect what I said earlier here.

smillst: Ah! compile-time constant are non-null.  Yes, I agree with that.  Please update the manual here to say that.

framework/src/main/java/org/checkerframework/common/value/ValueChecker.java
```diff
@@ -19,10 +19,21 @@
...
+})
 public class ValueChecker extends BaseTypeChecker {
+    /**
+     * Command-line option to warn the user if a @StaticallyExecutable method can't load and run at
+     * compile time.
+     */
     public static final String REPORT_EVAL_WARNS = "reportEvalWarns";
+    /** Command-line option to ignore the possibility of overflow for range annotations. */
     public static final String IGNORE_RANGE_OVERFLOW = "ignoreRangeOverflow";
+    /** Command-line option to allow elements of String concatenations to be null. */
```
smillst: This doesn't allow them to be null, it assume that most expressions could be null.

framework/src/main/java/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -538,6 +543,28 @@ private Range calculateLengthRangeAddition(Range leftLengths, Range rightLengths
         return leftLengths.plus(rightLengths).intersect(Range.INT_EVERYTHING);
     }
 
+    /**
+     * A light check for nodes that can be the null literal. Nodes that are certainly not the null
```
smillst: What's a `light check`?  I think you mean that this method isn't as precise as it could be, but this phrase isn't clear.   
The next sentence isn't clear either.  All nodes that are not an instance of `NullLiteralNode` are certainly not the null literal.  Neither sentence says what this method does.

framework/tests/value-null-strings-concatenation/Binaries.java
```diff
@@ -0,0 +1,350 @@
+import java.util.BitSet;
```
smillst: Why did you add this test case?  It doesn't have string concatenations.

framework/src/main/java/org/checkerframework/common/value/ValueTransfer.java
smillst: Right, the `null` literal should have `@StringVal("null")` even without `-AnullStringsConcatenation`.  This would be clearer if you add this check to `isNullable` and then repeat the check in an else branch like this:
```
if (nullStringConcat) {
  ...
} else {
  if (leftOperand instanceof StringConversionNode) {
    if (((StringConversionNode) leftOperand).getOperand().getType().getKind()
...
```

mernst:
@Maxi17 when you have addressed Suzanne's comments, could you please assign the issue to me so I can review it?  Thanks!

framework/tests/value-null-strings-concatenation/Binaries.java
Maxi17: It has one on line 79 ([link](https://github.com/typetools/checker-framework/blob/36cac07c83176088f7b589e200e222ec8cb13b49/framework/tests/value/Binaries.java#L79)), but that kind of assignment is overly tested in every other test, so I will remove `Binaries.java` from this test suite.

Maxi17:
@mernst It is ready for your review.

framework/src/main/java/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -538,6 +543,28 @@ private Range calculateLengthRangeAddition(Range leftLengths, Range rightLengths
         return leftLengths.plus(rightLengths).intersect(Range.INT_EVERYTHING);
     }
 
+    /**
+     * Checks whether or not the passed node is nullable. This superficial check assumes that every
+     * node is nullable unless it is a primitive, String literal, or compile-time constant.
```
mernst: The comment and code will be easier to read if they both list the cases in the same order.

Maxi17:
The change looks good to me. Thank you for the review!

framework/src/main/java/org/checkerframework/common/value/ValueTransfer.java
mernst: @Maxi17 Could you please address this comment?

Maxi17: I changed the order of checks in this method. Hopefully it's what was expected.

mernst:
@Maxi17 Can you add a changelog entry?
Also, what is the rationale for making the option default to unsound?
When possible, we prefer to have options that default to sound behavior, with a command-line option for unsound behavior.

Maxi17:
I thought the imprecision introduced by this option would be too high to leave it as default behaviour. Martin suggested to hide it behind a flag ([link](https://github.com/typetools/checker-framework/pull/2797#pullrequestreview-295312691)) and he also said to default it to true. I think I haven't read his comment carefully enough, so I did the opposite.
Since it is Checker Framework's style to add a command-line argument for unsound behaviour, I will stick to that.

Also, what do I need to do to add a changelog entry? Thanks.

mernst:
@Maxi17 You can create a branch with the option defaulted to sound, and observe whether any jobs fail in continuous integration.  (You might also want to rename the option in that case.)

For a changelog entry, just edit top-level file `changelog.txt`.  You should do that after you decide whether to change the default (or in the same pull request).

Thanks!

PR #2803, opened 2019-10-02
Add support for jdkN.astub files.

smillst:


docs/manual/creating-a-checker.tex
```diff
@@ -1656,7 +1656,9 @@
 To supply an annotated JDK that will be compiled, see Section~\ref{annotating-jdk}.
 
 To supply an annotated JDK as a stub file, create a file \<jdk.astub> in
-the checker's main source directory.  It will be automatically used by the
+the checker's main source directory.  You can also create \<jdkN.astub> files that contain methods
```
wmdietl: You renamed the existing `jdk.astub` files to be only for JDK 8, however, those classes would still exist in JDK 11.
Are those annotations already contained in the new annotated JDK for 11?
If so, a note somewhere explaining this would be good, to make sure we don't accidentally introduce inconsistencies.

docs/manual/creating-a-checker.tex
```diff
@@ -1725,7 +1727,7 @@
 \begin{itemize}
 
 \item \code{-Aignorejdkastub}:
-  ignore the \<jdk.astub> file in the checker directory. Files passed
+  ignore the \<jdk.astub> or \<jdkN.astub> file in the checker directory. Files passed
```
wmdietl: Should the `or` be an `and`? If there are three files `jdk.astub`, `jdk8.astub`, and `jdk11.astub`, are all three ignored? The sentence could use some tweaking to account for there being only a `jdk8.astub` that should be ignored (but there is no `jdk.astub`).

docs/manual/creating-a-checker.tex
smillst: Yes, the annotations in the existing `jdk.astub` have all been moved to `typetools/jdk`.  I think this should go both in https://checkerframework.org/manual/#annotating-jdk and https://checkerframework.org/manual/#reporting-bugs-annotated-libraries.  I need to update both for these to explain the current state of the JDK.  I'll do this in another PR.

PR #2805, opened 2019-10-02
Warning note about Lombok and the Checker Framework

kelloggm:


mernst:
This does not address other requests, such as distinguishing between the use cases of type-checking code that contains Lombok annotations vs. type-checking clients of such code.  It will be easier to review this short section all at once rather than having to re-assess how the whole section flows with each change.

Also, don't start out with what *not* to do.  Rather, start out by telling users what they should do.  You can then explain why, afterward, as a detail that not all users have to read.

mernst:
Can you please fix the build failure?

kelloggm:
@mernst ping on reviewing this

PR #2806, opened 2019-10-03
Update the manual and Maven examples for JDK 11.

smillst:


docs/examples/MavenExampleJDK11/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-site-plugin</artifactId>
+        <version>3.2</version>
+      </plugin>
+    </plugins>
+  </build>
+
+    <reporting>
```
wmdietl: Nit: the indentation in this file is inconsistent, e.g. between the end of `build` and the beginning of `reporting` and whether it uses 2 or 4 spaces as indentation.
Can you auto-format it? (probably the same problems for the existing .xml file.)

PR #2808, opened 2019-10-08
Changes to build files to work with JDK 11.

smillst:


build.gradle
```diff
@@ -569,19 +569,23 @@ subprojects {
...
                                 "-javacoptions:-target 8"
                                 ]
+                            if (project.name.is('checker')) {
+                                args += [
+                                        "-javacoptions:-Xbootclasspath/p:${projectDir}/dist/jdk8.jar",
+                                ]
+                            }
+                        } else {
+                            args += [
+                                    "-javacoptions:--add-modules jdk.jdeps",
```
wmdietl: Could you add a comment why this `jdk.jdeps` export is needed only here?
Why doesn't CheckerMain also need this?

smillst: I added a comment above this line.  `checker/jtreg/nullness/defaultsPersist/ReferenceInfoUtil.java` uses `jdk.deps` module.  CheckerMain doesn't need it because the Checker Framework doesn't use it.

PR #2810, opened 2019-10-10
For Java 11, parse typetools/jdk as stub files

smillst:
The Gradle plugin does not yet work with Java 11, so the plume-lib tests fail.


checker/jtreg/nullness/defaultsPersist/Fields.java
```diff
@@ -4,6 +4,7 @@
  *
  * @compile ../PersistUtil.java Driver.java ReferenceInfoUtil.java Fields.java
  * @run main Driver Fields
+ * @ignore This is fails for Java 11.
```
smillst: @wmdietl Can you take a look at this failure?  The bytecode looks to have the same annotations for both Java 8 and Java 11, but it fails on with Java 11

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1813,14 +1889,18 @@ private static void putOrAddToMap(
...
+    private void putNew(
             Map<Element, AnnotatedTypeMirror> m, Element key, AnnotatedTypeMirror value) {
         if (key == null) {
             throw new BugInCF("StubParser: key is null");
         }
         if (m.containsKey(key)) {
             AnnotatedTypeMirror value2 = m.get(key);
-            AnnotatedTypeMerger.merge(value, value2);
+            if (isJdkAsStub) {
+                AnnotatedTypeMerger.merge(value2, value);
```
wmdietl: This is another place where a comment would make it easier to understand why the order matters here.

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -542,6 +542,9 @@ private void checkSubtype(TypeMirror expected, Object givenValue) {
             found = elements.getTypeElement(givenValue.getClass().getCanonicalName()).asType();
             isSubtype = types.isSubtype(types.erasure(found), types.erasure(expected));
         }
+        if (!isSubtype) {
```
wmdietl: Is this a left-over from a different change?

framework/src/main/java/org/checkerframework/framework/StubTypes.java
```diff
@@ -0,0 +1,476 @@
...
+import org.checkerframework.framework.type.AnnotatedTypeFactory;
+import org.checkerframework.framework.type.AnnotatedTypeMirror;
+import org.checkerframework.javacutil.BugInCF;
+import org.checkerframework.javacutil.ElementUtils;
+import org.checkerframework.javacutil.PluginUtil;
+
+/** Holds information about types parsed from stub files. */
+public class StubTypes {
+    /** Types read from stub files (but not those from the annotated JDK jar file). */
+    private Map<Element, AnnotatedTypeMirror> typesFromStubFiles;
```
wmdietl: Can some/most of these fields be final?

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -173,6 +174,8 @@
     /** The line separator. */
     private static final String LINE_SEPARATOR = System.lineSeparator().intern();
 
+    private final boolean isJdkAsStub;
```
wmdietl: Some comment for the intuition for this field would be good.

checker/build.gradle
```diff
@@ -33,6 +34,53 @@ dependencies {
...
+task copyJdk11Files(dependsOn: cloneTypetoolsJdk, group: 'Build') {
+    description "Copy annotated jdk 11 files to ${buildDir}/jdk/jdk11/"
+    doLast {
+        FileTree tree = fileTree(dir: "${jdkHome}/src/")
+        SortedSet<String> annotatedForFiles = new TreeSet<>();
+        tree.visit {
+            if (!it.file.isDirectory() && it.file.name.matches(".*\\.java")) {
+                    it.file.eachLine { line ->
+                        if (line.contains("@AnnotatedFor") || line.contains("org.checkerframework")) {
+                            annotatedForFiles.add(it.file.absolutePath)
```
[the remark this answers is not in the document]
smillst: No, this is a for each in lambda, there's no way to stop it.

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
smillst: I added a comment.

checker/build.gradle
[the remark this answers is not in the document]
wmdietl: In Java I would use https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#anyMatch-java.util.function.Predicate-

Is the Groovy equivalent `.any`?

framework/src/main/java/org/checkerframework/framework/StubTypes.java
```diff
@@ -0,0 +1,476 @@
...
+import org.checkerframework.javacutil.PluginUtil;
+
+/** Holds information about types parsed from stub files. */
+public class StubTypes {
+    /** Types read from stub files (but not those from the annotated JDK jar file). */
+    private final Map<Element, AnnotatedTypeMirror> typesFromStubFiles;
+
+    /**
+     * Declaration annotations read from stub files (but not those from the annotated JDK jar file).
+     * Map keys cannot be Element, because a different Element appears in the stub files than in the
```
wmdietl: Why is this not an issue for `typesFromStubFiles` above?

framework/src/main/java/org/checkerframework/framework/StubTypes.java
```diff
@@ -0,0 +1,476 @@
...
+    /**
+     * Whether or not a stub file is currently being parsed. (If one is being parsed, don't try to
+     * parse another.)
+     */
+    private boolean parsing;
+
+    /** AnnotatedTypeFactory */
+    private final AnnotatedTypeFactory factory;
+
+    /** JDK stub files from the file system. */
```
wmdietl: For this field and the next it would be interesting to know the meaning of the keys/values.

framework/src/main/java/org/checkerframework/framework/StubTypes.java
```diff
@@ -0,0 +1,476 @@
...
+        return type == null ? null : type.deepCopy();
+    }
+
+    /**
+     * Returns the set of declaration annotations for {@code e} containing only annotations
+     * explicitly written in a stub file or the empty set if {@code e} does not appear in a stub
+     * file.
+     *
+     * @param elt element for which annotations are returned
+     * @param eltName name of the element
```
wmdietl: Why pass both the element and the element name?

framework/src/main/java/org/checkerframework/framework/StubTypes.java
```diff
@@ -0,0 +1,476 @@
...
+        return enclosingClass.getQualifiedName().toString();
+    }
+
+    /**
+     * Parses the stub file in {@code path}.
+     *
+     * @param path path to file to parse
+     */
+    private void parseStubFile(Path path) {
+        boolean oldParsing = parsing;
```
wmdietl: It isn't clear why the state of `parsing` needs to be handled like this. Can you add a comment on the field? Or here?
I would have expected that if parsing is already true, you raise an error.

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -715,11 +778,16 @@ private void processCallableDeclaration(CallableDeclaration<?> decl, ExecutableE
                 annotate(
                         methodType.getReceiverType(),
                         decl.getReceiverParameter().get().getAnnotations());
+                annotate(
```
wmdietl: A comment here would help explain why two `annotate` calls for the receiver type are here.

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1291,11 +1367,17 @@ private VariableElement findElement(
      *     or null if method element is not found
      */
     private ExecutableElement findElement(TypeElement typeElt, MethodDeclaration methodDecl) {
+        if (isJdkAsStub && methodDecl.getModifiers().contains(Modifier.privateModifier())) {
+            // Don't process private methods of the jdk.  They can't be referenced outside of the
+            // jdk
```
wmdietl: The line break is odd.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1204,6 +1170,7 @@ public AnnotatedTypeMirror getTypeOfExtendsImplements(Tree clause) {
     // They only include qualifiers explicitly inserted by the user.
     // **********************************************************************
 
+    StubTypes stubTypes = new StubTypes(this);
```
wmdietl: It's a bit hard to see in the diff... Is this a field? Why does it have default visibility? Can it be final? Why is it down here and not with other fields at the beginning?

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -542,6 +542,11 @@ private void checkSubtype(TypeMirror expected, Object givenValue) {
...
+            // Annotations in stub files sometimes are the same type, but Types#isSubtype fails
+            // anyways.
+            isSubtype = found.toString().equals(expected.toString());
+        }
 
         if (!isSubtype) {
             if (found.toString().equals(expected.toString())) {
```
wmdietl: After the change above, can this ever be true?

framework/src/main/java/org/checkerframework/framework/StubTypes.java
smillst: This is copied from the existing implementation.   I don't know why it's different.

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
smillst: Done above.

checker/jtreg/nullness/defaultsPersist/Fields.java
wmdietl: I've opened #2816 for this. I won't get to it before the November release.

Issue #2809, opened 2019-10-09
Cannot infer annotations of new operator with type arguments when using diamond operator <>


abslamp:
When using new operators to create an object with type arguments,  the annotations on the type arguments cannot be inferred when using the diamond operator <>.

For example:

```
class NewOpTest {

  void new2(MyType<int @Interned []> t, int @Interned [] non) {
    t.self(new MyType<>(non));  // error
  }

  void new2_1(MyType<int @Interned []> t, int @Interned [] non) {
    t.self(new MyType<int @Interned []>(non));  // good
  }
  void new3(MyType<@Interned MyType<Object>> t, @Interned MyType<Object> non) {
    t.self(new MyType<>(non));  // error
  }

  private class MyType<MyTypeParam2F8A> {
    MyType (MyTypeParam2F8A f2a8) {}
    void self(MyType<MyTypeParam2F8A> myType) {}
  }
}
```
Output:
```
NewOpTest.java:14: error: [argument.type.incompatible] incompatible types in argument.
    t.self(new MyType<>(non));  // error
           ^
  found   : NewOpTest.MyType<@Interned int []>
  required: NewOpTest.MyType<@Interned int @Interned []>
NewOpTest.java:22: error: [argument.type.incompatible] incompatible types in argument.
    t.self(new MyType<>(non));  // error
           ^
  found   : NewOpTest.MyType<NewOpTest.MyType<Object>>
  required: NewOpTest.MyType<NewOpTest.@Interned MyType<Object>>
2 errors
```
The `@Interned` is missing in the infered type but exists in the explicit type.

Tested with `@Interned`, `@Untainted`, `@RegexBottom`, and `@Nullable`, all of them are replaced with its `@Unknown***` version.

When preventing cache of  `NEW_CLASS` types in `AnnotatedTypeFactory::fromExpression(ExpressionTree)`  the annotations could be correctly inferred. (As in https://github.com/typetools/checker-framework/pull/2643/commits/4724d053999ac2d434008b253caf158f80551501)
https://github.com/typetools/checker-framework/blob/0668b259c5214440d49736a6f773508c8a37a466/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java#L1348

abslamp:
Wondering if it is similar to #2781

PR #2812, opened 2019-10-15
#2809: disable caching of NEW_CLASS

abslamp:
AnnotatedTypeFactory.fromExpressionTreeCache will return types with outdated type variables when visiting new operators. Disable caching of NEW_CLASS could be a solution for #2809.

### DIFF
--- a/checker/tests/interning/Issue2809.java
+++ b/checker/tests/interning/Issue2809.java
@@ -6,11 +6,11 @@ import org.checkerframework.checker.interning.qual.UnknownInterned;
 
 class Issue2809 {
 
-    void new2(MyType<int @Interned []> t, int @Interned [] non) {
+    void new1(MyType<int @Interned []> t, int @Interned [] non) {
         t.self(new MyType<>(non));
     }
 
-    void new2_1(MyType<int @Interned []> t, int @Interned [] non) {
+    void new2(MyType<int @Interned []> t, int @Interned [] non) {
         t.self(new MyType<int @Interned []>(non));
     }
 
--- a/checker/tests/interning/Issue2809.java
+++ b/checker/tests/interning/Issue2809.java
@@ -25,6 +25,7 @@ class Issue2809 {
 
     class MyType<T> {
         MyType(T p) {}
+
         void self(MyType<T> myType) {}
     }
 }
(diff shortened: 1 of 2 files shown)


framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1345,7 +1345,11 @@ private AnnotatedTypeMirror fromExpression(ExpressionTree tree) {
 
         AnnotatedTypeMirror result = TypeFromTree.fromExpression(this, tree);
 
-        if (shouldCache) {
+        if (shouldCache && tree.getKind() != Tree.Kind.NEW_CLASS) {
+            // Don't cache types of NEW operator
+            // Cached NEW_CLASS types may contain "Unknowned*" annotations
```
smillst: The cached types may contain incorrect annotations that were computed during dataflow.  The better solution is to only clear the cache if dataflow is running.

wmdietl: @smillst Thanks for the pointer to #602!
@abslamp Can you look into disabling caching during dataflow analysis and fixing #602?

abslamp: I will change the comments for now, but I want to look into the better solution.

abslamp:
TODO: disable cache in dataflow

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1345,7 +1345,11 @@ private AnnotatedTypeMirror fromExpression(ExpressionTree tree) {
...
+            // Such annotations are from previous dataflow analysis
```
wmdietl: @abslamp Can you update this PR and change this comment as suggested in https://github.com/typetools/checker-framework/pull/2643#discussion_r343972843

checker/tests/interning/Issue2809.java
```diff
@@ -0,0 +1,25 @@
+// test cases for #2809
+// https://github.com/typetools/checker-framework/issues/2809
+
+import org.checkerframework.checker.interning.qual.Interned;
+
+class Issue2809 {
```
wmdietl: Could you add an expected failure test, to show that things that should go wrong still cause an error?

abslamp: Do you mean a false negative one or just a normal expected failure?

PR #2819, opened 2019-10-22
Correctly infer bounds for single-ended @IntRange types with wrapper classes; fixes #2815

### LINKED ISSUE #2815: IntRange annotation defaults wrong for non-Long primitive wrapper types
When written as `@IntRange(from=0)` or `@IntRange(to=42)`, the other parameter is defaulted to the minimum/maximum value allowed for the type that is being annotated.

#1300 made it so that the correct extreme values are chosen for non-long primitive types: `int`, `short`, `byte`. However, their corresponding wrapper types still get `Long.MIN_VALUE` and `Long.MAX_VALUE` defaults, which is wrong.

Simple test:
```java
void good(@IntRange(from=0, to=Integer.MAX_VALUE) Integer baz,
          @IntRange(from=0) int quux) {
    quux = baz;
}
void bad(@IntRange(from=0) Integer foo, @IntRange(from=0) int bar) {
    bar = foo;
}
```
```
error: [argument.type.incompatible] incompatible types in argument.
    bar = foo;
          ^
  found   : @IntRange(from=0, to=9223372036854775807) Integer
  required: @IntRange(from=0, to=2147483647) int
```

Tested with Gradle plugin 0.4.5 and Checker Framework 2.11.1 (overridden manually from the plugin’s default 2.11.0).

kelloggm:
Fixes #2815


framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -351,6 +353,25 @@ public long getFromValueFromIntRange(AnnotatedTypeMirror atm) {
...
+                    case "java.lang.Short":
+                        from = Short.MIN_VALUE;
+                        break;
+                    case "java.lang.Byte":
+                        from = Byte.MIN_VALUE;
+                        break;
+                    case "java.lang.Character":
+                        from = Character.MIN_VALUE;
+                        break;
+                    default:
```
mernst: Use of `default` was a contributor to the problem.  Please make the `default` case throw an exception, and have an explicit test for `"java.lang.Long"`.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -386,6 +408,25 @@ public long getToValueFromIntRange(AnnotatedTypeMirror atm) {
...
+                    case "java.lang.Short":
+                        to = Short.MAX_VALUE;
+                        break;
+                    case "java.lang.Byte":
+                        to = Byte.MAX_VALUE;
+                        break;
+                    case "java.lang.Character":
+                        to = Character.MAX_VALUE;
+                        break;
+                    default:
```
mernst: Same comment as above regarding defaulting.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -386,6 +408,25 @@ public long getToValueFromIntRange(AnnotatedTypeMirror atm) {
...
+                        to = Long.MAX_VALUE;
+                }
+                break;
             default:
```
mernst: Same comment as above regarding defaulting.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
kelloggm: I can add an explicit case for `Long`. However, I'm not sure that the default case should throw an exception. Do we intend for `@IntRange` only to be applicable to `int`, `Integer`, `long`, etc.? I don't think so - for example, as a user I would expect to be able to write e.g `@IntRange(from=0, to=10) BigInteger x`.

mernst: What values do you propose as the default `to` and `from` for BigInteger? 

I think it's better for the checker to fail in unhandled cases rather than defaulting to a value that might not make sense.

kelloggm: I think it would be better to keep the current behavior, which is to default to `Long.MIN_VALUE` and `Long.MAX_VALUE` for `from` and `to`, respectively. In particular, I am concerned about breaking code using this checker that currently typechecks and is not incorrect.

As written, this change can only make the checker more precise. Adding an exception in the default case, as you suggest, would break any clients that use a single-ended `@IntRange` annotation on a class not on our whitelist (that is, the built in integral types). I don't see a compelling reason to break such clients, nor a compelling reason why they shouldn't exist.

If the Value Checker currently required that `@IntRange` only be written on built-in integral types, then I would support the change you're requesting. However, I don't think it would be wise to permit such annotations in general, but not in this particular use-case.

mernst: Please make the change.

As I noted, this is the reason for the bug.  It is wise, when you learn of a bug, to fix the root cause that might lead to other problems in the future.

You speculate about existing clients.  If any existing clients exist, the default behavior is as likely to be incorrect as to be correct.  It's better for clients to be notified of a problem than to silently preserve behavior that may be incorrect.

astiob:
Thanks!

Just to clarify based on the above comments and the final code (although this isn’t currently relevant to me personally): is `@IntRange` still allowed on non-built-in-wrapper class types such as `@BigInteger` as long as both bounds are explicit? The exception messages seem to suggest it is not.

mernst:
@astiob It is not.  There is no mechanism for specifying BigInteger bounds, so we don't permit long bounds on them either.

PR #2825, opened 2019-10-23
Enable Daikon typecheck with JDK 11.

smillst:


mernst:
The daikon_jdk11 job doesn't work right now because Daikon doesn't yet compile with JDK 11.
It is possible to compile Daikon with JDK 8, then type-check it with JDK 11, but the job doesn't do that (and we won't change it to do so; we will wait until Daikon compiles with JDK 11).

PR #2832, opened 2019-10-28
Improve DOTCFGVisualizer.

xingweitian:
This PR fixes:
1. Currently, in the pdf file generated from DOTCFGVisualizer, the shape of regular and exception Block is oval but not rectangle.
2. When we pass the argument `verbose` to args as a String, `this.verbose = false`. We should use `parseBoolean()` instead of `getBoolean()`.

In addition, some useless symbols and newlines are removed to make the generated dot file looks better, the method `dotOutputFileName()` is improved to generate a better file name when there is no `checkName`.


dataflow/src/main/java/org/checkerframework/dataflow/cfg/DOTCFGVisualizer.java
```diff
@@ -180,8 +177,10 @@ protected String dotOutputFileName(UnderlyingAST ast) {
         } else {
             throw new BugInCF("Unexpected AST kind: " + ast.getKind() + " value: " + ast);
         }
-        outFile.append("-");
-        outFile.append(checkerName);
+        if (!checkerName.isEmpty()) {
```
xingweitian: Here we don't need to check if `checkerName != null` since all the usages of it make sure that `checkerName` will never be `null`.

In playground, `checkerName` will be set as `""`:
https://github.com/typetools/checker-framework/blob/master/dataflow/src/main/java/org/checkerframework/dataflow/cfg/JavaSource2CFGDOT.java#L154

xingweitian:
@wmdietl This PR is ready for review.

wmdietl:
@smillst @mernst The Travis tests are passing, so I would guess the dependent projects are all up-to-date. However, the Azure build is failing.
Any hints for what else could be wrong?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/DOTCFGVisualizer.java
wmdietl: Can you improve error handling?
I would add a null check here and make that field nullable. Maybe the configuration map doesn't contain that key and that is ok.
In contrast, field `outDir` must be non-null and that configuration always must be provided. So add a check for that value to be non-null in the constructor.

smillst:
@wmdietl, I don't understand why the Travis test is passing: @xingweitian's fork of checker-framework-inference is forked from opprop/checker-framework-inference, so it isn't up-to-date:
https://github.com/xingweitian/checker-framework-inference/

wmdietl:
The corresponding branch is up-to-date with typetools/master: https://github.com/xingweitian/checker-framework-inference/tree/typetools-improve-dotcfgvisualizer

So maybe Azure isn't using the same logic as Travis to find the right branch?

xingweitian:
@wmdietl The comment has been resolved.

I think the failed Azure build is because I did not update my corresponding CFI branch on time, I updated it after the Azure building started. So Azure did not get the latest CFI code.

Now all the test groups pass.

xingweitian:
@wmdietl This PR also needs to be cleaned. I will also do the force-push.

xingweitian:
@wmdietl The clean-up is done, this PR is updated and ready for review.

wmdietl:
Thanks for all these improvements @xingweitian !

PR #2837, opened 2019-10-28
Use different ManualTaglets depending on JDK version.

smillst:
Also, run "misc" tests with JDK 11.

### DIFF
--- a/framework-test/src/tagletJdk11/java/org/checkerframework/taglet/ManualTaglet.java
+++ b/framework-test/src/tagletJdk11/java/org/checkerframework/taglet/ManualTaglet.java
@@ -1,0 +1,122 @@
+package org.checkerframework.taglet;
+
+import com.sun.source.doctree.DocTree;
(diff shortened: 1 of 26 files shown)


framework-test/src/tagletJdk11/java/org/checkerframework/taglet/ManualTaglet.java
```diff
@@ -0,0 +1,122 @@
...
+ * <p>The {@code @checker_framework.manual} tag is used as follows:
+ *
+ * <ul>
+ *   <li>{@code @checker_framework.manual #} expands to a top-level link to the Checker Framework
+ *       manual
+ *   <li>{@code @checker_framework.manual #anchor text} expands to a link with some text to a
+ *       particular part of the manual
+ * </ul>
+ */
+public class ManualTaglet implements Taglet {
```
smillst: @mernst This is copied from mernst/com-sun-javadoc-deprecated.

PR #2846, opened 2019-10-31
Don't cache default qualifiers for uses while parsing stub files

smillst:


framework/src/main/java/org/checkerframework/framework/type/typeannotator/DefaultQualifierForUseTypeAnnotator.java
```diff
@@ -75,7 +75,7 @@ public void clearCache() {
                 }
             }
         }
-        if (typeFactory.shouldCache) {
+        if (typeFactory.shouldCache && !typeFactory.stubTypes.isParsing()) {
```
wmdietl: @smillst Could you add a comment why this is needed? Is it to prevent the cache to get too large with elements that won't be interesting? Or some other reason?

smillst: Done a572867f4a.

Issue #2847, opened 2019-10-31
Assignment context and caching


wmdietl:
@smillst Using assignment context seems to have a problem with caching. If the caching is only disabled in dataflow, the nullness checker will crush when checking https://github.com/typetools/checker-framework/blob/master/framework/tests/all-systems/Issue2302.java. 
When visiting the return part in InitializationAnnotatedTypeFactory::visitNewClass (node=`new StrangeConstructorTypeArgs(new MyClass<>())`), the type of the assignment context is: `Pair(StrangeConstructorTypeArgs, @Initialized @NonNull StrangeConstructorTypeArgs<V extends @Initialized @Nullable Object>)`, which is not good enough for `MyClass<>()`. (But I suppose it is good enough for returning value?) By using the context, the type of `MyClass<>()` would be resolved into `MyClass<Object>` and cached, which would cause a crush later when checking the arguments, 
```
error: AnnotatedTypeMirrors aren't structurally equal.
    type1 = AnnotatedDeclaredType( @Initialized @NonNull Object )
    type2 = AnnotatedArrayType( @Initialized @NonNull byte @Initialized @NonNull [] )
    visitHistory = org.checkerframework.framework.type.StructuralEqualityVisitHistory@2a798d51
  Compilation unit: /home/pu/checker-framework/checker/tests/all-systems/Issue2302.java
  Last visited tree at line 17 column 16:
          return new StrangeConstructorTypeArgs(new MyClass<>());
  Exception: java.lang.Throwable; Stack trace: org.checkerframework.javacutil.BugInCF.<init>(BugInCF.java:14)
  org.checkerframework.framework.type.visitor.AbstractAtmComboVisitor.defaultAction(AbstractAtmComboVisitor.java:52)
  org.checkerframework.framework.type.StructuralEqualityComparer.defaultAction(StructuralEqualityComparer.java:68)
  org.checkerframework.framework.type.StructuralEqualityComparer.defaultAction(StructuralEqualityComparer.java:31)
  org.checkerframework.framework.type.visitor.AbstractAtmComboVisitor.visitDeclared_Array(AbstractAtmComboVisitor.java:123)
  org.checkerframework.framework.util.AtmCombo.accept(AtmCombo.java:303)
  org.checkerframework.framework.type.StructuralEqualityComparer.areEqual(StructuralEqualityComparer.java:110)
[...]

PR #2851, opened 2019-11-01
Disabling cache in data flow cause error

abslamp:
Related issue: #2847. The purpose of this pull request is to show that disabling cache in data flow would cause a crash in NullnessChecker while checking https://github.com/typetools/checker-framework/blob/79d0882e30d36a627f612d012361b055f6529a17/framework/tests/all-systems/Issue2302.java.

It is not trying to fix the issue (by now).


wmdietl:
This illustrates #602, but we're punting on fixing this larger issue for now.

PR #2852, opened 2019-11-02
`-AnullStringsConcatenation` sound by default

Maxi17:
@mernst 
This is how the option would be if it was sound by default. I'm not sure that moving existing tests around is a good idea, so I would keep the previous version. If you still recommend keeping the changes in this PR, I will rename the command-line option. I will add the entry log after a decision is taken.


kelloggm:
Please add a changelog entry to this pull request, and then re-assign it to Mike.

changelog.txt
```diff
@@ -1,3 +1,6 @@
+Every operand of a String concatenation can be "null", unless it is a String literal,
+primitive or compile-time constant.
```
mernst: This is a minor bug fix, which does not need to go in the changelog.
Any new or changed command-line argument needs to be in the changelog, however.  Could you please add a note about the command-line argument?

Maxi17:
@mernst I made the changes you requested. Hopefully I haven't missed any.

framework/src/main/java/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -574,13 +574,13 @@ private AnnotationMirror createAnnotationForStringConcatenation(
         List<String> rightValues = getStringValues(rightOperand, p);
 
         boolean nullStringConcat =
```
mernst: Please write documentation about what this means.  It's surprising that you could invert its meaning without renaming the variable.

Maxi17: This was something I overlooked when I changed the names of everything related to the command-line option. Thank you for noticing it.

Maxi17:
I made a small change to the manual. Otherwise, it looks good.

docs/manual/constant-value-checker.tex
```diff
@@ -334,30 +334,30 @@
...
 
-By default, the Constant Value Checker is unsound with respect to string
-concatenation.  The \code{-AnullStringsConcatenation} command-line option
-eliminates this unsoundness, though it may lead to false positive warnings.
+\sectionAndLabel{Strings can be null in concatenations}{non-null-strings-concats}
+
+By default, the Constant Value Checker is sound with respect to string
+concatenation and nullness.  It assumes that, in a string concatenation,
+every non-primitive argument might be null, except for String literals
+and compile-time constants. It ignores Nullness Checker annotations.
```
mernst: 1. String literals are a type of compile-time constant, so I don't think they need to be called out explicitly.
2. I don't see any test cases for compile-time constants that are not string literals.  Could you add some?  It might be enough to duplicate the `nullConcatenation` test (in both places it occurs) and change the local variables to `final` in the copy.

Maxi17:
@mernst I'm sorry for the delay. I added a test for integer compile-time constants.

mernst:
Thanks!

PR #2858, opened 2019-11-05
Perform comparisons between AnnotationMirror strings in reversed order

nhat-nguyen:
Comparisons between different `AnnotationMirror`'s names are done very frequently, and since
all `AnnotationMirrors` generated by CF always start with `org.checkerframework`, we usually
end up having to go very far into the strings only to realize that they are different. To take advantage of this property, we can perform the comparison backwards.

The only catch here is that `String.equals(s)` can access the internal array directly without having to go through bounds check which `charAt(i)` has to. So the performance gain might not be as much.


mernst:
Have you considered interning the strings (say, in `annotationName()`?  That is an alternative to the proposed change.  It might improve efficiency more (comparison becomes simply `==`).  It does not require you to hand-code a comparison function.

nhat-nguyen:
I initially thought about interning the strings, but it seems like we previously had some performance issues with it: https://github.com/typetools/checker-framework/pull/1631

smillst:
The `annotationName` field of `org.checkerframework.javacutil.AnnotationBuilder.CheckerFrameworkAnnotationMirror` is interned, so if both arguments to `areSameByName` are `CheckerFrameworkAnnotationMirrors` you could check that their names are `==`.   I don't know if that would be faster than this change though.

nhat-nguyen:
I ran the microbenchmark and nullnessfbtest 15 times on different variants of this approach:
+ always do comparisons from end
+ check the last few characters before falling back to the default .equals() method

Unfortunately, none of them showed any performance benefits.

PR #2859, opened 2019-11-05
Consolidate HashMap.containsKey() followed by HashMap.get() into a single get() call

nhat-nguyen:
Profiling results show that we are spending quite a bit of time getting the hashes of objects inside maps. Even though they are fast, the numbers add up very quickly since we have many maps used for different purposes.

This PR tries to consolidate a common code pattern in the codebase (`Map.containsKey()` followed by `Map.get()`), which result in 2 calls to `Object.hashCode()`, into just a single call to `Map.get()`.

Other Hash useful methods like `putIfAbsent` are also used where applicable.


smillst:
@nhat-nguyen Could you please merge typetools/master into this branch so that the CI tests will run again?

nhat-nguyen:
@smillst I have rebased from master but the tests still time out.

mernst:
@nhat-nguyen A small note:  in general, there is no need to rebase; you can just pull master.  (This creates a merge conflict, but when we squash-and-merge your branch, that will do the rebasing.)

smillst:
@nhat-nguyen Don't worry about the Travis time outs.  This is a general problem that I'm working to fix.

Also, never rebase or force any commits to a branch is in a pull request.  If you do, then all the review comments will point to nonsensical locations and make it impossible to determine which comments have been addressed.  (Luckily there were no comments on this PR, so no harm was done.)  We always squash pull requests so that they are pushed to master as one clean commit with a descriptive commit message, so the commits in the pull request aren't import.

smillst:
The daikon job on Azure timed out.  That's unexpected and could mean that this change actually makes the performance worse.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -3028,9 +3028,10 @@ private AnnotationMirror getDeclAnnotation(
      * @param elt the element for which to determine annotations
      */
     public Set<AnnotationMirror> getDeclAnnotations(Element elt) {
-        if (cacheDeclAnnos.containsKey(elt)) {
```
smillst: `if (map.containsKey(..) { ..}` is only equivalent to `value = map.get(...); if (value != null) {...}` if values in the map are never `null`.  Values in `cacheDeclAnnos` may be null, so this change should be reverted.  Can you check all the other changes to see if any of those maps contain null values?

mernst: A way to be sure about that would be to run the Nullness Checker, though it is challenging to run on some parts of the codebase (I remember the parser and AST, in particular, causing problems in the past).

nhat-nguyen: @smillst I think the values in `cacheDeclAnnos` shouldn't be null because the only place where we put new values into it is also in this same function:

```
// Add the element and its annotations to the cache.
cacheDeclAnnos.put(elt, results);
```

But the `results` variable is always initialized by a call to `AnnotationUtils.createAnnotationSet();` which never returns null.

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractStore.java
```diff
@@ -853,9 +853,9 @@ private S upperBound(S other, boolean shouldWiden) {
             // other are discarded, as one of store implicitly contains 'top'
             // for that variable.
             FlowExpressions.LocalVariable localVar = e.getKey();
-            if (localVariableValues.containsKey(localVar)) {
+            V thisVal = localVariableValues.get(localVar);
```
nhat-nguyen: @smillst I gave all of these another look again and it looks like all these maps might have null values if the `Store` is a `LockStore`. Inside `CFAbstractStore`, we always make sure that only non-null new entry is added to these maps. However, for `LockStore`, we simply use the return result of `changeLockAnnoToTop` (internally calls `createAbstractValue` which can return `null`) without checking for nullness.

wmdietl: The documentation for LockStore#changeLockAnnoToTop  makes it sound that it should also never return null. So I think the right thing to do is to see what goes wrong if you add a null check there, too.

PR #2861, opened 2019-11-05
Don't call getAnnotatedTypeLhs on lambda parameters (Fixes #2853)

### LINKED ISSUE #2853: performance regression between 2.9.0 and 2.10.0
Running the nullness checker on the following program takes ~50x longer with 2.10.0 than with 2.9.0:

```java

import com.google.common.collect.ImmutableSet;

abstract class A {

  abstract B getB();

  public abstract C getC();

  public abstract Object getD();

  public abstract Object getE();

  public abstract Object getF();

  public abstract Object getG();

  public abstract H getH();
}

abstract class B {}

abstract class C {
  abstract Object getI();
}

class I {
  static class J {}
}

abstract class H {

  abstract Object getK();

  abstract Object getL();

  abstract Object getM();

  abstract Object getN();
}

abstract class O {}

abstract class J {
  static O f(B b) {
    throw new AssertionError();
  }
}

abstract class K {
  abstract Object getL();
}

abstract class M {
  abstract N getN();
}

abstract class N {
  abstract Object f();
}

class Test {

  static final ImmutableSet<P> C =
      new ImmutableSet.Builder<P>()
          .add(R.c((A x) -> J.f(x.getB())))
          .add(R.c((A x) -> x.getC().getI()))
          .add(R.c((M x) -> x.getN().f()))
          .add(R.c((A x) -> x.getH().getK()))
          .add(R.c((A x) -> x.getH().getM()))
          .add(R.c((A x) -> x.getH().getL()))
          .add(R.c((A x) -> x.getH().getN()))
          .add(R.c((A x) -> x.getD()))
          .add(R.c((A x) -> x.getE()))
          .add(R.c((A x) -> x.getE()))
          .add(R.c((A x) -> x.getG()))
          .add(R.c((K x) -> x.getL()))
          .build();

  interface P {
  }

  interface Q<U, V> {
    V get(U message);
  }

  private static class R<T, U> implements P {

    static <T, U> R<T, U> c(Q<T, U> x) {
      throw new AssertionError();
    }
  }
}
```

```
$ time ./checker-framework-2.9.0/checker/bin/javac -cp guava.jar -processor Nullness Test.java
...
real	0m1.754s
$ time ./checker-framework-2.10.0/checker/bin/javac -cp guava.jar -processor Nullness Test.java
...
real	1m35.453s
```

smillst:
Calling getAnnotatedTypeLhs on a lambda parameter node is possibly expensive because caching is turned off and the type of every expression that contributes to the type of the lambda parameter is recomputed.  This should be fixed by #979.

### DIFF
--- a/framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
+++ b/framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
@@ -1014,26 +1014,23 @@ public class BaseTypeVisitor<Factory extends GenericAnnotatedTypeFactory<?, ?, ?
         warnAboutTypeAnnotationsTooEarly(node, node.getModifiers());
 
         Pair<Tree, AnnotatedTypeMirror> preAssCtxt = visitorState.getAssignmentContext();
-        visitorState.setAssignmentContext(
-                Pair.of((Tree) node, atypeFactory.getAnnotatedType(node)));
+        AnnotatedTypeMirror variableType;
+        if (getCurrentPath().getParentPath() != null
+                && getCurrentPath().getParentPath().getLeaf().getKind()
+                        == Tree.Kind.LAMBDA_EXPRESSION) {
+            // Calling getAnnotatedTypeLhs on a lambda parameter node is possibly expensive
+            // because caching is turned off.  This should be fixed by #979.
+            // See https://github.com/typetools/checker-framework/issues/2853 for an
+            // example.
+            variableType = atypeFactory.getAnnotatedType(node);
+        } else {
+            variableType = atypeFactory.getAnnotatedTypeLhs(node);
+        }
+        visitorState.setAssignmentContext(Pair.of(node, variableType));
 
         try {
             if (atypeFactory.getDependentTypesHelper() != null) {
-                if (getCurrentPath().getParentPath() != null
-                        && getCurrentPath().getParentPath().getLeaf().getKind()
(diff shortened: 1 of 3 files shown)


framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1019,9 +1019,21 @@ public Void visitVariable(VariableTree node, Void p) {
...
+                if (getCurrentPath().getParentPath() != null
+                        && getCurrentPath().getParentPath().getLeaf().getKind()
+                                == Tree.Kind.LAMBDA_EXPRESSION) {
+                    // Calling getAnnotatedTypeLhs on a lambda parameter node is possibly expensive
+                    // because caching is turned off.  This should be fixed by #979.
+                    // See https://github.com/typetools/checker-framework/issues/2853 for an
+                    // example.
+                    atypeFactory
+                            .getDependentTypesHelper()
+                            .checkType(visitorState.getAssignmentContext().second, node);
```
wmdietl: I'm trying to understand the change in #2649: why isn't the assignment context always the declared type?
That is, why did #2649 change this line to `getAnnotatedTypeLhs` instead of using the assignment context?
Is the real problem maybe that the assignment context isn't set correctly to the declared type?

smillst: > Is the real problem maybe that the assignment context isn't set correctly to the declared type?

Yes, that seems right.  I'll make that change.

PR #2870, opened 2019-11-07
Change toString of uninferred type arguments.

smillst:
Here's an example:
```
Streaming.java:9: error: [assignment.type.incompatible] incompatible types in assignment.
        List<String> listA = Stream.of("A").collect(Collectors.<String>toList());
                                                   ^
  found   : @Initialized @NonNull  /*INFERENCE FAILED ? extends java.lang.Object*/
  required: @UnknownInitialization @Nullable List<@Initialized @NonNull String>
```

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/type/DefaultAnnotatedTypeFormatter.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/DefaultAnnotatedTypeFormatter.java
@@ -386,36 +386,30 @@ public class DefaultAnnotatedTypeFormatter implements AnnotatedTypeFormatter {
         public String visitWildcard(AnnotatedWildcardType type, Set<AnnotatedTypeMirror> visiting) {
             StringBuilder sb = new StringBuilder();
             if (type.isUninferredTypeArgument()) {
-                sb.append(
-                        annoFormatter.formatAnnotationString(
-                                type.getAnnotationsField(), currentPrintInvisibleSetting));
-                sb.append(" /*INFERENCE FAILED for: ");
-                sb.append(type.getUnderlyingType());
-                sb.append("*/");
-            } else {
+                sb.append(" /*INFERENCE FAILED for:*/ ");
+            }
 
-                sb.append(
-                        annoFormatter.formatAnnotationString(
-                                type.getAnnotationsField(), currentPrintInvisibleSetting));
+            sb.append(
+                    annoFormatter.formatAnnotationString(
+                            type.getAnnotationsField(), currentPrintInvisibleSetting));
 
-                sb.append("?");
-                if (!visiting.contains(type)) {
+            sb.append("?");
+            if (!visiting.contains(type)) {
 
-                    try {
-                        visiting.add(type);
+                try {
+                    visiting.add(type);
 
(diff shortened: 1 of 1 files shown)


smillst:
@mernst  FYI, you can hide white space changes in a pull request:
https://github.blog/2018-05-01-ignore-white-space-in-code-review/.

mernst:
Good point.

I would like to use that setting, but I can't find the button in my browser window while doing a code review.  I wonder why.

`?w=1` does help, but it isn't as good as Patience Diff.  (And I am constantly forgetting about `?w=1`.)

smillst:
That blog post is out-of-date.  The button is now a gear:
https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-comparing-branches-in-pull-requests#diff-view-options.

mernst:
Thanks.  I can see it now.  I wish there was a way to default whitespace hiding.  (That is, it could be part of the "always" section of the settings.)

framework/src/main/java/org/checkerframework/framework/type/DefaultAnnotatedTypeFormatter.java
```diff
@@ -385,26 +385,37 @@ public String visitNull(AnnotatedNullType type, Set<AnnotatedTypeMirror> visitin
...
-            sb.append(
-                    annoFormatter.formatAnnotationString(
-                            type.getAnnotationsField(), currentPrintInvisibleSetting));
-            sb.append("?");
-            if (!visiting.contains(type)) {
+            if (type.isUninferredTypeArgument()) {
+                sb.append(
+                        annoFormatter.formatAnnotationString(
+                                type.getAnnotationsField(), currentPrintInvisibleSetting));
+                sb.append(" /*INFERENCE FAILED ");
```
wmdietl: What is the motivation for the comment syntax?
How about using:

```java
if (type.isUninferredTypeArgument()) {
    sb.append("INFERENCE FAILED for: ");
}
```

as the first thing and leaving all the rest unchanged?

smillst: We add `/*DECL*/` for declarations, so I was trying to be consistent with that.  It's kind of nice that it is a comment so that the printed type is still legal Java.  It's also good to make it clear that the `INFERENCE FAILED` is a comment and not part of the type.  I will change it to ` /*INFERENCE FAILED for: `.

wmdietl: Having the comment before the type instead of between the annotations and the type would make it more readable, at least to me. So I would rather add

```java
if (type.isUninferredTypeArgument()) {
    sb.append("/* INFERENCE FAILED for: */");
}
```

and then keep the output of the type unchanged.

smillst: So the output would be:
```
  /*INFERENCE FAILED for:*/ @Initialized @NonNull ? extends @Initialized @NonNull List<@Initialized @NonNull String>
```

PR #2875, opened 2019-11-08
Show more details about error and failure in jtreg tests.

xingweitian:
When running jtreg tests, if some tests fail, we can't see the error or failure messages, it is not helpful (especially when seeing travis's log).

I suggest to change from `-verbose:summary` to `-verbose:error,fail`, which will give us more details. But we may concern that the length of the log will increase.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -580,7 +580,7 @@ subprojects {
                                 "-dir:${projectDir}/jtreg",
                                 "-workDir:${jtregOutput}/${name}/work",
                                 "-reportDir:${jtregOutput}/${name}/report",
-                                "-verbose:summary",
+                                "-verbose:error,fail",
                                 "-javacoptions:-g",
                                 "-keywords:!ignore",
                                 "-samevm",


mernst:
@xingweitian Does the length of the log always increase, or increase only when there is a failure?  If the latter, I am not concerned.

xingweitian:
It will always increase, but not too much.

Before:
```
Passed: issue1133/Main.java
Passed: issue469/Main.java
Passed: multiplecheckers/NullnessInterning.java
```

After:
```
--------------------------------------------------
TEST: issue1133/Main.java
TEST RESULT: Passed. Compilation failed as expected
--------------------------------------------------
TEST: issue469/Main.java
TEST RESULT: Passed. Compilation successful
--------------------------------------------------
TEST: multiplecheckers/NullnessInterning.java
TEST RESULT: Passed. Compilation successful
--------------------------------------------------
```

When a test fails, we will get all the information we need, for example:
```
ACTION: compile -- Failed. Output does not match reference file: Issue347-con.out, line 1
REASON: User specified action: run compile/fail/ref=Issue347-con.out -XDrawDiagnostics -processor org.checkerframework.checker.nullness.NullnessChecker -Alint Issue347.java -AconcurrentSemantics 
TIME:   100.099 seconds
messages:
command: compile -XDrawDiagnostics -processor org.checkerframework.checker.nullness.NullnessChecker -Alint /checker-framework/checker/jtreg/nullness/Issue347.java -AconcurrentSemantics
reason: User specified action: run compile/fail/ref=Issue347-con.out -XDrawDiagnostics -processor org.checkerframework.checker.nullness.NullnessChecker -Alint Issue347.java -AconcurrentSemantics 
```

mernst:
Thanks for the information.  I'm not concerned about the increase in log size.

wmdietl:
Thanks!

Issue #1376, opened 2017-07-05
Better warning if a polymorphic qualifier is used on classes and fields


kelloggm:
Currently polymorphic qualifiers are only supported in methods. It is a reasonable intuition that polymorphic qualifiers written on classes and fields will behave similarly to polymorphic qualifiers written on methods, but that isn't the case. The documentation on polymorphic qualifiers should explicitly call out that this feature is _not_ supported.

kelloggm:
In addition, the framework itself should warn users who write polymorphic qualifiers on fields or classes that the feature is not supported, and should point back to this new documentation.

mernst:
This is already documented.  Section 23.2 "Qualifier polymorphism" (https://checkerframework.org/manual/#qualifier-polymorphism) says, in its second paragraph:

> It may not be used on a class or field.

Methods can have type parameters but fields cannot.  Therefore, it's an incorrect intuition to think that the syntax and semantics should be the same for fields and methods.

Nonetheless, programmers might have this misconception.  Your suggestion is a good one:  issue a clear error message if a user writes polymorphic qualifiers on a field.  It would also be nice to support polymorphic qualifiers on classes.

mernst:
Here are a few implementation suggestions if someone wants to address this.

Its related to issue #515.

However, the implementation would not use `@TargetLocations`, because we don't want to force every checker author who writes a checker to write.  Writing `@PolymorphicQualifier` should be enough.  To support this, the BaseTypeVisitor could have hard-coded logic to forbid annotations from being written in certain locations if the annotation is meta-annotated with `@Polymorphic`.

wmdietl:
This should also check for uses of polymorphic qualifiers in extends clauses.
For example `class C extends List<@PolyTainted String>` should be forbidden?

Also usages in field initializers, instance initializers, and static instance initializers should be forbidden.

PR #2883, opened 2019-11-12
Add support for class qualifier parameters

smillst:
The manual section added in the PR explains this feature.  The [Google Doc](https://docs.google.com/document/d/1nbJ4eTEs2kpysMbRyd0DJWH-BKuXawA77woncVgelGc/edit) gives more of the motivation for the semantics.

Merge with https://github.com/typetools/jdk/pull/58.


docs/manual/generics.tex
```diff
@@ -853,6 +856,95 @@
...
+
+If a class extends a \<@HasQualifierParameter> class (or implements a
+\<@HasQualifierParameter> interface), then that class must also be marked
+\<@HasQualifierParameter>.
+
+\subsectionAndLabel{Using class qualifier parameters in method signatures}{class-qual-param-method}
+
+In a class with a class qualifier parameter, a formal parameter type should
+sometimes have the same qualifier as the class qualifier parameter.  To express
+this, annotate both the receiver type and the parameter type with the
```
mernst: It seems that `@PolyX` in a method can have two different meanings:
 * method qualifier polymorphism
 * class implicit qualifier polymorphism

How does a programmer know what it means, and how does a programmer choose one of those meanings?
(I suspect that if the class has implicit qualifier polymorphism, then the meaning is always the latter, and no method qualifier polymorphism is possible.)
There should also be a mention of this and a cross-reference from the method qualifier polymorphism section.

docs/manual/generics.tex
```diff
@@ -853,6 +856,95 @@
   void m(@Nullable Object obj)
 \end{Verbatim}
 
+\sectionAndLabel{Classes with implicit qualifier parameters}{class-qualifier-polymorphism}
+
+The Checker Framework supports type \emph{qualifier} polymorphism for classes.  This is similar to
```
mernst: Could you be more specific than "similar to"?

Maybe giving a use case would be the best way to introduce the feature.

Is the key point of the feature to enable a field or method access to have the same type qualifier as the receiver argument (or at least to be controlled by the type of the receiver argument)?
That is, to make the type of `x.f` depend on the type of `x`?  And the same for a method parameter or return type?

Java generics can be viewed as giving a way to control method/field types by the type of the (type argument in the) receiver.

docs/manual/generics.tex
```diff
@@ -853,6 +856,95 @@
...
+this, annotate both the receiver type and the parameter type with the
+polymorphic qualifier for the type system. For example,
+\begin{Verbatim}
+@HasQualifierParameter(Tainted.class)
+class Buffer {
+  void append(@PolyTainted Buffer this, @PolyTainted String s) { ... }
+}
+\end{Verbatim}
+Because \<@PolyTainted> applies to a type with a qualifier parameter, then all
+other uses of \<@PolyTainted> are instantiated to the instantiation of
```
mernst: This is unclear.  I think you mean they depend on the actual qualifier argument (which is implicit) on the receiver.

docs/manual/generics.tex
smillst: I addressed this in the paragraph above this one.

docs/manual/generics.tex
smillst: I'm struggling to address this comment.  Would you please use [this doc](https://docs.google.com/document/d/1nbJ4eTEs2kpysMbRyd0DJWH-BKuXawA77woncVgelGc/edit?usp=sharing) and see if you can address this?

docs/manual/generics.tex
```diff
@@ -930,9 +939,10 @@
 At a field access where the declared type of the field has a polymorphic
 qualifier, that polymorphic qualifier is instantiated to the qualifier on the
 type of the receiver of the field access (or in the case of type variables, the
-qualifier on the upper bound).
+qualifier on the upper bound).  That is, the qualifier on \<myBuffer.field>
```
smillst: This doesn't match the example above.

smillst:
TODO: edit this section:
https://checkerframework.org/manual/#tainting-polymorphism-caution

abslamp:
The effect of `@HasQualifierParameter` seems not repeatable. Consider this hierarchy:

A foo(`@Poly`<sub>1</sub> SomeType someName)
|
B foo(`@Poly`<sub>2</sub> ...)
|
C foo(`@Poly`<sub>3</sub> ...)

What if the end user not want Poly<sub>2</sub> corresponds to Poly<sub>1</sub>, while Poly<sub>3</sub> still corresponds to Poly<sub>1</sub> (i.e, they are considered the same poly when performing substitution)? Is there a mechanism to allow the Poly substitution finds the `super.super` polymorphism annotations?

smillst:
@abslamp I don't understand your question.  Are A, B, and C classes and foo is a method in each class?  What's the meaning of `@Poly`<sub>1</sub>?

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -375,9 +375,111 @@ public void processClassTree(ClassTree classTree) {
...
 
+    /**
+     * Issues an error if {@code classTree} has polymorphic fields but is not annotated with
+     * {@code @HasQualifierParameter}. Always issue a warning if the type of a static field is
+     * annotated with a polymorphic qualifier.
+     *
+     * <p>Issues an error if (@code classTree} extends or implements a class/interface that has a
+     * qualifier parameter, but this class does not.
+     *
+     * @param classTree the ClassTree to check foor polymorphic fields
```
wmdietl: foor -> for

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -375,9 +375,111 @@ public void processClassTree(ClassTree classTree) {
...
+     */
+    private void checkQualifierParam(ClassTree classTree) {
```
wmdietl: Should the method be protected, to allow specific type systems to implement additional/other checks?

wmdietl: I would write out the name as `checkQualifierParameter`, to be consistent with the other new methods.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -375,9 +375,111 @@ public void processClassTree(ClassTree classTree) {
...
+                    atypeFactory.getQualifierHierarchy().getPolymorphicAnnotation(top);
+            if (poly != null) {
+                polys.add(poly);
+            }
+
+            if (!conflictingErrorReported
+                    && atypeFactory.hasExplicitQualifierParameterInHierarchy(classElement, top)
+                    && atypeFactory.hasExplicitNoQualifierParameterInHierarchy(classElement, top)) {
+                checker.report(Result.failure("conflicting.qual.param", top), classTree);
+                conflictingErrorReported = true;
```
wmdietl: Why does this report only the first conflicting top annotations? If there are multiple annotations on a class, wouldn't it be useful to highlight all top annotations that have conflicts?

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -375,9 +375,111 @@ public void processClassTree(ClassTree classTree) {
...
+                    }
+                }
+            }
+        }
+
+        for (Tree mem : classTree.getMembers()) {
+            if (mem.getKind() == Tree.Kind.VARIABLE) {
+                AnnotatedTypeMirror fieldType = atypeFactory.getAnnotatedType(mem);
+                Set<AnnotationMirror> polyAnnos;
+                if (ElementUtils.isStatic(TreeUtils.elementFromDeclaration((VariableTree) mem))) {
```
wmdietl: A comment about this logic would also be helpful.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1975,6 +2088,33 @@ protected boolean isTypeCastSafe(AnnotatedTypeMirror castType, AnnotatedTypeMirr
         }
     }
 
+    /**
+     * Return whether or not casting the exprType to castType is legal.
+     *
+     * @param castType an invariant type
+     * @param exprType type of the expressions that is cast may or may not be invariant
```
wmdietl: Should the `may or may not be invariant` be in parenthesis? Or make it two sentences to make it more readable.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1975,6 +2088,33 @@ protected boolean isTypeCastSafe(AnnotatedTypeMirror castType, AnnotatedTypeMirr
...
+     * @return whether or not casting the exprType to castType is legal.
+     */
+    private boolean isInvariantTypeCastSafe(
+            AnnotatedTypeMirror castType, AnnotatedTypeMirror exprType, AnnotationMirror top) {
+        if (!isTypeCastSafe(castType, exprType)) {
+            return false;
+        }
+        AnnotationMirror castTypeAnno = castType.getEffectiveAnnotationInHierarchy(top);
+        AnnotationMirror exprTypeAnno = exprType.getEffectiveAnnotationInHierarchy(top);
+        ;
```
wmdietl: Remove?

framework/src/main/java/org/checkerframework/common/basetype/messages.properties
```diff
@@ -96,6 +97,11 @@ contracts.conditional.postcondition.false.methodref.invalid=Conditional postcond
 lambda.unimplemented=This version of the Checker Framework does not type-check lambda expressions.
 methodref.inference.unimplemented=This version of the Checker Framework does not type-check method references with implicit type arguments.
 
+invalid.polymorphic.qualifier.use=Cannot use polymorphic qualifier on a field unless the class is annotated with @HasQualifierParameter
```
wmdietl: Wouldn't it be more useful to include the specific polymorphic qualifier? The field might be annotated with two different poly quals, and only one is invalid.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -3764,9 +3904,23 @@ protected boolean isAccessAllowed(
      */
     public boolean isValidUse(
             AnnotatedDeclaredType declarationType, AnnotatedDeclaredType useType, Tree tree) {
-        return atypeFactory
-                .getTypeHierarchy()
-                .isSubtype(useType.getErased(), declarationType.getErased());
+        Set<? extends AnnotationMirror> tops =
+                atypeFactory.getQualifierHierarchy().getTopAnnotations();
+        Set<AnnotationMirror> upperBounds =
```
wmdietl: Can you add a comment about this implementation? It's not clear to me why it needed to change.

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractValue.java
```diff
@@ -217,13 +217,37 @@ public V mostSpecific(@Nullable V other, @Nullable V backup) {
         return analysis.createAbstractValue(mostSpecific, mostSpecifTypeMirror);
     }
 
+    /** Computes the most specific annotations. */
     private class MostSpecificVisitor extends AnnotationSetAndTypeMirrorVisitor {
+        /**
+         * This is set to true, if there is no back up value, but one is required then the resulting
```
wmdietl: I do not understand this comment. Could you try rephrasing?

framework/src/main/java/org/checkerframework/framework/qual/HasQualifierParameter.java
```diff
@@ -0,0 +1,74 @@
...
+ * write it because the class hard-codes a Java basetype rather than taking a type parameter.
+ * Writing {@code HasQualifierParameter} on a package is the same as writing it on each class in
+ * that package.
+ *
+ * <p>Writing {@code @HasQualifierParameter} on a type declaration has two effects.
+ *
+ * <ol>
+ *   <li>Invariant subtyping is used for occurrences of the type: no two occurrences of the type
+ *       with different qualifiers have a subtyping relationship.
+ *   <li>The polymorphic qualifier is the default for all occurrences of that type in a method
```
wmdietl: Why only in method signatures? Why not in field types, local variables, etc.?

framework/src/main/java/org/checkerframework/framework/qual/HasQualifierParameter.java
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * <p>When {@code @HasQualifierParameter} is written on a package, it is equivalent to writing it on
+ * each class in that package with the same arguments, including classes in sub-packages. It can be
+ * disabled on a specific class by writing {@code @NoQualifierParameter} on that class.
+ *
+ * @see NoQualifierParameter
+ */
+@Target({ElementType.TYPE, ElementType.PACKAGE})
+@Documented
+public @interface HasQualifierParameter {
```
wmdietl: What retention should this annotation have?

framework/src/main/java/org/checkerframework/framework/qual/NoQualifierParameter.java
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * <p>One or more top qualifiers must be given for the hierarchies for which there are no qualifier
+ * parameters. This annotation may not be written on the same class as {@code HasQualifierParameter}
+ * for the same hierarchy.
+ *
+ * @see HasQualifierParameter
+ */
+@Target(ElementType.TYPE)
+@Documented
+public @interface NoQualifierParameter {
```
wmdietl: What retention should this annotation have?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -3309,6 +3317,154 @@ private void addOrMerge(Set<AnnotationMirror> results, AnnotationMirror annotati
         }
         return result;
     }
+    /**
```
wmdietl: Add empty line before and after this method, to separate it from others.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -3309,6 +3317,154 @@ private void addOrMerge(Set<AnnotationMirror> results, AnnotationMirror annotati
...
+     * Returns the set of top annotations representing all the hierarchies for which this element
+     * has a qualifier parameter.
+     *
+     * @param element Element to check
+     * @return the set of top annotations representing all the hierarchies for which this element
+     *     has a qualifier parameter
+     */
+    public Set<AnnotationMirror> getQualifierParameterHierarchies(@Nullable Element element) {
+        if (element == null || !ElementUtils.isTypeDeclaration(element)) {
```
wmdietl: Maybe explain this in the javadoc?
Why does it need to support null values?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -3309,6 +3317,154 @@ private void addOrMerge(Set<AnnotationMirror> results, AnnotationMirror annotati
...
+            return AnnotationUtils.createAnnotationSet();
```
wmdietl: Also, does it need to be a fresh, modifiable set? Or could this return an immutable singleton?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -3309,6 +3317,154 @@ private void addOrMerge(Set<AnnotationMirror> results, AnnotationMirror annotati
...
+     * @return the set of supported annotations with classes listed in the value element of an
+     *     annotation with class {@code annoClass} on the {@code element}. Returns an empty set if
+     *     {@code annoClass} is not written on {@code element} or {@code element} is null.
+     */
+    private Set<AnnotationMirror> getSupportedAnnotationsInElementAnnotation(
+            @Nullable Element element, Class<? extends Annotation> annoClass) {
+        Set<AnnotationMirror> found = AnnotationUtils.createAnnotationSet();
+
+        if (element == null) {
+            return found;
```
wmdietl: Return an immutable singleton instead of a fresh object for null?

abslamp:
@smillst Never mind. I overlooked that class with `@HasQualifierParameter` cannot be extended by a class without `@HasQualifierParameter`. As shown in your [doc](https://docs.google.com/document/d/1nbJ4eTEs2kpysMbRyd0DJWH-BKuXawA77woncVgelGc/edit):
 ```
@HasQualifierParameter
class Buffer {
  void append(@PolyTainted Buffer this, @PolyTainted String s) { ... }
}
class SubBuffer1 extends Buffer { ... } // error
```

File: framework/src/main/java/org/checkerframework/framework/qual/HasQualifierParameter.java (hunk 14 above)
Comment (smillst) [reply framework/src/main/java/org/checkerframework/framework/qual/HasQualifierParameter.java]: That's what we think will lead to the least errors, but the default may need to be changed.  An individual checker could also choose a different defaulting scheme.

**Comment by abslamp** [comment]
@smillst What is the expected behaviour when a class with `@NoQualifierParameter` extends a class with `@HasQualifierParameter`? I think there is no special handling with such case in the visitor.

**Comment by smillst** [comment]
> What is the expected behaviour when a class with `@NoQualifierParameter` extends a class with `@HasQualifierParameter`? 

@abslamp, in that case an error is issued, see the test case I just added, 67a555d3d.

**Comment by abslamp** [comment]
As Prof. @wmdietl point out, please consider this case:
```
class Pair<F, S> {
  F first;
  S second;
}

class Clazz<T> {
  Pair<@PolyTainted Object, T> f;
}

class Main {
  Pair<@Tainted Object, @PolyTainted String> foo(@Tainted Clazz<@PolyTainted String> c){
     return c.f;
  }
}
```
The result is:
```
Taint.java:17: error: [return.type.incompatible] incompatible types in return.
     return c.f;
             ^
  type of expression: @Tainted Pair<@Tainted Object, @Tainted String>
  method return type: @Tainted Pair<@Tainted Object, @PolyTainted String>
1 error
```
But actually in `Main::foo`, the second `@PolyTainted` in the returning type is undecidable, so it should remain `@PolyTainted` but not be resolved.

**Comment by abslamp** [comment]
Could a poly qualifier appear in extends/implements clause if the super class has `@HasQualifierParameter`? Should such use be forbidden?
(It make sense that some class extends a tainted buffer, but for poly...?)

Currently a poly qualifier in extends clause is compared with the annotation on class declaration, for this case:
```
@HasQualifierParameter(Tainted.class) class Super {}
@HasQualifierParameter(Tainted.class) class Sub extends @PolyTainted Super {}
```
The result shows:
```
Taint3.java:5: error: [declaration.inconsistent.with.extends.clause] Class @Tainted cannot extend @PolyTainted
@HasQualifierParameter(Tainted.class) class Sub extends @PolyTainted Super {}
                                                        ^
1 error
```

I think `@Tainted` is the default qualifier, so the result shows so.

**Comment by smillst** [comment]
@abslamp  In response to your first comment, it seems like resolving polymorphic annotations on field accesses will need a fix similar to https://github.com/typetools/checker-framework/pull/2876.

> Could a poly qualifier appear in extends/implements clause if the super class has @HasQualifierParameter?

That shouldn't be allowed.

File: checker/src/main/java/org/checkerframework/checker/tainting/jdk11.astub [hunk 21]
```diff
@@ -0,0 +1,7 @@
+import org.checkerframework.checker.tainting.qual.*;
+package java.lang;
+
+class Object {
+    // TODO: Move to typetools/jdk.
+    public @Untainted Object();
```
smillst, line 6: Move to typetools/jdk.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1,5 +1,7 @@
 package org.checkerframework.common.basetype;
 
+import static javax.tools.Diagnostic.Kind.ERROR;
```
smillst, line 3: Remove that.

framework/src/main/java/org/checkerframework/framework/qual/HasQualifierParameter.java
wmdietl: generics.tex says `Within a class with a qualifier parameter, the default qualifier for uses of that class is the polymorphic qualifier.` without a limitation to just method signatures. Which one is it?

framework/src/main/java/org/checkerframework/framework/qual/HasQualifierParameter.java
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * @see NoQualifierParameter
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE, ElementType.PACKAGE})
+public @interface HasQualifierParameter {
+
+    /**
+     * Class of the top qualifier for the hierarchy for which this class has a qualifier parameter.
```
wmdietl, line 72: It's an array, so maybe use plural?

framework/src/main/java/org/checkerframework/framework/source/AggregateChecker.java
```diff
@@ -1,5 +1,7 @@
 package org.checkerframework.framework.source;
 
+import static javax.tools.Diagnostic.Kind.ERROR;
```
wmdietl, line 3: The changes to this file seem unnecessary.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -3,6 +3,8 @@
 // The imports from com.sun are all @jdk.Exported and therefore somewhat safe to use.
 // Try to avoid using non-@jdk.Exported classes.
 
+import static javax.tools.Diagnostic.Kind.ERROR;
```
wmdietl, line 6: Also here I would continue using `Kind.ERROR` instead of the static import.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -3499,6 +3513,170 @@ private boolean isListForRepeatedAnnotationImplementation(DeclaredType annotatio
         return result;
     }
 
+    /**
+     * Whether or not the {@code annotatedTypeMirror} has an implicit qualifier parameter.
```
wmdietl, line 3517: The `implicit` in this comment is not meant as the opposite of `explicit` in the method names below?
Should it be `class qualifier parameter`? Or `implicit or explicit`?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -3499,6 +3513,170 @@ private boolean isListForRepeatedAnnotationImplementation(DeclaredType annotatio
...
+     * @return true if the type has a qualifier parameter
+     */
+    public boolean hasQualifierParameterInHierarchy(
+            AnnotatedTypeMirror annotatedTypeMirror, AnnotationMirror top) {
+        return AnnotationUtils.containsSame(
+                getQualifierParameterHierarchies(annotatedTypeMirror), top);
+    }
+
+    /**
+     * Whether or not the {@code element} has an implicit qualifier parameter.
```
wmdietl, line 3530: Same comment about `implicit`.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -3499,6 +3513,170 @@ private boolean isListForRepeatedAnnotationImplementation(DeclaredType annotatio
...
     }
 
+    /**
+     * Whether or not the {@code annotatedTypeMirror} has an implicit qualifier parameter.
+     *
+     * @param annotatedTypeMirror AnnotatedTypeMirror to check
+     * @param top the top of the hierarchy to check
+     * @return true if the type has a qualifier parameter
+     */
+    public boolean hasQualifierParameterInHierarchy(
```
wmdietl, line 3523: Many of these methods look like they could be static. Maybe we could move them to a new helper class instead of adding them to AnnotatedTypeFactory?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -3499,6 +3513,170 @@ private boolean isListForRepeatedAnnotationImplementation(DeclaredType annotatio
...
+    /**
+     * Returns the set of top annotations representing all the hierarchies for which this element
+     * has a qualifier parameter.
+     *
+     * @param element Element to check
+     * @return the set of top annotations representing all the hierarchies for which this element
+     *     has a qualifier parameter
+     */
+    public Set<AnnotationMirror> getQualifierParameterHierarchies(Element element) {
+        if (!ElementUtils.isTypeDeclaration(element)) {
```
wmdietl, line 3615: Could you use `TypeElement` to make this explicit and check the argument?

framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java
```diff
@@ -221,6 +221,11 @@ protected boolean isPrimarySubtype(
             AnnotatedTypeMirror subtype, AnnotatedTypeMirror supertype, boolean annosCanBeEmtpy) {
         final AnnotationMirror subtypeAnno = subtype.getAnnotationInHierarchy(currentTop);
         final AnnotationMirror supertypeAnno = supertype.getAnnotationInHierarchy(currentTop);
+        if (checker.getTypeFactory().hasQualifierParameterInHierarchy(supertype, currentTop)
+                && checker.getTypeFactory().hasQualifierParameterInHierarchy(subtype, currentTop)) {
+            return isAnnoSubtype(subtypeAnno, supertypeAnno, annosCanBeEmtpy)
```
wmdietl, line 226: Isn't this equivalent to `areSame`? Can you add a comment what is different, maybe whether one can be null?

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1558,6 +1559,61 @@ protected void applyInferredAnnotations(AnnotatedTypeMirror type, Value as) {
                 new DefaultInferredTypesApplier(getQualifierHierarchy(), this);
         applier.applyInferredType(type, as.getAnnotations(), as.getUnderlyingType());
     }
+    /**
```
wmdietl, line 1562: Empty line between methods.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1558,6 +1559,61 @@ protected void applyInferredAnnotations(AnnotatedTypeMirror type, Value as) {
...
+    }
+
+    /**
+     * Applies defaults for types in a class with an qualifier parameter.
+     *
+     * @param elt Element whose type is {@code type}
+     * @param type where the defaults are applied
+     */
+    protected void applyQualifierParameterDefaults(Element elt, AnnotatedTypeMirror type) {
+        if (elt == null) {
```
wmdietl, line 1579: Mark the element `@Nullable`.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1558,6 +1559,61 @@ protected void applyInferredAnnotations(AnnotatedTypeMirror type, Value as) {
...
+            return;
+        }
+        if (tops.isEmpty()) {
+            return;
+        }
+        new TypeAnnotator(this) {
+            @Override
+            public Void visitDeclared(AnnotatedDeclaredType type, Void aVoid) {
+                if (type.getUnderlyingType().asElement().equals(enclosingClass)) {
+                    for (AnnotationMirror top : tops) {
```
wmdietl, line 1607: Could this use `addMissingAnnotations` instead?

framework/src/main/java/org/checkerframework/framework/type/poly/AbstractQualifierPolymorphism.java
```diff
@@ -85,6 +86,13 @@
      */
     private Completer completer = new Completer();
 
+    /**
+     * Instantiations for poly qualifiers that are exactly the qualifier corresponding to poly
```
wmdietl, line 90: Can you reformulate this? I'm getting lost in what maps to what.

checker/tests/nullness/InvariantTypes.java
```diff
@@ -18,6 +18,7 @@ void testAsLocal() {
     }
 
     // Type for array creation is propagated from LHS
+    @SuppressWarnings("invalid.polymorphic.qualifier.use")
     @PolyNull Object[] po = new Object[5];
```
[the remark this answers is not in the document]
smillst: This test file is unrelated to my changes. It just uses a polymorphic qualifier on a field.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -366,9 +368,116 @@ public void processClassTree(ClassTree classTree) {
...
+        Set<AnnotationMirror> polys = AnnotationUtils.createAnnotationSet();
+
+        for (AnnotationMirror top : atypeFactory.getQualifierHierarchy().getTopAnnotations()) {
+            TypeElement classElement = TreeUtils.elementFromDeclaration(classTree);
+            if (classElement == null) {
+                continue;
+            }
+            AnnotationMirror poly =
+                    atypeFactory.getQualifierHierarchy().getPolymorphicAnnotation(top);
+            if (poly != null) {
```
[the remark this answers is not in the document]
smillst: No, you could have a type system without a polymorphic qualifier that still uses `@HasQualifierParameter`.

framework/src/main/java/org/checkerframework/common/basetype/messages.properties
```diff
@@ -24,6 +24,7 @@ type.invalid.conflicting.annos=invalid type: conflicting annotations %s in type
 type.invalid.too.few.annotations=invalid type: missing annotations %s in type "%s"
 type.invalid.annotations.on.use=invalid type: annotations %s conflict with declaration of type %s
 cast.unsafe=cast from "%s" to "%s" cannot be statically verified
+invariant.cast.unsafe=cast from "%s" to "%s" cannot be statically verified
```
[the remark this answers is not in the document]
smillst: It's reported as an error rather than a warning.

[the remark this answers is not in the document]
wmdietl: Then you don't need a separate error key. You could just only keep `cast.unsafe` and report it as an error or warning, if the message is the same in both cases.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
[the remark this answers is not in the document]
wmdietl: A comment would help make that case clearer.

checker/tests/nullness/InvariantTypes.java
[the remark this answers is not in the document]
wmdietl: And the polymorphic qualifier is now forbidden. Instead of suppressing this error, two valid tests that actually test what was intended to be tested would be more useful. Or keep this test and add the intended tests.

docs/manual/generics.tex
```diff
@@ -850,6 +854,174 @@
...
+\begin{tt}
+  @Tainted StringBuffer $\approx$ @Tainted Collection<@Tainted Character>\\
+  @Untainted StringBuffer $\approx$ @Untainted Collection<@Untainted Character>
+\end{tt}
+
+If two types have different qualifier arguments, they have no subtyping
+relationship.  (This is ``invariant subtyping'', also used by Java for
+generic classes.)  In particular, \<@Untainted StringBuffer> is not a
+subtype of \<@Tainted StringBuffer>; an attempt to cast between them, in
+either direction, will yield an \<invariant.cast.unsafe> error.
```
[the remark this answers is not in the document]
smillst: I don't understand this question.  Can you restate it?

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -366,9 +368,116 @@ public void processClassTree(ClassTree classTree) {
...
 
+        checkQualifierParameter(classTree);
+
         super.visitClass(classTree, null);
     }
 
+    /**
+     * Issues an error if {@code classTree} has polymorphic fields but is not annotated with
+     * {@code @HasQualifierParameter}. Always issue a warning if the type of a static field is
+     * annotated with a polymorphic qualifier.
```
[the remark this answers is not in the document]
smillst: Forbidding polymorphic qualifiers in those locations is outside the scope of this pull request.  Issue https://github.com/typetools/checker-framework/issues/1376 tracks that.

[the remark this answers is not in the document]
wmdietl: Ok, I added that comment there.

docs/manual/generics.tex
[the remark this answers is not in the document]
wmdietl: This states that `@Untainted StringBuffer` and `@TaintedStringBuffer` are not comparable.
How do I declare a field that can take either kind of `StringBuffer`, without also adding a qualifier parameter to that class?
How does local variable type refinement work? We usually start with the top type and then refine to subtypes. Now that the two possible types are incomparable, how are local variables handled?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
smillst: They can't be static; they all use `AnnotatedTypeFactory#isSupportedQualifier(AnnotationMirror)` which is not static.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
smillst: No. 
1. The check and cast would have to be added to every call site.
2. Currently only TypeElements can have a qualifier parameter, but at some point in the future, type parameters will be allowed to have qualifier parameters.  This would allow types with qualifier parameters to be type arguments.

docs/manual/generics.tex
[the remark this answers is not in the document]
smillst: >How do I declare a field that can take either kind of StringBuffer, without also adding a qualifier parameter to that class?

You can't.

>How does local variable type refinement work? We usually start with the top type and then refine to subtypes. Now that the two possible types are incomparable, how are local variables handled?

If the type of a local variable has a qualifier parameter, it can't be refined.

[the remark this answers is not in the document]
wmdietl: Mentioning these in the manual would be useful to understand how to use this feature.

PR #2884, opened 2019-11-12
Set enclosingType for the type of NewClassTrees (Fixes 2865)

### LINKED ISSUE #2865: Incorrect upper bound on type parameter with inner classes
Consider this test case:
```

public class TestTypeParamsCons<T extends @Nullable Object> {
    public class C {
        public C(T a) {}

        public void f(T a) {
            new C(a);
        }
    }
}
```

The constructor for `C` expects a `T`, which is exactly what is passed in `f`. These should both refer to the same type, `T extends @Nullable Object`. When I run the nullness checker with the command ` javac -processor nullness TestTypeParamsCons.java`, I get the following error:

```
./checker/tests/nullness/TestTypeParamsCons.java:8: error: [argument.type.incompatible] incompatible types in argument.
            new C(a);
                  ^
  found   : T extends @Initialized @Nullable Object
  required: T extends @Initialized @NonNull Object
1 error

```

smillst:
Also changed `TreeUtils#getReceiverTree` to return the receiver tree of a NewClassTree. 

Fixes #2865.

### DIFF
--- a/checker/tests/nullness/Issue2865.java
+++ b/checker/tests/nullness/Issue2865.java
@@ -17,4 +17,9 @@ public class Issue2865<T extends @Nullable Object> {
         s.new C(null);
         s.new C("");
     }
+
+    void test2(Issue2865<@Nullable String> s) {
+        s.new C(null);
+        s.new C("");
+    }
 }
--- a/checker/tests/nullness/Issue2865.java
+++ b/checker/tests/nullness/Issue2865.java
@@ -1,0 +1,20 @@
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public class Issue2865<T extends @Nullable Object> {
+    public class C {
+        public C(T a) {}
+
+        public void f(T a) {
+            new C(a);
+            // :: error: (argument.type.incompatible)
+            new C(null);
+        }
+    }
+
+    void test(Issue2865<@NonNull String> s) {
+        // :: error: (argument.type.incompatible)
+        s.new C(null);
+        s.new C("");
+    }
+}
--- a/framework/src/main/java/org/checkerframework/framework/util/typeinference/TypeArgInferenceUtil.java
+++ b/framework/src/main/java/org/checkerframework/framework/util/typeinference/TypeArgInferenceUtil.java
@@ -181,6 +181,10 @@ public class TypeArgInferenceUtil {
         } else if (assignmentContext instanceof NewClassTree) {
             // This need to be basically like MethodTree
             NewClassTree newClassTree = (NewClassTree) assignmentContext;
+            if (newClassTree.getEnclosingExpression() instanceof NewClassTree
+                    && (newClassTree.getEnclosingExpression() == path.getLeaf())) {
+                return null;
+            }
             ExecutableElement constructorElt = TreeUtils.constructor(newClassTree);
             AnnotatedTypeMirror receiver = atypeFactory.fromNewClass(newClassTree);
             res =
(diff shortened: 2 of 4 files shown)


javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -745,44 +745,44 @@ public static boolean isCompileTimeString(ExpressionTree node) {
...
 
-            if (receiver.getKind() == Tree.Kind.IDENTIFIER) {
-                // It's a method call "m(foo)" without an explicit receiver
+                if (receiver.getKind() == Tree.Kind.IDENTIFIER) {
+                    // It's a method call "m(foo)" without an explicit receiver
+                    return null;
+                } else if (receiver.getKind() == Tree.Kind.MEMBER_SELECT) {
+                    receiver = ((MemberSelectTree) receiver).getExpression();
+                } else {
+                    // Otherwise, e.g. a NEW_CLASS: nothing to do.
```
wmdietl: What situation is this? How can the overall `expression` be a method invocation and the receiver be a new class tree?
Is this for a constructor invocation? I would have expected `expression` to have type `NEW_CLASS` in that case.

javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -745,44 +745,44 @@ public static boolean isCompileTimeString(ExpressionTree node) {
...
+                break;
+            case NEW_CLASS:
+                receiver = ((NewClassTree) expression).getEnclosingExpression();
+                break;
+            case ARRAY_ACCESS:
+                receiver = ((ArrayAccessTree) expression).getExpression();
+                break;
+            case MEMBER_SELECT:
+                receiver = ((MemberSelectTree) expression).getExpression();
+                // Avoid int.class
```
wmdietl: Why is this only avoiding `int.class` and not also `Double.class` or other `.class` literals?

javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
smillst: I don't understand this comment either.  I think `IDENTIFIER` and `MEMBER_SELECT` are the only cases and this else branch isn't reached.  I'll make that change and see what happens.

checker/tests/nullness/Issue2865.java
```diff
@@ -0,0 +1,20 @@
...
+        public C(T a) {}
+
+        public void f(T a) {
+            new C(a);
+            // :: error: (argument.type.incompatible)
+            new C(null);
+        }
+    }
+
+    void test(Issue2865<@NonNull String> s) {
```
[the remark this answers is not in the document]
smillst: Done below.

framework/src/main/java/org/checkerframework/framework/util/typeinference/TypeArgInferenceUtil.java
```diff
@@ -181,6 +181,10 @@ public static AnnotatedTypeMirror assignedTo(AnnotatedTypeFactory atypeFactory,
         } else if (assignmentContext instanceof NewClassTree) {
             // This need to be basically like MethodTree
             NewClassTree newClassTree = (NewClassTree) assignmentContext;
+            if (newClassTree.getEnclosingExpression() instanceof NewClassTree
+                    && (newClassTree.getEnclosingExpression() == path.getLeaf())) {
```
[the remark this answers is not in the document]
smillst: It's equivalent to what happens in MethodInvocationTree.  I don't quite understand this myself.

javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
smillst: I don't know.  This comment was copied from the old code.

PR #2885, opened 2019-11-12
Correct rawtypes manual section.

smillst:


docs/manual/introduction.tex
```diff
@@ -689,6 +685,10 @@
   Passing this option uses a conservative value instead.
   See \href{https://github.com/typetools/checker-framework/issues/979}{Issue
   979}.
+\item \<-AignoreRawTypeArguments=false>
```
mernst: Please make the order of items in the manual be the same as the order of the variables in the source code, per the instructions on line 612.  That might mean moving text in the manual or fields in the source code.

Issue #1889, opened 2018-03-12
Too long commpilation


justmin:
Hello.
In our project we use java code generators which produce classes with hundreds of fields.
Consider synthetic example:
```
public class Test3 {
    private final Random fld0 = new Random();
    private final Random fld1 = new Random();
    private final Random fld2 = new Random();

    public int count(){
        int result = 0;
        result += fld0.nextInt();
        result += fld1.nextInt();
        result += fld2.nextInt();
        return result;
    }
}
```

I measured compilation time depends on fields count:

100.00 | 1.20
200.00 | 2.60
300.00 | 4.70
400.00 | 8.50
500.00 | 13.30
600.00 | 15.60
700.00 | 22.70
800.00 | 28.60
900.00 | 37.60
1000.00 | 44.60

for 2000 field compilation time was 200 sec with peak memory consumption 8 GB!!!
without CF compilation time was 3 sec.

We need to enable CF for hand-written code.
We generate code in annotation processors. So hand-written and auto-generated code are in the very same module. 
If it is not possible to speedup compilation may be it would be a good idea to have possibility to disable CF for some classes in module? 
For ex, `@SuppressWarning("nullness")` on a class should prevent CF to analyze methods' bodies.

wmdietl:
Thanks for sharing these numbers! We have a test that looks similar to yours:
https://github.com/typetools/checker-framework/blob/master/checker/jtreg/nullness/Issue1438.java
And at some point used it for performance profiling. Improving performance is high on our to-do list.

Would you be able to share your code generator? It would be helpful to experiment with larger instances.
Thanks!

wmdietl:
Regarding skipping classes: `@SuppressWarnings` performs all analyses and then suppresses error output. So it doesn't have a positive performance impact.

Did you try using `-AskipDefs`?
https://checkerframework.org/manual/#askipdefs
That option takes a pattern, so you might be able to suppress your generated code easily.
The skipDefs test happens earlier, so it should have a performance benefit. Let us know if you don't see an effect.

mernst:
> For ex, `@SuppressWarnings("nullness")` on a class should prevent CF to analyze methods' bodies.

Per the JDK's definition of `@SuppressWarnings`, it only suppresses warnings.
It is still necessary for the checker to perform some work, such as defaulting and implicit types.

[...]

PR #2887, opened 2019-11-12
Convert all type annotations to CheckerFrameworkAnnotationMirrors

smillst:


wmdietl:
@nhat-nguyen could you please have a look whether this has a positive impact on execution time (it uses interned strings) vs. memory consumption (it creates wrapper objects)?

nhat-nguyen:
> @nhat-nguyen could you please have a look whether this has a positive impact on execution time (it uses interned strings) vs. memory consumption (it creates wrapper objects)?

I'll make sure to profile this. This PR is very similar to my proposed idea of wrapping every annotation and perform a one-time pass over all annotations and assign each of them a hash value so that we don't have to repeatedly compare their string representation in AnnotationUtils.

@smillst I have one quick question, if now all type annotations are wrapped inside `CheckerFrameworkAnnotationMirror`, isn't the `instanceof` check in `areSameByName(AnnotationMirror a1, AnnotationMirror a2)` unnecessary? Or can there still be javac's annotation as well?

wmdietl:
> I'll make sure to profile this. This PR is very similar to my proposed idea of wrapping every annotation and perform a one-time pass over all annotations and assign each of them a hash value so that we don't have to repeatedly compare their string representation in AnnotationUtils.

Is there something in  `CheckerFrameworkAnnotationMirror` that could be improved to do that?

> @smillst I have one quick question, if now all type annotations are wrapped inside `CheckerFrameworkAnnotationMirror`, isn't the `instanceof` check in `areSameByName(AnnotationMirror a1, AnnotationMirror a2)` unnecessary? Or can there still be javac's annotation as well?

At the moment the parameters allow arbitrary `AnnotationMirror`s, so the instanceof is necessary. That probably isn't costing a lot.

smillst:
>@smillst I have one quick question, if now all type annotations are wrapped inside CheckerFrameworkAnnotationMirror, isn't the instanceof check in areSameByName(AnnotationMirror a1, AnnotationMirror a2) unnecessary? Or can there still be javac's annotation as well?

All _type_ annotations are converted to CheckerFrameworkAnnotationMirrors, but _declaration_ annotations are not.  So the util methods in AnnotationUtils still need to expect both.

smonteiro:
When will this be merged into master ? 
It looks better than it was.
Once this is merged I have an improvement to offer on this issue of *areSame*().

mernst:
@nhat-nguyen
> I'll make sure to profile this.

What was the result of your profiling?

Azure Pipelines shows a slowdown.
Either this branch degrades performance, or Azure Pipelines is not useful for profiling.
It would be useful to know which is the case.
Here is timing information from Azure Pipelines:
In this branch, the daikon_jdk8 job takes 47:33 :
https://dev.azure.com/typetools/checker-framework/_build/results?buildId=2046&view=logs&jobId=458a3e2d-91a0-5e1b-309f-8f24fd6215b9
In master, the daikon_jdk8 job takes 41:12 :
https://dev.azure.com/typetools/checker-framework/_build/results?buildId=2131&view=logs&jobId=458a3e2d-91a0-5e1b-309f-8f24fd6215b9

nhat-nguyen:
@wmdietl @mernst I'm sorry for the delay in getting the results. I just ran this branch 15 times using `time -v` with https://github.com/typetools/checker-framework/issues/1889 (1600 fields) as a micro-benchmark and nullnessfbctest (type-checking daikon took too long on my machine to run) and had the following results (attached).

[baseline.txt](https://github.com/typetools/checker-framework/files/3909174/baseline.txt)
[wrapper.txt](https://github.com/typetools/checker-framework/files/3909175/wrapper.txt)

The timings on my laptop look a bit odd since the wrapper is actually somehow a bit faster. But the memory usage seems pretty much the same.

wmdietl:
@nhat-nguyen Thanks for profiling!

Performance numbers seem to not show performance improvements and I dislike making `CheckerFrameworkAnnotationMirror` public.

@smonteiro You mentioned possible improvements that build on this change.
Was that for the changes that we already merged or for these remaining changes?
Can you give some more details what this would enable?
Thanks!

smonteiro:
@wmdietl  Definitely I can try to explain briefly. on a  separate issue so I do not cloud this one. It would be applicable either way but I did not want to interfere with this effort.. Let me do that.

smonteiro:
See #2958 for another approach to this.

smillst:
Since this doesn't improve performance, I'm closing it.  @nhat-nguyen, thanks for testing this!

PR #2889, opened 2019-11-13
Remove unnecessary ci-info invocations.

wmdietl:


wmdietl:
These invocations don't appear in the other scripts and I propose removing them for consistency in how these scripts are used.

PR #2891, opened 2019-11-13
Small fixes found by running the Interning Checker

smillst:
These are the true positives that the Interning Checker found.

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/i18nformatter/I18nFormatUtil.java
+++ b/checker/src/main/java/org/checkerframework/checker/i18nformatter/I18nFormatUtil.java
@@ -373,7 +373,7 @@ public class I18nFormatUtil {
 
             // Try trimmed lowercase.
             String ls = s.trim().toLowerCase(Locale.ROOT);
-            if (ls.equals(s)) {
+            if (!ls.equals(s)) {
                 for (int i = 0; i < list.length; ++i) {
                     if (ls.equals(list[i])) {
                         return i;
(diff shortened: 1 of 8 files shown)


framework/src/main/java/org/checkerframework/framework/util/typeinference/constraint/AFReducingVisitor.java
```diff
@@ -309,7 +309,7 @@ public Void visitIntersection_Declared(
         // at least one of the intersection bound types must be convertible to the param type
         final AnnotatedDeclaredType subtypeAsParam =
                 AnnotatedTypes.castedAsSuper(typeFactory, subtype, supertype);
-        if (subtypeAsParam != null && !subtypeAsParam.equals(subtype)) {
+        if (subtypeAsParam != null && !subtypeAsParam.equals(supertype)) {
```
smillst: This one wasn't actually found by the Interning Checker.  (I found it via an IntelliJ analysis.)

checker/src/main/java/org/checkerframework/checker/i18nformatter/I18nFormatUtil.java
```diff
@@ -373,7 +373,7 @@ private static final int findKeyword(String s, String[] list) {
 
             // Try trimmed lowercase.
             String ls = s.trim().toLowerCase(Locale.ROOT);
-            if (ls != s) {
+            if (!ls.equals(s)) {
```
wmdietl: I don't think this is needed. Both `trim` and `toLowerCase` return the receiver object if nothing needs to change. So comparing for reference equality is enough. Or why would this be needed?
Maybe instead add a comment explaining this.

PR #2892, opened 2019-11-13
Cache hash code for FlowExpressions.LocalVariable 

smillst:


dataflow/src/main/java/org/checkerframework/dataflow/analysis/FlowExpressions.java
```diff
@@ -721,17 +721,36 @@ public boolean isUnmodifiableByOtherCode() {
...
         }
 
-        public LocalVariable(Element elem) {
-            super(ElementUtils.getType(elem));
-            this.element = elem;
+        /** Create a {@link LocalVariable} from {@code element}. */
+        public LocalVariable(Element element) {
+            super(ElementUtils.getType(element));
+            this.element = element;
+            VarSymbol vs = (VarSymbol) this.element;
```
wmdietl: Instead of duplicating this logic, could you please introduce a helper method or helper constructor that extracts this?

dataflow/src/main/java/org/checkerframework/dataflow/analysis/FlowExpressions.java
```diff
@@ -721,17 +721,36 @@ public boolean isUnmodifiableByOtherCode() {
         }
     }
 
+    /** FlowExpression representing a local variable. */
     public static class LocalVariable extends Receiver {
```
wmdietl: Why do this only for this kind of Receiver? There are several other `hashCode` methods in the Receiver hierarchy that also look expensive. (I assume that this PR is a performance optimization.)
Would an alternative be to introduce the `hashcode` field in `Receiver`, override `hashCode` in `Receiver` to return that field, and rename the existing `hashCode` methods to `hashCodeCompute` or some such and use that method in the constructor?

smillst: Yes, this PR is a possible performance optimization. (I should have included that in the description.)  I'm not sure if the performance improvement of this is worth the increase in memory.  Is this something @nhat-nguyen and you would like to look into?  If not, then we can just close this pull request.

PR #2895, opened 2019-11-14
Only parse JDK 11 if source version is 11.

### LINKED ISSUE #2878: version 3.0.0 doesn't support cross compilation
When running on JDK 11, version 3.0.0 of the checker framework crashes when targeting the Java 8 language level, i.e. `--release 8`:

```
$ java -fullversion
openjdk full version "11.0.5+10"
$ ./checker-framework-3.0.0/checker/bin/javac -processor Nullness --release 8 T.java
error: Exception while parsing jdk11/src/java.base/share/classes/java/lang/Object.java
  Compilation unit: T.java
  Last visited tree at line 4 column 1:
  class FakeName implements Name {
  Exception: org.checkerframework.javacutil.BugInCF: Couldn't find since element in java.lang.Deprecated; Stack trace: org.checkerframework.javacutil.AnnotationBuilder.findElement(AnnotationBuilder.java:504)
  org.checkerframework.framework.stub.StubParser.handleExpr(StubParser.java:1710)
  org.checkerframework.framework.stub.StubParser.getAnnotation(StubParser.java:1531)
  org.checkerframework.framework.stub.StubParser.annotateDecl(StubParser.java:1139)
  org.checkerframework.framework.stub.StubParser.processCallableDeclaration(StubParser.java:727)
  org.checkerframework.framework.stub.StubParser.processTypeDecl(StubParser.java:569)
  org.checkerframework.framework.stub.StubParser.processCompilationUnit(StubParser.java:505)
  org.checkerframework.framework.stub.StubParser.processStubUnit(StubParser.java:488)
  org.checkerframework.framework.stub.StubParser.process(StubParser.java:482)
  org.checkerframework.framework.stub.StubParser.parse(StubParser.java:442)
  org.checkerframework.framework.stub.StubParser.parseJdkFileAsStub(StubParser.java:415)
  org.checkerframework.framework.stub.StubTypes.parseJarEntry(StubTypes.java:352)
  org.checkerframework.framework.stub.StubTypes.parseEnclosingClass(StubTypes.java:287)
```

I think one of the issues here is that `CheckerMain` sets up the stubs for the system classes based on `PluginUtil.getJreVersion`, i.e. the host runtime version, instead of the version being targeted by the current compilation. It should use bootclasspath stubs corresponding to the `--release` or `-target` flags, not the host runtime version.

smillst:
Also add jtreg test that are only run with JDK 11.

This fixes #2878, but there is no way to pass the annotated JDK if the JDK version is 11, but `--release 8` is used.  In a separate pull request, I'll work on parsing JDK 11 when using `--release 8`.


framework/src/main/java/org/checkerframework/framework/stub/StubTypes.java
```diff
@@ -59,19 +59,33 @@
...
+    /** Should the JDK be parsed? */
+    private final boolean shouldParseJdk;
 
     /** Creates a stub type. */
     public StubTypes(AnnotatedTypeFactory factory) {
         this.factory = factory;
         this.typesFromStubFiles = new HashMap<>();
         this.declAnnosFromStubFiles = new HashMap<>();
         this.parsing = false;
+        String release = PluginUtil.releaseValue(factory.getProcessingEnv());
```
wmdietl: Instead of returning a `@Nullable` string here and handling the fall-back at the call-site, how about moving that into `PluginUtil.releaseValue` and have that return an int?

smillst: If I make that change, the method won't return the value of `release`, so I'll need to rename it. What should the new name be?

wmdietl: How about `getReleaseVersion` to be symmetric with `getJreVersion`?

wmdietl: We discussed this and decided to keep the current name.

framework/src/main/java/org/checkerframework/framework/stub/StubTypes.java
```diff
@@ -427,6 +444,8 @@ private void prepJdk11FromFile(URL resourceURL) {
                     walk.filter(p -> Files.isRegularFile(p) && p.toString().endsWith(".java"))
                             .collect(Collectors.toList());
             for (Path path : paths) {
+                // TODO: revert.
```
wmdietl: Highlighting that there are two `TODO: revert` comments.

framework/src/main/java/org/checkerframework/framework/stub/StubTypes.java
wmdietl: Although something like `getReleaseValue` would be probably closer to the naming of other methods in `PluginUtil`...

smillst: Good point, I make this change.

framework/src/main/java/org/checkerframework/framework/stub/StubTypes.java
smillst: Thanks for catching this! I had merged in a different branch that had changes that should not be merged.  I've reverted them now.

PR #2899, opened 2019-11-16
Try Gradle 6.0

wmdietl:


smonteiro:
How is this coming along in the builds servers ?  I could not get 6.0.1 anywhere near even doing a local build in a Mac. It got so lost in the weeds of self configuration.

xingweitian:
> How is this coming along in the builds servers ? I could not get 6.0.1 anywhere near even doing a local build in a Mac. It got so lost in the weeds of self configuration.

Did you try `./gradlew wrapper --gradle-version=6.0.1`? It works well on my computer.

xingweitian:
I did some investigation about this PR.

On my own [branch](https://github.com/xingweitian/checker-framework/tree/gradle-6.0), most of the warnings have been fixed.

There is one left due to `biz.aQute.bnd`, which is the dependency of `checker-qual`.

Now we are using `biz.aQute.bnd:biz.aQute.bnd.gradle:4.1.0`, the warning is fixed by `4.4.0`, which is not available on maven yet. Related issue: https://github.com/bndtools/bnd/issues/3494

I think we should wait for the release of `4.4.0` and I will make a new PR to merge into this one to fix all the warnings.

smonteiro:
That is EXACTLY the version giving me trouble.  And the biz.aQute.bnd, message was one of the headaches.
So, this is not ready then.

smillst:
Closed in favor of #2959.

PR #2902, opened 2019-11-18
Fixes #1929: Determine whether toArray() is used with ArrayLen(0).

### LINKED ISSUE #1929: Empty array behavior with CollectionToArrayHeuristics
https://github.com/typetools/checker-framework/blob/6deea6321310e7b2d8eed197ef71e02f64230d81/checker/src/org/checkerframework/checker/nullness/CollectionToArrayHeuristics.java#L132

[CollectionToArrayHeuristics](https://checkerframework.org/api/org/checkerframework/checker/nullness/CollectionToArrayHeuristics.html) describes several heuristics for evaluating Collection.toArray. [Collection.toArray()](https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#toArray-T:A-) will possibly set the array element after the last value from the collection to null. This shouldn't actually be a problem in this case, because the EMPTY_MYCLASS_ARRAY is not large enough to be used directly, so Collection.toArray() will allocate a new array of just the right size.

```
private static final MyClass[] EMPTY_MYCLASS_ARRAY = new MyClass[0];
...
return result.values().toArray(EMPTY_MYCLASS_ARRAY);
```

Which does not satisfy 2. array creation tree of size 0, e.g. c.toArray(new String[0]), perhaps because the empty array is static/final? The following new allocation does:
```
return result.values().toArray(new MyClass[result.values().size()]);
```

We're not sure if this discrepancy is because CollectionToArrayHeuristics is operating directly on the abstract syntax tree level, rather than on the type level, or some other reason entirely. Perhaps the heuristic could be integrated with the [constant value checker](https://checkerframework.org/manual/#constant-value-checker) such as the ArrayLen annotation. For example if the checker framework inferred EMPTY_MYCLASS_ARRAY as ArrayLen(0) and CollectionToArrayHeuristics could check for ArrayLen(0). We aren't sure if this is a valid approach, or if our understanding of the behavior is correct. It would be great as a feature request if this is the case. Thanks!

xingweitian:
Contributors: @wmdietl @txiang61 @xingweitian.


checker/src/main/java/org/checkerframework/checker/nullness/CollectionToArrayHeuristics.java
```diff
@@ -32,6 +40,7 @@
  *         <li value="2">array creation tree of size 0, e.g. {@code c.toArray(new String[0])}, or
  *         <li value="3">array creation tree of the collection size method invocation {@code
  *             c.toArray(new String[c.size()])}
+ *         <li value="4">field access where the declaration has a @ArrayLen(0) annotation
```
wmdietl: Use `{@code ..}`

xingweitian: Don't worry about it, I will clean it soon.

checker/src/main/java/org/checkerframework/checker/nullness/CollectionToArrayHeuristics.java
```diff
@@ -67,11 +76,12 @@ public CollectionToArrayHeuristics(
     public void handle(MethodInvocationTree tree, AnnotatedExecutableType method) {
         if (TreeUtils.isMethodInvocation(tree, collectionToArrayE, processingEnv)) {
             assert !tree.getArguments().isEmpty() : tree;
-            Tree argument = tree.getArguments().get(0);
+            ExpressionTree argument = tree.getArguments().get(0);
             boolean argIsArrayCreation =
```
wmdietl: This lost a few changes I made in https://github.com/typetools/checker-framework/commit/f33adcdeaea4eb98105e104dacd20fff6627452c

wmdietl: I successfully cherry-picked that changeset.

wmdietl:
I'll add a command-line flag and update the manual.

smillst:
@xingweitian Please do not force push to branches under review.  It corrupts the review history. (https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests#about-pull-requests.) We always squash and merge pull request, so the commit history in the branch does not matter.

wmdietl:
@smillst I had incorrectly pulled in opprop/master into this PR so there were many unrelated changes.
Instead of force pushing, how should we fix the PR in the future?

smillst:
@wmdietl You can revert the merge commit.  Or you could create a new PR with a new branch.

xingweitian:
@smillst Thanks for the advice!

xingweitian:
@wmdietl Travis has been passed.

checker/src/main/java/org/checkerframework/checker/nullness/CollectionToArrayHeuristics.java
```diff
@@ -35,26 +43,48 @@
  *       </ol>
  * </ol>
  *
- * Note: The nullness of the returned array doesn't depend on the passed array nullness.
+ * <p>Additionally, when the lint option {@link NullnessChecker#LINT_TRUSTARRAYLENZERO} is provided,
+ * field accesses where the field declaration has a {@code @ArrayLen(0)} annotation are considered
```
mernst: Can you please clarify this documentation?
 * I'm not sure where the field access will be.  Is the field access the receiver of `toArray`, or is the field access the argument to `toArray`?
 * Please be more specific than "are considered".  What is the behavior?

checker/src/main/java/org/checkerframework/checker/nullness/CollectionToArrayHeuristics.java
```diff
@@ -35,26 +43,48 @@
...
+ * when determining whether a call to {@link Collection#toArray(Object[]) Collection.toArray(T[])}
+ * will return an array with non-null components. This trusts the {@code @ArrayLen(0)} annotation,
+ * but does not verify it. Run the Constant Value Checker to verify that annotation.
+ *
+ * <p>Note: The nullness of the returned array doesn't depend on the passed array nullness.
```
mernst: Is this a fact about nullness, or is it a fact about the heuristics defined in this routine?  Please clarify.

docs/manual/nullness-checker.tex
```diff
@@ -456,6 +465,44 @@
...
+returned array component is of type \<@NonNull> if the following conditions
+hold:
+
+\begin{itemize}
+\item The receiver collection type argument is \<@NonNull>, and
+\item The passed array size is less than the collection size. Here are heuristics to
+       handle the most common cases:
+
+  \begin{itemize}
+    \item an empty array initializer, e.g. \<c.toArray(new String[] \{\})>,
```
mernst: Why not combine the first two items to just be "the argument has length 0"?  Then elsewhere you can list what arrays are known to have size 0:  two specific patterns, plus a special case for those annotated as `@ArrayLen(0)`.

checker/src/main/java/org/checkerframework/checker/nullness/CollectionToArrayHeuristics.java
wmdietl: I've reformulated that sentence.

checker/src/main/java/org/checkerframework/checker/nullness/CollectionToArrayHeuristics.java
wmdietl: This is a fact about `toArray`. It's from the original description.

wmdietl:
I agree that discovering this new flag might be tricky. I hope that the errors are relatively local and then a search for `toArray` in the manual will lead one to this new section.

I tried implementing your suggestion, but at the moment we have no framework support to do this. I didn't find a simple way to change the error message depending on what shape some subtree has.

If you prefer, I can remove the longer discussion from the manual and just keep the short sentence about the lint option.
However, merging this PR would be very useful and we can work on discoverability separately, which e.g. would also help with keyfor errors.

checker/src/main/java/org/checkerframework/checker/nullness/CollectionToArrayHeuristics.java
```diff
@@ -35,26 +43,53 @@
  *       </ol>
  * </ol>
  *
- * Note: The nullness of the returned array doesn't depend on the passed array nullness.
+ * <p>Additionally, when the lint option {@link NullnessChecker#LINT_TRUSTARRAYLENZERO} is provided,
+ * a call to {@link Collection#toArray(Object[]) Collection.toArray(T[])} will return an array with
```
mernst: The option doesn't change what is returned.  It changes the checker's estimate of the return type.  Could you clarify the text?

checker/src/main/java/org/checkerframework/checker/nullness/CollectionToArrayHeuristics.java
mernst: Could you clarify it?  If it's unclear wherever the original description is, then please clarify that too.  (And please link to it.)

checker/src/main/java/org/checkerframework/checker/nullness/NullnessChecker.java
```diff
@@ -46,6 +47,15 @@
     /** Default for {@link #LINT_REDUNDANTNULLCOMPARISON}. */
     public static final boolean LINT_DEFAULT_REDUNDANTNULLCOMPARISON = false;
 
+    /**
+     * Should we unsoundly trust {@code @ArrayLen(0)} annotations to improve handling of {@link
```
mernst: Please don't use first person for a description of the checker.

docs/manual/nullness-checker.tex
```diff
@@ -456,6 +465,44 @@
 % TODO: explain more aspects, give more examples.
 
 
+\subsectionAndLabel{Nullness and conversions from collections to arrays}{nullness-collection-toarray}
+
+% Keep this in sync with the javadoc of org.checkerframework.checker.nullness.CollectionToArrayHeuristics
```
mernst: Javadoc, not "javadoc".

Also, the text is out of sync.  Please sync them up.

mernst:
A reminder about this comment:

> there is no note in the code to keep the text consistent with the manual,

checker/src/main/java/org/checkerframework/checker/nullness/CollectionToArrayHeuristics.java
```diff
@@ -28,33 +36,73 @@
...
+ *
+ * <p>Note: The nullness of the returned array doesn't depend on the passed array nullness. This is
+ * a fact about {@link Collection#toArray(Object[]) Collection.toArray(T[])}, not a limitation of
+ * this heuristic.
+ *
+ * @checker_framework.manual #nullness-collection-toarray Nullness and conversions from collections
+ *     to arrays
+ * @checker_framework.manual #constant-value-checker Constant Value Checker
  */
+// Note: The https://checkerframework.org/manual/#nullness-collection-toarray section in the manual,
```
mernst: I checked the two texts, and they contained some subtle differences in wording.  I think this will be difficult to maintain.  I suggest that you remove the documentation from here and instead just reference the manual.

checker/src/main/java/org/checkerframework/checker/nullness/CollectionToArrayHeuristics.java
```diff
@@ -81,6 +131,35 @@ public void handle(MethodInvocationTree tree, AnnotatedExecutableType method) {
         }
     }
 
+    /**
+     * Determine whether the argument is a field access expression of which the declaration has a
```
mernst: This is a description of the implementation strategy, but it is not a specification of the method.

I think the specification is:
 * return true if the given expression is an empty array (has type `@ArrayLen(0)`), using heuristics

Is that correct?
The current implementation only works for field accesses, but that is not worth putting in the Javadoc nor embedding in the name.  These changes will make the code clearer.

Also, the addition of this method makes the documentation and name of `isHandledArrayCreation` inaccurate.  It no longer handles all the heuristics.  (It does handle just some of the 0-length arrays.)  Why isn't the code `trustArrayLenZero && isArrayLenZeroFieldAccess(argument)` in the body of `isHandledArrayCreation`?  If you don't want to do that, then it seems that restructuring and renaming it is necessary.

checker/src/main/java/org/checkerframework/checker/nullness/CollectionToArrayHeuristics.java
```diff
@@ -28,33 +36,73 @@
...
- *         <li value="2">array creation tree of size 0, e.g. {@code c.toArray(new String[0])}, or
- *         <li value="3">array creation tree of the collection size method invocation {@code
- *             c.toArray(new String[c.size()])}
+ *         <li value="1">the argument has length 0:
+ *             <ol>
+ *               <li value="1">an empty array initializer, e.g. {@code c.toArray(new String[] {})},
+ *                   or
+ *               <li value="2">array creation tree of size 0, e.g. {@code c.toArray(new String[0])}.
+ *             </ol>
+ *         <li value="2">array creation tree with a collection {@code size()} method invocation as
```
mernst: The change in numbering makes the comments in `isHandledArrayCreation` inaccurate.  Please update them along with this update.  And add numbering for the code that calls `isArrayLenZeroFieldAccess`, though folding it into `isHandledArrayCreation` is probably a better idea, see below.

docs/manual/nullness-checker.tex
```diff
@@ -97,6 +97,15 @@
   See Section~\ref{nullness-arrays} for a discussion.
   % TODO: this option is temporary and the sound option
   % should become the default.
+
+\item
+  \label{nullness-lint-trustarraylenzero}%
+  If you supply the \code{-Alint=trustArrayLenZero} command-line option, then a
+  call to \code{Collection.toArray} will return an array with a non-null
```
mernst: "will return" implies a run-time behavior.  I think you mean that the type-checker will treat the type of an expression to be non-null.

Also, this sentence is long and hard to follow.  Maybe shorten it to "the checker will trust `@ArrayLen(0)` annotations" and the cross-reference.

mernst:
Thanks!  This is looking good.  I like the improvement and the tests.
Can you please address my comments about naming and code/documentation structure?

wmdietl:
@xingweitian Could you try addressing these comments?

xingweitian:
> @xingweitian Could you try addressing these comments?

Sure, I will try it.

xingweitian:
@wmdietl Could you please restart `all_tests_jdk11`?

changelog.txt
```diff
@@ -1,5 +1,11 @@
 Version 3.3.0, April 1, 2020
 
+New command-line options:
+  The Nullness Checker now supports the option -Alint=trustArrayLenZero to
```
mernst: Please start the entry with the command-line option, to make it easier to read.

changelog.txt
```diff
@@ -1,5 +1,11 @@
...
+  trust @ArrayLen(0) annotations when determining the type of
+  Collections.toArray. See the new manual section
+  http://checkerframework.org/manual/#nullness-collection-toarray
```
mernst: I would omit this for brevity.  (Because it should be self-evident to read the manual for any change that is briefly described here.)

checker/src/main/java/org/checkerframework/checker/nullness/CollectionToArrayHeuristics.java
```diff
@@ -159,6 +160,35 @@ private boolean isHandledArrayCreation(Tree argument, String receiver) {
         return false;
     }
 
+    /**
+     * Returns true if the argument is a field access expression, where the field has declared type
```
mernst: This documentation does not seem accurate.
 * It's not any field access expression, only the special "len" field access applied to an array.
 * It's not the field that has declared type `@ArrayLen(0)`, but the array whose field is being accessed.

wmdietl: There was some misunderstanding here and @xingweitian changes in response to this comment made the javadoc incorrect.
There is no special `len` field and this method didn't special-case such a field.
It is a field access where the declared type has `@ArrayLen(0)`. I'm not sure what "the array whose field is being accessed" should mean, it probably refers to the non-existent `len` field.

@xingweitian please undo your javadoc changes.

mernst:
@xingweitian CI is failing because there is a `fix-issue-1929` branch at https://github.com/xingweitian/checker-framework-inference/branches but it is not up to date.

Does this pull request need to be accompanied by a pull request into checker-framework-inference?
If so, make it from that branch (after pulling master into that branch).
If not, delete that branch.

xingweitian:
@mernst Thanks for pointing it out! I have updated that branch, now the CI should pass.
I am using a checker-framework forked from opprop, so every time I open a PR at typetools/checker-framework, I also need to maintain the corresponding branch of checker-framework-inference in my repo. Otherwise, the CI will try to build my checker-framework-inference which is also forked from opprop.

Issue #2905, opened 2019-11-19
Code duplicated for client invocation of new Range() with min and max


smonteiro:
The following pattern:
```
 long annoMinVal = Collections.min(values);
 long annoMaxVal = Collections.max(values);
 ... new Range(annoMinVal, annoMaxVal) ...
```
is seen  duplicated in many places.
That is troubling for multiple reasons: duplication, several local variables every time, and requiring two passes through every list of values to evaluate min and max.

A new convenience  constructor would greatly shrink client code 
``` 
    /**
     * Convenience construtor for often invoked pattern for Integer collections
     * @param collection the min and max value found in it will be used as from and to values in the range 
     */
    public Range(Collection<Integer> collection){
    	this(Collections.min(collection),Collections.max(collection));
    }
```
turning those invocations into simply new Range(values) 
and later one could tune the evaluation of min and max within the convenience constructor to only take one pass through the values and even  taking advantage of collections that are already sorted so first and last are min and max.

PR #2906, opened 2019-11-19
#2905 Issue2905 use new Range() constructors

smonteiro:
Minor improvement on client code that uses new Range() by use of new convenience constructors

### DIFF
--- a/framework/src/main/java/org/checkerframework/common/value/util/Range.java
+++ b/framework/src/main/java/org/checkerframework/common/value/util/Range.java
@@ -70,10 +70,11 @@ public class Range {
     /**
      * Create a Range from a collection of Integers.
      *
-     * @param collection from which min and max values will be used as from and to values in the
-     *     range
+     * @param list from which min and max values will be used as from and to values in the range
      */
-    public Range(Collection<Integer> collection) {
+    // The formal parameter type cannot be Collection<Integer> because that would erase to the same
+    // raw types as the Range(Collection<Long>) constructor.
+    public Range(List<Integer> list) {
         this(Collections.min(list), Collections.max(list));
     }
 
--- a/framework/src/main/java/org/checkerframework/common/value/util/Range.java
+++ b/framework/src/main/java/org/checkerframework/common/value/util/Range.java
@@ -64,7 +64,7 @@ public class Range {
     public Range(Collection<Long> collection){
         this(Collections.min(collection),Collections.max(collection));
     }
-    
+
     /**
      * Convenience construtor for often invoked pattern for Integer collections
      * @param collection the min and max value found in it will be used as from and to values in the range
@@ -72,7 +72,7 @@ public class Range {
     public Range(List<Integer> collection){
         this(Collections.min(collection),Collections.max(collection));
     }
-    
+
     /**
      * Constructs a range with its bounds specified by two parameters, {@code from} and {@code to}.
      *
@@ -93,7 +93,6 @@ public class Range {
         this.to = Long.MIN_VALUE;
     }
 
-   
     /**
      * Returns a range with its bounds specified by two parameters, {@code from} and {@code to}. If
      * {@code from} is greater than {@code to}, returns {@link #NOTHING}.
--- a/framework/src/main/java/org/checkerframework/common/value/util/Range.java
+++ b/framework/src/main/java/org/checkerframework/common/value/util/Range.java
@@ -58,20 +58,22 @@ public class Range {
     public static final Range NOTHING = new Range();
 
     /**
-     * Convenience construtor for a any collection of Long elements.
+     * Create a Range from a collection of Longs.
      *
-     * @param collection from which min and max values will be used as from and to values in the range.
+     * @param collection from which min and max values will be used as from and to values in the
+     *     range
      */
     public Range(Collection<Long> collection) {
         this(Collections.min(collection), Collections.max(collection));
     }
 
     /**
-     * Convenience construtor for List of Integer.
+     * Create a Range from a collection of Integers.
      *
-     * @param list from which min and max values will be used as from and to values in the range.
+     * @param collection from which min and max values will be used as from and to values in the
+     *     range
      */
-    public Range(List<Integer> list) {
+    public Range(Collection<Integer> collection) {
         this(Collections.min(list), Collections.max(list));
     }
 
--- a/framework/src/main/java/org/checkerframework/common/value/util/Range.java
+++ b/framework/src/main/java/org/checkerframework/common/value/util/Range.java
@@ -1065,7 +1065,7 @@ public class Range {
      * bound and upper bound.
      */
     public boolean isWithin(long lb, long ub) {
-        return lb <= from && to <= ub;
+        return from >= lb && to <= ub;
     }
 
     /**
--- a/framework/src/main/java/org/checkerframework/common/value/util/Range.java
+++ b/framework/src/main/java/org/checkerframework/common/value/util/Range.java
@@ -94,7 +94,7 @@ public class Range {
     }
 
    
-  /**
+    /**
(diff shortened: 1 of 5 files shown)


framework/src/main/java/org/checkerframework/common/value/util/Range.java
```diff
@@ -56,6 +57,26 @@
     /** The empty range. */
     public static final Range NOTHING = new Range();
 
+    /**
+     * Convenience construtor for often invoked pattern for Long collections
```
mernst: Please use correct grammar:  sentences end with a period.
Please say *what* the pattern is.  This documentation doesn't do that.
Please don't make value judgments like "often invoked" without evidence.  Just say what it is.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
```diff
@@ -56,6 +57,26 @@
...
+     *
+     * @param collection the min and max value found in it will be used as from and to values in the
```
mernst: Please say what the argument is.  This mentions some facts about elements in the collection but doesn't directly document the collection.  (The summary documentation on line 61 should do so too.)

mernst:
@smonteiro If your code is incorrectly formatted, you can fix it by running
```./gradlew reformat```

The [error message](https://dev.azure.com/typetools/checker-framework/_build/results?buildId=1793&view=logs&jobId=dff7b79b-ed6e-5b72-51c5-d49537c74f23&taskId=7dabb28f-aafd-5998-fbcd-46ef74c7ca82&lineStart=476&lineEnd=477&colStart=1&colEnd=1) told you to run this command, and the [developer manual](https://rawgit.com/typetools/checker-framework/master/docs/developer/developer-manual.html#Gradle_build_scripts) also mentions this.

File: framework/src/main/java/org/checkerframework/common/value/util/Range.java [hunk 3]
```diff
@@ -56,6 +57,26 @@
...
+        this(Collections.min(collection), Collections.max(collection));
+    }
+
+    /**
+     * Convenience construtor for often invoked pattern for Integer collections
+     *
+     * @param collection the min and max value found in it will be used as from and to values in the
+     *     range
+     */
+    public Range(List<Integer> collection) {
```
wmdietl: Is there a reason why this constructor uses `List` instead of `Collection`?

framework/src/main/java/org/checkerframework/common/value/util/Range.java
```diff
@@ -56,6 +57,26 @@
...
     public static final Range NOTHING = new Range();
 
+    /**
+     * Convenience construtor for often invoked pattern for Long collections
+     *
+     * @param collection the min and max value found in it will be used as from and to values in the
+     *     range
+     */
+    public Range(Collection<Long> collection) {
+        this(Collections.min(collection), Collections.max(collection));
```
wmdietl: In #2905 you also mention the two passes through the collection as a concern.
Is it worth documenting this future improvement?

framework/src/main/java/org/checkerframework/common/value/util/Range.java
smonteiro: The client code to Range() uses List<> on types Long and Integer.
Choosing arguments of Collection<Long> and Collection<Integer> does not compile as their signatures collide.
*<Long> will not match  *<Integer>
Collection<Long> does not match List<Integer> or anything *<Integer>
List<Integer> will not match List<Long>

Ideally we will coalesce the client code to focus on a convenience and a designated representation that makes this less of an eyesore.  I am taking small steps as I am new to contributing to this project.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
smonteiro: Is it worth documenting for a tiny improvement or is it worth implementing and submitting with a ticket for a pull request ?  I am new to the group so I still don't know which has been teh preferred way.

We could spend pages on the analysis if we try to document it.
On the other hand, the code is veryshort... if passed a non-sorted  Collection<Number>...
//  traverses the list once, updates min and max only when needed.
//        long min=values.get(0).longValue();
//        long max=min;
//        for (Number value : values) {
//        	long current=value.longValue();
//        	if (min>current)
//        		min=current;
//        	if (max<current)
//        		max=current;
//        }
//        return new Range(min,max);
Total combined assignments to variables min and max = N 
One internal iterator setup. 

And local evaluations of binary comparison of two long variables, with one wasted iteration on the very first element.

Using Math.min() and max(), as in min=Math.min(min,current) are very legible but they do mean 2(N+1) assignments to variables min and max and function calls and returns for each element in the list.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
smonteiro: Hi Michael.
Your help is most appreciated.
I will fix the syntax on a newer commit.
I am still getting the hang of the ticketing system here.
I tried documenting the "pattern" in the issue: 
https://github.com/typetools/checker-framework/issues/2905
The desire is to no longer have that in the client code using Range().

Documenting in javadoc what one wants to remove/go away/retire from the code seems odd.

The refactoring here was a "pull up" from the client code using Range() into convenience constructors in Range().

The examples of the repeated code being absorbed are found in the pull request.

I hope this helps.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
mernst: In my opinion, we don't need to document the potential improvement.  I suspect these lists will not be extremely long in general, and I suspect that if we find them to be a bottleneck, we will be able to figure out the improvement without a hint in a comment.

smonteiro: One can not declare both convenience signatures to be Collection<Long> and Collection<Integer> because of generic type erasure.  Type is discarded on generics so the two signatures end up the same Range(Collection<?>)  and therefore the compiler does not allow them to co-exist.

smonteiro: The build shows the problem: 
/__w/1/s/framework/src/main/java/org/checkerframework/common/value/util/Range.java:76: error: name clash: Range(Collection<Integer>) and Range(Collection<Long>) have the same erasure
    public Range(Collection<Integer> collection) {
           ^

framework/src/main/java/org/checkerframework/common/value/util/Range.java
```diff
@@ -58,20 +58,22 @@
...
     /**
-     * Convenience construtor for List of Integer.
+     * Create a Range from a collection of Integers.
      *
-     * @param list from which min and max values will be used as from and to values in the range.
+     * @param collection from which min and max values will be used as from and to values in the
+     *     range
      */
-    public Range(List<Integer> list) {
+    public Range(Collection<Integer> collection) {
```
smonteiro: One can not declare both convenience signatures to be Collection and Collection because of generic type erasure. Type is discarded on generics so the two signatures end up the same Range(Collection<?>) and therefore the compiler does not allow them to co-exist. Therefore on for Collection<Long> and one for List<Integer>

mernst: You are right; my mistake.  (It's too bad that both cannot be the more general type.)

smonteiro:
@mernst  What happens now with this pull request ? Who merges it ?

smillst:
>What happens now with this pull request ? Who merges it ?

Since @wmdietl left comments, he will need to approve this pull request.  That's why @mernst assigned this pull request to him.  Once that is done and assuming the tests are still passing, one of the maintainer will merge the pull request.  Thanks!

PR #2907, opened 2019-11-19
Actually run the Interning Checker.

wmdietl:
Follow-up to #2864


smillst:
Thanks!  I think this got lost in a merge in the original pull request.  We can run the Interning Checker on all subprojects, so I changed this a little bit.

PR #2911, opened 2019-11-19
Infer return types of abstract methods from the return types of their implementations in WPI

kelloggm:


framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInference.java
```diff
@@ -120,17 +122,20 @@ void updateInferredFieldType(
 
     /**
      * Updates the return type of the method {@code methodTree} based on {@code returnedExpression}.
+     * May also strengthen the return types of any abstract methods that this method overrides.
      *
      * @param returnedExpression the node that contains the expression returned
      * @param classSymbol the symbol of the class that contains the method
      * @param methodTree the tree of the method whose return type may be updated
+     * @param overriddenMethods the methods that the given method return overrides
```
mernst: The parameter's type is a map type, but this documentation doesn't explain what it maps from and to.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
```diff
@@ -451,6 +456,7 @@ public void updateInferredFieldType(
      * @param retNode the node that contains the expression returned
      * @param classSymbol the symbol of the class that contains the method
      * @param methodTree the tree of the method whose return type may be updated
+     * @param overriddenMethods the methods that this method overrides
```
mernst: Same comment about documentation.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
```diff
@@ -477,6 +485,38 @@ public void updateInferredMethodReturnType(
         AnnotatedTypeMirror rhsATM = atf.getAnnotatedType(retNode.getTree().getExpression());
         helper.updateAnnotationSetInScene(
                 method.returnType, atf, jaifPath, rhsATM, lhsATM, TypeUseLocation.RETURN);
+
+        // Now, update overridden methods based on the implementation we just saw.
+        // If the overridden method is abstract, then WPI cannot update it directly,
```
mernst: Clarify "directly".  I think you mean that there are no calls that resolve to it.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
```diff
@@ -477,6 +485,38 @@ public void updateInferredMethodReturnType(
...
+        // so updating based on implementations makes sense (cf. how method parameters
```
mernst: Please be more specific.  Are you referring to `updateInferredMethodParameterTypes`, or to something else?

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
```diff
@@ -477,6 +485,38 @@ public void updateInferredMethodReturnType(
...
+        // are updated).
+        for (Map.Entry<AnnotatedDeclaredType, ExecutableElement> pair :
+                overriddenMethods.entrySet()) {
+
+            AnnotatedExecutableType overriddenMethod =
+                    AnnotatedTypes.asMemberOf(
+                            atf.getProcessingEnv().getTypeUtils(),
+                            atf,
+                            pair.getKey(),
```
mernst: Rather than making multiple calls to `getKey` and `getValue`, make one call and assign to a local variable with a mnemonic name.  That will be easier to understand than making the reader remember what `pair` means.

framework/tests/whole-program-inference/non-annotated/MethodDefinedInSupertype.java
```diff
@@ -0,0 +1,14 @@
+import testlib.wholeprograminference.qual.*;
+
+abstract class MethodDefinedInSupertype {
+
+    void test() {
+        MethodDefinedInSupertype i = new MethodOverrideInSubtype();
```
mernst: What is the purpose of this line of the test?

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInference.java
```diff
@@ -120,17 +122,21 @@ void updateInferredFieldType(
 
     /**
      * Updates the return type of the method {@code methodTree} based on {@code returnedExpression}.
+     * May also strengthen the return types of any abstract methods that this method overrides.
```
mernst: This weakens them rather than strengthens them, I believe.  I believe that all changes made by WPI are inital assignment of a qualifier, or weakenings (LUBs).  (Is that necessary/desirable?  Why not propagate in the other direction as well and do GLBs?

Or do you mean compared to the default qualifier?  Even then, the initial assignment could be a weakening or a strengthening.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
```diff
@@ -452,6 +460,44 @@ public void updateInferredMethodReturnType(
...
+        // the inference procedure for method parameters: both are updated based
+        // only on the implementations (in this case) or call-sites (for method
+        // parameters) that are available to WPI.
+        for (Map.Entry<AnnotatedDeclaredType, ExecutableElement> pair :
+                overriddenMethods.entrySet()) {
+
+            AnnotatedDeclaredType superclassType = pair.getKey();
+            ExecutableElement superclassMethodDef = pair.getValue();
+
+            AnnotatedExecutableType overriddenMethod =
```
mernst: It's confusing to use two different prefixes "superclass" and "overridden" for the same concept.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
```diff
@@ -452,6 +460,44 @@ public void updateInferredMethodReturnType(
...
+                    AnnotatedTypes.asMemberOf(
+                            atf.getProcessingEnv().getTypeUtils(),
+                            atf,
+                            superclassType,
+                            superclassMethodDef);
+
+            if (overriddenMethod.getElement().getModifiers().contains(Modifier.ABSTRACT)) {
```
mernst: Why aren't facts about return values propagated to all overridden methods?

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
```diff
@@ -452,6 +460,44 @@ public void updateInferredMethodReturnType(
         AnnotatedTypeMirror rhsATM = atf.getAnnotatedType(retNode.getTree().getExpression());
         helper.updateAnnotationSetInScene(
                 method.returnType, atf, jaifPath, rhsATM, lhsATM, TypeUseLocation.RETURN);
+
+        // Now, update overridden methods based on the implementation we just saw.
```
mernst: Performing this for each return statement is inefficient.  It would be enough to propagate the information once at the end of WPI, for the final return type.
In addition, it's inefficient to propagate (elsewhere) when the overriding relationship is discovered.  More information may be discovered but not propagated.  This could lead to an extra iteration of WPI, which would be very costly.

It would be better to record overriding relationships as they are discovered, then do all propagation related to them at the end of WPI.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
kelloggm: You're right about `superclassMethodDef`. However, I think `superclassType` is a reasonable name for that `AnnotatedDeclaredType`, which is the type of the superclass which originally defined the method in question. To make it more clear, I've changed the name to `superclassDecl`.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
kelloggm: This seems like premature optimization to me. WPI is already inefficient. I will add a comment with your concern about performance, so that if performance becomes a problem for WPI we can revisit it.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
kelloggm: Because when I wrote this I was only interested in abstract methods. I've fixed it.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInference.java
mernst: I had this same question when re-reading the diffs.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
```diff
@@ -409,6 +412,8 @@ public void updateInferredFieldType(
      * Updates the return type of the method methodTree in the Scene of the class with symbol
      * classSymbol.
      *
+     * <p>Maybe also update the return types of abstract methods in the chain of overridden methods.
```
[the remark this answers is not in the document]
mernst: This comment has not been addressed.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
```diff
@@ -452,6 +459,52 @@ public void updateInferredMethodReturnType(
...
+        //
+        // TODO: Performing this for each return statement is inefficient. It would
+        //  be enough to propagate the information once at the end of WPI, for the
+        //  final return type.
+        //
+        // In addition, it's inefficient to propagate (elsewhere) when the overriding
+        // relationship is discovered. More information may be discovered but not propagated.
+        // This could lead to an extra iteration of WPI, which would be very costly.
+        //
+        // It would be better to record overriding relationships as they are discovered,
```
mernst: Rather than writing a TODO comment, could you just implement this change?  It's true that fixpoint iteration is expensive, but anything we can do to avoid a extra iteration is valuable.

kelloggm: Mike and I agreed that the approach described in this todo comment is no more efficient, and possibly less efficient, than the approach I actually implemented; neither can lead to an extra WPI pass (since the values aren't used until the next WPI pass anyway), and the approach Mike suggested requires another AST pass.

kelloggm:
Mike and I discussed this PR in detail on 1/13/20. In our discussion, we identified all the ways that method parameters and return types could reasonably be inferred, and identified the parts of WPI that handle all except two ("the green arrows" in our diagram):
* the type of each parameter of a method defined in a superclass should be at least the least upper bound of that parameter in each of its implementations, and
* the return type of a method defined in a superclass should be at least the least upper bound of the return types of each of its implementations

The former of the two is, in fact, handled by the aptly-named `updateFromOverride` method in `WholeProgramInference` (I don't remember how we missed that). The latter is handled by this PR. @mernst, can you re-review this PR with that in mind?

mernst:
@smillst Please merge when appropriate.

PR #2912, opened 2019-11-19
How to suppress initialization warnings; fixes #2840

### LINKED ISSUE #2840: Out-of-date documentation about the initialization checker
The introduction to the section about the initialization checker in the manual says:

> If you wish to disable the initialization checks, pass the command-line argument -AsuppressWarnings=uninitialized when running the Nullness Checker.

I don't think this is true, and I don't think it has been true for a very long time. From the changelog:

> Version 1.2.4, 15 Dec 2011
> Nullness Checker:
> - The -Alint=uninitialized command-line option now defaults to off, and
  applies only to initialization of primitive and `@Nullable` fields.  It is
  not possible to disable, from the command line, the check that all
  `@NonNull` fields are initialized.  Such warnings must be suppressed
  explicitly, for example by using `@SuppressWarnings`.

mernst:


mernst:
@smillst @wmdietl Could one of you review this?

docs/manual/nullness-checker.tex
```diff
@@ -1883,25 +1880,32 @@
 \end{sloppypar}
 
 
-\subsectionAndLabel{More details about initialization checking}{initialization-checking}
-
+\subsectionAndLabel{Suppressing warnings}{initialization-checking-suppressing-warnings}
 
-\paragraphAndLabel{Suppressing warnings}{initialization-checking-suppressing-warnings}
+To suppress most warnings related partially-initialized values, use the
```
wmdietl: Should it be `related to par...`?

PR #2913, opened 2019-11-20
Create a task that runs javadoc with -Xdoclint:all

smillst:


build.gradle
```diff
@@ -329,38 +329,51 @@ configurations {
...
+ * passed. Defaults to true.
+ * @return the new task
+ */
+def createJavadocTask(taskName, memberLevel, requireJavadoc = true) {
+    tasks.create(name: taskName, type: Javadoc) {
+        description = 'Ensures that Java elements have Javadoc documentation.'
+        destinationDir = file("${rootDir}/docs/tmpapi")
+        destinationDir.mkdirs()
+        subprojects.forEach {
+            if(!it.name.startsWith("checker-qual")) {
```
smillst: By not running javadoc on checker-qual, `javadocDoclintAll` now works with both JDK 8 and JDK 11.

build.gradle
```diff
@@ -329,38 +329,51 @@ configurations {
...
+ *
+ * @param taskName the name of the task to create
+ * @param memberLevel the JavadocMemberLevel to use
+ * @param requireJavadoc whether or not to run the RequireJavadoc doclet. If false -Xdoclint:all is
+ * passed. Defaults to true.
+ * @return the new task
+ */
+def createJavadocTask(taskName, memberLevel, requireJavadoc = true) {
+    tasks.create(name: taskName, type: Javadoc) {
+        description = 'Ensures that Java elements have Javadoc documentation.'
```
mernst: I would take the description as a command-line argument; otherwise all the tasks have the identical documentation and it's hard for a user to understand the differences among them.

build.gradle
smillst: The contents of checker-qual contains source code copied from the checker and framework projects, so it's not necessary to check that project separately.

PR #2918, opened 2019-11-21
Enforce static finals

smonteiro:
The constants were already in place. They needed to be used.

Also it becomes evident that isWithinInteger() was being reinvented in client invocations when all that was needed was for it to be public.

### DIFF
--- a/framework/src/main/java/org/checkerframework/common/value/util/Range.java
+++ b/framework/src/main/java/org/checkerframework/common/value/util/Range.java
@@ -1081,7 +1081,7 @@ public class Range {
             // This bound is adequate to guarantee no overflow when using long to evaluate.
             // Long.MIN_VALUE >> 1 + 1 = -4611686018427387903
             // Long.MAX_VALUE >> 1 = 4611686018427387903
-            return to - from + 1 > value;
+            return width() > value;
         } else {
             return BigInteger.valueOf(to)
                             .subtract(BigInteger.valueOf(from))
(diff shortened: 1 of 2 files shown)


framework/src/main/java/org/checkerframework/common/value/util/Range.java
```diff
@@ -151,29 +163,40 @@ public int hashCode() {
         return Objects.hash(from, to);
     }
 
+    /**
+     * Compare two ranges in a type safe manner for equality without incurring the cost of an
+     * instanceof check such as equals(Object) does.
```
wmdietl: How about using this method from `equals(Object)`, to make sure the logic is consistent and there is no duplication?

smonteiro: Yes. That change is in my local working directory.  
I am having a hell of a time getting gradlew reformat or simply format to work so my pace is slower than I wanted.

mernst: Go ahead and push.  I can reformat for you.

mernst: Your problems suggest that the formatter scripts don't run well under Windows.  Sorry about that.

I don't know whether the problem is in the Checker Framework's `build.gradle` file or in the [run-google-java-format](https://github.com/plume-lib/run-google-java-format) project itself.  If you could let us know the symptoms, perhaps we could help.

Another alternative is to install Windows Subsystem for Linux and run the reformat command there.  I know that works (though it's a rather heavyweight solution and doesn't help other Windows users).

smonteiro: I committed the change. If it fails for format, please reformat.

PR #2921, opened 2019-11-21
evaluate castType.getKind() only once and then compare many times

smonteiro:


framework/src/main/java/org/checkerframework/common/value/ValueVisitor.java
```diff
@@ -29,7 +29,7 @@
 import org.checkerframework.framework.type.AnnotatedTypeMirror;
 import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedDeclaredType;
 import org.checkerframework.framework.type.visitor.AnnotatedTypeScanner;
-import org.checkerframework.javacutil.AnnotationUtils;
+import static org.checkerframework.javacutil.AnnotationUtils.*;
```
wmdietl: In general we avoid wildcard imports and there are very few static imports.
Please only import the methods you actually use.

smonteiro: Can do. Next commit will have it split.

smonteiro:
This is as far as I want to  go with this pull request. No more commits.

mernst:
As a general rule, we like to make every pull request a logical unit.  This one combines unrelated changes:  using static imports, and avoiding recomputation of `getKind`.  I will merge it, but in the future could you make a separate pull request for each?  I realize that's a touch more work, but it makes each pull request (and commit in the history) easier to understand.
Thanks again.

smonteiro:
I agree completely. I have no issue complying with that as that is what I request also when I do reviews for merge :)
I appreciate the flexibility while I get my "programming legs" with the cadence. 

About the gradlew reformat with Windows: should I file that as an issue in this repository with details or what do you recommend ?

PR #2924, opened 2019-11-22
Introduce factory methods and use them via automated refactoring. 

smonteiro:
This is a follow up to the addition of convenience constructors to Range().

(The changes are better viewed in split view)

By introducing factory methods and refactor-renaming them all to create(*) the interface footprint to classes using it is greatly reduced. Now one passes whatever one has at hand to a **Range.create**(*)  and let the pertinent implementation handle the processing.

int, long, Integer, Long, BigInteger, List, Set ... the  `create(*,*)` and `create(*)` signatures handle them all.

There is a unified implementation for the create signature for Collection which does only one pass through the collection gathering min and max; and it is works for all the 8 previous  places that needed such construct.

The public factory methods validate arguments defensively while private ones do not validate arguments by design. See JavaDov within Range now for better insight.

Once this is approved, the next step (2) is to remove the now redundant client code that attempted to convert lists to create Range. _I think it belongs in this same pull request._

NOT DOING THIS PARAGRAPH ANYMORE:  Once step (2)  is approved,  an improvement to the evaluation of the hash is in order and can finally be done in one place instead of several. _And that would conclude this pull request._

### DIFF
--- a/framework/src/main/java/org/checkerframework/common/value/util/Range.java
+++ b/framework/src/main/java/org/checkerframework/common/value/util/Range.java
@@ -119,6 +119,7 @@ public class Range {
      *
      * @param from the lower bound (inclusive)
      * @param to the upper bound (inclusive)
+     * @return newly created Range or EVERYTHING
      */
     private static Range createOrNothing(long from, long to) {
         return createOrElse(from, to, NOTHING);
@@ -315,12 +316,22 @@ public class Range {
         return createOrElse((byte) this.from, (byte) this.to, BYTE_EVERYTHING);
     }
 
-    /** Returns true if the element is contained in this range. */
+    /**
+     * Returns true if the element is contained in this range.
+     *
+     * @param element the value to seek
+     * @return true or false
+     */
     public boolean contains(long element) {
         return from <= element && element <= to;
     }
 
-    /** Returns true if the element is contained in this range. */
+    /**
+     * Returns true if the element is contained in this range.
+     *
+     * @param other the range to seek
+     * @return true or false
+     */
     public boolean contains(Range other) {
         return other.isWithin(from, to);
     }
@@ -1078,6 +1089,10 @@ public class Range {
     /**
      * Determines if this range is completely contained in the range specified by the given lower
      * bound inclusive and upper bound inclusive.
+     *
+     * @param lb lower bound
+     * @param ub upper bound
+     * @return true or false
      */
     public boolean isWithin(long lb, long ub) {
         return lb <= from && to <= ub;
--- a/framework/src/main/java/org/checkerframework/common/value/util/Range.java
+++ b/framework/src/main/java/org/checkerframework/common/value/util/Range.java
@@ -373,12 +373,20 @@ public class Range {
         return createOrNothing(resultFrom, resultTo);
     }
 
-    /** @return the range with the lowest to and from values of this range and the passed range */
+    /** 
+     * 
+     * @param other
+     * @return the range with the lowest to and from values of this range and the passed range.
+     */
     public Range min(Range other) {
         return create(Math.min(this.from, other.from), Math.min(this.to, other.to));
     }
 
-    /** @return the range with the highest to and from values of this range and the passed range */
+    /** 
+     * 
+     * @param other
+     * @return the range with the highest to and from values of this range and the passed range.
+     */
     public Range max(Range other) {
         return create(Math.max(this.from, other.from), Math.max(this.to, other.to));
     }
@@ -1160,10 +1168,10 @@ public class Range {
      *
      * <p>Overflow and underflow here are any violations of {@code from<=to}.
      *
-     * @param from
-     * @param to
-     * @param underflow
-     * @return
+     * @param from given from value. 
+     * @param to given to value.
+     * @param underflow.
+     * @return a Range instance.
      */
     private static Range createOrElse(long from, long to, Range underflow) {
         if (from <= to) return new Range(from, to);
--- a/framework/src/main/java/org/checkerframework/common/value/util/Range.java
+++ b/framework/src/main/java/org/checkerframework/common/value/util/Range.java
@@ -65,10 +65,8 @@ public class Range {
     /**
      * Constructs a range with its bounds specified by two parameters, {@code from} and {@code to}.
      *
-     * <p>Private constructor does no validation of arguments so special instances can be created
-     * through it.
-     *
-     * <p>Example {@code #NOTHING}.
+     * <p>This is a private constructor that does no validation of arguments, so special instances
+     * (e.g., {@link #NOTHING}) can be created through it.
      *
      * @param from the lower bound (inclusive)
      * @param to the upper bound (inclusive)
@@ -80,12 +78,11 @@ public class Range {
 
     /**
      * Constructs a range with its bounds specified by two parameters, {@code from} and {@code to}.
-     *
-     * <p>Public factory method validates that {@code from <= to}
+     * Requires {@code from <= to}.
      *
      * @param from the lower bound (inclusive)
      * @param to the upper bound (inclusive)
-     * @return a valid Range for {@code from <= to}, otherwise IllegalArgumentException.
+     * @return the Range [from..to]
      */
     public static Range create(long from, long to) {
         if (!(from <= to)) {
@@ -95,18 +92,17 @@ public class Range {
     }
 
     /**
-     * Create a Range from a collection of Number, Longs, Integer.
+     * Create a Range from a collection of Numbers.
      *
-     * <p>This is a public convenience factory method that builds using the public factory for a
-     * range.
-     *
-     * @param values collection from which min and max values will be used as from and to values in
-     *     the range.
-     * @return NOTHING on empty collection; a valid Range on any collection.
+     * @param values collection whose min and max values will be used as the range's from and to
+     *     values
+     * @return a range that encompasses all the argument's values ({@link #NOTHING} if the argument
+     *     is an empty collection)
      */
     public static Range create(Collection<? extends Number> values) {
-        if (values.isEmpty()) return NOTHING;
-        // traverses the list once for min and max
+        if (values.isEmpty()) {
+            return NOTHING;
+        }
         long min = values.iterator().next().longValue();
         long max = min;
         for (Number value : values) {
--- a/framework/src/main/java/org/checkerframework/common/value/util/Range.java
+++ b/framework/src/main/java/org/checkerframework/common/value/util/Range.java
@@ -6,6 +6,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 /**
  * The Range class models a 64-bit two's-complement integral interval, such as all integers between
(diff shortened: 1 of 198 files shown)


framework/src/main/java/org/checkerframework/common/value/util/Range.java
```diff
@@ -1153,7 +1169,7 @@ public boolean isWithinInteger() {
      * @param bigTo the upper bound of the BigInteger range
      * @return a range with Long type bounds converted from the BigInteger range
      */
-    private static Range bigRangeToLongRange(BigInteger bigFrom, BigInteger bigTo) {
+    private static Range create(BigInteger bigFrom, BigInteger bigTo) {
```
wmdietl: Move this up to the other `create` methods?

framework/src/main/java/org/checkerframework/common/value/util/Range.java
```diff
@@ -40,65 +42,81 @@
...
-    private Range() {
-        this.from = Long.MAX_VALUE;
-        this.to = Long.MIN_VALUE;
+    /**
+     * Create a Range from a collection of Number, Longs, Integer.
+     *
+     * <p>This is a public convenience factory method that builds using the public factory for a
+     * range.
+     *
+     * @param values never null collection from which min and max values will be used as from and to
```
wmdietl: We usually use the non-null default that the Nullness Checker uses. So if you don't say that null is allowed, it's forbidden. If something does allow null, it should be annotated as `@Nullable` and explain in the javadoc.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
```diff
@@ -166,7 +184,7 @@ public int hashCode() {
      * Compare two ranges in a type safe manner for equality without incurring the cost of an
      * instanceof check such as equals(Object) does.
      *
-     * @param range to compare against
+     * @param range to compare against; never null
```
wmdietl: Same comment about non-null-by-default.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
```diff
@@ -40,65 +42,81 @@
...
-    public static final Range CHAR_EVERYTHING = new Range(Character.MIN_VALUE, Character.MAX_VALUE);
+    public static final Range CHAR_EVERYTHING = create(Character.MIN_VALUE, Character.MAX_VALUE);
 
     /** A range containing all possible 8-bit values. */
-    public static final Range BYTE_EVERYTHING = new Range(Byte.MIN_VALUE, Byte.MAX_VALUE);
+    public static final Range BYTE_EVERYTHING = create(Byte.MIN_VALUE, Byte.MAX_VALUE);
 
-    /** The empty range. */
-    public static final Range NOTHING = new Range();
+    /** The empty range singleton. */
```
wmdietl: In what way is this a singleton? I don't think we guarantee that this is the only Range object with that range.

smonteiro: Two parts to this.

1. Actually this is from the original source code's line 98 (excerpt below) for the private constructor 
` /** Creates the singleton empty range. */`
` private Range() {`
` this.from = Long.MAX_VALUE;`
`this.to = Long.MIN_VALUE;`
	   
2. As written there is the guarantee  that no other existing or publicly created Range will have a different instance with these same bounds as the public factory method enforces from<=to which this special instance violates on purpose.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
smonteiro: Thanks. I was not sure if this code base itself was subjected to the Nullness Checker .  
I will remove the runtime check on a commit to come.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
smonteiro: ok. will do.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
smonteiro: Agreed. 
I edited it in place to make the review comparison easier to tell this was being renamed.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
wmdietl: Ah, I had missed that! Maybe this should be highlighted in the `Range()` constructor? It should only be called once and after that `NOTHING` should be used.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
wmdietl: Yes, that's a good strategy. In these cases, maybe add a TODO comment that you plan to move the method after the review.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
wmdietl: We're not running the Nullness Checker yet (#2916 is starting work on that), but we already follow that style.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
smonteiro: @mernst handled that on the javadoc in this review.

It is getting better all the time. :)

wmdietl: Now I looked at the new code again and you're doing exactly that. So this looks good now.

smonteiro: It is better but I found an issue with the definition used for NOTHING : it does not allow a proper hash code to be computed.

smonteiro:
@mernst @wmdietl  Hurray!  this pull request passes everything :)  Your approval(s) and merge are anxiously awaited :)

smonteiro:
Again passes all checks. Ready to go.

smonteiro:
This was approved. When does it get pushed to master ?

framework/src/main/java/org/checkerframework/common/value/util/Range.java
wmdietl: These two new create methods should be moved up in the file to the other create methods, as discussed above.

mernst:
> This was approved. When does it get pushed to master ?

Two people are assigned to the pull request, so they each have to approve.  Sorry you had to wait 10 days for one of them.

smonteiro:
I am glad to move the renamed method up once this is merged into master as moving it with the many pull requests slated towards this same Range.java file will make a mess for diffs for merging changes.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
smonteiro: See https://github.com/typetools/checker-framework/pull/2924#issuecomment-560803366

wmdietl:
This was approved on Friday and I looked at it again today. I don't understand what you mean with "sit out this long". I gave comments within a day of opening of the PR and now within a workday of another review.

You had agreed to move the methods before, but you haven't done so yet.
I don't see how moving the methods becomes easier after merging into master. You will have the same merge problems as moving them now.

smonteiro:
The comment about sitting and becoming stale is about being 11 days from inception, not about you in any particular way.
Since I created this pull request I have developed quite a few changes to offer based on this being in place as is.
The desire was to make comparing for the additional changes very simple. 
By forcing a move now and to master, all of those isolated changes which were branched off separately from master to adhere to the policy of one-topic-of-change-per-pull-request become stale.
I hope this explains where I was coming from. 
It  is moot now. I edited the source file to move the methods up.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
```diff
@@ -331,35 +362,30 @@ public Range byteRange() {
...
-        return BYTE_EVERYTHING;
+        return createOrElse((byte) this.from, (byte) this.to, BYTE_EVERYTHING);
     }
 
     /**
      * Returns true if the element is contained in this range.
      *
      * @param element the value to seek
-     * @return true if {@code element} is in this range
+     * @return true or false
```
wmdietl: What was wrong with the previous explanation? I find `true or false` rather useless as explanation for a boolean, compared to what was here before.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
```diff
@@ -331,35 +362,30 @@ public Range byteRange() {
...
     }
 
     /**
-     * Returns true if the other range is contained in this range.
+     * Returns true if the element is contained in this range.
      *
-     * @param other the range that might be within this one
-     * @return true if {@code other} is within this range
+     * @param other the range to seek
+     * @return true or false
```
wmdietl: Same comment about javadoc.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
```diff
@@ -152,7 +199,7 @@ public String toString() {
     }
 
     @Override
-    public boolean equals(@Nullable Object obj) {
+    public boolean equals(Object obj) {
```
wmdietl: You should not remove the `@Nullable` annotation here. The Nullness Checker will complain about an illegal override. The Nullness Checker isn't run on this code yet, but there is no reason to remove this correct specification.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
```diff
@@ -41,96 +42,142 @@
...
+
+    /**
+     * Internal factory that handles creation of Range that may be in overflow or underflow as a
+     * result of an internal cast conversion.
+     *
+     * <p>Overflow and underflow here are any violations of {@code from<=to}.
+     *
+     * @param from given from value.
+     * @param to given to value.
+     * @param underflow.
```
wmdietl: This parameter name is a bit confusing. The documentation above talks about overflow and underflow, but then there is only an underflow parameter without explanation. Can you add javadoc and use a better name, maybe `alternative`?

framework/src/main/java/org/checkerframework/common/value/util/Range.java
```diff
@@ -41,96 +42,142 @@
...
 
     /**
      * Returns a range with its bounds specified by two parameters, {@code from} and {@code to}. If
-     * {@code from} is greater than {@code to}, returns {@link #EVERYTHING}.
+     * {@code from} is greater than {@code to}, returns {@link #NOTHING}.
      *
      * @param from the lower bound (inclusive)
      * @param to the upper bound (inclusive)
-     * @return newly-created Range or EVERYTHING
+     * @return newly created Range or EVERYTHING
```
wmdietl: Should the `EVERYTHING` be `NOTHING` as suggested by the method name?

framework/src/main/java/org/checkerframework/common/value/util/Range.java
```diff
@@ -331,35 +362,30 @@ public Range byteRange() {
...
-     * @return true if {@code element} is in this range
+     * @return true or false
      */
     public boolean contains(long element) {
         return from <= element && element <= to;
     }
 
     /**
-     * Returns true if the other range is contained in this range.
+     * Returns true if the element is contained in this range.
```
wmdietl: Isn't `other range` better than `element` when there is no `element` parameter?

framework/src/main/java/org/checkerframework/common/value/util/Range.java
```diff
@@ -1120,12 +1151,11 @@ public boolean isConstant() {
      * Determines if this range is completely contained in the range specified by the given lower
      * bound inclusive and upper bound inclusive.
      *
-     * @param lb lower bound for the range that might contain this one
-     * @param ub upper bound for the range that might contain this one
-     * @return true if this range is within the given bounds
+     * @param lb lower bound
+     * @param ub upper bound
+     * @return true or false
```
wmdietl: Is there a better explanation?

framework/src/main/java/org/checkerframework/common/value/util/Range.java
smonteiro: done for "alternate"

framework/src/main/java/org/checkerframework/common/value/util/Range.java
smonteiro: corrected

framework/src/main/java/org/checkerframework/common/value/util/Range.java
smonteiro: corrected

smonteiro:
Hurray! :) Done! :)

mernst:
Thanks @smonteiro for the changes and @wmdietl for the reviews that further improved it.

PR #2927, opened 2019-11-22
Declaration annotations don't apply to array component types (Fixes #2888)

### LINKED ISSUE #2888: Aliased declaration annotations on array component types
According to https://checkerframework.org/manual/#declaration-annotations-moved we are moving aliased declaration annotations from the array component to the array type.

The following example:

```java
import com.sun.istack.internal.Nullable;

class Demo {
    @Nullable Object [] noa;
    
    @Nullable Object [] foo(@Nullable Object [] p) {
        noa = p;
        noa = foo(noa);
        noa = foo(p);
        return p;
    }
}
```

(`com.sun.istack.internal.Nullable` only exists in JDK 8; I didn't look which of our other aliases might exist in both 8 and 11.)

Should compile without any error, but raises six:

```
Demo.java:3: error: [initialization.fields.uninitialized] the constructor does not initialize fields: noa
class Demo {
^
Demo.java:8: error: [assignment.type.incompatible] incompatible types in assignment.
        noa = foo(noa);
                 ^
  found   : @Initialized @NonNull Object @Initialized @Nullable []
  required: @Initialized @Nullable Object @Initialized @NonNull []
Demo.java:8: error: [argument.type.incompatible] incompatible types in argument.
        noa = foo(noa);
                  ^
  found   : @Initialized @Nullable Object @Initialized @NonNull []
  required: @Initialized @NonNull Object @Initialized @Nullable []
Demo.java:9: error: [assignment.type.incompatible] incompatible types in assignment.
        noa = foo(p);
                 ^
  found   : @Initialized @NonNull Object @Initialized @Nullable []
  required: @Initialized @Nullable Object @Initialized @NonNull []
Demo.java:9: error: [argument.type.incompatible] incompatible types in argument.
        noa = foo(p);
                  ^
  found   : @Initialized @Nullable Object @Initialized @NonNull []
  required: @Initialized @NonNull Object @Initialized @Nullable []
Demo.java:10: error: [return.type.incompatible] incompatible types in return.
        return p;
               ^
  type of expression: @Initialized @Nullable Object @Initialized @NonNull []
  method return type: @Initialized @NonNull Object @Initialized @Nullable []
6 errors
```

The field `noa` is `@Nullable`, as that annotation is transferred from the component type to the array, so there shouldn't be an initialization warning.
All the calls and assignments are between syntactically equal types, so the errors are equally unexpected.

smillst:
This is already the case for elements, so this fix is for trees.

Fixes #2888.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
@@ -13,6 +13,7 @@ import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.ExecutableElement;
 import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedExecutableType;
 import org.checkerframework.framework.util.AnnotatedTypes;
+import org.checkerframework.javacutil.AnnotationUtils;
 import org.checkerframework.javacutil.TreeUtils;
 
 /**
@@ -30,11 +31,17 @@ class TypeFromMemberVisitor extends TypeFromTreeVisitor {
         AnnotatedTypeMirror result = TypeFromTree.fromTypeTree(f, node.getType());
 
         // Add primary annotations
-        List<? extends AnnotationTree> annos = node.getModifiers().getAnnotations();
-        if (annos != null && !annos.isEmpty()) {
-            List<AnnotationMirror> ams = TreeUtils.annotationsFromTypeAnnotationTrees(annos);
+        List<? extends AnnotationTree> annoTrees = node.getModifiers().getAnnotations();
+        if (annoTrees != null && !annoTrees.isEmpty()) {
+            List<AnnotationMirror> annos = TreeUtils.annotationsFromTypeAnnotationTrees(annoTrees);
             AnnotatedTypeMirror innerType = AnnotatedTypes.innerMostType(result);
-            innerType.addAnnotations(ams);
+            for (AnnotationMirror anno : annos) {
+                if (AnnotationUtils.isDeclarationAnnotation(anno)) {
+                    result.addAnnotation(anno);
+                } else {
+                    innerType.addAnnotation(anno);
+                }
+            }
         }
 
         Element elt = TreeUtils.elementFromDeclaration(node);
--- a/framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
@@ -36,12 +36,16 @@ class TypeFromMemberVisitor extends TypeFromTreeVisitor {
             List<AnnotationMirror> annos = TreeUtils.annotationsFromTypeAnnotationTrees(annoTrees);
             AnnotatedTypeMirror innerType = AnnotatedTypes.innerMostType(result);
             for (AnnotationMirror anno : annos) {
+                // The code here is similar to
+                // org.checkerframework.framework.util.element.ElementAnnotationUtil.addDeclarationAnnotationsFromElement.
                 if (AnnotationUtils.isDeclarationAnnotation(anno)
                         // Always treat Checker Framework annotations as type annotations.
                         && !AnnotationUtils.annotationName(anno)
                                 .startsWith("org.checkerframework")) {
+                    // Declaration annotations apply to the outer type.
                     result.addAnnotation(anno);
                 } else {
+                    // Type annotations apply to the inner most type.
                     innerType.addAnnotation(anno);
                 }
             }
--- a/framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
@@ -65,7 +65,8 @@ class TypeFromMemberVisitor extends TypeFromTreeVisitor {
 
         // TODO: Needed to visit parameter types, etc.
         // It would be nicer if this didn't decode the information from the Element and
-        // instead also used the Tree.
+        // instead also used the Tree. If this is implemented, then care needs to be taken to put
+        // any alias declaration annotations in the correct place for return types that are arrays.
         ElementAnnotationApplier.apply(result, elt, f);
         return result;
     }
--- a/framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
@@ -74,6 +74,8 @@ class TypeFromMemberVisitor extends TypeFromTreeVisitor {
         // It would be nicer if this didn't decode the information from the Element and
         // instead also used the Tree. If this is implemented, then care needs to be taken to put
         // any alias declaration annotations in the correct place for return types that are arrays.
+        // This would be similar to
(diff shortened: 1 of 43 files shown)


framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
```diff
@@ -30,11 +31,17 @@ public AnnotatedTypeMirror visitVariable(VariableTree node, AnnotatedTypeFactory
         AnnotatedTypeMirror result = TypeFromTree.fromTypeTree(f, node.getType());
 
         // Add primary annotations
```
wmdietl: Can you add a comment to the method, similar to what org.checkerframework.framework.util.element.ElementAnnotationUtil.addAnnotationsFromElement is doing?
Cross-references between the two places are probably also a good idea, to make sure they get updated together.

framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
```diff
@@ -30,11 +31,17 @@ public AnnotatedTypeMirror visitVariable(VariableTree node, AnnotatedTypeFactory
...
-        List<? extends AnnotationTree> annos = node.getModifiers().getAnnotations();
-        if (annos != null && !annos.isEmpty()) {
-            List<AnnotationMirror> ams = TreeUtils.annotationsFromTypeAnnotationTrees(annos);
+        List<? extends AnnotationTree> annoTrees = node.getModifiers().getAnnotations();
```
wmdietl: This change only overrides `visitVariable`, wheras `addAnnotationsFromElement` is used in other contexts as well.
In particular, this probably doesn't do the right thing for method return types - they are not variables. Could you add a test to make sure?

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -801,4 +801,26 @@ public static Name getElementValueClassName(
...
     }
+
+    /**
+     * Returns true if anno is a declaration annotation. In other words, returns true if anno cannot
+     * be written on uses of types.
+     *
+     * @param anno the AnnotationMirror
+     * @return true if anno is a declaration annotation.
+     */
+    public static boolean isDeclarationAnnotation(AnnotationMirror anno) {
```
wmdietl: Move this method next to `org.checkerframework.framework.util.AnnotatedTypes.isTypeAnnotation(AnnotationMirror, Class<?>)` or move that one here? It would be better to have them next to each other or have cross-references.

framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
```diff
@@ -30,11 +31,17 @@ public AnnotatedTypeMirror visitVariable(VariableTree node, AnnotatedTypeFactory
...
+        if (annoTrees != null && !annoTrees.isEmpty()) {
+            List<AnnotationMirror> annos = TreeUtils.annotationsFromTypeAnnotationTrees(annoTrees);
             AnnotatedTypeMirror innerType = AnnotatedTypes.innerMostType(result);
-            innerType.addAnnotations(ams);
+            for (AnnotationMirror anno : annos) {
+                if (AnnotationUtils.isDeclarationAnnotation(anno)) {
```
wmdietl: Note how the logic in `addAnnotationsFromElement` is different: it only checks whether the annotation is part of the `org.checkerframework` package, not that it is a declaration annotation.
Shouldn't the logic between these two be the same?

smillst: No, for elements, the annotation in `@Nullable Object []` is placed in a different location in the element depending on whether it is a type annotation or a declaration annotation. For elements, we only need to move our compatibility annotations to the component type.   

For trees, the annotation in `@Nullable Object []` is the same place in the tree no matter whether the annotation is a declaration annotation or a type annotation.  So we have to move it to the correct place.  

I did forget to treat our compatibility annotations as type annotations, so I've made that change.

framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
smillst: There is a test, `foo2`.  Because annotations from methods are decoded using the element rather than the tree, this logic isn't needed there. (I added a comment in that method below.) All other places where a declaration annotation can be written are variables.  So this should be enough.

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
smillst: `isTypeAnnotation` is never used and it uses a static cache.  How about I just deprecate that method?  I moved `hasTypeQualifierElementTypes` here.

framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
smillst: Done below.

framework/src/main/java/org/checkerframework/framework/util/element/ElementAnnotationUtil.java
```diff
@@ -84,8 +84,10 @@ public static void applyAllElementAnnotations(
      * @param type the type to annotate
      * @param annotations the annotations to add
      */
-    static void addAnnotationsFromElement(
+    static void addDeclarationAnnotationsFromElement(
```
smillst: I'll probably merge this change separately, but I think this rename should help clarify what's happening here.

framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
wmdietl: Sounds good.

framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
```diff
@@ -58,7 +72,8 @@ public AnnotatedTypeMirror visitMethod(MethodTree node, AnnotatedTypeFactory f)
 
         // TODO: Needed to visit parameter types, etc.
         // It would be nicer if this didn't decode the information from the Element and
-        // instead also used the Tree.
+        // instead also used the Tree. If this is implemented, then care needs to be taken to put
+        // any alias declaration annotations in the correct place for return types that are arrays.
```
wmdietl: A reference to `ElementAnnotationUtil.addDeclarationAnnotationsFromElement` might be good here.

framework/src/main/java/org/checkerframework/framework/util/element/ElementAnnotationUtil.java
wmdietl: Yes, this is a better name.

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
wmdietl: Deprecating `isTypeAnnotation` sounds good and this is the more appropriate location for `hasTypeQualifierElementTypes`, so moving here is good.

framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
smillst: Done below.

PR #2928, opened 2019-11-22
Only infer types for code presented as source files

kelloggm:
Whole-program inference should not attempt to infer types for code presented only as bytecode. Doing so is 1) unnecessary, because whole-program inference should only be used to annotate code, not to produce stub files generally AND 2) dangerous, because actually using the outputs of WPI on unchecked code means that those annotations are trusted, but no human has ever looked at them and no typechecker has verified that they are correct.

This change should have no impact on users of jaif-based WPI if they are using `infer-and-annotate.sh`, because the jaif files that are created for classes presented as bytecode are never used.


kelloggm:
This PR is failing one of the WPI tests which uses a static block to update fields. I need to find a better way to check if field updates are to fields defined in bytecode before this can be reviewed.

kelloggm:
Problem has been fixed.

Issue #682, opened 2016-04-26
Whole-program inference: locations not inferred


smillst:
Whole-program inference ignores the following locations:
 - methods or fields in anonymous classes.
 - methods or fields declared in a superclass, but used in a subclass
 - type arguments
 - variable arguments  
 - static methods called in a static context

(This issue is referring to the whole-program inference in PR #612)

PR #2929, opened 2019-11-22
WPI: Add support for inferring the parameters of methods defined in a superclass

kelloggm:


framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
```diff
@@ -223,13 +229,6 @@ public void updateInferredMethodParameterTypes(
             // https://github.com/typetools/checker-framework/issues/682
             return;
         }
-        // TODO: We must handle cases where the method is declared on a superclass.
-        // Currently we are ignoring them. See ElementUtils#getSuperTypes.
-        // See Issue 682
-        // https://github.com/typetools/checker-framework/issues/682
```
smillst: Does this PR fix this issue?

kelloggm: Yes. The test case I added fails without the changes I made.

smillst: Then you should include `Fixes #682` in the pull request title.

kelloggm: Sorry, I misunderstood your question. This PR fixes *the part of 682 described in the code comment I removed*. #682 is a large issue detailing all the known problems with WPI. This PR doesn't fix all of those problems, so closing the issue would be inappropriate.

kelloggm: When this is merged, I can update #682 so that it no longer mentions this problem.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
```diff
@@ -214,7 +214,13 @@ public void updateInferredMethodParameterTypes(
             // https://github.com/typetools/checker-framework/issues/682
             return;
         }
-        ClassSymbol classSymbol = getEnclosingClassSymbol(receiverTree);
+        final MethodTree methodDefinitionTree = (MethodTree) atf.declarationFromElement(methodElt);
```
smillst: declarationFromElement only returns a tree if the methodElt was declared in the current complication unit.
What about using `ElementUtils.enclosingClass(methodElt);` or `((MethodSymbol) methodElt).enclClass()` intstead?

kelloggm: I just added a test case where the use of the method happens in a different compilation unit than the definition of the method, and the tests still pass. I think `declarationFromElement` will return non-null so long as the definition was provided as source, which is fine with me - I don't want to infer types for non-source files anyway.

Did you have a different example in mind than the test case that I just added?

smillst: The test is in the same compilation unit, but a different class.(https://docs.oracle.com/javase/specs/jls/se8/html/jls-7.html#jls-7.3) Try adding a method in a different source file.  If the method was declared in a different compilation unit, `declarationFromElement` will return null.  I've clarified its documentation adce909822.

kelloggm: Sorry, I forgot to actually push the new test: https://github.com/typetools/checker-framework/pull/2929/files#diff-77455f96d2150b3322a0e76c2df697a9R3

That test should fail if what you're saying about `declarationFromElement` is correct, because the method use is in a different compilation unit than the superclass with the method definition.

smillst: Oh, I see for members of a class, the tree can always be found.  So from a correctness point of view, this code is fine.  
However, this code gets the tree only to immediately convert it back to the element in `getEnclosingClassSymbol`.  So why do you need the tree?

kelloggm: I don't think I actually need the tree. The important thing is getting the declaration, and neither of the methods you suggested actually fetch the declaration.

Basically, I have a method invocation, and I need to find the element representing the class it was declared in. `declarationFromUse` does that, but it requires going through the tree.

smillst: Elements represent declarations, so `methodElement` is the method declaration.

kelloggm: Right - I understand your point now. I've simplified the code to use `ElementUtils.enclosingClass`

smillst:
I took a closer look at WholeProgramInferenceScenes and found several more place where the enclosing class of methods/variables was not computed correct.  I've changed all the places I found.  I think this pull request now fixes every in #682 except for varargs and type parameters.

kelloggm:
Thanks for those changes - I think they really improve the code.

smillst:
@kelloggm Can you resolve the merge conflict?

PR #2933, opened 2019-11-23
fewer local interim single use variables exposing implementation signatures

smonteiro:
This may not seem like much, but it is a step in a larger goal.

### DIFF
--- a/framework/src/main/java/org/checkerframework/common/reflection/ClassValAnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/common/reflection/ClassValAnnotatedTypeFactory.java
@@ -122,11 +122,9 @@ public class ClassValAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
             } else if (isSubtype(a2, a1)) {
                 return a1;
             } else {
-                List<String> a1ClassNames = getClassNamesFromAnnotation(a1);
-                List<String> a2ClassNames = getClassNamesFromAnnotation(a2);
                 Set<String> lubClassNames = new TreeSet<>();
-                lubClassNames.addAll(a1ClassNames);
-                lubClassNames.addAll(a2ClassNames);
+                lubClassNames.addAll(getClassNamesFromAnnotation(a1));
+                lubClassNames.addAll(getClassNamesFromAnnotation(a2));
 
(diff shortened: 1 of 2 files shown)


wmdietl:
Could you outline, maybe as an issue, what the larger goal is?
At the moment this doesn't look like an improvement.

smonteiro:
Currently the return type on those signatures is List<String> and the local variable is just a pass-through.
That locks the implementation of getClassNamesFromAnnotation of  into List<String> not allowing it to evolve to  Set<String>  or SortedSet<String>.

wmdietl:
Ok, but why not do that when that signature evolves?
This PR doesn't change all invocations of `getClassNamesFromAnnotation` so there is remaining work in any case.
What is the benefit of doing this now over waiting until we actually need that change?

smonteiro:
Much like one strives to program to the interface instead of the concrete class, programming to the concept of a returned collection instead of the structure of the collection.

Why bother with the pass trough single use  local variable with a narrowing type when it is used solely for this ? 
Why have a local temporary pass-through variable narrowing the type to List<String>  when Collection<String> will do ? 
Of all the answers, using the direct evaluation as the argument to addAll() has the fewest number of lines and no arbitrary narrowing local type influencing the implementation of getClassNamesFromAnnotation().

smonteiro:
"You are not going to need it" has limits.
Larger changes require a lot of review time.
This is a tiny, no negative impact refactoring today.

If we are to evaluated the merit of it today in its tiny scope all we get is tiny:
two fewer local variables per function, no intervening distraction with allocation a TreeSet to then get back to the conversation about the class names, fewer lines.

Tiny. Cheap today.

smonteiro:
The Travis CI build is clean but this is not. What is going on ?

mernst:
@smonteiro There was some debugging gone wrong, at a time I lost Internet connectivity.  My apologies for that.  I expect the current job to pass.

mernst:
> Why bother with the pass trough single use local variable

One reason is that naming a concept (a value or computation) can make it clearer.

It's not true for all code, but I find this particular code easier to read with named local variables.
(This is analogous to abstracting out a helper function that is only used once.)

I agree with your comment that making the type explicit does couple the implementation.
I'm not positive that that disadvantage outweighs the advantage of giving a name, though.

PR #2934, opened 2019-11-24
Don't unnecessarily call getDeclAnnotation

nhat-nguyen:
Work being done on both branches are the same, additionally, it seems like getDeclAnnotation doesn't return `null`.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
@@ -3055,12 +3055,7 @@ public class AnnotatedTypeFactory implements AnnotationProvider {
 
             // Retrieving annotations from stub files.
             Set<AnnotationMirror> stubAnnos = stubTypes.getDeclAnnotation(elt);
-            if (stubAnnos != null) {
-                results.addAll(stubAnnos);
-            } else {
-                stubAnnos = stubTypes.getDeclAnnotation(elt);
-                results.addAll(stubAnnos);
-            }
+            results.addAll(stubAnnos);
 
             if (elt.getKind() == ElementKind.METHOD) {
                 // Retrieve the annotations from the overridden method's element.


framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -3055,12 +3055,7 @@ private AnnotationMirror getDeclAnnotation(
 
             // Retrieving annotations from stub files.
             Set<AnnotationMirror> stubAnnos = stubTypes.getDeclAnnotation(elt);
-            if (stubAnnos != null) {
-                results.addAll(stubAnnos);
-            } else {
-                stubAnnos = stubTypes.getDeclAnnotation(elt);
```
wmdietl: @smillst Should this maybe have been a call to something else?

mernst:
> Work being done on both branches are the same, additionally, it seems like getDeclAnnotation doesn't return `null`.

This comment doesn't make sense.  Which are the "both branches" you are referring to?
Which line of code does the comment about getDeclAnnotation refer to, and why isn't it a comment on that line of code in this pull request?

nhat-nguyen:
> This comment doesn't make sense. Which are the "both branches" you are referring to?
I'm sorry it was ambiguous. Regardless of the nullness of `stubAnnos`, we still add the return value of `getDeclAnnotation`, which should  be the same unless `getDeclAnnotation` is not idempotent, to `results`.

> Which line of code does the comment about getDeclAnnotation refer to, and why isn't it a comment on that line of code in this pull request?
I'm not sure what comment you are referring to.

mernst:
I think by "both branches", you mean both branches of some if statement.  I had interpreted it as referring to two GitHub branches, since your comment was not attached to any particular line of code nor to any previous conversation in this pull request.  (Please try to attach comments to code whenever the comment refers to code.)

Regarding `getDeclAnnotation`, you made a remark about it, which seemed to be referring to some part of your pull request, but it was not clear to me what part.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
smillst: I think it was just a mistake.  I don't see anything else that this should be doing.

Issue #1908, opened 2018-03-22
Enforce @RelevantJavaTypes annotation


mernst:
The [`@RelevantJavaTypes`](https://checkerframework.org/api/org/checkerframework/framework/qual/RelevantJavaTypes.html) annotation can prevent users from writing annotations on incorrect locations.  However, it is not currently enforced.

The same `@RelevantJavaTypes` restrictions should be enforced on
 * source code
 * internal Checker Framework data structures

When a type is not relevant, defaulting should probably be different for wildcard upper and lower bounds than for other locations.

`@RelevantJavaTypes` should perhaps be made more fine-grained.

PR #2942, opened 2019-11-26
Reuse existing functions for TypeKind unboxing and range determination

smonteiro:


smonteiro:
Ready for review.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -2327,4 +2232,43 @@ public AnnotatedTypeMirror getDummyAssignedTo(ExpressionTree expressionTree) {
         }
         return null;
     }
+
+    /**
+     * Unboxes a type it needed all the way to a primitive integral.
+     *
+     * @param type TypeMirror to unbox.
```
mernst: Minor:  In a Javadoc `@param`, `@return`, etc. clause, the initial text is a sentence fragment that starts with a lowercase (not capital) letter (which you did, thanks) and does not end with a period unless followed by another sentence.

I don't make the rules; those are just the conventions.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -2327,4 +2232,37 @@ public AnnotatedTypeMirror getDummyAssignedTo(ExpressionTree expressionTree) {
         }
         return null;
     }
+
+    /**
+     * Unboxes a type if needed all the way to a primitive integral.
```
mernst: What does "all the way" mean?  The implementation takes at most one step.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -2327,4 +2232,37 @@ public AnnotatedTypeMirror getDummyAssignedTo(ExpressionTree expressionTree) {
...
+     *
+     * @param type TypeMirror to unbox
```
mernst: What are the legal values?

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -2327,4 +2232,37 @@ public AnnotatedTypeMirror getDummyAssignedTo(ExpressionTree expressionTree) {
...
+     * @return the TypeKind once unboxed
```
mernst: What are the possible values?

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -2327,4 +2232,37 @@ public AnnotatedTypeMirror getDummyAssignedTo(ExpressionTree expressionTree) {
...
+     */
+    private static TypeKind resolveToPrimitiveIntegralTypeKind(TypeMirror type) {
+        TypeKind typeKind = NumberUtils.unboxPrimitive(type);
+        if (isPrimitiveIntegral(typeKind)) {
+            return typeKind;
+        }
+        throw new BugInCF(type.toString() + " expected to be an integral type.");
```
mernst: This error message is confusing.  If bad client code passes in (say) String, the error is not that String is expected to be an integral type.  The error is that the argument should have been an integral type.  Could you reword it so it doesn't take reading a couple of times to understand?

framework/src/main/java/org/checkerframework/common/value/util/Range.java
```diff
@@ -1184,4 +1185,27 @@ private static Range bigRangeToLongRange(BigInteger bigFrom, BigInteger bigTo) {
...
         return createRangeOrEverything(longFrom, longTo);
     }
+
+    /**
+     * Returns a Range representing all possible values for the given primitive type.
+     *
+     * @param typeKind is a java.lang.TypeKind that isPrimitive() and integral.
+     * @return the matching range for that primitive type.
+     */
+    public static Range byPrimitiveTypeKind(TypeKind typeKind) {
```
mernst: This name doesn't reflect what the routine does.  Could you make it use some words from the first sentence of the documentation, or otherwise a bit more obvious?

framework/src/main/java/org/checkerframework/common/value/util/Range.java
```diff
@@ -1184,4 +1185,27 @@ private static Range bigRangeToLongRange(BigInteger bigFrom, BigInteger bigTo) {
         long longTo = bigTo.longValue();
         return createRangeOrEverything(longFrom, longTo);
     }
+
+    /**
+     * Returns a Range representing all possible values for the given primitive type.
+     *
+     * @param typeKind is a java.lang.TypeKind that isPrimitive() and integral.
```
mernst: You don't need to restate the Java type.  For brevity, please just give facts that aren't obvious from the method's signature.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
smonteiro: When read in context, the names of the new functionality refactor-extracted from them are not so foreign:
return Range.byPrimitiveTypeKind(resolveToPrimitiveIntegralTypeKind(type)).from;
and 
return Range.byPrimitiveTypeKind(resolveToPrimitiveIntegralTypeKind(type)).to;

smonteiro:
The goal here was to reduce the code duplication in the bodies found in ValueAnnotatedTypeFactory.java  at
public long getFromValueFromIntRange(AnnotatedTypeMirror atm) 
public long getToValueFromIntRange(AnnotatedTypeMirror atm) 
Their original bodies are clearly copied and edited except for text in the error message.
(https://github.com/typetools/checker-framework/pull/2942/files#r351958389)

When read in context, the names of the new functionality refactor-extracted from them are not so foreign:
return Range.byPrimitiveTypeKind(resolveToPrimitiveIntegralTypeKind(type)).from;
and 
return Range.byPrimitiveTypeKind(resolveToPrimitiveIntegralTypeKind(type)).to;

The same functionality was already coded elsewhere as a private function NumberUtils.java
private static TypeKind unboxPrimitive(TypeMirror type) 
that could be reused but for the missing error message made it a likely candidate of the BugInCF() invocation originally at ValueAnnotatedTypeFactory.java 
public long getFromValueFromIntRange(AnnotatedTypeMirror atm) 
public long getToValueFromIntRange(AnnotatedTypeMirror atm) 

It may be the case that you are reviewing the changes in unified mode instead of split view.
When read in split mode,  many of these become clear.

I am glad to try to explain it in a different way.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
smonteiro: See https://github.com/typetools/checker-framework/pull/2942#issuecomment-560397137

framework/src/main/java/org/checkerframework/common/value/util/Range.java
```diff
@@ -1168,4 +1185,28 @@ private static Range bigRangeToLongRange(BigInteger bigFrom, BigInteger bigTo) {
...
     }
+
+    /**
+     * Returns the instance of Range for the given Java language primitive.
+     *
+     * @param typeKind is a java.lang.TypeKind that isPrimitive() and integral.
+     * @return the matching range for that primitive type.
+     */
+    public static Range byPrimitiveTypeKind(TypeKind typeKind) {
+        assert typeKind.isPrimitive();
```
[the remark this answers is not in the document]
smonteiro: What is the policy on the use of assert in this framework ? 
I tried to cover both bases: assert for exercising the code logic with unit tests and if asserts are turned off on production compilation there is the throw new IllegalArgumentException()
I have seen both assert and explicit checking for  arguments with throw exception in the code base.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
smonteiro: OK

framework/src/main/java/org/checkerframework/common/value/util/NumberUtils.java
```diff
@@ -84,8 +84,17 @@ public static Range castRange(TypeMirror type, Range range) {
         }
     }
 
-    private static TypeKind unBoxPrimitive(TypeMirror type) {
-        if (type.getKind() == TypeKind.DECLARED) {
+    /**
+     * Easy unboxing of a type.
+     *
+     * <p>Made public as the need arose outside this package.
```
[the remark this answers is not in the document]
smonteiro: OK

framework/src/main/java/org/checkerframework/common/value/util/Range.java
smonteiro: I am open to suggestions for a better name.

smonteiro:
@mernst It seems that there are changes requested in this pull request but they are in outdated items.
What do we need to do to resolve this ?

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -2327,4 +2232,37 @@ public AnnotatedTypeMirror getDummyAssignedTo(ExpressionTree expressionTree) {
...
         return null;
     }
+
+    /**
+     * Evaluates a given annotation to its primitive integral TypeKind equivalent.
+     *
+     * @param type {@link javax.lang.model.type.TypeKind}
+     * @return one of INT, SHORT, BYTE, CHAR, LONG
+     */
+    private static final TypeKind toPrimitiveIntegralTypeKind(TypeMirror type) {
```
wmdietl: The `final` doesn't add anything for a `private` method.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -2327,4 +2232,37 @@ public AnnotatedTypeMirror getDummyAssignedTo(ExpressionTree expressionTree) {
         }
         return null;
     }
+
+    /**
+     * Evaluates a given annotation to its primitive integral TypeKind equivalent.
+     *
+     * @param type {@link javax.lang.model.type.TypeKind}
```
wmdietl: Instead of repeating the type, maybe `the type to convert`.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -2327,4 +2232,37 @@ public AnnotatedTypeMirror getDummyAssignedTo(ExpressionTree expressionTree) {
         }
         return null;
     }
+
+    /**
+     * Evaluates a given annotation to its primitive integral TypeKind equivalent.
```
wmdietl: The parameter is a `TypeMirror`, so talking about an `annotation` is confusing.

wmdietl: How about `Determine the primitive integral TypeKind for the given type`.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -2327,4 +2232,37 @@ public AnnotatedTypeMirror getDummyAssignedTo(ExpressionTree expressionTree) {
...
+    private static final TypeKind toPrimitiveIntegralTypeKind(TypeMirror type) {
+        TypeKind typeKind = NumberUtils.unboxPrimitive(type);
+        if (isPrimitiveIntegral(typeKind)) {
+            return typeKind;
+        }
+        throw new BugInCF(type.toString() + " expected to be an integral type.");
+    }
+
+    /**
+     * Return true if the argument is one of INT, SHORT, BYTE, CHAR, LONG.
```
wmdietl: How about `Determine whether the argument is one of INT, SHORT, BYTE, CHAR, LONG or not.`
Or use `if and only if`.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
```diff
@@ -1180,4 +1181,28 @@ private boolean isWithinHalfLong() {
...
         return isWithin(Integer.MIN_VALUE, Integer.MAX_VALUE);
     }
+
+    /**
+     * Returns a Range representing all possible values for the given primitive type.
+     *
+     * @param typeKind is java.lang.TypeKind of INT,SHORT,BYTE,CHAR,LONG
+     * @return the matching range for that primitive type
+     */
+    public static final Range create(TypeKind typeKind) {
```
wmdietl: No `final` needed on a `static` method.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -2327,4 +2232,37 @@ public AnnotatedTypeMirror getDummyAssignedTo(ExpressionTree expressionTree) {
...
+     *
+     * @param typeKind the TypeKind to inspect
+     * @return true if the argument is a primitive integral type
+     */
+    private static final boolean isPrimitiveIntegral(TypeKind typeKind) {
```
wmdietl: `final` is unnecessary.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -337,54 +335,8 @@ public long getFromValueFromIntRange(AnnotatedTypeMirror atm) {
...
-                        from = Byte.MIN_VALUE;
-                        break;
-                    case "java.lang.Character":
-                        from = Character.MIN_VALUE;
-                        break;
-                    case "java.lang.Long":
-                        from = Long.MIN_VALUE;
-                        break;
-                    default:
-                        throw new UserError(
```
wmdietl: Previously we got a user error. What happens now when an `@IntRange` is used on something invalid, e.g. a `String`?

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
smonteiro: You are right.  I will see that it is removed.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
smonteiro: Agreed again.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
smonteiro: Agreed.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
smonteiro: done

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
smonteiro: I opt for tacit over verbose. 
When dealing with private static functions of minimal logic, it is more burdensome than helpful to expand in prose in the javadoc.  There is only one argument and the return is boolean.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
smonteiro: IMHO, if the javadoc is supposed to be helpful linking to the class from which the constants are used is more helpful than just saying "the  type to convert"

mernst:
> @mernst It seems that there are changes requested in this pull request but they are in outdated items.
> What do we need to do to resolve this ?

GitHub has two separate concepts:
 * Whether changes are requested on the pull request as a whole.  This is a boolean (per reviewer) that is controlled manually by people.  When the code review is ready for re-review, ping the reviewer.
 * Whether comments are visible on the pull request webpage.  By default, even outdated comments and responses are visible, which is a bit confusing.  By convention, you shouldn't ping a reviewer until you have either made the suggested change (in which case no response on the pull request webpage is necessary) or have written a response on the pull request webpage.

Does that answer your question?  If I misinterpreted, please clarify.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
mernst: The Javadoc automatically links to the type of each parameter, so there is no need to add a second link adjacent to it.  I agree with @wmdietl that "the type to convert" is better.

(If Javadoc didn't already link to the type, I would agree that such a link is useful.)

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
mernst: I prefer to avoid the phrase "determine whether" because it isn't explicit about what the routine returns (it might be the negation of the determined value), so it makes me read the documentation twice to understand it.

I think "if and only if" is fine, and so is the current version is acceptable because "if and only if" is understood.  To avoid verbosity that can make the documentation harder to read, I sometimes use "iff" which is a standard mathematical abbreviation for "if and only if", but not everyone knows that abbreviation.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
mernst: @smonteiro If you accept a comment, you don't need to write a reply; just make the change.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
mernst: @smonteiro Could you add a test for this?

framework/src/main/java/org/checkerframework/common/value/util/Range.java
smonteiro: The change was  made and committed to this branch but the review references do not seem to get updated to show the latest about the file with the comments.
If one navigates to this file for editing, one finds that the change is there. **SHRUG**

framework/src/main/java/org/checkerframework/common/value/util/Range.java
```diff
@@ -1180,4 +1181,28 @@ private boolean isWithinHalfLong() {
...
         return isWithin(Integer.MIN_VALUE, Integer.MAX_VALUE);
     }
+
+    /**
+     * Returns a Range representing all possible values for the given primitive type.
+     *
+     * @param typeKind one of INT, SHORT, BYTE, CHAR, or LONG
+     * @return the range for the given primitive type
+     */
+    public static Range create(TypeKind typeKind) {
```
wmdietl: Could you move this method to the other `create` methods in this class?

smonteiro: I am not trying to defy you  but this thing about moving function bodies into an order in the source file affects negatively way one traverses the history of a function (see blame.)
Such move makes it unnecessarily hard to deal with merging large independent concurrent features in parallel as simple side by side changes in time.
I submit to your consideration that 
1.  when we look at code via IDEs, the location of the function body is irrelevant as we can navigate with ease via outlines in any order desired.
2. when we have to look at large file differences here in github having only the minimum amount of change to deal with is better.
I submit that in general, adding methods at the bottom of large files and or renaming them in place has no negative effect on deltas and facilitates merging.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
mernst: The review references are not updated, but there is a tag "Outdated" that shows you have made some change within one line of the comment.  GitHub reviewers are supposed to notice the "Outdated" text and then ignore the text, or click the "resolve conversation" button (which I will now do for this one).

In the past, a review comment was automatically "resolved" (that is, hidden from view) as soon as the submitter made a change within one line of the comment.  The current scheme is a bit more paranoid/untrusting, but we have no choice but to go with it.

framework/src/main/java/org/checkerframework/common/value/util/Range.java
mernst: Regarding the git history, I believe you made two arguments:
 * File organization does not matter.  I respectfully but strongly disagree.  I feel that it is much more important that the code in the file be logically organized, than that the git history be minimal.  I look at the code around a method more often than I look at its git history.  Even within an IDE, the context of a function matters.  It helps in understanding the code's design and how it fits together.
 * Moving the function would increase the size of the git history.  It would not.  Since this is new code, the git history won't be any bigger or smaller if the code is inserted into one location rather than another.

Regarding the difficulty of making multiple changes:  I agree with you, but that is a temporary problem that does not justify a messy organization of the file (which will hamper programmers long into the future).  What other pull requests depend on this one?  Is there anything that I can do to help with that merge?

smonteiro: None of those bullets as you stated them.

I wrote and re-wrote this reply many times.  Long, short, middle ground. Writing lacks tone and is very much subject to misinterpretation among strangers.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
smonteiro: The original functions to gather from and to values had different BuginCF error messages:
To:
```
throw new UserError( "Illegal type \"@IntRange "	
                                        + qualifiedName	
                                        + "\". @IntRange can be applied to Java integral types.");	
                }	
                break;	
            default:	
                throw new BugInCF(anno.toString() + " on a type of kind " + type.getKind());
```
vs
From:
```
default:	
                        throw new UserError(	
                                "Illegal type \"@IntRange "	
                                        + qualifiedName	
                                        + "\". @IntRange can be applied to Java integral types.");	
                }	
                break;	
            default:	
                throw new BugInCF(	
                        "Tried to apply a default to an IntRange annotation that was neither an integral primitive nor a declared type.");	
        }
``` 
NOTE: that UserError() message seems to be missing the word "only"

The choice of what an unified message should say is up to you guys.
My desire here was to reduce duplication and  reuse existing functions whenever possible.

wmdietl: There are three different ways to raise errors: type checking errors, on a particular line; `UserError` which are supposed to be used for invalid flags/configuration; and `BugInCF` for internal errors of the CF.
I would say that the original code here incorrectly used a `UserError`. Instead, this should be a type-checking error raising this problem.

Take the following code:

```java

class IntRangeUsage {
  @IntRange(to = 0) Object o;
}
```

At the moment we get the following error:

```
javac -processor org.checkerframework.common.value.ValueChecker IntRangeUsage.java
error: Illegal type "@IntRange java.lang.Object". @IntRange can be applied to Java integral types.
1 error
```

That's not particularly useful, because it doesn't point you to the right line in the code.
Changing the `UserError` here to a `BugInCF` will make the error even scarier, because we will think that it is an internal inconsistency and show a stacktrace etc.

So https://github.com/typetools/checker-framework/blob/master/framework/src/main/java/org/checkerframework/common/value/ValueVisitor.java#L285 should itself first validate the underlying java type and then the internal code in `ValueATF` can raise the `BugInCF`.

Could you add the above test and make this change in a separate PR? Then we'll know that this PR doesn't break that.

smonteiro: @wmdietl  @mernst 
While researching for that change in the preceding comment, something else seems to have surfaced in that snippet of code.

The definition of a checker framework annotation does not state in a programmatic way what classes the checker annotation is applicable to.
It is implied by naming conventions and javadoc documentation but not enforced as such.

Example on [(2.11.1 live demo )](http://eisop.uwaterloo.ca/live#mode=display) using StringVal as an annotation on an int field.
```
public class Magician  {
@StringVal({"a","b"}) int i;
@StringVal({"a","b"}) Integer bigI;
}
```
Produces no error message. IMHO, it should.

Enforcing an explicitly declared relationship of "this-annotation-is-applicable-to-these-classes-and-their-subclasses"  seems to in order.

At this time, I cannot tell if the current implementation of the framework checks that relationship for the other known checkers in imperative form in code.

mernst: @smonteiro I agree that the Constant Value Checker should issue a warning for
```@StringVal({"a","b"}) int i;```
Thanks for pointing that out.

You can declare that an entire checker is only relevant to certain types:
https://checkerframework.org/manual/#creating-relevant-java-types .
I think that annotation should be permitted on a type annotation as well, and should be enforced.

Let's move that discussion to a separate issue, probably https://github.com/typetools/checker-framework/issues/1908.

smonteiro:
@wmdietl 
> Could you add the above test and make this change in a separate PR? Then we'll know that this PR doesn't break that.

 I implemented the code that attempts to enforce the IntRange only applies to integers but I need a hand with where I should place the tests that are supposed to fail compilation by design.
I never wrote one that was meant to fail compilation for a build. What should that look like.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
wmdietl: https://github.com/typetools/checker-framework/pull/2942#discussion_r354910201
is being addressed in #2993.
Once that PR is merged into this PR, we can continue reviewing this.

wmdietl:
Work finished in #3087

PR #2949, opened 2019-12-02
Move env and types to CF level.

xingweitian:
When trying to run `ConstantPropagationPlayground.java`, we get an error:
```
	at org.checkerframework.dataflow.analysis.Analysis.<init>(Analysis.java:154)
	at org.checkerframework.dataflow.analysis.Analysis.<init>(Analysis.java:145)
	at org.checkerframework.dataflow.cfg.playground.ConstantPropagationPlayground.main(ConstantPropagationPlayground.java:24)
```
Line 144 - 146 in `Analysis.java`:

```java
    public Analysis(T transfer, ProcessingEnvironment env) {
        this(transfer, -1, env);
    }
```

Line 152 - 154 in `Analysis.java`:

```java
    public Analysis(@Nullable T transfer, int maxCountBeforeWidening, ProcessingEnvironment env) {
        this.env = env;
        this.types = env.getTypeUtils();
```

This error is because we pass `null` as the value of `env`, so `env.getTypeUtils()` will raise `NullPointerException`.

This PR fixes it by moving `env` and `types` to CF level.

Contributors: @CharlesZ-Chen @xingweitian


xingweitian:
`getTypes()` and `getEnv()` are used by initialization and nullness checkers, `CFAbstractStore` and `CFAbstractValue`, so I did not remove them.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/Analysis.java
```diff
@@ -132,26 +124,31 @@ public void setCurrentTree(Tree currentTree) {
     /**
      * Construct an object that can perform a org.checkerframework.dataflow analysis over a control
      * flow graph. The transfer function is set later using {@code setTransferFunction}.
+     *
+     * @param maxCountBeforeWidening number of times a block can be analyzed before widening
      */
-    public Analysis(ProcessingEnvironment env) {
-        this(null, -1, env);
+    public Analysis(int maxCountBeforeWidening) {
```
smillst: I think this should be marked as deprecated and then removed rather than removed in one release.

smillst: These are used by ErrorProne, so it's important to document this is the change log.
https://github.com/google/error-prone/blob/master/check_api/src/main/java/com/google/errorprone/dataflow/DataFlow.java#L84

dataflow/src/main/java/org/checkerframework/dataflow/analysis/Analysis.java
```diff
@@ -132,23 +140,46 @@ public void setCurrentTree(Tree currentTree) {
     /**
      * Construct an object that can perform a org.checkerframework.dataflow analysis over a control
      * flow graph. The transfer function is set later using {@code setTransferFunction}.
+     *
+     * @deprecated will be removed in next version
```
smillst: This comment isn't helpful. See https://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#@deprecated for how to write a deprecated comment.  Please update all `@deprecated` comments.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/Analysis.java
```diff
@@ -132,23 +140,46 @@ public void setCurrentTree(Tree currentTree) {
...
+     * @param env associated processing environment
      */
+    @Deprecated
     public Analysis(ProcessingEnvironment env) {
```
smillst: There should exist a no argument constructor to replace this one, but there isn't one.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/Analysis.java
```diff
@@ -132,23 +144,58 @@ public void setCurrentTree(Tree currentTree) {
...
         this(null, -1, env);
     }
 
+    /**
+     * Construct an object that can perform a org.checkerframework.dataflow analysis over a control
+     * flow graph. The transfer function is set later using {@code setTransferFunction}.
```
smillst: @wmdietl This comment was copied from the old method. `setTransferFunction` was removed in c6e8ae771fea7064fc4a8a669cb2210c3e1f52c4.  Do you know how the transfer function should be set later?

xingweitian: @smillst I just talked about this with Prof. Dietl, `setTransferFunction` should be removed from the comments. `transferfunction` can be set by the subclasses later. For example, in `CFAbstractAnalysis`, it is set by [createTransferFunction()](https://github.com/typetools/checker-framework/blob/master/framework/src/main/java/org/checkerframework/framework/flow/CFAbstractAnalysis.java#L72).

Thanks a lot for your reviews!

dataflow/src/main/java/org/checkerframework/dataflow/analysis/Analysis.java
```diff
@@ -164,20 +214,70 @@ public Analysis(@Nullable T transfer, int maxCountBeforeWidening, ProcessingEnvi
...
 
-    /** The current transfer function. */
+    /**
+     * Construct an object that can perform a org.checkerframework.dataflow analysis over a control
+     * flow graph, given a transfer function.
+     *
+     * @param transfer transfer function
+     * @param maxCountBeforeWidening number of times a block can be analyzed before widening
+     */
+    @SuppressWarnings("nullness")
```
wmdietl: Instead of suppressing the nullness warnings, is it enough to make the deprecated fields `env` and `types` `@Nullable`?

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractAnalysis.java
```diff
@@ -58,12 +60,30 @@
     /** Initial abstract types for fields. */
     protected final List<Pair<VariableElement, V>> fieldValues;
 
+    /** The associated processing environment. */
+    @SuppressWarnings("HidingField")
```
wmdietl: Can you add a TODO to the deprecated fields to remove these suppressions when they are removed?

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractAnalysis.java
```diff
@@ -169,4 +189,24 @@ public V createSingleAnnotationValue(AnnotationMirror anno, TypeMirror underlyin
         annos.add(anno);
         return createAbstractValue(annos, underlyingType);
     }
+
+    /**
+     * Get the types utility.
+     *
+     * @return {@link #types}
+     */
+    @SuppressWarnings("deprecation")
```
wmdietl: Is there a deprecation warning because this method overrides the deprecated supermethod?
Maybe also add a TODO to the supermethod so that we don't forget to remove this suppression once we remove that method.

xingweitian: Yes, the deprecation warning is caused by overriding the deprecated method.

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractAnalysis.java
smillst: @xingweitian TODOs should not be in the JavaDoc.  They should be in `//` comments.  If they are about a `@SuppressWarnings`, then they should appear after annotation, if there is room.  Also @wmdietl asked that you add the TODO to the deprecated field. So this TODO belongs on Analysis#env.

xingweitian:
`all_tests_jdk11` was failed because of the timeout when pulling the docker image...
I don't have the permission to restart the pipeline's job, @smillst @wmdietl could you please restart it for me? Thanks!

PR #2952, opened 2019-12-02
Clean checker/jdk.

smillst:


smillst:
@mernst, I think we intentionally did not clean the `checker/jdk` directory so that the whole jdk would not have to be rebuilt if just a few classes changed.

mernst:
> @mernst, I think we intentionally did not clean the `checker/jdk` directory so that the whole jdk would not have to be rebuilt if just a few classes changed.

That is only relevant if someone has run the clean task, right?  I don't expect that to be very common, in situations where just a few classes have recently changed.

mernst:
Thanks.  Feel free to merge, or to discard if you believe this behavior is the wrong thing.

PR #2957, opened 2019-12-03
Use ErrorProne version 2.3.4 and address new warnings.

smillst:


javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -334,26 +334,46 @@ public static boolean containsSameByName(
...
+    /**
+     * Provide ordering for {@link AnnotationMirror} based on their fully qualified name. The
+     * ordering ignores annotation values when ordering.
+     *
+     * <p>The ordering is meant to be used as {@link TreeSet} or {@link TreeMap} ordering. A {@link
+     * Set} should not contain two annotations that only differ in values.
+     *
+     * @return an ordering over AnnotationMirrors based on their name
+     */
+    public static int compareAnnotationMirrors(AnnotationMirror a1, AnnotationMirror a2) {
```
smillst: The documentation of this method, copied from `annotationOrdering` does not match its implementation. The ordering does not ignore annotation values.  I'm not sure if I should change the documentation or the implementation.

mernst:
@smillst There are still [Error Prone warnings](https://dev.azure.com/typetools/checker-framework/_build/results?buildId=2195&view=logs&jobId=dff7b79b-ed6e-5b72-51c5-d49537c74f23).

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
mernst: Code comments indicate:

1. `annotationOrdering` assumes that no set contains two annotations that differ only in their values.  Is this a good assumption, given that we have made more annotations (even some type annotations) repeatable?

2. The implementation of ANNOTATION_ORDERING is broken when an annotation has values.

I would be inclined to make ANNOTATION_ORDERING respect values, and not use toString but compare the number of values/elements, and if they are equal to sort and compare them.

smillst: RE 1. It's a good assumption if comparing AnnotationMirrors in sets in AnnotatedTypeMirrors, but I still think this method should work with annotations of the same class.

> I would be inclined to make ANNOTATION_ORDERING respect values, and not use toString but compare the number of values/elements, and if they are equal to sort and compare them.

I agree with this and will make that change in a separate PR.

smillst: See #2962.

build.gradle
```diff
@@ -754,6 +754,8 @@ subprojects {
                     '-Xep:EqualsGetClass:OFF',
                     // Not a real problem
                     '-Xep:MixedMutabilityReturnType:OFF',
+                    // Don't want to add a dependency to ErrorProne.
+                    '-Xep:AnnotateFormatMethod:OFF',
```
wmdietl: Instead, we should enable to Format String Checker https://checkerframework.org/manual/#formatter-checker
This would dog-food another checker and give use the same guarantees.
Maybe this is a nice simple starter task?

mernst: Independently, there is https://github.com/google/error-prone/issues/1440 .

smillst: I'll try adding Checker Framework FormatMethod annotations where ErrorProne suggest and then I'll try running the Format Checker.

PR #2960, opened 2019-12-03
Array-backing AnnotationMirrorSet

nhat-nguyen:
This PR attempts to change the underlying data structures of annotation sets from TreeSet and TreeMap to arrays instead to have faster memory access. All of the annotation sets in the benchmark have sizes ≤ 9, so the whole array can fit into a whole cache line, making accessing the whole array almost free. The experiment results in ~10% faster performance when running with the microbenchmark, however it does not have any improvements in the unit tests nor daikon. Prof. @wmdietl asked me to put this up anyway to see if it's worth it to change to this new implementation. The code is still rough but if we decide to go forward with it, I can change and update this PR accordingly.


nhat-nguyen:
It looks like my two new files are failing type checking, causing the unsuccesful pipeline checks. I'll try to fix them.

mernst:
I'm generally supportive of this change.  Thanks!
Before review, documentation needs to be added (that is another reason for the CI failure).

nhat-nguyen:
GitHub somehow automatically closed this, it's probably because I accidentally force-pushed my branch with no changes. I'm still working on documentation and refactoring and will reopen this again once my branch is fixed.

wmdietl:
Let's revisit this when we go to replace AnnotatedTypeMirror.

PR #2962, opened 2019-12-06
Implement annotationOrdering such that annotation values are compared.

smillst:
This overlaps with #2957, but either can be merged first.


javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -334,26 +334,52 @@ public static boolean containsSameByName(
...
+     *
+     * @param a1 the first annotation
+     * @param a2 the second annotation
+     * @return an ordering over AnnotationMirrors based on their name and values
+     */
+    public static int compareAnnotationMirrors(AnnotationMirror a1, AnnotationMirror a2) {
+        if (!AnnotationUtils.areSameByName(a1, a2)) {
+            return annotationName(a1).compareTo(annotationName(a2));
+        }
+        if (AnnotationUtils.sameElementValues(a1, a2)) {
```
wmdietl: This call already iterates once over the elements, expanding defaults, etc.
Is it worth doing that here and again below? If yes, a comment would be good.

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -334,26 +334,52 @@ public static boolean containsSameByName(
...
+                value2 = method.getDefaultValue();
+            }
+            if (value1 == value2) {
+                continue;
+            } else if (value1 == null) {
+                return -1;
+            } else if (value2 == null) {
+                return 1;
+            }
+            int result = value1.getValue().toString().compareTo(value2.getValue().toString());
```
wmdietl: Instead of doing the String comparison here, wouldn't it be nicer to change
`boolean sameAnnotationValueValue(Object val1, Object val2)` https://github.com/typetools/checker-framework/blob/master/javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java#L558
to be a comparator? That is, make a version that returns `int` and the `boolean` version compares against 0.

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -334,26 +334,52 @@ public static boolean containsSameByName(
...
-                    // Because the AnnotationMirror.toString prints the annotation as it appears
-                    // in source code, the order in which annotations of the same class are
-                    // sorted may be confusing.  For example, it might order
-                    // @IntRange(from=1, to=MAX) before @IntRange(to=MAX,from=0).
-                    return n1.compareTo(n2);
-                }
-            };
+    /**
+     * Provide ordering for {@link AnnotationMirror}s. AnnotationMirrors are first compared by their
+     * fully qualified names, then by their element values, in ordered of the name of the element.
```
wmdietl: Should it be `fully-qualified names`?
Should the `ordered` be `order` or is something else strange?

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -552,56 +638,7 @@ public static boolean sameElementValues(AnnotationMirror am1, AnnotationMirror a
...
-            // This case occurs because of the recursive call when comparing arrays of
-            // annotation values.
-            return sameAnnotationValue((AnnotationValue) val1, (AnnotationValue) val2);
-        } else if ((val1 instanceof Type.ClassType) && (val2 instanceof Type.ClassType)) {
-            // Type.ClassType does not override equals
-            return TypesUtils.areSameDeclaredTypes((Type.ClassType) val1, (Type.ClassType) val2);
-        } else {
-            return Objects.equals(val1, val2);
-        }
+        return compareAnnotationValue(av1, av2) != 0;
```
wmdietl: Shouldn't this be `== 0`?

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -335,33 +335,119 @@ public static boolean containsSameByName(
...
+        }
+        if (Objects.equals(val1, val2)) {
+            return 0;
+        }
+        int result = val1.toString().compareTo(val2.toString());
+        if (result == 0) {
+            result = -1;
+        }
+        return result;
+    }
```
wmdietl: Nit: Empty line between top-level elements.

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -535,7 +621,7 @@ public static boolean sameElementValues(AnnotationMirror am1, AnnotationMirror a
             if (aval2 == null) {
                 aval2 = meth.getDefaultValue();
             }
-            if (!sameAnnotationValue(aval1, aval2)) {
+            if (0 != compareAnnotationValue(aval1, aval2)) {
```
wmdietl: You don't have to change this line and the original is probably easier to read.

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
smillst: This is fixed.

PR #2966, opened 2019-12-08
Print information about git repository from which Checker Framework was compiled

mernst:


framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -347,6 +352,9 @@
     // and maybe have an interface for all the methods for which it's safe
     // to override.
 
+    /** True if the git.properties file has been printed. */
+    static boolean gitPropertiesPrinted = false;
```
wmdietl: This field should be `private`.

build.gradle
```diff
@@ -20,6 +22,10 @@ repositories {
     mavenCentral()
 }
 
+gitProperties {
+    gitPropertiesResourceDir = "${project.rootDir}/checker/src/main/resources"
```
wmdietl: Is there a reason to put this under `checker` instead of under `framework`?
The use is in `framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java` so putting the related file under `framework` would be more consistent.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -2292,4 +2304,23 @@ protected Properties getProperties(Class<?> cls, String filePath) {
...
     }
+
+    /** Print information about the git repository from which the Checker Framework was compiled. */
+    void printGitProperties() {
+        if (gitPropertiesPrinted) {
+            return;
+        }
+        gitPropertiesPrinted = true;
+
+        InputStream in = getClass().getResourceAsStream("/git.properties");
```
wmdietl: Wouldn't it be better to use try-with-resources for this, to make sure the files are closed in all cases?

PR #2967, opened 2019-12-08
Use `elementUtils` as the name for variables of type `Elements`

mernst:
Ignore failures in misc jobs; we don't update all documentation for large-scale renaming refactorings.
Any other failures are due to transient timeouts; this is ready for review.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,3 +1,10 @@
+Version 3.1.0
+
+The `elements` field has been renamed to `elementUtils` in classes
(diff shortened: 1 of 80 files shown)


mernst:
The motivation for this pull request is that our code uses `elements` for two different purposes (sometimes in the same file):  for a list of element objects, or for an Elements utilities object.  Also, the method to get an Elements object is called `getElementsUtils`.

However, it is the javac convention to use the method name `getElementUtils` but always use the variable name `elements`.  Therefore, closing this pull request because the Checker Framework already obeys javac's (weird) naming convention.

PR #2968, opened 2019-12-09
Refactor ContractUtils

mernst:


framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -188,24 +200,24 @@ public Postcondition(
          * {@code foo} is guaranteed to be {@code @NonNull} after a call to {@code method()} if that
          * call returns {@code false}.
          */
-        public final boolean annoResult;
+        public final boolean resultValue;
```
smillst: This is a breaking refactor.  I might be worth reverting just this renaming so me can merge the rest of the changes.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -230,25 +242,38 @@ public int hashCode() {
         }
     }
 
-    public List<Contract> getContracts(ExecutableElement element) {
+    /**
+     * Returns all the contracts on {@code methodElement}.
+     *
+     * @param methodElement the method or constructor whose contracts to retrieve
```
wmdietl: Would `executableElement` be a better name? I find `methodElement` misleading for something that can be either a method or a constructor.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -230,25 +242,38 @@ public int hashCode() {
...
     }
 
-    /** Returns the set of preconditions on the element {@code element}. */
-    public Set<Precondition> getPreconditions(Element element) {
+    /// Precondition methods (keep in sync with other two types)
+
+    /**
+     * Returns the preconditions on {@code methodElement}.
+     *
+     * @param methodElement the method or contract whose preconditions to return
```
wmdietl: Should the `contract` be `constructor`?
Same comment about the naming of the parameter.

wmdietl: Same for other methods later.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -367,7 +402,14 @@ public int hashCode() {
         return result;
     }
 
-    /** Returns the conditional postconditions on the method {@code methodElement}. */
+    /// Conditional postcondition methods (keep in sync with other two types)
+
+    /**
+     * Returns the conditional postconditions on {@code methodElement}.
+     *
+     * @param methodElement the method or contract whose unconditional postconditions to return
```
wmdietl: A constructor cannot have a conditional postcondition. So here `methodElement` should really only be for methods, so the name is appropriate.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
mernst: I noticed this and fixed it recently.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
mernst: This would be a reasonable change.  A search revealed that this is a widespread issue in the Checker Framework, not isolated to this class.

PR #2970, opened 2019-12-09
Check validity of AnnotationMirrors when building them; fixes #722

### LINKED ISSUE #722: AnnotationBuilder should issue a warning if a value without a default has not been set.
AnnotationBuilder should issue an error if an element value is not set and has no default value.  For example, the `@Format` annotation has no default value for `value`, so if is a bug `@Format` if built without setting `value`.  Issuing an error when it is built will make finding and fixing the bug easier.

mernst:
We won't merge this pull request as is.  Rather, we will do other refactorings to avoid the use of `fromClassNonsense`; then this pull request can be much simpler, since it will not define that method.
There are two files that use `fromClassNonsense`:
 - [x] StubParser.java: @mernst will fix; see #2981
 - [x] AnnotatedTypeFactory.java; @smillst will fix (one is ~#2990~ but others exist and MultiGraphHierarchy needs to be rewritten. See #3404. )
 - [x] Eliminate changes related to defining `fromClassNonsense`; @mernst will do

Merge together with https://github.com/typetools/checker-framework-inference/pull/173.

~This PR includes the changes in https://github.com/typetools/checker-framework/pull/3654. (Merge https://github.com/typetools/checker-framework/pull/3654 first.)~

Fixes #722.


smillst:
This fixes https://github.com/typetools/checker-framework/issues/722.

smillst (approved):

PR #2973, opened 2019-12-10
Refactorings of StubParser.java

mernst:


framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1904,7 +1922,7 @@ private static void putOrAddToMap(
         if (map.containsKey(key)) {
             map.get(key).addAll(annos);
         } else {
-            map.put(key, annos);
+            map.put(key, new HashSet<>(annos));
```
wmdietl: @mernst This change seems unnecessary. Both calls of this method provide a fresh set. So instead of creating another map, it would be easier to document the assumption that annos is not aliased.

(Moving the comment here from the commit https://github.com/typetools/checker-framework/commit/31d1cd341182486098f58aa07d618faa9cec3482#r36431425)

mernst: The Checker Framework crashes without this change.  `annos` might be an unmodifiable set, such as created by `Collections.singleton`.  This change ensures that it is modifiable.

PR #2871, opened 2019-11-08
Add whole-program inference mode using stub files instead of jaif files

kelloggm:
This PR adds a command-line options to support using .astub files as the output of whole-program inference instead of .jaif files, and permits `-Ainfer` to take an argument:
* `-AmergeStubsWithSource` causes stub types to be used, *even when source files are available*; the stronger type from the two is used.

For now, there is scant documentation on `-Ainfer=stubs`, because there is no practical way for a user to make use of it (for now). This is because the manual only describes how to run whole-program inference via the `infer-and-annotate.sh` script, which is not practical for large programs. Providing a version of `infer-and-annotate.sh` that uses stubs (or something more scalable) is explicitly future work for this PR.

The stub mode of WPI is tested with the same tests that the `.jaif` version of whole-program inference uses. Those tests also implicitly test `-AmergeStubsWithSource`, because it's not possible to overwrite (a lack of) annotations in source files otherwise.

Merge together with https://github.com/typetools/annotation-tools/pull/279

PR #2976, opened 2019-12-11
Refactor whole-program inference implementation

mernst:
This renames a helper class, combines two methods, renames methods, and avoids duplication and omissions in Javadoc.


framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInference.java
```diff
@@ -44,91 +54,112 @@ void updateInferredConstructorParameterTypes(
      * Updates the parameter types of the method {@code methodElt} based on the arguments in the
      * method invocation {@code methodInvNode}.
      *
+     * <p>For each method parameter in methodElt:
```
kelloggm: This comment should be explicit about whether this also updates the receiver parameter (I think it does).

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
```diff
@@ -31,20 +31,17 @@
 
 /**
  * WholeProgramInferenceScenes is an implementation of {@link
- * org.checkerframework.common.wholeprograminference.WholeProgramInference} that uses a helper class
- * ({@link org.checkerframework.common.wholeprograminference.WholeProgramInferenceScenesHelper})
- * that manipulates .jaif files to perform whole-program inference.
+ * org.checkerframework.common.wholeprograminference.WholeProgramInference} that uses a storage
+ * class ({@link
+ * org.checkerframework.common.wholeprograminference.WholeProgramInferenceScenesStorage}) that
+ * manipulates .jaif files to perform whole-program inference.
```
kelloggm: The reference to `.jaif` files here should probably be to the `AScene` library instead, since that's what's actually stored. The scenes are only written out to `.jaif` files once, at the end.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInference.java
```diff
@@ -30,12 +30,22 @@
...
+     *       creation call.
+     * </ul>
+     *
      * @param objectCreationNode the Node that invokes the constructor
      * @param constructorElt the Element of the constructor
      * @param atf the annotated type factory of a given type system, whose type hierarchy will be
      *     used to update the constructor's parameters' types
      */
-    void updateInferredConstructorParameterTypes(
+    void updateFromObjectCreation(
```
smillst: These renames are technically a breaking refactor, but I would guess no external checkers use them though.

mernst: Yes, that is my thought as well.

mernst:
> Overall I think these changes are significant improvements. Do you intend for them to be merged before or after #2871? Either way, there will almost certainly be merge conflicts between the two.

I agree there will be merges.  (Sorry about that.)
I think it will be easiest to merge this into master first.  But in any event I think it will be best for you to do the merge.  I'll leave this assigned to you, so that you can merge it into master and into #2871.

PR #2978, opened 2019-12-12
Error if the Checker Framework is built with a JDK version other than 8 or 11.

smillst:


wmdietl:
@smillst What is the reason for this?
Is the motivation the annotated JDK only being available for those versions? If so, it should also fail when trying to run on other versions.
Using Java 9 works without an issue and at least Java 12 worked in a quick test.
Could you document somewhere all the places that need updating when we support a new JDK?

mernst:
There was a recent thread in which a user tried to build the Checker Framework with JDK 12, it failed, and we told the user to use JDK 11.  We would like to avoid repeating that occurrence.
Once we have fully supported JDK 12, we can change this.  Though at this point I suggest that we don't support non-LTS versions of Java (that is, don't support JDK 12 or JDK 13) because their experimental features (eg, new language features) significantly add to the effort to support them, with little benefit because there are few users of those versions.

PR #2983, opened 2019-12-15
Improved nullness annotations for methods returning boolean, especially equals()

mernst:


checker/jdk/nullness/src/java/lang/Boolean.java
```diff
@@ -123,6 +124,7 @@ public Boolean(@Nullable String s) {
      * @since 1.5
      */
     @Pure
+    @EnsuresNonNullIf(expression="#1", result=true)
```
wmdietl: Shouldn't the same annotations appear in typetools/jdk?
https://github.com/typetools/jdk/blob/master/src/java.base/share/classes/java/lang/Boolean.java doesn't contain this annotation.

mernst: It contains it now.

PR #2985, opened 2019-12-16
Reduce duplicated code in ContractsUtils

mernst:
The main change here is to change `getPreconditions` and `getPostconditions` and `getConditionalPostconditions`, which had a lot of duplicated code, into one-line calls to a single generalized method.

Also, `Contract` is moved to its own file just to keep `ContractsUtils` from being too long and unweildy, and I added some fields to `Contract`.


mernst:
Although this is a breaking refactor, I would like to merge it when it passes review because it simplifies fixing #1307.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -3244,8 +3245,8 @@ private void checkPreAndPostConditions() {
 
             // Check postconditions
             ContractsUtils contracts = ContractsUtils.getInstance(atypeFactory);
-            Set<Postcondition> superPost = contracts.getPostconditions(overridden.getElement());
-            Set<Postcondition> subPost = contracts.getPostconditions(overrider.getElement());
+            Set<Contract> superPost = contracts.getPostconditions(overridden.getElement());
```
wmdietl: @mernst Did you get to try to improve the return types of these methods? It would be nicer if these methods and the code above didn't use the general `Contract` and casts and instead used the more precise types.
You can probably achieve this by having helper methods that have a generic return type and then the existing methods are simple wrappers around the generic method, with the precise return type.

mernst: @wmdietl I did not do so.

mernst: Feel free to make that change in a subsequent pull request.  Thanks!

mernst:
I'm merging this because it has been open over a week.

PR #2990, opened 2019-12-16
Change implementation of isSupportedQualifier to avoid using getTypeQualifiers

smillst:
* Deprecate QualifierHierarchy#getTypeQualifiers. It returns a list of AnnotationMirrors whose element values may not have been properly initialized. Also, for some hierarchies it is not possible to list all possible type qualifiers. 
* Change the implementation of AnnotatedTypeFactory#isSupportedQualifier to search a set of strings for a class name rather than QualifierHierarchy#getTypeQualifiers
* Add versions of AnnotatedTypeFactory#isSupportedQualifier that take a string or a class.
* Fixes one occurrence of AnnotationBuilder that creates invalid annotations. (#2970.)


framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -2401,8 +2423,31 @@ public boolean isSupportedQualifier(@Nullable AnnotationMirror a) {
...
+    /**
+     * Determines whether the given class is a part of the type system under which this type factory
+     * operates.
+     *
+     * @param className annotation class
+     * @return true if that class is a type qualifier in the type system under which this type
+     *     factory operates, false otherwise
+     */
+    public boolean isSupportedQualifier(Class<? extends Annotation> className) {
+        return isSupportedQualifier(className.getCanonicalName());
```
wmdietl: I like the change to remove `AnnotationMirrors` for the supported type qualifiers.
I'm not so sure why we also need the new String versions of these APIs.
Here I'm also confused why this uses the new set of Strings instead of the existing `supportedQuals`. This seems to introduce a duplication between `supportedQuals` and `supportedQualsNames` and it's not clear to me why we need it and what is good about it.
Could you expand on the motivation for this a bit?

smillst: We don't *need* the String version of this method.  We do need `supportedQualsNames` so that we can compare the String name against the AnnotationMirror name for `isSupportedQualifier(AnnotationMirror)`.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -205,6 +206,12 @@
      */
     private final Set<Class<? extends Annotation>> supportedQuals;
 
+    /**
+     * To cache the supported type qualifier names. call {@link #getSupportedTypeQualifierNames()}
+     * instead of using this field directly, as it may not have been initialized.
```
wmdietl: A cross-reference to the existing `supportedQuals` would be good. Do they contain corresponding elements?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -394,6 +401,7 @@ public AnnotatedTypeFactory(BaseTypeChecker checker) {
         this.visitorState = new VisitorState();
 
         this.supportedQuals = new HashSet<>();
+        this.supportedQualNames = new TreeSet<>();
```
wmdietl: Is there a reason for using `TreeSet` here and a `HashSet` for `supportedQuals`? As this set will be queried a lot, `HashSet` should be better.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -205,6 +206,12 @@
      */
     private final Set<Class<? extends Annotation>> supportedQuals;
 
+    /**
+     * To cache the supported type qualifier names. call {@link #getSupportedTypeQualifierNames()}
```
wmdietl: Can you document that these are the fully-qualified names, not the simple names.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -2401,8 +2423,31 @@ public boolean isSupportedQualifier(@Nullable AnnotationMirror a) {
...
+
+    /**
+     * Determines whether the given class is a part of the type system under which this type factory
+     * operates.
+     *
+     * @param className annotation class
+     * @return true if that class is a type qualifier in the type system under which this type
+     *     factory operates, false otherwise
+     */
+    public boolean isSupportedQualifier(Class<? extends Annotation> className) {
```
wmdietl: `className` is a bad name for a parameter of type `Class`. How about `annotationClass`?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
wmdietl: That didn't answer why this isn't just `return getSupportedTypeQualifiers().contains(className)`. The two sets contain corresponding values, so why query the String set instead of the Class set?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -923,6 +931,25 @@ public AnnotationFormatter getAnnotationFormatter() {
         return Collections.unmodifiableSet(supportedQuals);
     }
 
+    /**
+     * Returns an immutable set of the type qualifiers supported by this checker.
```
wmdietl: Here and in the `@return` it would be better to use something closer to the documentation in the field, like `Returns an immutable set of the fully-qualified names of the classes supported by this checker.`
At the moment this doesn't explain what the Strings represent.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -201,11 +201,18 @@
     protected TypeArgumentInference typeArgumentInference;
 
     /**
-     * To cache the supported type qualifiers. call {@link #getSupportedTypeQualifiers()} instead of
-     * using this field directly, as it may not have been initialized.
+     * Caches the supported type qualifiers classes. Call {@link #getSupportedTypeQualifiers()}
```
wmdietl: Should it be `type qualifier classes`?

PR #2996, opened 2019-12-17
Annotate Object.equals(Object) overrides as accepting null.

cpovirk:
Most overrides are already annotated that way, but I noticed that
Map.equals(Object) was not, and I grepped through for other instances.


mernst:
Thank you, Chris.  I must have missed those instances.  I appreciate your cleanup.

mernst:
Whenever a change is made to checker-framework/checker/jdk, a similar change must be made to https://github.com/typetools/jdk.  (We are working on eliminating the need for this duplicated work.)

I will open a pull request against that repository.

PR #2987, opened 2019-12-16
Constant Value Checker: avoid using areSameByClass.

smillst:
Use `areSameByName(AnnotationMirror, String)` instead of `areSameByClass`.  These changes only speed up the checker by ~5%. However, if/when we remove the static cache in `areSameByClass`, then these changes are needed so that the checker isn't any slower.

PR #2997, opened 2019-12-17
Remove the static annotationNameCache from AnnotationUtils

smillst:
* Also adds version of areSameByClass, containsSameByClass, and getAnnotationByClass to AnnotatedTypeFactory that use a (non-static) cache.
 * Replace most calls to the AnnotationUtil methods with calls to the AnnotatedTypeFactory versions. 

I didn't change the Value Checker because #2987 replaced most of those calls with calls to areSameByName.  So, that PR should be merged first.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
@@ -28,6 +28,7 @@ import java.lang.annotation.ElementType;
 import java.lang.annotation.Target;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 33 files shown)


framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -370,6 +371,11 @@ void checkRep(String aliasName) {
     /** The Object.getClass method. */
     protected final ExecutableElement objectGetClass;
 
+    /** Size of the annotationClassNames cache. */
+    private static final int ANNOTATION_CACHE_SIZE = 500;
+
+    /** Maps classes representing AnnotationMirrors to their names. */
```
wmdietl: Canonical names, to make the value description more precise?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -370,6 +371,11 @@ void checkRep(String aliasName) {
...
+    private final Map<Class<? extends Annotation>, String> annotationClassNames;
```
wmdietl: Nit: empty line.

PR #2998, opened 2019-12-18
use explicit SortedSet return signature instead of commenting about it.

smonteiro:


mernst:
Thanks, this is much less error-prone.

PR #2999, opened 2019-12-18
Don't override super with the same implementation as super.

smillst:


smillst:
I'm ignoring the unrelated Javadoc failures.

PR #3002, opened 2019-12-18
Throw an exception if SubtypeOf meta-annotation refers to qualifier not in hierarchy.

smillst:
Also, fix the I18nFormatterLubGlb and FormatterLubGlb Checkers.

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/formatter/FormatterAnnotatedTypeFactory.java
+++ b/checker/src/main/java/org/checkerframework/checker/formatter/FormatterAnnotatedTypeFactory.java
@@ -2,9 +2,7 @@ package org.checkerframework.checker.formatter;
 
 import com.sun.source.tree.LiteralTree;
 import com.sun.source.tree.Tree;
-import java.lang.annotation.Annotation;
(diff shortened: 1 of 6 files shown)


wmdietl:
@smillst Is this the nicest way to fix these two checkers? Why were the right qualifiers not in the subtype hierarchy so far? (If that is the aspect that is fixed about them.)
What is broken about the old code that we could improve instead of requiring all this code for simple subcheckers?

smillst:
@wmdietl 
>@smillst Is this the nicest way to fix these two checkers? Why were the right qualifiers not in the subtype hierarchy so far? (If that is the aspect that is fixed about them.)
What is broken about the old code that we could improve instead of requiring all this code for simple subcheckers?

When a checker extends another checker, then the qualifiers need to be explicitly stated because the super checker's qualifiers will not be found via automatic look up.  I think explicitly stating the qualifiers is the nicest way to do this; if you have another suggestion, please open an issue.  We don't generally recommend subclassing another checker, so I don't think it is worth our time to reduce this boiler plate.   Also, note that these checkers are only used for testing.

PR #3003, opened 2019-12-18
Allow type arg. inference to return type variable (Fixes #3001)

### LINKED ISSUE #3001: Feature request: option to unsoundly ignore more errors issued because of #979
False positives for checkers that only issue a few errors (like the [Object Construction Checker](https://github.com/kelloggm/object-construction-checker)) are dominated by errors caused by #979. When running these checkers, suppressing these warnings is time-consuming and inconvenient for users, because they are unrelated to the property or properties actually being checked.

Some errors related to #979 are already [unsoundly suppressed](https://github.com/typetools/checker-framework/issues/979#issuecomment-279578221) by default. This feature request is to suggest that we do something similar for another class of errors. Whether or not the default is the conservative behavior isn't important to me.

The errors arise from programs like the following:

```java
import java.lang.Class;
import com.sun.tools.javac.code.Symbol;
import com.sun.tools.javac.code.Symbol.ClassSymbol;
import javax.lang.model.element.ElementKind;
import com.sun.tools.javac.code.Scope;
import com.sun.tools.javac.code.Type;
import com.sun.tools.javac.code.Types;
import com.sun.tools.javac.util.Name;

class Issue979RemainingError {
    private final Types types = null;

    private <T extends Symbol> T getMember(
      Class<T> type, ElementKind kind, Symbol classSymbol, Name name) {
	if (classSymbol.type == null) {
	    return null;
	}
	for (Type t : types.closure(classSymbol.type)) {
	    Scope scope = t.tsym.members();
	    for (Symbol sym : scope.getSymbolsByName(name)) {
		if (sym.getKind().equals(kind)) {
		    return type.cast(sym);
		}
	    }
	}
	if (classSymbol.hasOuterInstance()) {
	    T sym = getMember(type, kind, classSymbol.type.getEnclosingType().asElement(), name);
	    if (sym != null) {
		return sym;
	    }
	}
	if (classSymbol.owner != null
	    && classSymbol != classSymbol.owner
	    && classSymbol.owner instanceof Symbol.ClassSymbol) {
	    T sym = getMember(type, kind, classSymbol.owner, name);
	    if (sym != null && sym.isStatic()) {
		return sym;
	    }
	}
	return null;
    }
}
```

Typechecking this class with e.g. the Value Checker, which should issue no errors, does issue two errors.

The command I ran:
```
javac -processor org.checkerframework.common.value.ValueChecker Issue979RemainingErrors.java 
```

The output:
```
Issue979RemainingErrors.java:29: error: [return.type.incompatible] incompatible types in return.
		return sym;
		       ^
  type of expression: T[ extends @UnknownVal Symbol super @UnknownVal Void]
  method return type: T[ extends @UnknownVal Symbol super @BottomVal Void]
Issue979RemainingErrors.java:37: error: [return.type.incompatible] incompatible types in return.
		return sym;
		       ^
  type of expression: T[ extends @UnknownVal Symbol super @UnknownVal Void]
  method return type: T[ extends @UnknownVal Symbol super @BottomVal Void]
2 errors
```

Note that in each case, the error has some similar properties:
1. the type of the expression is `T extends <Top> super <Top>`
2. the required type is `T extends <Top> super <Bottom>`

Running this example through another type system that should issue no errors, such as the Tainting Checker, shows similar behavior:

```
Issue979RemainingErrors.java:29: error: [return.type.incompatible] incompatible types in return.
		return sym;
		       ^
  type of expression: T[ extends @Tainted Symbol super @Tainted Void]
  method return type: T[ extends @Tainted Symbol super @Untainted Void]
Issue979RemainingErrors.java:37: error: [return.type.incompatible] incompatible types in return.
		return sym;
		       ^
  type of expression: T[ extends @Tainted Symbol super @Tainted Void]
  method return type: T[ extends @Tainted Symbol super @Untainted Void]
2 errors
```

I propose that we implement a check to prevent such errors from being issued, based on the pattern of the required type being `<GenericType> extends <Top> super <Bottom>` when the actual type is `<GenericType> extends <Top> super <Top>`.

I'm happy to do the implementation work, but I'm not exactly sure where this information is available that I'd be able to prevent the error from being issued. I've considered two options:
* in `SourceChecker#report`, but the top and bottom types in the qualifier hierarchy aren't available there (or at least, I don't know how to get to them)
* in the common assignment check, but not all errors go through the CAC. That still might be the best way to do it

@smillst, do you have any other ideas on how I should do this, or on which approach you would take?

@msridhar suggested that I do this in public so he could see our discussion rather than just sending an email, so I'm pinging him

smillst:
Type argument inference may infer a type variable that is the same as a type variable for which it is solving.

This fixes the warnings in #3001, but does not implement the feature requested.

### DIFF
--- a/framework/tests/value/Issue3001.java
+++ b/framework/tests/value/Issue3001.java
@@ -1,0 +1,7 @@
+class Issue3001 {
+
+    private <T> T getMember(Class<T> type) {
+        T sym = getMember(type);
+        return sym;
+    }
+}
--- a/framework/src/main/java/org/checkerframework/framework/util/typeinference/DefaultTypeArgumentInference.java
+++ b/framework/src/main/java/org/checkerframework/framework/util/typeinference/DefaultTypeArgumentInference.java
@@ -747,8 +747,7 @@ public class DefaultTypeArgumentInference implements TypeArgumentInference {
             final TypeVariable typeVar = atv.getUnderlyingType();
             if (targets.contains((TypeVariable) TypeAnnotationUtils.unannotatedType(typeVar))) {
                 final AnnotatedTypeMirror inferredType = inferredArgs.get(typeVar);
-                if (inferredType == null
-                        || TypeArgInferenceUtil.containsTypeParameter(inferredType, targets)) {
+                if (inferredType == null) {
                     AnnotatedTypeMirror dummy = typeFactory.getUninferredWildcardType(atv);
                     inferredArgs.put(atv.getUnderlyingType(), dummy);
                 }


kelloggm:
This clearly fixes a bug, and looks like a good addition. Thanks! But, based on @msridhar's comments on #3001, I don't think we should close #3001 with just this PR.

framework/tests/value/Issue3001.java
```diff
@@ -0,0 +1,7 @@
+class Issue3001 {
```
wmdietl: As there is nothing value-checker specific in this test, would it make sense to move it to `framework/tests/all-systems`?

PR #3005, opened 2019-12-19
Use range create effectively. depends on #3004

smonteiro:
This improvement is about using the one-pass for min and max in Range.create(Collection) and the  contains() where it fits.
Again, this change contains #3000 also.


mernst:
Thanks for pointing out the dependencies/inclusions of other pull requests.  That helps a lot.
And thanks for the code improvements!

Issue #2995, opened 2019-12-17
assignment.type.incompatible only with explicit "extends @Nullable Object" on type parameter


cpovirk:
I could be confused, but: I would expect `class Map<K>` and `class Map<K extends @Nullable Object>` to be [equivalent](https://checkerframework.org/manual/#null-defaults). It appears that they're not. Here's some code that uses the longer version:

```java
$ cat Map.java 
import org.checkerframework.checker.nullness.qual.Nullable;

interface Set<E> {}

class Map<K extends @Nullable Object> {
  Set<K> keySet = new KeySet();

  class KeySet implements Set<K> {}
}

$ checker/bin/javac -processor org.checkerframework.checker.nullness.NullnessChecker Map.java
Map.java:6: error: [assignment.type.incompatible] incompatible types in assignment.
  Set<K> keySet = new KeySet();
                  ^
  found   : @Initialized @NonNull Map<K extends @Initialized @NonNull Object>.@Initialized @NonNull KeySet
  required: @Initialized @NonNull Set<K extends @Initialized @Nullable Object>
1 error
```

However, if I tweak the class declaration to the "equivalent" shorter version, then the file compiles successfully. That's the behavior I would have expected in either case.

If I suppress the error, the `.class` files that I get for the long version is almost identical to that for the short version, according to `javap -v`. The only difference appears to be in the order of the `RuntimeVisibleTypeAnnotations`.

[Source file](https://github.com/typetools/checker-framework/files/3975408/Map.java.txt) and [`-version -verbose -AprintAllQualifiers` output](https://github.com/typetools/checker-framework/files/3975418/err.txt) attached.

cpovirk:
This turns out to affect implementations of the JDK's `Map` type, since it's defined with explicit `<K extends @Nullable Object, V extends @Nullable Object>`:

https://github.com/typetools/checker-framework/blob/64f05c44dec5da0e99e6a5d0bd72070495650a22/checker/jdk/nullness/src/java/util/Map.java#L139

(similarly [in typetools/jdk](https://github.com/typetools/jdk/blob/df0ed91f922b3f6c653b657bc6da364d50549ab0/src/java.base/share/classes/java/util/Map.java#L175))

[...]

PR #3007, opened 2019-12-19
Remove explicit "extends @Nullable Object."

cpovirk:
It is supposed to be a no-op. As such, the stub files usually omit it.

Currently, it actually changes behavior in a bad way:
https://github.com/typetools/checker-framework/issues/2995

Until that bug is fixed, this commit serves as a workaround.


cpovirk:
Corresponding change to typetools/jdk coming shortly.

cpovirk:
(https://github.com/typetools/jdk/pull/26)

smillst:
Thanks!

PR #3008, opened 2019-12-19
Warn when using JDK version 9, 10, or 12.

smillst:
Error on JDK versions below 8 and above 12.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
+++ b/framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
@@ -470,7 +470,8 @@ public abstract class SourceChecker extends AbstractTypeProcessor
         // This is used to trigger AggregateChecker's setProcessingEnvironment.
         setProcessingEnvironment(env);
 
-        // Keep in synce with check in checker-framework/build.gradle.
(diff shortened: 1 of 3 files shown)


framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -470,12 +470,22 @@ public final synchronized void init(ProcessingEnvironment env) {
         // This is used to trigger AggregateChecker's setProcessingEnvironment.
         setProcessingEnvironment(env);
 
+        // Keep in synce with check in checker-framework/build.gradle.
```
wmdietl: `synce` -> `sync`.

A third location that we should probably mention in these spots is
https://checkerframework.org/manual/#installation
There it currently says `Requirement: You must have JDK 8 or JDK 11 installed. (The Checker Framework should work with any JDK 8+, but we only test with JDK 8 and JDK 11.)`
which this PR could also bring in line.

mernst:
@wmdietl Please merge this if you approve of it.

sollyucko:
BTW, why is JDK 13+ completely disallowed? Does it have some sort of breaking change? I understand not testing it, but why completely disallow it?

Issue #3014, opened 2019-12-23
Support records


michaelhixson:
Add support for [JEP 359: Records (Preview)](https://openjdk.java.net/jeps/359) to the Checker Framework.

Records are a [preview feature](https://openjdk.java.net/jeps/12) in Java 14.  Java 14 is currently in early access and will be ready for general use in March 2020.  The earliest that records could become a non-preview feature in a non-early-access release is in September 2020 with Java 15, but no commitment has been made to target that release yet.

## What's a record?

Records are a new kind of class.

```java
record Foo(String a, @Nullable String b) {}
```

That code essentially translates to this:

```java
final class Foo extends java.lang.Record {
  private final String a;
  private final @Nullable String b;

  public Foo(String a, @Nullable String b) {
    this.a = a;
    this.b = b;
  }

  public String a() { return a; }
  public @Nullable String b() { return b; }

  // equals, hashCode, and toString based on the fields
}
```

They introduce one new tree kind, [`Tree.Kind.RECORD`](https://github.com/openjdk/jdk/blob/f4af0eadb6eaf9d9614431110ab7fc9c1588966d/src/jdk.compiler/share/classes/com/sun/source/tree/Tree.java#L651-L665), which may be [returned from `ClassTree#getKind()`](https://github.com/openjdk/jdk/blob/f4af0eadb6eaf9d9614431110ab7fc9c1588966d/src/jdk.compiler/share/classes/com/sun/tools/javac/tree/JCTree.java#L798-L799).  Fortunately they don't introduce a new subclass of `Tree` or any new methods to `TreeVisitor`.

They introduce two new element kinds, [`ElementKind.RECORD` and `ElementKind.RECORD_COMPONENT`](https://github.com/openjdk/jdk/blob/f4af0eadb6eaf9d9614431110ab7fc9c1588966d/src/java.compiler/share/classes/javax/lang/model/element/ElementKind.java#L112-L140).

They introduce one new annotation element type, [`ElementType.RECORD_COMPONENT`](https://github.com/openjdk/jdk/blob/f4af0eadb6eaf9d9614431110ab7fc9c1588966d/src/java.base/share/classes/java/lang/annotation/ElementType.java#L119-L137).

## What's involved in supporting records?

There are a handful of switch statements in the Checker Framework that need to be modified.  They do the wrong thing when encountering one of the new record-related enum constants.

That is, there's a bunch of code like this:

```java
switch (kind) {
  case CLASS:
  case ENUM:
[...]

PR #3017, opened 2019-12-25
Add baseline support for records

michaelhixson:
Treat records as regular classes.  Avoid throwing exceptions when records
are encounterd in source code.  Keep backwards compatibility with Java 8
and 11.

Related to #3014

### DIFF
--- a/dataflow/src/main/java/org/checkerframework/dataflow/analysis/FlowExpressions.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/analysis/FlowExpressions.java
@@ -287,6 +287,10 @@ public class FlowExpressions {
                 assert TreeUtils.isUseOfElement(identifierTree)
                         : "@AssumeAssertion(nullness): tree kind";
                 Element ele = TreeUtils.elementFromUse(identifierTree);
+                if (ElementUtils.isClassElement(ele)) {
+                    receiver = new ClassName(ele.asType());
+                    break;
+                }
                 switch (ele.getKind()) {
                     case LOCAL_VARIABLE:
                     case RESOURCE_VARIABLE:
@@ -308,10 +312,6 @@ public class FlowExpressions {
                                         fieldAccessExpression, typeOfId, (VariableElement) ele);
                         break;
                     default:
-                        if (ele.getKind().isClass() || ele.getKind().isInterface()) {
-                            receiver = new ClassName(ele.asType());
-                            break;
-                        }
                         receiver = null;
                 }
                 break;
@@ -370,6 +370,12 @@ public class FlowExpressions {
         }
         assert TreeUtils.isUseOfElement(memberSelectTree) : "@AssumeAssertion(nullness): tree kind";
         Element ele = TreeUtils.elementFromUse(memberSelectTree);
+        if (ElementUtils.isClassElement(ele)) {
+            // o instanceof MyClass.InnerClass
+            // o instanceof MyClass.InnerInterface
+            TypeMirror selectType = TreeUtils.typeOf(memberSelectTree);
+            return new ClassName(selectType);
+        }
         switch (ele.getKind()) {
             case METHOD:
             case CONSTRUCTOR:
@@ -380,10 +386,6 @@ public class FlowExpressions {
                 Receiver r = internalReprOf(provider, memberSelectTree.getExpression());
                 return new FieldAccess(r, fieldType, (VariableElement) ele);
             default:
-                if (ele.getKind().isClass() || ele.getKind().isInterface()) {
-                    TypeMirror selectType = TreeUtils.typeOf(memberSelectTree);
-                    return new ClassName(selectType);
-                }
                 throw new BugInCF("Unexpected element kind: %s element: %s", ele.getKind(), ele);
         }
     }
--- a/dataflow/src/main/java/org/checkerframework/dataflow/analysis/FlowExpressions.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/analysis/FlowExpressions.java
@@ -307,13 +307,11 @@ public class FlowExpressions {
                                 new FieldAccess(
                                         fieldAccessExpression, typeOfId, (VariableElement) ele);
                         break;
-                    case CLASS:
-                    case ENUM:
-                    case ANNOTATION_TYPE:
-                    case INTERFACE:
-                        receiver = new ClassName(ele.asType());
-                        break;
                     default:
+                        if (ele.getKind().isClass() || ele.getKind().isInterface()) {
+                            receiver = new ClassName(ele.asType());
+                            break;
+                        }
                         receiver = null;
                 }
                 break;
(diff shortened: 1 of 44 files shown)


mernst:
Thanks for the pull request.  CI is failing because the changes don't compile.  Could you please correct that and get the CI jobs passing?  Thanks!  Ask for help if you get stuck.

michaelhixson:
I'm trying to understand the latest [build failure in misc_jdk8/test-misc.sh](https://dev.azure.com/typetools/checker-framework/_build/results?buildId=2562&view=logs&jobId=dff7b79b-ed6e-5b72-51c5-d49537c74f23&j=dff7b79b-ed6e-5b72-51c5-d49537c74f23&t=7dabb28f-aafd-5998-fbcd-46ef74c7ca82).  Is it telling me to add javadoc to these methods?  If so, why?

```
/__w/1/s/javacutil/src/main/java/org/checkerframework/javacutil/ElementUtils.java:495: missing documentation for isTypeDeclaration
/__w/1/s/javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java:820: missing documentation for classTreeKinds

##[error]Bash exited with code '1'.
Finishing: test-misc.sh
```

I reverted the changes outside of dataflow and javacutil, since I realized I'm only using those two and can't vouch for any other changes.

mernst:
@michaelhixson The rule requires documentation for anything that a pull request affects.  The rule includes any lines adjacent to ones changed in the pull request.  I added documentation for those two lines.

michaelhixson:
I'll try to figure out if I caused the latest build errors and/or if I can fix them.

Also, I have a couple of questions about code that I'm not sure if I should change.

### 1. Resolver.findClassInPackage

Currently, `Resolver#findClassInPackage(String, PackageSymbol, TreePath)` finds an element and checks if its kind is exactly `ElementKind.CLASS`.  It looks suspicious to me.

https://github.com/typetools/checker-framework/blob/cff1a036a04f102079dba64f1fa9557aadea30d6/javacutil/src/main/java/org/checkerframework/javacutil/Resolver.java#L278-L289

Here is that method's only caller.
https://github.com/typetools/checker-framework/blob/cff1a036a04f102079dba64f1fa9557aadea30d6/framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java#L494-L498

Should that check in `findClassInPackage` be changed to one of these instead?

```java
if (res.getKind().isClass() || res.getKind().isInterface()) {
    return (ClassSymbol) res;
```

```java
if (res instanceof ClassSymbol) {
    return (ClassSymbol) res;
```

### 2. AnnotationUtils.getElementKindsForElementType

Currently, `AnnotationUtils#getElementKindsForElementType(ElementType)` contains code to accomodate Java 9 modules.  It uses a different strategy than the code I added for records, so I'm wondering if I should modify the existing code, my code, or neither.

https://github.com/typetools/checker-framework/blob/cff1a036a04f102079dba64f1fa9557aadea30d6/javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java#L556-L560

In this PR, I added the following code right below that.
```java
if (elementType.name().equals("RECORD_COMPONENT")) {
    return EnumSet.of(ElementKind.valueOf("RECORD_COMPONENT"));
}
```

Shouldn't the module-handling code be changed?  If the caller is passing `ElementType.MODULE`, that implies that `ElementKind.MODULE` also exists, doesn't it?
```java
if (elementType.name().contentEquals("MODULE")) {
    return EnumSet.noneOf(ElementKind.valueOf("MODULE"));
}
```

dataflow/src/main/java/org/checkerframework/dataflow/analysis/FlowExpressions.java
```diff
@@ -376,18 +374,16 @@ private static Receiver internalReprOfMemberSelect(
...
-                TypeMirror selectType = TreeUtils.typeOf(memberSelectTree);
-                return new ClassName(selectType);
             case ENUM_CONSTANT:
             case FIELD:
                 TypeMirror fieldType = TreeUtils.typeOf(memberSelectTree);
                 Receiver r = internalReprOf(provider, memberSelectTree.getExpression());
                 return new FieldAccess(r, fieldType, (VariableElement) ele);
             default:
+                if (ele.getKind().isClass() || ele.getKind().isInterface()) {
+                    TypeMirror selectType = TreeUtils.typeOf(memberSelectTree);
```
wmdietl: Could you copy over the two comments from the previous cases?

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -526,11 +526,13 @@ public static boolean hasInheritedMeta(AnnotationMirror anno) {
     public static EnumSet<ElementKind> getElementKindsForElementType(ElementType elementType) {
         switch (elementType) {
             case TYPE:
-                return EnumSet.of(
-                        ElementKind.CLASS,
-                        ElementKind.INTERFACE,
-                        ElementKind.ANNOTATION_TYPE,
-                        ElementKind.ENUM);
+                EnumSet<ElementKind> typeKinds = EnumSet.noneOf(ElementKind.class);
```
wmdietl: I'm wondering whether it's worth to also extract this into a field, like is done for trees.
This method isn't on a hot path as far as I can tell, but having the symmetry and separate code might still improve readability of the code.
What do you think?

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -558,6 +560,9 @@ public static boolean hasInheritedMeta(AnnotationMirror anno) {
                 if (elementType.name().contentEquals("MODULE")) {
                     return EnumSet.noneOf(ElementKind.class);
```
wmdietl: Regarding your second question: https://github.com/typetools/checker-framework/commit/3ef0af3636c7f6662c638699dddc6a48ce1378f5 added this work-around.
I'm not quite sure why we need it, as we use at least a Java 9 compiler, so using MODULE should always work.
I agree that using `ElementKind.valueOf("MODULE")` should also work here.
Could you try that as a separate PR?

javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -1174,19 +1179,9 @@ public static boolean isEnumSuper(MethodInvocationTree node) {
...
-            case CLASS:
-            case ENUM:
-            case INTERFACE:
-            case TYPE_PARAMETER:
-                return true;
-
-            default:
-                return false;
-        }
+        // These tree kinds are always declarations.  Uses of the declared types
```
wmdietl: You removed this comment in the `Element` version of this method. I agree that the comment doesn't add anything useful, so I would be fine with removing it here too.

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
michaelhixson: This sounds reasonable.  I'm wondering how far I should take the symmetry.

`TreeUtils` has:
```
private static final Set<Tree.Kind> classTreeKinds
public static Set<Tree.Kind> classTreeKinds()
public static boolean isClassTree(Tree)
```

I could add these to `ElementUtils`:
```
private static final Set<ElementKind> classElementKinds
public static Set<ElementKind> classElementKinds()
public static boolean isClassElement(Element)
```

and then I could refactor the places currently doing `element.getKind().isClass() || element.getKind().isInterface()` to instead do `ElementUtils.isClassElement(element)`.

Does that sound worthwhile, or is that going too far?

michaelhixson: Oops, one minor correction.  I would need to specify `EnumSet` as the type of `ElementUtils.classElementKinds` if I wanted to return it from `AnnotationUtils.getElementKindsForElementType`.  That, or return `EnumSet.copyOf(ElementUtils.classElementKinds())` (which would take an [optimized path in `EnumSet.copyOf(...)`](https://github.com/openjdk/jdk/blob/911846b49b85d71f41ec3b4d019e8873e64aee50/src/java.base/share/classes/java/util/EnumSet.java#L170-L172), since the collection it's copying also happens to be an `EnumSet`).

wmdietl: That sounds worthwhile. Using EnumSet sounds good and would probably also be good for the Tree.Kind sets. Also, we should make defensive copies to prevent somebody from modifying the definitions of these sets.
Feel free to open an issue for further cleanups that aren't relevant for this PR.
Thanks!

michaelhixson: Mind taking another look?  I did not end up changing the static set types from `Set` to `EnumSet` or making defensive copies, because that was spiraling out into too many unrelated changes (changing other similar sets in `TreeUtils`, changing internal callers to avoid the defensive copies, but realizing there are no external callers and so no one needs the defensive copy, and so on).  IMHO it's ok as is, but let me know if you disagree.

michaelhixson:
@wmdietl I think I addressed all the feedback.

PR #3019, opened 2019-12-26
Resolve TODO in AnnotationUtils to return ElementKind.MODULE

michaelhixson:
I noticed this code while working on #3017 and [I asked if it should be changed](https://github.com/typetools/checker-framework/pull/3017#issuecomment-569076959) (question 2).  [@wmdietl responded](https://github.com/typetools/checker-framework/pull/3017#discussion_r361505839) asking me to open a new PR, and here we are.


javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -553,10 +553,8 @@ public static boolean hasInheritedMeta(AnnotationMirror anno) {
             case TYPE_USE:
                 return EnumSet.noneOf(ElementKind.class);
             default:
-                // TODO: Add actual case to check for the enum constant and return Set containing
```
wmdietl: I think the original `TODO` was to actually use a `case MODULE:`. We are using the Java 9 compiler, so the constant should be available. When we tried that, some configuration didn't work. Maybe that configuration is fixed by now and we can actually use the proper `case`?

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -553,10 +553,8 @@ public static boolean hasInheritedMeta(AnnotationMirror anno) {
...
-                // ElementKind.MODULE.  (Java 11)
-                if (elementType.name().contentEquals("MODULE")) {
-                    return EnumSet.noneOf(ElementKind.class);
+                if (elementType.name().equals("MODULE")) {
+                    return EnumSet.of(ElementKind.valueOf("MODULE"));
```
wmdietl: Using `valueOf` to get the constant instead of using `noneOf` definitely seems better to me.

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
smillst: No, we are using the Java 9 compiler, but still using the Java 8 version of `ElementType`.

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
smillst: This is now inconsistent with the Javadoc.  Could you please just delete the following lines in the Javadoc:
```
* <p>If the Element is MODULE, the empty set is returned. This is so that this method can
* compile with Java 8.
```

michaelhixson: Oops, was just getting to this now.  Thanks for handling it.

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
wmdietl: The `TODO` comment should remain, to mark this as a place to change when we require a Java 9+ JVM. Isn't that what the original TODO was for? Or should we introduce a different TODO comment?

michaelhixson: @wmdietl I added this `TODO`.  Will this suffice?

```java
// TODO: Use MODULE enum constants directly instead of looking them up by name.
// (Java 11)
```

Looking back, I think the original TODO had two components:
* Add case for `ElementType.MODULE` instead of looking it up by name in the default case.
* Return a set containing `ElementKind.MODULE` instead of returning an empty set.

This PR addresses 2 (which I think is the more important one) but not 1 (which I missed in my original reading).

wmdietl: Yes, that sounds right.

PR #3023, opened 2019-12-31
Explicitly override createSupportedTypeQualifiers in CompilerMessagesAnnotatedTypeFactory

kelloggm:
#2828 added code that caused an exception to be thrown if no supported type qualifiers are found for a given checker. The CompilerMessagesChecker extends another checker, which in my experience causes the usual reflective-loading mechanism to behave strangely.

In #2871, the PR's tests are [failing](https://dev.azure.com/typetools/checker-framework/_build/results?buildId=2487&view=logs&j=e7da885e-38cc-54b9-8049-a12030020bfb&t=9f830bd7-3da0-547c-7985-0984f9314dee) because the CompilerMessagesChecker is failing to initialize itself, with the error message indicating that no supported type qualifiers were found. I have no idea why #2871 triggered that problem; it doesn't touch that part of the code.

Regardless, that failure is fixed locally by merging in the changes from this PR. I considered adding them to #2871, but that seems wrong: these have nothing to do with whole-program inference, which is the subject of that PR. That PR does add some minor changes to `AnnotatedTypeFactory`, where the error is raised, but that code is purely additive and in logic that shouldn't be run until after a checker is initialized, so I can't imagine how it would trigger this crash.

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/compilermsgs/CompilerMessagesAnnotatedTypeFactory.java
+++ b/checker/src/main/java/org/checkerframework/checker/compilermsgs/CompilerMessagesAnnotatedTypeFactory.java
@@ -1,6 +1,11 @@
 package org.checkerframework.checker.compilermsgs;
 
+import java.lang.annotation.Annotation;
+import java.util.Arrays;
+import java.util.LinkedHashSet;
+import java.util.Set;
 import org.checkerframework.checker.compilermsgs.qual.CompilerMessageKey;
+import org.checkerframework.checker.compilermsgs.qual.UnknownCompilerMessageKey;
 import org.checkerframework.checker.propkey.PropertyKeyAnnotatedTypeFactory;
 import org.checkerframework.common.basetype.BaseTypeChecker;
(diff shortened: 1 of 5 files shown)


kelloggm:
Fascinatingly, the build for this PR is now failing for the same reason, but caused by the Interning Checker. I'm going to update this PR with those changes too, but that validates my hypothesis that these failures are actually unrelated to #2871, where they first showed up.

mernst:
@kelloggm I'll review when the tests are passing.

kelloggm:
Tests were failing because the Signature Checker wasn't running properly, either. When I did get it running, it failed, so I had to change some annotations.

smillst:
@kelloggm Can you reproduce these failures locally?  I cannot.

smillst:
@kelloggm I can reproduce this if I rollback AFU to the same commit as the head of your fork.  Please merge typetools/master into your fork of annotation-tools and these errors along with those in  #2871 should go away.  (My guess is https://github.com/typetools/annotation-tools/commit/76d7af89017edb3a922e36465d66858f5978138f is the commit that fixes this problem.)

PR #3031, opened 2020-01-03
Allow Resolver#findClassInPackage Class kinds beside just CLASS. (Fixes #3018)

### LINKED ISSUE #3018: Suspicious check in Resolver.findClassInPackage
Document or fix this suspicious-looking element kind check in `Resolver.findClassInPackage`.

https://github.com/typetools/checker-framework/blob/cff1a036a04f102079dba64f1fa9557aadea30d6/javacutil/src/main/java/org/checkerframework/javacutil/Resolver.java#L278-L289

Why does it check for exactly `ElementKind.CLASS`?  Why doesn't it do something like this instead?

```java
if (res.getKind().isClass() || res.getKind().isInterface()) {
    return (ClassSymbol) res;
} else {
    return null;
}
```

Its only caller is in `FlowExpressionParseUtil`.

https://github.com/typetools/checker-framework/blob/cff1a036a04f102079dba64f1fa9557aadea30d6/framework/src/main/java/org/checkerframework/framework/util/FlowExpressionParseUtil.java#L494-L498

It's not clear why that code would not want to find the class if the class was an enum, for example.  If there's a reason for this, then perhaps an inline comment should be added to state that reason.

I'm filing this issue because [@wmdietl asked me to](https://github.com/typetools/checker-framework/pull/3017#pullrequestreview-336575990) in #3017.

smillst:
Fixes #3018

### DIFF
--- a/javacutil/src/main/java/org/checkerframework/javacutil/Resolver.java
+++ b/javacutil/src/main/java/org/checkerframework/javacutil/Resolver.java
@@ -282,7 +282,7 @@ public class Resolver {
                             pck,
                             names.fromString(name),
                             Kinds.KindSelector.TYP);
-            if (res.getKind().isClass() || res.getKind().isInterface()) {
+            if (ElementUtils.isClassElement(res)) {
                 return (ClassSymbol) res;
             } else {
                 return null;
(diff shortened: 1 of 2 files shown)


javacutil/src/main/java/org/checkerframework/javacutil/Resolver.java
```diff
@@ -282,7 +282,7 @@ public Element findClass(String name, TreePath path) {
                             pck,
                             names.fromString(name),
                             Kinds.KindSelector.TYP);
-            if (res.getKind() == ElementKind.CLASS) {
+            if (res.getKind().isClass() || res.getKind().isInterface()) {
```
wmdietl: This is fine, but `ElementUtils.isClassElement` would be shorter and easier to read.

PR #3034, opened 2020-01-06
Added callback in BaseTypeVisitor.checkExtendsImplements

abslamp:
I am working on a checker that does not use `isSubtype` when comparing annotations in class declaration and extends/implements clauses. Adding a callback to allow future checkers to override this logic seems a good idea.


wmdietl:
I think in further discussions we found that this isn't currently needed.
@abslamp Please re-open if I misunderstood.

Issue #3032, opened 2020-01-04
false positive with method reference


msridhar:
Extracted from https://github.com/typetools/checker-framework/issues/3001#issuecomment-570026011

See test case below, which was extracted from [nomulus](https://github.com/google/nomulus).  Output with the `UnitsChecker`:

```
$ ~/git-repos/checker-framework/checker/bin/javac -processor org.checkerframework.checker.units.UnitsChecker InvoicingPipeline.java
InvoicingPipeline.java:78: error: [argument.type.incompatible] incompatible types in argument.
              .via(BillingEvent::getInvoiceGroupingKey));
                  ^
  found   : @UnknownUnits MapElements</*INFERENCE FAILED for:*/ ? extends @UnknownUnits Object, /*INFERENCE FAILED for:*/ ? extends @UnknownUnits Object>
  required: @UnknownUnits PTransform<? extends @UnknownUnits PInput super @UnitsBottom PCollection<@UnknownUnits BillingEvent>, /*INFERENCE FAILED for:*/ @UnknownUnits ? extends @UnknownUnits POutput>
1 error
```

This has nothing to do with the units checker; the same error comes up with the object construction checker as well.  Test input:

```java
import java.io.Serializable;
public class InvoicingPipeline {
  public static class PCollection<T> implements PValue {
    public <OutputT extends POutput> OutputT apply(
        String name, PTransform<? super PCollection<T>, OutputT> t) {
      return null;
    }
  }
  public abstract static class PTransform<InputT extends PInput, OutputT extends POutput> {
    public abstract void expand(InputT input);
  }
  static interface PInput {}
  static interface POutput {}
  interface PValue extends PInput, POutput {}
  static class BillingEvent {
    InvoiceGroupingKey getInvoiceGroupingKey() {
      return null;
    }
  }
  public static class MapElements<InputT, OutputT>
      extends PTransform<PCollection<? extends InputT>, PCollection<OutputT>> {
    public static <OutputT> MapElements<?, OutputT> into(final TypeDescriptor<OutputT> outputType) {
      return null;
    }
    public static <InputT, OutputT> MapElements<InputT, OutputT> via(
        final ProcessFunction<InputT, OutputT> fn) {
      return null;
    }
    public void expand(PCollection<? extends InputT> input) {}
  }
  static class TypeDescriptor<T> {
    public static <T> TypeDescriptor<T> of(Class<T> type) {
      return null;
    }
  }
  static class InvoiceGroupingKey {}
[...]

PR #3035, opened 2020-01-07
Without -AconservativeUninferredTypeArguments, NEVER issue error about uninferred type arguments

kelloggm:
This is an extension to the [hack](https://github.com/typetools/checker-framework/issues/979#issuecomment-279578221) that prevents most errors related to #979, which did not prevent the CF from issuing errors involving method references (and possibly in other scenarios we don't know about). The test case in this PR was provided by @msridhar in a [comment](https://github.com/typetools/checker-framework/issues/3001#issuecomment-570026011) on #3001.

The technique I'm using is to check the message to be printed in the error reporter, and not issue it if it includes the text about inference failing added by [`DefaultAnnotatedTypeFormatter`](https://github.com/typetools/checker-framework/blob/master/framework/src/main/java/org/checkerframework/framework/type/DefaultAnnotatedTypeFormatter.java#L389) to wildcards for which inference has failed, which @mernst suggested.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
+++ b/framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
@@ -1369,9 +1369,12 @@ public abstract class SourceChecker extends AbstractTypeProcessor
                 return;
             } else if (this.processingEnv.getOptions().containsKey("nomsgtext")) {
                 for (Object arg : args) {
-                    if (arg != null && arg instanceof AnnotatedTypeMirror.AnnotatedWildcardType) {
-                        if (((AnnotatedTypeMirror.AnnotatedWildcardType) arg)
-                                .isUninferredTypeArgument()) {
+                    if (arg != null && arg instanceof String) {
+                        if (((String) arg).contains("INFERENCE FAILED")) {
+                            return;
+                        }
+                    } else if (arg != null && arg instanceof AnnotatedTypeMirror) {
+                        if (((AnnotatedTypeMirror) arg).containsUninferredTypeArguments()) {
                             return;
                         }
                     }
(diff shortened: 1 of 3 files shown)


msridhar:
FYI I also opened #3032 for the same test case.  If there is a proper fix for that case that doesn't require fixing #979, we may want to use a different test for this change

smillst:
@kelloggm all-tests is failing.  Is that expected?

kelloggm:
@smillst It was not! Sorry about that. The problem was that my original solution didn't take into account that the tests run with `-Anomsgtext`. It should be fixed now to account for that case.

mernst:
@kelloggm all_tests seems to still be failing.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -1358,6 +1359,29 @@ private void message(
...
+        if (this.processingEnv.getOptions() != null
+                && !this.processingEnv
+                        .getOptions()
+                        .containsKey("conservativeUninferredTypeArguments")) {
+
+            if (messageText.contains("INFERENCE FAILED")) {
+                return;
+            } else if (this.processingEnv.getOptions().containsKey("nomsgtext")) {
+                for (Object arg : args) {
+                    if (arg != null && arg instanceof String) {
```
wmdietl: The two null checks for `arg` are redundant, as `instanceof` is only true for non-null values.

kelloggm:
@smillst I eventually decided to leave this open and directly work around the problem I showed you this morning. It's imperfect, but I think it does what we want it to for the most part, and I've left a pretty detailed comment of my reasoning in the code.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeMirror.java
```diff
@@ -681,6 +682,36 @@ public void clearAnnotations() {
         annotations.clear();
     }
 
+    /**
+     * @return whether this type or any component type is a wildcard type for which Java 7 type
+     *     inference is insufficient. See issue 979, or the documentation on AnnotatedWildcardType.
+     */
+    public boolean containsUninferredTypeArguments() {
+        return uninferredTypeArgumentScanner.visit(this);
```
smillst: This throws a NPE when `this` is a declared type with out any type arguments.  #3044 has a version of the visitor that works correctly.

kelloggm:
Closing in favor of #3044, which solves the same problem in a cleaner way.

PR #3042, opened 2020-01-09
Don't try to infer type argument for a method invocation returned by a lambda (Fixes #3036)

### LINKED ISSUE #3036: False positive, possibly related to type inference and lambdas
Reduced from one of the errors reported by @msridhar in  [#3001-comment](https://github.com/typetools/checker-framework/issues/3001#issuecomment-567616512). An error is issued by most (all?) type systems, including the Signature Checker and the Object Construction Checker. I'm looking into the root cause.

Test case:

```
import java.util.List;
import java.util.Set;
import java.util.Map;
import java.util.HashMap;
import java.util.stream.Collectors;

public class LambdaInStream {

    public Set<MyInnerClass> getDsData() {
        return null;
    }

    public static class MyInnerClass {
        public int getKeyTag() {
            return 5;
        }

        public String getDigest() {
            return null;
        }
    }

    private void write() {
        List<Map<String, ?>> dsData = 
                        getDsData()
                        .stream()
                        .map(dsData1 ->
                                        ImmutableMap.of(
                                                "keyTag", dsData1.getKeyTag(),
                                                "digest", dsData1.getDigest()))
                        .collect(Collectors.toList());
    }

    public static class ImmutableMap<K, V> extends HashMap<K, V> {
        public static <K, V> ImmutableMap<K, V> of(K k1, V v1, K k2, V v2) {
            return null;
        }
    }
}
```

Error:
```
> javac -processor signature framework/tests/all-systems/java8inference/LambdaInStream.java
framework/tests/all-systems/java8inference/LambdaInStream.java:30: error: [return.type.incompatible] incompatible types in return.
                                        ImmutableMap.of(
                                                       ^
  type of expression: @SignatureUnknown ImmutableMap<@Identifier String, @SignatureUnknown Serializable & @SignatureUnknown Comparable<?[ extends @SignatureUnknown Serializable & @SignatureUnknown Comparable<?[ extends @SignatureUnknown Object super @SignatureUnknown Void]> super @SignatureUnknown Void]>>
  method return type: @SignatureUnknown ImmutableMap<@SignatureUnknown String, @SignatureUnknown Serializable & @SignatureUnknown Comparable<?[ extends @SignatureUnknown Serializable & @SignatureUnknown Comparable<?[ extends @SignatureUnknown Object super @SignatureBottom Void]> super @SignatureBottom Void]>>
1 error
```

smillst:
These require Java 8 type argument inference, #979.

Fixes #3036.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/util/typeinference/DefaultTypeArgumentInference.java
+++ b/framework/src/main/java/org/checkerframework/framework/util/typeinference/DefaultTypeArgumentInference.java
@@ -136,10 +136,8 @@ public class DefaultTypeArgumentInference implements TypeArgumentInference {
 
         final Set<TypeVariable> targets = TypeArgInferenceUtil.methodTypeToTargets(methodType);
 
-        if ((pathToExpression.getParentPath() != null
-                        && pathToExpression.getParentPath().getLeaf() != null
-                        && pathToExpression.getParentPath().getLeaf().getKind()
-                                == Tree.Kind.LAMBDA_EXPRESSION)
+        if (TreeUtils.enclosingNonParen(pathToExpression).first.getKind()
+                        == Tree.Kind.LAMBDA_EXPRESSION
                 || (assignedTo == null
                         && TreeUtils.getAssignmentContext(pathToExpression) != null)) {
             // If the type of the assignment context isn't found, but the expression is assigned,
--- a/framework/src/main/java/org/checkerframework/framework/util/typeinference/DefaultTypeArgumentInference.java
+++ b/framework/src/main/java/org/checkerframework/framework/util/typeinference/DefaultTypeArgumentInference.java
@@ -143,7 +143,7 @@ public class DefaultTypeArgumentInference implements TypeArgumentInference {
                 || (assignedTo == null
                         && TreeUtils.getAssignmentContext(pathToExpression) != null)) {
             // If the type of the assignment context isn't found, but the expression is assigned,
-            // then don't attempt to infere type arguments, because the Java type inferred will be
+            // then don't attempt to infer type arguments, because the Java type inferred will be
             // incorrect.  The assignment type is null when it includes uninferred type arguments.
             // For example:
             // <T> T outMethod()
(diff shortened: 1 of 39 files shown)


framework/src/main/java/org/checkerframework/framework/util/typeinference/DefaultTypeArgumentInference.java
```diff
@@ -135,7 +136,12 @@ public DefaultTypeArgumentInference(AnnotatedTypeFactory typeFactory) {
 
         final Set<TypeVariable> targets = TypeArgInferenceUtil.methodTypeToTargets(methodType);
 
-        if (assignedTo == null && TreeUtils.getAssignmentContext(pathToExpression) != null) {
+        if ((pathToExpression.getParentPath() != null
```
wmdietl: A comment about the first case would be good in the documentation below.

framework/src/main/java/org/checkerframework/framework/util/typeinference/DefaultTypeArgumentInference.java
```diff
@@ -135,7 +136,12 @@ public DefaultTypeArgumentInference(AnnotatedTypeFactory typeFactory) {
...
+                        && pathToExpression.getParentPath().getLeaf() != null
+                        && pathToExpression.getParentPath().getLeaf().getKind()
```
wmdietl: I'm wondering whether using `TreeUtils.enclosingMethodOrLambda` and then checking for `LAMBDA_EXPRESSION` would be a bit shorter.

smillst: The type arguments cannot be inferred if the immediate parent is lambda expression.  `TreeUtils.enclosingMethodOrLambda` might return a lambda expression that is the parent's parent, so it doesn't work.

wmdietl: Couldn't the method invocation be parenthesized?
Couldn't one write `mapper = dsData1 -> ((ImmutableMap.of(...)))`? Not the nicest style, but would be a bit odd not to handle.

PR #3044, opened 2020-01-10
If a type has an uninferred type arg, only check primary annotations

### LINKED ISSUE #3032: false positive with method reference

smillst:
This suppresses more warnings related to #979. 
This fixes #3032.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java
@@ -154,6 +154,25 @@ public class DefaultTypeHierarchy extends AbstractAtmComboVisitor<Boolean, Void>
     @Override
     public boolean isSubtype(
             final AnnotatedTypeMirror subtype, final AnnotatedTypeMirror supertype) {
+        if (subtype.atypeFactory.ignoreUninferredTypeArguments
+                && (subtype.containsUninferredTypeArguments()
+                        || supertype.containsUninferredTypeArguments())) {
+            if (ignoreUninferredTypeArgument(subtype) || ignoreUninferredTypeArgument(supertype)) {
+                // If either is an uninferred type argument don't check primary annotations.
+                return true;
+            }
+            // Check primary annotations.
+            for (final AnnotationMirror top : qualifierHierarchy.getTopAnnotations()) {
+                AnnotationMirror subAnno = subtype.getAnnotationInHierarchy(top);
+                AnnotationMirror superAnno = supertype.getAnnotationInHierarchy(top);
+                if (subAnno != null
+                        && superAnno != null
+                        && !qualifierHierarchy.isSubtype(subAnno, superAnno)) {
+                    return false;
+                }
+            }
+            return true;
+        }
         for (final AnnotationMirror top : qualifierHierarchy.getTopAnnotations()) {
             if (!isSubtype(subtype, supertype, top)) {
                 return false;
--- a/framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java
@@ -154,25 +154,6 @@ public class DefaultTypeHierarchy extends AbstractAtmComboVisitor<Boolean, Void>
     @Override
     public boolean isSubtype(
             final AnnotatedTypeMirror subtype, final AnnotatedTypeMirror supertype) {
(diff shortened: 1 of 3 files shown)


kelloggm:
I think this is a much better solution than #3035. I'm closing that PR in favor of this one.

framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java
```diff
@@ -154,6 +154,25 @@ protected StructuralEqualityComparer createEqualityComparer() {
...
+            for (final AnnotationMirror top : qualifierHierarchy.getTopAnnotations()) {
+                AnnotationMirror subAnno = subtype.getAnnotationInHierarchy(top);
+                AnnotationMirror superAnno = supertype.getAnnotationInHierarchy(top);
+                if (subAnno != null
+                        && superAnno != null
+                        && !qualifierHierarchy.isSubtype(subAnno, superAnno)) {
+                    return false;
+                }
+            }
+            return true;
```
wmdietl: I'm wondering whether this is a bit too coarse-grained.
With this change, even if somewhere deep within a generic type there is an uninferred type argument, only the primary annotations are checked.
Let's take a type `@A M<@B C, @D ?>` where the second type argument is uninferred.
With this change, only `@A` is checked. Wouldn't it be better if also `@B` is checked?
The first type argument isn't affected by the failure to infer the second argument.

What happens if you remove lines 157--159 and 164--174 from the patch? Methods like `isContainedBy` already check for uninferred type arguments, so maybe only the check in `isSubtype` was missing?
Then you wouldn't even need the new `containsUninferredTypeArguments` method.

smillst:
@wmdietl I was able to reduce the scope of this change a bit.  I'm going to go ahead and merge this.

PR #3051, opened 2020-01-15
Cleaner way to use a locally built version of the Checker Framework with Maven; fixes #3050

### LINKED ISSUE #3050: Maven instructions: suggest installing to local repo instead of using system dependence?
In the [instructions for using Maven](https://checkerframework.org/manual/#maven), the manual suggests to use a `system` dependence to make use of a local version of Checker Framework.  This approach is a bit problematic for a couple of reasons:
* The `system` dependence feature [is deprecated](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#System_Dependencies).
* When using third-party checkers by sticking them under `annotationProcessorPaths` in the `maven-compiler-plugin` config, the `system` dependence does not seem to override the dependence declared by the third-party checker.  E.g., when placing the Object Construction Checker in the `annotationProcessorPaths`, the compiler plugin still stuck v3.0.0 of the Checker Framework on the processor path, ignoring the declared `system` dependence.

An alternative approach that I found to work better was to install the relevant Checker Framework jars to the local Maven repository under a `SNAPSHOT` version number.  E.g., I did:
```
mvn install:install-file -DgroupId=org.checkerframework -DartifactId=checker -Dversion=3.1.1-SNAPSHOT -Dpackaging=jar -Dfile=path/to/checker-framework/checker/dist/checker.jar
```
Then, in my `annotationProcessorPaths` for `maven-compiler-plugin` I added:
```xml
            <path>
              <groupId>net.sridharan.objectconstruction</groupId>
              <artifactId>object-construction-checker</artifactId>
              <version>0.1.7-SNAPSHOT</version>
            </path>
            <path>
              <version>3.1.1-SNAPSHOT</version>
            </path>
```
This had the desired effect of overriding our checker's declared dependence on Checker Framework 3.0.  I think the `mvn install` approach should work fine for `checker-qual`, `jdk8`, etc. as well, though I haven't tested that.

Is there a strong reason to retain the `system` approach?

mernst:


msridhar:
I'd like to take some time to think through this a bit more, to see if there are any other alternatives.  A downside of this approach is that every time you make a change to the Checker Framework you need to re-run the `mvn install` command; maybe there is a better way.  Also we should make sure we change the Maven example to be consistent with the documentation and that everything works.  Assuming it's not urgent I'll try to take a closer look sometime over the next week or two.

mernst:
@msridhar Thanks!  There is no urgency.  We could consider creating a Gradle target that does the `mvn install` command, to make things simpler for developers.

mernst:
@msridhar A ping to prevent this from falling through the cracks.

msridhar:
Thanks @mernst I am looking into this

msridhar:
See #3112.  If we agree that's a good approach, we should able to simplify the instructions here.

mernst:
Closing in favor of #3112.

PR #3052, opened 2020-01-16
Add examples based on OOPSLA 2016 paper on Java unsoundness.

wmdietl:
I manually entered and adapted the examples.
Are there any copyright concerns to consider?


wmdietl:
Created #3251 as a follow-up to look at the ClassCastExceptions.

PR #3063, opened 2020-01-25
Sound or unsound treatment of collection Object parameters; fixes #3040

### LINKED ISSUE #3040: Add NullFriendlyCollections.astub stub file
Consider a method such as `Map.remove()`.

Some versions of `Map` are null-hostile and will throw an exception if null is passed to `remove()`.
Other versions of `Map` are null-friendly and do not throw an exception if null is passed to `remove()`.

It is unsound to annotate `Map.remove()` as taking nullable arguments:
```
  default boolean remove(@Nullable Object key, @Nullable Object value) {
```
because code that passes the Nullness Checker may still throw a NullPointerException.
Another way of explaining why this is unsound is that `@Nullable` means, "No null pointer exception will be thrown due to null being passed as an argument.", and that is not true in this case.

We would like to annotate it more precisely and permissively than only accepting non-null arguments, however.  https://github.com/typetools/checker-framework/issues/1326 is an issue for this.

Until #1326 is resolved, a workaround is to have a stub file that assumes that all collections are null-friesdly.
This stub file can be used by people who wish to reduce the number of false positives when type-checking code with null-friendly collections, and do not care about the unsoundness.

mernst:
Merge together with https://github.com/typetools/jdk/pull/34 and https://github.com/plume-lib/plume-util/pull/50


mernst:
@cpovirk Could you review this?

checker/jdk/nullness/src/java/util/AbstractList.java
```diff
@@ -181,7 +181,7 @@ public E remove(int index) {
      * @throws NullPointerException {@inheritDoc}
      */
     @Pure
-    public int indexOf(@Nullable Object o) {
+    public int indexOf(Object o) {
```
cpovirk: At a high level, I see 3 cases:

1. the interfaces themselves (like `List`)
2. the skeletal abstract classes (like `AbstractList`)
3. the concrete implementations (like `ArrayList`)

For (1), we're here exactly because we're aiming to make conservative assumptions by default (though allow them to be loosened with stub files). So we don't want `@Nullable` on parameters there.

Jumping ahead to (3), I would have guessed that classes like `ArrayList` _would_ have `@Nullable` on their parameters, since their implementations support null inputs and their docs reflect this (and the Checker Framework supports generalizing parameter types). I can see a couple possible arguments against that, though I have reservations about each:

- First, users rarely store objects in variables/fields whose static type is `ArrayList`, anyway, so the annotations on `ArrayList` itself rarely matter. (The argument about static types might become somewhat less true as users adopt `var`.)
- Second, users can extend `ArrayList` and override this method to reject null inputs. Taken to the extreme, this seems like an argument against putting `@Nullable` on a parameter of _any_ extensible method -- and an argument _for_ putting `@Nullable` on the _return value_ of every extensible method (since anyone might override it to return `null`). So I think at some point we have to take a stand on which of those overrides are valid and which aren't. Given that `ArrayList` doesn't document that its methods throw `NullPointerException`, I'd expect us to permit `null` here (unless the other argument above this one holds sway).

Coming back to (2), I think that's more like (1): Even though a given implementation might support null, the classes that override it might not, and the docs reflect this.

So the interesting question is likely to be (3).

checker/src/main/java/org/checkerframework/checker/nullness/collection-object-parameters-may-be-null.astub
```diff
@@ -0,0 +1,152 @@
...
+public interface Map<K, V> {
+    boolean containsKey(@Nullable Object key);
+    boolean containsValue(@Nullable Object value);
+    @Nullable V get(@Nullable Object key);
+    @Nullable V remove(@Nullable Object key);
+    default V getOrDefault(@Nullable Object key, V defaultValue);
+    default boolean remove(@Nullable Object key, @Nullable Object value);
+    }
+
+public class PriorityQueue<E> extends AbstractQueue<E>
```
cpovirk: Do these stubs still need bounds, like `<E extends @NonNull Object>` in this case?

checker/jdk/nullness/src/java/util/ArrayList.java
```diff
@@ -290,7 +290,7 @@ public boolean isEmpty() {
      * @return <tt>true</tt> if this list contains the specified element
      */
     @Pure
-    public boolean contains(@Nullable Object o) {
```
smillst: This should stay `@Nullable` because `null` is allowed according to the Javadoc. That's true for all the changes in this file.

checker/src/main/java/org/checkerframework/checker/nullness/collection-object-parameters-may-be-null.astub
```diff
@@ -0,0 +1,152 @@
...
+package java.util;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public abstract class AbstractList<E> extends AbstractCollection<E> implements List<E> {
+    public int indexOf(@Nullable Object o);
+    public int lastIndexOf(@Nullable Object o);
+}
+
+public class ArrayList<E> extends AbstractList<E>
```
smillst: This class can be removed if the annotations on ArrayList in the annotated JDK are changed.

docs/manual/nullness-checker.tex
```diff
@@ -96,6 +96,20 @@
   See Section~\ref{nullness-arrays} for a discussion.
   % TODO: this option is temporary and the sound option
   % should become the default.
+
+\item
+  \label{collection-object-parameters-may-be-null}
+  If you supply the
+  \code{-Astubs=collection-object-parameters-may-be-null.astub}
```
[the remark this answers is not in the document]
smillst: Because `collection-object-parameters-may-be-null.astub` is in the same directory as the Nullness Checker, it will automatically be found. You only need to add `checker.jar/` is the stub file is in the top level directory of the jar.  (Also, good to ask, though.)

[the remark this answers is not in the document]
mernst: `javadoc.astub` is at the top level because it has annotations for multiple type systems.

checker/jdk/nullness/src/java/util/AbstractList.java
mernst: My analysis is similar to yours:  @Nullable on any class that makes a commitment to handle null, such as `ArrayList`, but not on classes that don't make a commitment (like most interfaces and abstract classes).

You are right that `ArrayList` should have `@Nullable` on its parameters.  Sorry for that mistake; I have corrected it.  (It looks like I overlooked it because it uses slightly different wording about null than other classes.)

> users can extend ArrayList and override this method to reject null inputs.
Subtypes of `ArrayList` must also permit null values.  If a subclass of `ArrayList` were to forbid null elements, it would violate a commitment made by `ArrayList`, would violate behavioral subtyping, and the Nullness Checker would yield an error when type-checking it.

Are there any other classes that are incorrectly annotated?

checker/src/main/java/org/checkerframework/checker/nullness/collection-object-parameters-may-be-null.astub
mernst: They do not need them.  Because nothing is written there, the bounds from the annotated JDK are used.
It would do no harm to write them, and if you think it would be clearer to do so, I can do that.

checker/jdk/nullness/src/java/util/concurrent/ConcurrentLinkedDeque.java
```diff
@@ -1101,7 +1101,7 @@ public boolean removeLastOccurrence(Object o) {
      * @param o element whose presence in this deque is to be tested
      * @return {@code true} if this deque contains the specified element
      */
-    @Pure public boolean contains(@Nullable Object o) {
+    @Pure public boolean contains(Object o) {
```
cpovirk: Docs and testing suggest that the parameter to this `contains` can be `null`. Note that that's _not_ true of `remove` below, so definitely keep that one unchanged, as you currently have it.

This does raise an additional question -- whether it makes sense to put `@Nullable` here when `contains(null)` would always return safely but only by returning `false`. The advantage would be for a user who happens to have, say, a `@Nullable E` to pass to this method. Supporting such users is much the reason that Java [made `contains` accept an `Object`](https://errorprone.info/bugpattern/CollectionIncompatibleType#footnote-would-requiring-e-have-been-better) in the first place, rather than an `E`. But I'm not sure if it's best to emulate that or not.

checker/jdk/nullness/src/java/util/concurrent/ConcurrentLinkedQueue.java
```diff
@@ -491,7 +491,7 @@ else if (p == q)
      * @param o element to be removed from this queue, if present
      * @return {@code true} if this queue changed as a result of the call
      */
-    public boolean remove(@Nullable Object o) {
+    public boolean remove(Object o) {
```
cpovirk: In this file, both `contains` and `remove` can accept `null`.

checker/jdk/nullness/src/java/util/Collection.java
```diff
@@ -323,7 +323,7 @@
      * @throws UnsupportedOperationException if the <tt>remove</tt> operation
      *         is not supported by this collection
      */
-    boolean remove(@Nullable Object o);
+    boolean remove(Object o);
 
 
     // Bulk Operations
```
cpovirk: We'd probably want for `containsAll`, `removeAll`, and `retainAll` to require `Collection<? extends Object>` (or `Collection<? extends @NonNull Object>` if it's necessary to be explicit here, but I don't think so?).

That also goes for the redeclarations on `List` and `Set`.

checker/jdk/nullness/src/java/util/AbstractMap.java
```diff
@@ -153,7 +153,7 @@ public boolean containsValue(@Nullable Object value) {
      */
     @Pure
     @EnsuresKeyForIf(result=true, expression="#1", map="this")
-    public boolean containsKey(@Nullable Object key) {
+    public boolean containsKey(Object key) {
```
cpovirk: Other methods in this class that would benefit from `@Nullable Object` parameters: `containsValue`, `get`, `remove`.

checker/jdk/nullness/src/java/util/PriorityQueue.java
```diff
@@ -18,8 +18,8 @@
   public boolean add(E a1) { throw new RuntimeException("skeleton method"); }
   public boolean offer(E a1) { throw new RuntimeException("skeleton method"); }
   public @Nullable E peek() { throw new RuntimeException("skeleton method"); }
-  public boolean remove(@Nullable Object a1) { throw new RuntimeException("skeleton method"); }
-  @Pure public boolean contains(@Nullable Object a1) { throw new RuntimeException("skeleton method"); }
+  public boolean remove(Object a1) { throw new RuntimeException("skeleton method"); }
```
cpovirk: These `remove` and `contains` methods both accept `null`.

checker/jdk/nullness/src/java/util/Collection.java
mernst: Great point!  I overlooked those methods.  Thanks for pointing them out.

checker/jdk/nullness/src/java/util/AbstractMap.java
mernst: `AbstractMap` specifies that a NullPointerException is possible from all of those methods.  :-(

mernst:
> There are classes that inherit a method and ideally would redeclare it to mark types as `@Nullable`.

I agree, but this is not currently possible.  See #3094.

mernst:
@cpovirk Thanks for the helpful feedback.  I have revised the pull request.  Could you please take another look?

checker/jdk/nullness/src/java/util/Vector.java
```diff
@@ -880,7 +881,7 @@ public void clear() {
      * @throws NullPointerException if the specified collection is null
      */
     @Pure
-    public synchronized boolean containsAll(Collection<?> c) {
+    public synchronized boolean containsAll(Collection<? extends @NonNull Object> c) {
```
cpovirk: The docs suggest that these `Vector` bulk operations throw NPE only if the _collection_ is `null`. They do look like they accept null _element_.

checker/jdk/nullness/src/java/util/concurrent/ConcurrentHashMap.java
```diff
@@ -4519,7 +4519,7 @@ public final String toString() {
             return sb.append(']').toString();
         }
 
-        public final boolean containsAll(Collection<?> c) {
+        public final boolean containsAll(Collection<? extends @NonNull Object> c) {
```
cpovirk: It looks like these methods probably accept `null` elements in practice.

The only documentation of this fact is that they inherit the doc from `Collection.containsAll`, etc. and don't request to _also_ inherit the `@throws` doc.

That could easily be accidental, especially because this class didn't use to be publicly accessible. Well, _this class_ still isn't publicly accessible, but its subclass `ConcurrentHashMap.KeySetView` is as of 1.8 after being hidden before.

It's also a bit surprising that the collection views' query methods permit null elements when methods on `ConcurrentHashMap` itself do not.

I'm not sure what you want to do with all that :) But that's my read of the situation.

checker/jdk/nullness/src/java/util/concurrent/ConcurrentSkipListSet.java
```diff
@@ -335,7 +335,7 @@ public boolean equals(@Nullable Object o) {
      * @throws NullPointerException if the specified collection or any
      *         of its elements are null
      */
-    public boolean removeAll(Collection<?> c) {
+    public boolean removeAll(Collection<? extends @Nullable Object> c) {
```
cpovirk: I think you want `@NonNull` here.

checker/jdk/nullness/src/java/util/concurrent/SynchronousQueue.java
```diff
@@ -1028,7 +1028,7 @@ public boolean remove(Object o) {
      * @param c the collection
      * @return {@code false} unless given collection is empty
      */
-    @Pure public boolean containsAll(Collection<?> c) {
+    @Pure public boolean containsAll(Collection<? extends @NonNull Object> c) {
```
cpovirk: All the bulk methods on `SynchronousQueue`, plus the single-element `contains` and `remove`, are documented to suggest that null elements are permitted. (And an implementation would have to go out of its way to throw NPE for them.)

That said, I also grant that these bulk methods aren't documented to throw NPE for a null _collection_, and yet clearly doing so would be acceptable behavior (probably even _the only right behavior_ in the case of `containsAll`)....

checker/src/main/java/org/checkerframework/checker/nullness/collection-object-parameters-may-be-null.astub
```diff
@@ -0,0 +1,254 @@
...
+//
+//   ArrayDeque
+//     removeFirstOccurrence
+//     removeLastOccurrence
+//
+//   Collections
+//     frequency
+//
+// Special cases:
+//   EnumMap:
```
cpovirk: This section about `EnumMap` appears both here and above.

checker/src/main/java/org/checkerframework/checker/nullness/collection-object-parameters-may-be-null.astub
```diff
@@ -0,0 +1,254 @@
...
+    public boolean contains(@Nullable Object o);
+    public int indexOf(@Nullable Object o);
+    public int lastIndexOf(@Nullable Object o);
+    public boolean remove(@Nullable Object o);
+}
+
+public interface Collection<E> extends Iterable<E> {
+    boolean contains(@Nullable Object o);
+    boolean remove(@Nullable Object o);
+    boolean containsAll(Collection<? extends @Nullable Object> c);
```
cpovirk: I notice a mix of `Collection<?>` and `Collection<? extends @Nullable Object>` in this stub file. I don't know if that makes any difference to the tool (or to humans).

checker/src/main/java/org/checkerframework/checker/nullness/collection-object-parameters-may-be-null.astub
```diff
@@ -0,0 +1,254 @@
...
+    boolean addAll(Collection<? extends @Nullable E> c);
+    boolean retainAll(Collection<? extends @Nullable Object> c);
+}
+
+public interface Comparator<T> {
+    int compare(@Nullable T o1, @Nullable T o2);
```
cpovirk: I don't think we want `@Nullable` here: A `Comparator` for `@Nullable` `Foo` objects would be a `Comparator<@Nullable Foo>`.

checker/src/main/java/org/checkerframework/checker/nullness/collection-object-parameters-may-be-null.astub
```diff
@@ -0,0 +1,254 @@
...
+
+}
+
+class ConcurrentSkipListSet {
+    public boolean removeAll(Collection<?> c);
```
cpovirk: I believe this should forbid nulls.

checker/src/main/java/org/checkerframework/checker/nullness/collection-object-parameters-may-be-null.astub
cpovirk: I was wondering about this also. In fact, it looks like the annotations on `ArrayList` in the annotated JDK already match what's here.

Not that there's any harm in redeclaring them, especially if it's easier for you to generate or organize the stub file that way.

I think I saw some other classes that were technically redundant, like `PriorityQueue`.

checker/src/main/java/org/checkerframework/checker/nullness/collection-object-parameters-may-be-null.astub
```diff
@@ -0,0 +1,254 @@
...
+}
+
+public class ConcurrentSkipListSet<E>
+    extends AbstractSet<E>
+    implements NavigableSet<E>, Cloneable, java.io.Serializable {
+    public boolean contains(@Nullable Object o);
+    public boolean remove(@Nullable Object o);
+}
+
+public class CopyOnWriteArrayList<E>
```
cpovirk: I notice that you don't have this (and probably other classes?) in the annotated JDK currently. That seems perfectly fine -- the annotated JDK is known to not be 100% complete :) -- but this is one case in which the stub _would_ be redundant with the annotated JDK but isn't yet.

checker/jdk/nullness/src/java/util/concurrent/SynchronousQueue.java
mernst: > All the bulk methods on SynchronousQueue, plus the single-element contains and remove, are documented to suggest that null elements are permitted.

Can you point me at that documentation?  I must be overlooking it, because I can't seem to find it.

checker/src/main/java/org/checkerframework/checker/nullness/collection-object-parameters-may-be-null.astub
mernst: That would be the same as the annotation in the JDK.
This is trying to be more lenient and permit collections of nullable type.

Are you proposing that the stub file in general should not relax the signature of `removeAll`?

checker/jdk/nullness/src/java/util/concurrent/SynchronousQueue.java
cpovirk: Sorry, "documented" in the sense that these methods do not include `@throws NullPointerException`, in contrast to methods like `put`. But as noted, bulk methods like `containsAll` _also_ don't include `@throws NullPointerException` for when the _collection itself_ is `null`, so we might not want to read too much into the lack of a `@throws` clause.

checker/src/main/java/org/checkerframework/checker/nullness/collection-object-parameters-may-be-null.astub
cpovirk: Right, I'm proposing to match the annotated JDK. `ConcurrentSkipListSet`, even with a `Comparator` that permits `null`, has documentation on methods like `remove` and `removeAll` to suggest that they reject `null`. This is in contrast to `TreeSet`, which allows null queries if its `Comparator` does.

checker/jdk/nullness/src/java/util/concurrent/SynchronousQueue.java
mernst: Thanks for the clarification.  I don't want to read too much into that.  I'll leave them as is in the annotated JDK for now.

checker/src/main/java/org/checkerframework/checker/nullness/collection-object-parameters-may-be-null.astub
mernst: Thanks.  This comment only applies to ConcurrentSkipList and not to other methods, so I will only change ConcurrentSkipList by removing these lines from the stub file.

checker/jdk/nullness/src/java/util/concurrent/ConcurrentHashMap.java
mernst: As with the other example, I won't take (possibly accidental) omission of `@throws` to mean that null is a legal argument.

checker/jdk/nullness/src/java/util/Map.java
```diff
@@ -233,7 +233,7 @@
     // collections).  Some other implementation do accept nulls and are so
```
cpovirk: You probably want to remove the big comment here now (or revise it if you want to continue to call out this policy on `Map.get` specifically).

checker/jdk/nullness/src/java/util/concurrent/BlockingDeque.java
```diff
@@ -411,7 +411,7 @@ boolean offerLast(E e, long timeout, TimeUnit unit)
      * @throws NullPointerException if the specified element is null
      *         (<a href="../Collection.html#optional-restrictions">optional</a>)
      */
-    boolean removeFirstOccurrence(Object o);
+    boolean removeFirstOccurrence(@Nullable Object o);
```
cpovirk: I don't think we want `@Nullable` on these 4 methods in `BlockingDeque`, nor on the 2 in `BlockingQueue`.

checker/jdk/nullness/src/java/util/concurrent/ConcurrentSkipListSet.java
```diff
@@ -225,7 +225,7 @@ public ConcurrentSkipListSet(SortedSet<E> s) {
      *         compared with the elements currently in this set
      * @throws NullPointerException if the specified element is null
      */
-    @Pure public boolean contains(Object o) {
+    @Pure public boolean contains(@Nullable Object o) {
```
cpovirk: Similarly, I don't think we want `@Nullable` on `contains` and `remove` in `ConcurrentSkipListSet`.

checker/jdk/nullness/src/java/util/concurrent/CopyOnWriteArrayList.java
```diff
@@ -0,0 +1,1698 @@
...
+     * Removes from this list all of its elements that are contained in
+     * the specified collection. This is a particularly expensive operation
+     * in this class because of the need for an internal temporary array.
+     *
+     * @param c collection containing elements to be removed from this list
+     * @return {@code true} if this list changed as a result of the call
+     * @throws ClassCastException if the class of an element of this list
+     *         is incompatible with the specified collection
+     *         (<a href="../Collection.html#optional-restrictions">optional</a>)
+     * @throws NullPointerException if this list contains a null element and the
```
cpovirk: Given how detailed the spec is about `NullPointerException` for `removeAll` and `retainAll` (and arguably even `containsAll`), I could see accepting collections with null elements. But it's a judgment call.

(_If_ you buy that, it would apply to <code>CopyOnWriteArray**Set**</code>, as well.)

Also, this prompted me to file https://github.com/typetools/checker-framework/issues/3197, which might be helpful as part of mounting an argument for `Collection.contains` and friends to always accept `null`.

checker/src/main/java/org/checkerframework/checker/nullness/collection-object-parameters-may-be-null.astub
```diff
@@ -0,0 +1,215 @@
...
+    public synchronized boolean removeAll(Collection<?> c);
+    public synchronized boolean retainAll(Collection<?> c);
+}
+
+package java.util.concurrent;
+
+public class ConcurrentLinkedDeque<E>
+    extends AbstractCollection<E>
+    implements Deque<E>, java.io.Serializable {
+    public boolean removeFirstOccurrence(@Nullable Object o);
```
cpovirk: Both of these are documented to always reject nulls.

checker/src/main/java/org/checkerframework/checker/nullness/collection-object-parameters-may-be-null.astub
```diff
@@ -0,0 +1,215 @@
...
+}
+
+public class ConcurrentSkipListSet<E>
+    extends AbstractSet<E>
+    implements NavigableSet<E>, Cloneable, java.io.Serializable {
+    public boolean contains(@Nullable Object o);
+    public boolean remove(@Nullable Object o);
+}
+
+public class CopyOnWriteArrayList<E>
```
cpovirk: If you don't want to permit `null` in `CopyOnWriteArrayList.containsAll`+`removeAll`+`retainAll` in the annotated JDK, then you could do it here.

I haven't grepped through to find all classes that override those methods to see which others might benefit from similar changes; I just happened to notice this one.)

checker/src/main/java/org/checkerframework/checker/nullness/collection-object-parameters-may-be-null.astub
mernst: The .astub file is documented as
```
// This file relaxes the specifications for the following methods:
...
removeFirstOccurrence(Object)
...
```
It sounds like you would like to change that to,
"This file relaxes the specifications for the following methods when they are documented to possibly throw NPE, but does not change the specifications when they are documented to definitely throw NPE."

Is that correct?

checker/jdk/nullness/src/java/util/concurrent/CopyOnWriteArrayList.java
mernst: You are right (in #3197) that the current annotations are unsound.  Let's put that aside for the moment, because it requires a fix for the "annotations on receiver type parameters are ignored" problem.

I think this needs to stand as is in the annotated JDK, conservatively requiring non-null element types (the second condition in #3197).  Our goal is that if the client code satisfies the annotations, then there is no possibility of a NullPointerException.
The fix in #3197 would be more precise.

Does that make sense?  Or am I missing something?

cpovirk: Having thought about this more in #3197, I think the more conservative option is actually to _require_ a nullable element type for the parameter. That ensures that `copyOnWriteArraySet.removeAll(c)` can safely call `c.contains(null)`.

But I'm not sure if you want to go back and tweak a bunch of `removeAll` methods at this point, especially if you'd just be re-tweaking them after annotations on receiver type parameters start to work.

checker/src/main/java/org/checkerframework/checker/nullness/collection-object-parameters-may-be-null.astub
cpovirk: Yes, that's how I was thinking of it. I think that's most the useful approach for people who want to compromise safety, since it permits all potentially safe calls but no definitely unsafe calls. Plus, I think we are currently in an intermediate state, since we don't relax signatures for other classes like `ConcurrentHashMap` when we know some of their methods always reject null.

mernst: I agree that's better.  I will change the meaning of this pull request, as documented in that file.

checker/jdk/nullness/src/java/util/concurrent/CopyOnWriteArrayList.java
mernst: I would like to leave this until after annotations on receiver type arguments are no longer ignored.  Thanks for raising it, though.

checker/src/main/java/org/checkerframework/checker/nullness/collection-object-parameters-may-be-null.astub
cpovirk: You're generally postponing work on `removeAll` and `retainAll`, but you could consider handling `containsAll` now if you wanted.

I think that's the only thought I have left!

mernst: I added `removeAll` and `retainAll` to the stub file (maximally permissively), but I'm deferring work on them in the annotated JDK.  I made `containsAll` more permissive, either in the annotated JDK or in the stub file, depending on the specification.

checker/tests/nullness/WeakHasherMapNullable.java
```diff
@@ -0,0 +1,20 @@
+import java.util.AbstractMap;
+import java.util.Map;
+import org.checkerframework.checker.initialization.qual.*;
+import org.checkerframework.checker.nullness.qual.*;
+import org.checkerframework.checker.regex.qual.*;
```
wmdietl: Why import all these annotations, even though they aren't used?
Why not import `@Pure`, which is used?

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -2491,11 +2491,16 @@ private void printGitProperties() {
         }
         gitPropertiesPrinted = true;
 
-        try (InputStream in = getClass().getResourceAsStream("/git.properties");
-                BufferedReader reader = new BufferedReader(new InputStreamReader(in)); ) {
-            String line;
-            while ((line = reader.readLine()) != null) {
-                System.out.println(line);
+        try (InputStream in = getClass().getResourceAsStream("/git.properties")) {
```
wmdietl: Not clear how this is related to this PR?

PR #3077, opened 2020-02-03
Refactor and document BoundsInitializer.InitializerVisitor

smillst:
and the classes it uses.


framework/src/main/java/org/checkerframework/framework/type/BoundsInitializer.java
```diff
@@ -279,32 +276,63 @@ private static void initializeExtendsBound(
...
+        /**
+         * A mapping from IntersectionType to its {@link AnnotatedIntersectionType}. The first time
+         * this visitor encounters an intersection it creates an annotated type and adds it to this
+         * map. The next time the intersection is encounter, the annotated type in this map is
+         * returned.
+         */
         private final Map<IntersectionType, AnnotatedIntersectionType> intersections =
                 new HashMap<>();
+        /**
+         * Mapping from {@link TypeVariable} to {@link AnnotatedTypeMirror}. The annotated type
```
wmdietl: Could you add a comment why this is a mapping to `AnnotatedTypeMirror` instead of `AnnotatedTypeVariable`?
From the previous two maps I expected an `AnnotatedTypeVariable` here.

framework/src/main/java/org/checkerframework/framework/type/BoundsInitializer.java
```diff
@@ -279,32 +276,63 @@ private static void initializeExtendsBound(
...
-         * that kicked this visitation off.
+         * The RecursiveTypeStructure corresponding to the first wildcard or type variable bound
+         * initialization that kicked this visitation off.
          */
-        private final BoundStructure topLevelStructure;
+        private final RecursiveTypeStructure topLevelStructure;
 
-        private BoundStructure currentStructure = null;
+        /**
+         * The {@link RecursiveTypeStructure} corresponding to the wildcard or type variable that is
```
wmdietl: When should an `@link` be used in javadoc and when not?
The documentation above for `topLevelStructure` doesn't link to the field type, whereas this does. Similar inconsistency for other javadoc below, where e.g. the key type is not linked, but the value type is.
Is it ever worth to `@link` these, as one can simply use the type from the signature?

framework/src/main/java/org/checkerframework/framework/type/BoundsInitializer.java
```diff
@@ -823,47 +756,121 @@ private static AnnotatedTypeMirror createAndSetLowerBound(final AnnotatedTypeVar
...
+                        node, currentPath);
+            }
+            currentPath.removeLeaf();
+        }
 
         /**
-         * A mapping of all BoundPaths to TypeVariables for all type variables contained within
-         * annotatedTypeVar.
+         * For all type variables contained with in the type variable or wildcard that this
+         * structure represents, this a list of the replacement {@link AnnotatedTypeVariable} for
```
wmdietl: The `this a list` part is unclear.

framework/src/main/java/org/checkerframework/framework/type/BoundsInitializer.java
```diff
@@ -874,259 +881,302 @@ public void addTypeVar(final TypeVariable typeVariable) {
...
 
-    /** Represents an enclosing type in a path. */
-    private static class EnclosingNode extends BoundPathNode {
+        /**
+         * Internal implementation of {@link #getType(AnnotatedTypeMirror)}.
+         *
+         * @param parent type that is sideffected by this method
+         * @return the annotated type at the location represented by this node in {@code type}
+         */
+        protected abstract AnnotatedTypeMirror getTypeInternal(AnnotatedTypeMirror parent);
```
wmdietl: The overrides of this method are all `public`. Should they also be `protected`?

PR #3082, opened 2020-02-10
Value Checker: Better validation of types. Improve documentation.

wmdietl:


framework/src/main/java/org/checkerframework/common/value/util/NumberUtils.java
```diff
@@ -124,4 +124,59 @@ private static TypeKind unboxPrimitive(TypeMirror type) {
         }
         return type.getKind();
     }
+
+    /**
+     * Return true if the argument AnnotatedTypeMirror is a (possibly boxed) integral type.
```
smillst: Should `AnnotatedTypeMirror` be deleted?  The argument to the is method is a `TypeMirror` or am I missing something?

wmdietl: I updated the two instances to correctly refer to `TypeMirror`.

PR #3087, opened 2020-02-11
Reuse existing functions for TypeKind unboxing and range determination

wmdietl:
The repository in #2942 isn't available anymore. This PR tries to finish that work.

### DIFF
--- a/framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
@@ -395,10 +395,10 @@ public class ValueAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
     }
 
     /**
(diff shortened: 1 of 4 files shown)


framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -437,54 +390,22 @@ public long getToValueFromIntRange(AnnotatedTypeMirror atm) {
...
-            default:
-                throw new BugInCF(
-                        "Tried to apply a default to an IntRange annotation that was neither an integral primitive nor a declared type.");
+        return Range.create(toPrimitiveIntegralTypeKind(type)).to;
+    }
+
+    /**
+     * Determine the primitive integral TypeKind for the given type.
+     *
+     * @param type the type to convert
```
smillst: This should mention that `type` must be a boxed integral type.

PR #3088, opened 2020-02-12
Fix #3083 IDE failing to import com.sun.* packages under Java 8

### LINKED ISSUE #3083: IDEs cannot import any com.sun.* packages under Java 8
This issue only happens when trying to open the CF source code in an IDE with Java 8 as the project SDK. Although gradle does not have any issues with compiling, the IDE will not be able to process the files properly and consequently generate ubiquitous warnings. However, the issue is gone when switching the project SDK to Java 11.

Example:
image

I was able to reproduce this issue on IDEA and Eclipse, with oracle JDK 1.8.0_241.

I had a discussion with @xingweitian, and we think the followings are the possible solutions:
1. Simply ask developers to use Java 11 in IDEs
2. Add `compileOnly "com.google.errorprone:javac:9+181-r4173-1"` to `build.gradle`, which contains those com.sun.* packages (Thanks to @xingweitian)

zcai1:
Link to issue details: typetools#3083
Reference: https://github.com/typetools/templatefora-checker/pull/21

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -120,7 +120,7 @@ allprojects {
         if (isJava8) {
             javacJar group: 'com.google.errorprone', name: 'javac', version: "$errorproneJavacVersion"
 
-            // provides backward-compatible packages for IDE to import
+            // provides backward-compatible com.sun.* packages for IDE to import
             compileOnly "com.google.errorprone:javac:9+181-r4173-1"
         }
 
--- a/build.gradle
+++ b/build.gradle
@@ -119,12 +119,12 @@ allprojects {
     dependencies {
         if (isJava8) {
             javacJar group: 'com.google.errorprone', name: 'javac', version: "$errorproneJavacVersion"
+
+            // provides backward-compatible packages for IDE to import
+            compileOnly "com.google.errorprone:javac:9+181-r4173-1"
         }
 
         errorproneJavac("com.google.errorprone:javac:$errorproneJavacVersion")
-
-        // provides backward-compatible packages for IDE to import
-        compileOnly "com.google.errorprone:javac:9+181-r4173-1"
     }
 
     // After all the tasks have been created, modify some of them.
--- a/build.gradle
+++ b/build.gradle
@@ -122,6 +122,9 @@ allprojects {
         }
 
         errorproneJavac("com.google.errorprone:javac:$errorproneJavacVersion")
+
+        // provides backward-compatible packages for IDE to import
+        compileOnly "com.google.errorprone:javac:9+181-r4173-1"
     }
 
     // After all the tasks have been created, modify some of them.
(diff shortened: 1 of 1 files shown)


xingweitian:
One of the failed tests in `all_tests_jdk11`:
```
tests.SignatureTest > run[all-systems/java8/lambda] FAILED
    java.lang.AssertionError: 0 out of 0 expected diagnostics were found.
    :-1: other: Note: Multiple elements named 'com.sun.tools.javac.code' in modules 'unnamed module, jdk.compiler' were found by javax.lang.model.util.Elements.getPackageElement.
    :-1: other: Note: Multiple elements named 'com.sun.tools.javac.code.Type' in modules 'unnamed module, jdk.compiler' were found by javax.lang.model.util.Elements.getTypeElement.
    While type-checking /__w/1/s/checker/tests/all-systems/java8/lambda/Issue450.java, /__w/1/s/checker/tests/all-systems/java8/lambda/Issue573.java, /__w/1/s/checker/tests/all-systems/java8/lambda/Issue1817.java, /__w/1/s/checker/tests/all-systems/java8/lambda/Lambda.java, /__w/1/s/checker/tests/all-systems/java8/lambda/Issue1681.java
```
It's weird... It seems that we add a duplicate `errorprone:javac`, but it indeed solves the problem when using IDEs.
@zcai1 Could you please have a look at the failed tests?

xingweitian:
@zcai1 I think this pull request is almost finished. The failure is because that you are using opprop/checker-framework-inference. Please use typetools/checker-framework-inference instead.

You can create a  new branch named "backward-compatible-packages-for-IDE" which is consistent with typetools/checker-framework-inference's master branch. Push it to your repo and restart the tests.

zcai1:
@xingweitian Thanks a lot for your help! I will double check if the code can be compiled under JDK8 even without this fix, and if the answer is yes, I may need to figure out the reason.

zcai1:
Turns out `configurations.javacJar` in `build.gradle` provides the same library in compilation, but the way we add it to classpath hinders IDE's view of the dependency.

xingweitian:
@zcai1 Is there any update of this PR?

zcai1:
Replacing the “javacJar” configuration with “compileOnly” could be a better solution, so I’d like to check if we can achieve that.

On Feb 24, 2020, at 13:33, Weitian Xing <notifications@github.com> wrote:

﻿

@zcai1<https://github.com/zcai1> Is there any update of this PR?

zcai1:
@xingweitian After some experiments and investigations, I believe those `Xbootclasspath` are indispensable at the moment, so I guess, unfortunately, this is the simpler and less error-prone solution. This PR is now ready for final review.

wmdietl:
@smillst will work on a nicer solution for #3083.

PR #3092, opened 2020-02-16
Move purity checking to common.purity

### LINKED ISSUE #2581: Purity warning checker name
It is surprising that when `-AcheckPurityAnnotations` and `-ArequirePrefixInWarningSuppressions` are supplied, purity warnings are prefixed by a specific checker name, such as "fenum:purity..." or "formatter:purity...".  I feel it would be better if they were prefixed by "purity:".

### LINKED ISSUE #2582: Purity checks are performed multiple times
When a user supplies `-AcheckPurityAnnotations` and multiple checkers, purity warnings are issued multiple times, because the purity checker is run multiple times.  The purity checker should be run just once, no matter how many checkers are given on the command line.

Maxi17:
Fixes #2581 
Fixes #2582

This PR followed the steps needed to fix these issues as described in this comment: [link](https://github.com/typetools/checker-framework/issues/2581#issuecomment-510179727).

A few notes:

- I added `@skip-test` to `Purity.java` temporarily because I don't know how to fix it. The `PurityOverrideChecker` (subclass of `OverrideChecker` from `BaseTypeVisitor`) is not being used at all even though `AcheckPurityAnnotations` is present, which is why the test fails.

- Before, the `checker/jtreg/index/UnnededSuppressionsTest.java` test is run using the index checker, which also run the purity checker by default, without `-AcheckPurityAnnotations` being present. The two warnings I removed from the expected `.out` file are no longer issued by the `IndexChecker`, therefore they shouldn't be checked for. (manual: `You can use the -AwarnUnneededSuppressions command-line option to issue a warning if a @SuppressWarnings did not suppress any warnings issued by the current checker.`)

- The majority of purity checking is performed in `org.checkerframework.dataflow`. The annotations are also there. I haven't moved them because I thought this is enough work for an initial PR. If this is necessary, I can do it.

Also, the documentation needs to be improved.

@kelloggm helped me with feedback.

@mernst


Maxi17:
plume_lib_jdk11 fails because of a `java.lang.OutOfMemoryError`.
I don't know why the guava tests are failing. `messageStore` being null on line 489 in `BaseTypeChecker` triggers a `NullPointerException`. I will try to figure out why.

kelloggm:
#3093 looks like a similar problem in `BaseTypeChecker`

mernst:
@Maxi17 could you specify what feedback (if any) you want from me?
Are you asking for help with debugging the problems, or reviews of the code (though it sounds like you already know of some issues that need to be fixed, so maybe my comments there would be premature).

I find it confusing that there are two identically-named files `PurityChecker.java`:

```
dataflow/src/main/java/org/checkerframework/dataflow/util/PurityChecker.java
framework/src/main/java/org/checkerframework/common/purity/PurityChecker.java
```

By convention, a class named `FooChecker` is a checker (for example, a subclass of `BaseTypeChecker`), so the former should probably be renamed.
But, why are they distinct?  Can you document the relationship between them?

Also, the documentation of `org.checkerframework.dataflow.util.PurityChecker` says that it is a visitor, but it seems that `PurityCheckerHelper` is the actual visitor.

Maxi17:
@mernst Yes, I need some help with `framework/tests/all-systems/java8/memberref/Purity.java` test. I am not sure how to fix it. The problem is that the error expected is not issued because the `PurityOverrideChecker` is not created when `PurityChecker` is added as a subchecker.

> I find it confusing that there are two identically-named files `PurityChecker.java`

Yes, I agree. I have another working version in which the `PurityCheckerHelper` is integrated in `PurityVisitor`, but this was not what the initial behavior was expected to be (in #2581 and #2582). I will make the change now that you mentioned it, and will add some documentation to support it. I will also include the annotations (`@Pure`, `@Deterministic` etc.) in common.purity.qual, aliased to the ones from dataflow package, to avoid breaking something.

mernst:
@Maxi17 When you have made the changes you mentioned, could you ping me again?  Thanks!

smillst:
> messageStore being null on line 489 in BaseTypeChecker triggers a NullPointerException

The test is using your fork of Guava.  If you merge typetools/master into your fork, this error will go away.

mernst:
> > messageStore being null on line 489 in BaseTypeChecker triggers a NullPointerException
> 
> The test is using your fork of Guava. If you merge typetools/master into your fork, this error will go away.

You can either update your fork of Guava, or delete it.  The CI setup assumes that if you have one, you want to use it.

Maxi17:
@mernst I made the changes. (I'm sorry for the delay)

`PurityUtils.java` can't be moved from `dataflow.util` to `common.purity` because `dataflow.analysis.FlowExpressions` uses methods from this class and moving it would make them unavailable to `FlowExpressions`.

`all_testes_jdk11` and `plume_lib_jdk11` tests didn't start executing and I couldn't find how to manually start them.

I would appreciate some help with the `Purity.java` test, as described above. Thanks.

mernst:
There seem to have been upstream changes that are not reflected in your pull request.  For example, there is an assumeDeterministic, similar to assumeSideEffectFree, but your code does not use and drops that functionality.  Could you please reinstate it?

I have added a lot of Javadoc to the master branch, which should ease your task of creating a new file (which CI requires to be fully documented).  I don't know whether it will be easier to merge in the changes from upstream or to start from a fresh branch of master.

Maxi17:
I will start a new branch from master, since the tests are failing locally for me because of recent changes in `SourceChecker.java`.

mernst:
> I will start a new branch from master, since the tests are failing locally for me because of recent changes in `SourceChecker.java`.

Sorry about that!  On the positive side, I think we resolved several of your questions, and the underlying code that you are refactoring is much better now.

Thanks again for this refactoring.

I like the structure, which is definitely an improvement.

My biggest concern is understanding why there are timeouts for all_tests_jdk11 and plume_lib_jdk11 but not the jdk8 versions of them, and not downstream_jdk11 or guava_jdk11.  I suspect the problem might be memory:  something is using a lot of memory.  The obvious culprit is the annotated JDK, since it is a big difference between the JDK 8 and JDK 11 versions of the Checker Framework, but I have not yet investigated that lead.  I will do so later, on your new branch.

Maxi17:
New PR on a fresh branch: #3151

PR #3099, opened 2020-02-20
Run CFI tests under JDK 11

mernst:
This should be assigned to @zcai1, but GitHub doesn't let me do that.

This is not ready to merge until:
 * the tests pass, and
 * the CFI CI jobs run tests under JDK 11, because we don't want to block commits to the CF repository due to a problem in the CFI repository.


mernst:
Will be done by @zcai1 in a different pull request.

PR #3100, opened 2020-02-21
Avoid the false "cannot find stub file" warnings.

xingweitian:
I am currently working on a compound checker (ValueChecker as the sub checker). When I ran the compound checker with a stub file (passed by "-Astubs" argument) which is in the jar file of that compound checker, I got a "cannot find stub file" warning. It is because that ValueChecker cannot find that stub file (which is not needed by the ValueChecker).

Here is a potential solution: let all the parent checkers (the parent checker, the parent of the parent checker, ...) try to find the target stub file. If one can find, there is no need to report a "cannot find stub file" warning.


xingweitian:
@smillst @wmdietl Could you please restart `downstream_jdk11`? I don't have the permission to do it. The failure of that checking is due to a connection timed out to plume-script repo.

framework/src/main/java/org/checkerframework/framework/stub/StubTypes.java
```diff
@@ -202,15 +202,31 @@ public void parseStubFiles() {
...
-                                        + new File(stubPath).getAbsolutePath()
-                                        + (stubPathFull.equals(stubPath)
-                                                ? ""
-                                                : (" or at " + stubPathFull)));
+                        // When using a compound checker, the target stub file may be found by the
+                        // current checker's parent checkers. Also check this to avoid a false
+                        // warning.
+                        SourceChecker parentChecker = checker.parentChecker;
+                        boolean findByParentCheckers = false;
+                        while (parentChecker != null) {
```
wmdietl: I think you should try to pull this check out before the first use of `checker.getClass()` on line 192 and simply perform the same logic for either the first checker or one of the parent checkers.
This would include checking for the `topLevelResource` for each possible checker.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -515,6 +515,11 @@ protected void setParentChecker(SourceChecker parentChecker) {
         this.parentChecker = parentChecker;
     }
 
+    /** @return the parent checker of the current checker */
+    public SourceChecker getParentChecker() {
+        return this.parentChecker;
+    }
+
     /** Invoked when the current compilation unit root changes. */
```
wmdietl: @xingweitian Can you add the missing `@param` to make the tests pass. You didn't touch that method, but it is "close" to what you added.

PR #3103, opened 2020-02-22
Added @Nullable to add()

mrinal-pandey:
add() can take a null value and append it to the list.


wmdietl:
Please read through the section of the manual that explains generics https://checkerframework.org/manual/#polymorphism
A `List<@NonNull String>` should not accept `null` values.

mrinal-pandey:
Thanks for pointing me in the right direction, apologies for the irrelevant PR.

PR #3106, opened 2020-02-25
Fall back to getURLFromClasspaths (Fixes #2893)

### LINKED ISSUE #2893: When searching for qualifiers, fall back to classpath search if classLoader.getResource returns null
We believe we've found a case where in the process of reflectively loading qualifiers, the following call to `classLoader.getResource` returns `null`:

https://github.com/typetools/checker-framework/blob/11a2869526052061c13fba25ff6ac157cf91c41b/framework/src/main/java/org/checkerframework/framework/type/AnnotationClassLoader.java#L132

That is trying to get a resource for the qualifiers directory.  Further, we found that getting a resource for a specific qualifier class in that directory gives a non-null result.  Specifically in our case, `classLoader.getResource("org/checkerframework/checker/returnsrcvr/qual")` returns `null`, but `classLoader.getResource("org/checkerframework/checker/returnsrcvr/qual/MaybeThis.class")` gives a non-null result.  Unfortunately, I don't have a test case to reproduce at the moment.

Should the code in `AnnotationClassLoader` fall back on trying to get the qualifier directory via `getURLFromClasspaths` if the above call returns `null`?  Not sure if that makes sense or would be guaranteed to fail if the above call returns `null`.

/cc @lazaroclapp

wmdietl:
@msridhar This implements the change suggested in #2893, but I have no way to test whether this actually does any good.
Could you try whether this fixes your original issue?

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/type/AnnotationClassLoader.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/AnnotationClassLoader.java
@@ -139,23 +139,31 @@ public class AnnotationClassLoader {
 
         classLoader = getClassLoader();
 
+        URL localResourceURL;
         if (classLoader != null) {
             // if the application classloader is accessible, then directly
             // retrieve the resource URL of the qual package
             // resource URLs must use slashes
-            resourceURL = classLoader.getResource(packageNameWithSlashes);
+            localResourceURL = classLoader.getResource(packageNameWithSlashes);
 
             // thread based application classloader, if needed in the future:
             // resourceURL =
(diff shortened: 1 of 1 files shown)


msridhar:
@wmdietl I'll have to do some work to allow for reproducing this.  It only showed up in @lazaroclapp's config, and we worked around it by listing qualifiers explicitly in the checker.  I'll see what I can do but it may take some time.

smillst:
As this does no harm, we are going to go ahead and merge it so that it will be in the next release.

PR #3112, opened 2020-02-26
[WIP] Add task to publish checker jar with dependencies to local maven repository

msridhar:
See #3050 and #3051 

With this additional Gradle task, we could ease use of local builds of the Checker Framework.  Now, by running `./gradlew :checker:publishToMavenLocal`,  the `checker/dist/checker.jar` file gets pushed to the local Maven repository, and can be referenced from other Maven builds via:

```xml
               <version>3.1.2-SNAPSHOT</version>
               <classifier>all</classifier>
```

The change is clearly a work in progress and not ready for review.  I'd just like some feedback on the overall approach.  @mernst @smillst @wmdietl


build.gradle
```diff
@@ -89,6 +89,7 @@ task installGitHooks(type: Copy, dependsOn: 'setLocalRepo') {
 
 allprojects {
     apply plugin: 'java'
+    apply plugin: 'maven-publish'
```
msridhar: We probably only want to apply the plugin to the `checker` module, but it needs to be applied before the shadow plugin.  So we need to add some kind of conditional here.

checker/build.gradle
```diff
@@ -375,3 +375,39 @@ task jtregJdk11Tests(dependsOn: ':downloadJtreg', group: 'Verification') {
 
     }
 }
+
+group = 'org.checkerframework'
+version = '3.1.2-SNAPSHOT'
```
msridhar: We need some way to pull this information from elsewhere

checker/build.gradle
```diff
@@ -375,3 +375,39 @@ task jtregJdk11Tests(dependsOn: ':downloadJtreg', group: 'Verification') {
...
+
+
+publishing {
+    publications {
+        shadow(MavenPublication) { publication ->
+            project.shadow.component(publication)
+            artifactId = 'checker'
+            pom {
+                name = 'My Library'
```
msridhar: Copy-pasted from Gradle documentation for now.  Ideally we'd pull this from elsewhere too.  But since this is only for a local install, the contents are not that important.

build.gradle
msridhar: Or actually, maybe we want this to apply to all the modules?  Not sure.  If we want it everywhere, we'll have to move and/or copy the pom configuration from `build/checker.gradle`

msridhar:
Other thoughts here:

* This could also be useful for users of the Checker Framework Gradle plugin who want to use a locally-built Checker Framework.  Rather than having to following [these instructions](https://github.com/kelloggm/checkerframework-gradle-plugin#specifying-a-checker-framework-version) and pass a `cfLocal` flag, set the `CHECKERFRAMEWORK` environment variable, etc., they could just install a `SNAPSHOT` version and reference that from their Gradle config.
* This could also be the basis for automatically creating SNAPSHOT jars in CI.  We do this for NullAway and WALA and it's quite useful for helping users to test fixes.  These wouldn't be full-fledged Checker Framework releases but they could still be helpful.

mernst:
I strongly support this approach.  It's much nicer than the current instructions with the `cfLocal` flag and the `CHECKERFRAMEWORK` environment variable.  So please do push forward with this.  Thanks!

msridhar:
Ok, sounds good!  @smillst I think I'll need your guidance as to how to take this change forward.  I think one good way to do things would be to use this Gradle support to create artifact jars in all scenarios (local installs, snapshot builds, releases), replacing the current way of creating release jars.  This way there is one source of truth regarding version numbers, POM metadata, etc.  This would probably work best if Checker Framework kept the version in the repo as a `SNAPSHOT` version between releases, and only switched to a non-`SNAPSHOT` version when cutting a release.

I know the overall Checker Framework release process is complex, so not sure if the above proposal is feasible.  Let me know your thoughts.

smillst:
We discussed this today and I'll working on this.  We are going to use the existing Maven pom.xml and call out to Maven.

msridhar:
Thanks, @smillst. Let me know how I can help.

smillst:
Closing in favor of #3196 .

PR #3114, opened 2020-02-26
Improve error message of inconsistent.constructor.type

abslamp:
The error message of key `inconsistent.constructor.type` seems not clear enough, thus updated to improve it.
Update: explicit show user the cause: not returning types with top qualifier

### DIFF
--- a/framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
+++ b/framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
@@ -721,10 +721,7 @@ public class BaseTypeVisitor<Factory extends GenericAnnotatedTypeFactory<?, ?, ?
                     qualifierHierarchy.findAnnotationInHierarchy(constructorAnnotations, top);
             if (!qualifierHierarchy.isSubtype(top, constructorAnno)) {
                 checker.report(
-                        Result.warning(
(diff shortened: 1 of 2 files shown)


framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -721,7 +721,10 @@ protected void checkConstructorResult(
                     qualifierHierarchy.findAnnotationInHierarchy(constructorAnnotations, top);
             if (!qualifierHierarchy.isSubtype(top, constructorAnno)) {
                 checker.report(
-                        Result.warning("inconsistent.constructor.type", constructorAnno),
+                        Result.warning(
+                                "inconsistent.constructor.type",
+                                constructorAnno,
+                                qualifierHierarchy.getTopAnnotation(constructorAnno)),
```
smillst: Shouldn't this just be `top`? Or am I missing something?

wmdietl: @abslamp Can you fix this?

abslamp: Thanks for feedback. It's now fixed.

PR #3115, opened 2020-02-27
Change the meaning of @SuppressWarnings("initialization")

mernst:


checker/bin-devel/test-daikon.sh
```diff
@@ -23,5 +23,6 @@ source $SCRIPTDIR/build.sh ${BUILDJDK}
 # daikon-typecheck: 15 minutes
 /tmp/plume-scripts/git-clone-related codespecs daikon
 cd ../daikon
+git log | head -n 5
```
wmdietl: Should this change be in this PR?

mernst: Moved to master.

PR #3122, opened 2020-03-02
Mark Iterator.hasNext() and Enumeration.hasMoreElements() as @Pure

mernst:


mernst:
Merge together with https://github.com/typetools/jdk/pull/30

PR #3123, opened 2020-03-02
Keep stub warnings to one line; stack traces truncate subsequent lines

mernst:


framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1662,7 +1654,7 @@ private Object getValueOfExpressionInAnnotation(
 
             return typeElement.asType();
         } else if (expr instanceof NullLiteralExpr) {
-            stubWarn("Null found as value for %s. Null isn't allowed as an annotation value", name);
+            stubWarn("Ilegal annotation value null, for %s", name);
```
wmdietl: `Ilegal` should be `Illegal`.

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1013,14 +1007,12 @@ private void annotate(
                     stubWarn(
                             "Wildcard type <"
                                     + atype
-                                    + "> doesn't match type in stubs file: <"
+                                    + "> doesn't match type in stubs file"
```
wmdietl: Should we in general avoid `doesn't` and other contractions? Not really relevant for this PR, but notice that in a message above it's "does not".

PR #3127, opened 2020-03-03
Remove `@DefaultInUncheckedCodeFor` and `@DefaultQualifierInHierarchyInUncheckedCode`

mernst:
This only removes the annotations.  Once this is approved, I will proceed to other refactorings and improvements.


changelog.txt
```diff
@@ -1,3 +1,12 @@
+Version 3.3.0, April 1, 2020
+
+Implementation details:
+ * Removed `@DefaultInUncheckedCodeFor` and
+   `@DefaultQualifierInHierarchyInUncheckedCode`.  The conservative default
```
smillst: Type system designer can override it via `org.checkerframework.framework.util.defaults.QualifierDefaults#addUncheckedCodeDefault`.  I think we want to keep that option, but not document it in the manual.

PR #3131, opened 2020-03-05
SourceChecker reorganization

mernst:
This pull request moves methods around in the file to improve the file's organization.
It renames a few methods and fields.
It adds documentation.
It abstracts out two methods (to make the calling method, which used to inline the code shorter and clearer).

There are no logic changes in the pull request.


framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -499,6 +504,31 @@ public final synchronized void init(ProcessingEnvironment env) {
         }
     }
 
+    /** True if the git.properties file has been printed. */
+    private static boolean gitPropertiesPrinted = false;
+
+    /** Print information about the git repository from which the Checker Framework was compiled. */
+    void printGitProperties() {
```
wmdietl: Should this method be private? Or protected? I don't see a need for package visibility.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -499,6 +504,31 @@ public final synchronized void init(ProcessingEnvironment env) {
         }
     }
 
+    /** True if the git.properties file has been printed. */
+    private static boolean gitPropertiesPrinted = false;
+
+    /** Print information about the git repository from which the Checker Framework was compiled. */
```
wmdietl: Could we move/keep these support/debugging methods to the bottom of the file? Is there a reason to move them up?

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -1066,178 +895,79 @@ public void typeProcess(TypeElement e, TreePath p) {
...
-                    // For example, @SuppressWarnings("purity") suppresses errors with keys:
-                    // purity.deterministic.void.method, purity.deterministic.constructor, etc.
-                    if (errorKey.contains(userKey)) {
-                        reportUnneededSuppression(tree, fullUserKey);
-                    }
-                }
-            }
-        }
-    }
+    ///////////////////////////////////////////////////////////////////////////
```
wmdietl: I would also move shutdown logic towards the bottom.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -1066,178 +895,79 @@ public void typeProcess(TypeElement e, TreePath p) {
...
-                        + ")"
-                        + ((p == null)
-                                ? ""
-                                : " while processing "
-                                        + p.getCompilationUnit().getSourceFile().getName())
-                        + (t.getMessage() == null ? "" : "; message: " + t.getMessage()),
-                t);
-    }
+    ///////////////////////////////////////////////////////////////////////////
+    /// Reporting type-checking errors; the main entry point is report()
```
wmdietl: I'm not sure what the "main entry point" is supposed to mean. Can you clarify?
The two methods that visitors should use are `report()` and `message(Diagnostic.Kind kind, String msg, Object... args)`, so maybe list those two?

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -2053,100 +1634,551 @@ public final String getOption(String name, String defaultValue) {
...
+    /**
+     * Issues a warning about any {@code @SuppressWarnings} that isn't used by this checker, but
+     * contains a key that would suppress a warning from this checker.
+     *
+     * @param elementsSuppress elements with a {@code @SuppressWarnings} that actually suppressed a
+     *     warning
+     * @param checkerKeys suppress warning keys that suppress any warning from this checker
+     * @param errorKeys error keys that can be issued by this checker
+     */
+    protected void warnUnneedSuppressions(
```
wmdietl: Should this method be called `warnUnneededSuppressions`, like the other method above?
How does a developer chose between the two version?

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
mernst: I made it private.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
mernst: It is only called by `init`, so I put it next to `init`.  I have now moved it to the end of the file.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
mernst: `message` is documented as being "only for exceptional cases".  So `report` is the method that clients should prefer.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
mernst: That was the name before, but I have changed it.

PR #3133, opened 2020-03-06
Throw error rather than just printing a warning

mernst:


framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -792,23 +792,14 @@ public void initChecker() {
...
-                    javax.tools.Diagnostic.Kind.ERROR, "Refusing to process empty TypeElement");
-            return;
-        }
-        if (p == null) {
-            messager.printMessage(
-                    javax.tools.Diagnostic.Kind.ERROR,
-                    "Refusing to process empty TreePath in TypeElement: " + e);
-            return;
+        if (e == null || p == null) {
+            throw new BugInCF("SourceChecker.typeProcess: e=%s, p=%s", e, p);
```
wmdietl: You can't use `BugInCF` up here, because it is outside of the corresponding try/catch that would do something useful with it. So in the unlikely event this would ever happen, an ugly stack trace will be shown.
The previous code produces separate understandable errors, which seems preferable.
How about undoing this change?

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -962,8 +953,9 @@ private void message(
         try {
             messageText = String.format(fmtString, args);
         } catch (Exception e) {
-            messageText =
-                    "Invalid format string: \"" + fmtString + "\" args: " + Arrays.toString(args);
+            throw new BugInCF(
```
wmdietl: This change looks good and will indeed help us catch format string errors much easier.

PR #3134, opened 2020-03-06
Use `List<DiagMessage>` instead of `framework.source.Result`

mernst:
The `Result` class isn't necessary; `List<DiagMessage>` suffices.  This pull request removes `Result`.

Merge together with https://github.com/typetools/checker-framework-inference/pull/146 .


framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -879,71 +877,34 @@ public void typeProcess(TypeElement e, TreePath p) {
...
+     * @param source the source position information; may be an Element, a Tree, or null
+     * @param kind the type of message
+     * @param messageKey the message key
+     * @param args arguments for interpolation in the string corresponding to the given message key
      */
-    public void report(final Result r, final Object src) {
-        if (r.isSuccess()) {
-            return;
-        }
+    public void report(
```
wmdietl: What do you think of introducing two methods `reportError` and `reportWarning` and discouraging the use of  the version with a `javax.tools.Diagnostic.Kind` dependency? That would make it easier to use the API, removing the need for many of the imports this PR adds.

mernst: I considered that and thought this variant a bit cleaner.  But I'll change it to your suggestion.  You are right that it will limit the number of imports needed.

mernst: I have done this and I agree it's better.  Thanks for the suggestion.
Please go ahead and re-review.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -875,75 +874,77 @@ public void typeProcess(TypeElement e, TreePath p) {
...
-     * @param src the position object associated with the result; may be an Element, a Tree, or null
+     * @param source the source position information; may be an Element, a Tree, or null
+     * @param messageKey the message key
+     * @param args arguments for interpolation in the string corresponding to the given message key
      */
-    public void report(final Result r, final Object src) {
-        if (r.isSuccess()) {
-            return;
-        }
+    public void reportError(Object source, @CompilerMessageKey String messageKey, Object... args) {
```
smillst: I'm using the Result class in #2883 to collect multiple errors and report them once:
https://github.com/typetools/checker-framework/pull/2883/files#diff-31033326203125d5a0af0323436d39f4R463

So, I would like to keep the class.  

I support changing the report methods so that you don't have to create a Result object only to pass it to this method.  I do think we should deprecate the previous methods though as this is a rather large breaking change.

mernst: @smillst I deprecated the Result class instead of removing it.

Instead of `Result`, use `List<DiagMessage>`.  (Currently a `Result` is essentially just a `List<DiagMessage>` plus some baggage.)

The refactoring is easy.  It's two commands in Emacs; I will make a pull request for you.  I'm happy to do it for any other branches or forks where people are concerned about merge conflicts.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -894,56 +895,79 @@ public void report(final Result r, final Object src) {
...
+     */
+    public void report(Object source, DiagMessage d) {
+        report(source, d.getKind(), d.getMessageKey(), d.getArgs());
+    }
+
+    /**
+     * Reports a diagnostic message. By default, it prints it to the screen via the compiler's
+     * internal messager; otherwise, the method returns with no side effects.
+     *
+     * <p>Most clients should use {@link #reportError} or {@link #reportWarning}.
```
wmdietl: As this is a private method, no external clients can use this.
So maybe this comment is better on the public `void report(Object source, DiagMessage d)` above?

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -894,56 +895,79 @@ public void report(final Result r, final Object src) {
...
+     * @param source the source position information; may be an Element, a Tree, or null
+     * @param d the diagnostic message
+     */
+    public void report(Object source, DiagMessage d) {
+        report(source, d.getKind(), d.getMessageKey(), d.getArgs());
+    }
+
+    /**
+     * Reports a diagnostic message. By default, it prints it to the screen via the compiler's
+     * internal messager; otherwise, the method returns with no side effects.
```
wmdietl: I'm a bit confused about the `By default` and `otherwise` text. This is a private method, so subclasses can't adapt it. How/when do I reach the `otherwise` case?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -3,6 +3,8 @@
 // The imports from com.sun are all @jdk.Exported and therefore somewhat safe to use.
 // Try to avoid using non-@jdk.Exported classes.
 
+import static javax.tools.Diagnostic.Kind.ERROR;
```
wmdietl: Am I missing where this is used in this file?

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
mernst: Via `printOrStoreMessage`.  I'll clarify the comment.

framework/src/main/java/org/checkerframework/framework/source/Result.java
```diff
@@ -19,7 +20,9 @@
  * compiler interface.
  *
  * @see SourceChecker#report
+ * @deprecated use {@code List<DiagMessage>} instead
```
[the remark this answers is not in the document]
mernst: A `Result` contains a list of `DiagMessage`.  (`Result.failure` returns a `Result` that contains only a single `DiagMessage`, but other methods create non-singletons.)

checker/src/main/java/org/checkerframework/checker/formatter/FormatterTreeUtil.java
```diff
@@ -351,16 +351,26 @@ public Boolean visitNull(NullType t, Class<Void> p) {
         }
     }
 
-    /** Reports an error. Takes a {@link Result} to report the location. */
-    public final <E> void failure(Result<E> res, @CompilerMessageKey String msg, Object... args) {
-        checker.report(
-                org.checkerframework.framework.source.Result.failure(msg, args), res.location);
+    /**
```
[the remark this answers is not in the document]
mernst: This class contains `Result<E>` which is a *different* `Result` than `org.checkerframework.framework.source.Result.failure` which has been removed.  (Yes, that's very confusing.)
In the future, maybe `Result<E>` should be refactored, but that is not a matter for this pull request.

framework/src/main/java/org/checkerframework/framework/source/AggregateChecker.java
```diff
@@ -1,5 +1,7 @@
 package org.checkerframework.framework.source;
 
+import static javax.tools.Diagnostic.Kind.ERROR;
```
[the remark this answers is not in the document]
mernst: I'd prefer to keep this, if possible, because it makes all uses consistent throughout the codebase.

Using static imports was necessary in places because the codebase and its libraries contain multiple `Diagnostic` classes *and* multiple `Kind` classes, so only a static import would work.  So then I used it throughout.

I'll change it in this class, though.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
mernst: It's used in a `switch` statement where the import isn't needed.  Thanks for catching that.

mernst:
@wmdietl Thanks for the feedback.  Could you take another look?

framework/src/main/java/org/checkerframework/framework/source/Result.java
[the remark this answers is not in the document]
wmdietl: I found a single method `org.checkerframework.common.basetype.BaseTypeValidator#isValidType` that possibly returns a non-singleton list. The default suggestion should be to use a `DiagMessage` instead and exceptional places might use a `List<DiagMessage>`. It seems rather confusing to always suggest `List<DiagMessage>`.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeValidator.java
```diff
@@ -112,41 +118,48 @@ protected boolean shouldCheckTopLevelDeclaredType(AnnotatedTypeMirror type, Tree
...
-    protected Result isTopLevelValidType(
+    /**
+     * Checks every property listed in {@link #isValidType}, but only for the top level type. If
+     * successful, returns an empty list. If not successful, returns diagnostics.
+     *
+     * @param qualifierHierarchy the qualifier hierarchy
+     * @param type the type to be checked
+     * @return the diagnostics indicating failure, or an empty list if successful
+     */
+    protected List<DiagMessage> isTopLevelValidType(
```
wmdietl: This method either returns a singleton list or the empty list, right?
If so, I would rather model it like `FieldInvariants.isSuperInvariant` and return `DiagMessage` or null.
This makes the code and logic simpler, and the `Result` to `DiagMessage` transition more uniform.

mernst: Yes, you are right that the method returns a singleton list or the empty list.
Changing its type would make it slightly simpler, but would complicate its caller `isValidType`, which would need to immediately convert the null or DiagMessage into a list.  Take a look and let me know if you agree.

framework/src/main/java/org/checkerframework/framework/source/Result.java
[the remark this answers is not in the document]
mernst: I see your point.  I have made this change.

mernst:
@wmdietl Thanks for the reviews, which definitely improved the code.

PR #3136, opened 2020-03-07
Call typeProcessingOver even if an error occurred

mernst:
Merge after #3135 (update: #3135 has been merged), but ready for review now.


javacutil/src/main/java/org/checkerframework/javacutil/AbstractTypeProcessor.java
```diff
@@ -139,10 +139,10 @@ public void typeProcessingStart() {}
      * <p>Subclasses may override this method to do any aggregate analysis (e.g. generate report,
      * persistence) or resource deallocation.
      *
-     * <p>If an error (a Java error or a processor error) is reported, this method is not guaranteed
-     * to be invoked.
+     * @param hasError true if compilation issued an error, either from the Java compiler or from a
+     *     pluggable type-checker
      */
-    public void typeProcessingOver() {}
+    public void typeProcessingOver(boolean hasError) {}
```
wmdietl: What uses do you imagine for the parameter? Instead of nothing, this passes just the success/failure bit. Why not the number of errors or the error messages themselves?
Only two of the checkers actually do something with the boolean. As an alternative, those two checkers could create a `Log` instance and look at the number of errors themselves.
So if you don't have a larger plan for this parameter, just changing those two checkers would be a smaller change.

mernst: This is preparatory to #3137.  I separated it so that it was two smaller pull requests each with one topic.  This is what #3137 needs.  A future pull request could do more, like your suggestion.

mernst: Or, you could just review #3137 instead of this one.

wmdietl: How does #3137 need this? The implementation in SourceChecker there is:

```java
    public void typeProcessingOver(boolean hasError) {
        writeJsonDiagnostics();
        super.typeProcessingOver(hasError);
    }
```

So we could remove the boolean parameter, change the logic to always call `typeProcessingOver`, and change the two checkers that depend on success/failure to use the `Log` to do that.

mernst: #3137 depends on `typeProcessingOver` being called even if an error occurred (which is the topic of this pull request).  #3137 does not depend on the `hasError` boolean, though this pull request does depend on the `hasError` boolean.  Sorry if that was unclear.

I count three checkers that use the boolean, not two:  `AnnotationStatistics`, `CFGProcessor`, and `JavaCodeStatistics`.  
The boolean helps them preserve behavior, and is all they need.  I don't think it is simpler to make most clients create a log that they don't need, and I don't want this pull request to change their behavior.

If there is a need for the log in the future, then we should change the functionality, but it doesn't seem desirable to me right now.

wmdietl: I've made some changes in #3158. Please see whether you like them. I find it nicer without the boolean.
A few javadoc issues probably still need to be sorted out.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGProcessor.java
```diff
@@ -35,8 +37,8 @@
     /** Tree node for the specified method. */
     private @Nullable MethodTree methodTree;
 
-    /** Result of CFG process. */
-    private @Nullable CFGProcessResult result;
+    /** Result of CFG process; is set by {@link #typeProcessingOver}. */
+    private @MonotonicNonNull CFGProcessResult result = null;
```
wmdietl: This change should be in a separate PR.

PR #3137, opened 2020-03-07
Produce diagnostic message output as a JSON file

mernst:
Review after ~#3134~ and ~#3136~ have been merged.


mernst:
I agree that javac should have support for Language Server Protocol, like eclipsec does.  That would be a great improvement to javac.  It should really be built in to javac rather than provided by an external tool.

This pull request is not an attempt to create, or even to start, general LSP support for javac.  Its scope is more narrow:   it is a replacement for the `-Adetailedmsgtext` command-line option.  It happens to choose LSP JSON as its output format.

I don't have time to create general LSP support for javac.

mernst:
> A general question before I review this in detail:

This pull request will be easier to review after #3136 is merged.  Could you please approve that before reviewing this?

wmdietl:
> I agree that javac should have support for Language Server Protocol, like eclipsec does. That would be a great improvement to javac. It should really be built in to javac rather than provided by an external tool.
> 
> This pull request is not an attempt to create, or even to start, general LSP support for javac. Its scope is more narrow: it is an alternative to the `-Adetailedmsgtext` command-line option. It happens to use LSP JSON as its output format.
> 
> I don't have time to create general LSP support for javac.

I did not mean to provide general LSP support for javac.
I meant to take the logic you already wrote in this PR and instead of putting it into the CF, put it into a separate project that uses a simple wrapper that calls the compiler API to get diagnostic messages.
That would decouple the CF from this JSON formatter.

How urgent is this change for you? I can probably re-use the LSP CF server to produce the same JSON file, without needing any integration to the CF.
I'm a bit hesitant to add a JSON library as a dependency - we'll need to shadow it to avoid version mismatches and just add another dependency for a rather niche use case.

mernst:
Sorry, but I'm not sure how the simple wrapper would work.  I guess users would call it instead of javac, and it would would call javac programmatically and then query the compiler API.  I'm not sure how you would fill in the [`source`](https://github.com/typetools/checker-framework/pull/3137/files#diff-12c7683962794b9ac41a61cc01d83459R18) field, but you could leave it blank.  It sounds like you have something in mind already and could write that much faster than I could.  I can wait if you can do it within, say, a week.  Thanks.

mernst:
@wmdietl Can we merge this?  It has been open for over 3 weeks, and the release is less than 2 days away.

wmdietl:
I've put a stand-alone wrapper here: https://github.com/eisopux/javac-diagnostics-wrapper
The README contains an example for how to use it with the Checker Framework.
It should be easy to add more scripts/wrappers to make the combination easier.
Let me know if I should transfer that repo into typetools to give you direct access - or file issues, PRs, etc.

I think keeping this logic separate will make JSON and other formatting output easier to maintain outside of the Checker Framework.

mernst:
To simplify the diffs, merge pull requests #3229 #3230 #3231, then pull master into this branch.

mernst:
@wmdietl will extract parts of this into https://github.com/eisopux/javac-diagnostics-wrapper so that it produces LSP JSON.

wmdietl:
@mernst The javac-diagnostics-wrapper can now produce LSP JSON output.
I've added a reference to the external tools section https://github.com/typetools/checker-framework/pull/3250

PR #3148, opened 2020-03-09
Remove the deprecated constructors and methods as a follow-up of #2949.

xingweitian:


wmdietl:
@xingweitian Thanks for remembering to follow-up on this!

PR #3151, opened 2020-03-10
Moved purity checking to common.purity (2)

Maxi17:
This PR continues the discussion in #3092 in a new branch.

Fixed `Purity.java` by moving it to a new test directory, `purity`.

@mernst


Maxi17:
@mernst 
I think the work for the PR is done. There are still two tests (`plume_lib_jdk11` and `all_tests_jdk11`) that fail because of an `OutOfMemoryError`. I don't know why this happens.

Maxi17:
@mernst I found what the problem with `plume_lib` is. The problem is `plume-util` library. This library is type-checked using 7 checkers. The new implementation of purity checking adds the `PurityChecker` as a subchecker for each of these checkers (as it is supposed to do), therefore the purity checking is performed many times. This is the reason for the `OutOfMemoryError`. The test cases inside Checker Framework pass because they are run using only one checker at a time, so the `PurityChecker` is added only once as a subchecker.

I am sure of this because I tried to remove some of the checkers run when the gradle build starts for `plume-util`, and the build succeeds when fewer checkers are used.

I do not know how to continue fixing this.

Maxi17:
There is also a problem with `checkPurity` gradle task. This doesn't check purity of methods at all, because the `-AcheckPurityAnnotations` is not provided to it. The way purity verification works now, is that passing the `PurityChecker` as an annotation processor is not relevant unless `-AcheckPurityAnnotations` is provided, too. This is because of the check:

`        if (!checker.hasOption("checkPurityAnnotations")) {
            return;
        }`

from `BaseTypeVisitor.checkPurity(MethodTree)`. The methods doesn't do anything if the flag is not provided. As a result, there are some purity errors in CF code that are not issued by `checkPurity` task.

For example, `checker.formatter.qual.ConversionCategory.toString()` method is annotated as `@Pure`, even though it is not deterministic (the first line of this method creates a new instance of `StringBuilder`).

mernst:
With a heavy heart, I am closing this pull request.
The better abstraction in this pull request is desirable, but the performance degradation is unacceptable.
We would like to merge a pull request like this if we can find a way to do so without slowdowns.

PR #3153, opened 2020-03-11
Warn when toArray creates an array with nullable elements

mernst:


checker/src/main/java/org/checkerframework/checker/nullness/messages.properties
```diff
@@ -13,3 +13,5 @@ class.not.found.nullness.parse.error=class not found for field "%s"
 field.not.found.nullness.parse.error=Field "%s" not found
 new.array.type.invalid=annotations %s may not be applied as component type for array "%s"
 new.class.type.invalid=the annotations %s do not need be applied in object creations
+toArray.nullable.elements.not.newarray=collection of non-null elements is being converted into an array of possibly-null elements; omit the argument to toArray or make it an explicit array constructor
```
wmdietl: Could the first part of the two messages be more active? Maybe `call of toArray on collection of non-null elements results in an array possibly-null elements`?

PR #3155, opened 2020-03-11
Don't link from dataflow to framework, since framework depends on dataflow

mernst:


smillst:
I'm going to close this because this methods are removed by #3148.

PR #3158, opened 2020-03-11
Call typeProcessingOver even if an error occurred

wmdietl:
Possible changes to #3136
Still needs some documentation tweaks, but I need to run now.

### DIFF
--- a/framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
+++ b/framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
@@ -691,11 +691,11 @@ public abstract class BaseTypeChecker extends SourceChecker implements BaseTypeC
     }
 
     @Override
-    public void typeProcessingOver(boolean hasError) {
+    public void typeProcessingOver() {
         for (BaseTypeChecker checker : getSubcheckers()) {
-            checker.typeProcessingOver(hasError);
+            checker.typeProcessingOver();
         }
-        super.typeProcessingOver(hasError);
+        super.typeProcessingOver();
     }
 
     @Override
--- a/framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
+++ b/framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
@@ -695,6 +695,7 @@ public abstract class BaseTypeChecker extends SourceChecker implements BaseTypeC
         for (BaseTypeChecker checker : getSubcheckers()) {
             checker.typeProcessingOver();
         }
+
         super.typeProcessingOver();
     }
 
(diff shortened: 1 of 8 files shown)


framework/src/main/java/org/checkerframework/common/util/count/AnnotationStatistics.java
```diff
@@ -66,6 +67,12 @@
 
     @Override
     public void typeProcessingOver() {
+        Log log = getCompilerLog();
+        if (log.nerrors != 0) {
+            System.out.printf(
+                    "Compilation issued %d error(s). Report might be partial.%n", log.nerrors);
```
mernst: This is easy for a user to overlook.  I think it's better to produce no output than potentially incorrect output.  Same comment elsewhere.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
```diff
@@ -695,7 +695,6 @@ public void typeProcessingOver() {
         for (BaseTypeChecker checker : getSubcheckers()) {
             checker.typeProcessingOver();
         }
-
```
mernst: Nit: spacing change.

javacutil/src/main/java/org/checkerframework/javacutil/AbstractTypeProcessor.java
```diff
@@ -139,11 +139,20 @@ public void typeProcessingStart() {}
...
      *
-     * <p>If an error (a Java error or a processor error) is reported, this method is not guaranteed
-     * to be invoked.
+     * @param hasError true if compilation issued an error, either from the Java compiler or from a
+     *     pluggable type-checker
      */
     public void typeProcessingOver() {}
 
+    /**
+     * Return log.
```
mernst: Could you make this more descriptive?  A "log" is customarily unstructured text, but this has structure as explained at the com.sun.tools.javac.util.Log javadoc.

javacutil/src/main/java/org/checkerframework/javacutil/AbstractTypeProcessor.java
```diff
@@ -139,11 +139,20 @@ public void typeProcessingStart() {}
      * <p>Subclasses may override this method to do any aggregate analysis (e.g. generate report,
      * persistence) or resource deallocation.
      *
-     * <p>If an error (a Java error or a processor error) is reported, this method is not guaranteed
-     * to be invoked.
+     * @param hasError true if compilation issued an error, either from the Java compiler or from a
```
mernst: You don't want this change.
But do change to say that the method is guaranteed to be called.
The Javadoc should probably contain a hint about getCompilerLog.nerrors.

wmdietl:
Will be finished in #3136.

PR #3173, opened 2020-03-17
Correct method type argument inference with null; fixes #3056; fixes #3150.

### LINKED ISSUE #3056: Objects.requireNonNullElse is improperly annotated
It seems that `Objects.requireNonNullElse` has improper nullability annotations. Here are a few test cases that fail in various ways, with associated compilation log:
https://gist.github.com/octylFractal/7680eefabc242ca21f510b1fdfc42807

I think that a more correct signature would be `<T> T requireNonNullElse(@Nullable T obj, @NonNull T defaultObj)` or `<T extends @NonNull Object> T requireNonNullElse(@Nullable T obj, T defaultObj)`, but neither of these fail the third case, which seems weird to me.

### LINKED ISSUE #3149: Crashing invocation of Objects.requiresNonNullElse is permitted
The following code type-checks but should not:
```
    // This should fail typechecks, because it fails at run time.
    System.err.println((Object) requireNonNullElse(null, null));
```

A test case appears in `checker/tests/nullness/ObjectsRequireNonNullElse.java`.

### LINKED ISSUE #3150: null is permitted as a value of a type argument that extends `@NonNull Object`
Consider the following program.  I expect the Nullness Checker to issue warnings for both of the given calls.  However, the Nullness Checker issues no warnings -- it permits `null` as an argument to both methods.

```

class ObjectsRequireNonNull {
        // :: error: (argument.type.incompatible)
        requireNonNull1(null);
        // :: error: (argument.type.incompatible)
        requireNonNull2(null);
    }

    public static <T extends @NonNull Object> T requireNonNull1(T obj) {
        return obj;
    }

    public static <T> @NonNull T requireNonNull2(@NonNull T obj) {
        return obj;
    }
}
```

As another example, the following code type-checks but should not:
```
    // This should fail typechecks, because it fails at run time.
    System.err.println((Object) requireNonNullElse(null, null));
```

A test case appears in `checker/tests/nullness/ObjectsRequireNonNullElse.java`.

smillst:
Fixes #3150.   (~#3149~ is partially fixed by this pull request because the annotations in the JDK are wrong.  I'll fix them in a different pull request and enable the test.)

### DIFF
--- a/checker/tests/nullness/Issue3150.java
+++ b/checker/tests/nullness/Issue3150.java
@@ -1,5 +1,3 @@
-// @skip-test until the bug is fixed
-
 // Test case for https://tinyurl.com/cfissue/3150 .
 
 import org.checkerframework.checker.nullness.qual.NonNull;
@@ -7,9 +5,9 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 
 class Issue3150 {
     void foo(@Nullable Object nble, @NonNull Object nn) {
-        // :: error: (argument.type.incompatible)
+        // :: error: (type.argument.type.incompatible)
         requireNonNull1(null);
-        // :: error: (argument.type.incompatible)
+        // :: error: (type.argument.type.incompatible)
         requireNonNull1(nble);
         requireNonNull1("hello");
         requireNonNull1(nn);
--- a/framework/src/main/java/org/checkerframework/framework/util/typeinference/DefaultTypeArgumentInference.java
+++ b/framework/src/main/java/org/checkerframework/framework/util/typeinference/DefaultTypeArgumentInference.java
@@ -18,7 +18,9 @@ import java.util.Set;
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.type.TypeKind;
+import javax.lang.model.type.TypeMirror;
 import javax.lang.model.type.TypeVariable;
+import javax.lang.model.util.Elements;
 import javax.lang.model.util.Types;
 import javax.tools.Diagnostic.Kind;
 import org.checkerframework.framework.source.SourceChecker;
@@ -199,6 +201,11 @@ public class DefaultTypeArgumentInference implements TypeArgumentInference {
     }
 
     /**
+     * Holds an AnnotatedTypeMirror whose underlying type is {@code Object} and annotations are
+     * copied from null. It's lazily initialized in #handleNullTypeArguments.
+     */
+    private AnnotatedTypeMirror objectWithAnnosFromNull = null;
+    /**
      * If one of the inferredArgs are NullType, then re-run inference ignoring null method
      * arguments. Then lub the result of the second inference with the NullType and put the new
      * result back into inferredArgs.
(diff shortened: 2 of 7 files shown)


checker/tests/nullness/Issue3150.java
```diff
@@ -1,15 +1,13 @@
...
-
 // Test case for https://tinyurl.com/cfissue/3150 .
 
 import org.checkerframework.checker.nullness.qual.NonNull;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 class Issue3150 {
     void foo(@Nullable Object nble, @NonNull Object nn) {
-        // :: error: (argument.type.incompatible)
+        // :: error: (type.argument.type.incompatible)
```
wmdietl: I find this change in expected error message unintuitive and undesirable.
There is no type argument here, so raising an invalid type argument error is misleading.
Type argument inference should always produce type arguments that respect the bounds. Then supplied arguments can cause errors if they don't respect that bound.
Here, I would expect that type argument inference determines `@NonNull Object` as the type argument and then the argument should give an error.

For a similar reason I don't like https://github.com/typetools/jdk/pull/36 : it changes the signature just to account for bad type inference, even though that bad type inference shouldn't give different errors.

Would it be hard to do a GLB between each inferred type argument and the bounds?

framework/src/main/java/org/checkerframework/framework/util/typeinference/DefaultTypeArgumentInference.java
```diff
@@ -237,7 +244,14 @@ private void handleNullTypeArguments(
                 if (withoutNullResult == null) {
                     // withoutNullResult is null when the only constraint on a type argument is
                     // where a method argument is null.
-                    withoutNullResult = typeFactory.getUninferredWildcardType(atv);
+                    if (objectWithAnnosFromNull == null) {
+                        Elements elements = typeFactory.getProcessingEnv().getElementUtils();
+                        TypeMirror objectTM = elements.getTypeElement("java.lang.Object").asType();
+                        objectWithAnnosFromNull =
+                                AnnotatedTypeMirror.createType(objectTM, typeFactory, false);
+                        objectWithAnnosFromNull.addAnnotations(result.getAnnotations());
```
wmdietl: I'm confused why it is the right thing to use the annotations only from the first `result`. The loop goes through all type parameters, multiple of which might come from `null`, but only the first determines the value of `objectWithAnnosFromNull`.
Can you add an explanation?
Or should you ask the type factory what the type of `null` is, instead?

wmdietl: Also, should `objectWithAnnosFromNull` be set to `null` again at some point?

checker/tests/nullness/Issue3150.java
smillst: GLB the inferred type argument and its corresponding bound is tricky when the bound is another type variable. See  [checker/tests/nullness/InferNullType.java](https://github.com/typetools/checker-framework/blob/master/checker/tests/nullness/InferNullType.java).  As you can see in that example, the Checker Framework currently does infer type arguments that are outside its bounds.  I'll spend a little more time trying to change this so that it infers a type argument within the bound, but if it's not easy, then I think we should just open a separate issue.

wmdietl: Something like the following should work:

```
1. get possible incorrect assignment of type parameters to inferred type arguments
2. Iterate while there are changes:
2.1 Compute the type parameter bounds under the current assignment, to resolve type parameter uses
2.2 Go through all type parameters and GLB the inferred type argument with the current bound
```

Once #979 is fixed, this additional step shouldn't be necessary, but this might be a good time to improve the quality of type inference.

PR #3180, opened 2020-03-20
Make ValueQualifierHierarchy protected.

xingweitian:
So that new checkers can extend the value checker.


wmdietl:
@xingweitian has been working on a nicer solution. Closing this for now.

PR #3184, opened 2020-03-20
Add `@Nullable` to return types of `IdentityHashMap`.

cpovirk:
The public APIs in this class otherwise look fully annotated, with the exception of some missing `@Nullable` annotations on `Object` parameters, which are being addressed in https://github.com/typetools/checker-framework/pull/3063


mernst:
Merge together with https://github.com/typetools/jdk/pull/39.

mernst:

PR #3186, opened 2020-03-21
For IDE integration, import projects with gradle

wmdietl:


docs/developer/developer-manual.html
```diff
@@ -163,6 +163,9 @@ <h2 id="Code_style">Code style</h2>
 
 <h2 id="IDE_configuration">IDE configuration</h2>
 
+After <a href="../manual/manual.html">cloning and building all projects from their sources</a>,
+import them as gradle projects.
+
 If your IDE cannot find <code>com.sun.source.*</code> packages, try changing the project JDK to JDK 11.
 
 <h3 id="Configure_Eclipse_to_edit_the_Checker_Framework">Configure Eclipse to edit the Checker Framework</h3>
```
smillst: Are these instructions still correct or should we delete them?

wmdietl: I already removed them in #3185

PR #3187, opened 2020-03-21
Adapt to changes in IdentityHashMap annotations

mernst:


dataflow/src/main/java/org/checkerframework/dataflow/analysis/Analysis.java
```diff
@@ -809,10 +818,15 @@ public boolean isRunning() {
...
 
-    /** The transfer results for each return node in the CFG. */
-    public List<Pair<ReturnNode, TransferResult<A, S>>> getReturnStatementStores() {
+    /**
+     * The transfer results for each return node in the CFG.
+     *
+     * @return the transfer results for each return node in the CFG
+     */
+    @RequiresNonNull("cfg")
+    public List<Pair<ReturnNode, @Nullable TransferResult<A, S>>> getReturnStatementStores() {
```
wmdietl: @mernst Why is it desirable to make this `@Nullable TransferResult`?
It would have been nicer to check whether the `get` in the body returns an unexpected `null` and then abort with an error.

mernst: That would have been a code change.  I was only changing annotations.  I don't object to such a change in a different pull request.  Thanks for the suggestion.

PR #3191, opened 2020-03-24
Add nullability annotations to some JDK APIs.

cpovirk:
Merge together with https://github.com/typetools/jdk/pull/37

- `*Reference`: All constructors accept `null`, and all `get` methods can return `null`. (The `Reference` superclass itself already had an annotation on `get`, but the subclasses did not.)
- `ParameterizedType`: `getOwnerType` can return `null`.
- `URL`: The `host` parameters to the constructors can be `null`.
- `HashMap` and `Map`: `putIfAbsent` and `replace` can return `null`.
- `Objects`: `requireNonNull` ensures that its first argument is non-null, and it returns `@NonNull`, even if `<T>` might allow `null`.
- `PriorityQueue`, `SortedMap`, `SortedSet`, `TreeMap`, `TreeSet`, `ConcurrentSkipListSet`: `comparator` can returns `null`, and it is `@Pure`. Additionally, the `TreeMap` and `TreeSet` constructors accept a null `Comparator`. (`ConcurrentSkipListSet` already did.)
- `Spliterator` (and primitive specializations): `trySplit` and `getComparator` can return `null`, and `getComparator` is `@Pure`.

This PR replaces #3029.


mernst:
The associated pull request was failing, so I rebased and merged a separate pull request.
Thanks again for these annotations.

PR #3193, opened 2020-03-24
Rename AnnotatedTypeMerger to AnnotatedTypeReplacer

kelloggm:
There was already a class named `AnnotatedTypeReplacer`, so also renamed that.

Both had names that were somewhat confusing:
* `AnnotatedTypeMerger` suggests that information from both types will be merged. In fact, it replaces the contents of one annotated type with the contents of another. I therefore renamed it to `AnnotatedTypeReplacer`, especially because its documentation already used the phrase "replace" extensively.
* the old `AnnotatedTypeReplacer` copies an annotated type and makes replacements based on a separate map. I therefore renamed it to `AnnotatedTypeCopierWithReplacement`. As far as I can tell, it isn't used anywhere, so we may want to just delete it.

Rationale: I accidentally used `AnnotatedTypeMerger` when I wanted to merge two types, and got surprising results. I'll add a new class that actually merges two classes using the GLB as part of #2871, which requires that feature.


framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1965,9 +1965,9 @@ private void putNew(
                 // annotation exist in both types in the same location for the same hierarchy.
```
wmdietl: The line above this one should now refer to `AnnotatedTypeReplacer`.

framework/src/main/java/org/checkerframework/framework/type/TypeFromExpressionVisitor.java
```diff
@@ -353,10 +353,11 @@ public AnnotatedTypeMirror visitWildcard(WildcardTree node, AnnotatedTypeFactory
         // This ensures that the structure of the wildcard will match that created by
```
wmdietl: The line above talks about `merge`. It also says that it does that instead of replacing them. So it is confusing that this now uses a replacer.

checker/src/main/java/org/checkerframework/checker/nullness/KeyForPropagator.java
```diff
@@ -178,23 +178,30 @@ public void propagateNewClassTree(
      * An annotated type merger that merges @KeyFor annotations and only if the type that is
```
wmdietl: This comment still refers to `merger`.

checker/src/main/java/org/checkerframework/checker/nullness/KeyForPropagator.java
```diff
@@ -178,23 +178,30 @@ public void propagateNewClassTree(
...
+         */
+        private KeyForPropagationReplacer(ProcessingEnvironment env) {
             this.env = env;
         }
 
         @Override
-        protected void replaceAnnotations(AnnotatedTypeMirror from, AnnotatedTypeMirror to) {
-            final AnnotationMirror fromKeyFor = from.getAnnotationInHierarchy(UNKNOWN_KEYFOR);
-            final AnnotationMirror toKeyFor = to.getAnnotationInHierarchy(UNKNOWN_KEYFOR);
+        protected void replaceAnnotations(AnnotatedTypeMirror src, AnnotatedTypeMirror dst) {
```
wmdietl: The PR would be nicer if these parameter renamings happened separately, to focus on the essential changes.

framework/src/main/java/org/checkerframework/framework/type/visitor/AnnotatedTypeReplacer.java
```diff
@@ -69,14 +84,20 @@ protected Void combineRs(Void r1, Void r2) {
...
 
+    /**
+     * Replace the annotations in dst with the annotations in src
+     *
+     * @param src the source of the annotations
+     * @param dst the destination of the annotations
+     */
     protected void replaceAnnotations(
-            final AnnotatedTypeMirror one, final AnnotatedTypeMirror two) {
+            final AnnotatedTypeMirror src, final AnnotatedTypeMirror dst) {
```
wmdietl: It would have been more consistent to rename to `from` and `to`, like in the earlier methods.

kelloggm:
@wmdietl thanks for pointing out these inconsistencies. Since this has already been merged, I'll make a separate PR to address them and request your review on it.

framework/src/main/java/org/checkerframework/framework/type/TypeFromExpressionVisitor.java
kelloggm: I think this comment is just wrong - this PR didn't change the semantics of `AnnotatedTypeMerger`/`AnnotatedTypeReplacer` at all: the class always only did replacement.

PR #3196, opened 2020-03-25
Add Gradle tasks to deploy Maven artifacts (both locally and to Maven central)

smillst:
* `./gradlew deployArtifactsToLocalRepo`  installs the artifacts locally using the same pom files that are used to release the artifacts.
* `./gradlew deployArtifactsToSonatype` deploys the artifacts to Maven central.
* Moved pom files to `docs/developer/release/maven-artifacts`.  
* Renamed pom files to `artifactId-pom.xml` 

TODOs: (These don't block merging this PR)
- [ ] Need to update Maven instructions to run `./gradlew deployArtifactsToLocalRepo` instead of using the system dependency.  See #3050 )
- [ ] Delete all code related to upload artifacts from the release scripts
- [x] Add task that updates the version number in the pom files.


msridhar:
This looks great!  One thing that is still missing is the use of a `-SNAPSHOT` version number between releases.  It seems that the current version number on `master` is the number for the next release, but without `-SNAPSHOT`.  I think using the `-SNAPSHOT` suffix will make it clearer when a non-release build is being used.  It will also enable upload of snapshot builds to the Sonatype snapshot repository from CI (I can help set this up), which will let users try out the latest Checker Framework artifacts without cloning and building from source (e.g., to verify that a bug was fixed).

The `-SNAPSHOT` numbering could be adopted in a follow-up PR; just wanted to note the request.

smillst:
@msridhar 
>One thing that is still missing is the use of a -SNAPSHOT version number between releases.

I still need to add a Gradle task to update the version number in the pom files.  We currently use an ant task in the release scripts to do that.

> It will also enable upload of snapshot builds to the Sonatype snapshot repository from CI (I can help set this up), 

That would be great. Thanks!

build.gradle
```diff
@@ -54,6 +54,9 @@ ext {
     // The local git repository, typically in the .git directory, but not for worktrees.
     // This value is always overwritten, but Gradle needs the variable to be initialized.
     localRepo = ".git"
+
+    // Location of pom files used only to add meta-data to the artifacts published to Maven.
```
mernst: Is the location used only for that purpose, or are the pom files used only for that purpose?
(Using a complete sentence could have prevented my question.)

build.gradle
```diff
@@ -99,7 +102,7 @@ allprojects {
     //   * any new checkers have been added,
     //   * the patch level is 9 (keep the patch level as a single digit), or
     //   * backward-incompatible changes have been made to APIs or elsewhere.
-    version '3.2.0'
+    version '3.2.1'
```
mernst: Why is the release number already incremented?  Maybe it's a temporary artifact of the lack of SNAPSHOT version numbers.

build.gradle
```diff
@@ -969,3 +972,84 @@ task releaseAndTest {
 
 // Don't create an empty checker-framework-VERSION.jar
 jar.onlyIf {false}
+
+/**
+ * Deploys the given jar file to the local Maven repository.
+ * @param jar fully qualified file name of a jar file
+ * @param pom pom file describing the artifact
```
mernst: Is pom a filename, as opposed to the file contents?
(Reading on, I see that it is.)
Same comment a few other places.

build.gradle
```diff
@@ -969,3 +972,84 @@ task releaseAndTest {
...
+ */
+void mvnDeployToLocalRepo(String jar, String pom) {
+    exec {
+        executable 'mvn'
+        args += [
+                'install:install-file',
+                "-Dfile=${jar}",
+                "-DpomFile=${pom}",
+                // TODO: This changes the version number of the artifact, but not the version number in the pom file.
```
mernst: This isn't in the TODO list in the pull request.

build.gradle
```diff
@@ -969,3 +972,84 @@ task releaseAndTest {
...
+    mvnSignAndDeployToSonatype(sourceJar, pom, "sources")
+    mvnSignAndDeployToSonatype(javadocJar, pom,"javadoc")
+}
+
+/**
+ * Signs and deploys the classes jar, sources jar, and javadoc jar files for a single artifact.
+ * @param project with a jar, sourcesJar, and javadocJar tasks defined
+ * @param pom pom file describing the artifact
+ */
+void mvnSignAndDeployMultipleToSonatype(Project project, String pom) {
```
mernst: This could call the method that is defined on line 1040.  That wouldn't make the code much shorter, but it would clarify the relationship between the two methods, which would helped me to understand them.  The documentation could also be made more similar -- the difference in the Javadoc first sentence also made me think the methods were more different than they are.

mernst:
@smillst One more thing: There are `<dependencies>` in `dataflow-pom.xml` but not in `dataflow-shaded.xml`.

docs/developer/release/maven-artifacts/dataflow-pom.xml
```diff
@@ -37,12 +37,6 @@
             <url>http://www.gnu.org/software/classpath/license.html</url>
             <distribution>repo</distribution>
         </license>
-
-        <license>
```
wmdietl: The qualifiers in this package are under MIT license, aren't they?

docs/developer/release/maven-artifacts/dataflow-shaded-pom.xml
```diff
@@ -23,12 +23,6 @@
             <url>http://www.gnu.org/software/classpath/license.html</url>
             <distribution>repo</distribution>
         </license>
-
-        <license>
```
wmdietl: Same comment about qualifiers having MIT license.

docs/developer/release/maven-artifacts/dataflow-pom.xml
mernst: Yes, but the given license is the most restrictive one, and therefore the most useful one to state.
When a POM file contains two licenses, that generally means that the user is permitted to use either of the given licenses, at the user's option.  That is not the case for this project, and therefore giving two licenses was misleading.

docs/developer/release/maven-artifacts/dataflow-shaded-pom.xml
mernst: Same response. :-)

docs/developer/release/maven-artifacts/dataflow-pom.xml
wmdietl: Ah, yes, that makes sense. Thanks for clarifying. Maybe a comment here would prevent confusion.

mdernsta:
I'm OK with merging this pull request to avoid delay and because it is an improvement over the current state.

However, the right way to publish artifacts is to use the "maven-publish" Gradle plugin.  At one time I avoided it because it lacked support for signing artifacts, which Maven Central requires.  However, it seems to have that capability now.  I suggest opening an issue to further refactor Maven Central publishing to use the "maven-publish" Gradle plugin.

msridhar:
FWIW, WALA now uses "maven-publish" for creating snapshot and release artifacts and pushing them to Maven Central.  Here is the relevant Gradle code:

https://github.com/wala/WALA/blob/master/gradle-mvn-push.gradle

PR #3208, opened 2020-03-30
Warning suppression keys must be a logical subpart of the suppressed message

mernst:


changelog.txt
```diff
@@ -15,6 +15,11 @@ contains, indexOf, and remove, the annotated JDK now forbids null as an
 argument.  To make the Nullness Checker permit null, pass
 `-Astubs=checker.jar/collection-object-parameters-may-be-null.astub`.
 
+The argument to @SuppressWarnings can be a substring of a message key that
+is delimited by periods and/or either end of the key.  (Previously, any
```
wmdietl: I like the "substring extends at each end to a period or an end of the key" used below. Maybe similar text can be used here?
I couldn't come up with something succinct that's better. (The "substring extends at each end to a period (excluding the period) or an end of the key" makes it more explicit that the dots are not part of the substring, but that's nit-picking...)

PR #3210, opened 2020-03-30
Annotate bit patterns as @Unsigned

mernst:
Merge together with https://github.com/typetools/jdk/pull/43.

Maybe this isn't the right solution.
Maybe there should be a `@BitPattern` datatype that is incomparable with all the other types in the hierarchy (except top and bottom), and on which all (and only) bitwise operations are permitted.


mernst:
This is the wrong approach.  See https://github.com/typetools/checker-framework/issues/3211.

Issue #3201, opened 2020-03-29
IndexChecker InvocationTargetException on IntelliJ


smellslikekeenspirit:
## Steps to reproduce this on IntelliJ (using Checker 3.2.0)
 
-> Clone [this](https://github.com/smellslikekeenspirit/jalgorithms/tree/Annotations) repository on IntelliJ 
 
-> Set project SDK to 1.8
 
-> Change the language level to 8
 
-> Add the annotated JDK library to the bootclasspath. Under `File | Settings |  Build, Execution and Deployment | Compiler | Java Compiler` go to the  `Additional command line parameters` section, and add:

> -Xbootclasspath/p:".../checker-framework/checker/dist/jdk8.jar"
  
-> Configure an annotation profile under `File | Settings | Build, Execution and Deployment | Annotation Processors`
 
   1. Create an annotation profile.

   2. Associate the `jalgorithms` module with that profile.

   3. Add .../checker-framework/checker/dist/checker.jar to the processor path.

   4. Add "org.checkerframework.checker.index.IndexChecker" in the “Processor FQ Name” section.

   5. Add .../checker-framework/checker/dist/checker-qual.jar, as a dependency to the `jalgorithms` module)

## The bug you will likely see
```
Error:java: InvocationTargetException when invoking constructor for class org.checkerframework.checker.index.substringindex.SubstringIndexAnnotatedTypeFactory on args [org.checkerframework.checker.index.substringindex.SubstringIndexChecker@7e6a9f9c]; Underlying cause: java.lang.NoSuchFieldError: RELEASE
  ; The Checker Framework crashed.  Please report the crash.
  Exception: java.lang.reflect.InvocationTargetException; Stack trace: 
  sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
  sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
  sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
  java.lang.reflect.Constructor.newInstance(Constructor.java:423)
  org.checkerframework.common.basetype.BaseTypeChecker.invokeConstructorFor(BaseTypeChecker.java:294)
  org.checkerframework.common.basetype.BaseTypeVisitor.createTypeFactory(BaseTypeVisitor.java:254)
[...]

PR #3214, opened 2020-03-31
IntelliJ requires use of JDK 11

smellslikekeenspirit:
This addresses https://github.com/typetools/checker-framework/issues/3201, wherein following the current instructions for setting up Checker on IntelliJ was not sufficient for successful compilation.

### DIFF
--- a/docs/manual/external-tools.tex
+++ b/docs/manual/external-tools.tex
@@ -728,7 +728,7 @@ to run a checker within IntelliJ IDEA on every compile:
 \item Change the project SDK to 11, as explained at
   \url{https://www.jetbrains.com/help/idea/sdk.html#change-project-sdk}.
 
-\item Make sure the language level for your project is 11 (after the last step, it should show up as your SDK default), as explained at
+\item Make sure the language level for your project is 8 or higher, as explained at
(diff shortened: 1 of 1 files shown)


docs/manual/external-tools.tex
```diff
@@ -725,10 +725,10 @@
 
 \begin{enumerate}
 
-\item Change the project SDK to 1.8, as explained at
+\item Change the project SDK to 11, as explained at
   \url{https://www.jetbrains.com/help/idea/sdk.html#change-project-sdk}.
 
-\item Change the language level for your project to 8, as explained at
+\item Make sure the language level for your project is 11 (after the last step, it should show up as your SDK default), as explained at
```
smillst: Can you test if this works with the project SDK set to 11, but the language level set to 8?

smellslikekeenspirit: It does, but wouldn't it make sense to use the SDK default?

smillst: Not for all projects.  Some projects use JDK 11 to build, but set the release level to a lower version, such as 8.  This allows the project to be run on Java 8+.  So the text ` Make sure the language level for your project is 11`  should be changed to ` Make sure the language level for your project is 8 or higher`.

PR #3215, opened 2020-03-31
Update Maven example to use local repo. (Fixes #3050)

### LINKED ISSUE #3050: Maven instructions: suggest installing to local repo instead of using system dependence?

smillst:
Fixes #3050.

### DIFF
--- a/docs/manual/external-tools.tex
+++ b/docs/manual/external-tools.tex
@@ -928,7 +928,6 @@ section, or add it if it is not present.
   <plugin>
     <!-- This plugin will set properties values using dependency information -->
     <groupId>org.apache.maven.plugins</groupId>
-    <groupId>org.apache.maven.plugins</groupId>
     <artifactId>maven-dependency-plugin</artifactId>
     <executions>
       <execution>
--- a/docs/manual/external-tools.tex
+++ b/docs/manual/external-tools.tex
@@ -1011,18 +1011,22 @@ Warning:  adding
 \noindent
 may suppress warnings from the stub parser.
 
+\end{enumerate}
 
-\item To use a locally-modified version of the Checker Framework, first run:
+
+\subsectionAndLabel{Maven, with a locally-modified version of the Checker Framework}{maven-local-modifications}
+
+To use a locally-modified version of the Checker Framework, first run:
 
 \begin{alltt}
 ./gradlew deployArtifactsToLocalRepo
 \end{alltt}
 
 \noindent
-Then, in addition to the instructions above, use the release version output by running \<./gradlew version>
-as the version number for the Checker Framework artifacts.
+Then use the Maven instructions, but modify the version number for the
+Checker Framework artifacts.  Instead of \<\ReleaseVersion{}>, use the
+version number that is output when you run \<./gradlew version>.
 
-\end{enumerate}
 
 \sectionAndLabel{NetBeans}{netbeans}
 
--- a/docs/manual/external-tools.tex
+++ b/docs/manual/external-tools.tex
@@ -908,15 +908,14 @@ latest bug fixes and new features:
 \item
   To use a local version of the Checker Framework that you have built from
   source, install the relevant Checker Framework jars to your local Maven
-  repository under a SNAPSHOT version number.  For example, you could run:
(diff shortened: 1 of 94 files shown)


docs/manual/external-tools.tex
```diff
@@ -906,44 +906,32 @@
...
-      <artifactId>javac</artifactId>
-      <version>9+181-r4173-1</version>
-    </dependency>
-  </dependencies>
+\noindent
+Then, in the project you want to type-check, use
+\<\ReleaseVersion{}-SNAPSHOT> as the version number:
+
+\begin{alltt}
+   <artifactId>maven-compiler-plugin</artifactId>
```
mernst: This seems like the wrong place for these instructions.  This text appears in point 1 in a numbered list, but point 3 in the list is the first one that mentions the `checker` artifact.  I think it should appear in point 3, or more likely in a new subsection that appears at the end of the Maven section.

docs/manual/external-tools.tex
```diff
@@ -906,44 +906,32 @@
...
+        ...
+        <annotationProcessorPaths>
+          <path>
+            <groupId>org.checkerframework</groupId>
+            <artifactId>checker</artifactId>
+            <version>\ReleaseVersion{}-SNAPSHOT</version>
+          </path>
 \end{alltt}
 
+You can make similar changes if you have changed other artifacts besides
```
mernst: It may be least error-prone just to say "use \ReleaseVersion{}-SNAPSHOT everywhere that the above instructions say \ReleaseVersion".  That is less ambiguous, less error-prone, and requires less thought from the reader.

smillst: `\ReleaseVersion{}-SNAPSHOT` is the snapshot before the release, so it's not the correct version to use.  `./gradlew version` outputs the version of the artifacts that `./gradlew deployArtifactsToLocalRepo` uses.

mernst: Good point about the version number.  Can the text restructuring be done, with a correct version number?

mernst: (That is, is the text restructuring a good idea?)

docs/manual/external-tools.tex
```diff
@@ -968,6 +928,7 @@
   <plugin>
     <!-- This plugin will set properties values using dependency information -->
     <groupId>org.apache.maven.plugins</groupId>
+    <groupId>org.apache.maven.plugins</groupId>
```
mernst, line 931: Why does this line appear twice?

PR #3219, opened 2020-04-01
`@IntRange` annotation for `charAt` method return type. (Fixes #3195)

### LINKED ISSUE #3195: charAt LowerBound Unknown
`charAt` should return a non-negative char. Consider the following test case:

```java
class CharAtTest {
    @NonNegative char test1(@MinLen(1) String s) {
        return s.charAt(0);
    }
}
```

Output:
```
CharAtTest.java:6: error: [return.type.incompatible] incompatible types in return.
        return s.charAt(0);
                       ^
  type of expression: @LowerBoundUnknown char
  method return type: @NonNegative char
1 error
```
Expected: No errors
I'll be working on fixing this issue.

MrigankArora:
Fixes #3195.


mernst:
Issue 3195 is a real issue.  Thanks for working on a fix for it.  The test case is also a good one.

I feel that annotating one method in the JDK isn't the right way to fix this problem.  Rather, I think the Constant Value Checker (which checks `@IntRange` qualifiers) should treat all chars consistently as having type `@IntRange(from = 0, to = 65535)`.

kelloggm:
It is possible that the Value Checker is defaulting `char`s correctly, but that this annotation not being present in the stub file causes this *particular* `char` not to be.

As evidence, consider that this program compiles with no errors using the Value Checker:

```

class ValueChar {
    @IntRange(from = 0, to = 65535) int x(char c) {
	return c;
    }
}
```

mernst:
That's an interesting point.  Thanks for that information.

It would be unfortunate and error-prone to have to change every `char` in a stub file to `@IntRange(from=0, to=65535) char`.  That is perhaps necessary with the current state of the code, but can we think of a way to avoid needing that?

kelloggm:
> That is perhaps necessary with the current state of the code, but can we think of a way to avoid needing that?

This is a general property of stubs that I've noticed before, and believe is intentional in the design. From the manual: 

> A stub file takes precedence over bytecode, for elements in the stub file. If a stub file does not mention a method or field, its annotations are taken from bytecode. An un-annotated type variable in a stub file is used instead of annotations on a type variable in bytecode. This feature allows stub files to change the effective annotations in all possible situations.

The manual specifically calls out type variables, but in this case at least it seems that the unannotated type in the stub file is taking precedence over the defaulting rules, even though it is not a type variable. This might be a bug in how stub types are applied, but I can also imagine it being a feature: if the user provides a stub file, why should the type system assume that it is incomplete?

wmdietl:
Normal defaults are applied in stub files.
It should not be necessary to annotate `char` as `@IntRange(from=0, to=65535) char`.
So instead of this change, it would be good to figure out whether the value checker or stub parser is making a mistake.

PR #3221, opened 2020-04-01
Ignore getting classpath from SystemClassLoader on Java 9+

zcai1:
`ClassLoader.getSystemClassLoader()` no longer returns `URLClassLoader` starting from Java 9+, so this piece of code could potentially throw `ClassCastException`.

Unfortunately, I couldn't find a nice alternative approach, so it is left as a TODO.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
+++ b/framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
@@ -1939,7 +1939,10 @@ public abstract class SourceChecker extends AbstractTypeProcessor
                 }
                 userKey = userKey.substring(colonPos + 1);
             }
-            if (errKey.contains(userKey)) {
+            if (errKey.equals(userKey)
+                    || errKey.startsWith(userKey + ".")
+                    || errKey.endsWith("." + userKey)
+                    || errKey.contains("." + userKey + ".")) {
                 return true;
             }
         }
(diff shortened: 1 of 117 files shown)


framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -2367,9 +2367,15 @@ private void logBugInCF(BugInCF ce) {
             if (printClasspath) {
                 msg.add("Classpath:");
```
wmdietl, line 2368: As this code is building a diagnostic message, you should try to make it fail more gracefully.
How about moving this into the `then` part and adding an `else` part that mentions that Java 9+ is used?
The comment explaining the TODO should then move into the else part.

zcai1: feedback addressed

mernst:
There is something strange about this branch.
I cannot push commits to the branch.  (@zcai1 Is the "allow commits from maintainers" checkbox checked?)
I have made a pull request against the branch.  That pull request will be easier to understand if you pull master into the branch first.
@zcai1 Please try to avoid force-pushing (though I don't have evidence that that is the problem).

zcai1:
> There is something strange about this branch.
> I cannot push commits to the branch. (@zcai1 Is the "allow commits from maintainers" checkbox checked?)
> I have made a pull request against the branch. That pull request will be easier to understand if you pull master into the branch first.
> @zcai1 Please try to avoid force-pushing (though I don't have evidence that that is the problem).

Thank you for your feedback! The PR has been merged.

The "allow commits from maintainers" is always checked, so I'm not sure why it didn't work.

mernst:
> The "allow commits from maintainers" is always checked, so I'm not sure why it didn't work.

That's what I figured.  Maybe the `+` in the branch name was a problem?  Anyway, the pull request looks good now.  Thanks again.

PR #3226, opened 2020-04-04
Make all the obj parameters in Field @NonNull.

xingweitian:
Related PRs:

- https://github.com/plume-lib/options/pull/54
- https://github.com/codespecs/daikon/pull/232
- https://github.com/typetools/jdk/pull/45
- https://github.com/plume-lib/reflection-util/pull/51

### DIFF
--- a/checker/jdk/nullness/src/java/lang/reflect/Field.java
+++ b/checker/jdk/nullness/src/java/lang/reflect/Field.java
@@ -21,9 +21,9 @@ public final class Field extends AccessibleObject implements Member {
   @Pure public int hashCode() { throw new RuntimeException("skeleton method"); }
   @SideEffectFree public String toString() { throw new RuntimeException("skeleton method"); }
   @SideEffectFree public String toGenericString() { throw new RuntimeException("skeleton method"); }
-  // These Object formal parameters can be null if the field is static, or can be non-null if the
-  // field is field instance. We don't know which. To prevent a possible NullPointerException,
-  // the Nullness Checker should conservatively issue a warning whenever null is passed.
+  // The `obj` formal parameters can be null if the field is static, or must be non-null if the
+  // field is an instance field. We don't know which. To prevent a possible NullPointerException,
+  // the Nullness Checker conservatively issues a warning whenever null is passed.
   @SideEffectFree public @Nullable Object get(Object obj) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
   @Pure public boolean getBoolean(Object obj) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
   @Pure public byte getByte(Object obj) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
@@ -33,7 +33,7 @@ public final class Field extends AccessibleObject implements Member {
   @Pure public long getLong(Object obj) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
   @Pure public float getFloat(Object obj) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
   @Pure public double getDouble(Object obj) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
-  // It is OK to set value to a not fully initialized obj parameter
+  // It is OK to set a value in a possibly not fully-initialized `obj` parameter
   public void set(@UnknownInitialization Object obj, Object value) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
   public void setBoolean(@UnknownInitialization Object obj, boolean value) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
   public void setByte(@UnknownInitialization Object obj, byte value) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
--- a/checker/jdk/nullness/src/java/lang/reflect/Field.java
+++ b/checker/jdk/nullness/src/java/lang/reflect/Field.java
@@ -7,6 +7,14 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 import org.checkerframework.dataflow.qual.Pure;
 import org.checkerframework.dataflow.qual.SideEffectFree;
 
+// In general, the field value `get` methods should take a top-qualified `obj` parameter
+// and have a top-qualified return type; the field value `set` methods should take a
+// top-qualified `obj` parameter and a bottom-qualified `value` parameter.
+//
+// nullness: the `obj` parameter in `get`/`set` methods is @NonNull, because instance fields
+// require a receiver. Static field accesses need to suppress the errors.
+//
+// initialization: using fully-initialized types should make the typical use case easier.
 public final class Field extends AccessibleObject implements Member {
   protected Field() {}
   @SideEffectFree public Class<?> getDeclaringClass() { throw new RuntimeException("skeleton method"); }
--- a/checker/jdk/nullness/src/java/lang/reflect/Field.java
+++ b/checker/jdk/nullness/src/java/lang/reflect/Field.java
@@ -21,25 +21,28 @@ public final class Field extends AccessibleObject implements Member {
   @Pure public int hashCode() { throw new RuntimeException("skeleton method"); }
   @SideEffectFree public String toString() { throw new RuntimeException("skeleton method"); }
   @SideEffectFree public String toGenericString() { throw new RuntimeException("skeleton method"); }
-  // These Object formal parameters can be null if the field is static.
-  @SideEffectFree public @Nullable Object get(@UnknownInitialization @Nullable Object obj) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
-  @Pure public boolean getBoolean(@UnknownInitialization @Nullable Object obj) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
-  @Pure public byte getByte(@UnknownInitialization @Nullable Object obj) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
-  @Pure public char getChar(@UnknownInitialization @Nullable Object obj) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
-  @Pure public short getShort(@UnknownInitialization @Nullable Object obj) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
-  @Pure public int getInt(@UnknownInitialization @Nullable Object obj) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
-  @Pure public long getLong(@UnknownInitialization @Nullable Object obj) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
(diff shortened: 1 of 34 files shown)


checker/jdk/nullness/src/java/lang/reflect/Field.java
```diff
@@ -21,25 +21,27 @@ protected Field() {}
...
-  public void setLong(@UnknownInitialization @Nullable Object obj, long value) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
-  public void setFloat(@UnknownInitialization @Nullable Object obj, float value) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
-  public void setDouble(@UnknownInitialization @Nullable Object obj, double value) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
+  // These Object formal parameters can be null if the field is static, or can be non-null if the
+  // field is field instance. We don't know which. To prevent a possible NullPointerException,
+  // the Nullness Checker should conservatively issue a warning whenever null is passed.
+  @SideEffectFree public @Nullable Object get(@UnknownInitialization Object obj) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
```
wmdietl, line 27: Please also remove the `@UnknownInitialization`. We also don't want the top initialization qualifier, for the same reasons as why we don't want `@Nullable`.

mernst: Could you explain this comment?  We don't want `@Nullable` because it might lead to `NullPointerException`.  How does an `@UnknownInitialization` argument lead to an exception?

mernst:
@xingweitian  You need to make branches of daikon and options, in your own forks of those repositories.
However, I have made the changes for you.  Just create branches named `jdk-update` and pull in my changes from the branches I created:

 * Branch `field-set-nullness` at https://github.com/mernst/daikon/branches
 * Branch `field-set-nullness` at https://github.com/plume-lib/options/branches

Then please reference those pull requests from your commant at the top of this pull request.

mernst:
@xingweitian Thanks for doing that.  It seems that the only remaining error is in the `jdk_jar_jdk8` branch:

```
Note: NullnessChecker is type-checking src/java/util/TimerTask.java
src/java/util/TreeMap.java:2332: error: incompatible types: inference variable V#1 has incompatible equality constraints K#2,K#3,V#2,V#3
                    setColor(parentOf(parentOf(x)), RED);
                            ^
  where V#1,K#1,K#2,V#2,K#3,V#3 are type-variables:
    V#1 extends Object declared in method <K#1,V#1>parentOf(Entry<K#1,V#1>)
    K#1 extends Object declared in method <K#1,V#1>parentOf(Entry<K#1,V#1>)
    K#2 extends Object declared in method <K#2,V#2>setColor(Entry<K#2,V#2>,boolean)
    V#2 extends Object declared in method <K#2,V#2>setColor(Entry<K#2,V#2>,boolean)
    K#3 extends Object declared in class TreeMap
    V#3 extends Object declared in class TreeMap
```

xingweitian:
> @xingweitian Thanks for doing that. It seems that the only remaining error is in the `jdk_jar_jdk8` branch:
> 
> ```
> Note: NullnessChecker is type-checking src/java/util/TimerTask.java
> src/java/util/TreeMap.java:2332: error: incompatible types: inference variable V#1 has incompatible equality constraints K#2,K#3,V#2,V#3
>                     setColor(parentOf(parentOf(x)), RED);
>                             ^
>   where V#1,K#1,K#2,V#2,K#3,V#3 are type-variables:
>     V#1 extends Object declared in method <K#1,V#1>parentOf(Entry<K#1,V#1>)
>     K#1 extends Object declared in method <K#1,V#1>parentOf(Entry<K#1,V#1>)
>     K#2 extends Object declared in method <K#2,V#2>setColor(Entry<K#2,V#2>,boolean)
>     V#2 extends Object declared in method <K#2,V#2>setColor(Entry<K#2,V#2>,boolean)
>     K#3 extends Object declared in class TreeMap
>     V#3 extends Object declared in class TreeMap
> ```

Thanks for pointing it out! I tried to reproduce it on my computer and I failed... It is weird, this PR did not touch this part...

checker/jdk/nullness/src/java/lang/reflect/Field.java
wmdietl: It won't lead to an exception, that is true. But it would allow reading information from an object that is still being initialized.
So then the return type should also be `@UnknownInitialization`, shouldn't it?
It seems easier to normally only allow fully initialized objects and require a suppression if the receiver is not.

checker/jdk/nullness/src/java/lang/reflect/Field.java
```diff
@@ -21,25 +21,27 @@ protected Field() {}
...
+  @Pure public short getShort(@UnknownInitialization Object obj) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
+  @Pure public int getInt(@UnknownInitialization Object obj) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
+  @Pure public long getLong(@UnknownInitialization Object obj) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
+  @Pure public float getFloat(@UnknownInitialization Object obj) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
+  @Pure public double getDouble(@UnknownInitialization Object obj) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
+  public void set(@UnknownInitialization Object obj, @Nullable Object value) throws IllegalArgumentException, IllegalAccessException { throw new RuntimeException("skeleton method"); }
```
wmdietl, line 36: Related to the previous comment: allowing the `value` to be `@Nullable`, but not `@UnknownInitialization`, seems inconsistent.
I would change this to fully-initialized and non-null values only.
Otherwise, reflection can be used to set a field that is declared `@NonNull` to `null`, without a single warning that needs to be suppressed.

checker/jdk/nullness/src/java/lang/reflect/Field.java
mernst: A `@PolyInitialized` annotation seems appropriate in that case.

mernst: ... but until we do so, we could remove `@UnknownInitialization`.

xingweitian:
`jdk_jar_jdk8` check is successful now. The last error is: `nullness:unneeded.suppression` in `options`.

mernst:
> `jdk_jar_jdk8` check is successful now. The last error is: `nullness:unneeded.suppression` in `options`.

Only the JDK 11 job is failing, not the JDK 8 job.  A difference between JDK 8 and JDK 11 indicates a bug, typically a discrepancy in the two versions of the annotated JDK.  Can you ensure that you have  made corresponding edits to both?

xingweitian:
> > `jdk_jar_jdk8` check is successful now. The last error is: `nullness:unneeded.suppression` in `options`.
> 
> Only the JDK 11 job is failing, not the JDK 8 job. A difference between JDK 8 and JDK 11 indicates a bug, typically a discrepancy in the two versions of the annotated JDK. Can you ensure that you have made corresponding edits to both?

Thanks a lot! Now all the tests passed.

checker/jdk/nullness/src/java/lang/reflect/Field.java
wmdietl: Even if we had a `@PolyInitialized`, I think it would be unsound here.
Any field read from a fully-initialized object needs to be fully-initialized, but a field read from an under-initialization object may not be under initialization. So using `@PolyInitialized` would introduce an unsoundness.

checker/jdk/nullness/src/java/lang/reflect/Field.java
wmdietl: @mernst Do you agree with removing the `@Nullable` here?

mernst: Yes, that seems reasonable.  Thanks for the comment.

xingweitian:
OK, I think I have solved all the comments!

wmdietl:
@xingweitian I made a few tweaks to the `set` methods in https://github.com/typetools/jdk/pull/45/commits/2d0050670933a19cd052bec75c21e04f78d84f7e
Could you make sure they are reflected here/in the `lock` version of this file?

mernst:
@xingweitian Then ping me if any of the downstream jobs (plume-lib, daikon, etc.) fail.

xingweitian:
@mernst options failed. Azure pipelines log: https://dev.azure.com/checkerframework/checkerframework/_build/results?buildId=542&view=logs&j=6d7f356d-b53b-5741-11e4-46348384314f&t=91bd2f1c-c8dc-5ccf-3911-504b875c6660

mernst:
It seems to be failing for the same reason that all_tests_jdk11 is failing.  Could you fix that first, then let me know if the plume_lib_jdk11 job fails when the all_tests_jdk11 job passes.
More generally, it is suspicious that the plume_lib_jdk8 job passes but the plume_lib_jdk11 job fails.  Any difference between them is a bug (usually a mismatch between the annotated JDKs).

xingweitian:
Hi @mernst, now both `plume_lib_jdk8` and 11 failed while `all_tests_jdk8` and 11 passed. And also daikon jdk8 test failed. (We did not enable `daikon_jdk11` yet.)

mernst:
@xingweitian Can you make a branch `jdk-update` in your fork of plume-lib/reflection-util, and pull its `field-set-nullness` branch into yoru new `jdk-update` branch, and make a pull request for it and reference it from here?  Then re-run the failing tests.

xingweitian:
> @xingweitian Can you make a branch `jdk-update` in your fork of plume-lib/reflection-util, and pull its `field-set-nullness` branch into yoru new `jdk-update` branch, and make a pull request for it and reference it from here? Then re-run the failing tests.

I have made the same change in plume-util as in reflection-util: https://github.com/plume-lib/plume-util/pull/53/files
Otherwise, `plume-util`'s ReflectionPlume.java cannot pass the building.

wmdietl:
@xingweitian Can you have a look at the plume-lib and daikon failures? It complains about redundant suppressions.

xingweitian:
> @xingweitian Can you have a look at the plume-lib and daikon failures? It complains about redundant suppressions.

I think the warnings are expected. The downstream projects are using the master branch of checker-framework, so they actually don't have the updated code (after we merge this PR, they should not complain about the redundant suppressions anymore).

mernst:
> > @xingweitian Can you have a look at the plume-lib and daikon failures? It complains about redundant suppressions.
> 
> I think the warnings are expected. The downstream projects are using the master branch of checker-framework and the checker-framework gradle plugin, so they actually don't have the updated code (after we merge this PR, they should not complain about the redundant suppressions anymore).

Weitian's comment is correct.

wmdietl:
Ah, thanks, that makes sense! Please feel free to merge then. @mernst, I guess it's best for you to do this in the downstream projects.

wmdietl:
@mernst Pull requests after this one were merged are failing because of a plume-lib error:
https://dev.azure.com/checkerframework/checkerframework/_build/results?buildId=577&view=logs&j=6fd4313e-f60c-5399-cd7e-43e28962c2b7&t=3cc660ad-f376-52c4-61b1-c7e23795c8b8&l=794

What is the right way to resolve this?

xingweitian:
> @mernst Pull requests after this one were merged are failing because of a plume-lib error:
> https://dev.azure.com/checkerframework/checkerframework/_build/results?buildId=577&view=logs&j=6fd4313e-f60c-5399-cd7e-43e28962c2b7&t=3cc660ad-f376-52c4-61b1-c7e23795c8b8&l=794
> 
> What is the right way to resolve this?

This one https://github.com/plume-lib/plume-util/pull/53 should also be merged. I forgot to add it to the description of this PR...

Issue #3223, opened 2020-04-02
Daikon cannot pass the Nullness checking in JDK11.


xingweitian:
```
Note: NullnessChecker is type-checking daikon/Daikon.java
daikon/Daikon.java:1145: error: [nullness:argument.type.incompatible] incompatible types in argument.
              classesAsObject = f.get(Thread.currentThread().getContextClassLoader());
                                                                                  ^
  found   : @Initialized @Nullable ClassLoader
  required: @UnknownInitialization @NonNull Object
```

```
Note: NullnessChecker is type-checking daikon/chicory/DaikonVariableInfo.java
daikon/chicory/DaikonVariableInfo.java:790: error: [nullness:argument.type.incompatible] incompatible types in argument.
                value = Integer.toString(System.identityHashCode(field.get(null)));
                                                                           ^
  found   : null
  required: @UnknownInitialization @NonNull Object
```

```
Note: NullnessChecker is type-checking daikon/config/ParameterDoclet.java
daikon/config/ParameterDoclet.java:229: error: [nullness:argument.type.incompatible] incompatible types in argument.
      Object value = f.get(null);
                           ^
  found   : null
  required: @UnknownInitialization @NonNull Object
```

They are all about `Field.get(Field this, Object obj)`. I found that `@Nullable` was removed in this [commit](https://github.com/typetools/jdk/commit/ba84219e5d4bfa9e02ee73ffb9a8528bd1d5f703#diff-bd7c0c0650c4f25725d5b92eed2587b4L433).

The current master branch can pass the pipelines is because that `daikon_jdk11` is commented in [azure-pipelines.yml](https://github.com/typetools/checker-framework/blob/master/azure-pipelines.yml#L76), while it is not commented in travis.yml.

mernst:
`daikon_jdk11` is commented out in `azure-pipelines.yml` because Daikon currently [requires Java 8](http://plse.cs.washington.edu/daikon/download/doc/daikon.html#Requirements).  Since apparently Daikon compiles under Java 11, we could reinstate the `daikon_jdk11` job.

These are false positive warnings.  `Field.get`'s argument is `@NonNull` because `Field.get` might throw a NullPointerException.  The only way to guarantee no NullPointerException is to require the argument to be non-null.

The difference between the behavior of Java 8 and Java 11 is troubling, however.  That looks like a bug.  Why does Daikon type-check under Java 8 but not under Java 11?  I would expect the type-checker to issue the same warnings in each case.

[...]

PR #3227, opened 2020-04-04
Enable daikon_jdk11 on Azure Pipelines

xingweitian:
As #3223 discussed:

Since Daikon can compile under Java 11 now, to be consistent with travis (`daikon_jdk11` did not comment out in .travis.yml), we should enable it again.


wmdietl:
This is blocked on #3226 and on Daikon adapting to those changes.

PR #3229, opened 2020-04-06
Note about jtreg tests

mernst:


checker/jtreg/README
```diff
@@ -8,3 +8,7 @@ There are a few cases when jtreg is appropriate:
    in the test file itself.
  * The jtreg approach makes it possible to check that an error message
    was output two times.
+
+Sometimes you should put tests in a directory such as
```
wmdietl, line 12: The `nullness-extra` tests are not jtreg tests, so the reference is confusing.
Is this meant as an example for jtreg tests (they're not) or as an alternative to them?
A jtreg can also contain multiple files and even very flexibly indicate in which order they need to be compiled.
Can you clarify this paragraph?

mernst: This is intended to be an alternative.  I updated the text; thanks for pointing out that it was unclear.

jtreg can handle some but not all cases with multiple files.  I tried to clarify that as well.

checker/jtreg/README
```diff
@@ -8,3 +8,8 @@ There are a few cases when jtreg is appropriate:
    in the test file itself.
  * The jtreg approach makes it possible to check that an error message
    was output two times.
+
+Sometimes jtreg is not capable of executing a test, such as in some but not
+all cases when a test uses multiple Java files.  In such a case, you will
```
wmdietl, line 13: It would help to make this a bit more concrete.
There are many jtreg tests that use multiple files and it is quite flexible to configure the order in which these are compiled, what flags to use, and one can define test execution logic, as e.g. in:
https://github.com/typetools/checker-framework/blob/master/checker/jtreg/nullness/inheritDeclAnnoPersist/Implements.java#L8

It would be even better to migrate the existing `nullness-extra` tests to jtreg tests. OpenJDK javac now only uses jtreg tests. The benefit is platform independence - the Makefiles used in `nullness-extra` don't run the same everywhere.

So I would rather discourage adding things to `nullness-extra`. As a note here, I would probably highlight that in `nullness-extra` one can write their own Makefile, making it very easy to customize and adapt the testing logic. jtreg requires some getting used to. Is that a fair summary?

mernst: It's possible that I have not gotten used to jtreg yet. :-)

Can you show how to write the tests in https://github.com/typetools/checker-framework/pull/3137/files using jtreg?  I tried for hours, because I wanted to do it the right way, then wrote a makefile that did the job in a few minutes.

Seeing that would help me understand jtreg better, or would help to explain the problem that this comment is getting at.

wmdietl: Oh, I'm far from being a jtreg expert and I'm not sure how to do this.
The tricky part in the test for #3137 is that you perform some sed transformation on the generated output, before comparing it. I don't know whether there is a canonical way to do that in jtreg.

jtreg can run arbitrary things:
- https://github.com/typetools/checker-framework/blob/master/checker/jtreg/nullness/defaultsPersist/Constructors.java#L6
- https://github.com/typetools/checker-framework/blob/master/checker/jtreg/nullness/defaultsPersist/Driver.java

so I guess there would be a canonical way to do some post-processing on output.

But the difficult thing doesn't seem to be multiple files, it's performing post-processing and fine-grained error checks. Doing that is easier (for us) in Makefiles rather than in jtreg.
So how about leaving it at "we all know Makefiles better. If it is easy, add a jtreg test here. Otherwise, write a makefile and add the test under nullness-extra"?

mernst: I don't find that wording actionable.  It doesn't say in what circumstances to give up.  It also seems in conflict with your remark that jtreg is quite flexible and can run arbitrary things.

PR #3231, opened 2020-04-06
Separate "format of output" from "amount of detail"

mernst:


docs/manual/creating-a-checker.tex
```diff
@@ -1769,6 +1763,20 @@
...
+
+\subsectionAndLabel{Format of output}{creating-debugging-options-format}
+
+\begin{itemize}
+
+\item \code{-Adetailedmsgtext}: Output error/warning messages in a
+  stylized format that is easy for tools to parse.  This is useful for
+  tools that run the Checker Framework and parse its output, such as IDE
+  plugins.  See the source code of \<SourceChecker.java> for details about
+  the format.  This is now deprecated, given the existence of JSON output.
```
wmdietl, line 1775: I would remove the last sentence and add it once we integrate a pointer to the javac-diagnostics-wrapper.
Even with the JSON output, having more structure to the error messages could allow for better IDE integration, so I'm not quite sure whether it would be completely useless.

mernst: You are right.  That was just careless cut-and-paste.

PR #3234, opened 2020-04-07
Correct supertypes of CopyOnWriteArraySet.

cpovirk:
This fixes the warnings reported by StubParser.


cpovirk:
Thanks.

A belated realization/confession: What I have here _also_ produces a warning:

```
warning: StubParser: stub file does not match bytecode: could not find superinterface Set<E> from type CopyOnWriteArraySet<E extends Object>
```

I think that's because the class implements `Set` indirectly by extending `AbstractSet`.

I fear that I didn't test this at all because I switched away to another fix and got distracted. Alternatively, it could be that the new warning shows up only with `-AstubDebug`, which I'm now testing with.

Regardless, I think I have another PR to come, and I will fix this up then.

mernst:
Thanks for the fixes.  Your testing is helpful!

Issue #3236, opened 2020-04-07
collection-object-parameters-may-be-null.astub: Attempts to widen containsAll+retainAll parameter types appear not to work


cpovirk:
```java
$ cat ContainsAllRetainAll.java 
import java.util.Collection;

class ContainsAllRetainAll {
  void bulkOperations(Collection<?> a, Collection<?> b) {
    a.containsAll(b);
    a.retainAll(b);
  }
}
```

I would expect for this to fail with the new JDK stubs in 3.3.0 (and it does). However, I would expect it to pass when I supply `-Astubs=collection-object-parameters-may-be-null.astub`, yet it still fails:

```
$ checker/bin/javac -processor org.checkerframework.checker.nullness.NullnessChecker -Astubs=collection-object-parameters-may-be-null.astub ContainsAllRetainAll.java
warning: StubParser: annotateTypeParameters: mismatched sizes:  typeParameters (size 0)=[];  typeArguments (size 1)=[E extends Object];  decl=class AbstractCollection {      @Pure     public boolean contains(@Nullable Object o);      public boolean remove(@Nullable Object o);      @Pure     public boolean containsAll(Collection<?> c);      public boolean removeAll(Collection<?> c);      public boolean retainAll(Collection<?> c); };  elt=java.util.AbstractCollection (class com.sun.tools.javac.code.Symbol$ClassSymbol).; for more details, run with -AstubDebug
warning: StubParser: annotateTypeParameters: mismatched sizes:  typeParameters (size 0)=[];  typeArguments (size 2)=[K extends Object, V extends Object];  decl=class AbstractMap {      public boolean containsValue(@Nullable Object value);      public boolean containsKey(@Nullable Object key);      @Nullable     public V get(@Nullable Object key);      @Nullable     public V remove(@Nullable Object key); };  elt=java.util.AbstractMap (class com.sun.tools.javac.code.Symbol$ClassSymbol).; for more details, run with -AstubDebug
warning: StubParser: annotateTypeParameters: mismatched sizes:  typeParameters (size 0)=[];  typeArguments (size 1)=[E extends Object];  decl=class AbstractSet {      public boolean removeAll(Collection<?> c); };  elt=java.util.AbstractSet (class com.sun.tools.javac.code.Symbol$ClassSymbol).; for more details, run with -AstubDebug
[...]

PR #3237, opened 2020-04-07
Provide a stub for `Collection.removeAll`.

cpovirk:
However, note that this currently has no effect because of #3236.


cpovirk:
I think the intention was to do this for `Collection.removeAll`, like `Collection.containsAll` and _other_ `removeAll` methods, but I may have that wrong:

https://github.com/typetools/checker-framework/pull/3063#discussion_r398887563

PR #3239, opened 2020-04-08
Make per-checker options work for lint flags. Document the feature.

wmdietl:
We recently noticed that this feature isn't documented.

https://github.com/kelloggm/checker-framework/issues/237 recently noticed that it doesn't work for lint flags.


framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -783,6 +786,9 @@ public void initChecker() {
         this.messagesProperties = getMessagesProperties();
 
         this.visitor = createSourceVisitor();
+
+        // Validate the lint flags.
+        this.activeLints = createActiveLints(getOptions());
```
wmdietl, line 791: Without this change, executing `./checker/bin/javac -Alint=redundantNullComparison -processor nullness,regex  Repro.java` would not warn about an unknown lint flag, but running `./checker/bin/javac -Alint=redundantNullComparison -processor nullness,guieffect Repro.java` did produce a warning.
This now eagerly creates the active lint flags, always raising an error.
@smillst @mernst Does this sound good to you?

mernst: This seems good.  Thanks.

PR #3246, opened 2020-04-10
Move two visitors to visitor/ subdirectory

mernst:
Is this a desirable refactoring?


smillst:
Most of the other visitors in this package are interfaces, abstract classes, or otherwise intended to be subclassed.  (`AnnotatedTypeReplacer` is the only exception.)  So, I think it makes more sense to leave these two visitors where they are.  (And maybe to move `AnnotatedTypeReplacer` to `org.checkerframework.framework.type`.)

PR #3252, opened 2020-04-12
Fixes #3217: Lambdas: use erased type.

### LINKED ISSUE #3217: Lambda parameters for @Nullable are incorrect
The following code produces false positive with types changed quite dramatically in terms of nullness. 

#### Code:
```

public class Main<ModelA, ModelB, Value> {
    private final Function<Function<ModelA, @Nullable Value>, Function<ModelB, @Nullable Value>> proxyFunction;

    public Main(Function<Function<ModelA, @Nullable Value>, Function<ModelB, @Nullable Value>> proxyFunction) {
        this.proxyFunction = proxyFunction;
    }
}

public class SubClass<ModelA, Value> extends Main<ModelA, ModelA, Value> {
    public SubClass() {
        super(x -> x);
    }
}
```

#### The output of checker framework is:
```
> Task :example1:compileJava
/home/eirnym/t/example1/src/main/java/example1/SubClass.java:5: error: [lambda.param.type.incompatible] incompatible parameter types in lambda expression.
        super(x -> x);
              ^
  found   : @Initialized @NonNull Function<ModelA[ extends @Initialized @Nullable Object super @Initialized @NonNull Void], Value[ extends @Initialized @Nullable Object super @Initialized @Nullable Void]>
  required: @Initialized @NonNull Function<ModelA[ extends @Initialized @Nullable Object super @Initialized @NonNull Void], Value[ extends @Initialized @Nullable Object super @Initialized @NonNull Void]>
/home/eirnym/t/example1/src/main/java/example1/SubClass.java:5: error: [return.type.incompatible] incompatible types in return.
        super(x -> x);
                   ^
  type of expression: @Initialized @NonNull Function<ModelA[ extends @Initialized @Nullable Object super @Initialized @NonNull Void], Value[ extends @Initialized @Nullable Object super @Initialized @NonNull Void]>
  method return type: @Initialized @NonNull Function<ModelA[ extends @Initialized @Nullable Object super @Initialized @NonNull Void], Value[ extends @Initialized @Nullable Object super @Initialized @Nullable Void]>
2 errors
```
#### My Environment is:
* Gradle 6.3
* Java 8 & Java 11
* Gradle CheckerFramewrok plugin 0.4.13
* CheckerFramework 3.2.0
* Arch Linux

#### Additional notes:
You can also see the code and sample project in a [repository](https://github.com/eirnym/checkerframework-falsepositives/tree/master/example1) in module `example1`.

smillst:
Fixes #3217


framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
```diff
@@ -100,7 +101,8 @@ private static AnnotatedTypeMirror inferLambdaParamAnnotations(
             AnnotatedExecutableType functionType = f.getFunctionTypeFromTree(lambdaDecl);
             AnnotatedTypeMirror funcTypeParam = functionType.getParameterTypes().get(index);
             if (TreeUtils.isImplicitlyTypedLambda(declaredInTree)) {
-                if (f.types.isSubtype(funcTypeParam.actualType, lambdaParam.actualType)) {
+                if (TypesUtils.isErasedSubtype(
+                        funcTypeParam.actualType, lambdaParam.actualType, f.types)) {
                     // The Java types should be exactly the same, but because invocation type
```
wmdietl, line 106: Could you update the comment to explain why the erased type is enough?

Could you add a test case that uses generic types and make sure that this doesn't introduce false negatives? E.g. have a method that accepts a `List<String>` and pass something that is a `List<@Nullable String>`. If this only compares the `List` part, it will ignore the mismatch in the type arguments.
If this isn't a concern, please document why.

PR #3256, opened 2020-04-14
Enable cf-inference tests on jdk11

wmdietl:


xingweitian:
Should Travis also be enabled?

mernst:
If you are asking whether Travis should run on the `typetools` fork, the answer is "no".  The `typetools` for uses Azure Pipelines instead.  Travis is flaky and slow, so it wastes our time and slows us down.  These Travis tests are only for forks that choose not to enable Azure Pipelines and only enable Travis instead.  (I don't know of any reason to make that choice.)

xingweitian:
> If you are asking whether Travis should run on the `typetools` fork, the answer is "no". The `typetools` for uses Azure Pipelines instead. Travis is flaky and slow, so it wastes our time and slows us down. These Travis tests are only for forks that choose not to enable Azure Pipelines and only enable Travis instead. (I don't know of any reason to make that choice.)

Thanks for the reply! I am actually wondering how can we know that `cf-inference-jdk11` can pass on Travis now. It was commented out before, and now we are using Azure pipelines.

mernst:
Thanks for clarifying your question.
The simplest possibility is to create the branch in a fork that runs Travis, and only make the pull request after the branch passes CI.

wmdietl:
I've made the change in a fork with Travis enabled: https://github.com/typetools/checker-framework/compare/master...wmdietl:travis-cfi-11?expand=1
All tests pass:
https://travis-ci.org/github/wmdietl/checker-framework/builds/675321951

Even though we encourage Azure over Travis, it would seem nice to remove this inconsistency.

mernst:
Thanks for confirming!  I'm strongly in favor of this change (for consistency and to support people who still use Travis), and now there is no reason not to merge it.

Issue #3010, opened 2019-12-22
Minimize size of annotated JDK


zhangjiangqige:
At present the dependency of the annotated JDK 11 copied by the gradle task `copyJdk11Files` is ~24M, and it's possible to minimize those files with javaparser. After the minimization (of comments, body of methods, etc), the size of `checker.jar` can be decreased from ~14M to ~9M.

The minimizer can be found at [here](https://github.com/zhangjiangqige/try-javaparser/blob/master/src/main/java/org/zhangjiangqige/Minimizer.java), and the minimized JDK [here](https://github.com/zhangjiangqige/mini-jdk11).

Although it doesn't speed up the speed when using JDK 11, but it's still probably good. So I was wondering if there's a proper way to integrate the minimizer into this repo?

cc @wmdietl

wmdietl:
This was fixed in #3257.

PR #3257, opened 2020-04-14
Create stub files from Java source files

wmdietl:
Originally developed by @zhangjiangqige at https://github.com/zhangjiangqige/try-javaparser/blob/master/src/main/java/org/zhangjiangqige/Minimizer.java and mentioned in #3010.

Some refactorings and improved documentation compared to original.

Does `JavaStubifier` sound reasonable? Better name suggestions welcome!

@smillst I've run this on `checker/dist/jdk/jdk11` and it worked as expected. As discussed, could you separately look into integrating this into the build? We should try integration into the build even before Java 8 usage.

When merging, keep the co-author line below:

Co-authored-by: zhangjiangqige <zhangjiangqige@gmail.com>

### DIFF
--- a/docs/manual/annotating-libraries.tex
+++ b/docs/manual/annotating-libraries.tex
@@ -641,7 +641,7 @@ minimal stub files.
   mkdir project-stubs
   cp -R project/src project-stubs
   java -cp $CHECKERFRAMEWORK/checker/dist/checker.jar org.checkerframework.framework.stub.JavaStubifier project-stubs
-  find . -type f -name "*.java" -exec rename 's/.java$/.astub/' {} \;
+  find project-stubs -type f -name "*.java" -exec rename 's/.java$/.astub/' {} \;
 \end{Verbatim}
 
 Supply it a list of directories to process. Replacement happens in-place, so make sure to
--- a/docs/manual/annotating-libraries.tex
+++ b/docs/manual/annotating-libraries.tex
@@ -648,13 +648,14 @@ process a copy of your sources.
 
 You can now provide \<project-stubs> as a stub directory using \<-Astubs=project-stubs> as
 an additional command-line option.
-You can also concatenate all minimized \<.java> files into a single \<.astub> file:
+
+If you wish to use a single stub file, rather than a stub directory,
+you can concatenate all the minimized \<.java> files into a single \<.astub> file:
 
 \begin{Verbatim}
 find project-stubs/ -name "*.java" -type f | xargs cat > project.astub
 \end{Verbatim}
 
-and then pass just that single file.
 % That file will contain many non-unique import statements, but that shouldn't be harmful.
 
 
--- a/docs/manual/annotating-libraries.tex
+++ b/docs/manual/annotating-libraries.tex
@@ -633,6 +633,30 @@ it runs.
 % Furthermore, a programmer must search the stub file
 % for a given method rather than just skimming a few pages of method signatures.
 
+Alternatively, you can minimize source files to make them more suitable as stub files.
+Use the \<JavaStubifier> to convert, in-place, all \<.java> files in given directories into
+minimal stub files.
+
+\begin{Verbatim}
+  mkdir project-stubs
+  cp -R project/src project-stubs
+  java -cp $CHECKERFRAMEWORK/checker/dist/checker.jar org.checkerframework.framework.stub.JavaStubifier project-stubs
+\end{Verbatim}
+
+Supply it a list of directories to process. Replacement happens in-place, so make sure to
+process a copy of your sources.
+
+You can now provide \<project-stubs> as a stub directory using \<-Astubs=project-stubs> as
+an additional command-line option.
+You can also concatenate all minimized \<.java> files into a single \<.astub> file:
+
+\begin{Verbatim}
+find project-stubs/ -name "*.java" -type f | xargs cat > project.astub
+\end{Verbatim}
+
+and then pass just that single file.
+% That file will contain many non-unique import statements, but that shouldn't be harmful.
+
 
 \subsubsectionAndLabel{If you do not have access to the Java source code}{stub-creating-without-source}
 
--- a/docs/manual/annotating-libraries.tex
+++ b/docs/manual/annotating-libraries.tex
@@ -641,6 +641,7 @@ minimal stub files.
   mkdir project-stubs
   cp -R project/src project-stubs
   java -cp $CHECKERFRAMEWORK/checker/dist/checker.jar org.checkerframework.framework.stub.JavaStubifier project-stubs
+  find . -type f -name "*.java" -exec rename 's/.java$/.astub/' {} \;
 \end{Verbatim}
 
 Supply it a list of directories to process. Replacement happens in-place, so make sure to
--- a/framework/src/main/java/org/checkerframework/framework/stub/JavaStubifier.java
+++ b/framework/src/main/java/org/checkerframework/framework/stub/JavaStubifier.java
@@ -104,13 +104,21 @@ public class JavaStubifier {
 
     /** Visitor to processes one compilation unit. */
     private static class MinimizerVisitor extends ModifierVisitor<Void> {
+        /** Whether to consider members implicitly public. */
+        private boolean implicitlyPublic = false;
+
         @Override
         public ClassOrInterfaceDeclaration visit(ClassOrInterfaceDeclaration cid, Void arg) {
+            boolean prevIP = implicitlyPublic;
             if (cid.isInterface()) {
-                return cid;
+                // All members of interfaces are implicitly public.
+                implicitlyPublic = true;
             }
             super.visit(cid, arg);
             removeIfPrivateOrPkgPrivate(cid);
+            if (cid.isInterface()) {
+                implicitlyPublic = prevIP;
+            }
             return cid;
         }
 
@@ -174,6 +182,9 @@ public class JavaStubifier {
          */
         private boolean removeIfPrivateOrPkgPrivate(NodeWithAccessModifiers<?> node) {
             AccessSpecifier as = node.getAccessSpecifier();
+            if (implicitlyPublic) {
+                return false;
+            }
             if (as == AccessSpecifier.PRIVATE || as == AccessSpecifier.PACKAGE_PRIVATE) {
                 ((Node) node).remove();
                 return true;
--- a/framework/src/main/java/org/checkerframework/framework/stub/JavaStubifier.java
+++ b/framework/src/main/java/org/checkerframework/framework/stub/JavaStubifier.java
@@ -91,7 +91,8 @@ public class JavaStubifier {
                 mv.visit(cu, null);
                 if (cu.findAll(ClassOrInterfaceDeclaration.class).isEmpty()
                         && cu.findAll(AnnotationDeclaration.class).isEmpty()
-                        && cu.findAll(EnumDeclaration.class).isEmpty()) {
+                        && cu.findAll(EnumDeclaration.class).isEmpty()
+                        && !absolutePath.endsWith("package-info.java")) {
                     // All content is removed, delete this file.
                     new File(absolutePath.toUri()).delete();
                     res = Result.DONT_SAVE;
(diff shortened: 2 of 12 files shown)


framework/src/main/java/org/checkerframework/framework/stub/JavaStubifier.java
```diff
@@ -0,0 +1,181 @@
...
+import com.github.javaparser.utils.ProjectRoot;
+import com.github.javaparser.utils.SourceRoot;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Optional;
+
+/**
+ * Process Java source files in a directory to produce minimal stub files, by removing:
```
mernst, line 22: Does it do so in place?

smillst:
The tests are failing:
```
/__w/1/s/framework/src/main/java/org/checkerframework/framework/stub/JavaStubifier.java:174: warning: [unnecessary.equals] use of .equals can be safely replaced by ==/!=
            if (as.equals(AccessSpecifier.PRIVATE) || as.equals(AccessSpecifier.PACKAGE_PRIVATE)) {
                         ^
/__w/1/s/framework/src/main/java/org/checkerframework/framework/stub/JavaStubifier.java:174: warning: [unnecessary.equals] use of .equals can be safely replaced by ==/!=
            if (as.equals(AccessSpecifier.PRIVATE) || as.equals(AccessSpecifier.PACKAGE_PRIVATE)) {
```
```
framework/src/main/java/org/checkerframework/framework/stub/JavaStubifier.java:75:43: missing documentation for sm
framework/src/main/java/org/checkerframework/framework/stub/JavaStubifier.java:77:9: missing documentation for MinimizationCallback
```

I'll working on integrating this into the build, though that will be in another pull request.

framework/src/main/java/org/checkerframework/framework/stub/JavaStubifier.java
```diff
@@ -0,0 +1,181 @@
...
+
+        public MinimizationCallback(MinimizationVisitor sm) {
+            this.sm = sm;
+        }
+
+        @Override
+        public Result process(
+                Path localPath, Path absolutePath, ParseResult<CompilationUnit> result) {
+            Result res = Result.SAVE;
+            System.out.printf("Minimizing %s%n", absolutePath);
```
smillst, line 85: Can you either remove this or add a flag to turn it on/off?

wmdietl:
@smillst I'm running into a problem with the JavaParser: it does not process the `.java` files below `java.base`. It seems to be because that directory contains a `module-info.java` file.
I tried passing subdirectories, etc, but `java.base` was never processed.
I don't think the single `module-info.java` file contains anything useful, so you could filter it out:

```
diff --git a/checker/build.gradle b/checker/build.gradle
index 0c81c63c7..6152abe94 100644
--- a/checker/build.gradle
+++ b/checker/build.gradle
@@ -63,7 +63,8 @@ task copyJdk11Files(dependsOn: cloneTypetoolsJdk, group: 'Build') {
         FileTree tree = fileTree(dir: "${jdkHome}/src/")
         SortedSet<String> annotatedForFiles = new TreeSet<>();
         tree.visit { FileVisitDetails fvd ->
-            if (!fvd.file.isDirectory() && fvd.file.name.matches(".*\\.java")) {
+            if (!fvd.file.isDirectory() && fvd.file.name.matches(".*\\.java")
+                    && !fvd.file.name.matches("module-info.java")) {
                 fvd.getFile().readLines().any { line ->
                     if (line.contains("@AnnotatedFor") || line.contains("org.checkerframework")) {
                         annotatedForFiles.add(fvd.file.absolutePath)
```

I spent a bunch of time to get the `.java` files myself, but then I didn't figure out how to get JavaParser modify just that file without a `SourceRoot`...
Can you think of a better solution?

I have the issue with `ParserCollectionStrategy` minimized and will file an issue.

@mernst I added a section to the manual. Please have a look.

I didn't think of better javadocs. Most of the methods are private and the implementations are self-explanatory. I don't want to repeat the code for each, so "process" seems fine to me. Feel free to improve the javadoc if you feel strongly.

smillst:
>@smillst I'm running into a problem with the JavaParser: it does not process the .java files below java.base. 

@wmdietl, I can't reproduce this.  In fact, I was able to add JavaStubifier to the build.gradle to minimize the JDK.  I pushed those changes to this branch.  Can you give me steps to reproduce?

framework/src/main/java/org/checkerframework/framework/stub/JavaStubifier.java
```diff
@@ -0,0 +1,184 @@
...
+            }
+            return res;
+        }
+    }
+
+    /** Visitor to processes one compilation unit. */
+    private static class MinimizerVisitor extends ModifierVisitor<Void> {
+        @Override
+        public ClassOrInterfaceDeclaration visit(ClassOrInterfaceDeclaration cid, Void arg) {
+            if (cid.isInterface()) {
```
smillst, line 109: I had to add this; otherwise, the interface java files were just removed.  Methods in interfaces are public even if they are not marked as such.  An interface could have a default method whose body should be removed, so we may want to fix this some other way.

wmdietl:
> > @smillst I'm running into a problem with the JavaParser: it does not process the .java files below java.base.
> 
> @wmdietl, I can't reproduce this. In fact, I was able to add JavaStubifier to the build.gradle to minimize the JDK. I pushed those changes to this branch. Can you give me steps to reproduce?

Thanks for the build integration!
I've pulled your changes and then run `./gradlew clean assemble`. When I then look at file `checker/build/generated-resources/annotated-jdk/src/java.base/share/classes/java/lang/Object.java` it doesn't look minimized to me.
Once I removed `checker/build/generated-resources/annotated-jdk/src/java.base/share/classes/module-info.java` and run the minimizer again, it gets minimized.
Can you reproduce this?

framework/src/main/java/org/checkerframework/framework/stub/JavaStubifier.java
wmdietl: The logic in `removeIfPrivateOrPkgPrivate` looks at the visibility of the interface itself and removes an interface if it is private or package private.
If the interface itself isn't visible, why should we keep it?
Which interface did you look at?

smillst: `java.util.Map`, for example, is remove in MinimizerCallback#process because it has no methods marked as public or protected.  

 It's fine to remove the interface the interface is private or package private, which I suppose you can do by calling `removeIfPrivateOrPkgPrivate` before the return below.

smillst:
>I've pulled your changes and then run ./gradlew clean assemble. When I then look at file checker/build/generated-resources/annotated-jdk/src/java.base/share/classes/java/lang/Object.java it doesn't look minimized to me.
Once I removed checker/build/generated-resources/annotated-jdk/src/java.base/share/classes/module-info.java and run the minimizer again, it gets minimized.
Can you reproduce this?

No.  I tried it on buffalo, too.  Maybe your stubparser isn't up-to-date?

framework/src/main/java/org/checkerframework/framework/stub/JavaStubifier.java
wmdietl: I've added some logic to handle implicitly public members. Map looks fine to me and private interfaces are still removed.

wmdietl:
> > I've pulled your changes and then run ./gradlew clean assemble. When I then look at file checker/build/generated-resources/annotated-jdk/src/java.base/share/classes/java/lang/Object.java it doesn't look minimized to me.
> > Once I removed checker/build/generated-resources/annotated-jdk/src/java.base/share/classes/module-info.java and run the minimizer again, it gets minimized.
> > Can you reproduce this?
> 
> No. I tried it on buffalo, too. Maybe your stubparser isn't up-to-date?

I'm up-to-date with the last CF release. Pulled and rebuilt and same behavior. Nothing in java.base get's processed.

I made a repro-repo here: https://github.com/wmdietl/javaparser-module-bug
Could you let me know what output you see? I assume nothing changes for you, otherwise this would be even stranger...
@xingweitian Could you try the example, too?

xingweitian:
@wmdietl I can reproduce your demo on my computer. My java version: `openjdk version "11.0.1" 2018-10-16`

smillst:
@wmdietl I can reproduce it with your example.  I debugged through the JavaParser code.  Depending on when the module-info.java is encountered, the root may not be set to null.  So that's why I can't reproduce the problem with this pull request.  I think the best course of action is to skip module-info.java files, like you suggested:
```
         tree.visit { FileVisitDetails fvd ->
-            if (!fvd.file.isDirectory() && fvd.file.name.matches(".*\\.java")) {
+            if (!fvd.file.isDirectory() && fvd.file.name.matches(".*\\.java")
+                    && !fvd.file.name.matches("module-info.java")) {
                 fvd.getFile().readLines().any { line ->
                     if (line.contains("@AnnotatedFor") || line.contains("org.checkerframework")) {
                         annotatedForFiles.add(fvd.file.absolutePath)
```

wmdietl:
Thanks for confirming! I've filed https://github.com/javaparser/javaparser/issues/2615 and added the work-around.

framework/src/main/java/org/checkerframework/framework/stub/JavaStubifier.java
smillst: Thanks!

docs/manual/annotating-libraries.tex
```diff
@@ -633,6 +633,32 @@
...
 
+Alternatively, you can minimize source files to make them more suitable as stub files.
+Use the \<JavaStubifier> to convert, in-place, all \<.java> files in given directories into
+minimal stub files.
+
+\begin{Verbatim}
+  mkdir project-stubs
+  cp -R project/src project-stubs
+  java -cp $CHECKERFRAMEWORK/checker/dist/checker.jar org.checkerframework.framework.stub.JavaStubifier project-stubs
+  find project-stubs -type f -name "*.java" -exec rename 's/.java$/.astub/' {} \;
```
wmdietl, line 644: @mernst Is this the right command to rename the java files to astub files?

wmdietl:
@mernst plume_lib_jdk11 is failing https://dev.azure.com/checkerframework/checkerframework/_build/results?buildId=734&view=logs&j=3edf95e0-e0a0-576a-32b9-7f6083a2dca0&t=1e77a1c3-0666-5369-bfc3-ae7fc54485df&l=802 with:

```
tmp/plume-util/src/main/java/org/plumelib/util/StringBuilderDelimited.java:148: error: [lowerbound:return.type.incompatible] incompatible types in return.
    return delegate.length();
                          ^
  type of expression: @LowerBoundUnknown int
  method return type: @NonNegative int
1 error
```

but is passing for JDK 8. This looks unrelated to this PR.

mernst:
The build problem seems to be caused by the fact that `AbstractStringBuilder.java` does not appear in `build/generated-resources/annotated-jdk/src/java.base/share/classes/java/lang/`.  I think it should be there.

framework/src/main/java/org/checkerframework/framework/stub/JavaStubifier.java
```diff
@@ -0,0 +1,197 @@
...
+
+        @Override
+        public ClassOrInterfaceDeclaration visit(ClassOrInterfaceDeclaration cid, Void arg) {
+            boolean prevIP = implicitlyPublic;
+            if (cid.isInterface()) {
+                // All members of interfaces are implicitly public.
+                implicitlyPublic = true;
+            }
+            super.visit(cid, arg);
+            removeIfPrivateOrPkgPrivate(cid);
```
smillst, line 120: This is removing `AbstractStringBuilder` because it is a package private class.  However, some of the methods are exposed in public classes.

wmdietl: @mernst Thanks for noticing this!

So it looks like we can probably never remove a class. Even a private inner class can have a public subclass within the same outer class:

```java
class Demo {
    private class Internal {
        public void foo(@Nullable Object p) {}
    }
    public class Public extends Internal {}
  
    void bar(Public p) {
        p.foo(null);
    }
}
```

The public inner class could get exposed, so we need to keep it to maintain annotations from it.

Similarly, for `AbstractStringBuilder`, even though nobody from outside the package can refer to that class, there can be visible subclasses.
I'll remove that line for classes.

We kind-of have the same problem with methods: if a package-private method uses an inheritable annotation (like `@Pure`), the method affects overriding methods that might become public.
So we should also stop removing package-private methods.
It's still fine to remove private methods, because they cannot be overridden.

Removing private and package-private fields still seems ok to me.

Any other cases I'm missing?

mernst: That seems like a reasonable choice for now.  If the stubs remain too big (or if we have accidentally left something out), we can revisit the design later.

PR #3264, opened 2020-04-19
Remove unused methods from PluginUtil.

smillst:
Some of the methods were moved to `checkers.inference.InferenceLauncher`.

Merge with https://github.com/typetools/checker-framework-inference/pull/156.


javacutil/src/main/java/org/checkerframework/javacutil/PluginUtil.java
```diff
@@ -26,7 +26,10 @@
 /**
  * This file contains basic utility functions that should be reused to create a command-line call to
  * {@code CheckerMain}.
+ *
+ * @deprecated Renamed to {@link SystemUtil}
```
wmdietl, line 30: Not everything from here will be in `SystemUtil`. Should the document clarify that some methods will completely go away?

framework/src/main/java/org/checkerframework/framework/util/element/TargetedElementAnnotationApplier.java
```diff
@@ -134,8 +134,8 @@ protected void handleInvalid(List<Attribute.TypeCompound> invalid) {
                 remainingInfo.add(r.toString() + " (" + r.position + ")");
             }
             msg.add(remainingInfo.toString());
-            msg.add("Targeted annotations: " + PluginUtil.join(", ", annotatedTargets()));
-            msg.add("Valid annotations: " + PluginUtil.join(", ", validTargets()));
+            msg.add("Targeted annotations: " + SystemUtil.join(", ", annotatedTargets()));
```
wmdietl, line 137: Is there a reason not to use `String.join` for these? There is a `join` version with arrays and iterables, so aren't all of the `SystemUtil.join` uses replaceable?

smillst: No, String#join takes an iterable of CharSequences only.  `annotatedTargets()`, for example, is `TargetType[]`.

smillst:
@wmdietl I've assigned this to you because it can't be merged until you approve https://github.com/typetools/checker-framework-inference/pull/156.

PR #3265, opened 2020-04-22
Add task that tests the annotated jdk.

smillst:


checker/build.gradle
```diff
@@ -382,6 +382,12 @@ task jtregJdk11Tests(dependsOn: ':downloadJtreg', group: 'Verification') {
 
     }
 }
+
+task testAnnotatedJdk {
+    description 'Runs some of the tests that use the annotated jdk.'
```
mernst, line 387: Is this only some of the tests that use the annotated JDK?  Or is this some of the tests -- all those that use the annotated JDK?

mernst: Nit: capitalize JDK.

checker/build.gradle
```diff
@@ -382,6 +382,12 @@ task jtregJdk11Tests(dependsOn: ':downloadJtreg', group: 'Verification') {
 
     }
 }
+
+task testAnnotatedJdk {
+    description 'Runs tests that use the annotated JDK. (This does not run all the tests that use the annotated JDK.)'
```
mernst, line 387: What other tests exist?  Why does this task only run some of them?

smillst:
We decided to run allTests instead.

PR #3268, opened 2020-04-24
Fix memory leak in QualifierDefaults

zcai1:
@wmdietl @xingweitian 
This PR should fix the memory leak mentioned in https://github.com/opprop/checker-framework-inference/pull/263#issuecomment-618216146

The basic idea is to clear all cached `Element`s after each processor finished its job.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -717,8 +717,6 @@ subprojects {
             if (isJava8) {
                 jvmArgs "-Xbootclasspath/p:${configurations.javacJar.asPath}"
             } else {
-                // Without this, the test throw "java.lang.OutOfMemoryError: Java heap space"
-                forkEvery(1)
                 jvmArgs += [
                         "--illegal-access=warn",
                         "--add-opens", "jdk.compiler/com.sun.tools.javac.comp=ALL-UNNAMED",
--- a/framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
+++ b/framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
@@ -466,7 +466,7 @@ public abstract class SourceChecker extends AbstractTypeProcessor
      *
      * @see #activeOptions
      */
-    private static final String OPTION_SEPARATOR = "_";
+    protected static final String OPTION_SEPARATOR = "_";
 
     /**
      * The checker that called this one, whether that be a BaseTypeChecker (used as a compound
--- a/framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
+++ b/framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
@@ -62,6 +62,7 @@ import org.checkerframework.framework.type.AnnotatedTypeFactory;
 import org.checkerframework.framework.util.CFContext;
 import org.checkerframework.framework.util.CheckerMain;
 import org.checkerframework.framework.util.OptionConfiguration;
+import org.checkerframework.framework.util.defaults.QualifierDefaults;
 import org.checkerframework.javacutil.AbstractTypeProcessor;
 import org.checkerframework.javacutil.AnnotationProvider;
 import org.checkerframework.javacutil.AnnotationUtils;
@@ -465,7 +466,7 @@ public abstract class SourceChecker extends AbstractTypeProcessor
      *
      * @see #activeOptions
      */
-    protected static final String OPTION_SEPARATOR = "_";
+    private static final String OPTION_SEPARATOR = "_";
 
     /**
      * The checker that called this one, whether that be a BaseTypeChecker (used as a compound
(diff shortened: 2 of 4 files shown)


wmdietl:
Thanks for looking into this!

Can you remove the current hack at https://github.com/typetools/checker-framework/blob/master/build.gradle#L721 to get some indication that this fixes the issue?

Can you answer my question about the LRU cache from https://github.com/opprop/checker-framework-inference/pull/263#issuecomment-618457440 ?

zcai1:
> Thanks for looking into this!
> 
> Can you remove the current hack at https://github.com/typetools/checker-framework/blob/master/build.gradle#L721 to get some indication that this fixes the issue?
> 
> Can you answer my question about the LRU cache from [opprop/checker-framework-inference#263 (comment)](https://github.com/opprop/checker-framework-inference/pull/263#issuecomment-618457440) ?

1. Hack is now removed.
2. The cache never grows beyond size of 300. In terms of why it only happens in Java 11, I did reply in https://github.com/opprop/checker-framework-inference/pull/263#issuecomment-618607225. It seems the class is loaded multiple times in Java 8, but it's only loaded once in Java 11.

wmdietl:
> The cache never grows beyond size of 300.

How large are the objects in the cache?
What objects aren't being GCed b/c of the cache?


> It seems the class is loaded multiple times in Java 8, but it's only loaded once in Java 11.

That sounds a bit odd. Can you look into that a bit more? I didn't realize that there is such a class loading difference between the two.

zcai1:
> > The cache never grows beyond size of 300.
> 
> How large are the objects in the cache?
> What objects aren't being GCed b/c of the cache?
> 
> > It seems the class is loaded multiple times in Java 8, but it's only loaded once in Java 11.
> 
> That sounds a bit odd. Can you look into that a bit more? I didn't realize that there is such a class loading difference between the two.

@wmdietl 
1. In the memory dump, the cache occupied ~400M out of the 512M heap before crashing. However, this doesn't mean objects in the cache are large. From the screen shot in https://github.com/opprop/checker-framework-inference/pull/263#issuecomment-618216146, you can see that the elements in the cache have indirect reference to all the checkers, thus the checkers and everything used by the checkers can't be GCed (and that's as serious as leaking Activities in Android).

2. After some investigation, I think the root cause is we are loading external `javac` by the bootstrap class loader in Java 8, while we are using the internal `javac` in Java 11. <br/>
If you check the file "com/sun/tools/javac/file/BaseFileManager.java", there's a method called `getClassLoader(URL[] urls)` which is used to get the classloader for processor (checker). <br/>
In Java 8,
`ClassLoader thisClassLoader = getClass().getClassLoader();`
will assign `null` to the variable `thisClassLoader`, while in Java 11 the variable will be non-null. Citing the [java documentation](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getClassLoader--), "this method will return null in such implementations if this class was loaded by the bootstrap class loader." <br/>
`thisClassLoader` will be the parent of processor's classloader, and if it's null, processor's classloader, which is just instantiated using `new URLClassLoader`, has no idea if any classes were loaded before or not. This classloader will then reload everything it sees including `Checkers` and `QualiferDefaults`.<br/>
If you try running `(new URLClassLoader(urls, thisClassLoader)).loadClass("ostrusted.OsTrustedChecker")` in a debugger, it will always return a new class in Java 8, and the same class in Java 11. <br/>
I think this is a tricky difference between using the `bootstrapclasspath` and the internal library, and potentially it could cause more bugs in the future.

wmdietl:
@zcai1 Thanks for continuing this investigation!

As alternative to this PR I've started #3271, which makes the cache an instance field.
Could you look at that PR and confirm that it also fixes the memory issue? It works without the `forkEvery(1)`, so hopefully the issue is fixed.
We should work on removing all static caches throughout.

Could you file a follow-up issue with your observations about classloader differences between 8 and 11? Take your comment "2" above and expand on what we should investigate to fix the issue.

Thanks!

wmdietl:
@zcai1 Could you file a follow-up issue with your observations about classloader differences between 8 and 11? Take your comment "2" above and expand on what we should investigate to fix the issue.

zcai1:
> @zcai1 Could you file a follow-up issue with your observations about classloader differences between 8 and 11? Take your comment "2" above and expand on what we should investigate to fix the issue.

@wmdietl 
Sorry for the delay, and here's the issue I just created: https://github.com/typetools/checker-framework/issues/3282

I have also sent you an email for other PRs as well as scheduling a meeting.

zcai1:
Follow-up issue: https://github.com/typetools/checker-framework/issues/3282

zcai1:
Follow-up issue: https://github.com/typetools/checker-framework/issues/3282

PR #3271, opened 2020-04-26
Fix memory leak in QualifierDefaults

wmdietl:
By making a cache an instance field.
Original debugging and idea from @zcai1 in #3268 

Co-authored-by: zcai <12403206+zcai1@users.noreply.github.com>


zcai1:
@wmdietl @smillst Thank you! This PR looks good to me and I think it fixes the memory leak.

PR #3272, opened 2020-04-26
Add GenericAnnotatedTypeFactory#getExceptionalExitStore

smillst:
Add GenericAnnotatedTypeFactory#getExceptionalExitStore. Also, clean up the nearby Javadoc.


framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1195,13 +1224,21 @@ protected void analyze(
             if (regularExitStore != null) {
                 regularExitStores.put(method, regularExitStore);
             }
+            Store exceptionalExitStore = analysis.getExceptionalExitStore();
```
wmdietl, line 1227: It looks like that method implementation needs to be improved to me closer to how `getRegularExitStore` is implemented.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -828,6 +830,12 @@ public AnnotationMirror getAnnotationMirrorFromJavaExpressionString(
      */
     protected IdentityHashMap<Tree, Store> regularExitStores;
 
+    /**
+     * A mapping from methods (or other code blocks) to their exceptional exit store (used to check
```
wmdietl, line 834: We don't check postconditions on exception exit. JML would have syntax for that, if I remember correctly. But here we only check postconditions on regular exits.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -842,14 +850,34 @@ public AnnotationMirror getAnnotationMirrorFromJavaExpressionString(
...
+     * @param tree a MethodTree or other code block, such as a static initializer
+     * @return the exceptional exit store, or {@code null}, if there is no such store.
+     */
+    public @Nullable Store getExceptionalExitStore(Tree tree) {
+        return exceptionalExitStores.get(tree);
     }
 
-    /** @return all return node and store pairs for a given method */
+    /**
+     * Returns a list of all return statements of {@code method} paired with their the corresponding
```
wmdietl, line 873: The `their the` part is odd.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -842,14 +850,34 @@ public AnnotationMirror getAnnotationMirrorFromJavaExpressionString(
...
+     * {@link TransferResult}. If {@code method} has no return statement, then the empty list is
+     * returned.
+     *
+     * @param methodTree method whose return statements should be returned
+     * @return a list of all return statements of {@code method} paired with their the corresponding
```
wmdietl, line 878: Also a `their the`.

Issue #2047, opened 2018-06-21
java.util.List should be covariant


justmin:
```
        List<String> stringList = new ArrayList<>();
        List<@Nullable String> nullableStringList = stringList;

```
```
Error:(11, 53) java: [assignment.type.incompatible] incompatible types in assignment.
  found   : @Initialized @NonNull List<@Initialized @NonNull String>
  required: @UnknownInitialization @Nullable List<@Initialized @Nullable String>

```
It looks reasonable to make List be annotated with `@Covariant(0)`

mernst:
As [explained in the manual](https://checkerframework.org/manual/#invariant-arrays), this would be unsound.  If you translate the [example for arrays](https://checkerframework.org/manual/#invariant-arrays) to lists, you will see the problem.

PR #3278, opened 2020-04-27
Remove `@Covariant` from `Map`.

cpovirk:
It, like `List` (https://github.com/typetools/checker-framework/issues/2047), is not covariant:

```
import java.util.HashMap;
import java.util.Map;

class MapCovariant {
    Map<String, String> map = new HashMap<String, String>();
    Map<@Nullable String, String> sameMap = map;
    sameMap.put(null, "");
    map.keySet().iterator().next().toString(); // NPE
  }
}
```


cpovirk:
Merge with https://github.com/typetools/jdk/pull/49

PR #3283, opened 2020-05-01
Allow for parallel builds and tests

msridhar:
These Gradle configuration changes allow for parallel builds and parallel execution of tests (at the test class granularity) on multicore systems.  On my 6-core Mac Mini, this change reduces the time for a clean `./gradlew build` from ~6m50s to ~4m50s.  Improvements to incremental builds will of course depend on individual workflows (and on amount of RAM available, etc.).  If framework tests are run regularly in incremental builds, my feeling is this will yield a nice win for that case.


smillst:
Thanks!

PR #3284, opened 2020-05-01
Merge minimizeJdk with copyAnnotatedJdkFiles so that it only runs when needed.

smillst:


checker/build.gradle
```diff
@@ -58,8 +58,12 @@ task cloneTypetoolsJdk() {
 }
 
 task copyAnnotatedJdkFiles(dependsOn: cloneTypetoolsJdk, group: 'Build') {
+    // Make two copies of the annotated JDK. One to minimize and one to use in the up to date
+    // check of task minimizeJdk. See minimizeJdk for more details.
     def inputDir = "${jdkHome}/src"
-    def outputDir = "${buildDir}/generated-resources/annotated-jdk/"
+    def outputDir = "${buildDir}/generated-resources/not-minimized-annotated-jdk/"
```
smillst, line 64: Make sure this directory isn't in `checker.jar`

checker/build.gradle
```diff
@@ -50,8 +50,12 @@ task cloneTypetoolsJdk() {
 }
 
 task copyAnnotatedJdkFiles(dependsOn: cloneTypetoolsJdk, group: 'Build') {
+    // Make two copies of the annotated JDK. One to minimize and one to use in the up to date
```
smillst, line 53: Try creating an empty file instead of a copy

PR #3286, opened 2020-05-03
Test for issue 559 should now work with JDK 11.

smillst:


smillst:
This test case need to be rewritten.

wmdietl:
@smillst #559 is already closed. Should we open a new issue to keep track of these disabled tests?

PR #3292, opened 2020-05-05
Fix manual error to solve issue3258

QRXqrx:
Tha class name for PurityChecker is writen wrongly on page 125.
org.checkerframework.util.PurityChecker -> org.checkerframework.framework.util.PurityChecker

### DIFF
--- a/docs/manual/purity-checker.tex
+++ b/docs/manual/purity-checker.tex
@@ -10,7 +10,7 @@ All checkers utilize purity annotations on called methods.
 You do not need to run the Purity Checker directly.
 However you may run just the Purity Checker by
 supplying the following command-line options to javac:
-\code{-processor org.checkerframework.util.PurityChecker}.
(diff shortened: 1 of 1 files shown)


docs/manual/purity-checker.tex
```diff
@@ -10,7 +10,7 @@
 You do not need to run the Purity Checker directly.
 However you may run just the Purity Checker by
 supplying the following command-line options to javac:
-\code{-processor org.checkerframework.util.PurityChecker}.
+\code{-processor org.checkerframework.framework.util.PurityChecker}. # correct Class
```
mernst, line 13: As written, "# correct class" appears in the printed manual.  Please don't add comments about changes.

mernst:
@QRXqrx You don't need to close the pull request.  You can just make a change, commit it, and push the commit.  Then this pull request will be automatically updated.  This is a common workflow.

We appreciate your fix to the manual!

PR #3297, opened 2020-05-10
If two wildcards are equals, then one contains the other. (Fixes #3295)

### LINKED ISSUE #3295: error: Unexpected null second argument to areSameByName
```java
import java.util.List;

public class T {

  interface P<T> {

    Class<? super T> h();

    List<P<? super T>> g();
  }

  interface Q {}

  static void f(P<? extends Q> t) {
    t.g().stream().filter(x -> x.h().equals(Q.class));
  }
}
```

```shell
./checker-framework-3.4.0/checker/bin/javac -processor Nullness T.java
error: Unexpected null second argument to areSameByName
  ; The Checker Framework crashed.  Please report the crash.
  Compilation unit: T.java
  Last visited tree at line 17 column 27:
      t.g().stream().filter(x -> x.h().equals(Q.class));
  Exception: java.lang.Throwable; java.lang.Throwable
  	at org.checkerframework.javacutil.BugInCF.<init>(BugInCF.java:14)
  	at org.checkerframework.javacutil.AnnotationUtils.areSameByName(AnnotationUtils.java:116)
  	at org.checkerframework.javacutil.AnnotationUtils.areSame(AnnotationUtils.java:84)
  	at org.checkerframework.framework.type.StructuralEqualityComparer.arePrimeAnnosEqual(StructuralEqualityComparer.java:126)
  	at org.checkerframework.framework.type.StructuralEqualityComparer.defaultAction(StructuralEqualityComparer.java:66)
  	at org.checkerframework.framework.type.StructuralEqualityComparer.defaultAction(StructuralEqualityComparer.java:31)
...
error: Unexpected null second argument to areSameByName
  ; The Checker Framework crashed.  Please report the crash.
  Compilation unit: T.java
  Last visited tree at line 17 column 27:
      t.g().stream().filter(x -> x.h().equals(Q.class));
  Exception: java.lang.Throwable; java.lang.Throwable
  	at org.checkerframework.javacutil.BugInCF.<init>(BugInCF.java:14)
  	at org.checkerframework.javacutil.AnnotationUtils.areSameByName(AnnotationUtils.java:116)
  	at org.checkerframework.javacutil.AnnotationUtils.areSame(AnnotationUtils.java:84)
  	at org.checkerframework.framework.type.StructuralEqualityComparer.arePrimeAnnosEqual(StructuralEqualityComparer.java:126)
  	at org.checkerframework.framework.type.StructuralEqualityComparer.defaultAction(StructuralEqualityComparer.java:66)
  	at org.checkerframework.framework.type.StructuralEqualityComparer.defaultAction(StructuralEqualityComparer.java:31)
```

smillst:
Fixes #3295.


framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java
```diff
@@ -370,6 +370,11 @@ private boolean isContainedWildcard(
             AnnotatedTypeMirror outsideLowerBound,
             boolean canBeCovariant) {
 
+        if (inside.getKind() == TypeKind.WILDCARD
+                && outside.getKind() == TypeKind.WILDCARD
+                && inside.equals(outside)) {
+            return true;
```
wmdietl, line 376: Could you add a comment why this is the right check?
Or, as we discussed last week, add a comment to the method or maybe even directly on `AnnotatedWildcardType` that wildcards are used also for captured variables. All logic that uses AWT should be investigated once we have a separate captured type.

framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java
```diff
@@ -370,6 +370,11 @@ private boolean isContainedWildcard(
             AnnotatedTypeMirror outsideLowerBound,
             boolean canBeCovariant) {
 
+        if (inside.getKind() == TypeKind.WILDCARD
```
wmdietl, line 373: Why is this only for wildcards? Isn't it enough if `inside.equals(outside)` returns true?

PR #3298, opened 2020-05-10
Update Lombok example.

smillst:


smillst:
It seems that the Gradle Plugin stop working with JDK 11 starting with version 0.4.16.  @kelloggm can you take a look please?

PR #3299, opened 2020-05-10
Run cloneAndBuildDependencies if afu and stubparser jars don't exist.

smillst:


build.gradle
```diff
@@ -228,6 +228,13 @@ task cloneAndBuildDependencies(type: Exec) {
     description 'Clones (or updates) and builds all dependencies'
     executable 'checker/bin-devel/build.sh'
 }
+task maybeCloneAndBuildDependencies(type: Exec) {
+    description 'Clones (or updates) and builds all dependencies if they are not present.'
+    onlyIf {
+        !file(afuJar).exists() || !file(stubparserJar).exists()
```
mernst, line 234: Should this also depend on `jdk` being present?  cloneAndBuildDependencies creates it.

PR #3301, opened 2020-05-11
Explicitly calling constructors in KeyForAnnotatedTypeFactory

zcai1:
This allows the `KeyForAnnotatedTypeFactory` to be reused by other checkers. 
@wmdietl


wmdietl:
@zcai1 Thanks for the fix!

PR #3304, opened 2020-05-12
Add test for polymorphic qualifiers in SubtypeOf

lnsun:
While polymorphic qualifiers cannot have [`SubtypeOf`](https://github.com/typetools/checker-framework/blob/11262c72ef210d756b117c186fe62dff969e0832/framework/src/main/java/org/checkerframework/framework/qual/SubtypeOf.java), it should also not appears *in* the parameters of `SubtypeOf` of any other qualifiers. Otherwise, if a user do that, when building the qualifier hierarchy for the visitor, the constructor will try to find the subtype of the polymorphic qualifier (and it does not exists in the mapping), thus result into a NPE.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -400,7 +400,7 @@ task requireJavadoc(type: JavaExec) {
 
 
 /**
- * Creates a task named taskName that runs javadoc.
+ * Creates a task named taskName that runs javadoc with the -Xdoclint:all option.
  *
  * @param taskName the name of the task to create
  * @param taskDescription description of the task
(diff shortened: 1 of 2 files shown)


mernst:
Thanks for the fix!  Could you please add a test case to your pull request?  That will clarify its purpose, ensure that it works, and ensure that no one breaks it in the future.

mernst:
This is fine with me.  We should add tests when convenient, but it's OK not to write one here.  Thanks.

wmdietl:
@lnsun Thanks for the fix!

PR #3199, opened 2020-03-26
Returns Receiver Checker

nimakarimipour:
Add Returns Receiver Checker as a sub-checker to checkerframework.common package.

PR #3306, opened 2020-05-12
Add a generic accumulation checker

kelloggm:
This PR builds on ~#3199 (Returns Receiver Checker)~ by adding a generic checker that implements a basic general version of accumulation analysis, as described in our paper ["Verifying Object Construction"](https://homes.cs.washington.edu/~mernst/pubs/object-construction-icse2020.pdf) (ICSE 2020). 

A simple checker that implements a simplified version of the called methods type system described in the paper is included for testing.

This PR is the first step in what will be a series to generalize the implementation from the paper and make it re-usable.


kelloggm:
Sorry the commit log is a bit messy - I started working on this before the RR checker PR was merged, so I still have some of its intermediate commits in my history. It won't matter once this PR's commits are squashed when it is merged.

mernst:
Could you add documentation to the manual?

mernst:
There are a lot of CI failures.  Could you fix them before assigning reviewers (unless you are looking for feedback on the code that is causing the test failures)?

msridhar:
I think it may be premature to add manual documentation.  @kelloggm indicates that this is the first in a series of PRs that will result in a fully re-usable accumulation checker.  I think it's reasonable to wait until we get to a more final stage before requiring manual documentation.  In the meantime, this can live on the master branch without documentation, and this way we can do code reviews of more manageable chunks of code.  (Already this PR has a lot of code to look at.)

@kelloggm I'm also curious about the CI failures.  But I presume you just want any and all comments on the code at this point?

kelloggm:
@msridhar @mernst sorry for the CI failures - some code that was originally written for the object construction checker included some Java 11 syntax, which needed to be rewritten to Java 8. CI is now passing, except for the failures that are also on the master branch of the framework.

@msridhar is right that I intended this feature to be undocumented, for now. To allay @mernst's concern, I wrote a draft manual section but did not actually add it to the manual.

docs/manual/accumulation-checker.tex
```diff
@@ -0,0 +1,80 @@
...
+
+\paragraphAndLabel{Defining type qualifiers}{accumulation-qualifiers}
+
+The first step in building an accumulation checker is to decide what it is
+your checker will accumulate, and how to represent it. The Checker Framework's
+support for accumulation analysis requires you to accumulate a string representation
+of whatever you are accumulating. For example, when accumulating which methods have
+been called on an object, you might choose to accumulate method names.
+
+Define a type qualifier that with a single argument: a \<String[]> named \<value>.
```
msridhar, line 39: with -> has

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,280 @@
...
+    @Override
+    public QualifierHierarchy createQualifierHierarchy(MultiGraphFactory factory) {
+        return new AccumulationQualifierHierarchy(factory);
+    }
+
+    /**
+     * All accumulation analyses share a similar type hierarchy. This hierarchy implements the
+     * subtyping, LUB, and GLB for that hierarchy. The lattice looks like:
+     *
+     * <p>top / \ acc(x) acc(y) ... \ / acc(x,y) ... | bottom
```
msridhar, line 171: FWIW I can't really read this lattice.  Is this a standard notation?

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationChecker.java
```diff
@@ -0,0 +1,30 @@
...
+ * AccumulationAnnotatedTypeFactory}, which every subclass should override to provide custom
+ * annotations.
+ */
+public abstract class AccumulationChecker extends BaseTypeChecker {
+
+    @Override
+    protected LinkedHashSet<Class<? extends BaseTypeChecker>> getImmediateSubcheckerClasses() {
+        LinkedHashSet<Class<? extends BaseTypeChecker>> checkers =
+                super.getImmediateSubcheckerClasses();
+        checkers.add(ReturnsReceiverChecker.class);
```
msridhar, line 27: Is it possible someone might want to build an accumulation checker that does not require returns-receiver reasoning?  Or shall we just assume any accumulation checker is also dealing with fluent APIs?

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
```diff
@@ -0,0 +1,151 @@
...
+        AnnotatedTypeMirror oldType = typeFactory.getAnnotatedType(node.getTree());
+        AnnotationMirror newAnno = getNewAnno(oldType, values);
+        insertIntoStores(result, node, newAnno);
+    }
+
+    /**
+     * Updates the estimate for the receiver and any other receiver-returning methods in a chain
+     * with this MethodInvocationNode to include the new values.
+     *
+     * <p>For example, if the argument is the expression {@code a.b().c()}, the new value is "foo",
```
msridhar, line 52: This will only be true if methods `b` and `c` are deterministic.  If they are not, only the estimate of `a` will be updated, and the types of `a.b()` and `a.b().c()` are computed in the type factory.  We should update the documentation to be clearer about this

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
```diff
@@ -0,0 +1,151 @@
...
+            MethodInvocationNode node, TransferResult<CFValue, CFStore> result, String... values) {
+        Node receiver = node.getTarget().getReceiver();
+        AnnotatedTypeMirror oldType = typeFactory.getReceiverType(node.getTree());
+        // e.g. if the node being visited is static
+        if (oldType == null) {
+            return;
+        }
+        AnnotationMirror newAnno = getNewAnno(oldType, values);
+        while (receiver != null) {
+            insertIntoStores(result, receiver, newAnno);
```
msridhar, line 70: As noted above, this call will do nothing when `receiver` is a method call that is not deterministic.  Not sure if we should change the implementation logic to reflect this more clearly.  Nothing incorrect happens now, but it may confuse future readers of the code

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
kelloggm: The formatter messed this up. Sorry about that.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationChecker.java
kelloggm: I assumed that anyone building an accumulation checker would want RR reasoning - there's not much reason to *not* include it. I suppose we could make it configurable, somehow (but then again, it already is - someone who didn't want the RR checker could override this method again!).

I'll add something to the javadoc that mentions it.

kelloggm:
@msridhar thanks for the comments. I think I've addressed them all; let me know if I missed something.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationChecker.java
msridhar: There _is_ the whole including a sub-checker doubles the overall checking time issue... 🙂But this is fine for now

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,288 @@
...
+        return AnnotationUtils.areSameByClass(anm, ACC);
+    }
+
+    @Override
+    protected TreeAnnotator createTreeAnnotator() {
+        return new ListTreeAnnotator(
+                super.createTreeAnnotator(), new AccumulationTreeAnnotator(this));
+    }
+
+    /** Handles fluent APIs using the Returns Receiver Checker. */
```
msridhar, line 126: Can we add a bit more detail on what this tree annotator is actually doing?  Like the example we give in the transfer function documentation.  This can be in the class documentation or the docs of `visitMethodInvocation()`.

kelloggm: I added it to the docs for `getMethodInvocation()`

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
```diff
@@ -135,6 +135,14 @@ public AccumulationTreeAnnotator(AccumulationAnnotatedTypeFactory factory) {
             super(factory);
         }
 
+        /**
+         * If a method returns its receiver, and the receiver has an accumulation type, then the
```
msridhar, line 139: Wait, isn't the point that this annotator takes the receiver type and "accumulates in" the current method invocation?

kelloggm: Only the Object Construction Checker does that. This version is simpler - all it does it make the return type match the receiver type if the method returns `this`. The `TestAccumulationChecker` handles actually accumulating method calls differently than the OCC, also - note that it does the actual accumulation in the `Transfer` class.

msridhar: I get it now.  Since the transfer function computes the complete accumulated type for the most deeply-nested receiver, all the type factory has to do is to copy that type to the return value for each call and it will get bubbled up.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
```diff
@@ -0,0 +1,161 @@
...
+
+    /**
+     * Updates the estimate for the receiver and any other receiver-returning methods in a chain
+     * with this MethodInvocationNode to include the new values.
+     *
+     * <p>For example, if the argument is the expression {@code a.b().c()}, the new value is "foo",
+     * and b and c return their receiver (and are deterministic), all of the expressions {@code a},
+     * {@code a.b()}, and {@code a.b().c()} would have their estimates updated to include "foo". If
+     * any method in the chain is non-deterministic, its estimate will not be updated (but the rest
+     * of the chain is not affected).
```
mernst, line 62: Does "not affected" mean they are not updated, or they are updated as if all the methods had been deterministic, or they are updated (not "not updated"), but maybe in a different way than if all the methods had been deterministic?

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
```diff
@@ -0,0 +1,161 @@
...
+                receiver = ((MethodInvocationNode) receiver).getTarget().getReceiver();
+            } else {
+                // Do not continue, because the method does not return @This.
+                break;
+            }
+        }
+    }
+
+    /**
+     * Combines the values in oldType with the values in newValues to produce a single accumulator
```
mernst, line 110: Please be more specific.  Is this union?

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
```diff
@@ -0,0 +1,161 @@
...
+     * type qualifier.
+     *
+     * @param oldType an annotated type mirror whose values should be included
+     * @param newValues new values to include
+     * @return an annotation representing all the values
+     */
+    private AnnotationMirror getNewAnno(AnnotatedTypeMirror oldType, String[] newValues) {
```
mernst, line 117: `getNewAnno` is a vague name that I found confusing at call sites.  Please rename it.  Should "New" in the name be replaced by "Glb"?  (Or possibly by "Union"?)

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
```diff
@@ -0,0 +1,161 @@
...
+        AnnotationMirror oldAnno;
+        if (oldType == null) {
+            oldAnno = typeFactory.TOP;
+        } else {
+            oldAnno = oldType.getAnnotationInHierarchy(typeFactory.TOP);
+            if (oldAnno == null || !typeFactory.isAccumulatorAnnotation(oldAnno)) {
```
mernst, line 123: How can `!typeFactory.isAccumulatorAnnotation(oldAnno)` be true?  Is that an error?  Or is it a funny way to test for `BOTTOM` and (in some type systems) `TOP`?

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,296 @@
...
+        if (this.getClass() == AccumulationAnnotatedTypeFactory.class) {
+            this.postInit();
+        }
+    }
+
+    /**
+     * Creates a new instance of the accumulator annotation that contains the elements of {@code
+     * values} in sorted order.
+     *
+     * @param values the arguments to the annotation
```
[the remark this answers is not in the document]
kelloggm: I think that is only the case if an explicit array is passed. I added a note to the documentation.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,296 @@
...
+
+    /**
+     * Returns true if the return type of the given method invocation tree has an @This annotation
+     * from the Returns Receiver Checker.
+     *
+     * @param tree a method invocation tree
+     * @return true if the method being invoked returns its receiver
+     */
+    public boolean returnsThis(final MethodInvocationTree tree) {
+        ReturnsReceiverAnnotatedTypeFactory rrATF =
```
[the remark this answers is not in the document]
kelloggm: AnnotatedTypeFactories must not put subchecker type factories into fields. I have learned this the hard way before. In many other checkers, I've created a `getSpecificSubchecker()` method for each subchecker, but I didn't think that was really necessary here. See the documentation for `getTypeFactoryOfSubchecker()`: https://github.com/typetools/checker-framework/blob/master/framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java#L1585

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,296 @@
...
+        return rrType != null && rrType.hasAnnotation(This.class);
+    }
+
+    /**
+     * Is the given annotation an accumulator annotation?
+     *
+     * @param anm an annotation mirror
+     * @return true if the annotation mirror is an instance of this factory's accumulator annotation
+     */
+    public boolean isAccumulatorAnnotation(AnnotationMirror anm) {
```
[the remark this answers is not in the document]
kelloggm: It is definitely false of both. I think the documentation is clear, with the changes I've made elsewhere to make clear that the top, bottom, and accumulator annotations are expected to be distinct.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,296 @@
...
+    protected TreeAnnotator createTreeAnnotator() {
+        return new ListTreeAnnotator(
+                super.createTreeAnnotator(), new AccumulationTreeAnnotator(this));
+    }
+
+    /** Handles fluent APIs using the Returns Receiver Checker. */
+    protected class AccumulationTreeAnnotator extends TreeAnnotator {
+
+        /**
+         * Mandatory constructor.
```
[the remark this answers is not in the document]
kelloggm: Java requires you to write a constructor for classes that don't have a default constructor. It is therefore "mandatory" to write one. More to the point, there's nothing to say about this constructor - it just calls super - but Javadoc is required. I never know what to write in these cases, so I write something inane to satisfy the checker.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,296 @@
...
+         *
+         * @param factory the type factory
+         */
+        public AccumulationTreeAnnotator(AccumulationAnnotatedTypeFactory factory) {
+            super(factory);
+        }
+
+        /**
+         * If a method returns its receiver, and the receiver has an accumulation type, then the
```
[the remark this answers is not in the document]
kelloggm: I removed the comment on the method, and moved this into a list of the rules the annotator is responsible for implementing.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,296 @@
...
+            if (returnsThis(tree)) {
+                // There is a @This annotation on the return type of the invoked method.
+
+                // Fetch the current type of the receiver, or top if none exists.
+                ExpressionTree receiverTree = TreeUtils.getReceiverTree(tree.getMethodSelect());
+                AnnotatedTypeMirror receiverType;
+                AnnotationMirror receiverAnno;
+
+                if (receiverTree != null
+                        && (receiverType = getAnnotatedType(receiverTree)) != null) {
```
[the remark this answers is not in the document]
kelloggm: This particular code snippet is hard to read either way, and using an assignment within a conditional less verbose. Nevertheless, I've changed it to comply with the Checker Framework's typical code style, because an assignment within a conditional will be more confusing if they are not commonly used in a particular code base.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,296 @@
...
+                    receiverAnno = receiverType.getAnnotationInHierarchy(TOP);
+                } else {
+                    receiverAnno = TOP;
+                }
+
+                type.replaceAnnotation(receiverAnno);
```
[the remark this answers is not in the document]
kelloggm: Good catch!

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,296 @@
...
+            if (!AnnotationUtils.hasElementValue(a2, "value")) {
+                return a1;
+            }
+
+            if (isAccumulatorAnnotation(a1) && isAccumulatorAnnotation(a2)) {
+                Set<String> a1Val =
+                        new LinkedHashSet<>(
+                                ValueCheckerUtils.getValueOfAnnotationWithStringArgument(a1));
+                Set<String> a2Val =
+                        new LinkedHashSet<>(
```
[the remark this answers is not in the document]
kelloggm: To follow up on this, I just looked through how the Value Checker handles subtyping, etc. between `@StringVal` annotations, and I'm going to do the same thing here. It uses `List` rather than converting to sets in the `QualifierHierarchy`, and sorts and removes at the same time by converting to a `TreeSet` and back in the method that produces new `StringVal` annotations.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
```diff
@@ -0,0 +1,161 @@
...
+ * {@link #accumulate(MethodInvocationNode, TransferResult, String...)} methods to accumulate a
+ * string at a particular program point.
+ */
+public class AccumulationTransfer extends CFTransfer {
+
+    /** The type factory. */
+    protected final AccumulationAnnotatedTypeFactory typeFactory;
+
+    /**
+     * Required constructor.
```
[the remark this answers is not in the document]
kelloggm: see my comment about "mandatory" above. In my opinion, there's no need for Javadoc for these constructors.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
```diff
@@ -0,0 +1,161 @@
...
+        insertIntoStores(result, node, newAnno);
+    }
+
+    /**
+     * Updates the estimate for the receiver and any other receiver-returning methods in a chain
+     * with this MethodInvocationNode to include the new values.
+     *
+     * <p>For example, if the argument is the expression {@code a.b().c()}, the new value is "foo",
+     * and b and c return their receiver (and are deterministic), all of the expressions {@code a},
+     * {@code a.b()}, and {@code a.b().c()} would have their estimates updated to include "foo". If
```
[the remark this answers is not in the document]
kelloggm: Thanks for pointing this out. This method didn't actually do what it said it did. I've consolidated the two `accumulate` methods, and their documentation should now be accurate. There was a corresponding error in `TestAccumulationTransfer` where `accumulate` was called on the method invocation (i.e. the return type), not the receiver, as it should have been. These two errors together allowed the tests to pass.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
kelloggm: I prefer "combined", but if that's still confusing I can change it to "union", which I think sounds weird when you say it out loud

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
kelloggm: It was a funny way to test for bottom or top. It turns out it isn't even necessary, so I removed it.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
```diff
@@ -0,0 +1,152 @@
...
+     * type will also be updated. This chain will continue as long as each receiver is itself a
+     * receiver-returning method invocation.
+     *
+     * <p>For example, suppose the argument is the expression {@code a.b().c()}, the new value is
+     * "foo", and b and c return their receiver (and are deterministic). Then all of the expressions
+     * {@code a}, {@code a.b()}, and {@code a.b().c()} would have their estimates updated to include
+     * "foo".
+     *
+     * <p>If any method in the chain is non-deterministic, its estimate will not be updated (but the
+     * types of any deterministic methods, and the ultimate receiver, will be updated).
```
mernst, line 57: "any deterministic methods": they might follow it, as in `a.b().nondet().c().d()` where b, c, and d are deterministic.  Do the values go through `nondet()` and affect all of b(), c(), and d()?

mernst, line 57: "the ultimate receiver": this isn't clear.  I think you mean `a` in your examples, but I'm not sure.  The receiver value is identical, so I'm not sure what this means.  It's only the expressions that differ.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
```diff
@@ -0,0 +1,152 @@
...
+     * @param result the transfer result containing the store to be modified
+     * @param values the new accumulation values
+     */
+    public void accumulate(Node node, TransferResult<CFValue, CFStore> result, String... values) {
+        Tree tree = node.getTree();
+        if (tree == null) {
+            return;
+        }
+        AnnotatedTypeMirror oldType = typeFactory.getAnnotatedType(tree);
+        AnnotationMirror newAnno = getCombinedAnno(oldType, Arrays.asList(values));
```
mernst, line 69: Why is `asList` called?  The callee, `getCombinedAnno`, will make a copy anyway.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
```diff
@@ -0,0 +1,152 @@
...
+    /**
+     * Unions the values in oldType with the values in newValues to produce a single accumulator
+     * type qualifier.
+     *
+     * @param oldType an annotated type mirror whose values should be included, or null which is
+     *     treated as top
+     * @param newValues new values to include
+     * @return an annotation representing all the values
+     */
+    private AnnotationMirror getCombinedAnno(
```
mernst, line 112: The documentation is clear, but I still find "getCombinedAnno" unclear since it's not the annotations that are being combined (lubbed) but the values that are being combined (unioned).  I would change "Combined" in the method name to "glb" or "union" or "accumulated".

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
```diff
@@ -0,0 +1,152 @@
...
+                receiver = ((MethodInvocationNode) receiver).getTarget().getReceiver();
+            }
+        }
+    }
+
+    /**
+     * Unions the values in oldType with the values in newValues to produce a single accumulator
+     * type qualifier.
+     *
+     * @param oldType an annotated type mirror whose values should be included, or null which is
```
mernst, line 107: Why does this method treat null specially?  Are there places that null is more convenient to call?

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
```diff
@@ -0,0 +1,152 @@
...
+     *     treated as top
+     * @param newValues new values to include
+     * @return an annotation representing all the values
+     */
+    private AnnotationMirror getCombinedAnno(
+            @Nullable AnnotatedTypeMirror oldType, List<String> newValues) {
+        AnnotationMirror oldAnno;
+        if (oldType == null) {
```
mernst, line 115: Is the logic in these lines similar to that in visitMethodInvocation?  If so, consider abstracting it out.
Or, since getCombinedAnnotation is only called once, you could create a method getAnnotationMirror that calls getAnnotatedType and then calls this logic.  That seems like a logical place for it, and then this method could just take an AnnotationMirror as an argument.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
```diff
@@ -0,0 +1,152 @@
...
+            oldAnno = typeFactory.top;
+        } else {
+            oldAnno = oldType.getAnnotationInHierarchy(typeFactory.top);
+            if (oldAnno == null) {
+                oldAnno = typeFactory.top;
+            }
+        }
+        List<String> allValues = new ArrayList<>(newValues);
+        if (typeFactory.isAccumulatorAnnotation(oldAnno)) {
```
mernst, line 124: I think this is a way of saying "not top or bottom".  It's surprising that combining any A with top has the same effect as combining A with bottom.  Is that intentional?  Should combining A with bottom lead to bottom?

kelloggm: I rewrote this method to make this test unnecessary (it now uses GLB).

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
kelloggm: I rewrote the logic here to make it clear that they're not quite the same - in the annotated type factory, it's not necessary to check for the annotation mirror being null, but here it is. So I don't think there's any need to abstract the one or two lines they share.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
kelloggm: This is no longer a concern with the other changes I've made.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,279 @@
...
+     */
+    private final Class<? extends Annotation> accumulator;
+
+    /**
+     * Create an annotated type factory for an accumulation checker.
+     *
+     * @param checker the checker
+     * @param accumulator the accumulator type in the hierarchy. Must be an annotation with a single
+     *     argument named "value" whose type is a String array.
+     * @param top the top type in the hierarchy, which must be a supertype of {@code accumulator}.
```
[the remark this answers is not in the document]
kelloggm: This is a good idea. I modeled the general accumulation type system on `@CalledMethods`, which has a separate top type, but I'm no longer convinced it's needed there, either.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,279 @@
...
+     * with exactly one field named "value" whose type is a String array.
+     */
+    private final Class<? extends Annotation> accumulator;
+
+    /**
+     * Create an annotated type factory for an accumulation checker.
+     *
+     * @param checker the checker
+     * @param accumulator the accumulator type in the hierarchy. Must be an annotation with a single
+     *     argument named "value" whose type is a String array.
```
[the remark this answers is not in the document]
kelloggm: That's a good point. I've added tests for all these requirements.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
kelloggm: yes, as long as `nondet` returns `@This`. I think this is just an imprecision caused by the dataflow framework rather than something fundamental, since the RR information should make the fact that the method isn't known to be deterministic irrelevant.

kelloggm: I do mean `a`. Really, what was meant was the first receiver expression that isn't an invocation of a method that returns its receiver. I'll update the text.

kelloggm:
@mernst I've addressed your comments, and this is ready for another look. Thanks!

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,311 @@
...
+            Class<? extends Annotation> accumulator,
+            Class<? extends Annotation> bottom) {
+        super(checker);
+
+        this.accumulator = accumulator;
+
+        // Check that the requirements of the accumulator are met.
+        Method[] accDeclaredMethods = accumulator.getDeclaredMethods();
+        if (accDeclaredMethods.length != 1) {
+            rejectMalformedAccumulator("not have more than one argument");
```
mernst, line 74: "argument" should be "element".
The test also fails if it has zero elements, so the message may be misleading.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,311 @@
...
+        }
+        Method value = accDeclaredMethods[0];
+        if (value.getName() != "value") {
+            rejectMalformedAccumulator("name its argument \"value\"");
+        }
+        if (!value.getReturnType().isInstance(new String[0])) {
+            rejectMalformedAccumulator("have an argument of type String[]");
```
mernst: This check, and the next one, are not about the annotation but about its `value` element, so the string created by `rejectMalformedAccumulator` is imprecise and possibly confusing.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    /**
+     * Common error message for malformed accumulator annotation.
+     *
+     * @param missing what is missing from the accumulator, suitable for use in this string to
+     *     replace $MISSING$: "The accumulator annotation Foo must $MISSING$."
+     */
+    private void rejectMalformedAccumulator(String missing) {
+        throw new UserError(
```
mernst: A UserError is when a user passed wrong arguments to the Checker Framework.  It can be corrected by changing the run-time invocation.  A bug in the source code of a checker should be thrown as a BugInCF., even if that source code was written by someone other than the main Checker Framework developers.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
```diff
@@ -0,0 +1,140 @@
...
+     * <p>If the node is an invocation of a method that returns its receiver, then its receiver's
+     * type will also be updated. This chain will continue as long as each receiver is itself a
+     * receiver-returning method invocation.
+     *
+     * <p>For example, suppose the argument is the expression {@code a.b().c()}, the new value is
+     * "foo", and b and c return their receiver (and are deterministic). Then all of the expressions
+     * {@code a}, {@code a.b()}, and {@code a.b().c()} would have their estimates updated to include
+     * "foo".
+     *
+     * <p>If any method in the chain is non-deterministic, its estimate will not be updated (but the
```
mernst, line 53: "the chain" isn't really defined.  Can you define that term before you start discussing the behavior of the method?

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
```diff
@@ -0,0 +1,140 @@
...
+
+    /**
+     * Updates the estimate of how many things {@code node} has accumulated.
+     *
+     * <p>If the node is an invocation of a method that returns its receiver, then its receiver's
+     * type will also be updated. This chain will continue as long as each receiver is itself a
+     * receiver-returning method invocation.
+     *
+     * <p>For example, suppose the argument is the expression {@code a.b().c()}, the new value is
+     * "foo", and b and c return their receiver (and are deterministic). Then all of the expressions
```
mernst, line 49: Why is it important to mention that the method is deterministic?
I think every method that returns its receiver is deterministic, so this would be a vacuous test and it's confusing why it is mentioned as an extra condition.

framework/tests/accumulation/Generics.java
```diff
@@ -0,0 +1,45 @@
+// Copied from the Object Construction Checker.
```
mernst, line 1: Please remove these comments, or indicate how they are actionable.  For example, is the programmer supposed to keep the two versions of the test in sync by making a duplicate pull request on the other repository?

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
```diff
@@ -0,0 +1,140 @@
...
+    public AccumulationTransfer(CFAnalysis analysis) {
+        super(analysis);
+        typeFactory = (AccumulationAnnotatedTypeFactory) analysis.getTypeFactory();
+    }
+
+    /**
+     * Updates the estimate of how many things {@code node} has accumulated.
+     *
+     * <p>If the node is an invocation of a method that returns its receiver, then its receiver's
+     * type will also be updated. This chain will continue as long as each receiver is itself a
```
mernst, line 45: Are there tests of this functionality?  Test `SimpleFluent.java` does not seem to test it.  I expect a test would look like
```
  s.a().b().c();
  s.build();
```

kelloggm:
@mernst I don't entirely understand what some of your requests are asking for. Can you clarify the points I raise below? In the meantime, I will address your direct comments on the code.

>Make a similar change in `getUnionAnno`.

`getUnionAnno` never has access to the list of values in the old annotation; it defers to `greatestLowerBound`, which I think is important for clarity (and I think you suggested in an earlier review). I added a test that avoids creating a new annotation if the list of new values is empty, but I'm not sure what else you're asking for here.

> Rather than using the public `getAccumulatedValues`, can you create a private living-dangerously getter that doesn't do the casts, and catch errors in uses of it? Or can you create a private living-dangerously getter that returns an alias of existing data and therefore should never be side-effected?

I don't understand this request. There are no casts in `getAccumulatedValues`, which primarily defers to `ValueCheckerUtils#getValueOfAnnotationWithStringArgument`. If you think that `getValueOfAnnotationWithStringArgument` can be improved, I think that's a discussion for elsewhere - it is used extensively in the Value Checker and elsewhere.

The only defensive tests that `getAccumulatedValues` does are:
* that the passed annotation is an accumulator annotation, which is a single call to `AnnotationUtils#areSameByClass`, which should be cheap, and
* checks that the result of `getValueOfAnnotationWithStringArgument` is non-null, which is also cheap (it is null if the passed annotation doesn't have a value element for some reason).

`getAccumulatedValues` also does not duplicate existing data structures, but rather returns the result of `getValueOfAnnotationWithStringArgument` directly.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
kelloggm: I think Manu made this point at some point, which is why we added it to the documentation. I agree it is distracting. @msridhar do you think removing any mention of determinism is okay here - it is a general fact about how dataflow works rather than something specific to how accumulation analysis works.

msridhar: Yes we can remove the mention of determinism

kelloggm:
@mernst this is ready for another review

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
```diff
@@ -0,0 +1,146 @@
...
+        super(analysis);
+        typeFactory = (AccumulationAnnotatedTypeFactory) analysis.getTypeFactory();
+    }
+
+    /**
+     * Updates the estimate of how many things {@code node} has accumulated.
+     *
+     * <p>If the node is an invocation of a method that returns its receiver, then its receiver's
+     * type will also be updated. In a chain of method calls, this process will continue as long as
+     * each receiver is itself a receiver-returning method invocation.
```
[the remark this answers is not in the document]
kelloggm: Yes. See ~line 35 of `SimpleFluent.java`, for example.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
```diff
@@ -0,0 +1,146 @@
...
+     * (added by the accumulation analysis because of the {@code .c()} call) is "foo", and b and c
+     * return their receiver. Then all of the expressions {@code a.b().c()}, {@code a.b()}, and
+     * {@code a} would have their estimates updated to include "foo". Note that due to what kind of
+     * values can be held in the store, this information is lost outside the method chain. That is,
+     * the returns-receiver propagated information is lost outside the expression in which the
+     * returns-receiver method invocations are nested.
+     *
+     * <p>As a concrete example, consider the Called Methods accumulation checker: if {@code build}
+     * requires a, b, and c to be called, then {@code foo.a().b().c().build();} will typecheck (they
+     * are in one fluent method chain), but {@code foo.a().b().c(); foo.build();} will not -- the
```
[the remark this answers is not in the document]
kelloggm: The test case exists, in `SimpleFluent.java`. It passes, because the Test Accumulation called-methods checker includes the other rules that allows the code to typecheck. I'll clarify this documentation to explicitly mention the need for another rule from the user if they want to handle this kind of case.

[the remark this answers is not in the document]
mernst: The test case I had in mind contained a non-`@This`-returning method in the middle of a chain of `@This`-returning methods.  That is, there appear `@This`-returning methods both before and after it.  I added such a test case, and it fails.  It is poorly named `m8` and you might want to rename it.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
```diff
@@ -0,0 +1,152 @@
...
+        if (tree.getKind() == Kind.METHOD_INVOCATION) {
+            MethodInvocationNode methodInvocationNode = (MethodInvocationNode) node;
+            while (methodInvocationNode != null) {
+
+                Node receiver = methodInvocationNode.getTarget().getReceiver();
+
+                if (receiver == null || !typeFactory.returnsThis((MethodInvocationTree) tree)) {
+                    break;
+                }
+
```
mernst, line 91: Should the remainder of this loop be equivalent to a recursive call to `accumulate`, just more efficient?
If so, could you add a comment to that effect?  (Or maybe even make it a recursive call to `accumulate`, which will be shorter and simpler code, though a bit less efficient.  If you do that, you can remove the `while` loop.)

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
```diff
@@ -0,0 +1,152 @@
...
+                // Note that this call doesn't do anything if receiver is a method call
+                // that is not deterministic, though the code below can still continue to recurse.
+                // (The code above tested that `tree` returns `this` but did not test `receiver`.)
+                insertIntoStores(result, receiver, newAnno);
```
mernst, line 95: This use of `newAnno` looks wrong to me.  `newAnno` was computed earlier for a later method call, and if the recursion goes past a non-`@This`-returning call, then the value should be reset rather than carried across that call.  I added a failing test case, `m8`, that illustrates incorrect behavior.
Should the recursion go through non-`@This` returning methods at all?

It is also a bit suspicious that no call of `getUnionAnno` occurs.  Could you write a comment saying why it is not needed, and it's OK to do complete replacement rather than GLB?

kelloggm:
@mernst, thanks for the test case. I was able to rewrite `accumulate` to use dataflow directly rather than trying to go through the annotated type factory, which solved the problem when combined with making the `while` loop over receivers into a recursive call, as you suggested. I think the resulting code is cleaner - I did away with `getUnionAnno` entirely.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationTransfer.java
```diff
@@ -0,0 +1,125 @@
...
+     * @param values the new accumulation values
+     */
+    public void accumulate(Node node, TransferResult<CFValue, CFStore> result, String... values) {
+        List<String> valuesAsList = Arrays.asList(values);
+        // If dataflow has already recorded information about the target, fetch it and integrate
+        // it into the list of values in the new annotation.
+        CFValue flowValue = result.getResultValue();
+        if (flowValue != null) {
+            Set<AnnotationMirror> flowAnnos = flowValue.getAnnotations();
+            for (AnnotationMirror anno : flowAnnos) {
```
mernst, line 81: Is this `for` loop guaranteed to execute exactly 0 or 1 times?  Should it assert that about the size of `flowAnnos`?  I think that each one must be either the accumulator or bottom; is that correct?  Is that worth verifying?

kelloggm: > Is this for loop guaranteed to execute exactly 0 or 1 times

Yes. I added an assertion.

> I think that each one must be either the accumulator or bottom; is that correct?

Yes. My original version of this code included a test for it, but it was redundant with checking that the result of the call to `getValueOfAnnotationWithStringArgument` is non-null, so I removed it.

PR #3311, opened 2020-05-15
Remove compiler bootclasspath in CheckerDevelMain (Fixes #3310)

### LINKED ISSUE #3310: Missing specification for a path entry to prepend to the compiler bootclasspath in bin-devel/javac
Running the following command:
` $CHECKER_FRAMEWORK/checker/bin-devel/javac -processor index -implicit:class ArrayLengthLBC.java`
returns the following error:
```
Exception in thread "main" java.lang.AssertionError: CheckerDevelMain.compile.bcp must specify a path entry to prepend to the compiler bootclasspath
        at org.checkerframework.framework.util.CheckerDevelMain.main(CheckerDevelMain.java:49)
```
I think this is due to the deletion of the following line from `checker/bin-devel/javac` when the fix for #3273 was put in:
`     "-DCheckerDevelMain.compile.bcp=${jdkPaths} " \`

MrigankArora:
Fixes #3310

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -400,7 +400,7 @@ task requireJavadoc(type: JavaExec) {
 
 
 /**
- * Creates a task named taskName that runs javadoc.
+ * Creates a task named taskName that runs javadoc with the -Xdoclint:all option.
  *
  * @param taskName the name of the task to create
  * @param taskDescription description of the task
--- a/dataflow/src/main/java/org/checkerframework/dataflow/analysis/Analysis.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/analysis/Analysis.java
@@ -124,7 +124,7 @@ public class Analysis<
         this.currentTree = currentTree;
     }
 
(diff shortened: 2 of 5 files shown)


wmdietl:
Thanks for looking into this issue!
Variable `${jdkPaths}` doesn't seem to be declared, so this will pass an empty string.
I think a proper fix should look at adapting `CheckerDevelMain` and not require the flag at all: https://github.com/typetools/checker-framework/blob/master/framework/src/main/java/org/checkerframework/framework/util/CheckerDevelMain.java#L49

MrigankArora:
@wmdietl I did as you suggested and adapted `CheckerDevelMain` to not require the flag at all. The CI says that there is documentation missing that needs to be added, but everything else seems good to go.

kelloggm:
@wmdietl I think @MrigankArora is asking for our help in documenting `CheckerDevelMain`. I'm also unsure what the right comment here is, so your help would be appreciated.

wmdietl:
@MrigankArora Thanks for the fix! I've added the missing javadoc.

PR #3315, opened 2020-05-17
Fixed "javadoc" for Returns Receiver Checker

nimakarimipour:
This version fixes the warnings with javadoc for Returns Receiver Checker happened for "FluentAPIGenerator" class.


mernst:
Can you explain why this fixes the problem?  The problem is that public Javadoc refers to private implementation details.  Therefore, the links don't work when running `javadoc` without the `-private` command-line option.

mernst:
The problem is that the documentation at https://checkerframework.org/api/org/checkerframework/common/returnsreceiver/FluentAPIGenerator.html does not make sense on its own.  This public documentation makes reference to private fields that are not available to a reader of the documentation.  Please rewrite the documentation of `FluentAPIGenerator` to stand on its own.  The existing documentation could be useful as a code comment.

msridhar:
@mernst yes, @nimakarimipour and I discussed this.  He will rewrite the documentation as you indicate, to not make any reference to internal implementation details.

mernst:
@nimakarimipour Is this ready to merge?
As a general rule, please ping someone when you are done and they should take a look.

nimakarimipour:
> @nimakarimipour Is this ready to merge?
> As a general rule, please ping someone when you are done and they should take a look.

Yes it is ready now, I was waiting for a final review. Sure I will do that from now on.

PR #3317, opened 2020-05-18
Clarify `isSubtype` documentation

mernst:


mernst:
@smillst This makes documentation more explicit and consistent.
Is the added documentation correct?

framework/src/main/java/org/checkerframework/framework/type/QualifierHierarchy.java
```diff
@@ -396,14 +399,19 @@ public static boolean canHaveEmptyAnnotationSet(AnnotatedTypeMirror type) {
...
+     * Tests whether {@code subAnno} is a sub-qualifier of, or equal to, {@code superAnno},
+     * according to the type qualifier hierarchy. This checks only the qualifiers, not the Java
+     * type.
      *
      * <p>This method takes an annotated type to decide if the type variable version of the method
      * should be invoked, or if the normal version is sufficient (which provides more strict
      * checks).
      *
-     * @return true iff {@code subAnno} is a sub qualifier of {@code superAnno}
+     * @param subType the type that might be a subtype
```
smillst, line 410: The AnnotatedTypeMirror parameters are only used to decide if `isSubtypeTypeVariable` should be called instead of `isSubtype(AnnotatedTypeMirror, AnnotatedTypeMirror)`.  This is true for the other AnnotatedTypeMirrors parameters in this method and the method below.

PR #3321, opened 2020-05-22
Address delombok "cannot find symbol" warnings

msridhar:
Without this change, running `./gradlew clean :framework:delombok` leads to the following type of warning output:

```bash
> Task :framework:delombok
[ant:delombok] /home/mernst/research/types/checker-framework-branch-master/framework/tests/returnsreceiverlombok/BuilderTest.java:6: error: package org.checkerframework.common.returnsreceiver.qual does not exist
[ant:delombok] import org.checkerframework.common.returnsreceiver.qual.*;
[ant:delombok] ^
[ant:delombok] /home/mernst/research/types/checker-framework-branch-master/framework/tests/returnsreceiverlombok/BuilderTest.java:31: error: cannot find symbol
[ant:delombok]     BuilderTest.@This BuilderTestBuilder wrapperX() {
[ant:delombok]                  ^
[ant:delombok]   symbol:   class This
[ant:delombok]   location: class CustomBuilderTestBuilder
```

This change adds a Gradle dependence to ensure the relevant annotation is in the classpath.

Note that even with the above output, the build would succeed.  Ideally we would fail the build on such warnings, but I cannot find a setting for Lombok that would achieve this.

/cc @nimakarimipour

### DIFF
--- a/framework/build.gradle
+++ b/framework/build.gradle
@@ -210,9 +210,9 @@ task delombok {
     inputs.files file(srcJava)
     outputs.dir file(srcDelomboked)
 
-    // this dependency is required to ensure the checker-qual jar exists,
+    // This dependency is required to ensure the checker-qual jar exists,
     // to prevent lombok from emitting "cannot find symbol" errors for @This
-    // annotations in the test input code
+    // annotations in the test input code.
     dependsOn project(':checker-qual').tasks.jar
 
     doLast {
--- a/framework/build.gradle
+++ b/framework/build.gradle
@@ -210,10 +210,15 @@ task delombok {
     inputs.files file(srcJava)
     outputs.dir file(srcDelomboked)
 
+    // this dependency is required to ensure the checker-qual jar exists,
+    // to prevent lombok from emitting "cannot find symbol" errors for @This
+    // annotations in the test input code
+    dependsOn project(':checker-qual').tasks.jar
+
     doLast {
         def collection = files(configurations.testCompileClasspath)
         ant.taskdef(name: 'delombok', classname: 'lombok.delombok.ant.Tasks$Delombok',
-                classpath: configurations.testCompileClasspath.asPath)
+                classpath: collection.asPath)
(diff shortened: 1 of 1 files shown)


msridhar:
> I think this pull request should also ensure that any lombok errors (such as "cannot find symbol") should cause the build to fail, rather than being silently ignored as they currently are.

I cannot find any delombok or `lombok.config` setting that will cause the task to fail.  @kelloggm any ideas here?

msridhar:
I attempted to post to the Project Lombok Google Group to ask about making delombok fail on these errors (I don't see the post yet, but maybe it is being moderated).

kelloggm:
We could fail the build ourselves by checking for a line in the stdout that contains both "[ant.delombok]" and "error"

msridhar:
> We could fail the build ourselves by checking for a line in the stdout that contains both "[ant.delombok]" and "error"

Yikes 😄 Personally I don't know how to capture stdout from delombok as run from ant via gradle.  @kelloggm do you know?

msridhar:
CI failures look like network flakiness

kelloggm:
> Yikes 😄 Personally I don't know how to capture stdout from delombok as run from ant via gradle.

No. I suspect it's *possible* if we really want to, but I would wait for a response from the Lombok user group, and if the response is negative then I would recommend that we just merge this as-is. If @mernst thinks it's absolutely necessary to have this cause the build to fail, though, we could spend some time trying to figure out how to do what I suggested.

mernst:
I don't think it's *essential*, but it would be nice if possible.  I would be surprised if delombok doesn't set its return status when there is an error like this -- but if it doesn't, that seems like a bug in delombok.

msridhar:
This is the mailing list post:

https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/project-lombok/eD36Tgbui58/kFJ98YX5AAAJ

No answer yet.  Also I looked at the Delombok code and it wasn't obvious that it is checking for this type of failure.  So this could be a bug or missing capability in Delombok.

PR #3324, opened 2020-05-25
Improvements to CFG marker node text

d367wang:
This pull request mainly changes the marker text of catch block by adding concrete type of exception that the catch block can handle. Take the following example, 

    import java.io.IOException;
    class Exceptions {
        void foo() throws IOException {}
        void bar() {
                try {
                        foo();
                } catch (IOException ioe) {
                        l("IOE");
                } catch (NullPointerException e) {
                        l("NPE");
                } finally {
                        l("finally");
                }
        }
        void l(String p) {}   
  }

Marker texts of the two catch blocks should be updated to "start/end of catch block for **IOException** #hashcode"  and "start/end of catch block for **NullPointerException** #hashcode" separately. 

Additionally, other occurrences of marker texts are complemented to make CFG more readable, such as
- For marker node of `"start of try statement"`, `"start of finally block for Throwable"`, and `"end of finally block for Throwable"`, add `tree.hashCode()` to the messages.
- Add an end marker with message `"end of switch statement #hashcode"` corresponding to `"start of switch statement #hashcode"` marker.

### DIFF
--- a/dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGBuilder.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGBuilder.java
@@ -3426,7 +3426,9 @@ public class CFGBuilder {
 
                 extendWithNode(
                         new MarkerNode(
-                                switchTree, "start of switch statement", env.getTypeUtils()));
+                                switchTree,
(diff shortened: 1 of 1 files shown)


wmdietl:
Thanks for the improvement!
Don't forget to delete the branch from your fork.

PR #3332, opened 2020-05-27
Add Annotation File Utilities as an included build.

smillst:
Before this change it was possible to use dependencies of AFU without specifying them in the build files.


settings.gradle
```diff
@@ -6,4 +6,10 @@ include 'framework'
 include 'checker-qual'
 include 'checker-qual-android'
 include 'framework-test'
-
+includeBuild ('../annotation-tools/annotation-file-utilities'){
+    if(!file('../annotation-tools/annotation-file-utilities').exists()) {
```
smillst, line 10: spacing

framework/build.gradle
```diff
@@ -25,12 +25,24 @@ dependencies {
...
     implementation project(':checker-qual')
-    // As of 2019/12/16, the version of reflection-util in the Annotation
-    // File Utilities takes priority over this version, in the fat jar
-    // file. :-( So update it and re-build it locally when updating this.
     implementation 'org.plumelib:reflection-util:0.2.2'
+    implementation 'org.plumelib:plume-util:1.1.4'
+
+    // TODO: The annotated version of Guava gives an error:
+    // org.checkerframework.annotatedlib/guava/.../guava-28.2-jre.jar(/com/google/common/collect/ImmutableCollection.class):
+    // warning: Cannot find annotation method 'value()' in type 'DoNotMock': class file for com.google.errorprone.annotations.DoNotMock not found
```
mernst, line 39: This seems to be a bug in how the annotated version of Guava was made.  I don't understand it yet.
In the meanwhile, can you try the [workaround used in AFU](https://github.com/typetools/annotation-tools/commit/cad4722172f7fc93436d8af0ba9a6eb9b1ba6cd0), which is to add a dependency on `error_prone_annotations`?

PR #3336, opened 2020-05-28
Support unique classes (Fixes #3313)

### LINKED ISSUE #3313: Feature request: @AlwaysUnique (?) to mark classes whose instances must always be @Unique
**Checker**: [Aliasing Checker](https://checkerframework.org/manual/#aliasing-checker)

**Annotation name**: `@AlwaysUnique` or `@OnlyUnique` or `@Require{d,s,}Unique`

**Annotation target**: class and interface declarations

**Proposed behavior**:

All instances of the class/interface `X` marked (possibly via stubs) as `@AlwaysUnique` must be `@Unique` (i.e. no instance of `@MaybeAliased X` type is ever allowed in the checked code).

**Use cases**:

Certain mutable objects are inherently unsafe when leaked. A good example might be JGit's [`org.eclipse.jgit.revwalk.RevCommit` (as of v5.7.0)](https://download.eclipse.org/jgit/site/5.7.0.202003110725-r/apidocs/org/eclipse/jgit/revwalk/RevCommit.html), which has a mutable internal state. When the instance is leaked, the original holder of the reference may experience an unexpected behavior unless it calls `RevCommit#reset`.

aditya3434:
The Aliasing Checker gives "unique.leaked" for local variables only when they are explicitly annotated. This is because of the calling of hasExplicitAnnotation() type method in AliasingVisitor.java, in the canbeLeaked() method. By annotating class Data with @Unique, the local variables start as @MaybeAliased but eventually get refined to @Unique, due to the upper bound. The class objects are unique but since they are not explicitly annotated, they bypass the canbeLeaked() method and don't report errors. Removing this explicit requirement fixes the issue, and gives the expected output without giving any side effects (under my testing).

This PR should be merged along with the annotated JDK pull request.
(https://github.com/typetools/jdk/pull/57)

Fixes #3313.


framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
```diff
@@ -271,7 +271,13 @@ private boolean canBeLeaked(Tree exp) {
         AnnotatedTypeMirror type = atypeFactory.getAnnotatedType(exp);
         boolean isMethodInvocation = exp.getKind() == Kind.METHOD_INVOCATION;
         boolean isNewClass = exp.getKind() == Kind.NEW_CLASS;
-        return type.hasExplicitAnnotation(Unique.class) && !isMethodInvocation && !isNewClass;
+        boolean isArray = type.getUnderlyingType().getKind().toString().equalsIgnoreCase("array");
```
wmdietl, line 274: You should try to avoid operations on strings. Is there no TypeKind to compare against?

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
```diff
@@ -271,7 +271,13 @@ private boolean canBeLeaked(Tree exp) {
...
+        boolean isNull = exp.getKind() == Kind.NULL_LITERAL;
+        if (type.getUnderlyingType().toString().startsWith("java.lang") || isArray || isNull) {
```
wmdietl, line 276: What is the motivation for this special logic? You should explain the logic here or at least in the PR description.

framework/tests/aliasing/ExplicitAnnotationTest.java
```diff
@@ -0,0 +1,13 @@
...
+
+@Unique class Data {
+    @SuppressWarnings("unique.leaked")
+    Data() {} // All objects of Data are now @Unique
+}
+
+class Demo {
+    void check(Data p) { // p is @Unique Data Object
+        // :: error: (unique.leaked)
+        Data y = p; // @Unique p is leaked
```
wmdietl, line 11: Is the type of the left-hand side important? Can you add another method that does

`Object check2(Data p) { return p; }`

which I guess should also give a `unique.leaked` error.

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
wmdietl: Also, do add tests that cover such special logic, in at least a few important cases.

aditya3434: The main logic I was following that the Aliasing Checker doesn't check for explicit annotations for user defined classes. So only those classes that the user has specifically annotated as @Unique would construct @Unique classes. This would also prevent all trivial objects like strings, int, object, float, etc. from always becoming @Unique (it is more convenient to explicitly annotate them). 

However, I couldn't figure out how to make the AliasingVisitor check whether a class was user-defined or not. So, instead I listed classes and data types for which one would need explicit annotations. These included java.lang classes, arrays and null values. All this complicated conditions could be avoided if the Visitor could check for user-defined classes.

if (annotated class is user defined) {
            return type.hasAnnotation(Unique.class) && !isMethodInvocation && !isNewClass;
        } 
else {
            return type.hasExplicitAnnotation(Unique.class) && !isMethodInvocation && !isNewClass;
        }

wmdietl: Can you clarify what you mean with "user defined"?
Why should the handling of explicit annotations depend on whether the class is "user defined"?

aditya3434: In the stub file stubfile.astub in tests/aliasing, Object and String classes are annotated as @Unique by default. So, if I don't check for explicit annotations for all classes, then, even trivial operations like string assignments give a "unique.leaked" error in the test cases.

// part of TypeRefinementTest.java in tests/aliasing
void rule1() {
        String unique = new String();
        isUnique(unique);
        String notUnique = unique;  // gives unique.leaked error even though @Unique is never used
}

So I decided to limit this checking for "user-defined" classes (classes that are not part of any in-built java library, like java.lang or java.util). For example, in the code in the feature request #3313 (https://github.com/typetools/checker-framework/issues/3313), class Data and Demo are user-defined classes, and since Data is annotated as @Unique, all its instances don't require explicit annotations to be unique as well.

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
wmdietl: Ping.

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
wmdietl: We do not want different behavior for built-in classes and user-defined classes. The same logic should be applied to all source code, regardless of who wrote the code.

There is a difference between classes `String` and `Data` in these two examples.
In `String` the default constructor is annotated as returning a unique object. The class declaration is not annotated. Therefore, it is fine in the test case to have a non-unique String reference.
On the other hand class `Data` in this new test case is annotated as `@Unique` requiring that only `@Unique` references exist.
So your check should be for that difference between the two classes.

Also can you move the annotations from `framework/tests/aliasing/stubfile.astub` into https://github.com/typetools/jdk/pull/52?

framework/tests/aliasing/ExplicitAnnotationTest.java
wmdietl: Ping.

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
aditya3434: I have used your suggested logic in my latest commit. There is no need for array TypeKind comparison in the code.

framework/tests/aliasing/ExplicitAnnotationTest.java
aditya3434: No the type of the left-hand side isn't important. However, the method doesn't give an error, even when Data is explicitly annotated as @Unique.

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
aditya3434: I have added a new method for the above requested logic. Now the code doesn't need to deal with the special cases of arrays, null and java.lang classes. I have used the getDeclAnnotations method to get the set of declared annotations of the class as a whole, and then searched for the unique annotation mirror in the set. The code now gives leaked errors without explicit annotations only when the class is annotated as @Unique. Annotating just the constructor will allow non-unique object references.

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
```diff
@@ -271,7 +265,37 @@ private boolean canBeLeaked(Tree exp) {
...
+    /**
+     * Returns true if class of tree expression {@code exp} has type {@code @Unique}
+     *
+     * @param exp the Tree to check
+     * @return boolean true if class if unique and false otherwise
+     */
+    private boolean isUniqueClass(Tree exp) {
+        AnnotatedTypeMirror type = atypeFactory.getAnnotatedType(exp);
+        Element el = types.asElement(type.getUnderlyingType());
+        if (el != null) {
```
wmdietl, line 285: It might be easier as `if (el == null) { return false; }` and then the same for `annoMirrors`.

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
```diff
@@ -271,7 +265,37 @@ private boolean canBeLeaked(Tree exp) {
...
+            Set<AnnotationMirror> annoMirrors = atypeFactory.getDeclAnnotations(el);
+            if (annoMirrors != null) {
+                for (AnnotationMirror mirror : annoMirrors) {
+                    if (mirror.getAnnotationType()
+                            .asElement()
+                            .getSimpleName()
+                            .contentEquals("Unique")) {
```
wmdietl, line 292: Again, don't use Strings for things like this. Can you use `AnnotationUtils.containsSameByClass` https://github.com/typetools/checker-framework/blob/master/javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java#L230 ? Then you don't need the `for` loop.

framework/tests/aliasing/ExplicitAnnotationTest.java
wmdietl: I'm not sure how you mean this. Does `Object z = p;` not raise an error? Shouldn't it?

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
```diff
@@ -271,7 +265,37 @@ private boolean canBeLeaked(Tree exp) {
         AnnotatedTypeMirror type = atypeFactory.getAnnotatedType(exp);
         boolean isMethodInvocation = exp.getKind() == Kind.METHOD_INVOCATION;
         boolean isNewClass = exp.getKind() == Kind.NEW_CLASS;
-        return type.hasExplicitAnnotation(Unique.class) && !isMethodInvocation && !isNewClass;
+        boolean isUniqueClassFlag = isUniqueClass(exp);
+        if (isUniqueClassFlag) {
+            return type.hasAnnotation(Unique.class) && !isMethodInvocation && !isNewClass;
```
wmdietl, line 270: Why do you need the `type.hasAnnotation` again, if you already know that `isUniqueClass`?

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
```diff
@@ -271,7 +265,37 @@ private boolean canBeLeaked(Tree exp) {
         AnnotatedTypeMirror type = atypeFactory.getAnnotatedType(exp);
         boolean isMethodInvocation = exp.getKind() == Kind.METHOD_INVOCATION;
         boolean isNewClass = exp.getKind() == Kind.NEW_CLASS;
-        return type.hasExplicitAnnotation(Unique.class) && !isMethodInvocation && !isNewClass;
+        boolean isUniqueClassFlag = isUniqueClass(exp);
```
wmdietl, line 268: Remove the `Flag` part. Notice the style for the other `is` variables.

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
```diff
@@ -271,7 +265,37 @@ private boolean canBeLeaked(Tree exp) {
...
+        if (isUniqueClassFlag) {
```
wmdietl, line 269: Could this just be `return !isMethodInvocation && !isNewClass && (isUniqueClass || type.hasExplicitAnnotation(Unique.class));`

wmdietl: Maybe extract the `(isUniqueClass || type.hasExplicitAnnotation(Unique.class)` into a separate local and document why that is the logic.

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
```diff
@@ -271,7 +265,37 @@ private boolean canBeLeaked(Tree exp) {
         AnnotatedTypeMirror type = atypeFactory.getAnnotatedType(exp);
```
wmdietl, line 265: You should also update the javadoc with the change in behavior.
Do you understand what `canBeLeaked` expresses? The documentation is rather unclear.

aditya3434: I have changed the documentation for the canBeLeaked method in my latest commit

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
aditya3434: I have separated the (isUniqueClass || type.hasExplicitAnnotation(Unique.class) into a separate local variable and documented the logic.

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
aditya3434: I have removed the flag part from the variable name in my latest commit

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
aditya3434: I have used the AnnotationUtils.containtsSameByClass method in my latest commit

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
aditya3434: I have separated out el and annoMirrors checks instead of the nested if conditions. If either of them is null, the method returns false

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
```diff
@@ -271,7 +265,37 @@ private boolean canBeLeaked(Tree exp) {
...
+    }
+
+    /**
+     * Returns true if class of tree expression {@code exp} has type {@code @Unique}
+     *
+     * @param exp the Tree to check
+     * @return boolean true if class if unique and false otherwise
+     */
+    private boolean isUniqueClass(Tree exp) {
+        AnnotatedTypeMirror type = atypeFactory.getAnnotatedType(exp);
```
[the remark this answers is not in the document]
aditya3434: I replaced exp with type as an argument in the isUniqueClass in my latest commit

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
```diff
@@ -1,17 +1,11 @@
 package org.checkerframework.common.aliasing;
 
-import com.sun.source.tree.ExpressionTree;
-import com.sun.source.tree.MethodInvocationTree;
-import com.sun.source.tree.MethodTree;
-import com.sun.source.tree.NewArrayTree;
-import com.sun.source.tree.ThrowTree;
-import com.sun.source.tree.Tree;
+import com.sun.source.tree.*;
```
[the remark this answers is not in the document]
aditya3434: Instead of star import, I listed the library data types that were required in my latest commit

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
aditya3434: Yes, I see that was a redundant check. I have not used it in my latest commits

framework/tests/aliasing/ExplicitAnnotationTest.java
aditya3434:
```
@Unique
class Data {

    Object check2(Data p) { return p; }

    void check(Data p) {
        Object z = p;
    }
}
```
In the code given above, Object z = p; does raise an error. Hence, the left hand side isn't important. However, the check2 function (where you are returning a @Unique Data as an Object) doesn't raise an error both in the typetools/master and in my modified code, even if the Data p is explicitly annotated as @Unique Data p.

smillst:
@aditya3434 Is this ready to review again?  If so, please remove your assignment and assign @wmdietl.  Thanks!

aditya3434:
> @aditya3434 Is this ready to review again? If so, please remove your assignment and assign @wmdietl. Thanks!

Thanks for the heads up! I have unassigned myself and assigned @wmdietl for the review.

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
```diff
@@ -263,15 +264,39 @@ protected void checkThisOrSuperConstructorCall(
...
+     * {@code @Unique}
      *
      * @param exp the Tree to check
      */
     private boolean canBeLeaked(Tree exp) {
         AnnotatedTypeMirror type = atypeFactory.getAnnotatedType(exp);
         boolean isMethodInvocation = exp.getKind() == Kind.METHOD_INVOCATION;
         boolean isNewClass = exp.getKind() == Kind.NEW_CLASS;
-        return type.hasExplicitAnnotation(Unique.class) && !isMethodInvocation && !isNewClass;
+        boolean isUniqueType = (isUniqueClass(type) || type.hasExplicitAnnotation(Unique.class));
```
wmdietl, line 277: The outer parenthesis aren't necessary.

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
```diff
@@ -263,15 +264,39 @@ protected void checkThisOrSuperConstructorCall(
...
+        return isUniqueType && !isMethodInvocation && !isNewClass;
+    }
+
+    /**
+     * Returns true if class of annotated type {@code type} has annotation {@code @Unique}
```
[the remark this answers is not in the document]
wmdietl: Maybe use: `Return true if the class declaration for annotated type ...`

framework/tests/aliasing/ExplicitAnnotationTest.java
wmdietl: Ok, can you expand this test case with the `Object` local variable and the expected error.
Then file an issue that shows the problem with the `return` version.

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
```diff
@@ -10,8 +10,8 @@
 import com.sun.source.tree.VariableTree;
 import com.sun.source.util.TreePath;
 import java.util.List;
-import javax.lang.model.element.ExecutableElement;
-import javax.lang.model.element.VariableElement;
+import java.util.Set;
+import javax.lang.model.element.*;
```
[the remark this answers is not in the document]
wmdietl: We had this before: https://github.com/typetools/checker-framework/pull/3336#discussion_r445549913
Look how to change your IDE configuration to not make such changes.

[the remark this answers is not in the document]
aditya3434: I have changed the import statement in my latest commit. Turns out my IDE automatically * imports whenever it imports more than 3 packages.

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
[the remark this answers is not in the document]
aditya3434: I have adjusted the summary in my latest commit

framework/src/main/java/org/checkerframework/common/aliasing/AliasingVisitor.java
aditya3434: I have removed the parenthesis in my latest commit.

aditya3434:
I have added the ```Object``` local variable error in the test case and have filed an issue with the ```return``` version.

Issue Link : https://github.com/typetools/checker-framework/issues/3486

PR #3337, opened 2020-05-28
Aliasing: test constructor declarations

aditya3434:
The Aliasing checker gives a "unique.leaked" error when the super() class of a @Unique annotated class is @MaybeAliased. This is useful for subclasses whose unique objects can leak into parent class methods. Classes that don't inherit from other classes, inherit from Object class as default which is @MaybeAliased according to the Aliasing Checker, hence, their constructors give a redundant "unique.leaked" error. The solution is to annotate the Object class as @Unique in the annotated JDK (https://github.com/typetools/jdk). This would prevent @Unique parent classes to leak their objects in the super() function.

This PR should be merged along with the annotated JDK pull request.
(https://github.com/typetools/jdk/pull/52)


framework/src/main/java/org/checkerframework/common/aliasing/jdk.astub
```diff
@@ -0,0 +1,7 @@
+import org.checkerframework.common.aliasing.qual.*;
+
+package java.lang;
+
+class Object {
+    @Unique Object();
```
mernst, line 6: Why is this annotation being added in a `jdk.astub` file rather than in the annotated JDK, https://github.com/typetools/jdk?

framework/tests/aliasing/UniqueConstructorTest.java
```diff
@@ -0,0 +1,16 @@
+import org.checkerframework.common.aliasing.qual.Unique;
+
+class ParentUniqueClass {
```
wmdietl, line 3: How about adding a top-level class around these two classes that has the same name as the file?

framework/tests/aliasing/UniqueConstructorTest.java
```diff
@@ -0,0 +1,16 @@
...
+
+    // No need for @SuppressWarnings()
+    @Unique ParentUniqueClass() {
+        // Doesn't raise unique.leaked error since it's a parent class (top of heirarchy)
+    }
+}
+
+class ChildUniqueClass extends ParentUniqueClass {
+
+    @Unique ChildUniqueClass() {
```
wmdietl, line 13: Can you add another subclass that does raise an error?

framework/tests/aliasing/UniqueConstructorTest.java
aditya3434: I have added a top-level class around these two classes (with the same name) in my latest commit.

framework/tests/aliasing/UniqueConstructorTest.java
aditya3434: In my latest commit, the ChildUniqueClass will raise an error since the ParentClass is not @Unique.

framework/src/main/java/org/checkerframework/common/aliasing/jdk.astub
wmdietl: You should link related PRs from the PR descriptions, so that it is immediately clear that PRs need to be merged together.

You should remove this jdk.astub file from this PR now, as the changes from the related PR should be picked-up automatically.

Make the one suggested update in the jdk project and let's see whether the build just needs to re-run.

aditya3434: I have removed the jdk.astub file from this PR and provided the link for the PR in description which needs to be merged together.

framework/tests/aliasing/UniqueConstructorTest.java
```diff
@@ -0,0 +1,24 @@
+import org.checkerframework.common.aliasing.qual.Unique;
+
+public class UniqueConstructorTest {
```
wmdietl, line 3: We had discussed earlier having one top-level class. You reverted that and again have multiple top-level classes.

framework/tests/aliasing/UniqueConstructorTest.java
```diff
@@ -0,0 +1,24 @@
...
+    // No need for @SuppressWarnings()
```
wmdietl, line 4: That comment should be combined with the one in the body, where you give the reason why no warning is required.

framework/tests/aliasing/UniqueConstructorTest.java
```diff
@@ -0,0 +1,24 @@
...
+    @Unique UniqueConstructorTest() {
+        // Doesn't raise unique.leaked error since it's a top parent class (@Unique Object subclass
```
wmdietl, line 6: What does `top parent class` mean?

framework/tests/aliasing/UniqueConstructorTest.java
```diff
@@ -0,0 +1,24 @@
...
+        // by default)
+    }
+}
+
+class ParentClass extends UniqueConstructorTest {
+
+    ParentClass() {
+        // Doesn't raise unique.leaked error since it's a non unique class with a unique parent
```
wmdietl, line 14: `non-unique`

framework/tests/aliasing/UniqueConstructorTest.java
aditya3434: Sorry for misunderstanding. I thought that we were discussing about the inheritance hierarchy (Parent extends UniqueConstructorTest and Child extends Parent, making UniqueConstructorTest the "top" level class). I have committed the required changes.

framework/tests/aliasing/UniqueConstructorTest.java
aditya3434: I have combined the comment with the one in the body.

framework/tests/aliasing/UniqueConstructorTest.java
aditya3434: I have made the correction in the comment

framework/tests/aliasing/UniqueConstructorTest.java
aditya3434: The Parent class extends UniqueConstructorTest and the Child class extends the parent class. Hence, I referred to the UniqueConstructorTest class as the "top parent class" (not related to the Parent class in the code) in the inheritance hierarchy. Since, it explicitly doesn't extend any class, it extends java.lang.Object class by default.

wmdietl:
@mernst Do you want to look at this again? We removed the stub file and this is now only a test case.
The failures are all in other type systems, so they must be unrelated to this change.

mernst:
Since https://github.com/typetools/jdk/pull/52 is closed, its branch should be deleted.
You might want to enable this feature on all your forks (I have found it useful and have also enabled it on typetools forks):
https://help.github.com/en/github/administering-a-repository/managing-the-automatic-deletion-of-branches

mernst:
@aditya3434 Ping: Please delete the merged branch https://github.com/aditya3434/jdk/tree/3313-leak-in-constructor-error-fix.  Then re-run CI (for example, by pulling master into this branch).

mernst:
@aditya3434 Third request: Please delete the merged branch https://github.com/aditya3434/jdk/tree/3313-leak-in-constructor-error-fix.  Then re-run CI (for example, by pulling master into this branch).

aditya3434:
> @aditya3434 Third request: Please delete the merged branch https://github.com/aditya3434/jdk/tree/3313-leak-in-constructor-error-fix. Then re-run CI (for example, by pulling master into this branch).

I apologize for the delay. I have deleted the merged jdk branch and re-run CI on the pull request.

smillst:
@wmdietl I think this is ready to merge, but I don't think it fixes #3313. (I think  #3336 fixes that issue?)   Should I unlink that issue?

wmdietl:
@smillst I agree that this didn't fix that issue. I've unliked the issue.

PR #3338, opened 2020-05-28
Don't issue warnings about compiler-inserted modifiers; fixes #3305

### LINKED ISSUE #3305: Can't write annotations in try-with-resources statement
When I declare a variable with an annotation in a try statement, I get an error.
```java

public class C {
  public static void f() {
    try (@Nullable AutoCloseable obj = null) {
    } catch (Exception e) {
    }
  }
}
```
```
C.java:5: warning: [type.anno.before.modifier] write type annotation @Nullable() immediately before type, after modifiers [final]
    try (@Nullable AutoCloseable obj = null) {
                                 ^
```

This seems to be because there's an implicit "final" modifier after the annotation. This modifier doesn't actually exist, so it should be ignored.

mernst:


checker/tests/nullness/TryWithResourcesAnno.java
```diff
@@ -0,0 +1,25 @@
...
+
+    public static void g() {
+        try (@Nullable AutoCloseable obj1 = null;
+                AutoCloseable obj2 = null) {
+        } catch (Exception e) {
+        }
+    }
+
+    public static void h() {
+        try (AutoCloseable obj1 = null;
```
wmdietl, line 20: Do you want to add a test with an expected error, e.g. for `try (@Nullable final AutoCloseable obj = null) {`?

wmdietl: I now see that the implementation wouldn't find such an issue.

I'm wondering a bit why the same problem doesn't exist for implicitly public methods in interfaces...

```java
interface Foo {
    @Nullable public Object get();
}
```

correctly gives a warning, but when I remove the explicit `public` the warning goes away.
Maybe the `public` isn't added to the modifiers in that situation?

Anyways, the shortcoming is documented in the method below, so we don't need a todo test.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1051,13 +1051,25 @@ public Void visitVariable(VariableTree node, Void p) {
...
+        // These two tests could be made more precise and catch some user errors in enum constants
+        // and in try-with-resources declarations, but it doesn't seem worth the effort.
+        if (node.getKind() == Tree.Kind.VARIABLE) {
+            if (TreeUtils.elementFromDeclaration((VariableTree) node).getKind()
+                    == ElementKind.ENUM_CONSTANT) {
+                // Enum constants are "public static final" by default, so the annotation always
+                // appears to be before "public".
+                return;
+            }
+            Tree parent = getCurrentPath().getParentPath().getLeaf();
```
wmdietl, line 1065: I'm always a bit worried about using Paths, because they can be costly.
I think you can simplify this significantly:
- determine the variable kind once `kind = TreeUtils.elementFromDeclaration((VariableTree) node).getKind()`
- if kind is ENUM_CONSTANT it's an enum constant
- if kind is RESOURCE_VARIABLE it's a try resource

So you don't need a path and the two checks are more symmetric.

mernst: Yes, that is better.  I made that change.

PR #3341, opened 2020-05-29
Make special-casing of System.getProperty sound

mernst:
~There are no tests written yet, but~ the code is ready to review.


mernst:
Tests are created and passing now.

checker/src/main/java/org/checkerframework/checker/nullness/SystemGetPropertyHandler.java
```diff
@@ -71,27 +91,78 @@
...
+            String literal = literalFirstArgument(tree);
+            if (literal != null && predefinedSystemProperties.contains(literal)) {
+                AnnotatedTypeMirror type = method.getReturnType();
+                type.replaceAnnotation(factory.NONNULL);
             }
         }
+        if (TreeUtils.isMethodInvocation(tree, systemClearProperty, env)) {
+            String literal = literalFirstArgument(tree);
+            if (literal == null || predefinedSystemProperties.contains(literal)) {
+                checker.reportError(tree, "clear.system.property");
```
smillst, line 161: Errors should be reported in the visitor rather than the annotated type factory.

mernst:
@smillst This pull request has changed enough that I think it needs another review.

PR #3350, opened 2020-06-02
Use Error Prone version 2.4.0

mernst:
@smillst @wmdietl Do you have any interest in reviewing this monstrously big pull request?


framework/build.gradle
```diff
@@ -35,7 +35,7 @@ dependencies {
     implementation 'org.plumelib:plume-util:1.1.4'
 
     // TODO: org.checkerframework.annotatedlib:guava:28.2-jre requires the below dependency.
-    implementation 'com.google.errorprone:error_prone_annotations:2.3.4'
+    implementation 'com.google.errorprone:error_prone_annotations:2.4.0'
```
smillst, line 38: I'm not sure this should change since it is a dependency for Guava.

PR #3352, opened 2020-06-03
Update from typetools 2020-06-03

xingweitian:
Again, one deprecation warning blocks us from building opprop/master. We need a regular update.


xingweitian:
Sorry again...

PR #3361, opened 2020-06-08
Update JAVA_HOME export cmd in docs

leoliu49:
When setting up the system, I noticed an outdated command in the CF manual which didn't determine JAVA_HOME correctly for Ubuntu/Linux.

I updated the one-liner to be the same as in the [`bin-devel` build script](https://github.com/typetools/checker-framework/blob/master/checker/bin-devel/build.sh#L19). I verified that this works on Ubuntu 18.04+.


mernst:
Thanks for noticing and fixing this problem.

PR #3363, opened 2020-06-08
Remove javadoc.astub file; fixes #3357

### LINKED ISSUE #3357: Integrate javadoc.astub with typetools/jdk
https://github.com/typetools/checker-framework/blob/4ecaa4cbe17a14afd3dbcd009b2a442fc263fa65/checker/src/main/resources/javadoc.astub#L1 mentions that

```
// This file, javadoc.astub, models the JavaDoc classes in the JDK 8's
// com.sun.javadoc package, which were renamed/removed in JDK 11 and therefore
// cannot be part of an annotated JDK.
```

I looked at a few of these classes and they are still available:
https://github.com/typetools/jdk/blob/master/src/jdk.javadoc/share/classes/com/sun/javadoc/Doc.java
https://github.com/typetools/jdk/blob/master/src/jdk.javadoc/share/classes/com/sun/javadoc/RootDoc.java
etc.

Some of these classes already contain annotations.
Is there still a reason to keep `javadoc.astub` separate?

mernst:
Merge together with:
https://github.com/plume-lib/options/pull/66
https://github.com/codespecs/daikon/pull/254


checker/src/test/java/tests/NullnessJavadocTest.java
```diff
@@ -0,0 +1,51 @@
...
+                toolsJar(),
+                "-Anomsgtext");
+    }
+
+    /**
+     * Return a list that contains the pathname to the tools.jar file, if it exists.
+     *
+     * @returns a list that contains the pathname to the tools.jar file, or an empty list
+     */
+    private static List<String> toolsJar() {
```
smillst, line 29: Should this be in `org.checkerframework.javacutil.SystemUtil`?

checker/src/test/java/tests/NullnessJavadocTest.java
```diff
@@ -0,0 +1,51 @@
...
+        if (!System.getProperty("java.version").startsWith("1.8")) {
```
smillst, line 30: This should use `org.checkerframework.javacutil.SystemUtil#getJreVersion`.

checker/src/main/resources/javadoc.astub
```diff
@@ -1,127 +0,0 @@
-// This file, javadoc.astub, models the JavaDoc classes in the JDK 8's
```
smillst, line 1: There should be a pull request into typetools/jdk where these annotations are moved, right?

mernst: All the annotations are already in typetools/jdk.

smillst: So this branch, https://github.com/mernst/jdk/tree/remove-javadoc-astub, can be deleted?

PR #3366, opened 2020-06-09
AbstractQualifierPolymorphism: use AnnotationMirrors instead of sets of AnnotationMirrors

smillst:
Sets were used in the past because `@PolyAll` could be instantiated  with multiple qualifiers.


framework/src/main/java/org/checkerframework/framework/type/poly/AbstractQualifierPolymorphism.java
```diff
@@ -235,42 +235,46 @@ public void resolve(
...
+                    result.put(poly, typeQual);
                 }
             }
         }
         return result;
     }
 
     /**
-     * Returns an annotation set that is the merge of the two sets of annotations. The sets are
+     * Returns annotation set that is the combination of the two annotations. The annotations are
```
wmdietl, line 265: The `set` should be removed, right? `Returns the annotation that is....`

framework/src/main/java/org/checkerframework/framework/type/poly/AbstractQualifierPolymorphism.java
```diff
@@ -235,42 +235,46 @@ public void resolve(
...
      *
-     * @param polyQual polymorphic qualifier for which {@code a1Annos} and {@code a2Annos} are
-     *     instantiations
-     * @param a1Annos a set that is an instantiation of {@code polyQual}
-     * @param a2Annos a set that is an instantiation of {@code polyQual}
-     * @return the merge of the two sets
+     * @param polyQual polymorphic qualifier for which {@code a1} and {@code a2} are instantiations
+     * @param a1 an annotation that is an instantiation of {@code polyQual}
+     * @param a2 an annotation that is an instantiation of {@code polyQual}
+     * @return a annotation that is the combination of the two annotations
```
wmdietl, line 274: `an annotation`

PR #3368, opened 2020-06-10
Refactor private method in TypeFromTreeVisitor

smillst:
It was easier to refactor the method than document its previous behavior.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/type/TypeFromTypeTreeVisitor.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/TypeFromTypeTreeVisitor.java
@@ -246,14 +246,15 @@ class TypeFromTypeTreeVisitor extends TypeFromTreeVisitor {
     /**
      * If a tree is can be found for the declaration of the type variable {@code type}, then a
      * {@link AnnotatedTypeVariable} is returned with explicit annotations from the type variables
-     * declared bounds. If a tree cannot be found, then {@code type} is returned.
+     * declared bounds. If a tree cannot be found, then {@code type}, converted to a use, is
+     * returned.
      *
      * @param type type variable used to find declaration tree
      * @param f annotated type factory
      * @return the AnnotatedTypeVariable from the declaration of {@code type} or {@code type} if no
      *     tree is found.
      */
-    private AnnotatedTypeMirror getTypeVariableFromDeclaration(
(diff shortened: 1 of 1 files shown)


framework/src/main/java/org/checkerframework/framework/type/TypeFromTypeTreeVisitor.java
```diff
@@ -243,68 +244,56 @@ public AnnotatedTypeMirror visitWildcard(WildcardTree node, AnnotatedTypeFactory
...
-     * on the type parameter declaration and/or its upper bound.
+     * If a tree is can be found for the declaration of the type variable {@code type}, then a
+     * {@link AnnotatedTypeVariable} is returned with explicit annotations from the type variables
+     * declared bounds. If a tree cannot be found, then {@code type} is returned.
      *
-     * <p>Note for type variable uses in method signatures, explicit annotations on the declaration
-     * are added by {@link TypeFromMemberVisitor#typeVarAnnotator}.
+     * @param type type variable used to find declaration tree
+     * @param f annotated type factory
+     * @return the AnnotatedTypeVariable from the declaration of {@code type} or {@code type} if no
```
wmdietl, line 253: Inlining the call to `asUse()` sounds like a good simplification, but it would be good to document that change in behavior. At the moment it sounds like `type` is returned directly.
Maybe also the method name `getTypeVariable*Use*FromDeclaration` could highlight that.

framework/src/main/java/org/checkerframework/framework/type/TypeFromTypeTreeVisitor.java
```diff
@@ -243,68 +244,56 @@ public AnnotatedTypeMirror visitWildcard(WildcardTree node, AnnotatedTypeFactory
...
+     *     tree is found.
      */
-    private AnnotatedTypeMirror forTypeVariable(AnnotatedTypeMirror type, AnnotatedTypeFactory f) {
-        if (type.getKind() != TypeKind.TYPEVAR) {
-            throw new BugInCF(
-                    "TypeFromTree.forTypeVariable: should only be called on type variables");
-        }
-        TypeVariable typeVar = (TypeVariable) type.getUnderlyingType();
+    private AnnotatedTypeMirror getTypeVariableFromDeclaration(
```
wmdietl, line 256: Could the return type be `AnnotatedTypeVariable`?

PR #3370, opened 2020-06-13
Introduce backward analysis to dataflow framework.

xingweitian:
Refactor dataflow framework. Change `Analysis` to be an interface, and add `AbstractAnalysis`,
`ForwardAnalysis`, `ForwardTransferFunction`, `ForwardAnalysisImpl`, `BackwardAnalysis`, `BackwardTransferFunction`, and `BackwardAnalysisImpl`.

Live variable analysis is also added in #3371 to show that the introduced backward analysis can work.

Co-authored-by: Charles Chen <Charleszhuochen@gmail.com>


dataflow/src/main/java/org/checkerframework/dataflow/analysis/ForwardAnalysisImpl.java
```diff
@@ -0,0 +1,585 @@
...
+
+    // `@code`, not `@link`, because dataflow module doesn't depend on framework module.
+    /**
+     * Construct an object that can perform a org.checkerframework.dataflow forward analysis over a
+     * control flow graph. The transfer function is set by the subclass, e.g., {@code
+     * org.checkerframework.framework.flow.CFAbstractAnalysis}, later.
+     *
+     * @param maxCountBeforeWidening number of times a block can be analyzed before widening
+     */
+    public ForwardAnalysisImpl(int maxCountBeforeWidening) {
```
kelloggm, line 72: Why does the forward analysis contain widening support, but the backwards analysis does not? I know that there is no need for a widening operator in the liveness analysis, but to be generally useful I would expect that there will be times when a widening operator is needed - otherwise a dataflow problem might not terminate. The backwards analysis is not immune to this (as far as I can tell); it must re-add blocks to the worklist just as the forward analysis does. Am I missing something, or is widening support in the backwards analysis future work?

xingweitian:
@kelloggm Thanks for the review and helpful comments!

> Why doesn't the backwards analysis have support for a widening operator? I don't think one is needed to correctly implement live-variable analysis, but many other interesting dataflow problems that one might want to express as backwards analyses will need one.

I agree with you: backward analysis can support widening. In the previous code review of this pull request in opprop, @wmdietl and I decided to support widening for backward analysis in the future: https://github.com/opprop/checker-framework/pull/99#discussion_r439576896

> The live-variable analysis that #3371 implements is never tested, so I don't know how you've established that it is actually working. Can you add some tests that show that the live-variable analysis gets the correct results? That will also help us establish confidence that the implementation of backwards analysis here is correct.

Currently, I use `LiveVariablePlayground` to test the correctness of the live variable analysis and backward analysis: run the playground with a java file to generate a control flow graph. Then, I check the live variable information of each block to see whether it is correct.

I discussed with @wmdietl today, I will also try to create some test cases to test with live variable analysis and backward analysis.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/Analysis.java
```diff
@@ -1,898 +1,116 @@
...
-        Block entry = cfg.getEntryBlock();
-        thenStores.put(entry, initialStore);
-        elseStores.put(entry, initialStore);
-        inputs.put(entry, new TransferInput<>(null, this, initialStore));
-    }
-
-    /**
-     * Add a basic block to the worklist. If {@code b} is already present, the method does nothing.
+     * @param node the node to analyze
+     * @param before the boolean value to indicate which store to return
```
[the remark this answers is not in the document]
xingweitian: > I would also consider changing `before` to an enum, so that callsites say `Store.BEFORE` and `Store.AFTER` rather than just `true` and `false`. Distinguishing between `true` and `false` at callsites makes them harder to read.

This is a separate change, not related to the change of this pull request (introducing backward analysis). I will open an issue for future reference.

kelloggm:
> I agree with you: backward analysis can support widening. In the previous code review of this pull request in opprop, @wmdietl and I decided to support widening for backward analysis in the future: opprop#99 (comment)

Thanks for the context on this. I suspect we will eventually need it, but keeping the first pull request simple is a good idea. Can you document somewhere in the code (probably in the Javadoc for `BackwardsAnalysisImpl`, but possibly elsewhere) that adding widening support in the same way that the forwards analysis supports widening is future work?

> Currently, I use LiveVariablePlayground to test the correctness of the live variable analysis and backward analysis: run the playground with a java file to generate a control flow graph. Then, I check the live variable information of each block to see whether it is correct.
> I discussed with @wmdietl today, I will also try to create some test cases to test with live variable analysis and backward analysis.

Adding automated versions of those tests will really improve these PRs.

xingweitian:
> > I agree with you: backward analysis can support widening. In the previous code review of this pull request in opprop, @wmdietl and I decided to support widening for backward analysis in the future: [opprop#99](https://github.com/opprop/checker-framework/pull/99) (comment)
> 
> Thanks for the context on this. I suspect we will eventually need it, but keeping the first pull request simple is a good idea. Can you document somewhere in the code (probably in the Javadoc for `BackwardsAnalysisImpl`, but possibly elsewhere) that adding widening support in the same way that the forwards analysis supports widening is future work?
> 
> > Currently, I use LiveVariablePlayground to test the correctness of the live variable analysis and backward analysis: run the playground with a java file to generate a control flow graph. Then, I check the live variable information of each block to see whether it is correct.
> > I discussed with @wmdietl today, I will also try to create some test cases to test with live variable analysis and backward analysis.
> 
> Adding automated versions of those tests will really improve these PRs.

Hi @kelloggm, I have added a test case for live variable analysis in #3371, could you please have a look? Thanks!

dataflow/src/main/java/org/checkerframework/dataflow/analysis/AbstractAnalysis.java
```diff
@@ -307,7 +307,9 @@ public Direction getDirection() {
     }
 
     /**
-     * Call the transfer function for node {@code node}, and set that node as current node first.
+     * Call the transfer function for node {@code node}, and set that node as current node first. Be
+     * careful that {@code store} may be shared between nodes in a block. Passing a copied store to
+     * avoid the accident changing.
```
kelloggm, line 312: I don't fully understand this comment. I think it means that `store` should not be side-effected, because it is a copy of a store used elsewhere. If that's correct, can you change the documentation to say it directly.

Also, I find `store` to be a misleading name, because I would expect it to be a `Store`, but it is a `TransferInput`. Should we change the name?

xingweitian: What I intend to say here is that we should remember to pass a copied `store` when calling this method, otherwise the node's transfer input may be changed in the future. Here is an example: https://github.com/typetools/checker-framework/pull/3370/files#diff-9f052b8cabea4817c69534ba908adc75R349. The node in the next iteration maybe change the transfer input of the previous node. Could you please give me some advice?

`store` does not mean `Store` here, one parameter's name of this method is `store`. I can make it clear in Javadoc.

mernst: > store does not mean Store here, one parameter's name of this method is store. I can make it clear in Javadoc.

Martin suggested that you rename the formal parameter.  It is better to give the variable a meaningful name, rather than retain a bad name and add documentation to explain how it is bad.

wmdietl: Most of the code in `AbstractAnalysis` is what previously was in `Analysis`.
https://github.com/typetools/checker-framework/blob/master/dataflow/src/main/java/org/checkerframework/dataflow/analysis/Analysis.java#L427
Instead of just locally renaming this parameter, we can perform a clean-up later and make sure there are no `TransferInput` parameters with name `store` and that also the arguments have appropriate names.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/BackwardAnalysisImpl.java
```diff
@@ -0,0 +1,392 @@
...
+    public S runAnalysisFor(
+            Node node,
+            boolean before,
+            TransferInput<V, S> transferInput,
+            IdentityHashMap<Node, V> nodeValues,
+            Map<TransferInput<V, S>, IdentityHashMap<Node, TransferResult<V, S>>> analysisCaches) {
+        Block block = node.getBlock();
+        assert block != null : "@AssumeAssertion(nullness): invariant";
+        Node oldCurrentNode = currentNode;
+        // TODO: Understand why the Store of passing node is analysis.currentInput.getRegularStore()
```
smillst, line 325: Can this be removed?

xingweitian: Yes, thanks a lot! I have removed it.

PR #3372, opened 2020-06-14
Move Value Checker tests from framework/ to checker/, to use the annotated JDK

mernst:


wmdietl:
Would it be an alternative to move the annotated JDK? As the annotated JDK is now stub files, shouldn't the framework tests be able to use the same files?
We previously needed all checkers to be able to build the binary annotated `jdk8.jar` file, so the annotated JDK files were in `checkers`. We no longer have that constraint, so we should be able to simply clone `typetools/jdk` to a different location.

mernst:
Yes, moving the annotated JDK would also be possible, and it would keep the Value Checker tests in a more logical location.

PR #3373, opened 2020-06-14
Add postAnalyze method

wmdietl:


framework/src/main/java/org/checkerframework/framework/flow/CFAbstractAnalysis.java
```diff
@@ -183,11 +183,6 @@ public TypeHierarchy getTypeHierarchy() {
         return atypeFactory;
     }
 
-    /** Perform a visualization of the CFG and analysis info for inspection. */
```
wmdietl, line 186: Add changelog, possibly also for new method.

msridhar:
@wmdietl thanks a lot for this!  Our motivating use case is to do a CFG traversal and, based on inferred types, possibly report additional errors.  Traditionally I believe errors are reported from inside a checker's visitor.  But here, the APIs seem to be encouraging this type of CFG traversal to be done inside the type factory.  We were thinking having access to the CFG via the `getControlFlowGraph()` method would let us do checks and report errors from within the visitor.  Any thoughts on what is the more idiomatic way to make use of this functionality?

wmdietl:
Usually the recommendation is to not issue errors from the ATFs, because types might get recomputed, which could result in multiple error messages.
Method `postAnalyze` will only be called once per CFG, so it should be fine for you to issue errors in that method.
At the moment the SourceVisitors traverse only the AST and use information prepared by the ATF, so including your analysis over the CFG in the SourceVisitor would be harder.
As the kinds of errors you're trying to raise need CFG information it seems fine to include them in `postAnalyze`.

msridhar:
Sounds good!  It might be good to indicate in the documentation that `postAnalyze` will only be called once, and so it is safe to report errors from there.

msridhar:
As a follow up we realized we have to use `postAnalyze`.  By the time the visitor runs, CFGs for many of the methods in the class may not be readily available from the type factory.  So if we want to check all CFGs, `postAnalyze` seems to be the best way to go.  I don't think we'll be using the new `getControlFlowGraph` method though I see no big harm in adding it.

msridhar:
As an update this change is working great for us.  We are hooking into `postAnalyze()` and running our analysis there.

msridhar:
@wmdietl @smillst any chance this change can land in time to make the next release?

wmdietl:
Yes, I just need to improve the javadoc in a few places. I'll get to it this weekend.

wmdietl:
@msridhar I've tried improving the javadoc. What do you think?
I've also undone the changes to Analysis to not expose the CFG if it isn't necessary. If you have a reference to the Analysis, you should already know what CFG you're operating on.

PR #3379, opened 2020-06-15
pull-pico-change update: receiver sensitive poly resolution asSuper issue

lnsun:
After introducing receiver-sensitive poly resolution (RSPS) and changing the substitution order, the originally fixed subtype relationship between assignment l-value and return value does not hold anymore (e.g. comparing with type variable upper bound). So, compare the subtyping relation before calling asSuper.

TODO: when subtyping relation between assignment l-value and return value is undecidable, i.e. neither subtype nor supertype. Currently skipped RSPS for such case. So far the only known case is when returning type is a type var instantiated with a intersection:
```
interface I {}

class MyClass<T> {
    public T get() {...}
}

class Main {
    public <V extends Enum & I> foo(MyClass<V> myc) {
        I i = myc.get();  // l-value (I) and return value (T extends Object super null) neither subtype nor supertype
    }
}
```


lnsun:
Sorry, wrong repo

PR #3382, opened 2020-06-15
Clone annotated JDK in framework

mernst:
The purpose of this pull request is to permit framework tests (and in particular, Value Checker tests) to use the annotated JDK.


framework/build.gradle
```diff
@@ -1,9 +1,11 @@
+ext {
+    annotatedJdkHome = '../../jdk'
+}
 sourceSets {
     main {
         resources {
             // Stub files, message.properties, etc.
-            srcDirs = ['src/main/java']
-            exclude '**/*.java'
+            srcDirs += ['src/main/java', "${buildDir}/generated/resources"]
```
wmdietl, line 8: Will this now include both the `*.java` source files from `src/main/java` (which we don't want) and the annotated JDK source files (which we want)?
An update to the comment might make this clear.

smillst: No, Gradle excludes *.java files from the source set automatically.

smillst:
@mernst I think this can be merge.  It says two checks where canceled, but it looks they ran and passed:
https://dev.azure.com/checkerframework/checkerframework/_build/results?buildId=1215&view=logs&jobId=452cb463-5e9f-573e-a7d0-380aa0d1b158&j=e7da885e-38cc-54b9-8049-a12030020bfb

mernst:
@smillst You are correct.  I mistakenly canceled the jobs then re-started them, and the display on this webpage is wonky.

PR #3383, opened 2020-06-16
Remove statically-executable.astub

mernst:
~Merge after https://github.com/typetools/checker-framework/pull/3382 (and merge any changes to it into this branch).~

Merge with https://github.com/typetools/jdk/pull/56.


smillst:
The test is failing because of this change: 
https://github.com/typetools/jdk/compare/master...mernst:remove-statically-executable-astub-dont-move-jdk#diff-637753f5148ddd0f1baad0c38ecab33aR160

```
 public static @ArrayLenRange(from = 1) String toString(int i, @Positive @IntRange(from = 2, to = 36) int radix) {
```

The `@IntRange` annotation on radix was added.  I think that annotation is wrong because the Javadoc says that if `radix` is outside the range then 10 is used.

mernst:
@smillst You are right that an invalid radix doesn't throw an error there.  An invalid radix does cause an exception in the equivalent `toString` method of other classes (such as Byte), and does cause an exception in other methods in Integer (such as parsing routines).  An invalid radix still seems like a user error, so I am inclined to leave the annotation and fix the test.

PR #3388, opened 2020-06-17
Clarify and refactor how warnings are suppressed.

### LINKED ISSUE #3356: Flag to ignore suffix-less @SuppressWarnings("all"), or to require both prefix&suffix in suppressed keys
A mechanism to do either of the above would be pretty useful. Certain tools (like Lombok) generate `@SuppressWarnings("all")` on generated methods. (Lombok doesn't even allow to switch this generating this suppression off, unless `delombok` is being run... which in turn has its own host of problems, like completely mangling line numbers in compilation errors/runtime stack traces). This leads to false negatives under certain circumstances, let's e.g consider the following interface:

```java
interface IFoo {
  String getFoo();
}
```

and an implementing class:

```java

class Foo implements IFoo {
  @lombok.Getter
  @Nullable // overridden getter must NOT be nullable since it's declared non-null in IFoo
  private final String foo;
}
```

The Lombok-generated getter will be something like:

```java
  @org.checkerframework.checker.nullness.qual.Nullable
  @SuppressWarnings("all")
  @lombok.Generated
  public String getFoo() {
    return this.foo;
  }
```

**If it wasn't for `@SuppressWarnings("all")`, then `NullnessChecker` would issue an error, which it definitely should in this case**: a method declared as returning non-null in the interface can't be implemented as possibly returning null.

Either a flag to ignore suffix-less `all` key in `@SuppressWarnings`, or a flag similar to `-ArequirePrefixInWarningSuppressions` (maybe `-ArequirePrefixAndSuffixInWarningSuppressions`?) would alleviate the problem.

smillst:
* Use consistent terminology for SuppressWarnings string, prefixes, and message key.  See [SourceChecker#shouldSuppressWarningStringSuppressKey](https://github.com/typetools/checker-framework/pull/3388/files#diff-d6b7792163ff291b1071acc540950c65R2070-R2101) for definitions.  
 for definitions. 
* Use "allcheckers:" instead of "all:" as a prefix in a warning suppression key
* Renamed option -AshowSuppressWarningKeys to -AshowSuppressWarningsStrings 
* Renamed meta-annotation SuppressWarningsKeys to SuppressWarningsPrefix
* Renamed SourceChecker#getSuppressWarningsKeys(...) to getSuppressWarningsPrefixes
* Renamed SubtypingChecker#getSuppressWarningsKeys to getSuppressWarningsPrefixes

Fixes #3356.


framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -2430,19 +2426,6 @@ private void logBugInCF(BugInCF ce) {
         printMessage(msg.toString());
     }
 
-    /**
-     * Side-effects the array to make each string lowercase, then returns the array.
-     *
-     * @param a an array of strings
-     * @return {@code a}, but each element has been lowercased
-     */
-    private static String[] arrayToLowerCase(String[] a) {
```
mernst, line 2439: The use of `Arrays.setAll` and a lambda is shorter than this implementation and could be used here.
My preference would be to retain the `arrayToLowerCase` abstraction, though, because its purpose is obvious from its name, whereas I had to read the `Arrays.setAll` line each time I saw it to understand what it does.  Using the method is also shorter in the client.  Defining the method takes more characters than are saved at the client, but it's more important that clients are short.

Alternately, we could remove the undocumented behavior of lowercasing user-written `@SuppressWarnings` for the user.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -1790,27 +1775,27 @@ protected void warnUnneededSuppressions() {
...
+     * contains a string that would suppress a warning from this checker.
      *
      * @param elementsSuppress elements with a {@code @SuppressWarnings} that actually suppressed a
      *     warning
-     * @param checkerKeys suppress warning keys that suppress any warning from this checker
-     * @param errorKeys error keys that can be issued by this checker
+     * @param prefixes Suppresswarnings prefixes that suppress any warning from this checker
+     * @param allErrorKeys all error keys that can be issued by this checker
      */
     protected void warnUnneededSuppressions(
```
mernst, line 1793: The implementations of
  `SourceChecker.shouldSuppressWarningStringSuppressKey`
  `SourceChecker.warnUnneededSuppressions`
are very similar.
Either abstract out common logic, or eliminate one of these two methods.
(Note that there are some undesirable differences, such as the logic for permitting a user-written messagekey to be a substring of a full Checker Framework messagekey.)

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -2154,24 +2041,115 @@ public boolean shouldSuppressWarnings(@Nullable Element elt, String errKey) {
...
+     * <p>If {@code -ArequirePrefixInWarningSuppressions}, then {@code prefix} and {@code
+     * prefix:partial-message-key} are the only SuppressWarnings strings that have an effect.
+     *
+     * @param suppressWarningStrings the SuppressWarnings strings. May be null, in which case this
+     *     method returns false.
+     * @param messageKey the message key of the error the checker is emitting; a lowercase string,
+     *     without any "checkername:" prefix
+     * @return true if one of the {@code suppressWarningStrings} suppresses the error
+     */
+    private boolean shouldSuppressWarningStringSuppressKey(
```
mernst, line 2101: The implementations of
  `SourceChecker.shouldSuppressWarningStringSuppressKey`
  `SourceChecker.warnUnneededSuppressions`
are very similar.
Either abstract out common logic, or eliminate one of these two methods.
(Note that there are some undesirable differences, such as the logic for permitting a user-written messagekey to be a substring of a full Checker Framework messagekey.)

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -2151,24 +2016,133 @@ public boolean shouldSuppressWarnings(@Nullable Element elt, String errKey) {
...
+     * See {@link #shouldSuppressWarningsStringSuppressKey(String, String...)}
+     *
+     * @param prefixes the SuppressWarnings prefixes used by this checker
+     * @param messageKey the message key of the error the checker is emitting; a lowercase string,
+     *     without any "checkername:" prefix
+     * @param suppressWarningsStrings the SuppressWarnings strings that are in effect. May be null,
+     *     in which case this method returns false.
+     * @return true if one of the {@code suppressWarningsStrings} suppresses the error
+     */
+    private boolean shouldSuppressWarningsStringSuppressKey(
```
mernst, line 2095: How about renaming this to just "shouldSuppress" or to some other shorter name?  I find the current name a bit unwieldy, and it doesn't seem to convey more information than the shorter one.  Also, this method is not really about just keys as the method name implies.
I would rename the other overload as well.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -2151,24 +2016,133 @@ public boolean shouldSuppressWarnings(@Nullable Element elt, String errKey) {
...
+            Set<String> prefixes, String messageKey, String @Nullable ... suppressWarningsStrings) {
```
mernst, line 2096: I don't see any client that makes a varargs call to this method.  How about changing `...` to `[]` (also in the other overload).

After doing that, would it make sense to put the array as the first argument to both variants of the call?  That seems a bit more logical to me, and makes the two overloads a bit more consistent.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -2203,38 +2177,59 @@ private boolean isAnnotatedForThisCheckerOrUpstreamChecker(@Nullable Element elt
...
+     *
+     * @return a sorted set of SuppressWarnings prefixes
      */
-    protected final Collection<String> getStandardSuppressWarningsKeys() {
-        // TreeSet ensures keys are returned in a consistent order.
-        Set<String> result = new TreeSet<>();
-        result.add(SUPPRESS_ALL_KEY);
+    protected final NavigableSet<String> getStandardSuppressWarningsPrefixes() {
+        NavigableSet<String> prefixes = new TreeSet<>();
+        prefixes.add(SUPPRESS_ALL_PREFIX);
```
mernst, line 2201: I wonder if "allcheckers" should be put at the end of the list rather than the beginning, so it is less prominent in the output of `-AshowSuppressWarningStrings`.

smillst: It's a TreeSet, so it is ordered by name.

PR #3395, opened 2020-06-21
Improve CFGVisualizer.

xingweitian:
1. Improve the logic of visualizeBlockNode().
2. As graphviz suggested: We should add `\\l` to the end of each line. Adding to the start of the line will make the logic unclear. Also, two `\\l`s together will create a new line in DOT, which will make the generated graph looks bad. I tried to avoid and remove all the two `\\ls` in this PR.
3. Improve code and logic.
4. Remove extra blanks from the generated graph.


dataflow/src/main/java/org/checkerframework/dataflow/cfg/DOTCFGVisualizer.java
```diff
@@ -284,7 +277,7 @@ public String visualizeStoreKeyVal(String keyName, Object value) {
      * @param str the string to be escaped
      * @return the escaped version of the string
      */
-    private String escapeDoubleQuotes(final String str) {
+    protected String escapeDoubleQuotes(final String str) {
```
wmdietl, line 280: The two escape methods can stay private, right?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
```diff
@@ -261,6 +261,33 @@ protected String loopOverBlockContents(
         }
     }
 
+    /**
+     * Format the given object into a String.
```
wmdietl, line 265: How about `Format the given object as a String suitable for the output format, i.e. with format-specific characters escaped.`

dataflow/src/main/java/org/checkerframework/dataflow/cfg/DOTCFGVisualizer.java
```diff
@@ -206,21 +206,14 @@ protected String dotOutputFileName(UnderlyingAST ast) {
         return outFileName;
     }
 
+    @Override
+    protected String format(Object obj) {
+        return escapeDoubleQuotes(obj);
+    }
+
     @Override
     public String visualizeBlockNode(Node t, @Nullable Analysis<V, S, T> analysis) {
```
wmdietl, line 215: Now that the two `visualizeBlockNode` methods are exactly the same, should we just remove them and make the `Helper` the real implementation in `AbstractCFGVisualizer`?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
```diff
@@ -204,20 +201,17 @@ protected String visualizeBlockHelper(
...
                         sbStore.append(visualizeStore(store));
                     } else {
-                        sbStore.append("null store");
+                        sbStore.append("null store").append(escapeString);
                     }
                     sbBlock.append(sbStore);
                 }
             }
         }
-        if (!centered) {
```
wmdietl, line 218: What happens if analysis is `null`? Before, it looks like an `escapeString` is appended when it is a `SPECIAL_BLOCK`.
Is that missing now? Is that a problem or intended?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/StringCFGVisualizer.java
```diff
@@ -1,10 +1,6 @@
 package org.checkerframework.dataflow.cfg;
 
-import java.util.HashMap;
-import java.util.IdentityHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
wmdietl, line 3: Let's not use wildcard imports.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/StringCFGVisualizer.java
```diff
@@ -25,7 +20,7 @@
     @Override
     public Map<String, Object> visualize(
             ControlFlowGraph cfg, Block entry, @Nullable Analysis<V, S, T> analysis) {
-        String stringGraph = visualizeGraph(cfg, entry, analysis);
+        String stringGraph = visualizeGraph(cfg, entry, analysis).trim();
```
wmdietl, line 23: Would it be hard to track down where the trailing spaces are coming from and make the output proper by construction?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java
xingweitian: It is intended, this logic is no longer needed.

`visualizeSpecialBlockHelper` should be responsible for making special blocks centered. `DOTCFGVisualizer` now passes a separator to `visualizeSpecialBlockHelper`: https://github.com/typetools/checker-framework/pull/3395/files#diff-f708a265ff3b24e68538c35aef4dfb7eR136

Now in this situation, when the analysis is null, a special block in DOT will be like: `[shape=oval label="<entry>\n"]` (already centered).

dataflow/src/main/java/org/checkerframework/dataflow/cfg/StringCFGVisualizer.java
xingweitian: Thanks for catching it! Fixed.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/StringCFGVisualizer.java
xingweitian: I just came up with another way to avoid trailing spaces. Thanks for the remind!

dataflow/src/main/java/org/checkerframework/dataflow/cfg/StringCFGVisualizer.java
```diff
@@ -73,15 +65,25 @@ public String visualizeBlock(Block bb, @Nullable Analysis<V, S, T> analysis) {
 
     @Override
     public String visualizeSpecialBlock(SpecialBlock sbb) {
-        return super.visualizeSpecialBlockHelper(sbb, lineSeparator);
+        switch (sbb.getSpecialType()) {
+            case ENTRY:
+                return "<entry>" + lineSeparator;
```
wmdietl, line 70: Can you add a comment why the lineSeparator is only needed in this case?
Wasn't it nicer before with the `Helper` method that reduced duplication?

xingweitian: I have pushed a better solution... The previous change has been reverted.

PR #3398, opened 2020-06-22
ValueTest mytree.java update

lnsun:
mytree.java contains one false positive and cannot be resolved now. Marked for attention.
Added 2 more test cases to help future dev.

### DIFF
--- a/framework/tests/value/MyTree.java
+++ b/framework/tests/value/MyTree.java
@@ -21,9 +21,16 @@ public class MyTree<Value> {
 
         MyTree<@UnknownVal String> myTree1 = newTree("hello").put("bye");
         // :: error: (assignment.type.incompatible)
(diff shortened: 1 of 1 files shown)


framework/tests/value/MyTree.java
```diff
@@ -21,9 +21,16 @@ void uses() {
 
         MyTree<@UnknownVal String> myTree1 = newTree("hello").put("bye");
         // :: error: (assignment.type.incompatible)
+        MyTree<@StringVal("hello") String> myTree1b = newTree("hello").put("bye");
+
+        // Note: This is a false positive: the type newTree("hello").put("hello") is able to be
```
wmdietl, line 26: How about `Note: This is a false positive: the type of newTree("hello").put("hello") should be inferred as MyTree<@StringVal("hello") String> and the assignment should therefore pass.`

lnsun: Updated. Thanks for the review!

PR #3399, opened 2020-06-23
Don't use static imports.

smillst:


smillst:
This is passing even though GitHub shows it as failing.

PR #3401, opened 2020-06-24
Removed methods and classes marked deprecated in release 3.3.0 or earlier.

smillst:


checker/src/main/java/org/checkerframework/checker/propkey/PropertyKeyAnnotatedTypeFactory.java
```diff
@@ -176,7 +176,7 @@ private static boolean strContains(Set<String> messages, String messageKey) {
                 if (in == null) {
                     checker.message(Kind.WARNING, "Couldn't find the properties file: " + name);
                     // report(null, "propertykeychecker.filenotfound", name);
-                    // return Collections.emptySet();
+                    // return Collections.emptySet()
```
mernst, line 179: What is the purpose of this change?

PR #3404, opened 2020-06-26
Rewrite the implementation of QualifierHierarchy

smillst:
Deprecates `MultiGraphQualifierHierarchy` and `GraphQualifierHierarchy` and replaces them with three new implementations NoElementQualifierHierarchy, ElementQualifierHierarchy, and MostlyNoElementQualifierHierarchy.

This implements a fix that is blocking PR #2970.

Merge with https://github.com/typetools/checker-framework-inference/pull/164.


framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
```diff
@@ -0,0 +1,874 @@
...
+                qualifierKind.superTypes.add(qualifierKind.top);
+            } else {
+                qualifierKind.superTypes = findAllTheSupers(qualifierKind, directSuperMap);
+            }
+        }
+        for (QualifierKind qualifierKind : nameToQualifierKind.values()) {
+            for (QualifierKind top : tops) {
+                if (qualifierKind.isSubtype(top)) {
+                    if (qualifierKind.top == null) {
+                        qualifierKind.top = top;
```
mernst, line 545: As I noted elsewhere, I found it surprising that the `top` field is set for tops elsewhere, and here for all non-tops.  I would put them all in the same place.  Or, document the design.

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
```diff
@@ -0,0 +1,874 @@
...
+     * @throws UserError if there is a cycle in the hierarchy
+     */
+    private Set<QualifierKind> findAllTheSupers(
+            QualifierKind qualifierKind, Map<QualifierKind, Set<QualifierKind>> directSuperMap) {
+        Queue<QualifierKind> queue = new ArrayDeque<>(directSuperMap.get(qualifierKind));
+        Set<QualifierKind> allSupers = new TreeSet<>(directSuperMap.get(qualifierKind));
+        while (!queue.isEmpty()) {
+            QualifierKind superQual = queue.remove();
+            if (superQual == qualifierKind) {
+                throw new UserError("Cycle in hierarchy: %s", qualifierKind);
```
mernst, line 591: This logic does not find all cycles.  For example, suppose there is qualifierKind -> A -> B -> A.  This will infinite-loop because superQual will ping-pong between A and B but never be equal to qualifierKind.

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
```diff
@@ -0,0 +1,874 @@
...
+        Map<QualifierKindPair, QualifierKind> lubs = new TreeMap<>();
+        for (QualifierKind qual1 : nameToQualifierKind.values()) {
+            for (QualifierKind qual2 : nameToQualifierKind.values()) {
+                if (qual1.top != qual2.top) {
+                    continue;
+                }
+                QualifierKind lub = findLub(qual1, qual2);
+                QualifierKindPair pair = new QualifierKindPair(qual1, qual2);
+                QualifierKind otherLub = lubs.get(pair);
+                if (otherLub != null) {
```
mernst, line 615: When reading this, I was confused abouh how this can happen.  I think the answer is that QualifierKindPair canonicalizes the order of the elements of the pair.  With the elimination of that class that I suggested, this will not be relevant any more.

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
```diff
@@ -0,0 +1,874 @@
...
+     * Returns the least upper bound of {@code qual1} and {@code qual2}.
+     *
+     * @param qual1 a qualifier kind
+     * @param qual2 a qualifier kind
+     * @return the least upper bound of {@code qual1} and {@code qual2}
+     */
+    private QualifierKind findLub(QualifierKind qual1, QualifierKind qual2) {
+        if (qual1 == qual2) {
+            return qual1;
+        } else if (qual1.isSubtype(qual2)) {
```
mernst, line 639: This logic is a bit inefficient.  I think that's OK because this version is only ever called once per pair, at the beginning of the Checker Framework run, and the result is cached.  Explaining when various methods will be called would be useful to help readers understand the design.

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
```diff
@@ -0,0 +1,874 @@
...
+        Set<QualifierKind> allSuperTypes = new TreeSet<>(qual1.superTypes);
+        allSuperTypes.retainAll(qual2.superTypes);
+        Set<QualifierKind> lubs = findLowestQualifiers(allSuperTypes);
+        if (lubs.size() != 1) {
+            throw new BugInCF(
+                    "Not exactly 1 lub for %s and %s. Found lubs: [%s].",
+                    qual1, qual2, SystemUtil.join(", ", lubs));
+        }
+        QualifierKind lub = lubs.iterator().next();
+        if (lub.isPoly && !qual1.isPoly && !qual2.isPoly) {
```
mernst, line 653: Why isn't this an error?  (All comments about lub apply to corresponding glb methods too.)

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
```diff
@@ -0,0 +1,876 @@
...
+public class QualifierKindHierarchy {
+
+    /**
+     * Represents a qualifier class. It holds information about the relationship between itself and
+     * other {@link QualifierKind}s.
+     *
+     * <p>Exactly one qualifier kind is created for each annotation class.
+     */
+    // The private non-final fields of this class are set while creating the QualifierKindHierarchy.
+    public static @Interned class QualifierKind implements Comparable<QualifierKind> {
```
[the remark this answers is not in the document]
smillst: No.  Its constructor is private and the Interning Checker, which is run in CI, reports only one error in `#createQualifierKinds`.  See that method for why that warning is suppressed.

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
```diff
@@ -0,0 +1,870 @@
...
+     */
+    protected void specifyBottom(
+            Map<QualifierKind, Set<QualifierKind>> directSuperMap,
+            Class<? extends Annotation> bottom) {
+        if (bottom != null) {
+            String name = bottom.getCanonicalName();
+            QualifierKind bottomKind = getNameToQualifierKind().get(name);
+            Set<QualifierKind> superTypes = directSuperMap.get(bottomKind);
+            superTypes.addAll(directSuperMap.keySet());
+            superTypes.remove(bottomKind);
```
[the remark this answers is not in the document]
smillst: The line above adds the keySet of `directSuperMap` to `superTypes`, so it has to be removed.

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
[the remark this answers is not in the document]
smillst: I had originally named this class "QualifierClass", but it is confusing because a qualifier class would have type `Class<Annotation>`. For example, see the constructor of `QualifierKindHierarchy`.

This class does not represent a "Qualifier" because a qualifier may have values for the annotation elements, but QualifierKinds do not.  So `@Regex(1)` is a qualifier but it can't be represents by this class.

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
smillst: I added logic to break this loop.

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
smillst: They are all set here now.

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
```diff
@@ -0,0 +1,870 @@
...
+                    }
+                    directSupers.add(superQualifier);
+                }
+            }
+        }
+        return directSuperMap;
+    }
+
+    /**
+     * Explicitly set bottom to the given class.
```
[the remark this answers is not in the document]
smillst: Added below.

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
```diff
@@ -0,0 +1,876 @@
...
+import javax.lang.model.element.AnnotationMirror;
+import org.checkerframework.checker.interning.qual.Interned;
+import org.checkerframework.framework.qual.PolymorphicQualifier;
+import org.checkerframework.framework.qual.SubtypeOf;
+import org.checkerframework.javacutil.BugInCF;
+import org.checkerframework.javacutil.SystemUtil;
+import org.checkerframework.javacutil.UserError;
+
+/**
+ * This class holds information about the subtyping relationships between qualifiers.
```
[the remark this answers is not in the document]
smillst: [Just document that there may be more than one hierarchy.]

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
smillst: >Explaining when various methods will be called would be useful to help readers understand the design.

Any method in this class that does anything more than return a value or do a map.get is only called once.  I added 
```
    ///////////////////////////////////////////////////////////////////////////
    /// <editor-fold desc="Initialize hierarchy">
    /// Methods that initialize the hierarchy; only called once.
    ///
...
    // </editor-fold> End of methods that initialize hierarchy
```

Around these methods to make that clear.

framework/src/main/java/org/checkerframework/framework/util/ComplexQualifierHierarchy.java
```diff
@@ -0,0 +1,375 @@
...
+     * @param qual2 QualifierKind for {@code a2}
+     * @return the greatest lower bound between {@code a1} and {@code a2}.
+     */
+    protected abstract AnnotationMirror greatestLowerBound(
+            AnnotationMirror a1, QualifierKind qual1, AnnotationMirror a2, QualifierKind qual2);
+
+    @Override
+    public boolean isSubtypeTypeVariable(AnnotationMirror subAnno, AnnotationMirror superAnno) {
+        if (superAnno == null) {
+            // [] is a supertype of any qualifier, and [] <: []
```
mernst, line 322: I don't understand what "[]" means in this comment (and in other similar comments).

framework/src/main/java/org/checkerframework/framework/util/ComplexQualifierHierarchy.java
```diff
@@ -0,0 +1,375 @@
...
+            Collection<? extends AnnotationMirror> subAnnos,
+            Collection<? extends AnnotationMirror> superAnnos) {
+        int isSubtypeCount = 0;
+        for (AnnotationMirror subAnno : subAnnos) {
+            QualifierKind subKind = getQualifierKind(subAnno);
+            for (AnnotationMirror superAnno : superAnnos) {
+                QualifierKind superKind = getQualifierKind(superAnno);
+                if (subKind.areInSameHierarchy(superKind)) {
+                    if (isSubtype(subAnno, superAnno)) {
+                        isSubtypeCount++;
```
mernst, line 343: It is more efficient to `break;` right after this increment operation.

(Alternately, you could abstract out the inner loop into a separate method, and there would be no need for a count variable.  It might be a bit less efficient because of the method calls, but would be more efficient in being able to exit early rather than checking all type hierarchies.)

Same comment applies in 4 other places in the pull request.

framework/src/main/java/org/checkerframework/framework/util/SimpleQualifierHierarchy.java
```diff
@@ -0,0 +1,287 @@
...
+    @Override
+    public AnnotationMirror getBottomAnnotation(AnnotationMirror start) {
+        QualifierKind kind = getQualifierKind(start);
+        return qualifierMap.get(kind.getBottom());
+    }
+
+    @Override
+    public AnnotationMirror getPolymorphicAnnotation(AnnotationMirror start) {
+        QualifierKind kind = getQualifierKind(start);
+        QualifierKind poly = qualifierKindHierarchy.getTopToPoly().get(kind);
```
mernst, line 169: This code assumes that `kind` is a key in `getTopToPoly()`, but there is no guarantee that `start` is a top.  (The specification of this method in `QualifierHierarchy` does not demand it.)

framework/src/main/java/org/checkerframework/framework/util/ComplexQualifierHierarchy.java
```diff
@@ -0,0 +1,375 @@
...
+    @Override
+    public boolean isSubtypeTypeVariable(
+            Collection<? extends AnnotationMirror> subAnnos,
+            Collection<? extends AnnotationMirror> superAnnos) {
+        int isSubtypeCount = 0;
+        for (AnnotationMirror subAnno : subAnnos) {
+            QualifierKind subKind = getQualifierKind(subAnno);
+            for (AnnotationMirror superAnno : superAnnos) {
+                QualifierKind superKind = getQualifierKind(superAnno);
+                if (subKind.areInSameHierarchy(superKind)) {
```
mernst, line 341: `areInSameHierarchy` reads strangely to me, since the "subject" of the "sentence" is `subKind` which is singular.
How about renaming the method to `inSameHierarchyAs` or `isInSameHierarchyAs` or the like?

framework/src/main/java/org/checkerframework/framework/util/SimpleQualifierHierarchy.java
```diff
@@ -0,0 +1,287 @@
...
+                throw new UserError(
+                        "SimpleQualifierHierarchy cannot be used with annotations that have elements. Found %s: ",
+                        kind);
+            }
+            quals.put(kind, AnnotationBuilder.fromClass(elements, kind.getAnnotationClass()));
+        }
+        return Collections.unmodifiableMap(quals);
+    }
+
+    protected Set<AnnotationMirror> createTops() {
```
mernst, line 96: Please write Javadoc documentation for all methods.
The next two methods need it, too.

framework/src/main/java/org/checkerframework/framework/util/SimpleQualifierHierarchy.java
```diff
@@ -0,0 +1,287 @@
...
+                bottoms.add(entry.getValue());
+            }
+        }
+        return Collections.unmodifiableSet(bottoms);
+    }
+
+    protected QualifierKind getQualifierKind(AnnotationMirror anno) {
+        String name = AnnotationUtils.annotationName(anno);
+        QualifierKind kind = qualifierKindHierarchy.getQualifierKind(name);
+        if (name == null) {
```
mernst, line 119: I think this should be `kind == null`.

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
```diff
@@ -0,0 +1,839 @@
...
+     *
+     * @param directSuperMap a mapping from a {@link QualifierKind} to a set of its direct super
+     *     qualifier kinds; created by {@link #createDirectSuperMap()}
+     * @return the set of bottom {@link QualifierKind}s
+     */
+    // Subclasses should override createDirectSuperMap or specifyBottom to change the bottoms and
+    // not this method, because other methods expect the directSuperMap to be complete.
+    private Set<QualifierKind> createBottomsSet(
+            Map<QualifierKind, Set<QualifierKind>> directSuperMap) {
+        Set<QualifierKind> bottoms = new TreeSet<>(directSuperMap.keySet());
```
mernst, line 538: I wondered why this used a TreeSet.  It's not obviously more efficient.  Is it because you want the method to return the qualifiers in a deterministic order?  The Javadoc doesn't promise it.

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
```diff
@@ -0,0 +1,839 @@
...
+            } else if (!isPoly && !hasSubtypeOfAnno) {
+                throw new UserError(
+                        "AnnotatedTypeFactory: %s does not specify its super qualifiers.%n"
+                                + "Add an @SubtypeOf or @PolymorphicQualifier annotation to it,%n"
+                                + "or if it is an alias, exclude it from `createSupportedTypeQualifiers()`.%n",
+                        qualifierKind);
+            } else if (isPoly) {
+                if (qualifierKind.top == null) {
+                    throw new UserError(
+                            "PolymorphicQualifier, %s,  has to specify type hierarchy, if more than one exist; top types: [%s] ",
```
mernst, line 383: Be specific about the needed fix, similarly to the errors in the previous two cases.

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
```diff
@@ -0,0 +1,839 @@
...
+     * that is done by {@link #initializeQualifierKindFields(Map)}.
+     *
+     * @param qualifierClasses classes of annotations that are type qualifiers
+     * @return a mapping from the canonical name of an annotation class to {@link QualifierKind}
+     */
+    protected NavigableMap<@Interned String, QualifierKind> createQualifierKinds(
+            Collection<Class<? extends Annotation>> qualifierClasses) {
+        TreeMap<@Interned String, QualifierKind> nameToQualifierKind = new TreeMap<>();
+        for (Class<? extends Annotation> clazz : qualifierClasses) {
+            @SuppressWarnings("interning") // uniqueness is tested immediately below
```
mernst, line 412: Why is this required?  Since the QualifierKind class is annotated as `@Interned`, I expect its constructor to return an `@Interned` object.  Am I incorrect about that?

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
```diff
@@ -0,0 +1,839 @@
...
+            @Interned QualifierKind qualifierKind = new QualifierKind(clazz);
+            if (nameToQualifierKind.containsKey(qualifierKind.name)) {
+                throw new UserError("Duplicate QualifierKind " + qualifierKind.name);
```
mernst, line 415: You suggested adding a TypeSystemError to complement BugInCF and UserError.  I think that is a good thing to do, in a separate pull request that is merged before this one.

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
```diff
@@ -0,0 +1,839 @@
...
+            return this.bottom == this;
+        }
+
+        /**
+         * All the qualifier kinds that are a strict super qualifier of this qualifier. Does not
+         * include this qualifier kind itself.
+         *
+         * @return all the qualifier kinds that are a strict super qualifier of this qualifier
+         */
+        public Set<QualifierKind> getSuperTypes() {
```
mernst, line 163: It bugs me a bit that when we speak of subtypes, it is a reflexive operation, but when we speak of supertypes, it is strict (non-reflexive).  This lack of symmetry is confusing.  We could imagine renaming this method to `getStrictSuperTypes`.

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
smillst: The type of the constructor result is `@UnknownInterned`.  See [the manual](https://checkerframework.org/manual/#interning-interned-classes).

framework/src/main/java/org/checkerframework/framework/util/ComplexQualifierHierarchy.java
```diff
@@ -0,0 +1,380 @@
...
+    /**
+     * Returns the qualifier kind for the annotation with the fully qualified name {@code name}.
+     *
+     * @param name fully qualified annotation name
+     * @return the qualifier kind for the annotation with {@code name}
+     */
+    protected QualifierKind getQualifierKind(String name) {
+        QualifierKind kind = qualifierKindHierarchy.getQualifierKind(name);
+        if (kind == null) {
+            throw new BugInCF("QualifierKind not in hierarchy: %s", name);
```
mernst, line 169: Should this be TypeSystemError?  Similar question about some other uses of BugInCF.

smillst: No, the methods in this class are called by the Checker Framework during type checking.  So although these could be a bug in the type system, it is just as likely a bug in the Checker Framework.

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
```diff
@@ -0,0 +1,832 @@
...
+     * #nameToQualifierKind} has been initialized. This method is not static, so it can be
+     * overridden by subclasses.
+     *
+     * @return a mapping from each {@link QualifierKind} to a set of its direct super qualifiers
+     */
+    protected Map<QualifierKind, Set<QualifierKind>> createDirectSuperMap() {
+        Map<QualifierKind, Set<QualifierKind>> directSuperMap = new TreeMap<>();
+        for (QualifierKind qualifierKind : nameToQualifierKind.values()) {
+            SubtypeOf subtypeOfMetaAnno = qualifierKind.clazz.getAnnotation(SubtypeOf.class);
+            if (subtypeOfMetaAnno != null) {
```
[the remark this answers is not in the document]
smillst: > This code treats the absence of @SubtypeOf identically to @SubtypeOf() or @SubtypeOf({})

That's not true.  If an annotation class does not have a `@Subtype`, then it is not added to the directSuperType map.  If a qualifier is not a polymorphic qualifier nor in the directSuperTypes map, then QualifierKindHierarchy#verifyHierarchy throws a `TypeSystemError`.

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
```diff
@@ -0,0 +1,832 @@
...
+    protected Map<QualifierKind, QualifierKind> initializePolymorphicQualifiers() {
+        Map<QualifierKind, QualifierKind> topToPoly = new TreeMap<>();
+        for (QualifierKind qualifierKind : nameToQualifierKind.values()) {
+            Class<? extends Annotation> clazz = qualifierKind.getAnnotationClass();
+            PolymorphicQualifier polyMetaAnno = clazz.getAnnotation(PolymorphicQualifier.class);
+            if (polyMetaAnno == null) {
+                continue;
+            }
+            qualifierKind.isPoly = true;
+            String topName = polyMetaAnno.value().getCanonicalName();
```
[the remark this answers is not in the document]
smillst: The argument to `@PolymorphicQualifier` should always be a top qualifier.  I changed the documentation.  The old implementation assumes it is top. This implementation will throw a TypeSystemError (in `QualifierKindHierarchy#verifyHierarchy`) if it is not top.

framework/src/main/java/org/checkerframework/framework/util/SimpleQualifierHierarchy.java
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    /**
+     * Returns the {@link QualifierKind} for the given annotation.
+     *
+     * @param anno an annotation that is a qualifier in this hierarchy
+     * @return the {@code QualifierKind}
+     * @throws BugInCF if {@code anno} is not a qualifier in this hierarchy
+     */
+    protected QualifierKind getQualifierKind(AnnotationMirror anno) {
```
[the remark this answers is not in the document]
smillst: Yes, it's called often, but most `AnnotationMirrors` are instances of `CheckerFrameworkAnnotationMirror`.  In those cases, `AnnotationUtils.annotationName` returns the `annotationName` field.  If it's not a `CheckerFrameworkAnnotationMirror`, the it's an annotation from bytecode.  In those cases there are many AnnotationMirror objects for the same qualifiers, so a map isn't helpful.

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
```diff
@@ -0,0 +1,839 @@
...
+    public QualifierKind greatestLowerBound(QualifierKind q1, QualifierKind q2) {
+        return glbs.get(q1).get(q2);
+    }
+
+    /**
+     * Returns a collection of all {@link QualifierKind}s sorted in ascending order.
+     *
+     * @return a collection of all {@link QualifierKind}s sorted in ascending order
+     */
+    public Collection<QualifierKind> allQualifierKinds() {
```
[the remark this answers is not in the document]
smillst: Many methods iterate over the collection and throw TypeSystemErrors.  It's sorted so those exceptions are deterministic.

framework/src/main/java/org/checkerframework/framework/util/SimpleQualifierHierarchy.java
```diff
@@ -0,0 +1,304 @@
...
+/**
+ * A qualifier hierarchy where no qualifiers have arguments; that is, no qualifier is represented by
+ * an annotation with elements. The meta-annotation {@link
+ * org.checkerframework.framework.qual.SubtypeOf} specifies the hierarchy.
+ *
+ * <p>It uses a {@link QualifierKindHierarchy} to model the relationships between qualifiers.
+ * Subclasses can override {@link #createQualifierKindHierarchy(Collection)} to return a subclass of
+ * QualifierKindHierarchy.
+ */
+public class SimpleQualifierHierarchy extends QualifierHierarchy {
```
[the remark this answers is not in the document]
smillst: I'm still thinking on this.  
SimpleQualifierHierarchy:
* NoElementsQualifierHierarchy
* QualifierHierarchyWithElements
* DefaultQualifierHierarchy (Because it's the QualifierHierarchy used by default.)

ComplexQualifierHierarchy:
* ElementsQualifierHierarchy
* QualifierHierarchyWithElements
* QualifierWithValuesHierarchy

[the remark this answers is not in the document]
mernst: The suffixed versions QualifierHierarchyWithoutElements and QualifierHierarchyWithElements read best to me, among the options you list.

framework/src/main/java/org/checkerframework/framework/util/ComplexQualifierHierarchy.java
```diff
@@ -0,0 +1,309 @@
...
+        QualifierKind kind = getQualifierKind(start);
+        return bottomsMap.get(kind.getBottom());
+    }
+
+    @Override
+    public boolean isSubtype(AnnotationMirror subAnno, AnnotationMirror superAnno) {
+        QualifierKind subKind = getQualifierKind(subAnno);
+        QualifierKind superKind = getQualifierKind(superAnno);
+        if (subKind.isSubtype(superKind)) {
+            if (superKind.hasElements() && subKind.hasElements()) {
```
mernst, line 229: I expected this test to be `||` rather than `&&`.  There could be hierarchies like

```
    @Top
   /    \
@B(1)   @B(2)
```

or

```
@B(1)    @B(2)
  |        |
 @C       @D
```

framework/src/main/java/org/checkerframework/framework/util/ComplexQualifierHierarchy.java
```diff
@@ -0,0 +1,309 @@
...
+            } else {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
+     * Returns whether or not {@code subAnno} is a subtype of {@code superAnno}. Both {@code
+     * subAnno} and {@code superAnno} are annotations with elements.
```
mernst, line 240: I think "are" should be either "at least one has" or "may be".

framework/src/main/java/org/checkerframework/framework/util/ComplexQualifierHierarchy.java
smillst: I agree that such a hierarchy could exist.  (I don't know of any.)  The current implementation of ComplexQualifierHierarchy doesn't support such a hierarchy.  If we change the implementation, then all the subclasses would have to expect those cases and add boilerplate: 
```
if (!superKind.hasElements() || !subKind.hasElements()) {
    return true;
}
```

I could add another subclass of QualifierHierarchy that allows hierarchies like the one you mentioned.  But what should it be called? I guess in general renaming `ComplexQualifierHierarchy` to `QualifierHierarchyWithElements` isn't exactly correct because it can't represent the all hierarchies with qualifiers that have elements.

I'll continue to think on this.

framework/src/main/java/org/checkerframework/framework/util/ComplexQualifierHierarchy.java
smillst: This is the same question as https://github.com/typetools/checker-framework/pull/3404#discussion_r453075154, so I'm going to "resolve" this comment in favor of that one.

framework/src/main/java/org/checkerframework/framework/util/SimpleQualifierHierarchy.java
```diff
@@ -0,0 +1,239 @@
...
+     * Returns the {@link QualifierKind} for the given annotation.
+     *
+     * @param anno an annotation that is a qualifier in this
+     * @return the {@code QualifierKind}
+     */
+    protected QualifierKind getQualifierKind(AnnotationMirror anno) {
+        String name = AnnotationUtils.annotationName(anno);
+        QualifierKind kind = qualifierKindHierarchy.getQualifierKind(name);
+        if (kind == null) {
+            throw new BugInCF("Annotation not in hierarchy: %s", anno);
```
wmdietl, line 146: Is this a `TypeSystemError`?

checker/src/main/java/org/checkerframework/checker/fenum/FenumAnnotatedTypeFactory.java
```diff
@@ -77,37 +93,78 @@ public FenumAnnotatedTypeFactory(BaseTypeChecker checker) {
...
+                }
+                return FENUM_TOP;
+            } else if (qual1 == FENUM_KIND) {
+                return a1;
+            } else if (qual2 == FENUM_KIND) {
+                return a2;
             }
-            if (AnnotationUtils.areSameByName(subAnno, FENUM)) {
-                subAnno = FENUM;
+            throw new BugInCF("Unexpected QualifierKinds %s %s", qual1, qual2);
```
wmdietl, line 151: Should these now be `TypeSystemErrors`?

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
```diff
@@ -0,0 +1,920 @@
...
+    /**
+     * Returns the canonical name of {@code clazz}. Throws a {@link TypeSystemError} if {@code
+     * clazz} is anonymous or otherwise does not have a name.
+     *
+     * @param clazz annotation class
+     * @return the canonical name of {@code clazz}
+     */
+    public static String annotationClassName(Class<? extends Annotation> clazz) {
+        String name = clazz.getCanonicalName();
+        if (name == null) {
```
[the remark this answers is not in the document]
smillst: I care that name is not null.

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
```diff
@@ -0,0 +1,920 @@
...
+
+    /**
+     * Creates the set of top {@link QualifierKind}s by searching {@code directSuperMap} for
+     * qualifier kinds without any direct super qualifier kinds.
+     *
+     * @param directSuperMap a mapping from a {@link QualifierKind} to a set of its direct super
+     *     qualifier kinds; created by {@link #createDirectSuperMap()}
+     * @return the set of top {@link QualifierKind}s
+     */
+    // Subclasses should override createDirectSuperMap to change the tops and not this method,
```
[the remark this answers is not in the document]
smillst: This comment is to explain why it is private.

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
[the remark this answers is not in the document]
smillst: TODO: Document why it's called kind.

framework/src/main/java/org/checkerframework/framework/util/SimpleQualifierHierarchy.java
smillst: The framework itself could call this method with an incorrect annotation, so it could be a bug in the framework.

framework/src/main/java/org/checkerframework/framework/util/QualifierKindHierarchy.java
```diff
@@ -0,0 +1,920 @@
...
+        return lub;
+    }
+
+    /**
+     * Returns the lowest qualifiers in the passed set.
+     *
+     * @param qualifierKinds a set of qualifiers
+     * @return the lowest qualifiers in the passed set
+     */
+    protected static Set<QualifierKind> findLowestQualifiers(Set<QualifierKind> qualifierKinds) {
```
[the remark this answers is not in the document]
smillst: `findLowestQualifiers` is protected because it is used by the Units Checker.  It's static, so it's not an extension point.

framework/src/main/java/org/checkerframework/framework/type/QualifierHierarchy.java
```diff
@@ -5,25 +5,37 @@
...
  *
- * <p>This assumes that any particular annotated type in a program is annotated with at least one
- * qualifier from the hierarchy.
+ * <p>All type annotations need to be type qualifiers recognized within this hierarchy.
+ *
+ * <p>This assumes that every annotated type in a program is annotated with exactly one qualifier
+ * from each hierarchy.
  */
+@AnnotatedFor("nullness")
 public abstract class QualifierHierarchy {
```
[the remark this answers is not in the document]
smillst: I did this in #3445.

wmdietl:
@smillst Is this one targetted for the August release? Just so I can pick which PRs to look at first.

smillst:
@wmdietl I've marked the pull requests that are targeted for the next release as critical.

checker/src/main/java/org/checkerframework/checker/fenum/FenumAnnotatedTypeFactory.java
```diff
@@ -77,37 +94,78 @@ public FenumAnnotatedTypeFactory(BaseTypeChecker checker) {
...
+                QualifierKind qual1,
+                AnnotationMirror a2,
+                QualifierKind qual2) {
+            if (qual1 == FENUM_KIND && qual2 == FENUM_KIND) {
+                if (AnnotationUtils.areSame(a1, a2)) {
+                    return a1;
+                }
+                return FENUM_TOP;
+            } else if (qual1 == FENUM_KIND) {
+                return a1;
```
wmdietl, line 148: I don't understand this. If a1 is Fenum, but a2 is not Fenum, we could still have that a2 is FenumTop. Then, shouldn't the LUB be FenumTop?

checker/src/main/java/org/checkerframework/checker/fenum/FenumAnnotatedTypeFactory.java
wmdietl: Ping.

checker/src/main/java/org/checkerframework/checker/fenum/FenumAnnotatedTypeFactory.java
smillst: This method is only called when the LUB of the two QualifierKinds is a QualifierKind that has elements.  The `LUB(@Fenum("a"), @FenumTop)` is `@FenumTop`, so this method isn't called. I improved the documentation on the super class for this method.  It wasn't clear enough.

checker/src/main/java/org/checkerframework/checker/fenum/FenumAnnotatedTypeFactory.java
smillst: This could be a bug in QualifierKindHierarchy or ComplexQualifierHierarchy, so it should stay a BugInCF.

framework/src/main/java/org/checkerframework/framework/util/QualifierHierarchyWithoutElements.java
```diff
@@ -0,0 +1,239 @@
...
+        return subKind.isSubtype(superKind);
+    }
+
+    @Override
+    public @Nullable AnnotationMirror leastUpperBound(AnnotationMirror a1, AnnotationMirror a2) {
+        QualifierKind qual1 = getQualifierKind(a1);
+        QualifierKind qual2 = getQualifierKind(a2);
+
+        QualifierKind lub = qualifierKindHierarchy.leastUpperBound(qual1, qual2);
+        if (lub == null) {
```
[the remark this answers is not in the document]
smillst: `qualifierKindHierarchy.leastUpperBound(qual1, qual2)` returns null if the qualifier kinds are not in the same hierarchy. For example, qual1 is `@NonNull` and qual2 is `@Initialized`.  (So the if statement is required.)

framework/src/test/java/testlib/util/FlowTestAnnotatedTypeFactory.java
```diff
@@ -2,37 +2,42 @@
...
 
 public class FlowTestAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
-    protected final AnnotationMirror VALUE, BOTTOM;
+    protected final AnnotationMirror VALUE, BOTTOM, TOP;
 
     public FlowTestAnnotatedTypeFactory(BaseTypeChecker checker) {
         super(checker, true);
         VALUE = AnnotationBuilder.fromClass(elements, Value.class);
         BOTTOM = AnnotationBuilder.fromClass(elements, Bottom.class);
+        TOP = AnnotationBuilder.fromClass(elements, Unqualified.class);
```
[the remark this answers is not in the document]
smillst: That would be better to do in a separate pull request.

framework/src/main/java/org/checkerframework/framework/util/DefaultQualifierKindHierarchy.java
```diff
@@ -0,0 +1,802 @@
...
+        return glbs.get(q1).get(q2);
+    }
+
+    @Override
+    public List<? extends QualifierKind> allQualifierKinds() {
+        return qualifierKinds;
+    }
+
+    @Override
+    public @Nullable QualifierKind getQualifierKind(String name) {
```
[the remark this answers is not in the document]
smillst: No, this method could be used to check if an annotation is a qualifier in this hiearachy.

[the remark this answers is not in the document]
mernst: My comment was a bit unclear, sorry.  I was asking whether it is ever used that way in the current code.  If not, maybe it could throw the exception to simplify all current clients.  (That would require creating a different method if we ever come up with a client that needs to do just a check.)

framework/src/main/java/org/checkerframework/framework/util/QualifierHierarchyWithoutElements.java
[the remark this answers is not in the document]
mernst: I agree that `lub` may be null.
My point is that `kindToAnnotationMirror.get(lub)` evaluates to `null` if `lub` is `null`, so the code will behave the same if this `if` statement is removed.
I'm OK with leaving it in if you prefer it as a matter of style.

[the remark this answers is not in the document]
smillst: > My point is that `kindToAnnotationMirror.get(lub)` evaluates to `null` if `lub` is `null`, so the code will behave the same if this `if` statement is removed.

`kindToAnnotationMirror` will throw a NullPointerException if `get` is called with a `null` argument.

framework/src/main/java/org/checkerframework/framework/util/DefaultQualifierKindHierarchy.java
```diff
@@ -0,0 +1,802 @@
...
+     * @param directSuperMap directSuperMap
+     * @return the set of all qualifier kinds that are a strict supertype of {@code qualifierKind}
+     */
+    private Set<QualifierKind> findAllTheSupers(
+            @UnderInitialization DefaultQualifierKindHierarchy this,
+            @KeyFor("#2") QualifierKind qualifierKind,
+            Map<DefaultQualifierKind, Set<DefaultQualifierKind>> directSuperMap) {
+        Queue<DefaultQualifierKind> queue = new ArrayDeque<>(directSuperMap.get(qualifierKind));
+        Set<QualifierKind> allSupers = new TreeSet<>(directSuperMap.get(qualifierKind));
+        Set<DefaultQualifierKind> visited = new HashSet<>();
```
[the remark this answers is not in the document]
smillst: I don't think so. A `QualifierKind` is added to `visited` once its super qualifier kinds have been added to `allSupers`.  T `allSupers` contains `QualifierKind`s whose super qualifier kinds have not yet been added to allSupers.   But maybe I'm misunderstanding your comment.

checker/src/main/java/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
```diff
@@ -577,30 +578,59 @@ protected boolean hasFieldInvariantAnnotation(
...
+            NULLABLE = getQualifierKind(NullnessAnnotatedTypeFactory.this.NULLABLE);
+        }
+
+        @Override
+        protected boolean isSubtypeWithElements(
+                AnnotationMirror subAnno,
+                QualifierKind subKind,
+                AnnotationMirror superAnno,
+                QualifierKind superKind) {
+            if (!subKind.isInSameHierarchyAs(NULLABLE)
```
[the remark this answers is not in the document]
smillst: `isSubtypeWithElements` is only called when the subKind is a sub qualifier kind of superKind. (I just added that fact to the documentation.)  So the calls never return the different values.

framework/src/main/java/org/checkerframework/framework/util/QualifierHierarchyWithoutElements.java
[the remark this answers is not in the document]
mernst: OK.  Thanks for clarifying my misunderstanding.

framework/src/main/java/org/checkerframework/framework/util/DefaultQualifierKindHierarchy.java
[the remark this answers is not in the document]
mernst: The code as is works, but I was wondering if it could be simplified.  Maybe a comment about the relationship between the variables is enough to help people understand the code.

framework/src/main/java/org/checkerframework/framework/util/QualifierHierarchyMostlyWithoutElements.java
```diff
@@ -0,0 +1,149 @@
...
+import java.util.Collection;
+import javax.lang.model.element.AnnotationMirror;
+import javax.lang.model.util.Elements;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.checkerframework.framework.qual.AnnotatedFor;
+import org.checkerframework.framework.type.QualifierHierarchy;
+
+/**
+ * A {@link QualifierHierarchy} where qualifiers may be represented by annotations with elements,
+ * but most of the qualifiers do not have elements. In contrast to {@link
```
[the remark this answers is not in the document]
smillst: I don't really have any more specific advice.  When I was convert all of the MultiGraphHierarchy subclass to one of the new classes, I only used this class when all but 1 or 2 qualifiers have elements.  Otherwise, it was easier to use QualifierHierarchyWithElements and write out all cases.

framework/src/main/java/org/checkerframework/framework/util/QualifierHierarchyWithElements.java
```diff
@@ -0,0 +1,230 @@
...
+ * org.checkerframework.framework.qual.SubtypeOf} meta-annotation is used.
+ *
+ * <p>QualifierHierarchyWithElements uses a {@link QualifierKindHierarchy} to model the
+ * relationships between qualifiers. Subclasses can override {@link
+ * #createQualifierKindHierarchy(Collection)} to return a subclass of QualifierKindHierarchy.
+ */
+@AnnotatedFor("nullness")
+public abstract class QualifierHierarchyWithElements implements QualifierHierarchy {
+
+    /** {@link org.checkerframework.javacutil.ElementUtils}. */
```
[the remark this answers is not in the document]
smillst: I think it's better to keep them separate.  In the past, we had many bugs where we assumed that qualifiers did not have elements, but they did.  If we have a single abstract class, we might make the same mistakes again.

mernst (approved):
Thanks!

PR #3406, opened 2020-06-27
Issue#1395 : Dump on error fix

### LINKED ISSUE #1395: Implement Dump-on-Error alternative
One useful way of debugging Checker Framework issues was to pass `-doe` to javac, which "dumps-on-errors", meaning one sees a stack trace from where an error is raised.

The recent changes to collect all errors from a compilation unit and report them in `BaseTypeChecker::printCollectedMessages` https://github.com/typetools/checker-framework/commit/0c2c4f822778859f82002a1d5f8fa1e3b90b2e91 makes passing `-doe` useless, as all errors have the same stack trace.

It would be very useful for debugging if the `CheckerMessage` stored the stacktrace of where it was originally created and we either add a new `-AdumpOnErrors` flag (or reuse the existing `-AprintErrorStack` for this purpose) to output this additional information.

aditya3434:
Fix for issue typetools#1395 (dump-on-error). Compound checkers such as nullness and index checkers were not giving adequate stack trace to pin-point where the error in the code had occurred. This was because in BaseTypeChecker, errors due to checkers with subcheckers were stored in CheckerMessage but those without were printed right away giving a correct stack trace. This storage caused the stack trace to unwind and take a different path. Since, all errors were stored in the same place, they gave the same stack trace when the -doe option is used. To fix this problem, I added a CheckerMessage constructor to store the stack trace when the error occurred. Then, I added a method to print this stored stack trace for all the stored errors. The code gives the correct stack trace when used with the '-AdumpOnErrors' compiler option.

Fixes #1395.


docs/manual/creating-a-checker.tex
```diff
@@ -1783,6 +1783,10 @@
 already reported the bug, and you want to continue using the checker on a
 large codebase without being inundated in stack traces.
 
+\item \code{-AdumpOnErrors}: Outputs the stack trace of the program when the
+Checker Framework encounters an error or warning. The option also requires a
+boolean parameter to enables its functionality (e.g. \code{-AdumpOnErrors=true}).
```
wmdietl, line 1788: The boolean shouldn't be required, see the other boolean flags above. Just passing `-AdumOnErrors` should be equal to `-AdumOnErrors=true`. The user could turn off the flag using `-AdumOnErrors=false`.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
```diff
@@ -584,6 +586,22 @@ private void printStoredMessages(CompilationUnitTree unit) {
...
+     * Output the given stack trace if the "dumpOnErrors" option is enabled.
+     *
+     * @param trace stack trace when the checker encountered a warning/error
+     */
+    private void printStackTrace(StackTraceElement[] trace) {
+        boolean dumpOnErrors = getBooleanOption("dumpOnErrors", false);
+        if (dumpOnErrors) {
+            for (StackTraceElement elem : trace) {
+                String msg = "\tat " + elem;
+                message(Diagnostic.Kind.NOTE, msg);
```
wmdietl, line 604: We don't want individual messages for each element. Build up one string, including line breaks, and issue one NOTE.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -313,6 +313,10 @@
     // org.checkerframework.framework.util.typeinference.DefaultTypeArgumentInference
     "showInferenceSteps",
 
+    // Output stack trace when checker encounters errors
+    // org.checkerframework.common.basetype.BaseTypeChecker.printStackTrace()
+    "dumpOnErrors",
```
wmdietl, line 318: The `printOrStoreMessage` in this class should also handle this flag. So you might need to move the `printStackTrace` into this class and make it protected.

wmdietl:
Earlier reviews https://github.com/aditya3434/checker-framework/pull/2

docs/manual/creating-a-checker.tex
```diff
@@ -1783,6 +1783,9 @@
 already reported the bug, and you want to continue using the checker on a
 large codebase without being inundated in stack traces.
 
+\item \code{-AdumpOnErrors}: Outputs the stack trace of the program when the
```
wmdietl, line 1786: I think you can remove the `of the program` and shorten to `output a stack trace when reporting errors or warnings.`

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -315,6 +316,10 @@
     // org.checkerframework.framework.util.typeinference.DefaultTypeArgumentInference
     "showInferenceSteps",
 
+    // Output stack trace when checker encounters errors
```
wmdietl, line 319: Use the short message I proposed for the manual.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
```diff
@@ -619,10 +622,29 @@ private CheckerMessage(
                 String message,
                 @FindDistinct Tree source,
                 @FindDistinct BaseTypeChecker checker) {
+            this(kind, message, source, checker, null);
+        }
+
+        /**
+         * Constructor method for checker message if stack trace needs to be stored.
```
wmdietl, line 629: Don't say `constructor message`. The same `Create a new CheckerMessage.` as for the other constructor is fine.

wmdietl: Actually, there are no calls of the constructor without a trace, so why have two constructors?

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -1079,6 +1084,43 @@ protected void printOrStoreMessage(
...
+     * printOrStoreMessage method with an added stack trace argument. The stack trace is printed
+     * when the dumpOnErrors option is enabled.
+     *
+     * @param kind the kind of message to print
+     * @param message the message text
+     * @param source the souce code position of the diagnostic message
+     * @param root the compilation unit
+     * @param trace the stack trace where the checker encountered an error
+     */
+    protected void printOrStoreMessage(
```
wmdietl, line 1097: I don't see a call of this method in this class when the `dumpOnErrors` flag is provided.
Is the flag handled for a SourceChecker?

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -1079,6 +1084,43 @@ protected void printOrStoreMessage(
...
+            javax.tools.Diagnostic.Kind kind,
+            String message,
+            Tree source,
+            CompilationUnitTree root,
+            StackTraceElement[] trace) {
+        Trees.instance(processingEnv).printMessage(kind, message, source, root);
```
wmdietl, line 1103: Add an implementation comment why here you don't just call the other `printOrStoreMessage` overload.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -1079,6 +1084,43 @@ protected void printOrStoreMessage(
...
+    }
+
+    /**
+     * Output the given stack trace if the "dumpOnErrors" option is enabled.
+     *
+     * @param trace stack trace when the checker encountered a warning/error
+     */
+    private void printStackTrace(StackTraceElement[] trace) {
+        boolean dumpOnErrors =
+                getOptions().containsKey("dumpOnErrors") && getBooleanOption("dumpOnErrors", true);
```
wmdietl, line 1114: Shouldn't this just be `getBooleanOption("dumpOnErrors")`? That method already checks whether the option is given and false should be the default, shouldn't it?

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -1079,6 +1084,43 @@ protected void printOrStoreMessage(
...
+        if (dumpOnErrors) {
+            String msg = new String();
```
wmdietl, line 1116: Don't build up with just a String! In particular that initial `new String()` is very ugly.
Use a StringBuilder or a StringJoiner.

aditya3434: I have used a StringBuilder to build the stack trace and print it as a note in my latest commit.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
aditya3434: I have used the short message in my latest commit.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
aditya3434: I have updated the summary and removed the first redundant constructor from BaseTypeChecker.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
aditya3434: I originally tried ```getBooleanOption("dumpOnErrors", false)```, however it returns false whenever no boolean argument is given. ```getBooleanOption("dumpOnErrors", true)``` works for the most part but returns true even if the ```dumpOnErrors``` option is not provided. To prevent that, I used the ```getOptions().containsKey("dumpOnErrors")``` to check whether the dumpOnErrors option is given.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
aditya3434: We can alternatively call the other ```printOrStoreMessage``` and it wouldn't affect the output. So I don't have a particular reason for not calling the function. If you think I should call the function, I can make a commit accordingly.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
aditya3434: The ```dumpOnErrors``` flag stores the stack trace when the Checker Framework starts to print/store the error messages. The stack trace is stored in BaseTypeChecker, because after the error messages are stored, the stack trace starts to unwind for compound checkers. This is why the printOrStoreMessage is not seen in the stack trace because it wasn't part of it when the stack trace was stored. Hence, the functions calls that print and log the errors are not visible in the stack trace, but the user can observe where the error occurred.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
```diff
@@ -599,6 +600,8 @@ private void printStoredMessages(CompilationUnitTree unit) {
         final String message;
         /** The source code that the message is about. */
         final @InternedDistinct Tree source;
+        /** Stores the stack trace till the point the checker encounters an error. */
+        final StackTraceElement @Nullable [] trace;
```
wmdietl, line 604: Why is the field `@Nullable`? Same for the `trace` parameter that sets the field.
`Thread.currentThread().getStackTrace()` returns something non-null.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
wmdietl: This is a new message in `SourceChecker`, but I don't see a call to it in `SourceChecker`.
There should be a call in `SourceChecker` as well, in case somebody directly uses/extends it.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
wmdietl: I don't think that would give the same behavior, because `BaseTypeChecker` overrides that method and you would get two stack traces, wouldn't you?

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
wmdietl: That's a bit odd, looking at https://github.com/typetools/checker-framework/blob/34989702ef50f1ae3939d7acda3ad3b8dabc56e1/framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java#L1589 I wouldn't expect that behavior.
Can you investigate why this doesn't work as specified?

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
aditya3434: I think the IDE added it automatically when I initialized it. I have removed the annotation since it was, as you said, redundant.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
aditya3434: I called the other function in the ```printOrStoreMessage``` method. It printed the stack trace once but didn't print the error itself. I have added an implementation comment about this behaviour in my latest commit.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
aditya3434: If we write ```dumpOnErrors``` without a boolean argument, the ```value``` variable in the ```getBooleanOption()``` method would be null. Hence, the function would return the default value which, if we used ```getBooleanOption("dumpOnErrors", false)``` would be false. So the stack trace won't be printed if we don't give the boolean value ```true```. If we use ```getBooleanOption("dumpOnErrors", true)```, then we don't need a boolean argument to print the stack trace, however, then it would print the stack trace even if ```dumpOnErrors``` is not provided because ```value``` will be null, and it would return the default value which is ```true```.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
aditya3434: I have added the method call in this function so the stack trace will be printed if someone directly calls it.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
```diff
@@ -599,6 +600,8 @@ private void printStoredMessages(CompilationUnitTree unit) {
         final String message;
         /** The source code that the message is about. */
         final @InternedDistinct Tree source;
+        /** Stores the stack trace till the point the checker encounters an error. */
```
wmdietl, line 603: This is the stack trace when it encountered an error, so this should be something like `The stack trace when the message is created.`
Also change the parameter documentation to the same.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -1081,7 +1086,47 @@ protected void printOrStoreMessage(
             String message,
             Tree source,
             CompilationUnitTree root) {
+        StackTraceElement[] trace = Thread.currentThread().getStackTrace();
+        printOrStoreMessage(kind, message, source, root, trace);
+    }
+
+    /**
+     * printOrStoreMessage method with an added stack trace argument. The stack trace is printed
```
wmdietl, line 1094: Like for other overloaded methods, start out with the normal description of what the method is for and then describe the differences.
Also, `@see` cross-references can be useful.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -1081,7 +1086,47 @@ protected void printOrStoreMessage(
...
+     * @param root the compilation unit
+     * @param trace the stack trace where the checker encountered an error
+     */
+    protected void printOrStoreMessage(
+            javax.tools.Diagnostic.Kind kind,
+            String message,
+            Tree source,
+            CompilationUnitTree root,
+            StackTraceElement[] trace) {
+        // Not calling printOrStoreMessage(kind, message, source, root) since the error message
```
wmdietl, line 1109: That comment is out-of-date - you would now create an infinite loop, so this question no longer arises.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
wmdietl: Is simply `hasOption("dumpOnErrors")` enough? Then you don't need to introduce a local variable for this.
In a separate PR, can you look into what `getBooleanOption` is doing wrong? It is hardly used, so we should discuss what the correct behavior is.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
aditya3434: I have changed the field and parameter description in my latest commit.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
aditya3434: I have updated the documentation in my latest commit.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
aditya3434: I have deleted the comment in my latest commit.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
aditya3434: ```hasOption("dumpOnErrors")``` is the same as ```getOptions().containtsKey("dumpOnErrors")```. If we give the option ```-AdumpOnErrors=false```, ```hasOption("dumpOnErrors")``` will return true. Hence, we cannot use this directly. I have replaced ```getOptions().containtsKey("dumpOnErrors")``` with ```hasOption("dumpOnErrors")```.

PR #3409, opened 2020-06-29
Add `isPolymorphicQualifier` method to QualifierHierarchy class.

smillst:
* Add `isPolymorphicQualifier` method to QualifierHierarchy class. 
* Use `java.lang.annotation.Annotation.class` as the default value for the meta-annotation PolymorphicQualifier.  This makes the code less confusing.
* Move static package private methods closer to uses.


smillst:
@wmdietl FYI, this need not be in the July release.

framework/src/main/java/org/checkerframework/framework/type/QualifierHierarchy.java
```diff
@@ -73,6 +73,16 @@ public int getWidth() {
      */
     public abstract AnnotationMirror getPolymorphicAnnotation(AnnotationMirror start);
 
+    /**
+     * Returns {@code true} is the qualifier is a polymorphic qualifier; otherwise, returns {@code
+     * false}.
+     *
+     * @param qualifier qualifier
+     * @return {@code true} is the qualifier is a polymorphic qualifier; otherwise, returns {@code
```
wmdietl, line 81: First `is` should be `if`.

framework/src/main/java/org/checkerframework/framework/util/MultiGraphQualifierHierarchy.java
```diff
@@ -91,16 +92,71 @@ public void addQualifier(AnnotationMirror qual) {
...
             if (pqtopclass != null) {
-                AnnotationMirror pqtop =
-                        AnnotationBuilder.fromName(atypeFactory.getElementUtils(), pqtopclass);
-                // use given top (which might be PolymorphicQualifier) as key
+                AnnotationMirror pqtop;
+                if (pqtopclass.contentEquals(Annotation.class.getName())) {
+                    // @PolymorphicQualifier with no value defaults to Annotation.class.
+                    // That means there is only one top in the hierarchy. The top qualifier
+                    // may not be known at this point, so use the qualifier itself.
+                    // This changed to top in MultiGraphQualifierHierarchy.addPolyRelations
```
wmdietl, line 102: The `This changed` isn't clear. Maybe `It will be changed to...`? Or something else?

framework/src/main/java/org/checkerframework/framework/util/MultiGraphQualifierHierarchy.java
```diff
@@ -91,16 +92,71 @@ public void addQualifier(AnnotationMirror qual) {
...
+                } else {
+                    pqtop = AnnotationBuilder.fromName(atypeFactory.getElementUtils(), pqtopclass);
+                }
+                // use given top (which might be Annotation) as key
                 this.polyQualifiers.put(pqtop, qual);
             } else {
                 supertypesDirect.put(qual, AnnotationUtils.createAnnotationSet());
             }
         }
+        /**
```
wmdietl, line 113: Spacing.

framework/src/main/java/org/checkerframework/framework/util/MultiGraphQualifierHierarchy.java
```diff
@@ -564,77 +608,78 @@ protected void addPolyRelations(
             return;
         }
 
+        // Handle the case where @PolymorphicQualifier uses the default value Annotation.class.
+        if (polyQualifiers.size() == 1 && tops.size() == 1) {
+            AnnotationMirror poly = polyQualifiers.keySet().iterator().next();
```
wmdietl, line 613: I would probably use `entrySet` and then use the key for `poly` and the value for `maybeTop`.

framework/src/main/java/org/checkerframework/framework/qual/PolymorphicQualifier.java
```diff
@@ -25,8 +25,10 @@
      * When multiple type hierarchies are supported by a single type system, then each polymorphic
      * qualifier needs to indicate which sub-hierarchy it belongs to. Do so by passing a qualifier
      * from the given hierarchy, by convention the top qualifier.
+     *
+     * @return the top qualifier in the hierarchy of this qualifier
      */
-    // We use the meaningless PolymorphicQualifier.class as default value and
+    // We use the meaningless Annotation.class as default value and
```
wmdietl, line 31: Can you explain why using `Annotation.class` makes this clearer? Using `PolymorphicQualifier.class` makes it a bit clearer to me that this is about polymorphism... using just `Annotation.class` doesn't give any hint. Checking for either one shouldn't make a difference, does it? If it does, can you improve the comment?

smillst: It doesn't matter what the default value is to the user.  The implementation is confusing because it is expecting the top qualifier, but PolymorphicQualifier is a meta-annotation.  It's also confusing to get the PolymorphicQualifier meta-annotation and then look for PolymorphicQualifier again.

PR #3411, opened 2020-06-30
Improve Azure instructions in developer manual

jwaataja:
There seems to be an extra step in creating an Azure pipeline not in the developer manual. I added instructions for this step (I'm not sure exactly sure if they're correct or not).


mernst:
Thanks for fixing this oversight.

PR #3414, opened 2020-07-01
Correct the logic of dataflow framework.

### LINKED ISSUE #3447: NullPointerException in backward analysis.
Get `NullPointerException` when executing backward analysis:

```
> Task :dataflow:LiveVariablePlayground.main() FAILED
	at org.checkerframework.dataflow.analysis.BackwardAnalysisImpl.performAnalysisBlock(BackwardAnalysisImpl.java:134)
	at org.checkerframework.dataflow.analysis.BackwardAnalysisImpl.performAnalysis(BackwardAnalysisImpl.java:86)
	at org.checkerframework.dataflow.cfg.CFGVisualizeLauncher.generateDOTofCFG(CFGVisualizeLauncher.java:145)
	at org.checkerframework.dataflow.cfg.playground.LiveVariablePlayground.main(LiveVariablePlayground.java:31)

Execution failed for task ':dataflow:LiveVariablePlayground.main()'.
> Process 'command '/opt/jdk/jdk-11/bin/java'' finished with non-zero exit value 1
```

A minimal test case:

```java
public class Test {
    public void test() throws Exception {
        try {
            int[] myNumbers = {1};
            System.out.println(myNumbers[1]);
        } catch (Exception e) {}
    }
}
```

This issue will be fixed in #3414.

xingweitian:
BackwardAnalysis:
1. Add exception block's predecessor exception block to inputs (fix null pointer exception).
2. When running `runAnalysisFor()` with an exception block, pass a copied transfer input to `callTransferFunction()`.

ForwardAnalysis:
1. When running `runAnalysisFor()` with an exception block, pass a copied transfer input to `callTransferFunction()`. In the previous dataflow framework (before #3370), we passed a non-copied transfer input (https://github.com/typetools/checker-framework/pull/3370/files#diff-d4108228a0a5407e00c090c2e97d8b6dL401-L402). This seems a bug existing in the old dataflow framework.

Resolves #3447.


wmdietl:
This can wait until after the release.

xingweitian:
Sorry, this is closed by accident.

dataflow/build.gradle
```diff
@@ -59,3 +59,25 @@ task liveVariableTest(dependsOn: compileTestJava, group: 'Verification') {
         }
     }
 }
+
+task issue3447Test(dependsOn: compileTestJava, group: 'Verification') {
```
xingweitian, line 63: This is ugly... Is there a better way?

dataflow/build.gradle
```diff
@@ -59,3 +59,25 @@ task liveVariableTest(dependsOn: compileTestJava, group: 'Verification') {
...
+    description 'Test issue 3447 test case for backward analysis.'
+    inputs.file('tests/issue3447/Test.java')
+
+    outputs.file('tests/live-variable/Out.txt')
```
wmdietl, line 67: Should these two lines use `issue3447`?

dataflow/tests/issue3447/Test.java
```diff
@@ -0,0 +1,12 @@
+// Test case for Issue 3447:
+// https://github.com/typetools/checker-framework/issues/3447
+
+public class Test {
```
wmdietl, line 4: I would name the file and class Issue3447.

dataflow/build.gradle
```diff
@@ -59,3 +59,25 @@ task liveVariableTest(dependsOn: compileTestJava, group: 'Verification') {
...
+    outputs.file('tests/live-variable/Test.class')
+
+    delete('tests/issue3447/Out.txt')
```
wmdietl, line 70: Does the test even produce an `Out.txt` file?

dataflow/build.gradle
xingweitian: Thanks! I have fixed that.

dataflow/build.gradle
xingweitian: It does. This test relies on [dataflow/src/test/java/livevar/LiveVariable.java](https://github.com/typetools/checker-framework/blob/master/dataflow/src/test/java/livevar/LiveVariable.java).

dataflow/tests/issue3447/Test.java
xingweitian: It seems that I cannot change the file name or class name from `Test` to `Issue3447`. It is hardcoded at [LiveVariable.java#L23](https://github.com/typetools/checker-framework/blob/master/dataflow/src/test/java/livevar/LiveVariable.java#L23).

wmdietl: Ah, ok.

dataflow/build.gradle
wmdietl: You're just not comparing against a golden file, but the test suite still generates the Out.txt file. Sounds good.

dataflow/build.gradle
smillst: What is ugly?  The code duplication? There may be a better way, but let's worry about that when we have a third test case.

PR #3418, opened 2020-07-02
Add TypeSystemError exception.

smillst:
It's like UseError, but indicates a mistake made by a type system writer.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
+++ b/framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
@@ -2332,7 +2332,7 @@ public abstract class SourceChecker extends AbstractTypeProcessor
     /**
      * Log (that is, print) a type system error.
      *
-     * @param ce the user error to output
+     * @param ce the type system error to output
      */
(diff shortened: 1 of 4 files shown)


framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -2324,6 +2329,16 @@ private void logUserError(UserError ce) {
         printMessage(msg);
     }
 
+    /**
+     * Log (that is, print) a type system error.
+     *
+     * @param ce the user error to output
```
mernst, line 2335: "user" => "type system"

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -490,7 +491,7 @@ private void checkSupportedQuals() {
         if (supportedQuals.isEmpty()) {
             // This is throwing a CF bug, but it could also be a bug in the checker rather than in
```
mernst, line 492: Comment is no longer necessary.

PR #3430, opened 2020-07-05
License clarification; fixes #3429

### LINKED ISSUE #3429: Please consider providing expanded author information
Downstream distributors (such as Debian) like to provide detailed information about the origin of software packages in order to recognize the work of people who contribute to the overall system. The only attribution in Checker Framework is currently "Checker Framework developers" in [LICENSE.txt](https://github.com/typetools/checker-framework/blob/b323ab2d3dfcd0a3699ce61ad2c7f511db801afe/LICENSE.txt#L2). I see that there are additional names listed in the [POM files](https://github.com/typetools/checker-framework/blob/b323ab2d3dfcd0a3699ce61ad2c7f511db801afe/docs/developer/release/maven-artifacts/checker-qual-pom.xml#L34) but it is unclear if those are all-inclusive or merely the lead developers. 

If possible, if would be great to have an AUTHORS.txt file or similar to recognize all contributors.

mernst:


LICENSE.txt
```diff
@@ -6,34 +6,39 @@ Most of the Checker Framework is licensed under the GNU General Public
...
-   NullnessUtil.java, RegexUtil.java, SignednessUtil.java, etc.  In
-   addition, the cleanroom implementations of third-party annotations,
-   which the Checker Framework recognizes as aliases for its own
-   annotations, are licensed under the MIT License.
-
-Some external libraries that are included with the Checker Framework have
-different licenses.
+   (The text of this license also appears below.)  This applies to the
+   checker-qual*.jar and all the files that appear in it:  every file in a
+   qual/ directory, plus utility files including FormatUtil.java,
```
olekw, line 15: I recommend not using the word "including" because that implies that there are additional files that meet this criterion that are not listed. Just dropping that one word will make this perfect.

LICENSE.txt
```diff
@@ -6,34 +6,39 @@ Most of the Checker Framework is licensed under the GNU General Public
...
+   I18nFormatUtil.java, NullnessUtil.java, Opt.java,
+   PurityUnqualified.java, RegexUtil.java, SignednessUtil.java, and
+   UnitsTools.java.  It also applies to the cleanroom implementations of
+   third-party annotations (e.g., in checker/src/testannotations/ and in
```
olekw, line 20: I recommend not using "e.g." because, as above, it implies that additional files fall into this category. Again, just dropping that will make this sentence perfect. (Assuming that those are the only two instances.)

LICENSE.txt
```diff
@@ -6,34 +6,39 @@ Most of the Checker Framework is licensed under the GNU General Public
...
+   framework/src/main/java/org/jmlspecs/).
+
+The Checker Framework includes annotations for some libraries (e.g., in
+.astub files and in https://github.com/typetools/jdk, which appears in the
+annotated-jdk directory of file checker.jar).  Each annotated library uses
+the same license as the original upstream version of the library.
```
olekw, line 26: This is the only problematic part. I count 55 .astub files in this repository. None of them (of the random sampling I looked at) contains any license information. If they are under a different license than the main source code then that needs to be explicitly stated to allow legally-safe use of this library. Putting a LICENSE file in each directory would be fairly arduous but a summary in this file should suffice. e.g. 
```suggestion
annotated-jdk directory of file checker.jar).  Each annotated library uses
the same license as the original upstream version of the library.
Licenses for .astub files are summarized below:
checker/jtreg/*            license A
framework/tests/report/*   license B
```

LICENSE.txt
mernst: The reason I put "including" is that I'm worried that someone will add a new file without updating this list.  The general description is correct, and the list is correct at the moment.  I'll remove it nonetheless and will hope that it doesn't get out of date in the future.

mernst:
@olekw I made changes to address your comments.  Could you take another look?  Thanks!

LICENSE.txt
olekw: Thank you! I know it seems nit-picky but it really helps downstream users a lot!

LICENSE.txt
```diff
@@ -12,18 +12,18 @@ the parts that you might want to include with your own program.
  * The annotations and utility files are licensed under the MIT License.
    (The text of this license also appears below.)  This applies to the
    checker-qual*.jar and all the files that appear in it:  every file in a
-   qual/ directory, plus utility files including FormatUtil.java,
+   qual/ directory, plus utility files FormatUtil.java,
    I18nFormatUtil.java, NullnessUtil.java, Opt.java,
    PurityUnqualified.java, RegexUtil.java, SignednessUtil.java, and
    UnitsTools.java.  It also applies to other utility files, such as
    SignednessUtilExtra.java, and to the cleanroom implementations of
```
olekw: Oops, I missed one of the "weasel words" earlier. ;) Other than this, everything looks great. Thanks!

```suggestion
   UnitsTools.java.  It also applies to other utility files:
   SignednessUtilExtra.java and the cleanroom implementations of
```

olekw:
I've updated my Debian packaging to reflect these clarifications. I'm forwarding to our FTP Team (approval authorities) for their decision as soon as I rebuild it.

mernst:
Thanks again for pushing us to be more precise in our wording.
Also thanks for doing the packaging for Debian.

olekw:
You're welcome. Glad I could help. :)

PR #3433, opened 2020-07-06
Remove external links to images

olekw:
Images are already available locally. This change will prevent unnecessary Internet access and associated privacy concerns. (Imported from Debian packaging)


mernst:
Thanks for this improvement!

olekw:
You're welcome :)

PR #3440, opened 2020-07-07
Interning Checker: Use factory method to get declaration annotation.

smillst:
This is so that `@UsesObjectEquals` written stub files are read.


checker/tests/interning/SymbolCompleterUsesObjectEquals.java
```diff
@@ -0,0 +1,7 @@
+import com.sun.tools.javac.code.Symbol;
```
mernst, line 1: This won't compile without a command-line argument such as `--add-exports jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED`.  Could you remove the test or arrange for the extra command-line argument?

mernst: Or, use a different, public class that is `@UsesObjectEquals`.  (This would require renaming the test class, too.)

PR #3434, opened 2020-07-06
QualifierHierarchy documentation improvements

mernst:

Issue #2000, opened 2018-05-28
Java collection methods that accept Object should consider type hierarchy


wmdietl:
The ErrorProne check [Incompatible type as argument to Object-accepting Java collections method](http://errorprone.info/bugpattern/CollectionIncompatibleType) ensures that methods like `Collection.contains` take an argument of the right type.

It would be useful to extend this check to be aware of the type hierarchy of a particular checker.
For example, from #1999:

```java
    private final Map<String, String> map;

    public @Nullable String get(@Nullable String key) {
        return map.get(key);
    }
```

`Map.get` is declared to take a `@Nullable Object` and therefore the above code passes the Nullness Checker.
However, allowing `null` isn't useful, as the key type of `map` is `@NonNull String`.

It would be useful, for all type systems, if there were at least a warning about such incorrect usage.

PR #3445, opened 2020-07-07
Deprecate *TypeVariable methods in QualifierHierarchy and improve documentation

smillst:
These methods are barely used and are not correct.  I also improved the Javadoc of all the methods in QualifierHierarchy.

(Includes changes from ~#3434~)


framework/src/main/java/org/checkerframework/framework/type/QualifierHierarchy.java
```diff
@@ -138,298 +198,248 @@ public int numberOfIterationsBeforeWidening() {
...
     /**
-     * Tests whether {@code subAnno} is a sub-qualifier of, or equal to, {@code superAnno},
-     * according to the type qualifier hierarchy. This checks only the qualifiers, not the Java
-     * type.
-     *
-     * <p>This method works even if the underlying Java type is a type variable. In that case, a
-     * 'null' AnnnotationMirror and the empty set represent a meaningful value (namely, no
-     * annotation).
+     * Returns true if and only if {@link AnnotatedTypeMirror#getAnnotations()} can return an empty
+     * set even if the type has been fully-annotated.
```
wmdietl, line 276: Should it be `fully annotated`?

framework/src/main/java/org/checkerframework/framework/type/QualifierHierarchy.java
```diff
@@ -138,298 +198,248 @@ public int numberOfIterationsBeforeWidening() {
...
 
     /**
-     * Tests whether {@code subAnno} is a sub-qualifier of, or equal to, {@code superAnno},
-     * according to the type qualifier hierarchy. This checks only the qualifiers, not the Java
-     * type.
-     *
-     * <p>This method works even if the underlying Java type is a type variable. In that case, a
-     * 'null' AnnnotationMirror and the empty set represent a meaningful value (namely, no
-     * annotation).
+     * Returns true if and only if {@link AnnotatedTypeMirror#getAnnotations()} can return an empty
```
wmdietl, line 275: Is it really whether it can be an empty set? Shouldn't it be whether fewer than the qualifier hierarchy width qualifiers appear on the type?
For example in nullness, a `<T> @Nullable T foo()...` is valid, even though the qualifier hierarchy has width 2.

framework/src/main/java/org/checkerframework/framework/type/QualifierHierarchy.java
```diff
@@ -138,298 +198,248 @@ public int numberOfIterationsBeforeWidening() {
...
+     * @return true if the update was done; false if there was a qualifier hierarchy collision
      */
-    public abstract AnnotationMirror greatestLowerBoundTypeVariable(
-            AnnotationMirror a1, AnnotationMirror a2);
+    public <T> boolean updateMappingToMutableSet(
+            Map<T, Set<AnnotationMirror>> map, T key, AnnotationMirror qualifier) {
+        @SuppressWarnings("nullness:argument.type.incompatible") // key is of type T.
+        boolean mapContainsKey = map.containsKey(key);
+        if (mapContainsKey) {
+            @SuppressWarnings("nullness:assignment.type.incompatible") // key is a key for map.
```
wmdietl, line 339: Would using `@KeyFor(#1) T key` as parameter type work instead?

framework/src/main/java/org/checkerframework/framework/util/AtmLubVisitor.java
```diff
@@ -245,12 +245,14 @@ private void lubWildcard(
...
-                    qualifierHierarchy.greatestLowerBound(
-                            type1LowerBound,
-                            type2LowerBound,
-                            type1LowerBound.getAnnotationInHierarchy(top),
-                            type2LowerBound.getAnnotationInHierarchy(top));
+            AnnotationMirror anno1 = type1LowerBound.getAnnotationInHierarchy(top);
+            AnnotationMirror anno2 = type2LowerBound.getAnnotationInHierarchy(top);
+
+            AnnotationMirror glb = null;
+            if (anno1 != null && anno2 != null) {
```
wmdietl, line 252: Is this equivalent and a bit clearer:

```java
if (anno1 != null && anno2 != null) {
    AnnotationMirror glb = qualifierHierarchy.greatestLowerBound(anno1, anno2);
    lubLowerBound.replaceAnnotation(glb);
}
```

As anno1 and anno2 are from the same hierarchy, their glb must exist. The two ifs are confusing the logic for me.

framework/src/main/java/org/checkerframework/framework/type/QualifierHierarchy.java
smillst: No, `key` does not be a key in `map`.

framework/src/main/java/org/checkerframework/framework/type/QualifierHierarchy.java
```diff
@@ -5,23 +5,32 @@
...
 import org.checkerframework.javacutil.BugInCF;
 
 /**
- * Represents a type qualifier hierarchy.
+ * Represents multiple type qualifier hierarchies. {@link #getWidth} gives the number of hierarchies
+ * that this object represents. Each hierarchy has its own top and bottom, and subtyping
+ * relationships exist only within each hierarchy.
  *
- * <p>All method parameter annotations need to be type qualifiers recognized within this hierarchy.
+ * <p>Note the distinction in terminology between a qualifier hierarchy, which has one top and one
```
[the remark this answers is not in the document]
wmdietl: We discussed this.

framework/src/main/java/org/checkerframework/framework/type/QualifierHierarchy.java
```diff
@@ -38,40 +47,48 @@ public int getWidth() {
...
      * @return the top (ultimate super) type qualifiers in the type system
      */
     public abstract Set<? extends AnnotationMirror> getTopAnnotations();
 
     /**
      * Return the top qualifier for the given qualifier, that is, the qualifier that is a supertype
-     * of start but no further supertypes exist.
+     * of {@code qualifier} but no further supertypes exist.
+     *
+     * @param qualifier any qualifier from one of the type hierarchies represented by this
```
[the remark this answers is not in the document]
smillst: My thought was to leave it undefined.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeValidator.java
```diff
@@ -111,9 +111,9 @@ protected boolean shouldCheckTopLevelDeclaredType(AnnotatedTypeMirror type, Tree
      *
      * <ol>
      *   <li>There should not be multiple annotations from the same hierarchy.
```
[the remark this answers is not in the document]
smillst: No, this is a qualifier hierarchy.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeValidator.java
```diff
@@ -111,9 +111,9 @@ protected boolean shouldCheckTopLevelDeclaredType(AnnotatedTypeMirror type, Tree
...
-     *   <li>There should not be more annotations than the width of the qualifier hierarchy.
+     *   <li>There should not be more annotations than the width of the QualifierHierarchy.
```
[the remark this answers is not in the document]
smillst: I don't think it needs to be linked every place it is referring to the QualifierHierarchy.

framework/src/main/java/org/checkerframework/framework/type/QualifierHierarchy.java
```diff
@@ -32,46 +41,56 @@
...
      */
-    public abstract Set<? extends AnnotationMirror> getTopAnnotations();
+    Set<? extends AnnotationMirror> getTopAnnotations();
 
     /**
      * Return the top qualifier for the given qualifier, that is, the qualifier that is a supertype
-     * of start but no further supertypes exist.
+     * of {@code qualifier} but no further supertypes exist.
+     *
+     * @param qualifier any qualifier from one of the type hierarchies represented by this
```
wmdietl, line 63: Here and in a few other comments, should the "type hierarchies" be replaced with "qualifier hierarchies", to not confuse with TypeHierarchy instances? Sometimes it's the combined "type qualifier hierarchies".

framework/src/main/java/org/checkerframework/framework/type/QualifierHierarchy.java
```diff
@@ -81,361 +100,351 @@ public int getWidth() {
...
-     * <p>This method works even if the underlying Java type is a type variable. In that case, a
-     * 'null' AnnnotationMirror and the empty set represent a meaningful value (namely, no
-     * annotation).
+     * Returns true if and only if {@link AnnotatedTypeMirror#getAnnotations()} can return a set
+     * with fewer qualifiers than the width of the QualifierHierarchy.
      *
-     * @param subAnno a qualifier that might be a subtype
-     * @param superAnno a qualifier that might be a subtype
-     * @return true iff {@code subAnno} is a sub qualifier of, or equal to, {@code superAnno}
+     * @param type annotated type mirror
```
wmdietl, line 277: Just repeating the type doesn't add much. "the type to test"?

framework/src/main/java/org/checkerframework/framework/type/QualifierHierarchy.java
wmdietl: Could you update the description of the `key` parameter then? The `the key to update` description would be clearer as `the key to update or add`.

framework/src/main/java/org/checkerframework/framework/type/QualifierHierarchy.java
```diff
@@ -81,361 +100,351 @@ public int getWidth() {
...
+     * @param key the key to update
+     * @param qualifier the value to add
+     * @param <T> type of the map's keys
+     * @return true if the update was done; false if there was a qualifier hierarchy collision
      */
-    public abstract AnnotationMirror greatestLowerBoundTypeVariable(
-            AnnotationMirror a1, AnnotationMirror a2);
+    default <T> boolean updateMappingToMutableSet(
+            Map<T, Set<AnnotationMirror>> map, T key, AnnotationMirror qualifier) {
+        @SuppressWarnings("nullness:argument.type.incompatible") // key is of type T.
```
wmdietl, line 335: I'm confused about the suppression. Both the variable `key` and the key-type of `map` are `T`, so just saying `key is of type T`, doesn't explain why there is an error that can be suppressed.

framework/src/main/java/org/checkerframework/framework/type/QualifierHierarchy.java
```diff
@@ -81,361 +100,351 @@ public int getWidth() {
...
+     * superQualifier}, according to the type qualifier hierarchy.
+     *
+     * <p>This method works even if the underlying Java type is a type variable. In that case, a
+     * 'null' AnnotationMirror is a legal argument that represents no annotation.
+     *
+     * @param subQualifier a qualifier that might be a subtype
+     * @param superQualifier a qualifier that might be a subtype
+     * @return true iff {@code subQualifier} is a subqualifier of, or equal to, {@code
+     *     superQualifier}
+     * @deprecated Without the bounds of the type variable, it is not possible to correctly compute
```
wmdietl, line 398: For all the deprecated methods, can you point to the methods that should be used instead?

framework/src/main/java/org/checkerframework/framework/type/QualifierHierarchy.java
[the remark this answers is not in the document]
wmdietl: At some point it would probably be good to resolve all such undefined behavior, but I'm fine with ignoring it for now.

framework/src/main/java/org/checkerframework/framework/type/QualifierHierarchy.java
```diff
@@ -81,361 +100,351 @@ public int getWidth() {
...
+     * @param annos set of annotations to search
+     * @param annotationMirror annotation that is in the same hierarchy as the returned annotation
+     * @return annotation in the same hierarchy as annotationMirror, or null if one is not found
      */
-    // This method requires more revision.
-    public abstract boolean isSubtypeTypeVariable(
-            Collection<? extends AnnotationMirror> subAnnos,
-            Collection<? extends AnnotationMirror> superAnnos);
+    default @Nullable AnnotationMirror findAnnotationInSameHierarchy(
+            Collection<? extends AnnotationMirror> annos, AnnotationMirror annotationMirror) {
```
wmdietl, line 298: Other methods in this interface would call these `qualifiers` and `qualifier`, I think. Let's try to use consistent names within a class/interface. A few other places below.

framework/src/main/java/org/checkerframework/framework/type/QualifierHierarchy.java
smillst: The type of the parameter of `containsKey` is `@NonNull Object`.  Call is legal because key is of type `T`.  I added a reference to https://github.com/typetools/checker-framework/issues/2000.

PR #3450, opened 2020-07-09
Do not do inference on temporary variables in whole-program inference

### LINKED ISSUE #3442: Temporary variable written as field in static block with -Ainfer
When the increment operator is used inside of a Java `static` block, a temporary variable (which I believe is create by [this line](https://github.com/typetools/checker-framework/blob/dc530c37903c608f02e6adea752425eae7f06baf/dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGBuilder.java#L4751)) is recorded as a field and written out to stub/jaif files with -Ainfer.

With this class
```java
public class TempVars {
    static {
        int i = 0;
        i++;
    }
}
```
and this command
```bash
$CHECKERFRAMEWORK/checker/bin/javac -processor NullnessChecker -Ainfer=stubs TempVars.java
```
the generated stub file looks like
```java
import org.checkerframework.checker.initialization.qual.Initialized;

class TempVars {

  // fields:

  @Initialized @NonNull int tempPostfix#num0;

}
```
This class has no fields but there's a variable `tempPostfix` written under fields.

This also happens with `jaif` files, but with those it causes a crash.

Output for
```bash
$CHECKERFRAMEWORK/checker/bin/javac -processor NullnessChecker -Ainfer=stubs TempVars.java -version -verbose -AprintAllQualifiers
```

[javac_temp_vars.txt](https://github.com/typetools/checker-framework/files/4886656/javac_temp_vars.txt)

kelloggm:
fixes #3442 for both -Ainfer=stubs and -Ainfer=jaifs

### DIFF
--- a/framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
+++ b/framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
@@ -284,7 +284,11 @@ public class WholeProgramInferenceScenes implements WholeProgramInference {
                             + lhs.getClass());
         }
 
-        // Do not attempt to infer types for temporary variables.
+        // Do not attempt to infer types for fields that do not have valid
+        // names. For example, compiler-generated temporary variables will
+        // have invalid names. Recording facts about fields with
+        // invalid names causes jaif-based WPI to crash when reading the .jaif
+        // file, and stub-based WPI to generate unparseable stub files.
         // See https://github.com/typetools/checker-framework/issues/3442
         if (!SourceVersion.isIdentifier(fieldName)) {
             return;
--- a/framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
+++ b/framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
@@ -9,6 +9,7 @@ import com.sun.tools.javac.code.Symbol.MethodSymbol;
 import com.sun.tools.javac.code.Symbol.VarSymbol;
 import java.util.List;
 import java.util.Map;
+import javax.lang.model.SourceVersion;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.VariableElement;
@@ -281,6 +282,12 @@ public class WholeProgramInferenceScenes implements WholeProgramInference {
             throw new BugInCF(
                     "updateFromFieldAssignment received an unexpected node type: "
                             + lhs.getClass());
+        }
+
+        // Do not attempt to infer types for temporary variables.
+        // See https://github.com/typetools/checker-framework/issues/3442
+        if (!SourceVersion.isIdentifier(fieldName)) {
+            return;
         }
 
         // If the inferred field has a declaration annotation with the
--- a/framework/tests/whole-program-inference/non-annotated/Tempvars.java
+++ b/framework/tests/whole-program-inference/non-annotated/Tempvars.java
@@ -1,0 +1,8 @@
+// test case for https://github.com/typetools/checker-framework/issues/3442
+
+public class Tempvars {
+    static {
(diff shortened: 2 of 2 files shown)


framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
```diff
@@ -283,6 +284,12 @@ public void updateFromFieldAssignment(
                             + lhs.getClass());
         }
 
+        // Do not attempt to infer types for temporary variables.
```
jwaataja, line 287: It's unclear how this comment applies to the check being performed. Maybe something like "Do not attempt to infer types for variables with invalid names, which can be generated by temporary variables. See ...".

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
```diff
@@ -283,6 +284,12 @@ public void updateFromFieldAssignment(
...
+        // See https://github.com/typetools/checker-framework/issues/3442
+        if (!SourceVersion.isIdentifier(fieldName)) {
```
jwaataja, line 289: Two questions.

1. Is it having a valid identifier equivalent equivalent to being a temporary variable? If not, then this fix might work for the issue in question, but still do inference on other temporary nodes. I noticed near [this](https://github.com/typetools/checker-framework/blob/d3da57a5f9e0bc3b9bbf294159dc7283ec0a8c16/dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGBuilder.java#L4759) line from the code that creates the variable that `setInSource(false)` is called on the generated nodes. Would something like that work here instead?

2. Why is the temporary variable considered a field in the first place? I feel like we *would* want to do inference on temporary variables in an expression like `int n = i++` because the temporary variable would hold the inferred type of the expression `i++`. It seems to me like the root of the problem is that in static blocks in particular, the temporary variable is somehow "attached" to the class as a field.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
mernst, line 287: Can you define the term "temporary variable"?  Do you mean "compiler-generated"?

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
kelloggm: 1. I decided to check the identifier because I think it's broader than checking whether the tree was actually generated by the compiler (i.e. `setInSource` being set to `false`). My reasoning is that because we're going to write the results of inference out to a jaif/stub file (neither of which can include fields with invalid names), we should handle ANY invalid names that we encounter, regardless of where they came from.

2. 
> Why is the temporary variable considered a field in the first place
I think it's just a quirk of how the CFG is constructed.
> I feel like we would want to do inference on temporary variables in an expression like `int n = i++`
We'll still do WPI on `n` in this example, and we'll compute the correct results because *local* inference will still run on the `i++` node - but only `n`'s inferred type will be written to the jaif/stub file.

> It seems to me like the root of the problem is that in static blocks in particular, the temporary variable is somehow "attached" to the class as a field
Yes. I presume there's a reason that was done. If there isn't (and it's just an implementation detail that could be changed), I would still prefer this fix, because it has a much smaller blast radius (only WPI is impacted). Changing CFG construction could result in bugs elsewhere.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
kelloggm: Yes, that's what I was referring to. I will change this documentation to refer to variables with invalid names, and give compiler-generated temporary variables as an example.

PR #3456, opened 2020-07-09
Don't ignore type arguments (Fixes #3443.)

### LINKED ISSUE #3443: DefaultTypeHierarchy.visitTypevarSubtype can ignore type arguments
```java
$ cat OnlyTopLevelForTypeVariable.java

class OnlyTopLevelForTypeVariable {
  static <T extends Supplier<@Nullable String>> Supplier<String> passThrough(T t) {
    return t;
  }

    Supplier<@Nullable String> s1 = () -> null;
    Supplier<String> s2 = passThrough(s1);
    s2.get().toString();
  }
}

interface Supplier<T> {
  T get();
}

$ $HOME/checker-framework-3.4.1/checker/bin/javac -processor org.checkerframework.checker.nullness.NullnessChecker OnlyTopLevelForTypeVariable.java && java OnlyTopLevelForTypeVariable
        at OnlyTopLevelForTypeVariable.main(OnlyTopLevelForTypeVariable.java:11)
```

I would have expected an error in the implementation of `passThrough`, since a `Supplier<@Nullable String>` isn't a `Supplier<String>` (as the Checker Framework knows if I use `Supplier<@Nullable String>` directly instead of `T`).

I think this is a result of the code somewhere around here:
https://github.com/typetools/checker-framework/blob/4bd0591172fbf54cf128174589d68aebb2825eab/framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java#L1023

As I understand it, we end up checking `isPrimarySubtype` but not the arguments.

I didn't encounter this in "real" code.

smillst:
Fixes #3443.


checker/tests/nullness/Issue3443.java
```diff
@@ -0,0 +1,18 @@
...
+
+class Issue3443 {
+    static <T extends Supplier3443<@Nullable String>> Supplier3443<String> passThrough(T t) {
+        // :: error: (return.type.incompatible)
+        return t;
+    }
+
+    public static void main(String[] args) {
+        Supplier3443<@Nullable String> s1 = () -> null;
+        Supplier3443<String> s2 = passThrough(s1);
```
wmdietl, line 11: Shouldn't there also be an error on this line? Either that the inferred type argument is incompatible or that the argument is incompatible?
I guess that is masked by #979 work-arounds, but it might be good to highlight that we would expect an error here.
This is important if `passThrough` is in a library and we don't "see" that there is an error in the implementation.

smillst: No, the return type of pass through is `Supplier3443<String>`.

smillst: Yes, I added a comment.

PR #3460, opened 2020-07-09
Create util class for component discovery

lnsun:
The logic for finding visitors and factories by checker naming convention in base checker and visitor contains duplication.

By introducing a new util class to handle this, not only removes the duplication, but also benefits new checker components. (for example, maybe the `TypeVaildator`s will benefit from this...?)

Note: `TestChecker` and `LubGlbChecker` does not have explicit setters for visitor and factory and depends on the base to find, so this change does not need an additional test checker.

Update: Analysis and Transfer added.


lnsun:
@mernst @wmdietl `createFlowAnalysis` and `createFlowTransferFunction` updated.

framework/src/main/java/org/checkerframework/framework/util/ComponentFinderUtil.java
```diff
@@ -0,0 +1,83 @@
...
+         */
+        DefaultType getDefault(BaseTypeChecker checker);
+    }
+
+    /**
+     * Find a component named with the checker naming convention.
+     *
+     * @param <T> type of the component
+     * @param checker the current checker
+     * @param replacement the component suffix, e.g. "Visitor" or "AnnotatedTypeFactory"
```
mernst, line 34: If this is a suffix, why not name it "suffix".  The documentation doesn't indicate why it is named "replacement".  The fact that replacement is done within a method called by this one is an implementation detail.

framework/src/main/java/org/checkerframework/framework/util/ComponentFinderUtil.java
```diff
@@ -0,0 +1,83 @@
...
+     * @param defaultGetter a getter for the default value when cannot find the class
```
mernst, line 35: What is "the default value"?

framework/src/main/java/org/checkerframework/framework/util/ComponentFinderUtil.java
```diff
@@ -0,0 +1,83 @@
...
+        return defaultGetter.getDefault(checker);
+    }
+
+    /**
+     * Find a component named with the checker naming convention with its constructor invoked with
+     * the only argument {@code checker}
+     *
+     * @param <T> type of the component
+     * @param checker a checker
+     * @param replacement the component suffix, e.g. "Visitor" or "AnnotatedTypeFactory"
```
mernst, line 70: "replacement" is a poor name for this formal parameter.

framework/src/main/java/org/checkerframework/framework/util/ComponentFinderUtil.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+        return defaultGetter.getDefault(checker);
+    }
+
+    /**
+     * Find a component named with the checker naming convention with its constructor invoked with
+     * the only argument {@code checker}
+     *
+     * @param <T> type of the component
+     * @param checker a checker
```
mernst, line 69: How is this used?

framework/src/main/java/org/checkerframework/framework/util/ComponentFinderUtil.java
```diff
@@ -0,0 +1,83 @@
...
+     * @param replacement the component suffix, e.g. "Visitor" or "AnnotatedTypeFactory"
+     * @param defaultGetter a getter for the default value when cannot find the class
+     * @return the properly-named component found
```
mernst, line 72: I don't think "properly-named" adds anything; you could remove it and make the documentation clearer.

framework/src/main/java/org/checkerframework/framework/util/ComponentFinderUtil.java
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * A single method interface for getting default type for components by method reference.
+     *
+     * @param <DefaultType> the type for the default value
+     */
+    public interface DefaultGetter<DefaultType> {
+
+        /**
+         * The logic for getting the default value if all attempt fails
```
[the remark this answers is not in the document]
lnsun: The whole `DefaultGetter` does not make sense, so it is replaced with the built-in `Supplier`

framework/src/main/java/org/checkerframework/framework/util/ComponentFinderUtil.java
lnsun: I found both visitors and factories' constructors use the only parameter `checker`. I thought it would be useful to create a shortcut for that.

mernst:
@wmdietl Why was this closed?  Is there a problem with the logic?

PR #3464, opened 2020-07-11
Add -Aversion command-line option; fixes #3381

### LINKED ISSUE #3381: -version produces "javac (version info not available)"
Since the default issue template suggests including the output of `-version`, I've noticed that, for a while now, the output of that has been:

`javac (version info not available)`

If I go all the way back to Checker Framework 2.10.0, I see:

`javac 1.8.0-jsr308-2.4.0`

But every version I had handy for testing from 2.11.0 onward has "version info not available."

(I wonder if ideally the version output would include the version of the Checker Framework itself?)

PRITI1999:
#3381 Add -Aversion option


framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -253,6 +253,8 @@
 
     /// Amount of detail in messages
 
+    // Print the version of the checker framework
```
mernst, line 256: Please use correct capitalization for Checker Framework.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -781,6 +783,14 @@ public void run() {
                                     }
                                 });
             }
+            if (hasOption("version")) {
+                String version = getCheckerVersion();
+                if (version == null) {
+                    messager.printMessage(Kind.NOTE, "Version info not available!");
+                } else {
+                    messager.printMessage(Kind.NOTE, "checker-framework " + version);
```
mernst, line 791: Use correct name for Checker Framework.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -2530,4 +2540,28 @@ private void printGitProperties() {
             System.out.println("IOException while reading git.properties: " + e.getMessage());
         }
     }
+
+    /** Extract the version of the Checker Framework */
```
mernst, line 2544: You need to document that this can return null.  Under what circumstances can it do so?  Should it throw an exception rather than returning null?

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -2530,4 +2540,28 @@ private void printGitProperties() {
...
+        String startTag = p.getProperty("checkers.ver.0");
+        String endTag = p.getProperty("checkers.ver.1");
+        String XML_FILE = "/docs/examples/MavenExample/pom.xml";
+        try {
+            InputStream in = getClass().getResourceAsStream(XML_FILE);
+            BufferedReader reader = new BufferedReader(new InputStreamReader(in));
+            String line = null;
+            while ((line = reader.readLine()) != null && version == null) {
+                if (line.split(startTag).length > 1) {
+                    version = line.split(startTag)[1].split(endTag)[0];
```
mernst, line 2559: Why does this loop continue after setting `version`?  Could you use a `return` statement here?

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -2530,4 +2540,28 @@ private void printGitProperties() {
             System.out.println("IOException while reading git.properties: " + e.getMessage());
         }
     }
+
+    /** Extract the version of the Checker Framework */
+    private String getCheckerVersion() {
+        String version = null;
+        Properties p = new Properties();
+        String RLS_FILE = "/docs/developer/release/release.properties";
+        p.putAll(getProperties(getClass(), RLS_FILE));
```
mernst, line 2549: Rather than creating an empty Properties on line 2547 and then merging a Properties into it, can this line be
```
Properties p = getProperties(getClass(), RLS_FILE);
```
?

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -2530,4 +2540,28 @@ private void printGitProperties() {
...
+            InputStream in = getClass().getResourceAsStream(XML_FILE);
+            BufferedReader reader = new BufferedReader(new InputStreamReader(in));
+            String line = null;
+            while ((line = reader.readLine()) != null && version == null) {
+                if (line.split(startTag).length > 1) {
+                    version = line.split(startTag)[1].split(endTag)[0];
+                }
+            }
+        } catch (IOException e) {
+            System.out.println("IOException while reading version information: " + e.getMessage());
```
mernst, line 2563: Is it a bug if this line is reached?  If so, it is better to fail than to print a line that users are likely to overlook.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -781,6 +783,14 @@ public void run() {
                                     }
                                 });
             }
+            if (hasOption("version")) {
+                String version = getCheckerVersion();
+                if (version == null) {
+                    messager.printMessage(Kind.NOTE, "Version info not available!");
```
mernst, line 789: Under what circumstances is `version` null?  Is that a bug?

PRITI1999: Its never null for the current version. It can only be null if the tag in release.properties is not found in `XML_FILE`. Should I remove this check altogether and throw an exception if ever version is null?

mernst: Since that would indicate an unexpected problem, I think that is the best approach.

PRITI1999: Made the changes accordingly. Please re-review.

mernst: @PRITI1999 CI does not pass.  Please make it pass before requesting a review.

PRITI1999: I apologize. CI succeeds now.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -2530,4 +2535,33 @@ private void printGitProperties() {
             System.out.println("IOException while reading git.properties: " + e.getMessage());
         }
     }
+
+    /**
+     * Extract the version of the Checker Framework
```
mernst, line 2540: Minor: end sentences with punctuation (in this case, a period).

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -2530,4 +2535,33 @@ private void printGitProperties() {
...
+     *
+     * @throws IOException if unable to read docs/developer/release/release.properties or
+     *     docs/examples/MavenExample/pom.xml file
+     * @throws NullPointerException if tag present in release.properties for Checker Framework
+     *     version is not found in docs/examples/MavenExample/pom.xml
+     * @return Checker Framework version {@link String}
```
mernst, line 2546: It is usually not necessary to state the type, which is already stated nearby in the source code and in the generated API documentation.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -2530,4 +2535,33 @@ private void printGitProperties() {
...
+     */
+    private String getCheckerVersion() throws IOException {
+        String version = null;
```
mernst, line 2549: Why is this variable needed?

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
mernst, line 2540: If you use "Extract", say from where.  Even better, I would leave implementation details like that out of the specification and just use "Returns".

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -2530,4 +2535,33 @@ private void printGitProperties() {
...
+        InputStream in = getClass().getResourceAsStream(XML_FILE);
+        BufferedReader reader = new BufferedReader(new InputStreamReader(in));
+        String line = null;
+        while ((line = reader.readLine()) != null) {
+            if (line.split(startTag).length > 1) {
+                version = line.split(startTag)[1].split(endTag)[0];
+                return version;
+            }
+        }
+        throw new NullPointerException(
```
mernst, line 2564: This is an odd choice of exception.  Why not use `BugInCF`?

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -2530,4 +2535,33 @@ private void printGitProperties() {
...
+        Properties releaseProperties = getProperties(getClass(), RLS_FILE);
+        String startTag = releaseProperties.getProperty("checkers.ver.0");
+        String endTag = releaseProperties.getProperty("checkers.ver.1");
+        String XML_FILE = "/docs/examples/MavenExample/pom.xml";
+        InputStream in = getClass().getResourceAsStream(XML_FILE);
+        BufferedReader reader = new BufferedReader(new InputStreamReader(in));
+        String line = null;
+        while ((line = reader.readLine()) != null) {
+            if (line.split(startTag).length > 1) {
+                version = line.split(startTag)[1].split(endTag)[0];
```
mernst, line 2560: This line can be combined with the following one.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -253,6 +253,8 @@
 
     /// Amount of detail in messages
 
+    // Print the version of the Checker Framework
+    "version",
```
mernst, line 257: This user-visible change needs to be documented in the manual and in the changelog.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -2530,4 +2535,33 @@ private void printGitProperties() {
...
+     *     version is not found in docs/examples/MavenExample/pom.xml
+     * @return Checker Framework version {@link String}
+     */
+    private String getCheckerVersion() throws IOException {
+        String version = null;
+        String RLS_FILE = "/docs/developer/release/release.properties";
+        Properties releaseProperties = getProperties(getClass(), RLS_FILE);
+        String startTag = releaseProperties.getProperty("checkers.ver.0");
+        String endTag = releaseProperties.getProperty("checkers.ver.1");
+        String XML_FILE = "/docs/examples/MavenExample/pom.xml";
```
mernst, line 2554: Does this work?  It crashed for me when I tried to run it, because neither of the two files is not available in `checker.jar`.

mernst: You could use `git.properties` instead.  See method `printGitProperties`.

PRITI1999:
@mernst Added the suggested changes. Please re-review.

mernst:
Thanks for this feature!

PR #3465, opened 2020-07-11
Linear checker

aditya3434:
A modified version of the deleted linear checker in Checker Framework which acts as an extension of the Aliasing Checker. Objects annotated as @Linear are similar to @Unique objects, the only difference being that they can only be operated on once, after which, they are "used up". This "used up" state is denoted by the @Unusable annotation. The hierarchy is @MaybeAliased (top) <- @Unique <- @Linear <- @Unusable (bottom). This version of linear checker is created to combat various problems like the java.util.Stream issue where the Stream gives an error when it is operated more than once, without any warning from the compiler.


mernst:
CI passes, despite the red X.

wmdietl:
We're discussing whether this should be combined with the Aliasing Checker, to make one combined checker that handles aliasing and linear types.

wmdietl:
This needs more review before making a PR.

PR #3468, opened 2020-07-12
Add extraArgs argument to commonAssignmentCheck

mernst:
Merge together with https://github.com/typetools/checker-framework-inference/pull/165


framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -942,6 +942,21 @@ public void reportWarning(
...
 
+    /**
+     * Given an array and two values, creates an array with all the values.
+     *
+     * @param array the array to extend
+     * @param elt1 the first element to add
+     * @param elt2 the second element to add
+     * @return a copy of {@code array}, with two elements added
+     */
+    public static Object[] arrayPlusTwoElements(Object[] array, Object elt1, Object elt2) {
```
smillst, line 953: This is an unexpected place to put a utility method.  I think `org.checkerframework.javacutil.SystemUtil` would be a better location.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -942,6 +942,21 @@ public void reportWarning(
         report(source, Kind.MANDATORY_WARNING, messageKey, args);
     }
 
+    /**
+     * Given an array and two values, creates an array with all the values.
```
smillst, line 946: The Javadoc should say the two values are at the end of the new array.

javacutil/src/main/java/org/checkerframework/javacutil/SystemUtil.java
```diff
@@ -225,4 +226,21 @@ public static int getJreVersion() {
...
+    /**
+     * Given an array and two values, creates an array with all the values. The two extra values are
+     * at the end of the returned array.
+     *
+     * @param array the array to extend
+     * @param elt1 the first element to add
+     * @param elt2 the second element to add
+     * @return a copy of {@code array}, with two elements added at the end
+     */
+    public static Object[] arrayPlusTwoElements(Object[] array, Object elt1, Object elt2) {
```
wmdietl, line 239: This would be more general with a method type parameter `<T>` instead of `Object`.

wmdietl:
> I don't see any call that actually passes something for `extraArgs`, so I don't understand what the intention of this change is.
> Can you elaborate on the motivation?

Oh, I guess it is for #3469, which isn't linked to from here. Let me know if there is a different motivation.

mernst:
> > I don't see any call that actually passes something for `extraArgs`, so I don't understand what the intention of this change is.
> > Can you elaborate on the motivation?
> 
> Oh, I guess it is for #3469, which isn't linked to from here. Let me know if there is a different motivation.

That is correct.

javacutil/src/main/java/org/checkerframework/javacutil/SystemUtil.java
```diff
@@ -225,4 +226,22 @@ public static int getJreVersion() {
...
+     * Given an array and two values, creates an array with all the values. The two extra values are
+     * at the end of the returned array.
+     *
+     * @param <T> the type of the array elements
+     * @param array the array to extend
+     * @param elt1 the first element to add
+     * @param elt2 the second element to add
+     * @return a copy of {@code array}, with two elements added at the end
+     */
+    public static <T> T[] arrayPlusTwoElements(T[] array, T elt1, T elt2) {
```
wmdietl, line 240: Is there a reason to limit this to two elements instead of making it a varargs method?

PR #3469, opened 2020-07-12
Enhance message for argument.type.incompatible

mernst:
Review after ~https://github.com/typetools/checker-framework/pull/3468~ is merged.


framework/src/main/java/org/checkerframework/common/basetype/messages.properties
```diff
@@ -13,8 +13,8 @@ vector.copyinto.type.incompatible=incompatible component type in Vector.copyinto
 return.type.incompatible=incompatible types in return.%ntype of expression: %s%nmethod return type: %s
 annotation.type.incompatible=incompatible types in annotation.%nfound   : %s%nrequired: %s
 conditional.type.incompatible=incompatible types in conditional expression.%nfound   : %s%nrequired: %s
-type.argument.type.incompatible=incompatible types in type argument.%nfound   : %s%nrequired: %s
-argument.type.incompatible=incompatible types in argument.%nfound   : %s%nrequired: %s
+type.argument.type.incompatible=invalid type argument for type parameter %s of %s.%nfound   : %s%nrequired: %s
```
wmdietl, line 16: I would rather keep this as `incompatible type argument for type parameter %s of %s...` and `incompatible argument for parameter %s of %s.` to stay consistent with all the other `type.incompatible` messages.

checker/src/main/java/org/checkerframework/checker/guieffect/GuiEffectVisitor.java
```diff
@@ -487,6 +489,9 @@ private void scanUp(TreePath path) {
                 List<? extends ExpressionTree> args = invocationTree.getArguments();
                 ParameterizedExecutableType mType = atypeFactory.methodFromUse(invocationTree);
                 AnnotatedExecutableType invokedMethod = mType.executableType;
+                ExecutableElement method = TreeUtils.elementFromUse(invocationTree);
```
wmdietl, line 492: `invokedMethod.getElement()` should give you the same result.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeValidator.java
```diff
@@ -460,7 +461,16 @@ protected Void visitParameterizedType(AnnotatedDeclaredType type, ParameterizedT
         List<AnnotatedTypeParameterBounds> bounds =
                 atypeFactory.typeVariablesFromUse(type, element);
 
-        visitor.checkTypeArguments(tree, bounds, type.getTypeArguments(), tree.getTypeArguments());
+        DeclaredType declType = type.getUnderlyingType();
+        TypeElement typeElt = (TypeElement) declType.asElement();
```
wmdietl, line 465: There already is a `element` declared on line 456 that has the same value.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1368,11 +1369,19 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
             paramBounds.add(param.getBounds());
         }
 
-        checkTypeArguments(node, paramBounds, typeargs, node.getTypeArguments());
+        ExecutableElement method = TreeUtils.elementFromUse(node);
```
wmdietl, line 1372: `invokedMethod.getElement()` should give you the same result.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1681,15 +1690,24 @@ public Void visitNewClass(NewClassTree node, Void p) {
         List<AnnotatedTypeMirror> params =
                 AnnotatedTypes.expandVarArgs(atypeFactory, constructor, passedArguments);
 
-        checkArguments(params, passedArguments);
+        ExecutableElement method = TreeUtils.elementFromUse(node);
```
wmdietl, line 1693: `constructor.getElement()` should give you the same result.

wmdietl, line 1693: `method` and `methodName` are bad variable names for constructors.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -2921,25 +2948,46 @@ protected void checkConstructorInvocation(
...
      *     type.
      * @param passedArgs the expressions passed to the corresponding types
+     * @param methodName the name of the method being called
+     * @param paramNames the names of the callee's formal parameters
      */
     protected void checkArguments(
             List<? extends AnnotatedTypeMirror> requiredArgs,
-            List<? extends ExpressionTree> passedArgs) {
+            List<? extends ExpressionTree> passedArgs,
+            Name methodName,
```
wmdietl, line 2957: This is also used for constructors, so `methodName` is a bit misleading.

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractTransfer.java
```diff
@@ -156,7 +156,12 @@ protected V finishValue(V value, S store) {
      * This method is called before returning the abstract value {@code value} as the result of the
      * transfer function. By default, the value is not changed but subclasses might decide to
      * implement some functionality. The store at this position is also passed (two stores, as the
-     * result is a {@link ConditionalTransferResult}.
+     * result may be a {@link ConditionalTransferResult}.
```
wmdietl, line 159: Not sure why this is in this PR, but I'm not sure why this should be a `may` instead of `is` as before.

checker/src/main/java/org/checkerframework/checker/i18nformatter/I18nFormatterVisitor.java
```diff
@@ -87,9 +90,14 @@ private void checkInvocationFormatFor(I18nFormatCall fc) {
                                     break;
                                 default:
                                     if (!fc.isValidParameter(formatCat, paramType)) {
+                                        ExecutableElement method =
```
wmdietl, line 93: I would probably move this outside of the loop, to make clear that it doesn't change each iteration.

checker/src/main/java/org/checkerframework/checker/formatter/FormatterVisitor.java
```diff
@@ -88,9 +89,14 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
                                     default:
                                         if (!fc.isValidParameter(formatCat, paramType)) {
                                             // II.3
+                                            ExecutableElement method =
```
wmdietl, line 92: I would probably move this outside of the loop, to make clear that it doesn't change each iteration.

mernst: It's within the loop because it won't every be used in most loop executions.  Executing it outside the loop is only faster if on average every loop uses it more than 1 time, but is wasteful if each loop on average uses it less than 1 time.

checker/src/main/java/org/checkerframework/checker/i18nformatter/I18nFormatterVisitor.java
mernst: Same reasoning as for similar case above.

checker/src/main/java/org/checkerframework/checker/formatter/FormatterVisitor.java
```diff
@@ -88,9 +89,14 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
...
+                                                    TreeUtils.elementFromUse(node);
+                                            Name methodName = method.getSimpleName();
                                             tu.failure(
                                                     param,
                                                     "argument.type.incompatible",
+                                                    "", // parameter name is not useful
```
wmdietl, line 98: I don't see a simple way around this, but this will result in a message like

```
incompatible argument for parameter  of foo
```

with two spaces between `parameter` and `of`. (Interestingly, the markup doesn't show two spaces with inline code...)

mernst: You are right.  I didn't see a good way around it either.  It would be possible to fix by passing a single extra argument with text "parameter arg of foo" or "parameter of foo", but to me that looked uglier (in the source code) than leaving an extra space (in the output).  Let's fix it if a user complains.

PR #3473, opened 2020-07-14
Improve the documentation of Receiver

mernst:


dataflow/src/main/java/org/checkerframework/dataflow/analysis/FlowExpressions.java
```diff
@@ -447,8 +447,22 @@ private static Receiver internalReprOfMemberSelect(
     }
 
     /**
-     * The poorly-named Receiver class is actually a Java AST. Each subclass represents a different
-     * type of expression, such as MethodCall, ArrayAccess, LocalVariable, etc.
+     * This class represents a Java expression and its type. However, it does not represent all
+     * possible Java expressions (for example, it does not represent a ternary expression; use
+     * {@link FlowExpressions.Unknown} for unrepresentable expressions). Furthermore, it supports
```
smillst, line 452: Our "Java" expression language supports "#1", but it is parsed into a `LocalVariable`.  So this comment here doesn't really make sense.

mernst: @smillst Can you clarify your concern?  Do you mean that it doesn't make sense to call `#1` a formal parameter?  From the perspective of a user that's what it is used for.  The Java compiler and JVM do treat formal parameters as local variables, but I think of that as an implementation detail.  Can you suggest alternate wording if this comment is misleading?

smillst: The Javadoc seems to imply that there are Receiver classes for the extensions to the Java language, but there aren't.  `#1` is represented by a local variable and `<self>` is  converted before parsing by the Lock Checker.

mernst: Thanks for clarifying.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/FlowExpressions.java
```diff
@@ -447,12 +447,13 @@ private static Receiver internalReprOfMemberSelect(
...
-     * possible Java expressions (for example, it does not represent a ternary expression; use
-     * {@link FlowExpressions.Unknown} for unrepresentable expressions). Furthermore, it supports
-     * some <a
+     * This class represents a Java expression and its type. It does not represent all possible Java
+     * expressions (for example, it does not represent a ternary expression; use {@link
+     * FlowExpressions.Unknown} for unrepresentable expressions). Furthermore, its string syntax
+     * supports some <a
      * href="https://checkerframework.org/manual/#java-expressions-as-arguments">extensions</a>,
-     * notably {@code <self>} and {@code #1} for formal parameters.
+     * notably {@code <self>} and {@code #1} for formal parameters. (There are not special
```
wmdietl, line 455: From @smillst comment https://github.com/typetools/checker-framework/pull/3473#discussion_r454512286 it sounds like `<self>` isn't handled here at all. So we probably should instead highlight such checker-specific extensions?

wmdietl, line 455: Should the `not` be `no`?

docs/manual/advanced-features.tex
```diff
@@ -1140,7 +1140,8 @@
 \item \refqualclass{checker/lock/qual}{Holding}
 \end{itemize}
 
-The set of permitted expressions is a subset of all Java expressions:
+The set of permitted expressions is a subset of all Java expressions,
+with a few extensions, notably \code{<self>} and formal parameters like \<\#1>.
```
wmdietl, line 1144: Same concern about `<self>` as above.

docs/manual/advanced-features.tex
```diff
@@ -1246,20 +1247,12 @@
...
-% are variable references, but these expressions are not.
-\begin{itemize}
-\item String concatenation expressions.
-\item Mathematical operators (plus, minus, division, ...).
-\item Comparisons (equality, less than, etc.).
-\end{itemize}
-
-Additionally, it is not possible to write
-quantification over all array components (e.g. to express that all
+It is not possible to write a
```
wmdietl, line 1250: Aren't at least comparisons still forbidden? Or is this partially addressing #3474?

docs/manual/advanced-features.tex
mernst: It is important to include `<self>` here, because this section is about the Checker Framework Java expression syntax.

docs/manual/advanced-features.tex
mernst: Binary operations are possible, so I'm not positive that comparisons are explicitly forbidden.  Since part of the list became out of date, I removed it so that the rest does not get out of date.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/FlowExpressions.java
mernst: I will remove the mention from here, since it is specific to Checker Framework syntax.

docs/manual/advanced-features.tex
wmdietl: But it is only the Lock Checker that supports the `<self>` expression: https://github.com/typetools/checker-framework/blob/master/checker/src/main/java/org/checkerframework/checker/lock/LockVisitor.java#L78
So it seems rather confusing to include it like it is a generally supported expression.

docs/manual/advanced-features.tex
wmdietl: Ok, so let's leave it for #3474 to add proper restrictions again.

docs/manual/advanced-features.tex
mernst: I added a qualifier to its description.

PR #3476, opened 2020-07-14
Add predicate support to accumulation checker

kelloggm:
Differences from predicate support in the object construction checker:
* sensible LUB/GLB between predicate annotations
* subtyping between predicates does not use implication (because I didn't want a dependency on an SMT solver), so it conservatively underapproximates
* unicode support in identifiers
* no dependency on an external parsing library other than JavaParser, which the CF already depends on


kelloggm:
(sorry for the messy commit history)

framework/tests/accumulation/SimpleInferenceMerge.java
```diff
@@ -0,0 +1,37 @@
+import testaccumulation.qual.*;
+
+class SimpleInferenceMerge {
```
msridhar, line 3: This test case is unrelated to predicates, right?

docs/manual/accumulation-checker.tex
```diff
@@ -55,15 +55,21 @@
 \href{https://github.com/typetools/checker-framework/blob/master/framework/src/test/java/testaccumulation/qual/TestAccumulationBottom.java}{TestAccumulationBottom.java}.
 It should take no arguments, and should be a subtype of the accumulator type you defined earlier.
 
+You may also choose to define a predicate annotation, analogous to
+\href{https://github.com/typetools/checker-framework/blob/master/framework/src/test/java/testaccumulation/qual/TestAccumulationPredicate.java}{TestAccumulationPredicate.java}.
+A predicate annotation allows you to write specifications with disjunctions (the accumulation annotation implicitly already permits
```
msridhar, line 60: We probably want some documentation of the unsound handling of negation

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationVisitor.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    /** Checks each predicate annotation to make sure the predicate is well-formed. */
+    @Override
+    public Void visitAnnotation(final AnnotationTree node, final Void p) {
+        AnnotationMirror anno = TreeUtils.annotationFromAnnotationTree(node);
+        if (atypeFactory.isPredicate(anno)) {
+            String errorMessage = atypeFactory.isValidPredicate(anno);
+            if (errorMessage != null) {
+                checker.report(
```
mernst, line 33: Use `reportError` rather than `report`.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationVisitor.java
```diff
@@ -0,0 +1,41 @@
...
+                        node,
+                        new DiagMessage(Diagnostic.Kind.ERROR, "predicate.invalid", errorMessage));
+                return null;
```
mernst, line 36: Why not give the superclass the opportunity to issue more errors, if there are others?

framework/tests/accumulation/SimpleInferenceMerge.java
kelloggm: Yeah. It probably should have been included in the original AccumulationChecker PR, but was left off somehow. I included it here since it passes and I moved over a bunch of the other tests, so it didn't seem like it needed its own.

kelloggm:
@mernst I have addressed or responded to all of your comments. Thanks for being patient.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
```diff
@@ -58,38 +73,70 @@
...
+                rejectMalformedPredicate("have an element of type String");
+            }
+        }
+
         this.bottom = AnnotationBuilder.fromClass(elements, bottom);
         this.top = createAccumulatorAnnotation(Collections.emptyList());
 
-        // Every subclass must call postInit!  This does not do so for subclasses.
-        if (this.getClass() == AccumulationAnnotatedTypeFactory.class) {
-            this.postInit();
```
msridhar, line 91: Wall calling `postInit()` here before a bug?

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
```diff
@@ -248,6 +320,21 @@ public QualifierHierarchy createQualifierHierarchy(MultiGraphFactory factory) {
...
+     *
+     * Predicate subtyping is defined as follows:
+     *
+     * <ul>
+     *   <li>An accumulator is a subtype of a predicate if substitution from the accumulator to the
+     *       predicate makes the predicate true. For example, {@code Acc(A)} is a subtype of {@code
+     *       AccPred("A || B")}, because when A is replaced with {@code true} and B is replaced with
+     *       {@code false}, the resulting boolean formula evaluates to true.
+     *   <li>A predicate P is a subtype of an accumulator iff after converting the accumulator into
+     *       a predicate representing the conjunction of its elements, P is a subtype of that
```
msridhar, line 332: Does this mean that after converting the accumulator to a predicate, `P` must _equal_ that predicate (i.e., it falls into the next case of predicate subtyping)?

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
```diff
@@ -248,6 +320,21 @@ public QualifierHierarchy createQualifierHierarchy(MultiGraphFactory factory) {
      *        |
      *      bottom
      * </pre>
+     *
+     * Predicate subtyping is defined as follows:
```
msridhar, line 324: It might be good to document what the "right thing" to do would be, and why we deviate from that.  I think the "right thing" here is logical implication, i.e., P is a subtype of Q iff P implies Q.  But we don't bake in the implication check to keep things simple and minimize dependencies.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
kelloggm: The condition was always false, because this is an abstract class. All subclasses have to call `postInit`.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
kelloggm: In the default implementation, yes, *but* if the accumulation checker writer has implemented a more complex predicate subtyping rule that would also be taken into account.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
kelloggm: I'll add a comment.

PR #3478, opened 2020-07-15
StubParser documentation

mernst:


framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -85,20 +86,18 @@
 import org.checkerframework.javacutil.Pair;
 
 /**
- * Given a stub file, yields the annotated types in it and the declaration annotations in it. The
- * main entry point is {@link StubParser#parse(String, InputStream, AnnotatedTypeFactory,
- * ProcessingEnvironment, Map, Map)}, which side-effects its last two arguments.
- *
- * <p>The constructor acts in two parts. First, it calls the Stub Parser to parse a stub file. Then,
+ * This class represents a single stub file, notably its annotated types and its declaration
```
smillst, line 89: From the users (of this class) perspective, this class has two static methods that parse a stub file and adds annotations to two maps passed as arguments to the methods.  (The other method is StubParser#parseJdkFileAsStub.)

PR #3479, opened 2020-07-15
Signedness Checker: Fix local variable defaults for boxed integral types

smillst:


checker/src/main/java/org/checkerframework/checker/signedness/SignednessAnnotatedTypeFactory.java
```diff
@@ -163,11 +165,29 @@ private void addSignednessGlbAnnotation(Tree tree, AnnotatedTypeMirror type) {
...
+     * @param tree some tree
+     * @param type the tree's type; this method adds {@code @UnknownSignedness} to it
      */
     private void addUnknownSignednessToSomeLocals(Tree tree, AnnotatedTypeMirror type) {
         switch (type.getKind()) {
+            case DECLARED:
+                String qualifiedName =
+                        TypesUtils.getQualifiedName((DeclaredType) type.getUnderlyingType())
+                                .toString();
+                if (qualifiedName.equals("java.lang.Byte")
```
wmdietl, line 181: We have `isBoxedPrimitive` https://github.com/typetools/checker-framework/blob/master/javacutil/src/main/java/org/checkerframework/javacutil/TypesUtils.java#L102
and `isIntegral` https://github.com/typetools/checker-framework/blob/master/javacutil/src/main/java/org/checkerframework/javacutil/TypesUtils.java#L238.
The logic here is `isBoxedIntegral`, which we could add to TypesUtils.

smillst: It's not isBoxedIntegral because it excludes `char`.

smillst:
I'm going to fix this problem another way.

Issue #762, opened 2016-05-31
Refine the operand of instanceof and possibly issue a warning


smillst:
Similar to casts, the type of the operand of an instanceof should be refined to the reference type.  If the reference type is a subtype (but not equal to) the operand type, then a warning issued.  Also, reference types in the instance of should be defaulted to top so that a warning is not issued if the reference type is not annotated.

For example:

```
@A Object o = ...;
if (o instanceof @B String) {
// o: @B Object
```

if `@B <: @A`, then a warning should be issued.

smillst:
Note, defaulting the reference type to top may be difficult because in the current implementation the code that adds defaulted annotations does not know if the type is the reference type of an instanceof.

smillst:
An old implementation of refining the operands can be found here cbb21cca3a704eff7f2ff156ef4cece0ec0278bb.

PR #3481, opened 2020-07-19
Validate instanceof types (Fixes #3033 and #762)

### LINKED ISSUE #3033: Don't allow `instanceof @Nullable C`
`instanceof` is only true for a non-null expression https://docs.oracle.com/javase/specs/jls/se13/html/jls-15.html#jls-15.20.2.

Code like this:

```java
if (nbl instanceof @Nullable C)
```

is misleading: the then-branch is not taken for a null value.

A `@Nullable` annotation on the type of an `instanceof` should raise an error.
Having a `@NonNull` annotation is redundant, and could cause a warning.

aditya3434:
This PR is a fix to issue#3033 (https://github.com/typetools/checker-framework/issues/3033) and issue#762 (https://github.com/typetools/checker-framework/issues/762), regarding specific type rules for the instanceof tree. In case of the Nullness Checker, the added code will issue an error if the type of the instanceof is explicitly annotated as @Nullable and will issue a warning if it is explicitly annotated as @NonNull. For other checkers, the Framework would issue a warning if the reference type is a subtype of the expression type.

Fixes #3033 and #762.


checker/src/main/java/org/checkerframework/checker/nullness/NullnessVisitor.java
```diff
@@ -380,6 +381,20 @@ public Void visitIf(IfTree node, Void p) {
         return super.visitIf(node, p);
     }
 
+    @Override
+    public Void visitInstanceOf(InstanceOfTree node, Void p) {
+        if (node.getType().getKind() == Kind.ANNOTATED_TYPE) {
+            AnnotatedTypeMirror type = atypeFactory.fromExpression((ExpressionTree) node.getType());
```
wmdietl, line 387: Could you:
- cast to `AnnotatedTypeTree`
- call `TreeUtils.annotationsFromTree()`
- use `AnnotationUtils.containsSame` to look for the annotations

Then you don't need to change the visibility below.

aditya3434: Thanks for the help! I have used the above methods and reset the visibility of ```fromExpression``` to ```private``` in AnnotatedTypeFactory.

aditya3434:
> Can you add a few test cases, for cases without message, cases with warnings, and cases with errors?

I added a Issue3033.java test case file that contains all the possible error and warning cases in my latest commit

checker/src/main/java/org/checkerframework/checker/nullness/messages.properties
```diff
@@ -16,3 +16,5 @@ new.class.type.invalid=the annotations %s do not need be applied in object creat
 toArray.nullable.elements.not.newarray=call of toArray on collection of non-null elements yields an array of possibly-null elements; omit the argument to toArray or make it an explicit array constructor
 toArray.nullable.elements.mismatched.size=call of toArray on collection of non-null elements yields an array of possibly-null elements; cannot determine that the argument array has the same size as the receiver collection
 clear.system.property=call might clear a predefined system property; pass -Alint=permitClearProperty to permit it
+instanceof.nullable.error=instanceof is only true for a non-null expression
```
mernst, line 19: ".error" makes this key longer without making it more informative.  I suggest shortening the key.

framework/src/main/java/org/checkerframework/common/basetype/messages.properties
```diff
@@ -111,3 +111,4 @@ field.invariant.not.found.superclass=the field invariant annotation is missing f
 field.invariant.not.subtype.superclass=the qualifier for field %s is not a subtype of the qualifier in the superclass field invariant%nfound: %s%nsuperclass type: %s
 
 invalid.annotation.location.bytecode=found annotation in unexpected location in bytecode on element: %s %nUse -AignoreInvalidAnnotationLocations to suppress this warning
+operand.instanceof.subtype=reference type is a subtype of the operand type.%nOperand Type : %s%nReference Type : %s
```
mernst, line 114: I find this error message confusing.  The only point of performing `x instanceof C` is when `C` is a subtype of `x`'s compile-time type, so it's weird for the message to say only that.

I think this message is trying to say that, at run time, `instanceof` will not check the annotations.  Please make it do so.  That is similar to the "cast.unsafe" message earlier in the file.  You could make the error key "instanceof.unsafe" for parallelism.

Also, please use standard terms in the diagnostic message.  The JLS uses "expression" where you use "operand"

mernst (approved):
Thanks for this improvement!  It's great to have those issues fixed.

By the way, thanks for addressing my comments.  I had not noticed that you did so.  As a general rule, when you have addressed all comemnts, please ping for another round of review.

PR #3490, opened 2020-07-22
Fix issue 3482: Add method and class tree to CFGLambda, handle lambda case in DOTCFGVisualizer to avoid crash.

### LINKED ISSUE #3482: CFG visualizer crashes when creating file name for lambda
When running the CFG visualizer, I am encountering an error:
[`error: Unexpected AST kind: LAMBDA value`](https://github.com/typetools/checker-framework/blob/master/dataflow/src/main/java/org/checkerframework/dataflow/cfg/DOTCFGVisualizer.java#L197).
In method [`dotOutputFileName`](https://github.com/typetools/checker-framework/blob/master/dataflow/src/main/java/org/checkerframework/dataflow/cfg/DOTCFGVisualizer.java#L161), there are three possibilities for `ast.getKind()`.  The code handles `UnderlyingAST.Kind.ARBITRARY_CODE` and `UnderlyingAST.Kind.METHOD` but not `UnderlyingAST.Kind.LAMBDA`.

xingweitian:
Add method and class tree to `CFGLambda`, so that we can get the enclosing method and class name in `DOTCFGVisualizer`.
Handle lambda case in `dotOutputFileName(UnderlyingAST ast)` to avoid crash.

Resolves #3482


framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1172,10 +1172,17 @@ protected void performFlowAnalysis(ClassTree classTree) {
...
+                            TreeUtils.enclosingOfKind(getPath(lambdaPair.first), Kind.METHOD);
+                    Tree enclosingClass =
+                            TreeUtils.enclosingOfKind(getPath(enclosingMethod), Kind.CLASS);
                     analyze(
                             queue,
                             lambdaQueue,
-                            new CFGLambda(lambdaPair.first),
+                            new CFGLambda(
+                                    lambdaPair.first,
+                                    (ClassTree) enclosingClass,
```
mernst, line 1184: Nit: I prefer giving the variables the more specific type rather than casting at the use -- that is, move these casts to lines 1175-1178.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/DOTCFGVisualizer.java
```diff
@@ -193,6 +194,28 @@ protected String dotOutputFileName(UnderlyingAST ast) {
...
+            outFile.append(hashCode);
+
+            srcLoc.append("<");
+            srcLoc.append(enclosingClassName);
+            srcLoc.append("::");
+            srcLoc.append(enclosingMethodName);
+            srcLoc.append("(");
+            srcLoc.append(cfgLambda.getEnclosingMethod().getParameters());
+            srcLoc.append(")::");
+            srcLoc.append(hashCode);
```
wmdietl, line 215: Maybe the hashCode isn't necessary here - in ARBITRARY_CODE it's also not included.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/UnderlyingAST.java
```diff
@@ -75,22 +75,62 @@ public String toString() {
     /** If the underlying AST is a lambda. */
     public static class CFGLambda extends UnderlyingAST {
 
+        /** The lambda expression. */
         private final LambdaExpressionTree lambda;
 
-        public CFGLambda(LambdaExpressionTree lambda) {
+        /** The enclosing class of the lambda. */
+        private final ClassTree enclosingClass;
```
wmdietl, line 82: I would use the same field and method names as in CFGMethod.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1172,10 +1172,17 @@ protected void performFlowAnalysis(ClassTree classTree) {
 
                 while (!lambdaQueue.isEmpty()) {
                     Pair<LambdaExpressionTree, Store> lambdaPair = lambdaQueue.poll();
+                    MethodTree enclosingMethod =
+                            (MethodTree)
+                                    TreeUtils.enclosingOfKind(
+                                            getPath(lambdaPair.first), Kind.METHOD);
+                    ClassTree enclosingClass =
```
wmdietl, line 1179: I'm wondering whether this could just be `classTree`.
(I'm also wondering what the difference between `ct` and `classTree` is and whether this code should use `ct` instead...)

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1172,10 +1172,17 @@ protected void performFlowAnalysis(ClassTree classTree) {
 
                 while (!lambdaQueue.isEmpty()) {
                     Pair<LambdaExpressionTree, Store> lambdaPair = lambdaQueue.poll();
+                    MethodTree enclosingMethod =
```
wmdietl, line 1175: Possible alternative to think about: change `Queue<Pair<LambdaExpressionTree, Store>> lambdaQueue` to also contain the MethodTree within which the lambda occurred.

wmdietl:

dataflow/src/main/java/org/checkerframework/dataflow/cfg/DOTCFGVisualizer.java
```diff
@@ -193,6 +194,28 @@ protected String dotOutputFileName(UnderlyingAST ast) {
             srcLoc.append(")::");
             srcLoc.append(((JCTree) cfgMethod.getMethod()).pos);
```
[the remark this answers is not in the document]
xingweitian: Get it. I will have another look.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
xingweitian: I am not quite sure which one is better... It seems that we still need to compute the enclosing method if we change `lambdaQueue`. Does the current solution good enough?

mernst:
@xingweitian Please make the pull request title be descriptive (not just "fix issue 3482", though it should include text like "fixes #3482").  Thanks!

xingweitian:
> @xingweitian Please make the pull request title be descriptive (not just "fix issue 3482", though it should include text like "fixes #3482"). Thanks!

Thanks for pointing it out! Already changed.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/DOTCFGVisualizer.java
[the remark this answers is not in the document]
xingweitian: I have opened a new PR to solve this: https://github.com/typetools/checker-framework/pull/3521

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1172,10 +1172,14 @@ protected void performFlowAnalysis(ClassTree classTree) {
...
                     Pair<LambdaExpressionTree, Store> lambdaPair = lambdaQueue.poll();
+                    MethodTree mt =
+                            (MethodTree)
+                                    TreeUtils.enclosingOfKind(
+                                            getPath(lambdaPair.first), Kind.METHOD);
                     analyze(
                             queue,
                             lambdaQueue,
-                            new CFGLambda(lambdaPair.first),
+                            new CFGLambda(lambdaPair.first, ct, mt),
```
wmdietl, line 1182: Can you use `classTree` instead of `ct`? And open an issue that we should look into which one of the two we should be using here. It is definitely a bit odd to use both within the same expression.

PR #3503, opened 2020-07-24
Improve defaults for local variables with a qualifier parameter

jwaataja:
This pull request fixes the following issue. Consider this class,
```java

public class QualParamLocals {
    public void method(@Untainted C param) {
        C local = param;
        requiresUntainted(local);
    }

    public void requiresUntainted(@Untainted C param) {
    }

    static class C {
    }
}
```
Currently, the framework produces two errors for this code:
```
QualParamLocals.java:6: error: [assignment.type.incompatible] incompatible types in assignment.
        C local = param;
                  ^
  found   : @Untainted C
  required: @Tainted C
QualParamLocals.java:7: error: [argument.type.incompatible] incompatible types in argument.
        requiresUntainted(local);
                          ^
  found   : @Tainted C
  required: @Untainted C
2 errors
```

This shows that dataflow doesn't work as expected in the presence of qualifier parameters. There are two issues here. The first is that assignments are checked using the declared type of the left hand side, which will fail whenever the left-hand side is unnanotated (defaulting to top), and the right-hand side is not top. This causes the first error. The second error is that though dataflow can determine that `local` should have type `@Untainted`, dataflow prevents using that refined value because it compares it to the declared type (which by default has top), and rejects the refined value because it's not a subtype.

As a result, local variables must be manually annotated, which increases the annotation burden on programmers. In my experience with the determinism checker, which heavily uses qualifier parameters, this issue has been a nuisance.

This resolves these issues by changing the defaulting for local variables with a qualifier parameter. Instead of defaulting to top, local variables with an initialializer default to the type of the right hand side of the assignment.


framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -162,6 +162,29 @@
...
+     */
+    private Set<Element> variablesUnderInitialization;
+
+    /**
+     * Caches types of initializers for local variables with a qualifier parameter, so that they
+     * aren't computed each time the type of a variable is looked up.
+     *
+     * @see GenericAnnotatedTypeFactory#applyLocalVariableQualifierParameterDefaults
+     */
+    private Map<Tree, AnnotatedTypeMirror> initializerCache;
```
smillst, line 186: Did you have to add this cache so that the Daikon test would not time-out?  We have avoided caching expressions because dataflow can change the type of an expression; see https://github.com/typetools/checker-framework/issues/602.  I don't think this cache makes the problem worse than it already is though.

docs/manual/generics.tex
```diff
@@ -950,6 +950,40 @@
...
+default \<local> has type \<@Tainted StringBuffer>, and
+\<@Untainted StringBuffer> is not a subtype, no type refinement would be
+performed, leading to an error. Fixing this would require manually annotating
+\<local> as an \<@Untainted StringBuffer>, increasing the annotation burden on
+programmers.
+
+For this reason, local variables with types that have a qualifier parameter use
+different defaulting rules. When a local variable has an initializer, the type
+of that initializer is used as the default type of that variable if no other
+types are written. For example, in the above code, the type of \<local> would be
```
smillst, line 984: `types are written` should be `annotations are written`.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
mernst: Yes, the Daikon test timed out before Jason added this cache.

jwaataja: I actually added two optimizations. The first was to not calculate the type of the initializer if the type of the local variable it's assigned to has no qualifier parameter in any hierarchy. Previously, it calculated the type of the initializer for every local variable and then filtered to retain only the annotations in hierarchies with a qualifier parameter. The second  optimization was to add a cache.

The first fix alone would have been sufficient to fix the CI because it's likely that no types in the daikon test use a qualifier parameter. However, it indicates that there's some performance penalty for this check when used. Retaining the cache would probably be helpful for situations when writing `@HasQualifierParameter` on a package, which applies it to all classes in the package.

smillst: Thanks for the explanation!

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1617,6 +1652,8 @@ protected void applyQualifierParameterDefaults(
                 return;
         }
 
+        applyLocalVariableQualifierParameterDefaults(elt, type);
```
smillst, line 1655: Move this call to line 1666 and pass in `tops` rather than recomputing it.   Then in `applyLocalVariableQualifierParameterDefaults` you don't need to check that type has a qualifier parameter.

jwaataja: The `tops` variable seems to store qualifier parameter hierarchies for the enclosing class of the local variable, rather than on the type of the local variable itself. So in
```java
ClassWithQualParam1 {
    void m() {
        ClassWithQualParam2 local;
    }
}
```
We would need to apply the defaults to `local` regardless of the qualifier parameters on ClassWithQualParam1. I think this means moving it might be incorrect, because the enclosing class might not have a qualifier parameter at all and we would still apply the new defaults.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -162,6 +162,29 @@
...
+     * <p>Local variables with a qualifier parameter get their declared type from the type of their
+     * initializer. Sometimes the initializer's type depends on the type of the variable, such as
+     * during type variable inference or when a variable is used in its own initializer as in
+     * "Object o = (o = null)". This creates a circular dependency resulting in infinite recursion.
+     * To prevent this, variables in this set should not be typed based on their initializer, but by
+     * using normal defaults.
+     *
+     * @see GenericAnnotatedTypeFactory#applyLocalVariableQualifierParameterDefaults
+     */
+    private Set<Element> variablesUnderInitialization;
```
wmdietl, line 178: Can you use the more precise `VariableElement`?

checker/tests/tainting/HasQualParamDefaults.java
```diff
@@ -44,6 +44,21 @@ public Buffer append(@PolyTainted String s) {
...
         }
+
+        void initializeLocalTainted(@Tainted Buffer b) {
+            Buffer local = b;
+            @Tainted Buffer copy = local;
+        }
+
+        void initializeLocalUntainted(@Untainted Buffer b) {
+            Buffer local = b;
+            @Untainted Buffer copy = local;
```
wmdietl, line 55: How about adding a test with a local variable without initializer, showing how that will produce an expected error when the default and the RHS-type mismatch?

wmdietl: More generally, maybe a test with an expected error would be good, to show that the expected type is inferred.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1588,6 +1615,10 @@ protected void applyInferredAnnotations(AnnotatedTypeMirror type, Value as) {
     /**
      * Applies defaults for types in a class with an qualifier parameter.
      *
+     * <p>Within a class with {@code @HasQualifierParameter}, types with that class default to the
+     * polymorphic qualifier rather than the typical default. Local variables with a type that has a
+     * qualifier parameter are initialized to the type of their initializer, rather than top.
```
wmdietl, line 1620: Maybe instead of `top` say the default for local variables, as a type system could change that default.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1598,6 +1629,10 @@ protected void applyQualifierParameterDefaults(Tree tree, AnnotatedTypeMirror ty
...
```
wmdietl, line 1634: Same comment about top.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1645,6 +1682,59 @@ public Void visitDeclared(AnnotatedDeclaredType type, Void aVoid) {
         }.visit(type);
     }
 
+    /**
+     * Defaults local variables with types that have a qualifier parameter to type of their
```
wmdietl, line 1686: "to type" should be "to the type".

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1645,6 +1682,59 @@ public Void visitDeclared(AnnotatedDeclaredType type, Void aVoid) {
...
+     * initializer.
```
wmdietl, line 1687: Add ", if an initializer is present" and explain what happens for local variables without initializers?

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1645,6 +1682,59 @@ public Void visitDeclared(AnnotatedDeclaredType type, Void aVoid) {
...
+        if (initializer == null) {
+            return;
+        }
+
+        variablesUnderInitialization.add(elt);
+        AnnotatedTypeMirror initializerType;
+        if (shouldCache && initializerCache.containsKey(initializer)) {
+            initializerType = initializerCache.get(initializer);
+        } else {
+            // When this method is call by getAnnotatedTypeLhs, flow is turned off.
```
wmdietl, line 1715: "call" -> "called".

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1645,6 +1682,59 @@ public Void visitDeclared(AnnotatedDeclaredType type, Void aVoid) {
...
+            // Turn it back on so the type of the initializer is the refined type.
+            boolean oldUseFlow = useFlow;
+            useFlow = everUseFlow;
+            initializerType = getAnnotatedType(initializer);
```
wmdietl, line 1719: We're usually careful and put such code in try/finally blocks, to ensure that the pre-state is restored even if there is an error.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -162,6 +162,29 @@
...
+     * Elements representing variables for which the type of the initializer is being determined in
+     * order to apply qualifier parameter defaults.
+     *
+     * <p>Local variables with a qualifier parameter get their declared type from the type of their
+     * initializer. Sometimes the initializer's type depends on the type of the variable, such as
+     * during type variable inference or when a variable is used in its own initializer as in
+     * "Object o = (o = null)". This creates a circular dependency resulting in infinite recursion.
+     * To prevent this, variables in this set should not be typed based on their initializer, but by
+     * using normal defaults.
+     *
```
wmdietl, line 175: Maybe add a comment why this set doesn't need to be cleared out: it's expected to be empty outside an variable initializer, right?

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
jwaataja: The documentation on this seems sketchy, the [javadoc](https://docs.oracle.com/javase/8/docs/api/javax/lang/model/element/ElementKind.html#LOCAL_VARIABLE) for this doesn't seem guarantee that `e.getKind() == ElementKind.LOCAL_VARIABLE` implies `e instanceof VariableElement`. However, the only line in the jdk source code `return ElementKind.LOCAL_VARIABLE` is within a `VariableElement` implementation, so I think your suggestion makes sense.

PR #3506, opened 2020-07-24
Improve DefaultTypeHierarchy documentation

mernst:


framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java
```diff
@@ -281,17 +284,22 @@ protected boolean isBottom(final AnnotatedTypeMirror subtype) {
     }
 
     /**
-     * Check and subtype first determines if the subtype/supertype combination has already been
-     * visited. If so, it returns true, otherwise add the subtype/supertype combination and then
-     * make a subtype check
+     * Like {@link #isSubtype(AnnotatedTypeMirror, AnnotatedTypeMirror)}, but uses a cache for
+     * performance.
```
smillst, line 288: This cache is to prevent infinite recursion on recursive types.

PR #3508, opened 2020-07-24
Make @DefaultFor and @DefaultQualifierForUse consistent.

smillst:
The implementation of `@DefaultFor`, a meta-annotation, and `@DefaultForUse`, a class declaration annotation is inconsistent with [the manual](https://checkerframework.org/manual/#defaults).  This pull request fixes that.
  
1. Writing `@DefaultFor(types=MyClass.class)`  as a meta-annotation for annotation `A` should be the same as writing `@DefaultQualifierForUse(A.class)`  on the declaration of MyClass.
2. Defaults specified by `@DefaultFor(types=...)`, `@DefaultFor(typeKinds=...)`, or `@DefaulQualifierForUse` do not apply to local variables. This is currently implemented for local variables with declared types, but not primitive types.
3. These defaults apply to types, not expressions.

Casts and new arrays: 
The qualifier for unannotated casts and new array expressions is "propagated", quasi-inference, in the `PropagationTreeAnnotator`. The "propagated" qualifier should not be above the upper bound qualifier for the type.  This fixes a bug that was uncovered by the other changes in this pull request.

The changes to the Signedness Checker tests the changes in this pull request.  The Lock Checker and the Nullness Checker both needed to adapt to theses changes.

Merge with https://github.com/typetools/checker-framework-inference/pull/166.

(@mernst this includes the changes to the Signedness Checker from your `signedness-upperboundfor` branch. I think you can delete it now.)


framework/src/main/java/org/checkerframework/common/basetype/BaseTypeValidator.java
```diff
@@ -95,7 +95,7 @@ public boolean isValid(AnnotatedTypeMirror type, Tree tree) {
      * @return whether or not the top-level type should be checked
      */
     protected boolean shouldCheckTopLevelDeclaredType(AnnotatedTypeMirror type, Tree tree) {
-        if (type.getKind() != TypeKind.DECLARED) {
+        if (type.getKind() != TypeKind.DECLARED && !type.getKind().isPrimitive()) {
```
wmdietl, line 98: Could you clarify the javadoc? It's confusing given the method name that this applies to anything other than declared types.
Can you call out explicitly what happens with ARRAY (and what else? intersection?)

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1513,9 +1527,15 @@ protected void addComputedTypeAnnotations(
                 : "GenericAnnotatedTypeFactory.addComputedTypeAnnotations: "
                         + " root needs to be set when used on trees; factory: "
                         + this.getClass();
-        addAnnotationsFromDefaultQualifierForUse(TreeUtils.elementFromTree(tree), type);
+
+        if (!TreeUtils.isExpressionTree(tree)) {
+            addAnnotationsFromDefaultQualifierForUse(TreeUtils.elementFromTree(tree), type);
```
wmdietl, line 1532: Can you add a comment why the order is different for expressions?

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1839,20 +1859,28 @@ public void postAsMemberOf(
      */
```
wmdietl, line 1859: Could you update the javadoc?
This now also applies the `defaultForTypeAnnotator`, which is surprising given the method name.
It looks like `null` is passed as element to change what would happen by default, so calling that out would be good.
I was also surprised by the special case for primitives, so giving a reason for that would be good.

javacutil/src/main/java/org/checkerframework/javacutil/TypesUtils.java
```diff
@@ -248,9 +248,25 @@ public static boolean isIntegral(TypeMirror type) {
         }
     }
 
+    /**
+     * Returns true iff the argument is a boxed floating point type.
+     *
+     * @param type type mirror
```
wmdietl, line 254: I would prefer something more descriptive than the type, although "type to test" is also not much better.

javacutil/src/main/java/org/checkerframework/javacutil/TypesUtils.java
```diff
@@ -248,9 +248,25 @@ public static boolean isIntegral(TypeMirror type) {
...
+     * @return whether the argument is a boxed floating point type
+     */
+    public static boolean isBoxedFloating(TypeMirror type) {
+        if (type.getKind() != TypeKind.DECLARED) {
+            return false;
+        }
+
+        String qualifiedName = getQualifiedName((DeclaredType) type).toString();
```
wmdietl, line 262: At some point we should figure out and document what the proper way is. `Name` defines a `contentEquals`, which might be more direct than calling `toString`. However, for multiple comparisons one approach might be more efficient.

checker/src/main/java/org/checkerframework/checker/nullness/NullnessVisitor.java
```diff
@@ -648,4 +652,38 @@ public Void visitAnnotation(AnnotationTree node, Void p) {
...
+    }
+
+    /**
+     * Nullness Validator. Check that primitive types are annotated with {@code @NonNull} even if
+     * they are the type of a local variable.
+     */
+    private static class NullnessValidator extends BaseTypeValidator {
+
+        /**
+         * Create Nullness validator.
```
wmdietl, line 668: Nit. Make the `validator` capitalization consistent with the use in the class javadoc.

checker/tests/signedness/CastedShifts.java
```diff
@@ -36,35 +36,25 @@ public void CastedIntShifts(@Unsigned int unsigned, @Signed int signed) {
...
 
         // Cast to char.
-        @UnknownSignedness char charRes;
+        char charRes;
 
         // Shifting right by 23, the introduced bits are cast away
         charRes = (@Unsigned char) (unsigned >>> 23);
         charRes = (@Unsigned char) (unsigned >> 23);
-        charRes = (@Signed char) (signed >>> 23);
-        charRes = (@Signed char) (signed >> 23);
```
wmdietl, line 45: It looks like all the signed char casts are removed from this test. Should at least one remain and expect an error? Or what is supposed to happen if somebody writes that?

mernst: It is an error to write `@Signed char`.  We'll add a test later, in pull requests after this one.

checker/src/main/java/org/checkerframework/checker/nullness/NullnessVisitor.java
mernst: I think it's OK to write "NullnessValidator" or "Nullness validator".  I wouldn't write "Nullness Validator", though "Nullness BaseTypeValidator" would also be OK.

javacutil/src/main/java/org/checkerframework/javacutil/TypesUtils.java
smillst: That does sound like a good future improvement.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeValidator.java
```diff
@@ -86,16 +86,19 @@ public boolean isValid(AnnotatedTypeMirror type, Tree tree) {
...
      *
      * <p>Top-level type is not checked if tree is a local variable or an expression tree.
      *
      * @param type AnnotatedTypeMirror being validated
      * @param tree a Tree whose type is {@code type}
-     * @return whether or not the top-level type should be checked
+     * @return whether or not the top-level type should be checked, if {@code type} is a declared or
+     *     primitive type.
      */
     protected boolean shouldCheckTopLevelDeclaredType(AnnotatedTypeMirror type, Tree tree) {
```
wmdietl, line 100: Could you deprecate this method and introduce a better name?
It makes no sense to use `DeclaredType` in the method name if it actually is also for primitive types.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1841,25 +1868,37 @@ public void postAsMemberOf(
      * Adds default qualifiers bases on the underlying type of {@code type} to {@code type}. If
      * {@code element} is a local variable, then the defaults are not added.
      *
+     * <p>(This uses both the {@link DefaultQualifierForUseTypeAnnotator} and {@link
+     * DefaultForTypeAnnotator}.)
+     *
      * @param element possibly null element whose type is {@code type}
      * @param type the type to which defaults are added
      */
+    // TODO: rename in the next minor release.
```
wmdietl, line 1877: The August release will be the next minor release. So isn't now the right time to make this change?

framework/src/main/java/org/checkerframework/framework/type/treeannotator/PropagationTreeAnnotator.java
```diff
@@ -204,4 +209,27 @@ private boolean hasPrimaryAnnotationInAllHierarchies(AnnotatedTypeMirror type) {
...
+    /**
+     * Adds the qualifiers in {@code annos} to {@code type} that are below the qualifier upper bound
+     * of type and for which type does not already have annotation in the same hierarchy. If a
+     * qualifier in {@code annos} is above the bound, then the bound is added to {@code type}
+     * instead.
+     *
+     * @param type annotations are added to this type
+     * @param annos annotations to add to type
+     */
+    private void addAnnoOrBound(AnnotatedTypeMirror type, Set<? extends AnnotationMirror> annos) {
```
wmdietl, line 222: In other recent pull requests you used `quals` as the name for such qualifier sets.
Maybe `addQualOrBound` would also be a better method name?

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeValidator.java
smillst: We can't deprecate methods the might be overridden by subclass.  I can rename it and mention it in the change log.

framework/src/main/java/org/checkerframework/framework/type/treeannotator/PropagationTreeAnnotator.java
smillst: This file uses "annotation" rather "qualifier".  So `Anno` is more consistent.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
smillst: Yes, I didn't realize breaking changes were merged.  I'll fix it.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeValidator.java
wmdietl: You can deprecate it and have the new method call the deprecated method (suppressing the warning).
All CF overrides can change to override the new method and external overrides would still be considered by the new method in this class.
But as this will be a minor release anyways, directly renaming sounds good to me.

PR #3509, opened 2020-07-24
Add a new parameter "-string" to CFGVisualizeLauncher to print the string representation of the control flow graph.

xingweitian:
In the past, the CFGVisualizeLauncher can only generate the dot representation of the cfg (You can call `CFGVisualizeLauncher.generateStringOfCFG()` but there is no entry from the command line). Now we can use this parameter to print the string representation in the terminal.


dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizeLauncher.java
```diff
@@ -116,6 +127,25 @@ protected void generateDOTofCFGWithoutAnalysis(
         generateDOTofCFG(inputFile, outputDir, method, clas, pdf, verbose, null);
     }
 
+    /**
+     * Generate the String representation of the CFG for a method without analysis.
```
mernst, line 131: What does "without analysis" mean?  It's here and in the method name.

I realize this phrase exists elsewhere in the code (it's not your invention), but I don't understand it in the other places and its documentation could be improved throughout.  Thanks!

mernst:
@xingweitian You have created a xingweitian:typetools-improve-cfgvisualizelauncher fork in checker-framework-inference.  It is behind master and I cannot push to it.  Could you pull master into it?  Then CI should pass.
(But, it doesn't contain any changes from master, so I'm not sure why it exists.  You could also delete it, and then CI will also pass.)

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizeLauncher.java
xingweitian: "without analysis" means that `analysis=null`. So `analysis.performAnalysis(cfg);` will not be executed. Only the control flow graph will be visualized.

Before this PR, you can generate the dot representation of the control flow graph by:

```
java -Xbootclasspath/p:$CHECKERFRAMEWORK/checker/dist/javac.jar \
  -cp $CHECKERFRAMEWORK/checker/dist/checker.jar \
  org.checkerframework.dataflow.cfg.CFGVisualizeLauncher \
Test.java cfg -class Test -method test
```

Then the dot files will be generated. (You can also add `-pdf` to generate pdf files.)

But there is no way to see the string representation from the command line.

After this PR (add -string at the end of the command):

```
java -Xbootclasspath/p:$CHECKERFRAMEWORK/checker/dist/javac.jar \                                                                           checker-framework -> typetools-improve-cfgvisualizelauncher $ ?
  -cp $CHECKERFRAMEWORK/checker/dist/checker.jar \
  org.checkerframework.dataflow.cfg.CFGVisualizeLauncher \
Test.java cfg -class Test -method test -string
```

A string graph will be printed:

```
2 -> 3
3 -> 4
4 -> 8
4 -> 10
8 -> 0
10 -> 0

2:
<entry>

3:
a   [ VariableDeclaration ]
1   [ IntegerLiteral ]
a = 1   [ Assignment ]
b   [ VariableDeclaration ]
2   [ IntegerLiteral ]
b = 2   [ Assignment ]
c   [ VariableDeclaration ]
3   [ IntegerLiteral ]
c = 3   [ Assignment ]
a   [ LocalVariable ]
0   [ IntegerLiteral ]
(a > 0)   [ GreaterThan ]

4:
ConditionalBlock: then: 8, else: 10

8:
d   [ VariableDeclaration ]
a   [ LocalVariable ]
c   [ LocalVariable ]
(a + c)   [ NumericalAddition ]
d = (a + c)   [ Assignment ]

10:
e   [ VariableDeclaration ]
a   [ LocalVariable ]
b   [ LocalVariable ]
(a + b)   [ NumericalAddition ]
e = (a + b)   [ Assignment ]

0:
<exit>
```

I think this may be convenient when we only want to visualize a simple program.

mernst: Thanks for the explanation.  I wrote some text for the manual (which is what I really wanted.)  Please check to see that it is accurate.

I am not convinced I would want to use this, but I don't object to providing the feature.

Could you please document it in the manual as well as in the changelog?

docs/manual/creating-a-checker.tex
```diff
@@ -1892,6 +1892,7 @@
   \<-Acfgviz=org.checkerframework.dataflow.cfg.DOTCFGVisualizer,outdir=\emph{somedir}>.
   % TODO: create the directory if it doesn't exist.
   The directory must already exist.
+  The CFG is generated and output, but no dataflow analysis is performed.
```
xingweitian, line 1895: `javac -processor myProcessor -Aflowdotdir=. MyClass.java` will do the dataflow analysis. So I think we should remove this line.

docs/manual/creating-a-checker.tex
```diff
@@ -1904,6 +1905,7 @@
   analyzed by the dataflow analysis (Section~\ref{creating-dataflow}).
   The graph also contains information about flow-sensitively refined
   types of various expressions at many program points.
+  The CFG is generated and output, but no dataflow analysis is performed.
```
xingweitian, line 1908: The same for here. `CFGVisualizeLauncher` is not used in this case.

docs/manual/creating-a-checker.tex
```diff
@@ -1926,7 +1927,7 @@
 java -Xbootclasspath/p:$CHECKERFRAMEWORK/checker/dist/javac.jar \
   -cp $CHECKERFRAMEWORK/checker/dist/checker.jar \
   org.checkerframework.dataflow.cfg.CFGVisualizeLauncher \
-  MyClass.java output/ -class MyClass -method test -pdf
+  MyClass.java -output output/ -class MyClass -method test -pdf
```
mernst, line 1930: To simplify the command, how about leaving off the `-output` command-line argument?

docs/manual/creating-a-checker.tex
```diff
@@ -1937,17 +1938,29 @@
 \begin{Verbatim}
 java -cp $CHECKERFRAMEWORK/checker/dist/checker.jar \
   org.checkerframework.dataflow.cfg.CFGVisualizeLauncher \
-  MyClass.java output/ -class MyClass -method test -pdf
+  MyClass.java -output output/ -class MyClass -method test -pdf
```
mernst, line 1941: To simplify the command, how about leaving off the `-output` command-line argument?

docs/manual/creating-a-checker.tex
```diff
@@ -1937,17 +1938,31 @@
...
 \end{smaller}
 
 \end{itemize}
 
 \noindent
 The above command will generate the corresponding dot and pdf files for the
-method \code{test} in the class \code{MyClass} in the directory \<output>. For more
-information, run \refclass{dataflow/cfg}{CFGVisualizeLauncher} with no arguments to
-see the usage.
+method \code{test} in the class \code{MyClass} in the directory \<output>.
```
mernst, line 1949: If you drop `-output`, you would need to adjust this text too.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizeLauncher.java
```diff
@@ -53,12 +52,22 @@ public static void main(String[] args) {
...
+        String output = ".";
         boolean pdf = false;
         boolean error = false;
         boolean verbose = false;
+        boolean string = false;
 
-        for (int i = 2; i < args.length; i++) {
+        for (int i = 1; i < args.length; i++) {
             switch (args[i]) {
+                case "-output":
```
mernst, line 63: This name suggests a file to me.  I would call it "--outputdir".

changelog.txt
```diff
@@ -8,6 +8,10 @@ Added an overloaded version of NullnessUtil.castNonNull that takes an error mess
 
 Added a new option `-Aversion` to print the version of the Checker Framework.
 
+New CFGVisualizeLauncher command-line arguments:
+ * `-outdir`: directory in which to write output files
```
mernst, line 12: The command-line arguments to this program not follow the convention that `-` is for single-character flags and `--` is for multi-character flags?  I think that would make the program easier to follow.

I realize this isn't your fault!  It could be fixed in this pull request or in a different pull request that is merged before this one.

docs/manual/creating-a-checker.tex
xingweitian: Thanks for the comments!
If we remove `--output`, then the `outputdir` becomes a mandatory argument (`String outputdir = args[1];`). Although users want to generate a string graph, they still need to provide `outputdir` in the command. E.g., `MyClass.java output/ -class MyClass -method test -string`.

mernst: Sorry that my comment was ambiguous.  I think that the tool should support the command-line argument.  I was just suggesting that this example invocation could leave it off, using the default, to make the command line shorter and to reduce the difference between the `--string` version below.

xingweitian: Oh! I get it. Thanks a lot! I will do it.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizeLauncher.java
```diff
@@ -38,12 +38,11 @@
      */
     public static void main(String[] args) {
         CFGVisualizeLauncher cfgVisualizeLauncher = new CFGVisualizeLauncher();
-        if (args.length < 2) {
+        if (args.length < 1) {
```
wmdietl, line 41: Is `== 0` clearer?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizeLauncher.java
```diff
@@ -53,34 +52,48 @@ public static void main(String[] args) {
...
+        boolean string = false;
 
-        for (int i = 2; i < args.length; i++) {
+        for (int i = 1; i < args.length; i++) {
             switch (args[i]) {
-                case "-pdf":
+                case "--outputdir":
+                    if (i >= args.length - 1) {
+                        cfgVisualizeLauncher.printError(
+                                "Did not find <outputdir> after -outputdir.");
```
wmdietl, line 66: Use `--` for all mentions of flags.

changelog.txt
```diff
@@ -8,6 +8,10 @@ Added an overloaded version of NullnessUtil.castNonNull that takes an error mess
 
 Added a new option `-Aversion` to print the version of the Checker Framework.
 
+New CFGVisualizeLauncher command-line arguments:
+ * `--outputdir`: directory in which to write output files
+ * `--string`: print the control flow graph in the terminal
```
wmdietl, line 13: Mention that in general flags now use `--`.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizeLauncher.java
```diff
@@ -116,6 +136,26 @@ protected void generateDOTofCFGWithoutAnalysis(
         generateDOTofCFG(inputFile, outputDir, method, clas, pdf, verbose, null);
     }
 
+    /**
+     * Generate the String representation of the CFG for a method, only. Does no dataflow analysis
+     * analysis.
```
wmdietl, line 141: Duplicate `analysis`.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizeLauncher.java
```diff
@@ -116,6 +136,26 @@ protected void generateDOTofCFGWithoutAnalysis(
...
+     *
+     * @param inputFile java source input file
+     * @param method name of the method to generate the CFG for
+     * @param clas name of the class which includes the method to generate the CFG for
+     * @param verbose show verbose information in CFG
+     * @return the String representation of the CFG
+     */
+    protected @Nullable String generateStringOfCFGWithoutAnalysis(
```
wmdietl, line 149: Make non-null.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGVisualizeLauncher.java
```diff
@@ -116,6 +136,26 @@ protected void generateDOTofCFGWithoutAnalysis(
...
+            String inputFile, String method, String clas, boolean verbose) {
+        @Nullable Map<String, Object> res = generateStringOfCFG(inputFile, method, clas, verbose, null);
+        if (res != null) {
+            return (String) res.get("stringGraph");
+        } else {
+            return "null";
```
wmdietl, line 155: Return useful error message?

PR #3516, opened 2020-07-27
Visualize CFGs without package names

mernst:
This makes the CFG smaller and easier to read.  In rare cases when there are multiple classes with the same name, it might hide important information.  If that comes up, we can introduce a flag to print classes in fully-qualified form.

~Contains changes from https://github.com/typetools/checker-framework/pull/3514, which should be merged first.~


framework/src/main/java/org/checkerframework/framework/flow/CFAbstractValue.java
```diff
@@ -51,9 +55,21 @@
     /** The analysis class this value belongs to. */
     protected final CFAbstractAnalysis<V, ?, ?> analysis;
 
+    /** The underlying (Java) type in this abstract value. */
     protected final TypeMirror underlyingType;
+    /** The annotations in this abstract value. */
     protected final Set<AnnotationMirror> annotations;
 
+    /** Formats annotations without their package names. */
+    AnnotationFormatter annoFormatter = new DefaultAnnotationFormatter();
```
wmdietl, line 64: Can this field also be `protected final`?

wmdietl, line 64: I'm a bit concerned about adding this field. We create many abstract values and this will make that more costly.
We only call `toString`/`toStringSimple` when we debug/visualize the CFG, so it would overall probably be much better to only create the formatter in `toStringSimple`.
What do you think?

mernst: I made the field static.  It has no state, but since its methods are non-static, there must be an instance of AnnotationFormatter.  But that instance can be the same for all uses.

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractValue.java
```diff
@@ -48,12 +51,25 @@
  */
 public abstract class CFAbstractValue<V extends CFAbstractValue<V>> implements AbstractValue<V> {
 
+    /** Formats annotations without their package names. */
+    protected static DefaultAnnotationFormatter defaultAnnotationFormatter =
```
wmdietl, line 55: Any reason to not make it `final`?
I would probably still move this into `toStringSimple`, because that's the only place that uses this field and that method won't be used much. But I'm fine with a static field.

javacutil/src/main/java/org/checkerframework/javacutil/ElementUtils.java
```diff
@@ -215,20 +215,27 @@ public static String getSimpleName(ExecutableElement element) {
      * @param type a type
      * @return the simple type name, without annotations
      */
-    private static String simpleTypeName(TypeMirror type) {
+    public static String simpleTypeName(TypeMirror type) {
```
wmdietl, line 218: Now that this method is public, we should move it to `TypesUtils` - the parameter type is `TypeMirror`, so that's where I would expect such a method.

PR #3518, opened 2020-07-27
Clarify `@Covariant` documentation in terms of clients

mernst:
This is an attempt to clarify the documentation of `@Covariant`.  Is it a correct interpretation?


docs/manual/generics.tex
```diff
@@ -445,16 +445,17 @@
 invariantly.  For example, \code{List<@Nullable String>} is not a subtype
 of \code{List<String>}.
 
-When a type parameter is used in a read-only way --- that is, when values
-of that type are read but are never assigned --- then it is safe for the
+When a type parameter is used in a read-only way --- that is, when clients
+read values of that type from the class are never pass values of that type
```
smillst, line 449: I can't parse this. "from the class are never pass values of that type"

docs/manual/generics.tex
```diff
@@ -445,16 +445,17 @@
...
 type to be \emph{covariant} in the type parameter.  Use the
 \refqualclass{framework/qual}{Covariant} annotation to indicate this.
 When a type parameter is covariant, two instantiations of the class with
 different type arguments have the same subtyping relationship as the type
 arguments do.
 
-For example, consider \<Iterator>.  Its elements can be read but not
-written, so \code{Iterator<@Nullable String>} can be a subtype of
+For example, consider \<Iterator>.  A client can read elements but not
+write them, so \code{Iterator<@Nullable String>} can be a subtype of
```
wmdietl, line 458: @mernst This problem already existed before, but I just notice it when reading through this: isn't this the wrong way around? A `@Nullable String` is a supertype of `@NonNull String`, so it is safe for `Iterator<@Nullable String>` to be a supertype of `List<String>`.
As described, it sounds like one can assign an `Iterator<@Nullable String>` to a `Iterator<String>`, which would then allow to read a value as non-null, even though the iterator can contain nulls.
Am I misreading the example?

PR #3520, opened 2020-07-27
Refine instanceof operands (Fixes #762)

### LINKED ISSUE #762: Refine the operand of instanceof and possibly issue a warning

aditya3434:
This PR is a fix to issue#762 (#762), regarding refinement of ```instanceof``` operands to the level of the annotated type expression, along with a warning (similar to type casts).
```
@A Object o = ...;

if (o instanceof @B String) {
// o: @B Object    // Refined to @B if @B <: @A
}
```
If ```@B```  is a subtype of ```@A```, then object o will be refined and a warning will be issued.


wmdietl:
This is the PR I meant earlier today https://github.com/typetools/checker-framework/pull/3481#pullrequestreview-457780564

Why should they be two separate changes?

I'm confused by the "refine". We discussed simply performing the same checks as for the Nullness Checker more generally.

PR #3521, opened 2020-07-28
Add parameters to the output dot file's name.

xingweitian:
As https://github.com/typetools/checker-framework/pull/3490#discussion_r459773537 mentioned, to handle the overloaded methods, we also add parameters to the output dot file's name.

If we have two methods: `test(String a, String b)` and `test(String a, String b, String c)` in class `Test`, now the output dot files are:
~~`Test::test(String a, String b)-Nullness.dot` and~~
~~`Test::test(String a, String b, String c)-Nullness.dot`.~~
`Test-test-String,String-Nullness.dot` and
`Test-test-String,String,String-Nullness.dot` and


xingweitian:
Thanks for the comments! I have replaced all the reserved characters.

> As a nit, some operating systems require quoting for file names containing `<`, `(`, and ` ` (space), and certain other characters with special meaning for the shell.
> 
> With this change, the commands given in the [manual](https://checkerframework.org/manual/#creating-debugging-dataflow-graph) no longer work. I suggest changing or removing the problematic characters. If you don't do that, you'll need to edit certain parts of the manual.

Thanks for the comments! I have replaced all the reserved characters.

PR #3522, opened 2020-07-29
Routines to format AnalysisResult objects for debugging/logging

mernst:
~Incorporates changes from https://github.com/typetools/checker-framework/pull/3527, which should be merged first.~


dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java
```diff
@@ -443,4 +445,124 @@ public S getStoreAfter(Block block) {
...
+        return result.toString();
+    }
+
+    /**
+     * Return a printed representation of a node.
+     *
+     * @param n a node to format
+     * @return a printed representation of the given node
+     */
+    public static String nodeRepr(Node n) {
```
wmdietl, line 496: Any reason for not making this `Node.toString`?

dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java
```diff
@@ -443,4 +445,124 @@ public S getStoreAfter(Block block) {
...
+                n, n.getClass().getSimpleName(), n.hashCode(), System.identityHashCode(n));
+    }
+
+    /**
+     * Return a printed representation of a collection of nodes.
+     *
+     * @param nodes a collection of nodes to format
+     * @return a printed representation of the given collection
+     */
+    public static String nodeCollectionRepr(Collection<? extends Node> nodes) {
```
wmdietl, line 508: You wouldn't need this if you moved `nodeRepr` into `Node.toString`, right?

dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java
```diff
@@ -443,4 +445,124 @@ public S getStoreAfter(Block block) {
...
+        return result.toString();
+    }
+
+    /** Checks representation invariants on this. */
+    public void checkRep() {
+        // Require that each node in treeLookup exists in nodeValues.
+        for (Map.Entry<Tree, Set<Node>> entry : treeLookup.entrySet()) {
+            for (Node n : entry.getValue()) {
+                if (!nodeValues.containsKey(n)) {
+                    sleep(100);
```
wmdietl, line 546: I'm a bit confused why you would need that here. We throw `BugInCF` in many places, so if this is a general problem, should that be solved where `BugInCF` is output?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/ControlFlowGraph.java
```diff
@@ -235,7 +237,7 @@ public UnderlyingAST getUnderlyingAST() {
     }
 
     /**
-     * Returns the copied tree-lookup map.
+     * Returns the copied tree-lookup map. Ignores convertedTreeLookup.
```
wmdietl, line 240: I don't understand the new comment. The method returns a copy of that one field, like `getUnaryAssignNodeLookup` returns a copy of that field.
Is this more a to-do comment that we need a method to also access the `convertedTreeLookup`?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/ControlFlowGraph.java
```diff
@@ -303,4 +305,32 @@ public String toString() {
         String stringGraph = (String) res.get("stringGraph");
         return stringGraph == null ? super.toString() : stringGraph;
```
wmdietl, line 306: If `stringGraph` is null, you could use `repr` instead of the un-helpful super implementation.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/ControlFlowGraph.java
```diff
@@ -303,4 +305,32 @@ public String toString() {
...
     }
+
+    /**
+     * Returns a string representation of this.
+     *
+     * @return a string representation of this
+     */
+    public String repr() {
```
wmdietl, line 314: There already is a `toString` method in this class. Can you add cross-references and explain when to use which?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/ControlFlowGraph.java
mernst: The way that the tree-lookup map is accessed is via `getNodesCorrespondingToTree`.  That method makes use of both `convertedTreeLookup` and `treeLookup`.  I'm not sure why there is a need for a public method that makes a copy of `treeLookup`, but any client will be inconsistent with `getNodesCorrespondingToTree`.  So it is useful to provide a warning to clients.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java
mernst: That doesn't work.  I'm not sure why.  Without this, printf output is interspersed.  Understanding why would be useful, but is outside the scope of this pull request.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java
mernst: Yes, I agree.  However, that would make general use of `toString` excessively verbose.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java
```diff
@@ -443,4 +445,124 @@ public S getStoreAfter(Block block) {
         return transferInput.analysis.runAnalysisFor(
                 node, before, transferInput, nodeValues, analysisCaches);
     }
+
+    /**
+     * Returns a string representation of this.
+     *
+     * @return a string representation of this
+     */
+    public String repr() {
```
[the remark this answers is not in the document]
mernst: This is much more verbose.  It would clutter most output and is useful for debugging.  I have updated the description.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java
mernst: Same reason.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java
wmdietl: Can you at least add a TODO comment here to remark why it is here? This is very surprising and should be highlighted and fixed.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java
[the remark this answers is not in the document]
mernst: It is not the case that nothing useful is output.  `Object.toString` outputs the run-time class name and a unique identifier.  Users may prefer that to dozens of lines of output.  If `toString` is overridden, users don't have a way to get less information.  Creating a different method gives each user the option of how much output to put in logs.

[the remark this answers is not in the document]
mernst: Another customary difference between `toString()` and `repr()` is that `toString()` reflects the abstraction or client view, whereas `repr()` shows the implementation or representation.  (Not every implementation of `toString()` and `repr()` follows the convention perfectly.)

dataflow/src/main/java/org/checkerframework/dataflow/cfg/ControlFlowGraph.java
mernst: That's a change to `toString()`, and `repr()` is more about the internal implementation.  (Also, this pull request doesn't make any changes to `toString()`.)  Should the code instead throw `BugInCF` if `stringGraph` is null?  A separate pull request could do that.

mernst:
@wmdietl Thanks for the feedback.  I think I have addressed it (except for a change to an existing `toString()` routine that I propose to defer to a later pull request).  Could you see whether I did so?  Thanks.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java
```diff
@@ -443,4 +445,85 @@ public S getStoreAfter(Block block) {
...
+        return result.toString();
+    }
+
+    /** Checks representation invariants on this. */
+    public void checkRep() {
+        // Require that each node in treeLookup exists in nodeValues.
+        for (Map.Entry<Tree, Set<Node>> entry : treeLookup.entrySet()) {
+            for (Node n : entry.getValue()) {
+                if (!nodeValues.containsKey(n)) {
+                    SystemUtil.sleep(100); // without this, printf output is sometimes interleaved
```
wmdietl, line 522: What `printf` output is this referring to? Throwing a `BugInCF` outputs the message using the compiler messager and should happen one at a time. In general, the compiler is single-threaded and a `sleep` seems very inappropriate.

Is this maybe from running gradle concurrently?
How can I reproduce this issue?

As this `checkRep` method isn't called anywhere, how about moving this method to a separate PR?

dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java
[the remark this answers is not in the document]
wmdietl: This is the first class where we add a `repr` method, so I would like to understand what the motivation is and don't understand what you mean with "customary".
I see 483 matches for such a `repr` function on GitHub for Java code. This seems popular in Python code.

In general it seems more useful to output concrete information instead of what `Object.toString` gives. If the behavior of `Object.toString` is really desired in some place, one can simply call `getClass` and `hashCode` instead.

So far in similar classes we have overriden `toString`. What makes these classes special to not do the same?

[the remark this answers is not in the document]
mernst: > This is the first class where we add a `repr` method, so I would like to understand what the motivation is 

I added text from this conversation to the developer manual.

> In general it seems more useful to output concrete information instead of what `Object.toString` gives.

It is useful to give clients a choice about what information to output.

> So far in similar classes we have overriden `toString`. What makes these classes special to not do the same?

The difference is not the classes, the difference is the methods.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/node/Node.java
```diff
@@ -157,4 +158,32 @@ public void setAssignmentContext(AssignmentContext assignmentContext) {
         }
         return transitiveOperands;
     }
+
+    /**
+     * Returns a verbose printed representation of this, useful for debugging.
```
wmdietl, line 163: For consistency, in a few places in this PR, I would replace the `printed` with `string`.

mernst: I agree.  That was for consistency with existing methods' documentation, but I have changed those too.

mernst:
@wmdietl Would you prefer that I rename all the `repr` methods to `debugToString`?  That would be consistent with naming in some other parts of the Checker Framework.  And it's less concise and therefore less obscure.

PR #3524, opened 2020-07-29
Propagate length information from argument to return value of Arrays.copyOf()

### LINKED ISSUE #3224: `Arrays.copyOf` should propagate length information from argument to return value
Consider this source code.

```
import java.util.Arrays;
import org.checkerframework.common.value.qual.MinLen;

public class ArraysCopyOfLength {

  public static void m1(String @MinLen(1) [] args) {
    String @MinLen(1) [] args2 = Arrays.copyOf(args, args.length);
  }

  public static void m2(String @MinLen(1) ... args) {
    String @MinLen(1) [] args2 = Arrays.copyOf(args, args.length);
  }
}
```

The Index Checker issues these warnings:
```
ArraysCopyOfLength.java:7: error: [assignment.type.incompatible] incompatible types in assignment.
    String @MinLen(1) [] args2 = Arrays.copyOf(args, args.length);
                                              ^
  found   : @UnknownVal String @UnknownVal []
  required: @UnknownVal String @ArrayLenRange(from=1, to=2147483647) []
ArraysCopyOfLength.java:11: error: [assignment.type.incompatible] incompatible types in assignment.
    String @MinLen(1) [] args2 = Arrays.copyOf(args, args.length);
                                              ^
  found   : @UnknownVal String @UnknownVal []
  required: @UnknownVal String @ArrayLenRange(from=1, to=2147483647) []
2 errors
```

However, these are false positive warnings, because the result of these calls to `Arrays.copyOf` should have all the same properties as its argument.

PRITI1999:
Fixes #3224;
The Value checker did not infer array length information from the arguments of `Arrays.copyOf()` method leading to false positives.

To avoid those false positives, check if the expression `valueTree` is a call to `Arrays.copyOf()` method in `commonAssignmentCheck()` of `ValueVisitor.java` and then add appropriate `@ArrayLen` annotations to the return value of `Arrays.copyOf()`.


framework/src/main/java/org/checkerframework/common/value/ValueVisitor.java
```diff
@@ -76,6 +79,26 @@ protected void commonAssignmentCheck(
                     getTypeFactory().createIntRangeAnnotation(Range.CHAR_EVERYTHING));
         }
 
+        if (valueTree.getKind() == Kind.METHOD_INVOCATION
+                && TreeUtils.isArrayscopyOfMethodInvocation((MethodInvocationTree) valueTree)
+                && valueType.getKind() == TypeKind.ARRAY) {
+            List<? extends ExpressionTree> args = ((MethodInvocationTree) valueTree).getArguments();
+            AnnotatedTypeMirror arrType = atypeFactory.getAnnotatedType(args.get(0));
+            if (TreeUtils.isArrayLengthAccess(args.get(1))) {
```
kelloggm, line 87: This only tests that the second argument is any array length access expression. The correct test needs to check that it is an array length access for the array that is the first argument. You can add a test case for this:

```
    public static void m4(String @MinLen(1) [] args, String[] otherArray) {
        String @MinLen(1) [] args2 = Arrays.copyOf(args, otherArray.length);
    }
```

framework/src/main/java/org/checkerframework/common/value/ValueVisitor.java
```diff
@@ -76,6 +79,26 @@ protected void commonAssignmentCheck(
...
+                valueType = arrType;
+            } else if (atypeFactory.getAnnotatedType(args.get(1)).getAnnotation(IntVal.class)
+                    != null) {
```
kelloggm, line 90: Why only `IntVal`? If there is an `IntRange` annotation, you could take its lower bound (`from` field) and use that.

kelloggm: (and, you should add a test with `IntRange` when you do)

framework/src/main/java/org/checkerframework/common/value/ValueVisitor.java
PRITI1999: Do we want to check this now? As suggested above I have added to `valuetype` any `@IntVal` or `@IntRange` annotation on the second argument. Suppose the user intends to have `args2` of the same length as `otherArray`, then this check might lead to a false positive.

javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -1273,6 +1273,27 @@ public static boolean isAnonymousConstructor(final MethodTree method) {
         return false;
     }
 
+    /**
+     * Determines if the tree is a call to Arrays.copyOf()
+     *
+     * @param tree tree to check
+     * @return true if the given tree is a call to Arrays.copyOf() method
+     */
+    public static boolean isArrayscopyOfMethodInvocation(MethodInvocationTree tree) {
```
kelloggm, line 1282: nit: in camel-case, this should be `isArraysCopyOfMethodInvocation` (that is, capitalize the "c" in "copy")

javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -1273,6 +1273,27 @@ public static boolean isAnonymousConstructor(final MethodTree method) {
...
+        if (tree.getMethodSelect().getKind() != Kind.MEMBER_SELECT) {
+            return false;
+        }
+
+        MemberSelectTree memberSelectTree = (MemberSelectTree) tree.getMethodSelect();
+
+        if (memberSelectTree.getExpression().toString().equals("Arrays")
+                && getMethodName(memberSelectTree).equals("copyOf")) {
+            return true;
```
kelloggm, line 1291: This test isn't strong enough - it will return true on any method named `copyOf` in a class named `Arrays`, even if it is not the one from the JDK.

You should look at how `IndexMethodIdentifier` in the Index Checker identifies specific methods; you should be able to use the same kind of code here to identify `Arrays.copyOf` with confidence.

(It would probably also be worthwhile to add an inner static class named "Arrays" with a static "copyOf" method to the tests and then ensure that your code is not invoked when that method is called.)

framework/src/main/java/org/checkerframework/common/value/ValueVisitor.java
```diff
@@ -76,6 +80,32 @@ protected void commonAssignmentCheck(
                     getTypeFactory().createIntRangeAnnotation(Range.CHAR_EVERYTHING));
         }
 
+        if (valueTree.getKind() == Kind.METHOD_INVOCATION
+                && TreeUtils.isArrayscopyOfMethodInvocation((MethodInvocationTree) valueTree)
+                && valueType.getKind() == TypeKind.ARRAY) {
+            List<? extends ExpressionTree> args = ((MethodInvocationTree) valueTree).getArguments();
```
kelloggm, line 86: Please add a check here that the length of `args` is what you expect (i.e. 2) and throw a `BugInCF` otherwise

framework/src/main/java/org/checkerframework/common/value/ValueVisitor.java
```diff
@@ -76,6 +80,32 @@ protected void commonAssignmentCheck(
...
+            if (getTypeFactory().getAnnotatedType(args.get(1)).getAnnotation(IntVal.class)
```
kelloggm, line 87: Rather than checking for `IntVal` and `IntRange` by hand, I suggest using `ValueCheckerUtils.getPossibleValues`, which returns a `Range` or `null` if the annotation isn't integral. It already has support for both `IntVal` and `IntRange`, and you should be able to just use its lower bound directly rather than doing the computations below. That will greatly simplify the code.

framework/src/main/java/org/checkerframework/common/value/ValueVisitor.java
kelloggm: In the test case I suggested, that is a true positive: `otherArray` might not be `@MinLen(1)`, even if the user hopes that it is - its length could be zero. I think you can add that test method verbatim into the existing test (except for changing the name).

framework/src/main/java/org/checkerframework/common/value/ValueTreeAnnotator.java
```diff
@@ -398,6 +399,20 @@ public Void visitMethodInvocation(MethodInvocationTree tree, AnnotatedTypeMirror
...
 
+        if (atypeFactory
+                .getMethodIdentifier()
+                .isArraysCopyOfInvocation(tree, atypeFactory.getProcessingEnv())) {
+            List<? extends ExpressionTree> args = tree.getArguments();
+            if (args.size() != 2) {
+                throw new BugInCF(
+                        "Arrays.copyOf() should have 2 arguments. This point should not have reached");
+            }
+            Range range =
```
kelloggm, line 410: `range` can be null here, but `createArrayLenRange` assumes its input in non-null. Can you add a null check and only do the `type.replaceAnnotation(atypeFactory.createArrayLenRangeAnnotation(range));` if the range is non-null?

PR #3527, opened 2020-07-29
Add SystemUtil.sleep method

mernst:


javacutil/src/main/java/org/checkerframework/javacutil/SystemUtil.java
```diff
@@ -242,4 +242,20 @@ public static int getJreVersion() {
         System.arraycopy(array2, 0, result, array1.length, array2.length);
         return result;
     }
+
+    /**
+     * Like Thread.sleep, but does not throw any exceptions, so it is easier fro clients to use.
```
wmdietl, line 247: It looks like it just wraps the `InterruptedException` into a `RuntimeException`, so the documentation could be clearer about what happens with exceptions.

wmdietl, line 247: `fro` -> `for`.

PR #3529, opened 2020-07-30
Called Methods Checker

kelloggm:
This is a port of [the Object Construction Checker](https://github.com/kelloggm/object-construction-checker).

There are ~~two~~ three goals of this PR:
* make the Object Construction Checker a fully-supported part of the Checker Framework, and
* convert the Object Construction Checker to use the new generic accumulation checker
* rename the checker to the Called Methods Checker because it is useful for more than just object construction


checker/build.gradle
```diff
@@ -18,13 +27,29 @@ dependencies {
     // file. :-( So update it and re-build it locally when updating this.
     implementation 'org.plumelib:reflection-util:0.2.2'
 
+    // Object Construction Checker AutoValue + Lombok support
+    implementation "com.google.auto.value:auto-value-annotations:1.7.4"
```
msridhar, line 31: Like we did with the Returns Receiver Checker we likely want to remove this `implementation` dependence and use reflection to identify AutoValue annotations instead

kelloggm: Thanks for pointing that out. I removed the dependency.

kelloggm:
@msridhar @mernst I think this is ready for reviews when you have a chance.

kelloggm:
Mike, Manu, and Martin decided on a call that the name of this checker should be the Called Methods Checker rather than the Object Construction Checker when it is merged into the framework

checker/build.gradle
```diff
@@ -18,13 +27,28 @@ dependencies {
...
     testImplementation group: 'junit', name: 'junit', version: '4.13'
     testImplementation project(':framework-test')
     testImplementation sourceSets.testannotations.output
 
     testannotationsImplementation project(':checker-qual')
 }
 
+dependencyManagement {
+    imports {
+        mavenBom "com.amazonaws:aws-java-sdk-bom:1.11.808"
```
msridhar, line 48: What is this for?  Add a comment?

checker/build.gradle
```diff
@@ -176,6 +200,11 @@ task allNullnessTests(type: Test, group: 'Verification') {
     include '**/Nullness*.class'
 }
 
+task allCalledMethodsTests(type: Test, group: 'Verification') {
```
msridhar, line 203: Why do we need this new task?

checker/src/main/java/org/checkerframework/checker/builder/qual/CalledMethods.java
```diff
@@ -0,0 +1,25 @@
+package org.checkerframework.checker.builder.qual;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * For compatibility with Lombok, retain the old annotations it generates.
```
msridhar, line 10: Maybe rephrase?  Something like "An old version of the annotation, retained for compatibility with Lombok"

Can we add a `@Deprecated` annotation here?

checker/src/main/java/org/checkerframework/checker/builder/qual/ReturnsReceiver.java
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * This declaration annotation indicates that the method on which it is written returns exactly the
+ * receiver object.
+ *
+ * <p>This annotation can only be written on a method declaration. It is inherited by all overrides
+ * of that method.
+ *
+ * <p>This annotation has been replaced by {@link
+ * org.checkerframework.common.returnsreceiver.qual.This}. It is retained only for
```
msridhar, line 17: Maybe `@Deprecated` for this one also?

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,334 @@
...
+import org.checkerframework.framework.type.typeannotator.ListTypeAnnotator;
+import org.checkerframework.framework.type.typeannotator.TypeAnnotator;
+import org.checkerframework.javacutil.ElementUtils;
+import org.checkerframework.javacutil.TreeUtils;
+
+/** The annotated type factory for the Called Methods checker. */
+public class CalledMethodsAnnotatedTypeFactory extends AccumulationAnnotatedTypeFactory {
+
+    /** {@link java.util.Collections#singletonList} is treated specially by the EC2 logic. */
+    private final ExecutableElement collectionsSingletonList;
```
msridhar, line 43: Is it a meaningful performance optimization to keep this in a field?

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,334 @@
...
+
+    /**
+     * Whether to use the Value Checker as a subchecker to reduce false positives when analyzing
+     * calls to the AWS SDK. Defaults to false. Controlled by the command-line option {@code
+     * -AuseValueChecker}.
```
msridhar, line 48: Do we need to rename this option to mention it's for this checker?

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,334 @@
...
+                        }
+                    }
+                }
+            }
+        }
+        return methodName;
+    }
+
+    /** Necessary for the type rule for called methods described below. */
+    private class calledmethodsTreeAnnotator extends AccumulationTreeAnnotator {
```
msridhar, line 230: Camel case

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,334 @@
...
+            }
+            return super.visitNewClass(tree, type);
+        }
+    }
+
+    /**
+     * Adds @CalledMethod annotations for build() methods of AutoValue and Lombok Builders to ensure
+     * required properties have been set.
+     */
+    private class calledmethodsTypeAnnotator extends TypeAnnotator {
```
msridhar, line 278: Camel case

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsChecker.java
```diff
@@ -0,0 +1,94 @@
...
+import org.checkerframework.common.basetype.BaseTypeChecker;
+import org.checkerframework.common.returnsreceiver.ReturnsReceiverChecker;
+import org.checkerframework.common.value.ValueChecker;
+import org.checkerframework.framework.qual.StubFiles;
+import org.checkerframework.framework.source.SupportedOptions;
+import org.checkerframework.framework.source.SuppressWarningsPrefix;
+
+/**
+ * The primary typechecker for the Called Methods checker, which allows programmers to specify
+ * unsafe combinations of options to builder or builder-like interfaces and prevent dangerous
```
msridhar, line 13: Update to say this is just one use case for the checker?

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsChecker.java
```diff
@@ -0,0 +1,94 @@
...
+     * If this option is supplied, use the Value Checker to reduce false positives when analyzing
+     * calls to the AWS SDK.
+     */
+    public static final String USE_VALUE_CHECKER = "useValueChecker";
+
+    /**
+     * If this option is supplied, count the number of analyzed calls to build() in supported
+     * frameworks and print it when analysis is complete. Useful for collecting metrics.
+     */
+    public static final String COUNT_FRAMEWORK_BUILD_CALLS = "countFrameworkBuildCalls";
```
msridhar, line 45: Do we still want to keep this code for collecting metrics?

checker/tests/calledmethods-lombok/BuilderTest.java
```diff
@@ -0,0 +1,126 @@
+// Generated by delombok at Thu May 16 14:44:28 PDT 2019
+
+import lombok.NonNull;
+
+public class BuilderTest {
```
msridhar, line 5: We should change these Lombok tests to run delombok on the original source rather than committing the delombok'd code.  We did this for returns receiver

checker/build.gradle
kelloggm: This is a convenience task for local testing, because this checker has multiple test configurations (i.e. with different sets of framework support enabled, or with the `-AuseValueChecker` option enabled, etc.). The Nullness Checker has a similar task, on which this one is modeled.

checker/src/main/java/org/checkerframework/checker/builder/qual/CalledMethods.java
kelloggm: We decided against it when we merged https://github.com/kelloggm/object-construction-checker/pull/122. The reason is that it would cause deprecation warnings in delomboked code, which is annoying. I'll add a comment to that effect.

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsAnnotatedTypeFactory.java
kelloggm: I don't have any evidence that it is. It's in a field now for code readability, I think.

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsAnnotatedTypeFactory.java
kelloggm: Good catch, we definitely should. I'm going to change it to `-AcalledMethodsUseValueChecker`, but I'm open to other names.

Do you think we should also rename `-AdisableFrameworkSupports` in a similar way?

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsChecker.java
kelloggm: I don't see any reason to remove it

kelloggm:
@msridhar I've addressed your comments

checker/src/main/java/org/checkerframework/checker/builder/qual/CalledMethods.java
msridhar: Ok.  I would still rephrase the first sentence of this Javadoc comment

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsAnnotatedTypeFactory.java
msridhar: Ok.  Since we don't create a lot of type factory objects this is fine

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsChecker.java
msridhar: Ok

docs/manual/called-methods-checker.tex
```diff
@@ -0,0 +1,147 @@
+\htmlhr
+\chapterAndLabel{Called Methods Checker for the builder pattern}{called-methods-checker}
```
msridhar, line 2: Can we rewrite this material to indicate that checking the builder pattern is one use case for the called methods checker?  I'd like to bring out that you can use this checker anytime you need to check a property "Always call `a()` before `b()`"

msridhar: In particular I think the title of the chapter should just be "Called Methods Checker"

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsAnnotatedTypeFactory.java
wmdietl: There is a general mechanism to pass an option only to a particular checker: https://github.com/typetools/checker-framework/blob/master/docs/manual/introduction.tex#L572
So including the checker name in the option seems like overkill to me.

kelloggm: Thanks for pointing this out, @wmdietl. I've made the change.

framework/src/main/java/org/checkerframework/common/returnsreceiver/DescribeImages.astub
```diff
@@ -0,0 +1,24 @@
+// Support for the EC2 classes that are automatically handled by the Called Methods Checker.
```
msridhar, line 1: @kelloggm just confirming, by placing this `.astub` file in the same directory as the source code, I think the checker jar will include the stub file and it will automatically get loaded without passing an `-Astubs` option.  Is this correct?

docs/manual/called-methods-checker.tex
```diff
@@ -0,0 +1,171 @@
...
+Collection.
+\item If the object has a \<toBuilder()> method (for example, if the \<toBuilder = true> option is
+passed to Lombok's \<@Builder> annotation), then the return type of that method is annotated with
+the same \<@CalledMethods> annotation as the receiver of \<build()>, using the same rules as above.
+\item An \<@This> annotation is placed on the return type of each setter in the builder's implementation.
+\end{itemize}
+
+You can disable the framework supports by specifying them in a comma-separated list to the
+command-line flag \<disableFrameworkSupports>.  For example, to disable both Lombok and AutoValue supports,
+use \<-AdisableFrameworkSupports=AutoValue,Lombok>.
```
msridhar, line 142: Should we encourage the user to write `-ACalledMethodsChecker_disableFrameworkSupports` instead?

framework/src/main/java/org/checkerframework/common/returnsreceiver/DescribeImages.astub
kelloggm: Combined with the `@StubFiles` annotation that I added on `ReturnsReceiverChecker`, yes.

docs/manual/called-methods-checker.tex
kelloggm: Probably. I'll change it - nice catch.

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsChecker.java
```diff
@@ -0,0 +1,96 @@
...
+                        .containsKey(this.getClass().getSimpleName() + "_" + USE_VALUE_CHECKER)) {
+            checkers.add(ValueChecker.class);
+        }
+        return checkers;
+    }
+
+    @Override
+    public void typeProcessingOver() {
+        if (getBooleanOption(COUNT_FRAMEWORK_BUILD_CALLS)) {
+            System.out.printf("Found %d build() method calls.\n", numBuildCalls);
```
mernst, line 92: Never use `\n` in printed output, because it is not portable across operating systems.  Use `%n` instead.

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsTransfer.java
```diff
@@ -0,0 +1,40 @@
...
+import org.checkerframework.framework.flow.CFValue;
+
+/** A basic transfer function that accumulates the names of methods called. */
+public class CalledMethodsTransfer extends AccumulationTransfer {
+
+    /** The type factory. */
+    private final CalledMethodsAnnotatedTypeFactory atypeFactory;
+
+    /**
+     * default constructor
```
mernst, line 19: Please write a meaningful comment.  This is worse than nothing.  (Though a good comment is better than nothing, even if the comment is short.)

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsTransfer.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    @Override
+    public TransferResult<CFValue, CFStore> visitMethodInvocation(
+            final MethodInvocationNode node, final TransferInput<CFValue, CFStore> input) {
+        TransferResult<CFValue, CFStore> result = super.visitMethodInvocation(node, input);
+        String methodName = node.getTarget().getMethod().getSimpleName().toString();
+        methodName = atypeFactory.adjustMethodNameUsingValueChecker(methodName, node.getTree());
+        Node receiver = node.getTarget().getReceiver();
+        if (!"<init>".equals(methodName) && receiver != null) {
```
mernst, line 35: This test against methodName seems to come too late.  For efficiency, it could occur after the assignment and before the reassignment.  More importantly, can `methodName` ever be `"<init>"`?  If the method is `<init>`, wouldn't `node` have to be a ConstructorInvocationNode?

Also, the reassignment could occur within the `receiver != null` test.

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsVisitor.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * Custom error message key for finalizer invocations. The custom error message makes it easier
+     * for users to tell what methods weren't called.
+     */
+    public static final @CompilerMessageKey String FINALIZER_INVOCATION_INVALID =
+            "finalizer.invocation.invalid";
+
+    /**
+     * Constructor matching super.
```
mernst, line 29: Write a meaningful comment.

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsVisitor.java
```diff
@@ -0,0 +1,85 @@
...
+                    ((CalledMethodsChecker) checker).numBuildCalls++;
+                    break;
+                }
+            }
+        }
+        return super.visitMethodInvocation(node, p);
+    }
+
+    /**
+     * Adds special reporting for method.invocation.invalid errors to turn them into
```
mernst, line 54: I see no transformation that turns some errors into others.

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsVisitor.java
```diff
@@ -0,0 +1,85 @@
...
+            Set<String> foundMethods =
+                    foundCM == null
+                            ? Collections.emptySet()
+                            : new HashSet<>(atypeFactory.getAccumulatedValues(foundCM));
+            List<String> expectedMethods = atypeFactory.getAccumulatedValues(expectedCM);
+            StringBuilder missingMethods = new StringBuilder();
+            for (String expectedMethod : expectedMethods) {
+                if (!foundMethods.contains(expectedMethod)) {
+                    missingMethods.append(expectedMethod);
+                    missingMethods.append("() ");
```
mernst, line 73: Rather than having a trailing space at the end of missingMethods, use a StringJoiner.

checker/src/main/java/org/checkerframework/checker/calledmethods/DescribeImages.astub
```diff
@@ -0,0 +1,27 @@
+package com.amazonaws.services.ec2;
+
+import org.checkerframework.checker.calledmethods.qual.*;
+
+// Interface
+class AmazonEC2 {
+    DescribeImagesResult describeImages(
+    // any combination of withX/setX has to be permitted if an owner has been set or an imageId has been set or an executable user list has been set
```
mernst, line 8: I don't understand this comment.  Is it trying to say, "The receiver must have its owner, imageId, or executable user set."?

checker/src/main/java/org/checkerframework/checker/calledmethods/GenerateDataKey.astub
```diff
@@ -0,0 +1,8 @@
+package com.amazonaws.services.kms;
+
+import org.checkerframework.checker.calledmethods.qual.*;
+
+interface AWSKMS {
+     // The predicate implements a modified one-hot detector that permits e.g. setNumberOfBytes and withNumberOfBytes to be called together, but not setNumberOfBytes and withKeySpec. Basically, this predicate enforces that only one of the two categories are represented.
```
mernst, line 6: What is "modified" about this?  It looks standard to me.
What are the "two categories"?  That seems like the key information that should be in the comment but isn't.
Nit: "one ... is represented", not "one ... are represented".

checker/src/main/java/org/checkerframework/checker/calledmethods/framework/AutoValueSupport.java
```diff
@@ -0,0 +1,451 @@
...
+     * Simple constructor.
+     *
+     * @param atypeFactory the typechecker's type factory
+     */
+    public AutoValueSupport(CalledMethodsAnnotatedTypeFactory atypeFactory) {
+        this.atypeFactory = atypeFactory;
+    }
+
+    /**
+     * For frameworks to add any necessary @{@link
```
mernst, line 54: This comment is copied from the superclass.  If so, you might as well leave it off, since it will be inherited automatically.
Better, say what this method does.

checker/src/main/java/org/checkerframework/checker/calledmethods/framework/FrameworkSupport.java
```diff
@@ -0,0 +1,50 @@
+package org.checkerframework.checker.calledmethods.framework;
+
+import com.sun.source.tree.NewClassTree;
+import javax.lang.model.element.ExecutableElement;
+import org.checkerframework.framework.type.AnnotatedTypeMirror;
+
+/** add CalledMethods annotations to code generated by framework */
```
mernst, line 7: Please expand this comment.

checker/src/main/java/org/checkerframework/checker/calledmethods/framework/FrameworkSupport.java
```diff
@@ -0,0 +1,50 @@
...
+public interface FrameworkSupport {
```
mernst, line 8: The term "framework" is too broad.  It's confusing in a project named the "Checker Framework" that has a component called "framework".  I suggest that you change "framework" to "builder framework" (and Framework to BuilderFramework, etc.) throughout.  I think that will be clearer.

checker/src/main/java/org/checkerframework/checker/calledmethods/framework/FrameworkSupport.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Allows for supporting a framework's {@code toBuilder} routine. Typically, the returned
```
mernst, line 11: This doesn't say what the method is or when it is called.  A discussion of intended use is helpful, but first describe the method.

checker/src/main/java/org/checkerframework/checker/calledmethods/framework/FrameworkSupport.java
```diff
@@ -0,0 +1,50 @@
...
+     * this fact.
+     *
+     * @param t a method that is possibly the {@code toBuilder} method for a type that has an
+     *     associated builder. An implementation must check that this is actually the {@code
+     *     toBuilder} method for the corresponding framework.
+     */
+    void handlePossibleToBuilder(AnnotatedTypeMirror.AnnotatedExecutableType t);
+
+    /**
+     * For {@code build} methods on {@code Builder} types, the framework support should determine
```
mernst, line 23: I was confused by this comment.  I now think "should determine" may describe what an implementation of this method should do, but that was not clear.  On rereading it makes sense, but please rephrase it to make sense the first time through.  This comment applies to all of the methods in this interface.

checker/src/main/java/org/checkerframework/checker/calledmethods/framework/FrameworkSupportUtils.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /**
+     * Given an annotation name, return true if the element has the annotation of that name
+     *
+     * @param element the element
+     * @param annotName name of the annotation
+     * @return true if the element has the annotation of that name
+     */
+    public static boolean hasAnnotation(Element element, String annotName) {
```
mernst, line 42: I suggest putting this in `ElementUtils`.

checker/src/main/java/org/checkerframework/checker/calledmethods/framework/FrameworkSupportUtils.java
```diff
@@ -0,0 +1,76 @@
...
+    /** The frameworks supported by the checker. */
+    public enum Framework {
+        AUTO_VALUE,
+        LOMBOK;
+    }
+
+    /**
+     * Gets the list of supported that are currently enabled frameworks.
+     *
+     * @param option the value of the framework support option
```
mernst, line 56: Be specific:  give the name of the command-line option here.

docs/manual/called-methods-checker.tex
```diff
@@ -0,0 +1,187 @@
...
+
+If you use AutoValue or Lombok, most specifications are automatically
+inferred by the Called Methods Checker, from field annotations such as
+\<@Nullable> and field types such as \<Optional>. See the
+section on defaulting rules for Lombok and AutoValue for more details
+(Section~\ref{called-methods-framework-details}).
+
+In some cases, you may need to specify your code. You do so by writing one of the following type
+annotations:
+\begin{description}
```
mernst, line 72: The manual should list all the annotations.  (It lacks the bottom annotation.)  It should also show the subtyping relationship.  Please draw a type hierarchy diagram like those in other manual sections.

checker/src/main/java/org/checkerframework/checker/calledmethods/framework/FrameworkSupportUtils.java
kelloggm: The `withValues` bit is an example - it could have been any method in `Filter`. The comment is a bit confusing, so I will rewrite it to be more clear.

checker/build.gradle
```diff
@@ -102,6 +127,7 @@ artifacts {
 
 clean {
     delete "${projectDir}/dist"
+    delete "tests/calledmethods-delomboked"
```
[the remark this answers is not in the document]
kelloggm: It must be placed under the `tests` directory in order to be usable with `CheckerFrameworkPerDirectoryTest`. I'll add this directory to the `.gitignore`, however.

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,337 @@
...
+                .equals(ElementUtils.enclosingClass(invokedMethod).getQualifiedName().toString())) {
+            return methodName;
+        }
+
+        if ("withFilters".equals(methodName) || "setFilters".equals(methodName)) {
+            for (Tree filterTree : tree.getArguments()) {
+                // Search the arguments to withFilters for a Filter constructor invocation,
+                // passing through as many method invocation trees as needed. This code is searching
+                // for code of the form:
+                // new Filter("owner").withValues("...")
```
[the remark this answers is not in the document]
kelloggm: That's an example, not the actual target. I've clarified the comment.

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsChecker.java
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * The primary typechecker for the Called Methods Checker, which tracks the methods that have
+ * definitely been called on an object. One common use case for the Called Methods Checker is to
+ * specify safe combinations of options to builder or builder-like interfaces, preventing objects
+ * from being instantiated incompletely.
+ */
+@SuppressWarningsPrefix({
+    // Preferred checkername.
+    "calledmethods",
+    "called.methods",
```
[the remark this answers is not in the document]
kelloggm: The object construction checker supports the other variants, so to be fully compatible so does this checker. I would prefer backwards compatibility over purity in our internal consistency, so I'd prefer to keep those (I did remove the new, weird one).

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsChecker.java
```diff
@@ -0,0 +1,96 @@
...
+                super.getImmediateSubcheckerClasses();
+        checkers.add(ReturnsReceiverChecker.class);
+
+        // BaseTypeChecker#hasOption calls this method (so that all subcheckers' options are
+        // considered),
+        // so the processingEnvironment must be checked for the option directly.
+        if (this.processingEnv.getOptions().containsKey(USE_VALUE_CHECKER)
+                || this.processingEnv
+                        .getOptions()
+                        .containsKey(this.getClass().getSimpleName() + "_" + USE_VALUE_CHECKER)) {
```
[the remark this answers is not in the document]
kelloggm: There is already a comment here explaining why: `BaseTypeChecker#hasOption` calls `getImmediateSubcheckers`, so using the CF's standard option code doesn't work. We have to implement the logic ourselves (for just this option) here, because the effect of the option is to enable a new subchecker.

It might be possible to refactor the logic for options so that this hack can be avoided, but I'd prefer to do that separately (if we do it at all, which I don't see much reason to - it is quite rare for a subchecker to be guarded by an option flag).

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsTransfer.java
kelloggm: > This is worse than nothing

The original code had nothing, and I would have preferred to keep it, but that causes require javadoc to fail.

> a good comment is better than nothing, even if the comment is short

I don't think that's true, in this case. I'll try to improve the comment, but there really isn't anything to say about this constructor - all it does is call super.

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsTransfer.java
kelloggm: > If the method is <init>, wouldn't node have to be a ConstructorInvocationNode

That's what I would expect. I thought this had caused a failure when I was testing the accumulation checker, but I can't reproduce it. I'll remove the test.

> Also, the reassignment could occur within the receiver != null test.

I think that is bad style.

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsVisitor.java
kelloggm: It is implicit: this method in `BaseTypeVisitor` normally would report a `method.invocation.invalid` error every time it is called. This implementation doesn't do that, and instead reports `finalizer.invocation.invalid` errors in the same places.

checker/src/main/java/org/checkerframework/checker/calledmethods/DescribeImages.astub
kelloggm: Yes, but it is also saying that any other combination of methods is permitted. I'll rewrite to make it more clear.

checker/src/main/java/org/checkerframework/checker/calledmethods/GenerateDataKey.astub
kelloggm: I don't know. I think Martin Schäf wrote this comment. I'll rewrite it to be more clear what the predicate represents.

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,406 @@
...
+        Set<String> disabledFrameworks = new HashSet<>();
+        if (checker.hasOption(CalledMethodsChecker.DISABLE_BUILDER_FRAMEWORK_SUPPORTS)) {
+            disabledFrameworks.addAll(
+                    Arrays.asList(
+                                    checker.getOption(
+                                                    CalledMethodsChecker
+                                                            .DISABLE_BUILDER_FRAMEWORK_SUPPORTS)
+                                            .split(","))
+                            .stream()
+                            .map(String::toUpperCase)
```
mernst, line 100: Don't normalize case.  It is confusing that the examples in the documentation and in the tests are different.  Permitting arbitrary casing leads to such confusion, and makes it more difficult for a user to find all uses of a particular string in their code or the code of other people (because it could appear in multiple different forms).  It does not prevent any hidden problems, because this code warns if a user misspells a framework name.  In other words, I see disadvantages but no benefits to this undocumented feature.

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,406 @@
...
+                            .collect(Collectors.toList()));
+        }
+        builderFrameworkSupports = new ArrayList<>();
+        enableFramework(CalledMethodsChecker.LOMBOK_SUPPORT, disabledFrameworks);
+        enableFramework(CalledMethodsChecker.AUTOVALUE_SUPPORT, disabledFrameworks);
+
+        if (!disabledFrameworks.isEmpty()) {
+            StringJoiner sj = new StringJoiner(", ");
+            disabledFrameworks.iterator().forEachRemaining(s -> sj.add(s));
+            String unrecognized = sj.toString();
```
mernst, line 110: Rather than these 3 lines of code, call [`UtilPlume.join`](http://plumelib.org/plume-util/api/org/plumelib/util/UtilPlume.html#join-java.lang.CharSequence-java.lang.Iterable-).
Actually, it would be fine to just issue an error about the first incorrect string found.  That would lead to simpler code, and it is adequate for users.

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,406 @@
...
+    /**
+     * Enables support for the given builder-generation framework, unless it is listed in the
+     * disabled builder frameworks parsed from the -AdisableBuilderFrameworkSupport option's
+     * arguments.
+     *
+     * @param framework the builder framework to enable
+     * @param disabledFrameworks the set of disabled builder frameworks. This argument will be
+     *     side-effected to remove a builder framework if it was actually disabled.
+     */
+    private void enableFramework(String framework, Set<String> disabledFrameworks) {
```
mernst, line 136: This method can be inlined (and specialized) where it is used, and I think the code will be simpler.

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsChecker.java
```diff
@@ -0,0 +1,94 @@
...
+     * specifications on custom builder objects (such as the AWS SDK examples).
+     */
+    public static final String DISABLE_BUILDER_FRAMEWORK_SUPPORTS =
+            "disableBuilderFrameworkSupports";
+
+    /**
+     * The key for the {@link #DISABLE_BUILDER_FRAMEWORK_SUPPORTS} option to disable Lombok support.
+     * Use it via {@code -AdisableBuilderFrameworkSupports=LOMBOK}.
+     */
+    public static final String LOMBOK_SUPPORT = "LOMBOK";
```
mernst, line 50: This string only needs to be used once or maybe in the code, and it is unlikely to change.  (Also, the name `LOMBOK_SUPPORT` is a bit misleading since this is a string or name, not a BuilderFrameworkSupport.)  Just inline it where it is used.  That will simplify the code and make it easier to read.

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsVisitor.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * This visitor implements the custom error message finalizer.invocation.invalid, and supports
+ * counting the number of framework build calls.
+ */
+public class CalledMethodsVisitor extends AccumulationVisitor {
+
+    /** Error message key for incorrect finalizer invocations. */
+    public static final @CompilerMessageKey String FINALIZER_INVOCATION_INVALID =
+            "finalizer.invocation.invalid";
```
mernst, line 27: This is only used once.  There is no danger of misspelling because of the Compiler Message Checker.  So, please inline it to make the code shorter and easier to read.

checker/src/main/java/org/checkerframework/checker/calledmethods/builder/BuilderFrameworkSupport.java
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * Hook for automatically adding annotations to a build() method (i.e. a finalizer) generated by
+     * a builder framework.
+     *
+     * <p>For {@code build} methods on {@code Builder} types, implementations of this method should
+     * determine the required properties and add a corresponding {@link
+     * org.checkerframework.checker.calledmethods.qual.CalledMethods} annotation to the type of the
+     * receiver parameter.
+     *
+     * @param t a method that is possibly the {@code build} method for a builder. The only
```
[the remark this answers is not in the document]
kelloggm: I've made this change. I think it does improve the code overall, but it also made some of the code in `AutoValueSupport` a bit unpleasant - the logic for `isToBuilder` and `handleToBuilderMethod` were coupled, and I didn't find a way to uncouple them entirely.

checker/src/main/java/org/checkerframework/checker/calledmethods/builder/BuilderFrameworkSupport.java
```diff
@@ -0,0 +1,65 @@
...
+     *     guaranteed condition is that the enclosing class for the method is itself an inner class.
+     */
+    void handlePossibleBuilderBuildMethod(AnnotatedExecutableType t);
+
+    /**
+     * Hook to add any necessary @{@link
+     * org.checkerframework.checker.calledmethods.qual.CalledMethods} annotation to a constructor
```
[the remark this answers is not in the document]
kelloggm: The specification doesn't require annotations to be added in a particular place, and in fact the support for Lombok and for AutoValue implement this method differently (there are no annotations to infer for Lombok, but for AutoValue the generated copy constructor is treated the same as a `toBuilder` method).

checker/src/main/java/org/checkerframework/checker/calledmethods/builder/AutoValueSupport.java
```diff
@@ -0,0 +1,448 @@
...
+     * Create a new AutoValueSupport.
+     *
+     * @param atypeFactory the typechecker's type factory
+     */
+    public AutoValueSupport(CalledMethodsAnnotatedTypeFactory atypeFactory) {
+        this.atypeFactory = atypeFactory;
+    }
+
+    /**
+     * This method ensures that the type of a copy constructor matches the type of the AutoValue
```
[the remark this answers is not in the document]
kelloggm: It performs side effects to make the fact true. I'll clarify the documentation.

checker/src/main/java/org/checkerframework/checker/calledmethods/builder/AutoValueSupport.java
```diff
@@ -0,0 +1,448 @@
...
+     * toBuilder method.
+     *
+     * @param tree AST for the call
```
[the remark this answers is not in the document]
kelloggm: Not necessarily - this is a call to any constructor. This method's body decides if that constructor is a copy constructor generated by AutoValue, and side-effects its type if so.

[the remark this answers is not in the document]
kelloggm: I've clarified the documentation.

checker/src/main/java/org/checkerframework/checker/calledmethods/builder/AutoValueSupport.java
```diff
@@ -0,0 +1,448 @@
...
+     * @param classElement corresponding AutoValue class
+     */
+    private void handleToBuilderType(
+            AnnotatedTypeMirror type, TypeMirror builderType, Element classElement) {
+        Element builderElement = TypesUtils.getTypeElement(builderType);
+        Set<String> avBuilderSetterNames = getAutoValueBuilderSetterMethodNames(builderElement);
+        List<String> requiredProperties =
+                getAutoValueRequiredProperties(classElement, avBuilderSetterNames);
+        AnnotationMirror calledMethodsAnno =
+                createCalledMethodsForAutoValueProperties(requiredProperties, avBuilderSetterNames);
```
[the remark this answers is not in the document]
kelloggm: I followed the former suggestion, but I disagree with the latter. Because there is no need for such a method outside the implementation of a class that is implementing `BuilderFrameworkSupport`, adding it to the interface would clutter the interface and make implementation of new builder framework support classes more annoying. (I agree that it's common to all builder supports, but I don't think it's helpful to proscribe e.g. the inputs for such a method, since different builder frameworks might use different information to decide on the required properties.)

checker/src/main/java/org/checkerframework/checker/calledmethods/builder/AutoValueSupport.java
```diff
@@ -0,0 +1,448 @@
...
+     * @return {@code "com.google.auto.value"}
+     */
+    private String getAutoValuePackageName() {
+        String com = "com";
+        return com + "." + "google.auto.value";
+    }
+
+    /**
+     * @param symbol symbol for a class
+     * @return list including the class and all its supertypes, with a guarantee that subtypes
```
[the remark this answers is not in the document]
kelloggm: I think the weaker guarantee is sufficient: AutoValue isn't going to generate anything related to interfaces, and this code is only used to find all the methods that might be AutoValue-generated. I'll update the documentation.

checker/src/main/java/org/checkerframework/checker/calledmethods/builder/LombokSupport.java
```diff
@@ -0,0 +1,204 @@
...
+     * with @lombok.NonNull annotations.
+     *
+     * @param lombokClassElement the class with the @lombok.Builder annotation
+     * @return a list of required property names
+     */
+    private List<String> getLombokRequiredProperties(final Element lombokClassElement) {
+        List<String> requiredPropertyNames = new ArrayList<>();
+        List<String> defaultedPropertyNames = new ArrayList<>();
+        for (Element member : lombokClassElement.getEnclosedElements()) {
+            if (member.getKind() == ElementKind.FIELD) {
```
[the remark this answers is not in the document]
kelloggm: Lombok never generates fields with initializers in classes that use `@Builder`, unless that field has an `@Default` or `@Singular` annotation, which are handled elsewhere (and don't need to be included here, because they can never be null).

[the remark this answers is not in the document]
kelloggm: I'll add a comment to this effect.

checker/src/main/java/org/checkerframework/checker/calledmethods/builder/LombokSupport.java
```diff
@@ -0,0 +1,204 @@
...
+                    && ElementUtils.hasAnnotation(member, "lombok.Generated")) {
+                String methodName = member.getSimpleName().toString();
+                // Handle fields with @Builder.Default annotations.
+                // If a field foo has an @Builder.Default annotation, Lombok always generates a
+                // method called $default$foo.
+                if (methodName.startsWith("$default$")) {
+                    String propName = methodName.substring(9); // $default$ has 9 characters
+                    defaultedPropertyNames.add(propName);
+                }
+            } else if (member.getKind().isClass() && member.toString().endsWith("Builder")) {
```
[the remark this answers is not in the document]
kelloggm: This is checking for the generated Builder class' element. It's a heuristic in that unless the `builderClassName` option to the `@Builder` annotation is passed, Lombok is guaranteed to generate an inner class with a name ending in `Builder` (e.g. `FooBuilder` for a `Foo` object). I've never seen anyone use that option in practice (though I'm sure some do), but it's not a soundness concern because the worst that happens is the code misses a singular/default annotation and expects you to set a field that's defaulted.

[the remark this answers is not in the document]
kelloggm: I added a comment.

checker/src/main/java/org/checkerframework/checker/calledmethods/builder/LombokSupport.java
```diff
@@ -0,0 +1,204 @@
...
+                            defaultedPropertyNames.add(propName);
+                        }
+                    } else if (builderMember.getKind() == ElementKind.FIELD) {
+                        VariableTree variableTree =
+                                (VariableTree) atypeFactory.declarationFromElement(builderMember);
+                        if (variableTree != null && variableTree.getInitializer() != null) {
+                            String propName = variableTree.getName().toString();
+                            defaultedPropertyNames.add(propName);
+                            defaultedElements.put(builderMember, propName);
+                        } else if (defaultedElements.containsKey(builderMember)) {
```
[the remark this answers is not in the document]
kelloggm: I don't think this code relies on the order of elements, but even if it does it's not a concern because Lombok will always generate things in the same order (and it's not a soundness concern, so we'd get FP warnings that would indicate the problem if the generator changes).

checker/src/main/java/org/checkerframework/checker/calledmethods/builder/LombokSupport.java
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    @Override
+    public boolean isBuilderBuildMethod(ExecutableElement element) {
+        TypeElement enclosingElement = (TypeElement) element.getEnclosingElement();
+
+        if ((ElementUtils.hasAnnotation(enclosingElement, "lombok.Generated")
+                        || ElementUtils.hasAnnotation(element, "lombok.Generated"))
+                && enclosingElement.getSimpleName().toString().endsWith("Builder")) {
+            return "build".equals(element.getSimpleName().toString());
```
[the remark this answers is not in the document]
kelloggm: On line 199 above, the `toString` is necessary because `Name` doesn't have an `endsWith` method or equivalent.

Most of the uses of `Name#toString` precede such a call. In other cases, a mix of `Name`s and `String`s are used, so using `Name#toString` is sensible so that they can all be stored together. I don't think it's possible to remove most of the `toString` calls in this class (or even many of them), but I've removed the ones that can be replaced with `contentEquals`.

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsAnnotatedTypeFactory.java
kelloggm: I rewrote the whole method based on your other comments. I think it's clearer now as a separate method, but I can also inline it into the constructor if you still think that would be better.

kelloggm:
@mernst I think this is ready for another review

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,387 @@
...
+                    && filterTreeAsMethodInvocation.getArguments().size() >= 1) {
+                Tree withNameArgTree = filterTreeAsMethodInvocation.getArguments().get(0);
+                String withNameArg =
+                        ValueCheckerUtils.getExactStringValue(withNameArgTree, valueATF);
+                return filterKindToMethodName(withNameArg);
+            }
+
+            if (TreeUtils.isMethodInvocation(
+                    filterTree, collectionsSingletonList, getProcessingEnv())) {
+                // Descend into a call to Collections.singletonList()
```
mernst, line 235: Descending into singletonList is only relevant at the top level (as the argument `filterTree`), so it doesn't belong within the `while` loop.

In fact, it doesn't belong in `filterTreeToMethodName`.  Move it into `adjustMethodNameUsingValueChecker`, since it is logically part of the `for (Tree filterTree : tree.getArguments())` loop.  I think this will clarify the logic.

mernst (changes requested):

checker/src/main/java/org/checkerframework/checker/calledmethods/builder/AutoValueSupport.java
```diff
@@ -0,0 +1,458 @@
...
+     * @return true if {@code member} is required
+     */
+    private boolean isAutoValueRequiredProperty(Element member, Set<String> avBuilderSetterNames) {
+        String name = member.getSimpleName().toString();
+        // Ignore java.lang.Object overrides, constructors, and toBuilder methods in AutoValue
+        // classes.
+        // Strictly speaking, this code should check return types, etc. to handle strange
+        // overloads and other corner cases. They seem unlikely enough that we are skipping for now.
+        if (ArraysPlume.indexOf(
+                        new String[] {"equals", "hashCode", "toString", "<init>", "toBuilder"},
```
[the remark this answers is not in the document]
kelloggm: In your previous review (here: https://github.com/typetools/checker-framework/pull/3529#discussion_r476477088), you suggested inlining the existing final static field because it is only used once, which directly contradicts this suggestion.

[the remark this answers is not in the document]
mernst: Oops!  Sorry about that.  I must not have been thinking about the cost of recomputing it.  I don't have a feel for whether the run-time cost is actually a problem -- I just worried about the extra allocation and garbage collection.

checker/src/main/java/org/checkerframework/checker/calledmethods/builder/LombokSupport.java
```diff
@@ -0,0 +1,202 @@
...
+                            "org.jetbrains.annotations.NotNull",
+                            "org.jmlspecs.annotation.NonNull",
+                            "org.netbeans.api.annotations.common.NonNull",
+                            "org.springframework.lang.NonNull"));
+
+    /**
+     * A map from elements that have a lombok.Builder.Default annotation to the simple property name
+     * that should be treated as defaulted.
+     *
+     * <p>This cache is kept so that when declarationFromElement doesn't work, defaults are still
```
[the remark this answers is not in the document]
kelloggm: I added a comment, copied below:

>     * <p>This cache is kept because the usual method for checking that an element has been defaulted
>     * (calling declarationFromElement and examining the resulting VariableTree) only works if a corresponding
>     * Tree is available (for code that is only available as bytecode, no such Tree is available and that
>     * method returns null). See the code in {@link #getLombokRequiredProperties(Element)} that handles
>     * fields.

checker/src/main/java/org/checkerframework/checker/calledmethods/builder/LombokSupport.java
```diff
@@ -0,0 +1,202 @@
...
+     * @return a list of required property names
+     */
+    private List<String> getLombokRequiredProperties(final Element lombokClassElement) {
+        List<String> requiredPropertyNames = new ArrayList<>();
+        List<String> defaultedPropertyNames = new ArrayList<>();
+        for (Element member : lombokClassElement.getEnclosedElements()) {
+            if (member.getKind() == ElementKind.FIELD) {
+                // Lombok never generates non-null fields with initializers in builders, unless
+                // the field is annotated with @Default or @Singular, which are handled elsewhere.
+                // So, this code doesn't need to consider fields with initializers.
```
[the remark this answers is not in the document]
kelloggm: Yes. I've clarified the comment.

checker/src/main/java/org/checkerframework/checker/calledmethods/builder/AutoValueSupport.java
```diff
@@ -0,0 +1,460 @@
...
+        return ElementUtils.hasAnnotation(element, getAutoValuePackageName() + ".AutoValue");
+    }
+
+    /**
+     * Add, to {@code type}, a CalledMethods annotation with all required methods called. The type
+     * can be the return type of toBuilder or the corresponding generated "copy" constructor.
+     *
+     * @param type type to update
+     * @param builderType type of abstract @AutoValue.Builder class
+     * @param classElement corresponding AutoValue class
```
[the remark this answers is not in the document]
kelloggm: The comment was unclear - that was the intended meaning, but I see (and have fixed) the ambiguity.

checker/src/main/java/org/checkerframework/checker/calledmethods/builder/AutoValueSupport.java
```diff
@@ -0,0 +1,460 @@
...
+                    };
+            for (String setterName : setterNamesToTry) {
+                if (builderSetterNames.contains(setterName)) {
+                    return setterName;
+                }
+            }
+        }
+
+        // Could not find a corresponding setter.  This is likely because an AutoValue Extension is
+        // in use.  See https://github.com/kelloggm/object-construction-checker/issues/110 .
```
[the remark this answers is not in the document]
kelloggm: Eventually. Let's wait to move issues over until the checker actually exists, though. I'll go through the open OCC issues and move all the pertinent ones after this PR is merged (and fix this reference when I do).

checker/src/main/java/org/checkerframework/checker/calledmethods/builder/AutoValueSupport.java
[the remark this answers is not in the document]
kelloggm: I think the benefit to readability is worth the little bit of extra allocation/deallocation. This method is only called when actually analyzing a class that AutoValue generated (which shouldn't be that many classes), so the runtime overhead shouldn't be too bad.

checker/src/main/java/org/checkerframework/checker/calledmethods/builder/AutoValueSupport.java
```diff
@@ -0,0 +1,458 @@
...
+                != -1) {
+            return false;
+        }
+        TypeMirror returnType = ((ExecutableElement) member).getReturnType();
+        if (returnType.getKind() == TypeKind.VOID) {
+            return false;
+        }
+        // shouldn't have a nullable return
+        boolean hasNullable =
+                Stream.concat(
```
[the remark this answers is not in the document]
kelloggm: I think either way is fine. I'm going to leave it as-is since I don't see a compelling reason to change it and this version works.

checker/src/main/java/org/checkerframework/checker/calledmethods/builder/AutoValueSupport.java
```diff
@@ -0,0 +1,458 @@
...
+     * @param type some type
+     * @return true if type is an Optional type
+     */
+    static boolean isOptional(TypeMirror type) {
+        if (type.getKind() != TypeKind.DECLARED) {
+            return false;
+        }
+        DeclaredType declaredType = (DeclaredType) type;
+        TypeElement typeElement = (TypeElement) declaredType.asElement();
+        // This list of classes that AutoValue considers "optional" comes from AutoValue's source
```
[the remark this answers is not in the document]
kelloggm: Similar to the other potential static field in this class, you specifically asked me to inline this in your previous review. I think the benefit to readability that is gained from having this list at the exact place that it is used is worth the small efficiency loss.

checker/src/main/java/org/checkerframework/checker/calledmethods/builder/AutoValueSupport.java
```diff
@@ -0,0 +1,458 @@
...
+        String com = "com";
+        return com + "." + "google.auto.value";
+    }
+
+    /**
+     * Get all the supertypes of a given class, including that class.
+     *
+     * @param symbol symbol for a class
+     * @return list including the class and all its supertypes, with a guarantee that direct
+     *     subtypes (i.e. those that appear in extends clauses) appear before supertypes
```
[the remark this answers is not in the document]
kelloggm: That is much clearer, thanks.

kelloggm:
@mernst I've addressed your comments, and I think this is ready for another round of review. Thanks!

mernst (approved):

Issue #3528, opened 2020-07-30
The Check Framework assumes annotation processor options only contain one underscore


cushon:
In the following example the Checker Framework crashes after seeing a `-A` annotation processor option with multiple underscores. This prevents it from being used in combination with other annotation processors that define their own `-A` flags.

I think it should probably just ignore `-A` options it doesn't recognize?

```
$ ./checker-framework-3.5.0/checker/bin/javac -processor nullness -Anot_a_checker_framework_option=true T.java
An annotation processor threw an uncaught exception.
Consult the following stack trace for details.
org.checkerframework.javacutil.UserError: Invalid option name: not_a_checker_framework_option At most one separator _ expected, but found 5.
        at org.checkerframework.framework.source.SourceChecker.createActiveOptions(SourceChecker.java:1502)
        at org.checkerframework.framework.source.SourceChecker.getOptions(SourceChecker.java:1597)
        at org.checkerframework.common.basetype.BaseTypeChecker.getOptions(BaseTypeChecker.java:726)
        at org.checkerframework.framework.source.SourceChecker.hasOption(SourceChecker.java:1538)
        at org.checkerframework.framework.source.SourceChecker.init(SourceChecker.java:515)
        at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment$ProcessorState.<init>(JavacProcessingEnvironment.java:679)
        at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment$DiscoveredProcessors$ProcessorStateIterator.next(JavacProcessingEnvironment.java:779)
        at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.discoverAndRunProcs(JavacProcessingEnvironment.java:874)
        at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.access$2200(JavacProcessingEnvironment.java:111)
        at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment$Round.run(JavacProcessingEnvironment.java:1215)
        at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.doProcessing(JavacProcessingEnvironment.java:1327)
        at jdk.compiler/com.sun.tools.javac.main.JavaCompiler.processAnnotations(JavaCompiler.java:1234)
        at jdk.compiler/com.sun.tools.javac.main.JavaCompiler.compile(JavaCompiler.java:918)
[...]

PR #3530, opened 2020-07-30
Improve UserError message in createActiveOptions()

PRITI1999:
The `UserError` thrown in `createActiveOptions()` of `SourceChecker.java` displays the number of `OPTION_SEPARATOR` given as the number of strings obtained on splitting. Generally this value would be 1 less than the number of strings.


wmdietl:
Thanks for this fix!

wmdietl:
I guess this change was motivated by the numbers in #3528. Adding cross-reference.

PR #3531, opened 2020-07-30
Change block labels from long to Object; this makes formatting more flexible

mernst:


dataflow/src/main/java/org/checkerframework/dataflow/cfg/DOTCFGVisualizer.java
```diff
@@ -124,7 +124,7 @@ public String visualizeNodes(
     }
 
     @Override
-    protected String addEdge(long sId, long eId, String flowRule) {
+    protected String addEdge(Object sId, Object eId, String flowRule) {
         return "    " + sId + " -> " + eId + " [label=\"" + flowRule + "\"];" + lineSeparator;
```
wmdietl, line 128: Here and below, now that the input can be an arbitrary object, we should call `format` on them to make sure they are escaped properly for output.

mernst: Can you clarify?  Currently, toString() is being called implicitly on `sId` and `eId`.  Are you suggesting that `String.format` should be used instead of String concatenation `+`?  I don't understand the problem that would prevent, so I must be missing something.

wmdietl: Sorry for not being explicit. I meant [`AbstractCFGVisualizer.format`](https://github.com/typetools/checker-framework/blob/master/dataflow/src/main/java/org/checkerframework/dataflow/cfg/AbstractCFGVisualizer.java#L269) which takes an `Object` and turns it into a `String` usable for the corresponding format. This is important for DOT, where `"` needs to be escaped.

PR #3533, opened 2020-07-31
CFGVisualizer: Don't return a string with the escapeString.

xingweitian:
Remove escapeString from `visualizeStore*` methods. Use `StringJoiner` to add escapeString at a higher level.


mernst:
@wmdietl Why was this closed?  The overall goal was sound, I believe.

PR #3534, opened 2020-07-31
Move getSuccessors() into Block

mernst:


dataflow/src/main/java/org/checkerframework/dataflow/cfg/block/Block.java
```diff
@@ -41,4 +42,11 @@
      * @return the predecessors of this basic block
      */
     Set<Block> getPredecessors();
+
+    /**
+     * Returns all the successors of this basic block.
+     *
+     * @return all the successors of this basic block
+     */
+    Collection<Block> getSuccessors();
```
wmdietl, line 51: Should this be `Set` like for `getPredecessors()`? We don't need duplicates in the result.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/block/Block.java
```diff
@@ -41,4 +42,11 @@
      * @return the predecessors of this basic block
      */
     Set<Block> getPredecessors();
+
+    /**
+     * Returns all the successors of this basic block.
```
wmdietl, line 47: In contrast to `getPredecessors` there is an `all` in this sentence, which made me think this gives all transitive successors. It doesn't look like that is the case or intention.
So how about replacing all with `direct` or removing the `all`?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/block/Block.java
mernst: This forces use of `LinkedHashSet` throughout instead of the more efficient `ArrayList`.  I don't see anywhere that clients of `getPredecessors` require (for correctness) that the result is a set.

mernst:
> Should the title be `Add getSuccessors() to Block`, because this adds a new method as far as I can tell.

It already existed in ControlFlowGraph, which was not a logical place for it.

PR #3535, opened 2020-07-31
Improve documentation

mernst:


dataflow/src/main/java/org/checkerframework/dataflow/cfg/ControlFlowGraph.java
```diff
@@ -21,14 +21,22 @@
...
 import org.checkerframework.dataflow.cfg.node.AssignmentNode;
 import org.checkerframework.dataflow.cfg.node.Node;
 import org.checkerframework.dataflow.cfg.node.ReturnNode;
 
-/** A control flow graph (CFG for short) of a single method. */
+/**
+ * A control flow graph (CFG for short) of a single method.
+ *
+ * <p>The graph is represented by the successors (methods {@link SingleSuccessorBlock#getSuccessor},
+ * {@link ConditionalBlock#getThenSuccessor}, {@link ConditionalBlock#getElseSuccessor}, {@link
```
wmdietl, line 36: After #3534 this can be simplified.

mernst: Even when getSuccessors exists, this is still accurate and more informative.

PR #3536, opened 2020-07-31
Add Block.getNodes() method

mernst:
Merge after https://github.com/typetools/checker-framework/pull/3534


dataflow/src/main/java/org/checkerframework/dataflow/cfg/block/Block.java
```diff
@@ -41,4 +43,18 @@
...
+     * @return all the successors of this basic block
+     */
+    Collection<Block> getSuccessors();
+
+    /**
+     * Returns the nodes contained within this basic block.
+     *
+     * @return the nodes contained within this basic block
+     */
+    Collection<Node> getNodes();
```
wmdietl, line 59: As the order of `Node`s is important, returning a `List` is more appropriate.

mernst:
> What is the motivation

The motivation is the need to check and repair ControlFlowGraphs.  Currently, CFG construction creates broken data structures with maps that do not correspond to one another.

> Doesn't this make the Block interface misleading?

I don't see how it is misleading.  I suspect you have some concern in mind.  Could you please state that concern concretely?

wmdietl:
> I don't see how it is misleading. I suspect you have some concern in mind. Could you please state that concern concretely?

3 out of 5 blocks don't have nodes. I find it misleading to add a method that makes it seem like most blocks contain nodes.

mernst:
> > I don't see how it is misleading. I suspect you have some concern in mind. Could you please state that concern concretely?
> 
> 3 out of 5 blocks don't have nodes. I find it misleading to add a method that makes it seem like most blocks contain nodes.

The method returns all the nodes that each block contains.  It does not promise to return a non-empty list.  It does not imply that most blocks contain nodes.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/block/RegularBlockImpl.java
```diff
@@ -39,6 +39,11 @@ public void addNodes(List<? extends Node> ts) {
         return Collections.unmodifiableList(contents);
     }
 
+    @Override
+    public List<Node> getNodes() {
```
wmdietl, line 43: Instead of now having two methods, `getNodes` and `getContents`, in regular blocks, I would pick one name. I've gotten used to `getContents`, but `getNodes` is also fine with me.
So if you prefer `getNodes`, I would deprecate `getContents` for removal to not have this duplication.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/block/Block.java
```diff
@@ -41,4 +43,18 @@
...
+
+    /**
+     * Returns the successors of this basic block.
+     *
+     * @return the successors of this basic block
+     */
+    Set<Block> getSuccessors();
+
+    /**
+     * Returns the nodes contained within this basic block.
```
wmdietl, line 55: At the moment there is no documentation which blocks return something non-empty or with just a single element. Maybe documenting that would be useful with the individual implementations of this method.

mernst: I added that documentation.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/block/RegularBlockImpl.java
mernst: Good suggestion.  I did that, and that led me to notice other cleanups that could be performed (eg, a client that implemented getNodes and can now use the new method).  I also added a `getLastNode()` method since multiple clients implemented it.

I chose "getNodes" because I think it's more descriptive to clients.  (I see why getContents was chosen -- because a private field is named `contents`.)

Do you want to review the latest changes, or should I merge this?

PR #3539, opened 2020-07-31
Run Error Prone on every compile.

smillst:


smillst:
Don't merge until I check that ErrorProne is not in checker.jar.

mernst:
> Don't merge until I check that ErrorProne is not in checker.jar.

Confirmed.

PR #3540, opened 2020-07-31
Remove dead nodes from treeLookup when constructing CFG

mernst:
Merge after ~https://github.com/typetools/checker-framework/pull/3536~ and ~https://github.com/typetools/checker-framework/pull/3538~.


dataflow/src/main/java/org/checkerframework/dataflow/cfg/ControlFlowGraph.java
```diff
@@ -97,6 +96,7 @@ public ControlFlowGraph(
         this.returnNodes = returnNodes;
         this.declaredClasses = declaredClasses;
         this.declaredLambdas = declaredLambdas;
+        removeDeadNodesFromTreeLookup();
```
wmdietl, line 99: This change should be done when the CFG is built, not in this class.
`CFGBuilder.CFGTranslationPhaseThree` removes elements from a CFG, so it is probably the place that should correctly update the `treeLookup` information.

mernst: I tried doing it elsewhere, and it was uglier.  Doing the work in `CFGTranslationPhaseThree` would leave phase 2 making CFGs with dead nodes in its treeLookup field.  I would prefer that each stage creates a valid CFG.

wmdietl: I completely agree with `I would prefer that each stage creates a valid CFG.`, but that makes this the wrong solution.
Class `ControlFlowGraph` is supposed to model the final CFG, after all changes are done in the `CFGBuilder`.
`CFGBuilder.CFGTranslationPhaseThree` calls this constructor, so why is it more work to add the hack there?
It's still a hack, but at least it is in the CFGBuilder.

mernst: > This change should be done when the CFG is built, not in this class.

I don't understand this comment.  The line of code we are commenting on is in the constructor for ControlFlowGraph.  So it is definitely being done when the CFG is built.

mernst: > I completely agree with `I would prefer that each stage creates a valid CFG.`, but that makes this the wrong solution.

Can you say why it is wrong?

> Class `ControlFlowGraph` is supposed to model the final CFG, after all changes are done in the `CFGBuilder`.

Can you point me at documentation that states that?  ControlFlowGraph is used as the output of phase 2, not just the output of CFGBuilder.  So it also models the output of phase 2.

mernst: > This change should be done when the CFG is built, not in this class.

From discussion, I think this comment means "This change should be done in the CFGBuilder, before the ControlFrowGraph constructor is called."

mernst: @wmdietl The implementation of `removeDeadNodesFromTreeLookup` uses member methods `getBlocks` and `getNodes`.  Moving the definition of `removeDeadNodesFromTreeLookup` to `CFGBuilder` would require coping `getBlocks` and `getNodes` to `CFGBuilder`, or making the methods static so they can be called from `CFgBuilder`, or similar changes.  Do you think that it is still better to move the call, given the need for those changes?  Which of those code refactorings did you have in mind?

PR #3545, opened 2020-08-02
More concise visualization of empty store

mernst:


framework/src/main/java/org/checkerframework/framework/flow/CFAbstractStore.java
```diff
@@ -1024,11 +1024,14 @@ public String visualize(CFGVisualizer<?, S, ?> viz) {
...
-        StringBuilder sbVisualize = new StringBuilder();
-        sbVisualize.append(castedViz.visualizeStoreHeader(this.getClass().getSimpleName()));
-        sbVisualize.append(internalVisualize(castedViz));
-        sbVisualize.append(castedViz.visualizeStoreFooter());
-        return sbVisualize.toString();
+        String header = castedViz.visualizeStoreHeader(this.getClass().getSimpleName());
+        String internal = internalVisualize(castedViz);
+        String footer = castedViz.visualizeStoreFooter();
+        if (internal.trim().isEmpty()) {
+            return header.replaceAll("\\s+$", "") + footer.replaceAll("^\\s+", "");
```
wmdietl, line 1031: This is removing leading/trailing spaces, right? Would just calling `trim` on the result be easier? A comment might be useful to document the intent.

mernst: It removes trailing spaces from `header` and leading spaces from `footer`.  That is different than `trim()`.  I added a comment.

PR #3550, opened 2020-08-03
Correct implementation of IdentityMostlySingleton.

### LINKED ISSUE #3484: Local variable gets incorrectly unrefined in finally block

smillst:
This change on `t-rasmud/repro-casting-issue` branch makes the test case in for #3484 pass.  I was also reproduce the issue using an existing test checker.

Fixes #3484.


wmdietl:
This is a cool find! I wonder whether we could extend the Interning Checker to allow us to prevent this bug...

I've created #3558 with an alternative, that might be a tad more efficient and I find it a bit clearer what the actual problem is.
It might still benefit from some comment. See what you think and comment there.

PR #3552, opened 2020-08-03
Use StringJoiner and join()

mernst:
~Merge after #3564~

Merge with https://github.com/typetools/checker-framework-inference/pull/169

### DIFF
--- a/dataflow/build.gradle
+++ b/dataflow/build.gradle
@@ -1,6 +1,7 @@
 dependencies {
     implementation project(':javacutil')
     implementation project(':checker-qual')
+    implementation 'org.plumelib:plume-util:1.1.5'
 }
 
 shadowJar {
--- a/framework/src/main/java/org/checkerframework/framework/stub/StubGenerator.java
+++ b/framework/src/main/java/org/checkerframework/framework/stub/StubGenerator.java
@@ -25,6 +25,7 @@ import javax.lang.model.util.ElementFilter;
 import org.checkerframework.javacutil.ElementUtils;
 import org.checkerframework.javacutil.SystemUtil;
 import org.checkerframework.javacutil.TypesUtils;
+import org.plumelib.util.UtilPlume;
 
 /**
  * Generates a stub file from a single class or an entire package.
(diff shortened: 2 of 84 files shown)


dataflow/build.gradle
```diff
@@ -48,7 +48,7 @@ task liveVariableTest(dependsOn: compileTestJava, group: 'Verification') {
             if (!JavaVersion.current().java9Compatible) {
                 jvmArgs += "-Xbootclasspath/p:${configurations.javacJar.asPath}"
             }
-            classpath = sourceSets.test.compileClasspath
+            classpath = sourceSets.test.runtimeClasspath
```
wmdietl, line 51: I guess these are leftovers that are already in master?

framework/src/main/java/org/checkerframework/framework/stub/StubGenerator.java
```diff
@@ -367,21 +368,14 @@ private void indent() {
...
-        boolean isFirst = true;
-        for (Object o : lst) {
-            if (!isFirst) {
-                sb.append(", ");
-            }
-            sb.append(o);
-            isFirst = false;
-        }
-        return sb.toString();
+        return UtilPlume.join(", ", lst);
```
wmdietl, line 378: Why can this one not use `SystemUtil.join`? Should we add a corresponding overload, to have one place to look for such `join` methods, instead of looking for either `SystemUtil` or `UtilPlume`?

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -31,6 +31,7 @@
 import org.checkerframework.checker.interning.qual.Interned;
 import org.checkerframework.checker.nullness.qual.Nullable;
 import org.checkerframework.dataflow.qual.SideEffectFree;
+import org.plumelib.util.UtilPlume;
```
wmdietl, line 34: Same question about `UtilPlume`.

dataflow/build.gradle
mernst: No, this is a bug fix that was exposed by these changes.

framework/src/main/java/org/checkerframework/framework/stub/StubGenerator.java
mernst: Yes, it can use SystemUtil.join, but SystemUtil.join delegates to UtilPlume and this avoid the extra step.  I feel that eventually it would be good to remove the duplication and have only one version of the method.

mernst:
@wmdietl I gave some answers (and a question about whether you want this split into two pull requests).

dataflow/build.gradle
smillst: These fixes were already merged to master, but master had not been merged into this branch.  I fixed that.

framework/src/main/java/org/checkerframework/framework/stub/StubGenerator.java
mernst: https://github.com/typetools/checker-framework/pull/3564 takes a step toward that.  I would like to reduce usage of plume-util in favor of Guava, in the future.

mernst:
> I recommend that this pull request either adds calls to SystemUtil or adds calls to UtilPlume, but not both.

I had done that in a different pull request, #3564.  It is now merged (and merged into this pull request).

PR #3557, opened 2020-08-04
Update Checker Framework Gradle plugin version.

smillst:


mernst:
@kelloggm @smillst The CI failure is

```
A problem occurred configuring root project 'lombok'.
> Could not create task ':compileJava'.
   > Project#afterEvaluate(Closure) on root project 'lombok' cannot be executed in the current context.
```

which occurs under JDK 8 only.

docs/examples/lombok/build.gradle
```diff
@@ -6,7 +6,7 @@ plugins {
     id 'java'
     id "io.freefair.lombok" version "5.1.0"
     // Checker Framework pluggable type-checking
-    id 'org.checkerframework' version '0.5.5' apply false
+    id 'org.checkerframework' version '0.5.8' apply false
```
smillst, line 9: version 0.5.6 is the last version that worked with the lombok plugin.

mernst: https://github.com/kelloggm/checkerframework-gradle-plugin claims to work with the Lombok plugin.
(Ah, maybe your information is from experimentation rather than the plugin's documentation.)

smillst: Yes, I tried `0.5.7` and it fails with the same error. I tried `0.5.5` and it works as expected.  So, some change between 0.5.6 and 0.5.7 broke this example.

PR #3558, opened 2020-08-04
Optimize IdentityMostlySingleton

### LINKED ISSUE #3484: Local variable gets incorrectly unrefined in finally block

wmdietl:


dataflow/src/main/java/org/checkerframework/dataflow/util/IdentityMostlySingleton.java
```diff
@@ -21,21 +22,30 @@ public IdentityMostlySingleton(T value) {
...
                 state = State.SINGLETON;
                 value = e;
                 return true;
             case SINGLETON:
+                if (value == e) {
+                    return false;
+                }
                 state = State.ANY;
+                // Use ArrayList, but make sure to use reference comparisons.
                 set = new ArrayList<>();
```
mernst, line 37: I am partial to the version in #3550 because the variable named `set` is a set, and because it's shorter -- it doesn't need to re-implement logic for `add` and `contains` and therefore I find it easier to read and understand.  As a minor benefit, #3550 type-checks under the Nullness Checker without the need for assumptions about use of the `set` variable in other subclasses of `AbstractMostlySingleton` (namely, `@MonotonicNonNull` on that variable).

dataflow/src/main/java/org/checkerframework/dataflow/util/IdentityMostlySingleton.java
```diff
@@ -21,21 +22,30 @@ public IdentityMostlySingleton(T value) {
...
             case EMPTY:
                 state = State.SINGLETON;
                 value = e;
                 return true;
             case SINGLETON:
+                if (value == e) {
+                    return false;
+                }
                 state = State.ANY;
+                // Use ArrayList, but make sure to use reference comparisons.
```
smillst, line 36: In the [other pull request](https://github.com/typetools/checker-framework/pull/3550#issuecomment-668849223), you said that you think using an array list is more efficient.  Can you explain more why you think it is more efficient? `MostlySingleton` uses a HashSet, should we change that one to make it more efficient, too? 

When I was debugging this code, I was very surprised that `set` was in fact an ArrayList.  I wondered why that choice was made. Given that using an ArrayList and using List#contains is not the correct behavior, it made more sense to me to use an identity hash set.  This also matches the implementation match MostlySingleton.

PR #3559, opened 2020-08-05
Don't typecheck dead code

mernst:
Typechecking dead code can lead to false positive warnings.
This pull request fixes a false positive warning in checker/tests/nullness/TryCatch.java.

~~I'm not sure why the checker-framework-inference tests are failing. @wmdietl Can you correct them?~~
Suzanne fixed this.

Merge with https://github.com/typetools/checker-framework-inference/pull/170.


mernst:
This attempt to fix `TryCatch.java` without #3550 failed; the CFG construction bug must be fixed before this can be merged.  A review of these diffs would still be welcome.

smillst:
`checker/tests/nullness/TryCatch.java.` is still showing the false positive.  I'm happy to take a look when the tests are passing again.

mernst:
@smillst The tests are passing, other than checker-framework-inference tests.  I don't know enough about that codebase to fix the tests.  typetools/checker-framework-inference also doesn't seem to be where development is happening on the CFI project.  https://github.com/opprop/checker-framework-inference has about 400 commits that typetools/checker-framework-inference lacks.  Thus I don't even know whether it would be desirable to fix the problem in typetools/checker-framework-inference, or whether it would be easier to do it on the more modern codebase in https://github.com/opprop/checker-framework-inference.  @wmdietl can you provide guidance about the status of typetools/checker-framework-inference?  Thanks!

dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java
```diff
@@ -195,6 +201,52 @@ private static void mergeTreeLookup(
...
+     * Returns true if the given tree is dead code (will never be executed at run time).
+     *
+     * <p>This method is conservative. If it returns true, the tree is dead code. If it returns
+     * false, the tree may or may not be dead code.
+     *
+     * @param tree a tree
+     * @return true if the tree is dead code
+     */
+    public boolean isDeadCode(Tree tree) {
+        if (!(tree instanceof ExpressionTree)) {
```
smillst, line 214: You could use TreeUtils#isExpressionTree. (Though the implementation of that method is `tree instanceof ExpressionTree`.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java
```diff
@@ -195,6 +201,52 @@ private static void mergeTreeLookup(
...
+        }
+        switch (tree.getKind()) {
+            case MEMBER_SELECT:
+                return false;
+            case ASSIGNMENT:
+                // This may be an expression or a statement.
+                return false;
+            case IDENTIFIER:
+            case ARRAY_ACCESS:
+                // There might be a missing "this." prefix.  (Can I test that?)
```
smillst, line 225: I don't understand what a missing "this" prefix is.  Do you mean a `this` tree created by dataflow for a field access?

dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java
```diff
@@ -195,6 +201,52 @@ private static void mergeTreeLookup(
...
+     */
+    public boolean isDeadCode(Tree tree) {
+        if (!(tree instanceof ExpressionTree)) {
+            return false;
+        }
+        switch (tree.getKind()) {
+            case MEMBER_SELECT:
+                return false;
+            case ASSIGNMENT:
+                // This may be an expression or a statement.
```
smillst, line 221: AssignmentTrees represent assignment expression; see their Javadoc.  `int x = 0;` is not an AssignmentTree, it's a VariableTree.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1618,6 +1618,19 @@ protected void checkAndPerformFlowAnalysis(Tree tree) {
...
     }
 
+    @Override
+    public boolean isDeadCode(@Nullable Tree tree) {
+        if (tree == null || flowResult == null || analysis.isRunning()) {
+            return false;
+        } else {
+            boolean result = flowResult.isDeadCode(tree);
+            if (result) {
+                // System.out.printf("dead: %s %s%n", tree.getKind(), tree);
```
smillst, line 1628: Should this be removed?

smillst:
@mernst I was able to fix the CFI failure.  See https://github.com/typetools/checker-framework-inference/pull/170.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java
mernst: An assignment expression `x = 0` is also a legal statement.  This not a remark about variable declarations such as `int x = 0`.  I'll remove the comment since it doesn't seem to shed light.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java
mernst: This is explaining why both these cases, and MEMBER_SELECT, are all necessary.  I'll remove it.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java
```diff
@@ -195,6 +201,45 @@ private static void mergeTreeLookup(
...
+     * <p>This method is conservative. If it returns true, the tree is dead code. If it returns
+     * false, the tree may or may not be dead code.
+     *
+     * @param tree a tree
+     * @return true if the tree is dead code
+     */
+    public boolean isDeadCode(Tree tree) {
+        // Return false for statements.
+        if (!TreeUtils.isExpressionTree(tree)) {
+            return false;
```
smillst, line 216: Why are statements never dead code?

mernst: `isDeadCode` is retrieving information about expressions from dataflow.  I didn't see an easy way to determine whether a statement is dead code, so `isDeadCode` conservatively returns false for them.  The spec of `isDeadCode` always permits returning false.  (I can put some subset, or all, of this in a comment.)

smillst: That makes sense.  A comment here about dataflow have no information about statements would be helpful.  
Does dataflow also not have information about the kinds of trees in the switch below?

PR #3563, opened 2020-08-06
Remove spurious enclosing class for types of NewClassTrees and provide sane error messages

### LINKED ISSUE #3266: Spurious enclosing class for types of NewClassTrees
The fix for #3057 adds an enclosing class for types that do not have them.

```

public class Enclosing  {
    public Enclosing() {}
    void f(){
        @Untainted Enclosing o = new Enclosing();
    }
}
```

```
Enclosing.java:6: error: [assignment.type.incompatible] incompatible types in assignment.
        @Untainted Enclosing o = new Enclosing();
                                 ^
  found   : @Tainted Enclosing.@Tainted Enclosing
  required: @Untainted Enclosing
1 error
```

PRITI1999:
Fixes #3266;

The function `isMostEnclosingThisDeref()` returns true for new Object creation of the same class leading to the spurious enclosing class.

Fix this by checking tree is not a `NewClassTrees` and then invoking `isMostEnclosingThisDeref()`

Code:
```

class C<T> {
    C<String> ls = new C<@Nullable String>();
}
```

Old error
```
Error: [assignment.type.incompatible] incompatible types in assignment.
   found   : @UnderInitialization(java.lang.Object.class) @NonNull C<T extends @Initialized @Nullable Object>.@Initialized @NonNull C<@Initialized @Nullable String>
   required: @Initialized @NonNull C<@Initialized @NonNull String>
```
New error
```
checker/tests/all-systems/Issue3326.java:4: error: [assignment.type.incompatible] incompatible types in assignment.
    C<String> ls = new C<@Nullable String>();
                   ^
  found   : @Initialized @NonNull C<@Initialized @Nullable String>
  required: @Initialized @NonNull C<@Initialized @NonNull String>
```


mernst:
Thanks for looking into this issue.  I appreciate it.

Your code change resolves the given problem.  However, I am concerned that a hard-coded check for `NEW_CLASS` here may not be the best solution.

A few lines earlier is this code:

```
        if (!ElementUtils.hasReceiver(element)) {
            return null;
        }
```

The `new` expression in your program has no receiver, so I would have expected that `return` statement to be executed.
In other words, there seems to be a bug in `hasReceiver`.

Could you look into whether there is a bug in `hasReceiver` and how to fix it?  I suspect that might improve more of the Checker Framework, and would make the change you have proposed unnecessary.  Thanks.

PRITI1999:
@mernst The element sent to `hasReceiver` in this case is of kind `CONSTRUCTOR`. The `hasReceiver` method explicitly checks for elements of kind `CONSTRUCTOR` and returns `true`. Probably there is a missing condition to check if the constructor was called from the same class to which it belongs as you mentioned this `new` expression does not have a receiver. The only parameter of `hasReceiver` is `Element` and I could not yet figure out the missing condition with only information of `Element`.

smillst:
Thanks for attempting to fix this issue!  While fixing another issue, I completely rewrote `getImplicitReceiverType`; see #3659 .  That also fixed this #3266, so I'm going to close this pull request.

PR #3564, opened 2020-08-06
Deprecate SystemUtil.join

mernst:
System.join's added functionality over UtilPlume.join (handling null arguments) is never used.

More generally, it is a goal to use libraries rather than re-inventing them:
JDK > Guava > plume-lib > local re-implementations

Merge with https://github.com/typetools/checker-framework-inference/pull/167

### DIFF
--- a/checker/build.gradle
+++ b/checker/build.gradle
@@ -17,6 +17,7 @@ dependencies {
     // File Utilities takes priority over this version, in the fat jar
     // file. :-( So update it and re-build it locally when updating this.
     implementation 'org.plumelib:reflection-util:0.2.2'
+    implementation 'org.plumelib:plume-util:1.1.5'
 
     testImplementation group: 'junit', name: 'junit', version: '4.13'
     testImplementation project(':framework-test')
--- a/dataflow/build.gradle
+++ b/dataflow/build.gradle
@@ -1,6 +1,7 @@
 dependencies {
     implementation project(':javacutil')
     implementation project(':checker-qual')
+    implementation 'org.plumelib:plume-util:1.1.5'
 }
 
 shadowJar {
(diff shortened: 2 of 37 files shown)


dataflow/build.gradle
```diff
@@ -1,6 +1,7 @@
 dependencies {
     implementation project(':javacutil')
     implementation project(':checker-qual')
+    implementation 'org.plumelib:plume-util:1.1.5'
```
smillst, line 4: This dependency needs to be added to `checker-framework/docs/developer/release/maven-artifacts/dataflow-pom.xml`.  (Once we start using the Maven publish gradle plugin, the pom file will be automatically generated and we won't have to do this.)

framework-test/build.gradle
```diff
@@ -12,6 +12,8 @@ dependencies {
     implementation project(':javacutil')
     implementation project(':checker-qual')
 
+    implementation 'org.plumelib:plume-util:1.1.5'
```
smillst, line 15: This dependency needs to be added to `checker-framework/docs/developer/release/maven-artifacts/framework-test-pom.xml`.

checker/build.gradle
```diff
@@ -17,6 +17,7 @@ dependencies {
     // File Utilities takes priority over this version, in the fat jar
     // file. :-( So update it and re-build it locally when updating this.
     implementation 'org.plumelib:reflection-util:0.2.2'
+    implementation 'org.plumelib:plume-util:1.1.5'
```
smillst, line 20: Just to be clear, do NOT add this to checker-framework/docs/developer/release/maven-artifacts/checker-pom.xml.  That is the pom file for the fat jar that includes all dependencies.

PR #3571, opened 2020-08-06
Check for task name correctly so that ErrorProne runs.

smillst:


dataflow/src/main/java/org/checkerframework/dataflow/cfg/ControlFlowGraph.java
```diff
@@ -235,6 +235,7 @@ public UnderlyingAST getUnderlyingAST() {
      * @param cur a block whose successors to retrieve
      * @return a Deque of successor Blocks
      */
+    @SuppressWarnings("UnusedMethod")
```
smillst, line 238: @mernst You add this method, should we just delete it? Or suppress the warning?

mernst: I deleted it in d5d90696b9e69a74cdca29ef340f5ec551e9d531; a bad merge brought it back in a74139468e65c8810ae61a9bac1f22b93795b421.  I have deleted it again.

PR #3572, opened 2020-08-07
Make Error Prone versions consistent

mernst:


framework/build.gradle
```diff
@@ -39,7 +39,7 @@ dependencies {
     // TODO: org.checkerframework.annotatedlib:guava:28.2-jre requires the below dependency.
     // Get the version number from the "compile dependencies" section of
     // https://mvnrepository.com/artifact/com.google.guava/guava/28.2-jre .
-    implementation 'com.google.errorprone:error_prone_annotations:2.3.4'
+    implementation 'com.google.errorprone:error_prone_annotations:2.4.0'
```
smillst, line 42: This seems to be in conflict with the comment above the changed line.  So either we shouldn't make this change or the comment here needs to change.

PR #3573, opened 2020-08-07
Correct GenericAnnotatedTypeFactory#getStoreAfter(Node)

### LINKED ISSUE #3566: Store returned by GenericAnnotatedTypeFactory.getStoreAfter() sometimes missing expected variables
This issue was observed within the Object Construction Checker.  Example:
```java
class Test {
    class Foo {
    }
    Foo makeFoo(){
        return new Foo();
    }
    void test() {
        Foo foo = makeFoo();
    }
}
```
We are overriding `GenericAnnotatedTypeFactory#postAnalyze` and analyzing the CFG for `test()`.  Given the `AssignmentNode` `n` for `Foo foo = makeFoo()`, we invoke `GenericAnnotatedTypeFactory.getStoreAfter(n)` and expect to find an entry in the `CFStore` for `foo`.  But, `foo` is missing from the store.  Interestingly, if we invoke `getAnnotatedType(n.getTree())` we do get the expected type for `foo`.  We would expect this same type to be in the store immediately after the statement.

To reproduce, check out this branch:

https://github.com/msridhar/object-construction-checker/tree/repro-store-bug

And then run:
```
./gradlew :object-construction-checker:test --tests AlwaysCallTest
```
The test will fail with a RuntimeException that is thrown when the above case is encountered.

smillst:
Make the implementation of `GenericAnnotatedTypeFactory#getStoreAfter(Node)` match the implementations of `#getStoreAfter(Tree)`, `#getStoreBefore(Node)`, and `#getStoreBefore(Tree)`.

The Checker Framework doesn't have any calls to `#getStoreAfter(Node)`, so there isn't an easy way to add a test case for this.

@msridhar I think this fixes #3566.  (Or at least with this fix, the RuntimeException in  "repro-store-bug" is not hit...). Can you confirm that this fixes the whole issue?

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
@@ -1019,6 +1019,9 @@ public abstract class GenericAnnotatedTypeFactory<
      * @return the store immediately after a given {@link Node}
      */
     public Store getStoreAfter(Node node) {
+        if (!analysis.isRunning()) {
+            return flowResult.getStoreAfter(node);
+        }
         Store res =
                 AnalysisResult.runAnalysisFor(
                         node,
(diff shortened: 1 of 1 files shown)


msridhar:
@smillst I've confirmed that with this PR the issue we were seeing is fixed, and the type we obtain from calling `getStoreAfter` and looking up the variable type in the resulting store is always consistent with what we obtain from the type factory.  Thanks! /cc @Nargeshdb

PR #3575, opened 2020-08-08
checkRep() method for ControlFlowGraph

mernst:


framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1453,15 +1453,27 @@ public AnnotatedTypeMirror getAnnotatedTypeLhs(Tree lhsTree) {
     /**
      * Returns the type of a varargs array of a method invocation or a constructor invocation.
      *
+     * <p>A null return value might mean:
+     *
+     * <ul>
+     *   <li>This type-checker is not using dataflow analysis.
+     *   <li>The given tree is dead code, so dataflow analysis never reached it to refine its type.
```
smillst, line 1460: This method is only using the `Node` to get the synthetic tree.  The returned type is computed using the tree, not the dataflow value.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/ControlFlowGraph.java
```diff
@@ -330,4 +364,21 @@ public String toStringDebug() {
         result.add("declaredLambdas=" + declaredLambdas);
         return result.toString();
     }
+
+    /** Checks representation invariants on this. */
+    public void checkRep(@UnknownInitialization(ControlFlowGraph.class) ControlFlowGraph this) {
+        List<Node> allNodes = getAllNodes();
+
+        // Require that each node in treeLookup exists in this ControlFlowGraph
```
smillst, line 372: This is a new invariant. It's useful to know to which nodes a tree maps to even if the tree is in dead code.  (See the other comment for an example.)  Can you explain more why this new invariant is worth the loss of information?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/ControlFlowGraph.java
```diff
@@ -203,6 +216,27 @@ public UnderlyingAST getUnderlyingAST() {
...
+        // Remove references to dead code.
+        for (Set<Node> nodeSet : treeLookup.values()) {
+            for (Iterator<Node> i2 = nodeSet.iterator(); i2.hasNext(); ) {
+                Node n = i2.next();
+                if (!allNodes.contains(n)) {
+                    i2.remove();
+                }
+            }
+            // nodeSet, which is treeLookup.get(tree) for the tree, might be empty.  That indicates
+            // that the tree is in dead code, as opposed to being a tree that doesn't produce a
```
smillst, line 235: This change should be reflected in the Javadoc of `treeLookup`.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/ControlFlowGraph.java
mernst: It would be less expensive to precompute the information once, if in the future clients want to know whether an expression is dead.

But, if you think the change is a bad idea we can drop the pull request.  (There should still be some kind of documentation for this data structure, which I found underdocumented and confusing.)

PR #3578, opened 2020-08-08
Treat Name as interned when type-checking the Checker Framework

mernst:
Also contains some cleanup, which could be separated into a different pull request if desired.


build.gradle
```diff
@@ -307,7 +307,12 @@ def createCheckTypeTask(projectName, checker, shortName, args = []) {
                 '-Xmaxerrs', '10000',
                 '-Xmaxwarns', '10000',
                 '-ArequirePrefixInWarningSuppressions',
-                ]
+        ]
+        if (shortName.equals("Interning")) {
```
smillst, line 311: This should be passed when creating the `checkIntering` class:
`createCheckTypeTask(project.name, 'org.checkerframework.checker.interning.InterningChecker', 'Interning', ['-Astubs=javax-lang-model-element-name.astub'])`

framework/src/main/java/org/checkerframework/common/reflection/DefaultReflectionResolver.java
```diff
@@ -506,7 +506,6 @@ private boolean isUnknownMethod(MethodInvocationTree tree) {
      * @param env the environment
      * @return the (potentially empty) set of corresponding method Symbol(s)
      */
-    @SuppressWarnings("interning:not.interned") // bug?
```
smillst, line 509: The errors in this method are because `com.sun.tools.javac.util.Name` is used. `com.sun.tools.javac.util.Name` implements  `javax.lang.model.element.Name`, so I think we can also annotate it as `@Interned`.

PR #3580, opened 2020-08-08
Add toStringDebug methods

mernst:


dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGBuilder.java
```diff
@@ -370,6 +371,13 @@ public void setBlock(BlockImpl b) {
         public String toString() {
             throw new BugInCF("DO NOT CALL ExtendedNode.toString(). Write your own.");
         }
+
+        /**
+         * Returns a string representation of this.
```
smillst, line 376: This comment doesn't differentiate it from `toString()`.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGBuilder.java
```diff
@@ -1501,6 +1543,32 @@ private String resolveLabel(Label label) {
             }
             return nodeToString(nodeList.get(index));
         }
+
+        /**
+         * Returns a string representation of this.
```
smillst, line 1548: This comment doesn't differentiate it from `toString()`.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGBuilder.java
```diff
@@ -545,6 +568,25 @@ public Label getLabel() {
...
+        }
+    }
+
+    /**
+     * Return a printed representation of a collection of extended nodes.
+     *
+     * @param nodes a collection of extended nodes to format
+     * @return a printed representation of the given collection
+     */
+    public static String extendedNodeCollectionToString(Collection<? extends ExtendedNode> nodes) {
```
smillst, line 584: I think it would be clearer to call this `extendedNodeCollectionToStringDebug`.  Then it's clear in `PhaseOneResult#toStringDebug` why `PhaseOneResult#toString` isn't called.

PR #3583, opened 2020-08-10
Decode annotations on enclosing types correctly

### LINKED ISSUE #3562: Inability to write annotations on the outer class portion of an inner class name
Dataflow seems to break with inner class types.
```java

public class C {
    void method(@Tainted Inner param) {
        @Tainted C. @Untainted Inner local = param;
    }

    class Inner {
    }
}
```
I get this error,
```
C.java:5: error: [type.invalid.conflicting.annos] invalid type: conflicting annotations [@Tainted, @Untainted] in type "@Tainted C.@Tainted @Untainted Inner"
        @Tainted C. @Untainted Inner local = param;
                                     ^
```
I would expect to instead get an invalid assignment error. It seems dataflow tries to replace the inner type but doesn't remove the old annotation.

smillst:
Fixes #3562 .

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
@@ -6,11 +6,15 @@ import com.sun.source.tree.MethodTree;
 import com.sun.source.tree.Tree;
 import com.sun.source.tree.Tree.Kind;
 import com.sun.source.tree.VariableTree;
+import java.util.ArrayList;
 import java.util.List;
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.type.DeclaredType;
+import javax.lang.model.type.TypeKind;
+import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedDeclaredType;
 import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedExecutableType;
 import org.checkerframework.framework.util.AnnotatedTypes;
 import org.checkerframework.javacutil.AnnotationUtils;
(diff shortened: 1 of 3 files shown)


framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
```diff
@@ -26,17 +30,52 @@
...
+        }
+
+        if (result.getKind() == TypeKind.DECLARED
+                &&
+                // Annotations on enum constants are not in the TypeMirror and always apply to the
+                // innermost type, so handle them in the else block.
+                elt.getKind() != ElementKind.ENUM_CONSTANT) {
+
+            // Decode the annotations from the type mirror because the annotations are already in
+            // the correct place for enclosing types. For example, @Tainted Outer.Inner and @Tainted
```
mernst, line 56: "For example" sounds like it is following up on a previous point, but I think here it might be introducing a new point.

framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
```diff
@@ -26,17 +30,52 @@
...
+            // Inner. For both types @Tainted is stored in node.getModifiers(), but they apply to
```
mernst, line 57: Are they stored on the same node (that is, both are stored on the node for `Inner`)?  If so, I would say that.  If not, the fact that they apply to different types may not be so surprising.

framework/src/main/java/org/checkerframework/framework/type/TypeFromMemberVisitor.java
```diff
@@ -26,17 +30,52 @@
...
+            AnnotatedTypes.applyAnnotationsFromDeclaredType(annotatedDeclaredType, declaredType);
+
+            // Handle declaration annotations
+            for (AnnotationMirror anno : modifierAnnos) {
+                if (AnnotationUtils.isDeclarationAnnotation(anno)) {
+                    // This does not treat Checker Framework compatqual annotations differently,
+                    // because it's not clear whether the annotation should apply to the outermost
+                    // enclosing type or the innermost.
+                    result.addAnnotation(anno);
+                } // If anno is not a declaration annotation, it should have been applied in the
```
mernst, line 72: I don't understand why this comment is attached to the curly brace rather than on a separate line.  Can you clarify why?

framework/src/main/java/org/checkerframework/framework/util/AnnotatedTypes.java
```diff
@@ -1444,4 +1445,25 @@ public static void copyOnlyExplicitConstructorAnnotations(
             }
         }
     }
+
+    /**
+     * Add all the annotations in {@code declaredType} to {@code annotatedDeclaredType}.
```
mernst, line 1450: Maybe state explictly that this includes both declaration and type annotations.

PR #3587, opened 2020-08-11
Improve ControlFlowGraph documentation

mernst:


framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1457,9 +1457,17 @@ public AnnotatedTypeMirror getAnnotatedTypeLhs(Tree lhsTree) {
     /**
      * Returns the type of a varargs array of a method invocation or a constructor invocation.
      *
+     * <p>A null return value might mean:
+     *
+     * <ul>
+     *   <li>This type-checker is not using dataflow analysis.
+     *   <li>The given tree is dead code, so dataflow analysis never reached it to refine its type.
```
smillst, line 1464: This is incorrect.  This method does not use the refined type from dataflow. Rather it use the synthetic NewArrayTree created by dataflow and computes a type of that NewArrayTree via the AnnotatedTypeFactory.  It does not return null if the method call is in dead code.

This method only returns null if useFlow is false.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/ControlFlowGraph.java
```diff
@@ -55,10 +55,18 @@
...
+     * Maps from AST {@link Tree}s to sets of {@link Node}s.
+     *
+     * <ul>
+     *   <li>Most Trees that produce a value will have at least one corresponding Node.
+     *   <li>Trees that undergo conversions, such as boxing or unboxing, can map to two distinct
+     *       Nodes. The Node for the pre-conversion value is stored in {@link #treeLookup}, while
+     *       the Node for the post-conversion value is stored in {@link #convertedTreeLookup}.
+     * </ul>
+     *
+     * Some of the mapped-to nodes (in both maps) do not appear in {@link #getAllNodes} because
```
smillst, line 67: I was confused about which maps the "both maps" was referring to.  I think it's `treeLookup` and `convertedTreeLookup`.

PR #3593, opened 2020-08-14
Add test cases for -AwarnUnneededSuppressions

mernst:


checker/jtreg/index/UnneededSuppressionsTest.java
```diff
@@ -36,4 +36,12 @@ void method5() {}
 
     @SuppressWarnings("purity")
     void method6() {}
+
+    // TODO: It is a bug that this is not reported as an unused suppression
+    @SuppressWarnings("index:foo.bar.baz")
+    void method7() {}
+
+    // TODO: It is a bug that this is not reported as an unused suppression
+    @SuppressWarnings("allcheckers:purity.not.deterministic.call")
```
smillst, line 45: A checker besides the Index Checker could use this warning suppression, so it isn't reported.

checker/jtreg/index/UnneededSuppressionsTest.java
```diff
@@ -36,4 +36,12 @@ void method5() {}
 
     @SuppressWarnings("purity")
     void method6() {}
+
+    // TODO: It is a bug that this is not reported as an unused suppression
+    @SuppressWarnings("index:foo.bar.baz")
```
smillst, line 41: Because `foo.bar.baz` is not a message key for the Index Checker, this is not reported as unused.

checker/jtreg/index/UnneededSuppressionsTest.java
mernst: 1. Users get no warning about unused warning suppressions with the allcheckers: prefix, which is not documented.
2. Most users run just a few checkers, so they can suppress the warning for each relevant checker.

Do you think the current undocumented behavior is better than the behavior proposed in this TODO?  (Your comments don't indicate your position on this.)

checker/jtreg/index/UnneededSuppressionsTest.java
mernst: 1. This behavior is not documented.
2. Do you believe this is desirable behavior?  It means that a user who typos a message gets no warning about that typo.

smillst: 1. This behavior was documented, but was removed 1abced10d7df1e80a7909fed92b6900739cb1c03.  I suggest we add back the line `These warnings are only issued if the \<@SuppressWarnings>
  contains a key that would suppress any warning from this checker.`  in the introduction.tex. 

2. The purpose of this option is to find warning suppressions that at one time did suppress a checker warning.  So at the time the suppression is written a checker is issuing a warning.  If a user tries to suppress that warning, but typos the suppression string, then the checker will still issue a warning.  And thus that warning tells the user there's a typo.

checker/jtreg/index/UnneededSuppressionsTest.java
smillst: I understood the TODO as a question of whether this is a bug or desired behavior.  My point was that it is not a bug that there was no warning on this suppression.  We designed this option to only issue warnings when the suppression was clearly not needed, so this is the desired behavior.

PR #3595, opened 2020-08-14
Refactor AnnotatedTypeCombiner to make it more extensible

kelloggm:
Makes AnnotatedTypeCombiner extensible, so that a type system implementer can easily change the combination function by overwriting one method (`combineAnnotations(AnnotatedTypeMirror, AnnotatedTypeMirror)`).

Also make the constructor `protected` rather than `private`, so that the class can actually be extended (and the extension can call the `super` constructor).


framework/src/main/java/org/checkerframework/framework/type/visitor/AnnotatedTypeCombiner.java
```diff
@@ -39,7 +39,7 @@ public static void combine(
      *
      * @param hierarchy the hierarchy used to the compute the GLB
```
mernst, line 40: Mentions of GLB and "greatest lower bound" in the documentation need to be generalized, just as the code has been generalized.

mernst (changes requested):

PR #3596, opened 2020-08-15
Add visitIdentifier() to ValueTreeAnnotator; fixes #3105

### LINKED ISSUE #3105: Field in ValueChecker cannot be refined.
Consider these two cases:

```java

class C {
    public static final String DEMO = "foo";
}

class Demo {  
  @StringVal("foo") String m() {
      return C.DEMO;
  }
}
```

```java

class C {
    public static final String DEMO = "foo";
}

class Demo extends C{  
  @StringVal("foo") String m() {
      return DEMO; // error
  }
}
```

The first one can pass the typechecking while the second one can't (error at `return DEMO;`):

```
Error: [return.type.incompatible] incompatible types in return.
   type of expression: @UnknownVal String
   method return type: @StringVal("foo") String
```

We should consider why the field in this case cannot be refined.

PRITI1999:
Fixes #3105;
Fields parsed as `IdentifierTree` could not be refined due to lack of `visitIdentifier()` method in `ValueTreeAnnotator`. 
Add `visitIdentifier()` method to `ValueTreeAnnotator` by reusing the logic of `visitMemberSelect` to refine fields.


mernst:
> Add `visitIdentifier()` method to `ValueTreeAnnotator` by reusing the logic of `visitMemberSelect` to refine fields.

This pull request only reuses some of the logic of `visitMemberSelect`.  Why doesn't it include logic for handlng the case that the identifier names a static final field?

smillst (approved):
@mernst I added a different way to compile Issue3105B.java before it is used.  I think the test now test the issue.

Issue #2059, opened 2018-06-29
Stubs missing imports produce unclear errors


nicbec:
Checker Framework version 2.5.2

We wanted to add a stub to our codebase that would annotate java.lang.reflect.Constructor in keeping with the Checker-Framework stubs version and the related issue below
https://github.com/typetools/checker-framework/issues/1365
Checker Framework stubs:
https://github.com/typetools/checker-framework/blob/460863bf1612581de76a783b4ad254b816bfe8b7/checker/jdk/nullness/src/java/lang/reflect/Constructor.java

I overlooked to add the @NonNull import to the stub. Static analysis tests that reference the stub below will compile and run, but will produce unusual errors ([assignment.type.incompatible], [return.type.incompatible], [type.argument.type.incompatible]) in classes that reference this stub that don't make it apparent that the real issue is a missing import.

It would be great to receive more clear messages for cases where stubs are missing imports.

Please let me know if other information is needed, thanks.
##
package java.lang.reflect;

// Missing import below.
// import org.checkerframework.checker.nullness.qual.NonNull;
import org.checkerframework.checker.nullness.qual.Nullable;

public final class Constructor<T> extends Executable {

  public @NonNull T newInstance(Object ... initargs) throws InstantiationException,IllegalAccessException,IllegalArgumentException,InvocationTargetException {}
}

wmdietl:
Can you try adding `-AstubWarnIfNotFound` as an option?

```
    // Whether to print warnings about types/members in a stub file
    // that were not found on the class path
    // org.checkerframework.framework.stub.StubParser.warnIfNotFound
    "stubWarnIfNotFound",
```

This should give you a warning about the unknown `@nonnull` annotation.

nicbec:
Thanks for the response. I added "-AstubWarnIfNotFound" and see warnings for other locations in our code base, but nothing referencing @NonNull, Constructor, newInstance, or java.lang.reflect. There are no new warnings between the output when I comment out/uncomment the NonNull import though there is an unclear "error: [assignment.type.incompatible] incompatible types in assignment" when the import is commented out.

nicbec:
Just wanted to check in on this to see if there's anything else I should try. Please let me know, thanks.

wmdietl:
Pull request https://github.com/typetools/checker-framework/pull/2072 should fix this.
Now, when using `-AstubWarnIfNotFound` one also gets a warning about unknown annotations.
Please give it a spin and let us know whether this solves the issue.

[...]

PR #3599, opened 2020-08-17
Restore ability to use unannotated stub file to override annotations in bytecode

### LINKED ISSUE #3597: Handling of annotation defaulting in stubs between 3.4.0 and 3.4.1
Prior to `3.4.0`, a stub file without annotations could be used to override annotations present in the corresponding classpath dependency (in the example below, the stub file for `p.B#f` has an implicitly non-null return type, which overrides the nullable type in the source file).

Was this change deliberate? The most likely culprit I see is https://github.com/typetools/checker-framework/commit/3ce0ab57300439da3646cdf5b93408440a048a2b. The repro doesn't use `-Ainfer`, and passing `-AmergeStubsWithSource` has no effect. If this change was deliberate, is there a way to restore the previous behaviour?

```
=== stubs/p.astub
package p;

class B {
  Object f();
}
=== p/B.java
package p;

class B {
  @Nullable
  Object f() {
    return new Object();
  }
}
=== p/A.java
package p;

class A {
  void f() {
    System.err.println(new B().f().toString());
  }
}
```

Compiling with `p.B` on the classpath, and with its corresponding stub, produces a diagnostic with `3.4.1` but not with `3.4.0`:

```
./checker-framework-3.4.0/checker/bin/javac p/B.java
./checker-framework-3.4.0/checker/bin/javac -Astubs=stubs -processor Nullness -sourcepath : -cp . p/A.java 
./checker-framework-3.4.1/checker/bin/javac -Astubs=stubs -processor Nullness -sourcepath : -cp . p/A.java 
p/A.java:5: error: [dereference.of.nullable] dereference of possibly-null reference new B().f()
    System.err.println(new B().f().toString());
                                ^
1 error
```

kelloggm:
Fixes #3597, which was an unintended side-effect of #2871. 

The actual fix is making the call to `clearAnnotations` in `StubParser#annotate` unconditional. Removing that code revealed some other problems, though, which this PR also contains fixes for.

The first other problem was that one of the tests for stub-based WPI (`UsesAnonymous.java`) failed with an extra `override.param.invalid` error. Examining the stub file made it clear that the generated stub was (mostly) correct, but that annotations that were actually present in the source code were being ignored (and it seems that the guard for the `clearAnnotations` call was a hack to avoid the problem), because they were printed as fully-qualified annotations. The stub parser, however, ignored any annotations that weren't imported directly. I therefore modified the stub parser to also read fully-qualified annotations in stub files.

That change fixed the initial failing test, but caused another WPI test case (`MultiDimensionalArrays.java`) to fail. Examining the generated stub files for that class made it clear that 3+ dimensional arrays were being printed incorrectly. The test hadn't failed before because the fully-qualified annotations weren't being used. I discovered that our belief about the order in which javac stores arrays was *wrong*; scene-lib and javac do actually store arrays in the same order. I'm not sure how we made that mistake. That allowed me to remove/simplify a bunch of code in `SceneToStubWriter`, though I did have to change the specification of `formatArrayTypeImpl`: it now only prints the "array" part of the type, and relies on `formatArrayType` to print the component type.


kelloggm:
The test failure on this PR is interesting. It's not clear to me that the behavior in this PR is wrong, so I'm going to leave the failing test (for now) and instead describe the behavior here. I'd appreciate input from reviewers on which behavior is preferable.

The failing test is the one for #2059, which checks that `-AstubWarnIfNotFound` issues an error if (e.g.) `org.checkerframework.checker.nullness.qual.NonNull` (only) is imported in a stub file, but `org.checkerframework.checker.nullness.qual.Nullable` is used. The test fails because instead of issuing one error for an unknown annotation, it issues two errors: the expected one for the unknown annotation and another "type not found" error.

The latter, new error is caused by the change in this PR that attempts to treat an unimported annotation as a fully-qualified name. In this case, what that means is that the stub parser tries to load the class `Nullable` (in the default package). Since this doesn't exist as an annotation, the "type not found" error is issued.

The new error is therefore sensible given what the stub parser is trying to do, but I also think it's somewhat annoying for users. It would require a bit of refactoring to avoid the error, but it's not technically challenging. The question is whether we want to expose to the user that the stub parser tried (and failed) to load the default package's `@Nullable`, or not.

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1204,6 +1185,18 @@ private void annotate(AnnotatedTypeMirror type, List<AnnotationExpr> annotations
         }
         for (AnnotationExpr annotation : annotations) {
             AnnotationMirror annoMirror = getAnnotation(annotation, allStubAnnotations);
+            if (annoMirror == null) {
+                // If the annotation was not imported, then #getAllStubAnnotations does
```
smillst, line 1189: This should be moved into `getAnnotation`.  Or is there some reason it can't be?

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -1204,6 +1185,18 @@ private void annotate(AnnotatedTypeMirror type, List<AnnotationExpr> annotations
...
+                // not add it to the allStubAnnotations field. This code compensates for
+                // that deficiency by adding the annotation when it is encountered (i.e. here).
+                TypeElement annoTypeElt = getTypeElement(annotation.getNameAsString());
```
smillst, line 1192: Just call `TypeElement classElement = elements.getTypeElement(typeName);` directly so that the more sensible warning will be issued below.  (Make sure to document why getTypeElement isn't called.)

framework/src/main/java/org/checkerframework/framework/stub/StubParser.java
```diff
@@ -142,9 +142,8 @@
     private final Elements elements;
 
     /**
-     * The set of annotations found in the stub file. Keys are simple (unqualified) names. (This may
-     * be a problem in the unlikely occurrence that a type-checker supports two annotations with the
-     * same simple name.)
+     * The set of annotations found in the stub file. Keys are names. There are two entries for each
```
smillst, line 145: In several places you deleted `simple`, instead, could you please change it to `both fully-qualified and simple names`.  That's important information.

framework/src/main/java/org/checkerframework/common/wholeprograminference/SceneToStubWriter.java
```diff
@@ -173,79 +173,29 @@ private static String formatAnnotations(Collection<? extends Annotation> annos)
      * @return the type formatted to be written to Java source code, followed by a space character
      */
     private static String formatArrayType(ATypeElement scenelibRep, ArrayType javacRep) {
-        int levels =
```
smillst, line 176: The changes in this file should be merged to master in a separate commit. Can you open a separate PR with these changes?   I'll merge that PR first, so it's fine to leave these changes in this PR.  Also, can you add a test case for this that is unrelated to the other changes in this file?

kelloggm: > The changes in this file should be merged to master in a separate commit

I agree that would be cleaner. The reason I didn't do it is:

> can you add a test case for this that is unrelated to the other changes in this file?

I cannot. The error here only affects the order of the parts of the type that are contained in the javac representation. With respect to arrays, that means that the only parts that are incorrectly ordered are the `[]`s - even the inferred annotations were ordered correctly before. The error only shows up when there are printable, but not inferred, annotations - which javac prints using fully-qualified names. So it's not possible to test these changes without the fix to the stub parser, but fixing the stub parser causes the existing tests to fail (because the algorithm is wrong).

kelloggm: > even the inferred annotations were ordered correctly before

What I mean by this is that the `sceneLibRep` and its associated structures in the old code caused inferred annotations on the inner arrays types to be printed in the correct order, even though the array parts themselves (the `[]`s and any annotations in the javac types) were printed in the wrong order.

smillst: Thanks for the explantation.  I understand why you can't add a test case (or rather why the existing test case didn't fail until the other bugs where fixed).   Could you still open a separate PR without a test case?  Then we can just merge that one first, then merge this one.

kelloggm: #3601 has these changes

PR #3600, opened 2020-08-18
Read inherited declaration annotations from stub files.

### LINKED ISSUE #3594: @Inherited annotations written in a stub file
If we have an @Inherited annotation written in a stub file, subclasses in source code will not inherit the annotation! 
e.g. Consider InheritedAnnotationType:
```
@Inherited
@Target(ElementType.TYPE)
public @interface InheritedAnnotationType {
 ...   
}
```

Add @InheritedAnnotationType to Socket in a stub file:

```
@InheritedAnnotationType
class Socket implements Closeable {...}

```
Then, SubSocket doesn't inherit @InheritedAnnotationType!
`class SubSocket extends Socket {}`

smillst:
- [x] TODO: Add a test case.

Fixes #3594.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
@@ -26,6 +26,7 @@ import com.sun.source.util.Trees;
 import com.sun.tools.javac.code.Type;
 import java.lang.annotation.Annotation;
 import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
 import java.lang.annotation.Target;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


msridhar:
@Nargeshdb tested this and it works great for our scenario

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -3318,6 +3340,45 @@ private AnnotationMirror getDeclAnnotation(
         return results;
     }
 
+    /**
+     * Adds into {@code results} the declaration annotations found in all elements of the super
```
msridhar, line 3344: Should this say it adds the declaration annotations that are marked to be inherited?

PR #3607, opened 2020-08-25
Make TryFinallyScopeMap.accessedFields deterministic

mernst:


dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGBuilder.java
```diff
@@ -844,10 +845,12 @@ public String toString() {
      */
     @SuppressWarnings("serial")
     protected static class TryFinallyScopeMap extends HashMap<Name, Label> {
+        /** New labels within a try block that were added by this implementation. */
         private final Map<Name, Label> accessedNames;
 
+        /** Create a new TryFinallyScopeMap. */
         protected TryFinallyScopeMap() {
-            this.accessedNames = new HashMap<>();
+            this.accessedNames = new LinkedHashMap<>();
```
smillst, line 853: I don't see any place where a TryFinallyScopeMap is iterated over.  Did I miss it or is there some other reason this needs to be a LinkedHashMap?

jwaataja: The fields are accessed [here](https://github.com/typetools/checker-framework/blob/c28ede31c365c472d60823ee34d5afa226d2d244/dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGBuilder.java#L4612) and [here](https://github.com/typetools/checker-framework/blob/c28ede31c365c472d60823ee34d5afa226d2d244/dataflow/src/main/java/org/checkerframework/dataflow/cfg/CFGBuilder.java#L4663). In the for loops following each of those lines, `extendWithExtendedNode` modifies a list that should be in deterministic order.

smillst: Thanks!

PR #3626, opened 2020-09-05
Support Janino nullanalysis annotations

mernst:


vlsi:
Does that mean `org.codehaus.commons.nullanalysis.NotNullByDefault` needs to be added somehow as well?

mernst:
> Does that mean `org.codehaus.commons.nullanalysis.NotNullByDefault` needs to be added somehow as well?

There is no analogous annotation in the Nullness Checker.  It always defaults unannotated types to `@NonNull`.  So, the Nullness Checker ignores that annotation.

PR #3639, opened 2020-09-08
Documentation improvements

mernst:


framework/src/main/java/org/checkerframework/framework/type/visitor/AnnotatedTypeCombiner.java
```diff
@@ -8,8 +8,8 @@
 import org.checkerframework.javacutil.BugInCF;
 
 /**
- * Combines all the annotations in the parameter with the annotations from the visited type, taking
- * the greatest lower bound at each point.
+ * Changes each parameter annotation to be the GLB of the parameter annotation and visited type
```
smillst, line 11: `parameter annotation` doesn't make sense because the type of the parameter is an AnnotatedTypeMirror.

smillst (approved):

framework/src/main/java/org/checkerframework/framework/type/visitor/AnnotatedTypeCombiner.java
```diff
@@ -7,10 +7,7 @@
 import org.checkerframework.javacutil.AnnotationUtils;
 import org.checkerframework.javacutil.BugInCF;
 
-/**
- * Combines all the annotations in the parameter with the annotations from the visited type, taking
- * the greatest lower bound at each point.
- */
+/** Changes each parameter type to be the GLB of the parameter ytpe and visited type. */
```
smillst, line 10: `ytpe`

Issue #3630, opened 2020-09-06
Value Checker cannot find class for @StaticallyExecutable on static methods


ThiagoTGM:
Trying to use the `@StaticallyExecutable` on a static method doesn't seem to work, it always fails to find the class.
The class I'm trying to compile is:

```java
package com.github.thiagotgm.forester;

import org.checkerframework.common.value.qual.IntVal;
import org.checkerframework.common.value.qual.StaticallyExecutable;
import org.checkerframework.dataflow.qual.Pure;

public class App {
    public static void main( String[] args ) {
        System.out.println( "Hello World!" );
        new App().bar();
    }
    
    @StaticallyExecutable @Pure
    public static int myAdd( int a, int b ) {
        return a + b;
    }
    
    public void bar() {
        @IntVal({ 1 }) int a = 1;
        @IntVal({ 2 }) int b = 2;
        @IntVal({ 3 }) int v = myAdd( a, b );
    }
}
```

Before compiling with Checker, I used Maven to generate a jar of the project (`mvn package`), copied the jar to the project basedir (`/mnt/c/Users/thiag/GitRepositories/Forester`), then added that jar to the classpath. The compiler command I used is:

```bash
javac -d /mnt/c/Users/thiag/GitRepositories/Forester/target/classes -classpath /mnt/c/Users/thiag/GitRepositories/Forester/target/classes:/home/thiagotgm/.m2/repository/org/checkerframework/checker-qual/3.6.1/checker-qual-3.6.1.jar:/mnt/c/Users/thiag/GitRepositories/Forester/forester-0.0.1-SNAPSHOT.jar: -sourcepath /mnt/c/Users/thiag/GitRepositories/Forester/src/main/java:/mnt/c/Users/thiag/GitRepositories/Forester/target/generated-sources/annotations: -s /mnt/c/Users/thiag/GitRepositories/Forester/target/generated-sources/annotations -processor org.checkerframework.common.value.ValueChecker -processorpath /home/thiagotgm/.m2/repository/org/checkerframework/checker/3.6.1/checker-3.6.1.jar: -g --release 11 -encoding UTF-8 -Xmaxerrs 10000 -Xmaxwarns 10000 -AreportEvalWarns -AprintAllQualifiers -version -verbose src/main/java/com/github/thiagotgm/forester/App.java
```

(Note: Also tried to set the .class output itself by copying `./target/classes` to basedir then replacing the jar with that directory in the classpath, still didn't work)
[...]

PR #3647, opened 2020-09-09
Improve documentation about processorpath

mernst:
Fixes https://github.com/typetools/checker-framework/issues/3630.

I wonder whether we should merge this pull request, or replace it by a pull request that recommends *not* using classpath and using processorpath instead.


smillst:
>I wonder whether we should merge this pull request, or replace it by a pull request that recommends not using classpath and using processorpath instead.

The built classes need to be on the same path as the checker.  So, I would make changes that say that.

smillst (approved):

PR #3658, opened 2020-09-11
Remove IntelliJ work around.

smillst:


build.gradle
```diff
@@ -113,14 +113,6 @@ allprojects {
...
         // Holds the combined classpath of all subprojects including the subprojects themselves.
         allProjects
-
-        // There's a bug in IntelliJ that adds the processor path to the classpath when building
-        // a Gradle project.  Because ErrorProne uses old versions of some of our jars, IntelliJ uses
-        // those jars instead of the source code. The next 3 lines exclude Checker Framework from the
-        // from the processor path as a work around for this problem.
-        annotationProcessor.exclude group:'org.checkerframework', module:'javacutil'
-        annotationProcessor.exclude group:'org.checkerframework', module:'dataflow'
-        annotationProcessor.exclude group:'org.checkerframework', module:'checker-qual'
```
smillst, line 123: At least this line is still required, but not because of IntelliJ.  We need to update the comment above and test if we can remove the javacutil and dataflow exclustion.

mernst (approved):

Issue #904, opened 2016-09-05
No error about clear NPE when calling function in annonymous class from constructor


mihai-dumitresq:
No error about clear NPE when calling function in annonymous class from constructor.
Discovered while discussing #903. Tested with v2.1.1.

```Java
class Foo {
  private final Object mBar;
  private final Runnable mRunnable =
        new Runnable() {
          @Override
          public void run() {
            // No problem, even though it NPEs.
            mBar.toString();
          }
        };

  public Foo() {
    mRunnable.run();
    mBar = "";
  }
}
```

wmdietl:
Final fields without initializers might get set after the anonymous class is used, resulting in an NPE.
Final fields with initializers should be OK to access.
In general, the `mBar` access in run should be treated as `Foo.this.mBar` where the `Foo` instance is under initialization.

Test case added at `checker/tests/nullness/init/Issue904.java`.

mihai-dumitresq:
Also applies to inner classes. This related case also doesn't have any errors:

```Java
class Foo {
  private final Object mBar;
  class Inner {
      void run() {
        // No problem, even though it NPEs.
        mBar.toString();
      }
    };

  public Foo() {
    new Inner().run();
    mBar = "";
  }

  public static void main(String[] args) {
    new Foo();
  }
}
```

mihai-dumitresq:
Also, to avoid too many false positives if this issue is fixed, can the following be added as a test case as well?

```Java
class Foo {
  private final Object mBar;
  private final Runnable mRunnable =
        new Runnable() {
          @Override
          public void run() {
            mBar.toString();
          }
        };
  class Inner {
      void run() {
        mBar.toString();
      }
    };

  public Foo() {
    mBar = "";
  }

  public void shouldGiveNoErrors() {
    mRunnable.run();
    new Inner().run();
  }
}
```
(It currently has no errors, as intended)

NazguL86:
Resuming back this thread as it's still an open issue, here's another example I'm running into:
```
public class AccountStateCheck  {

    @Nullable
    private  JobFinishedCallback mJobFinishedCallback;

    private JobFinishedCallback mClassJobFinishedListener = new JobFinishedCallback() {
       @Override
        public void onSuccess() {
            mJobFinishedCallback = null;
            mJobFinishedCallback.onSuccess();  // <-- NOT CAUGHT
        }
    };
...
```

This is still an open issue, can we please have an update?
[...]

Issue #409, opened 2015-07-03
Implicit "this" passed through non-static inner class loses @UnderInitialization


GoogleCodeExporter:
```
The following code fails at runtime with an NPE, but does not cause an error in 
the checker framework. I'm guessing, but I believe this is because when I call 
p.call(), the "this" which is passed through to the inner class loses it's 
@UnderInitialization.

class Callback {
  static interface Proc {
    void call();
  }

  Proc p;

  class MyProc implements Proc {
    @Override
    public void call() {
      doStuff();
    }
  }

  String foo;

  Callback() {
    p = new MyProc();
    p.call();
    foo = "hello";
  }

  void doStuff() {
    System.out.println(foo.toLowerCase());
  }

  public static void main(String[] args) {
    new Callback();
  }
}
```

Original issue reported on code.google.com by `eatnumb...@google.com` on 9 Mar 2015 at 8:41

mernst:
I have created a smaller version of this test case and committed it as `checker/tests/nullness/Issue409.java`.

wmdietl:
There already is a `checker/tests/nullness/init/Issue409.java`. Should the smaller test case also go into `checker/tests/nullness/init/`?

mernst:
Thanks for pointing that out!  I have moved the smaller test case, replacing the larger one.
So. the test can now be found in `checker/tests/nullness/init/Issue409.java`.

PR #3659, opened 2020-09-11
Correct the type of "this"

### LINKED ISSUE #352: Incorrect annotations on nested class receivers
```
1.

class Outer {

    class Nested {
        @Untainted Outer context(@Untainted Outer.@Untainted Nested this) {
            return Outer.this;
        }
    }
}

OuterThisBug.java:8: error: [return.type.incompatible] incompatible types in 
return.
            return Outer.this;
                               ^
  found   : /*DECL*/ @Tainted Outer
  required: @Untainted Outer
1 error


The type of the returned Outer should come from context's receiver.

What version of the product are you using? On what operating system?
Checker Framework revision 261dee828a0a

Please provide any additional information below.

```

Original issue reported on code.google.com by `mcart...@cs.washington.edu` on 4 Sep 2014 at 7:41

### LINKED ISSUE #354: Nullness Checker nested class initialization
```
class Outer {

    String a;
    {
        Object o = new Object() {
            @Override
            public String toString() {
                return a.toString();
            }
        }.toString();

        // This is needed to avoid the initialization.fields.uninitialized warning.
        // The NPE still occurs
        a = ""; 
    }
}

I expect a warning on the "a" dereference, but there is none.

What version of the product are you using? On what operating system?
261dee828a0a tip

Please provide any additional information below.
This might require adjusting dataflow to pass propagate the current 
initialization store from the outer class to the evaluations in the inner class.
```

Original issue reported on code.google.com by `mcart...@cs.washington.edu` on 4 Sep 2014 at 9:49

### LINKED ISSUE #2208: Treat `ClassName.this` like `this`

```
import org.checkerframework.checker.initialization.qual.UnderInitialization;
import org.checkerframework.checker.nullness.qual.Raw;

public class ThisLiteralQualified {
    public ThisLiteralQualified() {
        super();
        @UnderInitialization @Raw ThisLiteralQualified a = this;
        @UnderInitialization @Raw ThisLiteralQualified b = ThisLiteralQualified.this;
    }
}
```

The expressions `this` and `ThisLiteralQualified.this` should be treated identically.

However, the Nullness Checker issues a warning for the latter but not for the former:

```
ThisLiteralQualified.java:13: error: [assignment.type.incompatible] incompatible types in assignment.
        @UnderInitialization @Raw ThisLiteralQualified b = ThisLiteralQualified.this;
                                                                               ^
  found   : @Initialized @NonNull ThisLiteralQualified
  required: @UnderInitialization @Nullable ThisLiteralQualified
1 error
```

### LINKED ISSUE #3266: Spurious enclosing class for types of NewClassTrees

### LINKED ISSUE #3408: False negative: Nullness Checker allows leaking `@UnderInitialization this` to anonymous class instance
Pretty similar to #3407, but still a distinct issue (i.e. it's likely that fixing #3407 won't automatically fix this issue and vice versa).

Let's consider the following class:

```java
class SampleClass {
  final String foo;
  String getFoo() { return foo; }

  SampleClass() {
    var anon = new Object() {
      String bar() {
        return SampleClass.this.getFoo().substring(1);
      }
    };
    anon.bar(); /// WHOOPS... NPE, `getFoo()` returns `foo` which is still null
    this.foo = "Hello world";
  }
}
```

This compiles with no error, despite leading to a NPE in the runtime.

### LINKED ISSUE #3561: Framework ignores annotations on outer types
If a class `Outer` contains an inner class `Inner`, then each instance of `Inner` has an instance of `Outer` separate from itself. This is often ignored, consider this class:
```java

public class Outer {
    void outerMethod(@Untainted Outer this) {
    }

    class Inner {
        void innerMethod(@Untainted Outer. @Untainted Inner this) {
            Outer.this.outerMethod();
        }
    }
}
```
This seems like it should type check because everything is `@Untainted`. However, I get this error
```
Outer.java:9: error: [method.invocation.invalid] call to outerMethod() not allowed on the given receiver.
            Outer.this.outerMethod();
                                  ^
  found   : @Tainted Outer
  required: @Untainted Outer
```

Currently, there's no way to specify that the type of `Outer.this` is `@Untainted` rather than `@Tainted`. The checker is overly conservative in this case.

I propose that, for a type `@A Outer. @B Inner`, where `@A` and `@B` are annotations, that the type of `Outer.this` is `@A` and the type of `Inner.this` is `@B`. If implemented, this would mean the following:

- Member accesses are checked against the correct enclosing class type. For example, in the above code, the framework would recognize that in `innerMethod` where the receiver has type `@Untainted Outer. @Untainted Inner`, that the type of `Outer.this` is `@Untainted Outer`, and the class should type check.
- Outer types must be checked during inner class creation. In a method `void myMethod(@Tainted Outer this)`, the line `@Untainted Outer. @Untainted Inner o = new @Untainted Inner()`. The reason is that the Outer instance for `o` is the receiver of the method it was created in, which has type `@Tainted Outer`, meaning the assignment would be invalid.
- New subtyping rules. For an inner class `C_1 . C_2. ... . C_n`, where each of `C_1` through `C_{n - 1}` is an enclosing class of `C_n`, then the type `@A1 C_1. @A_2 C_2. ... @A_n C_n` is a subtype of `@B_1 C_1. @B_2 C_2. ... . @B_n C_n` if and only if `@A_1` is a subtype of `@B_1`, `@A_2` is a subtype of `@B_2`, ..., `@A_n` is a subtype of `@B_n`.

smillst:
This pull request corrects the type of both implicit and explicit `this` references by correctly computing the enclosing type of `this`.

This pull request fixes two kinds of bugs: 
1. The type of `Outer.this` when used in an inner class.   (Issues #352, #2208, and #3561.)
2. For the Initialization Checker, the type of `Outer.this` when used in an anonymous class declared in a constructor. (Issues #354, part of #904, and #3408)

#409 (and other duplicate issues) is a related bug in the Initialization Checker, but isn't fixed by this PR.

Closes #352, closes #354, closes #2208, closes #3266, closes #3408 and closes #3561.

Merge with https://github.com/typetools/checker-framework-inference/pull/174.


mernst (approved):
Looks reasonable, but tests are not yet passing.

PR #3663, opened 2020-09-14
Warnings about `@SuppressWarnings` on inappropriate methods

### LINKED ISSUE #3648: Constant Value Checker warnings about `@StaticallyExecutable`
Prompted by the discussion in https://github.com/typetools/checker-framework/issues/3630 (thanks @ThiagoTGM!):

 * The Constant Value Checker should warn if a `@StaticallyExecutable` method is not `@Pure`.
 * The Constant Value Checker should warn if a `@StaticallyExecutable` method has a formal parameter (including the receiver) of a type that cannot be a compile-time constant.

mernst:
Fixes #3648


framework/src/main/java/org/checkerframework/common/value/ValueVisitor.java
```diff
@@ -341,4 +347,71 @@ public boolean validateType(Tree tree, AnnotatedTypeMirror type) {
...
+            }
+
+            // Ways to determin the receiver type.
+            // 1. This definition of receiverType is null when receiver is implicit and method has
+            //    class com.sun.tools.javac.code.Symbol$MethodSymbol.  WHY?
+            //        TypeMirror receiverType = method.getReceiverType();
+            // 2. Can't use the tree, because the receiver might not be explicit.
+            // 3. Check whether method is static and use the declaring class.  Doesn't handle all
+            //    cases, but handles the most common ones.
+            TypeMirror receiverType = method.getReceiverType();
```
smillst, line 388: The following should work:        
```
TypeMirror receiverType = TreeUtils.elementFromDeclaration(node).getReceiverType();
```

If the method has no receiver, the type returned with have `TypeKind.NONE`.

smillst (approved):

framework/src/main/java/org/checkerframework/common/value/ValueVisitor.java
mernst: That doesn't seem to work for me.  It returns a null value.  Please see https://dev.azure.com/mernst22/checker-framework/_build/results?buildId=5190&view=results .

smillst: That's strange.  The documentation of ExecutableType does not match its implementation.  You can still use that method, just check for `null`.

smillst: Or you can merge this pull request as is.

PR #3664, opened 2020-09-15
Compiling on every change or save in IntelliJ

mernst:


docs/manual/external-tools.tex
```diff
@@ -750,6 +753,21 @@
...
+To make IntelliJ compile on every change or save,
+access Settings (Preferences on macOS).
+Select Build, Execution, Deployment > Compiler.
+Enable Build project automatically. Press OK.
+
+You can also configure IntelliJ to automatically save (and thus
+automatically compile) your work periodically.
+
+You may also wish to do the following:
+Press Ctrl+Shift+A (Cmd+Shift+A on macOS) and search for Registry.
```
smillst, line 767: When I do this, the dialog box, in red letters, says "Changing these values may cause unwanted behavior of IntelliJ IDEA.  Please do not change these values unless you have been asked."  So, I don't think we should recommend this.

docs/manual/external-tools.tex
```diff
@@ -750,6 +753,21 @@
 \url{https://youtrack.jetbrains.com/issue/IDEA-159286}.
 
 
+\subsectionAndLabel{Running a checker on every IntelliJ change or save}{intellij-every-save}
+
+To make IntelliJ compile on every change or save,
+access Settings (Preferences on macOS).
+Select Build, Execution, Deployment > Compiler.
```
smillst, line 760: I prefer linking to the IntelliJ help page instead of writing text here.  The help page detects if you are using a Mac and shows the correct commends.  Plus it gives more information and stays up-to-date.

smillst (changes requested):

docs/manual/external-tools.tex
smillst: I think this is the link for this setting:

https://www.jetbrains.com/help/idea/compiling-applications.html#run_packaged_jar

mernst:
@smillst Could you complete this pull request?  Thanks.

smillst (approved):

PR #3678, opened 2020-09-22
Mention TreeAnnotator and TypeAnnotator in "creating a checker"

mernst:


docs/manual/creating-a-checker.tex
```diff
@@ -1182,29 +1182,77 @@
...
-%TODO: document tree and type annotators here, which should be used instead of override addComputedTypeAnnotations if possible.
+\begin{Verbatim}
+  @Override
+  protected TreeAnnotator createTreeAnnotator() {
+      return new ListTreeAnnotator(super.createTreeAnnotator(), new MyTreeAnnotator(this));
+  }
+\end{Verbatim}
+
+  \noindent
+  (or put your TreeAnnotator first; note that \<PropagationTreeAnnotator>,
```
smillst, line 1205: PropagationTreeAnnotator isn't the only annotator that super runs.

docs/manual/creating-a-checker.tex
```diff
@@ -1182,29 +1182,77 @@
...
+  which is run by default, adds annotations to unannotated trees, but has
```
smillst, line 1206: It adds annotations to AnnotatedTypeMirrors that do not have an annotation.

docs/manual/creating-a-checker.tex
```diff
@@ -1182,29 +1182,77 @@
...
+  your logic to be run even earlier or even later.
+
+  \<AnnotatedTypeFactory>, when given a program
+  expression, returns the expression's type.  This should include not only
+  the qualifiers that the programmer explicitly wrote in the source code, but
+  also default annotations and type
+  refinement (see Section~\ref{effective-qualifier} for explanations of these
+  concepts).
+
+  To add type introduction rules, you should override
```
smillst, line 1244: I think you should have deleted the rest of this section?

smillst (approved):

docs/manual/creating-a-checker.tex
mernst: That text is unchanged from the previous manual (just indented).  I think it adds information and should be retained.  But maybe I'm missing a reason it should be deleted.  Could you clarify?

smillst: I noticed it is unchanged, that's why I thought you meant to delete it.  The text reads as if it is the only way to add type introduction rules when it is just one of three ways to do so.  It's helpful to explicitly state the methods to override. 

I think you can just delete the sentence below or move it earlier in the section.  The other two items in the lists assume the read knows what an `AnnotatedTypeMirror` is. 
>  The methods operate on \refclass{framework/type}{AnnotatedTypeMirror},
  which is the Checker Framework's representation of an annotated type.

PR #3679, opened 2020-09-22
Refactor TypesUtils and NumberUtils; add TypeKindUtils

mernst:
The diff in `changelog.txt` explains the main changes (besides addition of some new methods).
The messy diff in `TypesUtils.java` is mostly rearranging existing methods to bring related methods near one another.

### DIFF
--- a/javacutil/src/main/java/org/checkerframework/javacutil/TypesUtils.java
+++ b/javacutil/src/main/java/org/checkerframework/javacutil/TypesUtils.java
@@ -37,39 +37,6 @@ public final class TypesUtils {
 
     /// Creating types
 
-    // TODO: What order of arguments for typeFromClass is desirable?  Current code is inconsistent,
-    // and before this refactoring TypesUtils already contained both overloads.  Remove one of the
-    // overloads.
-
-    /**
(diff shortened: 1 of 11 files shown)


javacutil/src/main/java/org/checkerframework/javacutil/TypesUtils.java
```diff
@@ -27,14 +28,91 @@
...
 
-    // Class cannot be instantiated
+    /** Class cannot be instantiated */
     private TypesUtils() {
         throw new AssertionError("Class TypesUtils cannot be instantiated.");
     }
 
+    /// Creating types
+
+    // TODO: What order of arguments for typeFromClass is desirable?  Current code is inconsistent,
```
mernst, line 40: Which order of arguments do you prefer?

smillst: I don't have a preference. It looks like the  `Elements elements, Types types` version was added on 9/18/20,  c1632554dcf1f24d3da6236a677d875299edebbf so I would keep the preexisting one.

smillst (approved):
Thanks!

PR #3690, opened 2020-09-25
extendWithNode is called for side effect, so don't return a value

mernst:
The trick of returning the argument makes some code more concise, but it is
confusing since the method is primarily called for side effect, and it is
used inconsistently throughout the codebase.  Removing the return type
makes clients more uniform and explicit, if a bit more verbose.


smillst (approved):
The code is much easier to read.

PR #3718, opened 2020-09-30
Add a debugging/logging switch to GenericAnnotatedTypeFactory

mernst:
This can be useful when debugging, but it also clutters the code.  What do you think?


smillst (approved):
I don't use logging output to debug, so I probably wouldn't use this.  But if you find it helpful, I don't have a problem with this being merged.

Issue #2953, opened 2019-12-03
Enhancement request: new @EnumVal() that would behave like @StringVal()


smonteiro:
I could be wrong but I did not find documentation for a `@EnumVal`
The concept is not too far-fetched: one wants to enforce domain, co-domain, and range:

> Example of use case:
import javax.lang.model.type.TypeKind;
static @EnumVal({TypeKind.BYTE,
                    TypeKind.BOOLEAN,
                    TypeKind.CHAR,
                    TypeKind.DOUBLE,
                    TypeKind.FLOAT,
                    TypeKind.INT,
                    TypeKind.LONG}) TypeKind toPrimitiveTypeKind(...)
...
static  int fromPrimitiveTypeKind( @EnumVal({TypeKind.BYTE,
                    TypeKind.BOOLEAN,
                    TypeKind.CHAR,
                    TypeKind.DOUBLE,
                    TypeKind.FLOAT,
                    TypeKind.INT,
                    TypeKind.LONG}) TypeKind tk) {...)

kelloggm:
This is probably related to #2156. I agree that it would be useful for the Value Checker to support an `@EnumVal` annotation; see e.g. 
[this comment](https://github.com/awslabs/aws-kms-compliance-checker/blob/master/src/main/java/com/amazon/checkerframework/compliance/kms/ComplianceChecker.java#L18) in a third-party checker, which describes a desire to write annotations in a stub file on enum constants. The linked CF issue (#2147) was closed in favor of the more general #2156.

smonteiro:
While the topic of  #2156 is Enum, it is about something else.
My interest here is to clamp the domain  of arguments to functions and clamp their return values.

mernst:
I agree that this is a valuable enhancement, and that it's unrelated to #2156.
I don't think this would be very challenging to do:  the logic should be very similar to the other `*Val` annotations.

mernst:
Fixed in 483bd239c6d3755cdd62166f0bdd9e2e89d81acc

PR #3722, opened 2020-09-30
Add @EnumVal annotation to Constant Value Checker

kelloggm:
Implements #2953 by re-using the logic for `@StringVal`.

Some design notes/possible concerns:
* @mernst and I discussed whether it might be possible to use the actual enum constants as arguments to the annotation, in the hopes that the built-in `toString` implementation would convert them automatically. This doesn't work, because the arguments have to be constants, and the call to `toString` makes it an expression.
* I had to suppress three warnings in `Issue1218.java` related to type arguments. I'm pretty sure these are false positives; they're all of the form:
  ```
  framework/tests/value/Issue1218.java:75: error: [type.argument.type.incompatible] incompatible type argument for type 
  parameter E extends @UnknownVal Enum<@StringVal("A") MyEnum> of enums.
          enums(MyEnum.A);
               ^
    found   : @StringVal("A") MyEnum
    required: @UnknownVal Enum<@StringVal("A") MyEnum>
  ```
  If a reviewer has any ideas about how to get rid of these false positives in the checker itself, I'd appreciate it - I'm at a bit 
  of a loss.
* I didn't include `@EnumVal` in the type hierarchy diagram because it's treated identically to `@StringVal`, so it seemed redundant.


mernst:
@kelloggm Please ensure that CI passes before assigning the pull request.  Thanks. :-)

kelloggm:
@mernst, I've fixed the CI failures (including the new test you added, with some slight modifications), so this is now ready for review.

kelloggm:
This PR should not be merged before **https://github.com/typetools/jdk/pull/91**

mernst (approved):

PR #3726, opened 2020-10-01
Split getValueWithSameAnnotations() into getWidenedValue() and getNarrowedValue()

mernst:


checker/bin-devel/test-misc.sh
```diff
@@ -21,7 +21,7 @@ source "$SCRIPTDIR"/build.sh
 # Javadoc documentation
 # Uncomment this line temporarily for refactorings that touch a lot of code that
 # you don't understand.  Then, recomment it as soon as the pull request is merged.
-# SKIPJAVADOC=1
+SKIPJAVADOC=1
```
smillst, line 24: Should this be reverted?

checker/src/main/java/org/checkerframework/checker/signedness/SignednessAnnotatedTypeFactory.java
```diff
@@ -184,28 +188,39 @@ private void addSignednessGlbAnnotation(Tree tree, AnnotatedTypeMirror type) {
...
+            // TODO: A future enhancement will make the widened type indicate the signed basetype
+            // from which it was widened.
+            result.add(SIGNEDNESS_GLB);
+            return result;
+        }
+    }
+
+    @Override
+    public Set<AnnotationMirror> getNarrowedAnnotations(
+            Set<AnnotationMirror> annos, TypeKind typeKind, TypeKind widenedTypeKind) {
```
smillst, line 218: Should `widenedTypeKind` be renamed to `narrowedTypeKind`?

changelog.txt
```diff
@@ -1,3 +1,10 @@
+Version 3.8.0, November 2, 2020
+
+Renamed CFAbstractTransfer.getValueWithSameAnnotations() to two methods
```
smillst, line 3: I wouldn't call this a renaming.  `CFAbstractTransfer.getValueWithSameAnnotations()` was removed.  Also, we don't necessarily need to remove that method.  If we keep it in, then this change wouldn't be a breaking change.

smillst (approved):

changelog.txt
mernst: @smillst I think it's still a breaking change because `getValueWithSameAnnotations()` is no longer called.  If we leave it in, then clients of the Checker Framework will continue to compile, but their behavior will change.  If you think that's OK, we could make the version number 3.7.1.  What do you think?

smillst: The documentation of `getValueWithSameAnnotations` doesn't say that it is called for narrowing and widening conversions, so clients shouldn't rely on that fact. So, I'm ok with this.

PR #3731, opened 2020-10-02
Use dependabot v1 config instead of v2

### LINKED ISSUE #3730: Your .dependabot/config.yml contained invalid details
Dependabot encountered the following error when parsing your `.dependabot/config.yml`:

```
The property '#/' did not contain a required property of 'update_configs'
The property '#/' contains additional properties ["updates"] outside of the schema when none are allowed
The property '#/version' value 2 did not match one of the following values: 1
```

Please update the config file to conform with Dependabot's specification using our [docs](https://dependabot.com/docs/config-file/) and [online validator](https://dependabot.com/docs/config-file/validator/).

kelloggm:
Fix #3730, which seems to have been caused by having the wrong version of dependabot enabled for the config script that we added. This change uses the old config script format (v1) rather than the newer v2, since we're apparently running v1.

Background: funnily enough, I wasn't able to figure out which version of dependabot we were using, so I guessed in #3724 that it was the newer version. Apparently dependabot doesn't update itself!

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,10 +1,11 @@
-version: 2
+version: 1
 
 # The com.amazonaws:aws-java-sdk-* dependencies are updated every day.
 # That is too many pull requests.
 
 updates:
-  - package-ecosystem: "java:gradle"
+  - package-manager: "java:gradle"
     directory: "/"
-    ignore:
-      - dependency-name: "com.amazonaws:aws-java-sdk-*"
+    ignore-updates:
+      - match:
+          dependency-name: "com.amazonaws:aws-java-sdk-*"


mernst:
@kelloggm I tried to fix the problem in https://github.com/typetools/checker-framework/commit/edca48452370a216c84cb130ad998020a15534a6 .  Is that not the right fix?

kelloggm:
I'm not sure. Your commit was definitely necessary, but I'm not sure if it is sufficient. The error message in #3730 says `The property '#/version' value 2 did not match one of the following values: 1`, which I understood to mean that we had to use the v1 syntax. The two config files (this one and the one that's currently checked in) should be identical, so we should only merge this PR if the current one still isn't working after your fix.

mernst:
@kelloggm Do you know how we can test whether the current one (after my attempted fix) is working?

kelloggm:
Let's just wait until Monday and see if we get anything from dependabot in the meantime.

I think this issue was mainly caused by using dependabot without a config (i.e. via the old GitHub integration), which will hopefully not be a problem in the future.

PR #3743, opened 2020-10-07
Support disabling returns receiver checker in an accumulation checker

kelloggm:
Also includes a (possibly over-engineered...) framework for supporting pluggable alias analyses in accumulation checkers. We don't have any other than returns receiver yet, but I anticipate adding more in the future as we build more accumulation analyses.

Because the only alias analysis available is RR, I didn't document the feature in the manual. I will do so once there is more than one alias analysis to choose from.

### DIFF
--- a/framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
@@ -19,6 +19,7 @@ import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.util.Elements;
 import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.checkerframework.common.accumulation.AccumulationChecker.AliasAnalysis;
 import org.checkerframework.common.basetype.BaseAnnotatedTypeFactory;
 import org.checkerframework.common.basetype.BaseTypeChecker;
 import org.checkerframework.common.returnsreceiver.ReturnsReceiverAnnotatedTypeFactory;
@@ -45,6 +46,12 @@ import org.checkerframework.javacutil.UserError;
  * #postInit()}.
  */
 public abstract class AccumulationAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
+
+    /**
+     * The typechecker associated with this factory. All AccumulationAnnotatedTypeFactory objects
+     * should use an accumulation checker rather than a standard base type checker.
+     */
+    public final AccumulationChecker accumulationChecker;
 
     /**
      * The canonical top annotation for this accumulation checker: an instance of the accumulator
@@ -84,6 +91,13 @@ public abstract class AccumulationAnnotatedTypeFactory extends BaseAnnotatedType
             Class<? extends Annotation> bottom,
             @Nullable Class<? extends Annotation> predicate) {
         super(checker);
+        if (!(checker instanceof AccumulationChecker)) {
+            throw new BugInCF(
+                    "AccumulationAnnotatedTypeFactory cannot be used with a checker "
+                            + "class that is not a subtype of AccumulationChecker. Found class: "
+                            + checker.getClass());
+        }
+        this.accumulationChecker = (AccumulationChecker) checker;
 
         this.accumulator = accumulator;
         // Check that the requirements of the accumulator are met.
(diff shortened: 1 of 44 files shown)


framework/src/main/java/org/checkerframework/common/accumulation/AccumulationChecker.java
```diff
@@ -10,20 +13,71 @@
...
+     * implementation of {@link #createAliasAnalyses()} to be as simple as possible.
+     */
+    private @MonotonicNonNull EnumSet<AliasAnalysis> aliasAnalyses;
+
     @Override
     protected LinkedHashSet<Class<? extends BaseTypeChecker>> getImmediateSubcheckerClasses() {
         LinkedHashSet<Class<? extends BaseTypeChecker>> checkers =
                 super.getImmediateSubcheckerClasses();
-        checkers.add(ReturnsReceiverChecker.class);
+        AliasAnalysis[] aliasAnalyses = createAliasAnalyses();
```
msridhar, line 38: This is coded as if enabling any `AliasAnalysis` will require enabling a sub-checker.  Is that the intention?  I'm not really sure how future alias reasoning might look.  In any case, this code might be clearer for now by just checking if `aliasAnalyses` contains `RETURNS_RECEIVER`, rather than looping and doing a switch.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationChecker.java
```diff
@@ -10,20 +13,71 @@
...
+    public enum AliasAnalysis {
+        RETURNS_RECEIVER
+    }
+
+    /**
+     * Get the alias analyses that this checker should employ.
+     *
+     * @return the alias analyses
+     */
+    protected AliasAnalysis[] createAliasAnalyses(@UnderInitialization AccumulationChecker this) {
```
msridhar, line 64: Should this return an `EnumSet`?

msridhar (commented):
Mostly looks good, but I have a couple questions.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationChecker.java
kelloggm: It returns an array rather than an enumset by design, so that overriding it is as easy as possible. With this signature, you can use the array constant syntax (as this version of the method does), which means that regardless of how many alias analyses you want to enable the method will always be very short and understandable. An enumset wouldn't be that much harder, of course, so if we think that's a lot clearer than I can change it.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationChecker.java
kelloggm: That's a fair point. I'll change this.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationChecker.java
msridhar: This is fine as is I think

msridhar (approved):

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationChecker.java
mernst: I had the same question as Manu when I read the code.  I think that using an array makes this less rather than more concise -- the `EnumSet.of()` and `EnumSet.allOf()` calls are slightly more concise.  Furthermore, it feels a bit convoluted, and clutters the code, to use an array that will be immediately converted to an EnumSet.

mernst (approved):

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
```diff
@@ -84,6 +88,13 @@ protected AccumulationAnnotatedTypeFactory(
             Class<? extends Annotation> bottom,
             @Nullable Class<? extends Annotation> predicate) {
         super(checker);
+        if (!(checker instanceof AccumulationChecker)) {
+            throw new BugInCF(
```
mernst, line 92: I think this should be `TypeSystemError` instead.

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationChecker.java
kelloggm: Since this was confusing for both of you, I'll change it to use an EnumSet throughout.

PR #3744, opened 2020-10-07
Update documentation for isValid and related methods

mernst:


framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -4076,6 +4076,11 @@ protected boolean isAccessAllowed(
      * <p>In most cases, {@code useType} simply needs to be a subtype of {@code declarationType}. If
      * a type system makes exceptions to this rule, its implementation should override this method.
      *
+     * <p>This method is not called if checkTopLevelDeclaredOrPrimitiveType is false -- by default,
```
smillst, line 4079: This method is not called if ` {@link #shouldCheckTopLevelDeclaredOrPrimitiveType(AnnotatedTypeMirror, Tree)}` returns false.

smillst (approved):

PR #3751, opened 2020-10-08
`@MatchesRegex` annotation for Constant Value Checker

kelloggm:


framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -936,6 +953,19 @@ public AnnotationMirror createArrayLenRangeAnnotation(Range range) {
         return createArrayLenAnnotation(ValueCheckerUtils.getLengthsForStringValues(values));
     }
 
+    /**
+     * Converts an {@code StringVal} annotation to an {@code MatchesRegex} annotation that matches
+     * exactly the string values listed in the {@code StringVal}.
```
mernst, line 958: This code does not do that.  To achieve this, you would need to call `Pattern.quote` and also add `^` and `$` anchors.

Also note that the strings are not necessarily legal regular expressions.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -917,6 +922,18 @@ public AnnotationMirror createArrayLenRangeAnnotation(Range range) {
         }
     }
 
+    /**
+     * Creates an {@code MatchesRegex} annotation for the given regular expressions.
+     *
+     * @param regexes a list of Java regular expressions
+     * @return a MatchesRegex annotation with those values
+     */
+    private AnnotationMirror createMatchesRegexAnnotation(List<String> regexes) {
```
mernst, line 931: Add a `@Regex` annotation.  That would have caught the problem I noted below.

mernst (changes requested):

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -936,7 +968,30 @@ public AnnotationMirror createArrayLenRangeAnnotation(Range range) {
...
+     * @param stringValAnno a StringVal annotation
+     * @return an equivalent MatchesReges annotation
+     */
+    /* package-private */ AnnotationMirror convertStringValToMatchesRegex(
+            AnnotationMirror stringValAnno) {
+        List<String> values = getStringValues(stringValAnno);
+        List<@Regex String> valuesAsRegexes = new ArrayList<>();
+        for (String value : values) {
+            // Quote the String, and surround it with start-of-string and end-of-string markers.
+            valuesAsRegexes.add("^" + Pattern.quote(value) + "$");
```
mernst, line 984: Because your implementation uses `matches` rather than `Matcher.find`, you do not need these anchors.  The tests all pass without them.

mernst (changes requested):

framework/src/main/java/org/checkerframework/common/value/qual/MatchesRegex.java
```diff
@@ -0,0 +1,28 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import org.checkerframework.framework.qual.SubtypeOf;
+
+/**
+ * An annotation indicating the possible values for a String type. If an expression's type has this
+ * annotation, then at run time, the expression evaluates to a string that matches at least one of
```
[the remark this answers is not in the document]
kelloggm: Rather than adding examples here, I've changed this to refer to the documentation of `java.lang.String#matches`. That absolves us of the need to maintain examples here, where they might become out of date. I will add the tests you describe, though, since those will start to fail if the behavior of `matches` changes.

mernst (approved):
Thanks!

PR #3755, opened 2020-10-09
Warn about nullness annotations on outer class scoping constructs

mernst:
Also adds method BaseTypeVisitor#visitAnnotatedType.


checker/tests/nullness/ScopingConstruct.java
```diff
@@ -0,0 +1,28 @@
...
+
+    ScopingConstruct.@Nullable StaticNested scnsn;
+
+    // :: error: (nullness.on.outer)
+    @Nullable ScopingConstruct.Inner nsci;
+
+    ScopingConstruct.@Nullable Inner scni;
+
+    // :: error: (nullness.on.outer)
+    @Nullable ScopingConstruct.Inner @Nullable [] array;
```
smillst, line 26: There is currently no error on this line.  Should there be?

smillst (changes requested):

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -2257,6 +2265,29 @@ public Void visitThrow(ThrowTree node, Void p) {
...
+        return super.visitAnnotatedType(node, p);
+    }
+
+    /**
+     * Checks an annotated type. Invoked by {@link #visitAnnotatedType(AnnotatedTypeTree, Void)} and
+     * also by {@link #visitVariable}. Exists to prevent code duplication between the two. Checking
+     * in visitVariable is needed because there isn't an AnnotatedTypeTree within a variable
+     * declaration -- all the annotations are attached to the VariableTree.
+     *
+     * @param declAnnos the user-written type annotations on a variable/method declaration, if this
```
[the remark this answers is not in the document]
mernst: Can you be explicit about what change you are suggesting?  I can't tell from your comment.

I agree they are not declaration annotations; they are annotations stored in the declaration location.  I used `declAnnos` for that concept.  Are you asking for a different variable name, and if so what name do you suggest?

I don't agree with your suggestion to change "variable declaration" to "variable".  Strictly speaking, they are not stored as modifiers of a *variable*, but as modifiers of a *variable declaration* (which is represented as a `VariableTree`).

mernst:
Thanks for the review; the code is cleaner and more general now.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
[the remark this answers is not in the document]
smillst: When I see `declAnnos`, I think thats a collection/set of `AnnotationMirror`s that are declaration annotations, as in annotation without `@Target(ElementType.TYPE_USE)`.  I would change `declAnnos` to `annoTrees`.  Also, the annotations could be declarations annotations or type annotations, so I would change `the user-written type annotations` to just `the user-written annotations`.

javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -1541,4 +1540,47 @@ public static boolean isWideningBinary(BinaryTree node) {
...
+    /**
+     * Returns the annotations explicitly written on the given type.
+     *
+     * @param declAnnos annotations written before a variable/method declaration; null if this type
+     *     is not from such a location
+     * @param typeTree the type whose annotations to return
+     * @return the annotations explicitly written on the given type.
+     */
+    public static List<? extends AnnotationTree> getExplicitAnnotationTrees(
+            List<? extends AnnotationTree> declAnnos, Tree typeTree) {
```
smillst, line 1553: I would also change `declAnnos` to `annoTree` here, too.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -2257,6 +2264,33 @@ public Void visitThrow(ThrowTree node, Void p) {
...
+     * belong to the variable type or method return type.
+     */
+    @Override
+    public Void visitAnnotatedType(AnnotatedTypeTree node, Void p) {
+        visitAnnotatedType(null, node);
+        return super.visitAnnotatedType(node, p);
+    }
+
+    /**
+     * Checks an annotated type. Invoked by {@link #visitAnnotatedType(AnnotatedTypeTree, Void)} and
```
smillst, line 2281: >`Checks an annotated type.`  

Is it `all` annotated types, or could we still be missing some?  It's fine if you aren't confident, I'm just curious.

smillst (approved):
The commit comment should say something about adding `BaseTypeVisitor#visitAnnotatedType`.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
mernst: I'm pretty sure it is invoked for all annotated types.

PR #3766, opened 2020-10-13
Warn about Index Checker annotations on floating-point types

mernst:


checker/tests/index/LessThanFloat.java
```diff
@@ -0,0 +1,24 @@
...
+    // :: error: (anno.on.nonintegral)
+    @LessThan("bigger") float f;
+
+    // :: error: (anno.on.nonintegral)
+    @LessThan("bigger") double d;
+
+    // :: error: (anno.on.nonintegral)
+    @LessThan("bigger") boolean bool;
+
+    @LessThan("bigger") char c;
```
kelloggm, line 23: Can you add tests for the primitive wrapper types (e.g. `java.lang.Integer`) here, as well? I think the code as-is handles them correctly, but it would be good to be sure.

kelloggm (approved):
Thanks for this change! I had only one minor comment; once that change is made I think this can be merged.

checker/tests/index/LessThanFloat.java
mernst: The code did *not* handle primitive wrapper types.  (There was a todo comment about that.)  I have enhanced the code to do so, removed the todo, and added tests.

checker/src/main/java/org/checkerframework/checker/index/inequality/LessThanVisitor.java
```diff
@@ -131,4 +138,47 @@ protected boolean isTypeCastSafe(AnnotatedTypeMirror castType, AnnotatedTypeMirr
 
         return super.isTypeCastSafe(castType, exprType);
     }
+
+    // TODO: This belongs in an IndexVisitor.
+    @Override
+    public void visitAnnotatedType(
+            @Nullable List<? extends AnnotationTree> annoTrees, Tree typeTree) {
+        // Look for a PRIMITIVE within the type.
```
smillst, line 146: This is now primitive or boxed primitive.

checker/src/main/java/org/checkerframework/checker/index/inequality/LessThanVisitor.java
```diff
@@ -131,4 +138,47 @@ protected boolean isTypeCastSafe(AnnotatedTypeMirror castType, AnnotatedTypeMirr
 
         return super.isTypeCastSafe(castType, exprType);
     }
+
+    // TODO: This belongs in an IndexVisitor.
```
smillst, line 142: Is there some reason not to do this now?

checker/src/main/java/org/checkerframework/checker/index/inequality/LessThanAnnotatedTypeFactory.java
```diff
@@ -336,4 +348,58 @@ public AnnotationMirror createLessThanQualifier(String expression) {
...
+    }
+
+    /**
+     * Returns true if type has an index annotation that is only applicable to integral types, such
+     * as @LessThan.
+     *
+     * @param type a type
+     * @return true if the type has an index annotation that is only applicable to integral types
+     */
+    protected boolean hasIntegralIndexAnnotation(AnnotatedTypeMirror type) {
```
smillst, line 383: This method isn't used.  Should it be deleted?

checker/src/main/java/org/checkerframework/checker/index/inequality/LessThanAnnotatedTypeFactory.java
```diff
@@ -336,4 +348,58 @@ public AnnotationMirror createLessThanQualifier(String expression) {
...
+
+    /**
+     * Returns true if the given annotation is an index annotation that is only applicable to
+     * integral types, such as @LessThan.
+     *
+     * @param am an annotation
+     * @return true if the given annotation is an index annotation that is only applicable to
+     *     integral types
+     */
+    protected boolean isIntegralIndexAnnotation(AnnotationMirror am) {
```
smillst, line 400: Calls to this method could be replaced with calls to `AnnotatedTypeFactory#isSupportedQualifier(AnnotationMirror)`.

smillst (changes requested):

checker/src/main/java/org/checkerframework/checker/index/inequality/LessThanVisitor.java
mernst: There is no `IndexVisitor` class.  We would need to create one just to contain this check.
I wanted to get your opinion about whether that is worth doing.

checker/src/main/java/org/checkerframework/checker/index/inequality/LessThanAnnotatedTypeFactory.java
mernst: That would be correct for most annotations, but the error message "Annotation %s is only applicable to integral types" would be funny for an annotation like `@SameLen` which is applicable to sequences.

checker/src/main/java/org/checkerframework/checker/index/inequality/LessThanVisitor.java
smillst: > There is no IndexVisitor class. We would need to create one just to contain this check. That would add a bit of complexity, but more seriously it would add another pass, slowing down the checker.

`IndexVisitor` should be a subclass of `UpperBoundVisitor`. (just like `IndexChecker` is a subclass of `UpperBoundChecker`.). So this wouldn't be a full additional pass.

>I wanted to get your opinion about whether that is worth doing.

So in this pull request, you've only added warnings about the LessThanChecker.  I'm assuming you want to add warnings for all of the other Index Checker annotations.  I can see the benefits of doing that all in one place, but it would also make sense for each checker to issue warnings about its own annotations.

Also, it probably makes sense to generalize this more.  One thought is to add warnings if an annotation is written on a type that is not in the list of its `@RelevantJavaTypes` meta-annotation.  We can discuss more at our next meeting.

checker/src/main/java/org/checkerframework/checker/index/inequality/LessThanAnnotatedTypeFactory.java
smillst: `@SameLen` is not a supported qualifier of the LessThanChecker, so `AnnotatedTypeFactory#isSupportedQualifier(AnnotationMirror)` would return false.

mernst: I agree.  I was thinking forward to centralizing the logic in one place.  (But, I realize that was not clear; sorry about that.)

mernst:
Closed in favor of #3787

PR #3773, opened 2020-10-13
More precise type for `someIdentifier.class.getName()`

mernst:


checker/src/main/java/org/checkerframework/checker/signature/SignatureAnnotatedTypeFactory.java
```diff
@@ -247,6 +266,24 @@ public Void visitMethodInvocation(MethodInvocationTree tree, AnnotatedTypeMirror
                     type.replaceAnnotation(BINARY_NAME);
                 }
             }
+
+            if (TreeUtils.isMethodInvocation(tree, classGetName, processingEnv)) {
+                ExpressionTree receiver = TreeUtils.getReceiverTree(tree);
+                if (receiver.getKind() == Tree.Kind.MEMBER_SELECT) {
```
smillst, line 272: You can replace this if and the next with `org.checkerframework.javacutil.TreeUtils#isClassLiteral`.

smillst (approved):

PR #3783, opened 2020-10-15
Add detail to error messages

mernst:


framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeMirror.java
```diff
@@ -870,16 +870,24 @@ private AnnotatedDeclaredType(
...
+            switch (encl.getKind()) {
+                case NONE:
+                    this.enclosingType = null;
+                    break;
+                case DECLARED:
+                    this.enclosingType =
+                            (AnnotatedDeclaredType) createType(encl, atypeFactory, declaration);
+                    break;
+                case ERROR:
+                    throw new BugInCF(
```
smillst, line 882: Maybe I'm missing something, but I don't see how this is different from the default case.

mernst: An error is not unexpected, but indicates a problem earlier (such as code that is not compilable).

mernst: OK, I must have been running into that bug.

Issue #3569, opened 2020-08-06
ClassCastException in AnnotatedTypeMirror$AnnotatedIntersectionType.directSuperTypes


msridhar:
Input (reduced from code in https://github.com/h2oai/h2o-3):
```java
import java.io.Serializable;
import java.util.Map;

class ClassCastCrash {
    static class AutoBuffer {
        public <T extends Freezable> T get() {
            return null;
        }

        public String getStr() {
            return "hi";
        }
    }

    public interface Freezable<T extends Freezable> extends Cloneable {}

    public abstract static class Iced<D extends Iced>
            implements Freezable<D>, java.io.Externalizable {
        @Override
        public void readExternal(java.io.ObjectInput ois)
                throws java.io.IOException, ClassNotFoundException {}

        @Override
        public void writeExternal(java.io.ObjectOutput oos) throws java.io.IOException {}
    }

    public abstract static class IcedHashMapBase<K, V> extends Iced
            implements Map<K, V>, Cloneable, Serializable {
        public enum KeyType {
            String(String.class),
            Freezable(Freezable.class),
            ;
            Class _clazz;

            KeyType(Class clazz) {
                _clazz = clazz;
            }
        }

        @SuppressWarnings("unchecked")
        protected K readKey(AutoBuffer ab, KeyType keyType) {
            switch (keyType) {
                case String:
                    return (K) ab.getStr();
                case Freezable:
                    return ab.get();
                default:
                    return null;
            }
        }
    }
}
```
I get a crash with the tainting checker:
```
$ ./checker/bin/javac -processor TaintingChecker checker/tests/tainting/ClassCastCrash.java
error: SourceChecker.typeProcess: unexpected Throwable (ClassCastException) while processing checker/tests/tainting/ClassCastCrash.java; message: org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedTypeVariable cannot be cast to org.checkerframework.framework.type.AnnotatedTypeMirror$AnnotatedDeclaredType
  ; The Checker Framework crashed.  Please report the crash.
  Compilation unit: checker/tests/tainting/ClassCastCrash.java
  Last visited tree at line 4 column 1:
  class ClassCastCrash {
[...]

PR #3790, opened 2020-10-16
Correct AnnotatedIntersectionTypes

### LINKED ISSUE #868: AnnotatedIntersectionTypes are not defaulted correctly

### LINKED ISSUE #3349: Qualifiers in intersection types
The qualifiers in intersecton types is confusing me by allowing two exclusive qualifiers in one type. For example:

```
import java.io.Serializable;

class MyClass<T extends @NonNull Object & @Nullable Serializable> {
    void foo(T p1){
        @NonNull Serializable s = p1;  // OK, but checker issues warning
        @NonNull Object o = p1;
    }
}
```
The first assignment operation of above code yields this error:
```
Error: [assignment.type.incompatible] incompatible types in assignment.
   found   : T extends @Initialized @NonNull Object & @Initialized @Nullable Serializable
   required: @UnknownInitialization @NonNull Serializable
```
That indicates `T` have both qualifiers, but `T` is only one type and cannot be instantiated with `@Nullable` arguments. So, it is quite confusing to have the error.

### LINKED ISSUE #3362: Warn about surprising behavior in annotated intersection types
This is a feature request to issue a warning when the type qualifier for an intersection type is not obvious from the types written on its component types, also known as its bound types.

In an intersection type like `@A MyClass & @B MyInterface`, any instantiation must be at GLB(`@A`, `@B`) or lower.  This may surprise a programmer who is just looking at the `@A MyClass` part, who may get an error message saying that `@A MyClass` is not a legal instantiation.

Another surprising behavior is in code like this:

```
class MyClass<T extends @Nullable Object & Serializable> { ... }
```

where every instantiation must be `@NonNull` even though the programmer explicitly wrote `@Nullable`.  The reason is that `Serializable` defaults to `@NonNull` serializable, and GLB(`@Nullable`, `@NonNull`) = `@NonNull`.

Here is the feature request:

 * If the GLB is lower than some explicitly-written type qualifier:
    * If the GLB is the default for all the component types, suggest omitting qualifiers from all the component types.
    * If the GLB can be written in source code on all the component types, then suggest that the user write the GLB on all the component types.

smillst:
General facts about intersections in javac that should make this pull request easier to review:
1. In source code, intersections can only be written as a cast type or an upper bound of a type parameter.
a. For cast types, the intersection type is parsed into an IntersectionTypeTree.
b. For type variable upper bounds, the intersection type is parsed into a list of trees
2. In byte code, the Checker Framework only reads annotations on intersections for type parameter upper bounds.
3. Intersection types can also arise from capture conversion and least upper bounds.
4. The bounds of explicitly written intersections must be declared types, but intersections from LUB or capture conversion can be any kind of type.

Changes in this pull request:
Defaulting rules for the primary annotation of an intersection type:
1. One or more of the bounds have an explicit annotation: the default annotation is the first explicit annotation.  (I tried making the default the glb of the (explicit or defaulted) annotations on the bounds, but I found this confusing in the case of casts.  We can discuss this further.)
2. Zero bounds have an explicit annotation: the default annotation of the location is applied.  (So, either explicit upper bound defaults, or for casts, the annotation of the cast expression.)

Warning:
If any explicit annotation on a bound is not the same as the primary annotation, then a warning is issued.  (I was going to put this in a separate pull request, but it needs to be in this one so that the test cases are easier to understand.)  

Changes to AnnotatedIntersectionType:
1. Adding a primary annotation to an intersection type causes that annotation to be copied to its bounds.  This is similar to what happens when an annotation is added to a type variable.
2. New method `getBounds` returns a list of the bounds of the intersection type.  (This used to be done via `directSuperTypes` method.
3.  The bounds can be any AnnotatedTypeMirror if the intersection arises from capture conversion.

 Fixes #868; Fixes #3349; Fixes #3362; Fixes #3659.


framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java
```diff
@@ -1016,6 +990,26 @@ protected boolean visitIntersectionSupertype(
         return result;
     }
 
+    /**
+     * An intersection is a subtype if one of its bounds is a subtype of {@code supertype}.
```
mernst, line 994: When this is called, have the primary annotations already been applied to every bound, so that each bound stands on its own without any need to refer to the primary annotations?

mernst (approved):
Overall, this looks good.  Thanks!  I have a couple clarifying questions.

msridhar:
Does this also fix #3569?  Just curious.

mernst:
Yes, it does.  Good catch.  I have updated the pull request description and the linked issues.

framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java
smillst: Yes.  This method is only called on fully-annotated types.  That means that the intersection type has a primary annotation for every qualifier hierarchy.

smillst: TOOD: Add comment.

PR #3798, opened 2020-10-18
Abstract out getSimpleClassName() and getMethodName() methods

mernst:
This looked more compelling on a branch where I was doing some other work in these methods.
In this pull request, I could take it or leave it.  What do you think?


smillst (approved):
Seems like they are nice to have methods.

PR #3803, opened 2020-10-19
Rename Block.getId() to Block.getUid()

mernst:


mernst:
Closed in favor of #3804

PR #3812, opened 2020-10-20
Refactor isAccessAllowed to checkAccessAllowed

mernst:


framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -4138,38 +4138,46 @@ protected void checkAccess(IdentifierTree node, Void p) {
...
         if (unused == null) {
-            return true;
+            return;
         }
 
         String when = AnnotationUtils.getElementValueClassName(unused, "when", false).toString();
-        if (!AnnotationUtils.containsSameByName(receiver.getAnnotations(), when)) {
-            return true;
+
+        // TODO: Don't just look at the receiver type, but at the receiver declaration.  (That will
```
smillst, line 4163: I think you mean check the declaration annotations on the receiver rather than the type qualifiers.

smillst (approved):
These changes seems fine, but the `@Unused` annotation and the `unallowed.access` error are not tested.  

Also, the documentation on `@Unused` doesn't match its implementation.

>  Declares that the field may not be accessed if the receiver is of the specified qualifier type (or any supertype). 

It's not a supertype it must be exactly the same annotation.

mernst:
> These changes seems fine, but the `@Unused` annotation and the `unallowed.access` error are not tested.

I wrote some tests and they suggest that the feature doesn't work. :-(  So I am leaving that for a future pull request.

I addressed your other comments; thanks for them.

PR #3823, opened 2020-10-22
Enrich treatment of `@PolyNull`; fixes #3614

### LINKED ISSUE #3614: False alarm when using "return null" and type is known to support nulls
Sample code:

```java
  public static @PolyNull Boolean not(@PolyNull Boolean b) {
    return (b == null) ? null : !b;
  }
```

Error:

```
SqlFunctions.java:2571: error: [return.type.incompatible] incompatible types in return.
    return (b == null) ? null : !b;
                       ^
  type of expression: @Initialized @Nullable Boolean
  method return type: @Initialized @PolyNull Boolean
```

Could checker be a bit smarter, so it could see that {{b == null}} implies {{@PolyNull}} is in {{@Nullable}} mode, so it should be perfectly valid to return {{null}}.

I understand I can rewrite the method like {{return (b == null) ? b : !b;}}, however, it would be nice if checker could figure out that automatically.

Another example might involve regular generics without Poly annotations:

```java
  public <T> static T of(T a) {
    return a == null ? null : a;
  }
```

It would be nice if checker could figure out `return null` is reachable only in the case `T is nullable`.

Another example:

```java
  /** Converts a JDBC array to a list. */
  public static @PolyNull List arrayToList(final java.sql.@PolyNull Array a) {
    if (a == null) {
      return null; // <- error here
    }
    ...
```

mernst:


checker/src/main/java/org/checkerframework/checker/nullness/NullnessValue.java
```diff
@@ -3,18 +3,26 @@
...
+import org.checkerframework.dataflow.qual.SideEffectFree;
 import org.checkerframework.framework.flow.CFAbstractAnalysis;
 import org.checkerframework.framework.flow.CFAbstractValue;
 import org.checkerframework.framework.flow.CFValue;
+import org.checkerframework.javacutil.AnnotationUtils;
+import org.checkerframework.javacutil.TypesUtils;
 
 /**
  * Behaves just like {@link CFValue}, but additionally tracks whether at this point {@link PolyNull}
  * is known to be {@link Nullable}.
```
smillst, line 19: This needs to be updated.

smillst (approved):

mernst:
@smillst There have been enough changes that I think re-review is merited.  Thanks.

checker/src/main/java/org/checkerframework/checker/nullness/NullnessStore.java
```diff
@@ -44,22 +48,26 @@ public NullnessStore(
...
             boolean sequentialSemantics) {
         super(analysis, sequentialSemantics);
+        isPolyNullNonNull = false;
         isPolyNullNull = false;
     }
 
+    /**
+     * Create a NullnessStore (copy constructor).
+     *
+     * @param s a store te copy
```
smillst, line 58: `te` -> `to`

checker/src/main/java/org/checkerframework/checker/nullness/NullnessTransfer.java
```diff
@@ -205,6 +225,78 @@ protected NullnessValue finishValue(
...
+        return true;
+    }
+
+    /**
+     * Returns true if there is an occurrence of @PolyNull that is not at the top level.
+     *
+     * @param t a type
+     * @return true if there is an occurrence of @PolyNull that is not at the top level
+     */
+    private boolean containsPolyNullNotAtTopLevel(AnnotatedTypeMirror t) {
```
smillst, line 261: You can do this by using a SimpleAnnotatedTypeScanner:
```
new SimpleAnnotatedTypeScanner<>(
                    (type, p) ->
                            typeArg.hasAnnotation(POLYNULL),
                    Boolean::logicalOr,
                    false);
```

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractAnalysis.java
```diff
@@ -51,7 +51,7 @@
     protected final DependentTypesHelper dependentTypesHelper;
 
     /** A type factory that can provide static type annotations for AST Trees. */
-    protected final GenericAnnotatedTypeFactory<V, S, T, ? extends CFAbstractAnalysis<V, S, T>>
+    public final GenericAnnotatedTypeFactory<V, S, T, ? extends CFAbstractAnalysis<V, S, T>>
```
smillst, line 54: Should this change be reverted?  You can call `getTypeFactory()` to access this field.

smillst (approved):

PR #3824, opened 2020-10-23
New scripts for WPI that use dljc and -Ainfer=stubs

kelloggm:
From https://github.com/kelloggm/wpi-many.

- [x] Scripts and documentation
- [x] Add tests

I didn't remove `infer-and-annotate.sh` because someone might still want to use it if these scripts fail for some reason. If reviewers think it's not necessary at all, I'm open to removing it.


kelloggm:
@mernst, if you have any comments on the way I've incorporated these scripts and documentation into the CF/manual, I would appreciate them.

docs/manual/inference.tex
```diff
@@ -140,46 +142,86 @@
...
+  or Maven buildfile.
+
+\item[-t TIMEOUT]
+  The timeout for running the checker, in seconds
+
+\item[DLJC-ARGS]
+  Arguments that are passed directly to
+  \ahref{https://github.com/kelloggm/do-like-javac}{do-like-javac}'s
+  \<dljc> program without
+  modification.  One argument is required: \<--checker>, which indicates
```
mernst, line 205: The `dljc` usage string does not include `--checker` as a permitted command-line option.

checker/bin/wpi.sh
```diff
@@ -0,0 +1,170 @@
...
+# This script performs whole-program inference on a project directory.
+
+# For usage and requirements, see the "Whole-program inference"
+# section of the Checker Framework manual:
+# https://checkerframework.org/manual/#whole-program-inference
+
+
+while getopts "d:u:t:" opt; do
+  case $opt in
+    d) DIR="$OPTARG"
```
mernst, line 12: Is the `-d` command-line argument to `wpi.sh` needed?
Here are two ideas, either of which may simplify invocation of the script:
 * Make it optional, using the current directory if not supplied.
 * Remove it, and take the directory as the first non-command-line argument.  Then there is no need for the `--` on the wpi.sh command line, either.

checker/bin/wpi.sh
```diff
@@ -0,0 +1,170 @@
...
+      WPI_RESULTS_AVAILABLE="no"
+  fi
+}
+
+#### Check and setup dependencies
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+# clone or update DLJC
+if [ ! -d "${SCRIPTDIR}/../do-like-javac" ]; then
```
mernst, line 140: Running `wpi.sh` creates a new directory `checker/do-like-javac/`.
 * Please put it under `checker/bin/`.
 * Start the name with a period, as is done for the clone of `plume-scripts` and `run-google-java-format` in `checker/bin-devel`.
 * Put it in the `.gitignore`

mernst (changes requested):
`wpi.sh` crashed when I ran it.
I ran these commands:
```
cd ~/tmp
rm -rf bcel-util
git clone https://github.com/plume-lib/bcel-util.git
cd bcel-util
$CHECKERFRAMEWORK/checker/bin/wpi.sh -d . -- --checker nullness
```

Here is the output of the last command.  It looks like a bug in `do-like-javac`.

```
Already up to date.
~/tmp/bcel-util ~/tmp/bcel-util

Welcome to Gradle 6.6.1!

Here are the highlights of this release:
 - Experimental build configuration caching
 - Built-in conventions for handling credentials
 - Java compilation supports --release flag

For more details see https://docs.gradle.org/6.6.1/release-notes.html

Starting a Gradle Daemon (subsequent builds will be faster)
> Task :clean UP-TO-DATE

Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/6.6.1/userguide/command_line_interface.html#sec:command_line_warnings

BUILD SUCCESSFUL in 15s
1 actionable task: 1 up-to-date
/home/mernst/research/types/checker-framework/checker/bin/../do-like-javac/dljc -t wpi --checker nullness -- ./gradlew clean compileJava -g .gradle -Dorg.gradle.java.home=/home/mernst/java/jdk11
Traceback (most recent call last):
  File "/home/mernst/research/types/checker-framework/checker/bin/../do-like-javac/dljc", line 5, in <module>
    do_like_javac.run()
  File "/home/mernst/research/types/checker-framework-fork-kelloggm-branch-wpi-many/checker/do-like-javac/do_like_javac/command.py", line 33, in main
    tools.run(args, javac_commands, jars)
  File "/home/mernst/research/types/checker-framework-fork-kelloggm-branch-wpi-many/checker/do-like-javac/do_like_javac/tools/__init__.py", line 46, in run
    TOOLS[tool].run(args, javac_commands, jars)
  File "/home/mernst/research/types/checker-framework-fork-kelloggm-branch-wpi-many/checker/do-like-javac/do_like_javac/tools/wpi.py", line 152, in run
    stubs = os.listdir(wpiDir)
OSError: [Errno 2] No such file or directory: '/home/mernst/tmp/bcel-util/build/whole-program-inference'
~/tmp/bcel-util
```

kelloggm:
With regard to the crash you experienced when trying the scripts on `bcel-util`: what you encountered wasn't a bug in DLJC, though DLJC should probably been more friendly in how it reported the error. DLJC crashed because the first round of WPI failed, which you could tell by examining the `dljc-out` directory in `~/tmp/bcel-util`. There, I found a `wpi.log` file that explained why WPI fails. Its contents (on my machine) were:

```
?master ~/tmp/bcel-util/dljc-out> cat wpi.log 
Running echo "-----------------------------------------------------------"

"-----------------------------------------------------------"
Running /Users/kelloggm/jsr308/checker-framework/checker/bin/javac -Ainfer=stubs -AmergeStubsWithSource -Xmaxerrs 10000 -Xmaxwarns 10000 -J-Xbootclasspath/p:/Users/kelloggm/jsr308/checker-framework/checker/dist/javac.jar -Xbootclasspath/p:/Users/kelloggm/jsr308/checker-framework/checker/dist/jdk8.jar --release 8 -classpath /Users/kelloggm/tmp/bcel-util/.gradle/caches/modules-2/files-2.1/org.checkerframework.annotatedlib/bcel/6.5.0/ec3cad9d51bf4f6ee4c851129e6cedcfbd1c07c6/bcel-6.5.0.jar:/Users/kelloggm/tmp/bcel-util/.gradle/caches/modules-2/files-2.1/org.checkerframework/checker-qual/3.7.0/b35f5bd883a3f9cd1af39fa5a64ba12c9c5f0440/checker-qual-3.7.0.jar:/Users/kelloggm/tmp/bcel-util/.gradle/caches/modules-2/files-2.1/org.plumelib/reflection-util/1.0.2/6220181a7456a56e316049071d8e026440a1b007/reflection-util-1.0.2.jar -processor org.checkerframework.checker.nullness.NullnessChecker -Xlint:-processing -XDuseUnsharedTable=true -Xplugin:ErrorProne -Xep:ReferenceEquality:OFF -d /Users/kelloggm/tmp/bcel-util/build/classes/java/main -g -AcheckPurityAnnotations -sourcepath  -h /Users/kelloggm/tmp/bcel-util/build/generated/sources/headers/java/main -ArequirePrefixInWarningSuppressions -Xlint:all --release 8 -s /Users/kelloggm/tmp/bcel-util/build/generated/sources/annotationProcessor/java/main --add-opens=jdk.compiler/com.sun.tools.javac.comp=ALL-UNNAMED -XDcompilePolicy=simple -XepExcludedPaths:.*/org/plumelib/bcelutil/StackVer.java -Werror -AwarnUnneededSuppressions --release 8 -XepExcludedPaths:.*/org/plumelib/bcelutil/StackVer.java /Users/kelloggm/tmp/bcel-util/src/main/java/org/plumelib/bcelutil/NoConstraintsVisitor.java /Users/kelloggm/tmp/bcel-util/src/main/java/org/plumelib/bcelutil/JvmUtil.java /Users/kelloggm/tmp/bcel-util/src/main/java/org/plumelib/bcelutil/StackVer.java /Users/kelloggm/tmp/bcel-util/src/main/java/org/plumelib/bcelutil/StackMapUtils.java /Users/kelloggm/tmp/bcel-util/src/main/java/org/plumelib/bcelutil/StackTypes.java /Users/kelloggm/tmp/bcel-util/src/main/java/org/plumelib/bcelutil/SimpleLog.java /Users/kelloggm/tmp/bcel-util/src/main/java/org/plumelib/bcelutil/BcelUtil.java /Users/kelloggm/tmp/bcel-util/src/main/java/org/plumelib/bcelutil/package-info.java /Users/kelloggm/tmp/bcel-util/src/main/java/org/plumelib/bcelutil/InstructionListUtils.java

error: invalid flag: -Xep:ReferenceEquality:OFF
Usage: javac <options> <source files>
use --help for a list of possible options
```

It looks like `bcel-util` needs some special configuration of javac, or a special javac version, that dljc isn't able to detect. That's fine - the scripts aren't (and were never intended to be) foolproof ways to run arbitrarily-complex projects. They should work for all simple cases, and some complicated ones. (This is part of why I had left the instructions for `infer-and-annotate.sh`, but I see that you have removed them. I'm fine with this, but we should add a disclaimer to the effect of "WPI might not work. If it doesn't, too bad, use the instructions elsewhere and annotate your code by hand.")

I'll also mention that it looks like `bcel-util` is running the CF itself via the `checkerframework-gradle-plugin`. I did not design the WPI tool in dljc with this possibility in mind, and I don't really expect it to work - javac will end up with a number of similar or identical duplicate arguments, which may very well conflict. I don't think testing that rare scenario is worthwhile until we have this working on a project that doesn't use any of our other tooling.

mernst:
Thanks for the useful explanation.  That is very helpful.

The `-Xep:ReferenceEquality:OFF` command-line argument is for Error Prone.
([Here](https://github.com/plume-lib/bcel-util/blob/master/build.gradle#L55) is the relevant part of the buildfile.)

> it looks like bcel-util is running the CF itself via the checkerframework-gradle-plugin. I did not design the WPI tool in dljc with this possibility in mind, and I don't really expect it to work - javac will end up with a number of similar or identical duplicate arguments, which may very well conflict. I don't think testing that rare scenario is worthwhile until we have this working on a project that doesn't use any of our other tooling.

I don't think this is a rare scenario, but will be the common case.  I expect a common use case to be:
 * Set up build support for the Checker Framework.
 * See vast numbers of warnings.
 * Run inference.

This is a different use case than the original "run WPI on large numbers of untouched projects" use case.  I suspect that original use case to be a niche use case (though important in its own right).  The vast majority of users will want to run inference on a single project at a time.


1. Can you make the error reporting more user-friendly?  Now I know where to look, but other users won't.

2. Would it be possible for the scripts (`wpi` or `do-like-javac`, I'm not sure which) to discard all `-Xep:` command-line arguments?  I don't see any potential problems with that (though maybe I am missing something), and it will make the scripts handle projects that use Error Prone.  (Fuller support for Error Prone would be nice because its compiler fixes bugs in the javac in JDK 8.  However, I wouldn't spend time on that fuller support.)

3. Can you find a way to support projects that already use the Checker Framework in their build system?

kelloggm:
> I don't think this is a rare scenario, but will be the common case. I expect a common use case to be:
> 
> * Set up build support for the Checker Framework.
> * See vast numbers of warnings.
> * Run inference.

The tooling here is not appropriate for this scenario, because it does not insert annotations into the source code. Instead, it produces `.astub` side-files that contain the annotations.

This tooling was designed for a user who wants to run a typechecker without modifying the original source code. If they want to run the checker again, such a user would re-infer all the annotations on the new version; they would not commit the inferred annotations to their repository or keep track of them in any way. If this is the scenario you think is the primary use-case for these tools, then the tools will need to be significantly modified. I don't think that's a good idea, because the tools already handle the use-case they were designed for (which I described above) quite well.

> Can you make the error reporting more user-friendly? Now I know where to look, but other users won't.

I have already done so.

> Would it be possible for the scripts (wpi or do-like-javac, I'm not sure which) to discard all -Xep: command-line arguments? I don't see any potential problems with that (though maybe I am missing something), and it will make the scripts handle projects that use Error Prone. (Fuller support for Error Prone would be nice because its compiler fixes bugs in the javac in JDK 8. However, I wouldn't spend time on that fuller support.)

I could do this at the `do-like-javac` level. But, I think the right answer is for the user to set their `JAVA_HOME` variables to point at the errorprone compiler rather than standard JDK8/11 distributions if the target projects require an errorprone compiler to compile rather than modifying the scripts.

> Can you find a way to support projects that already use the Checker Framework in their build system?

I just haven't tested it. I don't know for sure that it will cause problems (though I suspect it will, as I said above), but I wanted you to know that the scripts weren't designed with this use-case in mind.

mernst:
> it does not insert annotations into the source code. Instead, it produces `.astub` side-files that contain the annotations.
> This tooling was designed for a user who wants to run a typechecker without modifying the original source code.

This is a crucial fact that I lost track of, because the new WPI scripts are in a section of the manual that states, "The inferred type qualifiers are inserted into your program."

More generally, the wpi script lacks documentation.  I wrote some documentation for it, but there is an unresolved TODO in the manual that carries over from that.  Could you fill that in, and also complete the documentation about what wpi-many outputs?

kelloggm:
> This is a crucial fact that I lost track of, because the new WPI scripts are in a section of the manual that states, "The inferred type qualifiers are inserted into your program."

This is partially my fault, because I edited the existing text about WPI without reading the surrounding context. As-is, I don't think these scripts fall under either of "kinds of type inference" that the section's introduction describes:
* "Type inference during type-checking: The type-checker fills in an appropriate type where the programmer didn't write one, but does not change the source code."
* "Type inference to annotate a program: As a separate step before type-checking, a type inference tool inserts type qualifiers into the source code."

Mechanically, these WPI scripts do the former - they fill in appropriate types where the program didn't write one. However, they do so using the mechanism described as the second kind of inference ("a separate step"), but without actually inserting the qualifiers into source code.

More generally, I don't think these two categorizations cover the entirely of what we mean colloquially when we say "type inference." I would argue that type inference is any process that takes a program and produces a set of types that are consistent with that program - whether that happens during typechecking or not, whether the types are then transferred to the source code or not, etc. I'd also argue that those two dimensions ("during typechecking or not" and "inserted into source or not") are both valid, orthogonal ways to split up type inference algorithms. The current text, however, confuses the two independent dimensions and conflates "inserting into source code" with "not running during typechecking". The new WPI scripts in this PR do not insert into source code, and also do not run during typechecking.

I propose that we restructure the entire section. First, we can keep the existing differentiation between "during typechecking" and "not during typechecking", and discuss local inference first and all global type inference tools second. The introduction should also be more clear about *why* one would want a global inference tool: it overcomes the limitation of local inference (i.e. that typechecking is modular), but does so at the cost of more compute time, because a whole-program analysis of some kind is required (this is true of CFI and of WPI, and I assume of the other tools mentioned (Cascade, nullness-specific tools). Any mention of whether annotations from global tools are inserted into source should be moved to that subsection, rather than the introduction, because it is only relevant for those tool.

In the "global inference tools" section, we can move some of the content from the existing manual section about the advantages and disadvantages of various approaches. I would then argue that we should include all the tools (including `infer-and-annotate.sh`) with a description of the use-case that each tool was designed for.

Related, I do NOT think that we should try to rewrite these scripts to be a drop-in replacement for `infer-and-annotate.sh`. In every case where I have wanted to use WPI, I've found these scripts to be more useful than `infer-and-annotate.sh`, but they do fill a slightly different role. For a user who wants to have a tool write annotations into their source code that they will then maintain, `infer-and-annotate.sh` is still the best option.

mernst:
Thanks for thinking this through.  Your suggested reorganization sounds good.  I'm not sure that infer-and-annotate.sh is worth keeping (due to some bugs in it that I know of), but we can remove it later.

I am hopeful that the wpi scripts can replace infer-and-annotate.sh eventually.  Once Jason's rewrite of the  `-Astubs` feature is complete, it will create `.ajava` files.  A user can just rename the .ajava files to .java.  (Or, it is easy to write a program that copies annotations from .ajava files into .java files.)

kelloggm:
> I am hopeful that the wpi scripts can replace infer-and-annotate.sh eventually

So am I, but I don't think we *can* get there without Jason's work on the `.ajava` version of WPI, because we don't have a method for automatically transferring annotations from stub files into source code. So, for now, we should keep `infer-and-annotate.sh` around for users who absolutely need that functionality, but also provide these WPI scripts for those who don't - that's the intention of this PR, not replacement.

kelloggm:
@mernst, I've re-written the WPI chapter in the manual, and re-added `infer-and-annotate.sh`. Can you take a look at the documentation and let me know if it makes sense now?

docs/manual/inference.tex
```diff
@@ -126,22 +92,209 @@
...
+\sectionAndLabel{Running whole-program inference on a single project}{wpi-one}
+
+A typical invocation of \<wpi.sh> is
+
+\begin{Verbatim}
+  wpi.sh -- --checker nullness
+\end{Verbatim}
+
+The result is a set of log files placed in the \<dljc-out> folder of the
+target project. The results of type-checking with each candidate set of
```
mernst, line 150: Can you further explain "each candidate set of annotations"?  Is that from each iteration of type-checking?  Users only want to see the results of the final iteration.  Say how to see that.

docs/manual/inference.tex
```diff
@@ -126,22 +92,209 @@
...
+\item If you are using a checker that is distributed with the Checker
+Framework, use \<wpi-many.sh> directly.
+\item If you are using a checker that is not distributed with the Checker
+Framework (also known as a "custom checker"), file
+\<custom-checker-example.sh> is a no-arguments
+script that serves as an example of how to use \<wpi-many.sh>.
+\end{itemize}
+
+Log files are copied into a results directory.
+For a failed run, the log file indicates the reason that WPI could not
```
mernst, line 236: This text belongs in the "single project" section, I think.  This text also differs slightly from what is already there.  Please move the text and reconcile the versions.

mernst (changes requested):
This looks good.  Thanks for all the improvements.  I have just a few small comments.

docs/manual/inference.tex
kelloggm: It does not. `wpi-many.sh` processes the log files produced by `wpi-one.sh` and filters out those for which no actual results are available (and putting a log file with the reason a project was removed in the results directory). It also copies the `dlcj-out/wpi.log` file for projects for which results actually were available. This pre-processing makes it easier for users running large experiments to go through all the results quickly, by placing the top-level result for each project in a single directory.

checker/bin/wpi.sh
```diff
@@ -0,0 +1,175 @@
...
+function configure_and_exec_dljc {
+
+  if [ -f build.gradle ]; then
+      if [ -f gradlew ]; then
+    chmod +x gradlew
+    GRADLE_EXEC="./gradlew"
+      else
+    GRADLE_EXEC="gradle"
+      fi
+      CLEAN_CMD="${GRADLE_EXEC} clean -g .gradle -Dorg.gradle.java.home=${JAVA_HOME}"
```
mernst, line 81: The `-g .gradle` command causes gradle to core dump, when the .gradle directory does not exist.  Is it needed?  Or, can it be supplied only if a `.gradle/` directory exists?

docs/manual/inference.tex
```diff
@@ -126,22 +84,209 @@
...
+  or Maven buildfile. The default is the current working directory.
+
+\item[-t TIMEOUT]
+  The timeout for running the checker, in seconds
+
+\item[DLJC-ARGS]
+  Arguments that are passed directly to
+  \ahref{https://github.com/kelloggm/do-like-javac}{do-like-javac}'s
+  \<dljc> program without
+  modification.  One argument is required: \<-\relax-checker>, which indicates
```
mernst, line 166: https://github.com/kelloggm/do-like-javac says:

> The --checker option is required. Its argument is the fully-qualified name of the checker you want to run.

Can `--checker` be supplied multiple times?  Or can I supply multiple checkers in comma-separated form, as in javac's -processor argument (https://checkerframework.org/manual/#shorthand-for-checkers)?

(Please fix that in do-like-javac rather than here, though this text might need to be tweaked slightly.)

docs/manual/inference.tex
```diff
@@ -126,22 +84,209 @@
...
+  what type-checker to run (use the fully-qualified name of the checker).
```
mernst, line 167: It seems to work to use shorthands as described in https://checkerframework.org/manual/#shorthand-for-checkers .

checker/bin/wpi.sh
```diff
@@ -0,0 +1,174 @@
...
+fi
+
+# support wpi-many.sh's ability to delete projects without usable results
+# automatically
+if [ "${WPI_RESULTS_AVAILABLE}" = "no" ]; then
+    echo "dljc could not run the build successfully"
+    touch .cannot-run-wpi
+fi
+
+popd || exit 1
```
mernst, line 174: It is distracting and confusing that the pushd and popd commands produce user-visible output.

checker/bin/wpi.sh
```diff
@@ -0,0 +1,174 @@
...
+    export JAVA_HOME="${JAVA8_HOME}"
+    echo "couldn't build using Java 11; trying Java 8"
+    configure_and_exec_dljc "$@"
+    export JAVA_HOME="${JAVA11_HOME}"
+fi
+
+# support wpi-many.sh's ability to delete projects without usable results
+# automatically
+if [ "${WPI_RESULTS_AVAILABLE}" = "no" ]; then
+    echo "dljc could not run the build successfully"
```
mernst, line 170: Please add output telling the user where to look for diagnostic output.
I think the answer is "in the `dljc-out` directory".

kelloggm:
> > > Would it be possible for the scripts (wpi or do-like-javac, I'm not sure which) to discard all -Xep: command-line arguments? I don't see any potential problems with that (though maybe I am missing something), and it will make the scripts handle projects that use Error Prone. (Fuller support for Error Prone would be nice because its compiler fixes bugs in the javac in JDK 8. However, I wouldn't spend time on that fuller support.)
> > 
> > 
> > I could do this at the `do-like-javac` level. But, I think the right answer is for the user to set their `JAVA_HOME` variables to point at the errorprone compiler rather than standard JDK8/11 distributions if the target projects require an errorprone compiler to compile rather than modifying the scripts.
> 
> This does not seem to support the use case of running WPI on GitHub projects that already use Error Prone.
> 
> Also, I think the `JAVA_HOME*` environment variables should not be changed. The goal is to use a different executable for `javac`, but the rest of the JDK is unchanged, and Error Prone does not ship with a JDK.

I've added a check in DLJC that will ignore ErrorProne compiler options, as you suggested. I didn't realize that ErrorProne only provides a compiler, not a full JDK distribution.

checker/bin/wpi.sh
kelloggm: That's what the `pushd` and `popd` commands do, and I'm under the impression that they're still the preferred way to switch directories (vs `cd`). Personally, I don't find it distracting - especially because the user isn't meant to look at the output of `wpi.sh` or `wpi-many.sh` unless something goes wrong. The user should look at the produced log files instead.

Do you have a specific suggestion for how to avoid the output, such as a different set of commands to use? One option, suggested by [this stack overflow post](https://stackoverflow.com/questions/25288194/dont-display-pushd-popd-stack-across-several-bash-scripts-quiet-pushd-popd) is to redefine the commands to pipe their output to `/dev/null`, but doing that always makes me nervous.

mernst: > Personally, I don't find it distracting

Maybe because that's because you know the implementation.  I had to read the script to determine that this was normal output and not part of an error message.

> The user should look at the produced log files instead.

That's reasonable.  Maybe the script should produce:
 * no output when it succeeds.
 * a single line saying to look at the log file when it fails.

> Do you have a specific suggestion for how to avoid the output, such as a different set of commands to use?

I don't have a specific suggestion for you.  When the directories are known (as in this case), I use `cd`, because I don't see that pushd and popd provide any benefit.  (Am I missing some benefit?)

kelloggm: > I don't have a specific suggestion for you. When the directories are known (as in this case), I use cd, because I don't see that pushd and popd provide any benefit. (Am I missing some benefit?)

Using `pushd`/`popd` is safer, because it stores the history of the directories that have been the current working directory. `cd` only stores the most recent directory (`-`). If the script encounters a problem we didn't anticipate that causes the directory structure to be messed up, then the `pushd`/`popd` approach is more likely to recover, in my experience. That's the only benefit I know of.

> That's reasonable. Maybe the script should produce:
> * no output when it succeeds.
> * a single line saying to look at the log file when it fails.

Even when the script succeeds, the user should look at the log file. The documentation indicates this: 

> The result is a set of log files placed in the \<dljc-out> folder of the
target project. The results of type-checking with each candidate set of
annotations will be concatenated into the file \<dljc-out/wpi.log>; the final
results (i.e. those obtained using the most precise, consistent set of annotations)
will appear at the end of this file.

A simpler option than changing the scripts would be to modify the documentation to note that any output from the script itself is merely informational. Would that alleviate your concern?

mernst: I didn't see any error-handling code that uses the pushd/popd stack, but maybe I missed it.

Maybe the script should prefix and suffix its output with a remark to look elsewhere.  (Because users might not remember all details of the documentation.)  Even if so, it's friendly to users to remove output that isn't useful.

checker/bin/wpi.sh
kelloggm: As we discussed on our call, I intended to just guard this behind a flag. But I did some digging and now I remember why this is here. There is a relevant Gradle issue (https://github.com/gradle/gradle/issues/1319), which I'll link from a TODO comment in the script. The gist is that Gradle doesn't play nicely with distributed file systems (like those used on the cycle servers) because it tries to put its caches in the user's home directory by default (which can cause either contention if you're using multiple servers or use too much space. And in the cycle servers case, it's also slow! The `-g` option controls where those caches go, in addition to controlling where to look first for user configuration etc. So this is here to force gradle to cache everything in the `.gradle` directory relative to the project, which means each project gets its own cache.

I've implemented a check for whether `.gradle` exists, and create it if not. This leaves the gradle command line invocation unchanged. This should fix your immediate problem, but it leaves the larger problem of not sharing caches; I'm not sure if we want to fix that or not.

mernst: If it is convenient to do, I would prefer to use a location in /scratch/$USER/ or /tmp/$USER/.  This would permit sharing the cache and would reduce repeated downloads.  This isn't critical but would be nice to do.

kelloggm: That would work on the CSE cycle servers, but these scripts should be portable and we can't know a priori what's part of a shared file system and what is not. For that reason, I think it's better to default this to a directory that's exclusive to the project, even if it makes these scripts less efficient.

You're right that on a system like the CSE file servers it would be nice to have an option that will change where this default is. I've added one: `-g`. Hopefully that will address your concerns.

smillst (approved):

checker/bin-devel/Makefile
```diff
@@ -0,0 +1,18 @@
+remove-annotations.sh:
```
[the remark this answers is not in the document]
mernst: A Makefile target (unless it is marked as PHONY) is a file that the rule creates.  So, this rule creates the file `remove-annotations-sh`.

kelloggm:
@mernst automated testing is definitely working now, and CI is green. I think this is ready to be merged. Can you take another look and let me know if you agree? (I'd like to get this in for this release.)

mernst (approved):

Issue #3400, opened 2020-06-23
Inheritance of classes with a qualifier parameter


smillst:
If a class extends a `@HasQualifierParameter` class (or implements a
`@HasQualifierParameter` interface), then that class must also be marked
`@HasQualifierParameter`. A particular subclass should be able to specify which qualifier
parameter should be used for the super class.  (This is similar to `class StringList implements List<String> {...}`.) For example,
```java
 @HasQualifierParameter(Tainted.class)
 class Buffer {
     void append(@PolyTainted Buffer this, @PolyTainted String s) { ... }
 }
 @HasQualifierParameter(Tainted.class)
 @Tainted class TaintedBuffer extends @Tainted Buffer {
     @Override
     void append(@Tainted TaintedBuffer this, @Tainted String s) { ... } // legal override
 }
```
`@Untainted TaintedBuffer` is an invalid type.

t-rasmud:
#3832 makes `@HasQualifierParameter` an inherited type annotation addressing the requirement: "If a class extends a `@HasQualifierParameter` class (or implements a `@HasQualifierParameter` interface), then that class must also be marked
`@HasQualifierParameter`. "
The requirement "A particular subclass should be able to specify which qualifier parameter should be used for the super class." remains to be addressed.

PR #3832, opened 2020-10-25
Makes HasQualifierParameter an inherited type; Fixes #3831

### LINKED ISSUE #3831: @HasQualifierParameter doesn't apply to anonymous classes
```
import org.checkerframework.framework.qual.HasQualifierParameter;

public class InnerHasQualifierParameter {

    interface TestInterface {
        public void testMethod();
    }

    public void test() {
        TestInterface test =
                new TestInterface() {
                    public void testMethod() {}
                };
    }
}
```
Running the tainting checker with the command `javac -processor tainting checker/tests/tainting/InnerHasQualifierParameter.java` results in the following error:
```
checker/tests/tainting/InnerHasQualifierParameter.java:14: error: [missing.has.qual.param] Missing @HasQualifierParameter for hierarchy @Tainted.
                new TestInterface() {
                                    ^
  Class extends or implements a type that has a qualifier parameter.
1 error

```

While this is an expected output, there is no way to explicitly annotate the anonymous class with `@HasQualifierParameter`.
Proposed solution: Make `@HasQualifierParameter` an inherited annotation type.

t-rasmud:


framework/src/main/java/org/checkerframework/framework/qual/HasQualifierParameter.java
```diff
@@ -1,11 +1,6 @@
 package org.checkerframework.framework.qual;
 
-import java.lang.annotation.Annotation;
-import java.lang.annotation.Documented;
-import java.lang.annotation.ElementType;
-import java.lang.annotation.Retention;
-import java.lang.annotation.RetentionPolicy;
-import java.lang.annotation.Target;
+import java.lang.annotation.*;
```
mernst, line 3: Our style is to not use wildcard imports.  Please just add one import statement.

mernst (changes requested):
I don't see a test case that shows that `@HasQualifierParameter` is inherited when not written explicitly.  Could you add one?

How is this related to #3400?

framework/src/main/java/org/checkerframework/framework/qual/NoQualifierParameter.java
```diff
@@ -17,6 +17,9 @@
  * parameters. This annotation may not be written on the same class as {@code HasQualifierParameter}
  * for the same hierarchy.
  *
+ * <p>It is an error for a superclass to be {@code @HasQualifierParameter} but a subclass to be
+ * {@code @NoQualifierParameter} for the same hierarchy.
+ *
  * @see HasQualifierParameter
```
smillst, line 23: This annotation should be inherited, too.

t-rasmud:
> How is this related to #3400?
This PR addresses the first requirement of #3400:  "If a class extends a `@HasQualifierParameter` class (or implements a `@HasQualifierParameter` interface), then that class must also be marked `@HasQualifierParameter`."
It does not address the second requirement: "A particular subclass should be able to specify which qualifier parameter should be used for the super class."

mernst (changes requested):
Maybe I missed it, but still I don't see the test case that I asked about:

> I don't see a test case that shows that @HasQualifierParameter is inherited when not written explicitly. Could you add one?

I see that there in an unannotated subclass:
```class SubNoQualifierParameter extends InheritNoQualifierParameter {}```
but I don't see a test that confirms that the class `SubNoQualifierParameter` is treated as having a qualifier parameter.

t-rasmud:
> Maybe I missed it, but still I don't see the test case that I asked about:
> 
> > I don't see a test case that shows that @HasQualifierParameter is inherited when not written explicitly. Could you add one?
> 
> I see that there in an unannotated subclass:
> `class SubNoQualifierParameter extends InheritNoQualifierParameter {}`
> but I don't see a test that confirms that the class `SubNoQualifierParameter` is treated as having a qualifier parameter.

I have now added tests that confirm `SubHasQualifierParameter` is treated as having a qualifier parameter and `SubNoQualifierParameter` is treated as not having a qualifier parameter.

mernst (approved):

mernst:
> > How is this related to #3400?
> 
> This PR addresses the first requirement of #3400: "If a class extends a `@HasQualifierParameter` class (or implements a `@HasQualifierParameter` interface), then that class must also be marked `@HasQualifierParameter`."
> It does not address the second requirement: "A particular subclass should be able to specify which qualifier parameter should be used for the super class."

When this PR lands, please add a note at #3400 to explain what has been done and what remains to be done. Thanks!

t-rasmud:
@mernst That was an oversight on my part. @smillst Could you please re-open the PR if there are more changes? Thanks.

PR #3834, opened 2020-10-25
Abstract out fields to avoid re-computation

mernst:


dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java
```diff
@@ -3645,12 +3650,12 @@ public Node visitOther(Tree tree, Void p) {
...
-     * Returns the TypeElement for the given class.
+     * Returns the TypeMirror for the given class.
      *
      * @param clazz a class
-     * @return the TypeElement for the class
+     * @return the TypeMirror for the class
      */
-    private TypeElement getTypeElement(Class<?> clazz) {
-        return elements.getTypeElement(clazz.getCanonicalName());
+    private TypeMirror getTypeMirror(Class<?> clazz) {
```
smillst, line 3658: This could be replaced with `org.checkerframework.javacutil.TypesUtils#typeFromClass`.

smillst (approved):

Issue #3287, opened 2020-05-03
[BUG] Checker allows to unsoundly override a @SafeEffect method as @UIEffect if the method is inherited indirectly


PawelLipski:
**Checker version**: 3.4.0

**Affected checkers**: `GuiEffectChecker`, but possibly a similar problems occurs in other checkers where method annotations are taken into account.

**The problem**:

Given the 2 following interfaces (actually, most likely those can be classes as well):

```java
public interface IBaseInterface {
  void run();
}
```

```java
interface IDerivedInterface extends IBaseInterface { }
```

and a class implementing/overriding the `@SafeEffect` method as `@UIEffect`:

```java
public class ImplementingClass implements IDerivedInterface {
  @Override
  @UIEffect
  public void run() {}
}
```

we should get:

`error: [[all, guieffect]:override.effect.invalid] A method override may only be @UI if it overrides an @UI method.`

but instead the compilation completes successfully.

**Sample repo where the problem can be reproduced**: see the following PR, just run `./gradlew compileJava`: https://github.com/PawelLipski/git-machete-intellij-plugin/pull/1

PawelLipski:
Also, out of curiosity: why does the message say `A method override may only be @UI if it overrides an @UI method` instead of `A method override may only be @UIEffect if it overrides an @UIEffect method`? `@UI` annotation doesn't even target methods.

csgordon:
That message text may date back to before ```@UIEffect``` was named as it is now (a long time ago), and should probably be fixed as well.

csgordon:
The bug only arises in cases where the inherited method comes from an interface; if a direct super*class* defines the method as ```@SafeEffect``` the inheritance check catches invalid overrides at ```@UIEffect```.  A method I thought reflected semantically present elements only reflects elements present directly in the textual source.  A naive fix is relatively straightforward: search all transitively-reachable parent interfaces (whether the class implements an interface or subinterface, or extends a class that implements such an interface).  I'll prototype a naive solution (brute force search) as a baseline, but then I'll implement a caching search so running time doesn't tank on more complex inheritance hierarchies.

PawelLipski:
@csgordon any progress on this one? :)

csgordon:
[...]

PR #3841, opened 2020-10-28
Fix for #3287

csgordon:
This is a work-in-progress PR to correct #3287.

The old override checking code first traversed the direct class inheritance chain to check overrides, then iterated over the direct supertypes (i.e., base class and directly declared interfaces).  This left a hole when a method was inherited from either:

- An interface extended (possibly transitively) by a directly implemented interface, where the directly-implemented interface did not contain its own override; or
- An interface implemented by a class which did not provide its own override of the relevant method.

Basically, a method ```m``` defined in ```U``` and overridden in class or interface ```T``` wasn't correctly checked for overrides unless T and U were both class types with T <: U, or if U was an interface, if class T directly implemented U or interface T directly extended U.

The fix was to replace the traversal code (originally from 2012) with a use of comparatively newer helper functions that already correctly handle transitive inheritance along all relevant paths.

This fix appears functionally complete, but requires a bit more cleanup. I will remove the "WIP" and update the PR when it is actually ready for review.


mernst:
@csgordon Is this ready for review now?  You seem to have disabled the "Work in progress tickbox" (though I can't be sure because GitHub records no notifications about that), but the title and text still mention work in progress.

checker/src/main/java/org/checkerframework/checker/guieffect/GuiEffectTypeFactory.java
```diff
@@ -548,7 +519,7 @@ public AnnotatedTypeMirror getAnnotatedType(Tree tree) {
         // We don't need to issue warnings for inheriting from poly and a concrete effect.
```
mernst, line 519: Should "inheriting from" be "overriding"?

checker/src/main/java/org/checkerframework/checker/guieffect/GuiEffectTypeFactory.java
```diff
@@ -420,126 +441,76 @@ public AnnotatedTypeMirror getAnnotatedType(Tree tree) {
         ExecutableElement polyOverride = null;
```
mernst, line 441: As a minor comment, I think the above 3 variables would be better named *Overridden rather than *Override, since they contain a method of the given effect that is overridden.

mernst (approved):

checker/src/main/java/org/checkerframework/checker/guieffect/GuiEffectTypeFactory.java
csgordon: I've fixed this bit.

checker/src/main/java/org/checkerframework/checker/guieffect/GuiEffectTypeFactory.java
csgordon: Good suggestions, I've renamed them.

csgordon:
Since @mernst marked it approved, once the CI succeeds for the style fixes, I'll merge.

PR #3842, opened 2020-10-28
Don't clear type parameters in StubParser

jwaataja:
The `StubParser` class keeps track of the type variables in scope for the elements it's processing. When processing a class, it should add the class's type parameters on entry and remove them when leaving. The current implementation clears *all* type parameters when it finishes processing a class. This means after processing an inner class, all type parameters are cleared, including those of the enclosing class.


mernst (approved):

mernst:
Do not merge yet.  @jwaataja will add a test.

mernst:
@jwaataja This commit causes there to be two files `StubparserTest.java` and `StubParserTest.java` with no clear distinction between them.  Why do they need to be separate, and can you document the difference between them?  (Or merge them if they don't need to be separate.)

PR #3843, opened 2020-10-29
sideEffectsUnrefineAliases feature

mernst:
Read the comment in SourceChecker.java first.

The diffs are not as big as they look.  (GitHub ought to pass the `--minimal` command-line argument to diff.)

This pull request lacks a test case, but there is one here:
https://github.com/mernst/checker-framework/compare/sideEffectsUnrefineAliases...mernst:hasnext-checker?expand=1

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/nullness/KeyForSubchecker.java
+++ b/checker/src/main/java/org/checkerframework/checker/nullness/KeyForSubchecker.java
@@ -9,11 +9,4 @@ import org.checkerframework.common.basetype.BaseTypeChecker;
  * @checker_framework.manual #map-key-checker Map Key Checker
  * @checker_framework.manual #nullness-checker Nullness Checker
  */
-public class KeyForSubchecker extends BaseTypeChecker {
-    {
-        // While strictly required for soundness, this leads to too many false positives.  Printing
-        // a key or putting it in a map erases all knowledge of what maps it was a key for.
-        // TODO: Revisit when side effect annotations are more precise.
-        // this.sideEffectsUnrefineAliases = true;
-    }
-}
+public class KeyForSubchecker extends BaseTypeChecker {}
--- a/checker/src/main/java/org/checkerframework/checker/nullness/KeyForSubchecker.java
+++ b/checker/src/main/java/org/checkerframework/checker/nullness/KeyForSubchecker.java
@@ -9,4 +9,8 @@ import org.checkerframework.common.basetype.BaseTypeChecker;
  * @checker_framework.manual #map-key-checker Map Key Checker
  * @checker_framework.manual #nullness-checker Nullness Checker
  */
-public class KeyForSubchecker extends BaseTypeChecker {}
+public class KeyForSubchecker extends BaseTypeChecker {
+    {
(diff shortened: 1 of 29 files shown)


framework/src/main/java/org/checkerframework/framework/flow/CFAbstractStore.java
```diff
@@ -199,54 +199,73 @@ public void updateForMethodCall(
...
+                        .removeIf(e -> !e.getKey().isUnmodifiableByOtherCode());
+            }
+
+            // update this value
+            if (analysis.checker.sideEffectsUnrefineAliases) {
+                thisValue = null;
+            }
+
+            // update field values
+            if (analysis.checker.sideEffectsUnrefineAliases) {
```
smillst, line 218: Nit: I would probably move move this if/else to a helper method to reduce the indentation.

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -479,6 +479,17 @@
...
+    /**
+     * Should the analysis assume that side effects to a value can change the type of aliased
+     * references?
+     *
+     * <p>For many type systems, once a local variable's type is refined, side effects to the
+     * variable's value do not change the variable's type annotations. For some type systems, a side
+     * effect to the value could change them; set this field to true.
+     */
+    // Not final so that subclasses can set it.
+    public boolean sideEffectsUnrefineAliases = false;
```
smillst, line 491: Since this field is only used by checkers that use dataflow, it should be in GenericAnnotatedTypeFactory.

smillst (changes requested):

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractStore.java
mernst: Let's put this off for now, and revisit the indentation issue later.

mernst:
@smillst I moved the field you requested.

checker/src/main/java/org/checkerframework/checker/nullness/KeyForSubchecker.java
```diff
@@ -9,4 +9,11 @@
  * @checker_framework.manual #map-key-checker Map Key Checker
  * @checker_framework.manual #nullness-checker Nullness Checker
  */
-public class KeyForSubchecker extends BaseTypeChecker {}
+public class KeyForSubchecker extends BaseTypeChecker {
+    {
+        // While strictly required for soundness, this leads to too many false positives.  Printing
+        // a key or putting it in a map erases all knowledge of what maps it was a key for.
+        // TODO: Revisit when side effect annotations are more precise.
+        // this.sideEffectsUnrefineAliases = true;
```
smillst, line 17: This needs to move, too.

smillst (approved):

checker/src/main/java/org/checkerframework/checker/nullness/KeyForSubchecker.java
mernst: Thanks for catching that.  I found one other reference that needed to be changed, too.

PR #3847, opened 2020-11-02
Define classes `SubtypeIsSubsetQualifierHierarchy` and `SubtypeIsSupersetQualifierHierarchy`

mernst:
~~Merge after #3846~~

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/type/SubtypeIsSubsetQualifierHierarchy.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/SubtypeIsSubsetQualifierHierarchy.java
@@ -123,8 +123,9 @@ public class SubtypeIsSubsetQualifierHierarchy extends MostlyNoElementQualifierH
      *
      * @param kind the qualifier kind
      * @param values the annotation's {@code values} element/argument
+     * @return an annotation of the given kind and values
      */
-    public AnnotationMirror createAnnotationMirrorWithValue(
+    private AnnotationMirror createAnnotationMirrorWithValue(
             QualifierKind kind, LinkedHashSet<String> values) {
         AnnotationBuilder builder = new AnnotationBuilder(processingEnv, kind.getAnnotationClass());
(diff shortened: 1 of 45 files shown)


framework/src/main/java/org/checkerframework/framework/type/SubtypeIsSubsetQualifierHierarchy.java
```diff
@@ -0,0 +1,134 @@
...
+import javax.lang.model.element.ExecutableElement;
+import org.checkerframework.framework.qual.AnnotatedFor;
+import org.checkerframework.framework.util.QualifierKind;
+import org.checkerframework.javacutil.AnnotationBuilder;
+import org.checkerframework.javacutil.AnnotationUtils;
+import org.checkerframework.javacutil.BugInCF;
+
+/**
+ * A {@link org.checkerframework.framework.type.QualifierHierarchy} where, when a qualifier has
+ * arguments, the subtype relation is determined by a subset test on the arguments.
```
smillst, line 21: `arguments` should be `elements`.

framework/src/main/java/org/checkerframework/framework/type/SubtypeIsSubsetQualifierHierarchy.java
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }
+
+    /**
+     * Returns a mutable list containing the {@code values} element of the given annotation.
+     *
+     * @param anno an annotation
+     * @return a mutable list containing the {@code values} element; may be the empty list
+     */
+    private List<String> extractValues(AnnotationMirror anno) {
```
smillst, line 111: This class could be changed to operate on lists of Objects rather than Strings. If you don't want to make that change, then the class comment should mention that the annotation element but be an array of strings.
\

smillst (changes requested):

smillst (approved):

PR #3849, opened 2020-11-03
Use relative path when passing filenames to reformat and checkformat

smillst:


build.gradle
```diff
@@ -600,6 +600,7 @@ subprojects {
         description 'Format the Java source code'
         // checker-qual* projects have no source, so skip
         onlyIf { !project.name.startsWith('checker-qual') }
+//        workingDir projectDir
```
mernst, line 603: Should this comment be deleted?

mernst (changes requested):
I have a question about a comment that was added.

build.gradle
smillst: Yes, thanks!

mernst (approved):

PR #3852, opened 2020-11-03
Reset uninferredTypeArgumentScanner.

### LINKED ISSUE #3851: 3.7.1: uninferredTypeArgumentScanner consumes significant amount of memory in AnnotatedTypeMirror
Apparently, verification of Calcite codebase requires to bump `-Xmx` to 2g or something like that, and it turns out noticeable memory is taken by `SimpleAnnotatedTypeScanner`:

https://github.com/typetools/checker-framework/blob/9dbf0c7c06210a9cb926242a8181eb0dd888ab04/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeMirror.java#L813-L820

Could it be a static instance and/or `ThreadLocal`?

`NullnessAnnotatedTypeFactory` consumed ~50MiB

Снимок экрана 2020-11-04 в 1 10 30

~ 50% of that was consumed by empty `IdentityHashMap` in `SimpleAnnotatedTypeScanner`

Снимок экрана 2020-11-04 в 1 11 01

There are 72226 `SimpleAnnotatedTypeScanner` instances which are referenced from `AnnotatedTypeMirror$...` classes

Снимок экрана 2020-11-04 в 1 11 29

smillst:
Fixes #3851.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
@@ -90,6 +90,7 @@ import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedPrimitiv
 import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedTypeVariable;
 import org.checkerframework.framework.type.AnnotatedTypeMirror.AnnotatedWildcardType;
 import org.checkerframework.framework.type.visitor.AnnotatedTypeCombiner;
+import org.checkerframework.framework.type.visitor.SimpleAnnotatedTypeScanner;
 import org.checkerframework.framework.util.AnnotatedTypes;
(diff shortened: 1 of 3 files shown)


vlsi:
It won't help, since:
1) `IdentityHashMap` does not de-allocate backing array
2) It won't reduce the number of `SimpleAnnotatedTypeScanner` instances

smillst:
@vlsi How about not saving the anonymous class in a field?

vlsi:
That would reduce memory consumption, however, I wonder if that would impact verification time (e.g. re-instantiation might take time).

smillst:
@vlsi I moved `uninferredTypeArgumentScanner` to the AnnotatedTypeFactory.  In that location, there should only be one scanner.  Thanks for your comments!

mernst (approved):

mernst:
Let's see how this impacts memory and run time.  If its impact is negative, we can revert it.

PR #3854, opened 2020-11-04
Define glb more precisely for the Nullness Checker

mernst:
This is important for whole-program inference, which uses GLB to combine stubs and source code.

The code for new method `greatestLowerBoundInitialization` is very similar to that for existing method `leastUpperBoundInitialization`.

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
+++ b/checker/src/main/java/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
@@ -946,7 +946,7 @@ public abstract class InitializationAnnotatedTypeFactory<
          * Returns the greatest lower bound of two types.
          *
          * @param a the first argument
-         * @param a the second argument
+         * @param b the second argument
          * @return the glb of the two arguments
          */
         protected TypeMirror glbTypeFrame(TypeMirror a, TypeMirror b) {
--- a/checker/src/main/java/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
+++ b/checker/src/main/java/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
@@ -888,5 +888,75 @@ public abstract class InitializationAnnotatedTypeFactory<
 
             return TypesUtils.leastUpperBound(a, b, processingEnv);
         }
+
+        /**
+         * Compute the greatest lower bound of two initialization qualifiers. Returns null if one of
+         * the qualifiers is not in the initialization hierarachy. Subclasses should override
+         * greatestLowerBound and call this method for initialization qualifiers.
+         *
+         * @param anno1 an initialization qualifier
+         * @param qual1 a qualifier kind
+         * @param anno2 an initialization qualifier
+         * @param qual2 a qualifier kind
+         * @return the glb of anno1 and anno2
+         */
+        protected AnnotationMirror greatestLowerBoundInitialization(
+                AnnotationMirror anno1,
+                QualifierKind qual1,
+                AnnotationMirror anno2,
+                QualifierKind qual2) {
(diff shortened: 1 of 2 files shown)


checker/src/main/java/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -888,5 +888,75 @@ protected TypeMirror lubTypeFrame(TypeMirror a, TypeMirror b) {
...
+         * @return the glb of the two arguments
+         */
+        protected TypeMirror glbTypeFrame(TypeMirror a, TypeMirror b) {
+            if (types.isSubtype(a, b)) {
+                return a;
+            } else if (types.isSubtype(b, a)) {
+                return b;
+            }
+
+            return TypesUtils.greatestLowerBound(a, b, processingEnv);
```
smillst, line 959: If a and b are unrelated, then this will return an intersection type or an error type.  `createUnderInitializationAnnotation` uses the erased type of the intersection type and will throw an exception on the error type (I think).  Should bottom be returned instead?

checker/src/main/java/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
```diff
@@ -888,5 +888,75 @@ protected TypeMirror lubTypeFrame(TypeMirror a, TypeMirror b) {
...
+
+        /**
+         * Returns the greatest lower bound of two types.
+         *
+         * @param a the first argument
+         * @param b the second argument
+         * @return the glb of the two arguments
+         */
+        protected TypeMirror glbTypeFrame(TypeMirror a, TypeMirror b) {
+            if (types.isSubtype(a, b)) {
```
smillst, line 953: This is duplicated in `TypesUtils.greatestLowerBound`, so you could just call it directly.

smillst (changes requested):

checker/src/main/java/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
mernst: I don't see this special case in [`TypesUtils.greatestLowerBound`](https://github.com/typetools/checker-framework/blob/master/javacutil/src/main/java/org/checkerframework/javacutil/TypesUtils.java#L751).  Am I missing something?

checker/src/main/java/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
mernst: By "bottom", do you mean the type of `void`?  That doesn't seem like a good argument/element to `@UnderInitialization`.
Maybe it would be better to do this computation and then issue a warning if the result is an intersection or error type.

checker/src/main/java/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
smillst: `TypesUtils.greatestLowerBound` calls `com.sun.tools.javac.code.Types#glb(com.sun.tools.javac.code.Type, com.sun.tools.javac.code.Type)` which has the special case.

checker/src/main/java/org/checkerframework/checker/initialization/InitializationAnnotatedTypeFactory.java
smillst: I mean that the glb of the two annotation mirrors should be `@FBCBottom` if the glb of the elements of the annotation mirrors is an intersection or an error type.

mernst: That makes sense, thanks.

smillst (approved):

PR #3861, opened 2020-11-05
New method ElementUtils.getSimpleNameOrDescription

mernst:
I'm not thrilled about the method name; suggestions for improving it are welcome.

### DIFF
--- a/javacutil/src/main/java/org/checkerframework/javacutil/ElementUtils.java
+++ b/javacutil/src/main/java/org/checkerframework/javacutil/ElementUtils.java
@@ -206,6 +206,25 @@ public class ElementUtils {
     }
 
     /**
+     * Returns a user-friendly name for the given method. Does not return {@code "<init>"} or {@code
+     * "<clinit>"} as ExecutableElement.getSimpleName() does.
+     *
+     * @param element a method declaration
+     * @return a user-friendly name for the method
+     */
+    public static CharSequence getFriendlyName(ExecutableElement element) {
+        Name result = element.getSimpleName();
+        switch (result.toString()) {
+            case "<init>":
+                return element.getEnclosingElement().getSimpleName();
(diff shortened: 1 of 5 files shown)


javacutil/src/main/java/org/checkerframework/javacutil/ElementUtils.java
```diff
@@ -205,6 +205,25 @@ public static String getSimpleName(ExecutableElement element) {
...
     }
 
+    /**
+     * Returns a user-friendly name for the given method. Does not return {@code "<init>"} or {@code
+     * "<clinit>"} as ExecutableElement.getSimpleName() does.
+     *
+     * @param element a method declaration
+     * @return a user-friendly name for the method
+     */
+    public static CharSequence getFriendlyName(ExecutableElement element) {
```
smillst, line 215: I agree this name isn't great.  I think `getUserFriendlyName` would be an improvement.  Here are a few other, mostly bad names that might spark an idea of a better name:

* `getUserFriendlySimpleName`
* `getUserFriendlyName`
* `getPlainSimpleName`
* `getUserUnderstandableSimpleName`
* `getPrintableSimpleName`
* ` getLegibleSimpleName`

smillst (approved):
The code changes look good.  I added a comment about the name.

javacutil/src/main/java/org/checkerframework/javacutil/ElementUtils.java
mernst: I like `getUserFriendlyName` best among these suggestions (and it's better than the current name).
Another idea is:
 * `getSimpleNameOrDescription`

smillst: I like `getSimpleNameOrDescription` best.

PR #3863, opened 2020-11-05
In `updateFromReturn`, remove Java expressions that contain local variables

mernst:
`updateFromReturn` transfers a type from a return expression to a method return type.  References to local variables make sense within the method body but not in the method signature.

Similar standardization might be necessary elsewhere.

Reinstate Javadoc checking after merging.

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/lock/LockAnnotatedTypeFactory.java
+++ b/checker/src/main/java/org/checkerframework/checker/lock/LockAnnotatedTypeFactory.java
@@ -170,9 +170,10 @@ public class LockAnnotatedTypeFactory
                     String expression,
                     FlowExpressionContext context,
                     TreePath localScope,
-                    boolean useLocalScope) {
+                    boolean useLocalScope,
+                    boolean removeErroneousExpressions) {
                 if (DependentTypesError.isExpressionError(expression)) {
-                    return expression;
+                    return null;
                 }
 
                 // Adds logic to parse <self> expression, which only the Lock Checker uses.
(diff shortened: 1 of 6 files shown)


checker/src/main/java/org/checkerframework/checker/lock/LockAnnotatedTypeFactory.java
```diff
@@ -170,9 +170,10 @@ protected String standardizeString(
                     String expression,
                     FlowExpressionContext context,
                     TreePath localScope,
-                    boolean useLocalScope) {
+                    boolean useLocalScope,
+                    boolean removeErroneousExpressions) {
                 if (DependentTypesError.isExpressionError(expression)) {
-                    return expression;
+                    return null;
```
smillst, line 176: This should only return `null` if `removeErroneousExpressions` is true.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -600,8 +684,18 @@ private AnnotationMirror standardizeDependentTypeAnnotation(
...
-                standardizedStrings.add(
-                        standardizeString(expression, context, localScope, useLocalScope));
+                String standardized =
+                        standardizeString(
+                                expression,
+                                context,
+                                localScope,
+                                useLocalScope,
+                                removeErroneousExpressions);
+                if (standardized == null) {
```
smillst, line 694: Instead of passing `removeErroneousExpressions` to `standardizeString`, this if statement could check if `standardized` is an error string via `DependentTypesError#isExpressionError`.

smillst (changes requested):

smillst (approved):

Issue #3850, opened 2020-11-03
Regression in 3.7.1: InternalUtils.symbol: tree is null


vlsi:
Code: https://github.com/vlsi/calcite/blob/78d97778f938ab435851487f2e91fcfd1d293914/core/src/main/java/org/apache/calcite/sql/parser/SqlParserPos.java#L41

The error is not present in 3.7.0

Error:

```
error: InternalUtils.symbol: tree is null
  ; The Checker Framework crashed.  Please report the crash.
  Compilation unit: .../calcite/core/src/main/java/org/apache/calcite/sql/parser/SqlParserPos.java
  Last visited tree at line 41 column 1:
  public class SqlParserPos implements Serializable {
  Exception: java.lang.Throwable; java.lang.Throwable
  	at org.checkerframework.javacutil.BugInCF.<init>(BugInCF.java:16)
  	at org.checkerframework.javacutil.TreeUtils.elementFromTree(TreeUtils.java:459)
  	at org.checkerframework.javacutil.TreeUtils.elementFromDeclaration(TreeUtils.java:517)
  	at org.checkerframework.checker.nullness.NullnessTransfer.strengthenAnnotationOfEqualTo(NullnessTransfer.java:206)
  	at org.checkerframework.checker.nullness.NullnessTransfer.strengthenAnnotationOfEqualTo(NullnessTransfer.java:62)
  	at org.checkerframework.framework.flow.CFAbstractTransfer.visitEqualTo(CFAbstractTransfer.java:744)
  	at org.checkerframework.framework.flow.CFAbstractTransfer.visitEqualTo(CFAbstractTransfer.java:94)
  	at org.checkerframework.dataflow.cfg.node.EqualToNode.accept(EqualToNode.java:31)
  	at org.checkerframework.dataflow.analysis.AbstractAnalysis.callTransferFunction(AbstractAnalysis.java:338)
  	at org.checkerframework.dataflow.analysis.ForwardAnalysisImpl.callTransferFunction(ForwardAnalysisImpl.java:390)
  	at org.checkerframework.dataflow.analysis.ForwardAnalysisImpl.performAnalysisBlock(ForwardAnalysisImpl.java:130)
  	at org.checkerframework.dataflow.analysis.ForwardAnalysisImpl.performAnalysis(ForwardAnalysisImpl.java:107)
  	at org.checkerframework.framework.flow.CFAbstractAnalysis.performAnalysis(CFAbstractAnalysis.java:109)
[...]

PR #3864, opened 2020-11-05
getContainingMethod returns null for lambdas; fixes issue #3850

mernst:


smillst (approved):
It looks like you can add a `getContainingLambda()` to analysis (and cfg) and then add handling to polyNullIsNonNull for lambdas.  But let's fix the crash first.

PR #3873, opened 2020-11-07
Move ContractsUtils instance from visitor to type factory

mernst:
Merge after https://github.com/typetools/checker-framework/pull/3872.  Or, merge this one and close #3872.


smillst (approved):
I like this version better.

framework/src/main/java/org/checkerframework/framework/util/ContractsUtils.java
```diff
@@ -39,33 +38,16 @@
...
-     * Returns an instance of the {@link ContractsUtils} class for the given factory. Also sets it
-     * as the currently-used ContractsUtils object.
+     * Creates a ContractsUtils for the given factory.
      *
-     * @param factory the factory to create a ContractsUtils for
-     * @return a ContractsUtils for the given factory
+     * @param factory the type factory associated with the newly-created ContractsUtils
      */
-    @SuppressWarnings("interning")
-    public static ContractsUtils getInstance(GenericAnnotatedTypeFactory<?, ?, ?, ?> factory) {
```
smillst, line 64: This change should be mentioned in the change log.

mernst:
> I like this version better.

Do you mean compared to #3872?  This incorporates all the changes in #3872, plus more.  The point of merging #3872 was to make two smaller and simpler commits rather than one larger one.  If you prefer just one commit in the history, that is OK.

smillst:
> Do you mean compared to #3872? This incorporates all the changes in #3872, plus more. The point of merging #3872 was to make two smaller and simpler commits rather than one larger one. If you prefer just one commit in the history, that is OK.

This pull request removes methods that were added in #3872, so I thought this was just an alternative to #3872.  It doesn't make sense to add `BaseTypeVistor#getContractsUtils` in one comment and then remove it in the next.  So, I would just merge this one.

PR #3877, opened 2020-11-09
Define method getDefaultValueAnnotatedType

mernst:
The CI failures are due to missing "WARNING: Illegal reflective access" in `expected.txt` files.  I'd like to find a way to avoid those warnings.

They occur because this implementation uses reflection to access private members.  What is a better implementation?

In case it helps, my end goal is to determine whether a field's annotation is compatible with the default value that Java uses for uninitialized fields.  That is, the client code looks like this:

```java
            AnnotatedTypeMirror fieldType = defaultValueAtypeFactory.getAnnotatedType(field);
            AnnotatedTypeMirror defaultValueType =
                    defaultValueAtypeFactory.getDefaultValueAnnotatedType(
                            fieldType.getUnderlyingType());
            if (!defaultValueAtypeFactory
                    .getTypeHierarchy()
                    .isSubtype(defaultValueType, fieldType)) {
```

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
@@ -2208,6 +2208,7 @@ public abstract class GenericAnnotatedTypeFactory<
      * @param typeMirror a type
      * @return the annotated type of {@code type}'s default value
      */
+    // TODO: Cache results to avoid recomputation.
     public AnnotatedTypeMirror getDefaultValueAnnotatedType(TypeMirror typeMirror) {
         return getAnnotatedType(getDefaultValueTree(typeMirror));
     }
--- a/framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
@@ -16,12 +16,8 @@ import com.sun.source.tree.Tree.Kind;
 import com.sun.source.tree.UnaryTree;
 import com.sun.source.tree.VariableTree;
 import com.sun.source.util.TreePath;
-import com.sun.tools.javac.code.Type;
 import com.sun.tools.javac.code.TypeTag;
-import com.sun.tools.javac.tree.JCTree.JCLiteral;
 import java.lang.annotation.Annotation;
-import java.lang.reflect.Constructor;
-import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Comparator;
@@ -30,7 +26,6 @@ import java.util.HashSet;
 import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.NoSuchElementException;
 import java.util.Queue;
 import java.util.Set;
 import java.util.StringJoiner;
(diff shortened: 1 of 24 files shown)


framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -2174,4 +2181,76 @@ private boolean isRelevantHelper(TypeMirror tm) {
...
+
+    /**
+     * Calls the protected JCLiteral constructor.
+     *
+     * @param typeTag the literal's type tag
+     * @param value the literal's value
+     * @param typeMirror the typeMirror for the literal
+     * @return a JCLiteral tree for the given type tag and value
+     */
+    private static JCLiteral createJCLiteral(TypeTag typeTag, Object value, TypeMirror typeMirror) {
```
smillst, line 2247: `org.checkerframework.javacutil.trees.TreeBuilder#buildLiteral` seems to build a literal tree from a given value.  

You should be able to do something like this:

```
    public static LiteralTree buildLiteral(Object value, ProcessingEnvironment env) {
        Context context = ((JavacProcessingEnvironment) env).getContext();
         TreeMaker maker = TreeMaker.instance(context);
        return maker.Literal(value);
    }
```

(This sort of method should be in TreeUtils.)

In case it's not easy to figure out.  Here are the imports for the code above:
```
import com.sun.tools.javac.tree.TreeMaker;
import javax.annotation.processing.ProcessingEnvironment;
import com.sun.tools.javac.util.Context;
import com.sun.tools.javac.processing.JavacProcessingEnvironment;
```

mernst:
@smillst This is ready for re-review, but don't merge it yet because I am still running tests on clients of it.

javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -1641,4 +1644,25 @@ public static boolean isWideningBinary(BinaryTree node) {
...
+     * @return a LiteralTree for the given type tag and value
+     */
+    public static LiteralTree createLiteral(
+            TypeTag typeTag,
+            Object value,
+            TypeMirror typeMirror,
+            ProcessingEnvironment processingEnv) {
+        Context context = ((JavacProcessingEnvironment) processingEnv).getContext();
+        TreeMaker maker = TreeMaker.instance(context);
+        LiteralTree result = maker.Literal(TypeTag.BOT, null);
```
smillst, line 1664: ` maker.Literal(TypeTag.BOT, null)` should be ` maker.Literal(typeTag, value)`, right?  Why doesn't `maker.Literal(typeTag, value)` work? I don't think there are any test for this, so I can't check myself.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -2198,4 +2200,46 @@ private boolean isRelevantHelper(TypeMirror tm) {
...
+    }
+
+    /**
+     * Return a tree for the default value of the given type. The default value is 0, false, or
+     * null.
+     *
+     * @param typeMirror a type
+     * @return a tree for {@code type}'s default value
+     */
+    private LiteralTree getDefaultValueTree(TypeMirror typeMirror) {
```
smillst, line 2223: This should be in `TreeUtils` because it imports a class from outside the support complier API, (com.sun.tools.javac.code.TypeTag).

smillst (approved):
It's fine with me if you want to merge this, after addressing my last round of comments.

mernst:
Thanks.  That fixed the downstream jobs.

PR #3882, opened 2020-11-09
Make an implicit behavior explicit

mernst:
I don't know whether this optimization is worthwhile.
(The documentation changes are probably worthwhile even if the optimization is not.)

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java
@@ -215,6 +215,8 @@ public class DefaultTypeHierarchy extends AbstractAtmComboVisitor<Boolean, Void>
     /**
      * Compare the primary annotations of {@code subtype} and {@code supertype}.
      *
+     * @param subtype the type that might be a subtype
+     * @param supertype the type that might be a supertype
      * @param annosCanBeEmtpy indicates that annotations may be missing from the typemirror
      * @return true if the primary annotation on subtype {@literal <:} primary annotation on
      *     supertype for the current top. Returns true if both annotations are null. Returns false
--- a/framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java
@@ -219,8 +219,8 @@ public class DefaultTypeHierarchy extends AbstractAtmComboVisitor<Boolean, Void>
      * @param supertype the type that might be a supertype
      * @param annosCanBeEmtpy indicates that annotations may be missing from the typemirror
      * @return true if the primary annotation on subtype {@literal <:} primary annotation on
-     *     supertype for the current top. Returns true if both annotations are null. Returns false
-     *     if one annotation is null and the other is not.
+     *     supertype for the current top. Returns true if neither annotated type has a primary
+     *     annotation. Returns false if one annotation is does and the other does not.
      */
     protected boolean isPrimarySubtype(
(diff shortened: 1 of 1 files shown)


framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java
```diff
@@ -215,10 +215,12 @@ protected boolean isPrimarySubtype(AnnotatedTypeMirror subtype, AnnotatedTypeMir
     /**
      * Compare the primary annotations of {@code subtype} and {@code supertype}.
      *
+     * @param subtype the type that might be a subtype
+     * @param supertype the type that might be a supertype
      * @param annosCanBeEmtpy indicates that annotations may be missing from the typemirror
      * @return true if the primary annotation on subtype {@literal <:} primary annotation on
-     *     supertype for the current top or both annotations are null. False is returned if one
-     *     annotation is null and the other is not.
+     *     supertype for the current top. Returns true if both annotations are null. Returns false
```
smillst, line 222: This method operates on AnnotatedTypeMirrors, so instead of, `if both annotations` this should be if `neither annotated type has a primary annotation`.

framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java
```diff
@@ -247,8 +249,13 @@ protected boolean isAnnoSubtype(
             @Nullable AnnotationMirror subtypeAnno,
             @Nullable AnnotationMirror supertypeAnno,
             boolean annosCanBeEmtpy) {
-        if (annosCanBeEmtpy && subtypeAnno == null && supertypeAnno == null) {
-            return true;
+        if (annosCanBeEmtpy) {
```
smillst, line 252: Without this change, if only one of the annotations is null, ` qualifierHierarchy.isSubtype(subtypeAnno, supertypeAnno);` would throw an exception.  But the Javadoc says that `false` should be returned in that case, so I think this is a good change.

smillst (approved):

framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java
mernst: If that is the case, then no client has ever taken advantage of this documented behavior.  Would it be better to simplify the specification rather than making the implementation more complex?  (I'm not sure which way I lean on that question.)

smillst: I don't have a good argument for either option.  Really if both annotations are `null` this method shouldn't be called, so throwing an exception in that case makes sense.  So maybe it is better to change the documentation.

smillst: Mention type variables in javadoc.  Change the Javadoc to match implementation.

framework/src/main/java/org/checkerframework/framework/type/DefaultTypeHierarchy.java
```diff
@@ -209,80 +208,18 @@ protected String defaultErrorMessage(
...
-
-    /**
-     * Compare the primary annotations of {@code subtype} and {@code supertype}.
-     *
-     * @param annosCanBeEmtpy indicates that annotations may be missing from the typemirror
-     * @return true if the primary annotation on subtype {@literal <:} primary annotation on
-     *     supertype for the current top or both annotations are null. False is returned if one
-     *     annotation is null and the other is not.
-     */
-    protected boolean isPrimarySubtype(
```
smillst, line 223: Both this method and `isAnnoSubtype` either called with `annosCanBeEmtpy = false` or the code checked that the annotations were nonnull before calling them.  So I just deleted these methods.

PR #3887, opened 2020-11-10
Initialized Fields Checker

mernst:


framework/src/main/java/org/checkerframework/common/initializedfields/qual/InitializedFieldsPredicate.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * A predicate annotation for the Initialized Fields type system.
+ *
+ * @checker_framework.manual #initialized-fields-checker Initialized Fields Checker
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE_USE, ElementType.TYPE_PARAMETER})
+@SubtypeOf({InitializedFields.class})
+public @interface InitializedFieldsPredicate {
```
kelloggm, line 17: Why do you need `InitializedFieldsPredicate`? I can't think of a situation where it would be useful.

kelloggm: (There is a constructor in `AccumulationAnnotatedTypeFactory` that doesn't require a predicate annotation.)

framework/src/main/java/org/checkerframework/framework/qual/PostconditionAnnotation.java
```diff
@@ -33,10 +33,10 @@
...
  * {@literal @}PostconditionAnnotation(qualifier = MinLen.class)
  * {@literal @}Target({ElementType.METHOD, ElementType.CONSTRUCTOR})
- * public {@literal @}interface EnsuresMinLen {
- *   String[] value();
- *   {@literal @}QualifierArgument("value")
- *   int targetValue() default 0;
+ *  public {@literal @}interface EnsuresMinLen {
+ *    String[] value();
+ *    {@literal @}QualifierArgument("value")
+ *    int targetValue() default 0;
```
kelloggm, line 39: Extraneous whitespace changes?

framework/tests/all-systems/GenericExtendsTypeVars.java
```diff
@@ -20,11 +20,20 @@ void context() {
...
+    }
 }
 
 @SuppressWarnings("initialization.fields.uninitialized")
 class RecursiveTypevarClass<T extends RecursiveTypevarClass<T>> {
     T t;
+
+    RecursiveTypevarClass(T t) {
+        this.t = t;
+    }
```
kelloggm, line 36: Does this new constructor remove the need for the suppress warnings above?

kelloggm (changes requested):

checker/src/main/java/org/checkerframework/checker/index/lowerbound/LowerBoundAnnotatedTypeFactory.java
```diff
@@ -175,7 +176,9 @@ public void addComputedTypeAnnotations(Tree tree, AnnotatedTypeMirror type, bool
         // If dataflow shouldn't be used to compute this type, then do not use the result from
         // the Value Checker, because dataflow is used to compute that type.  (Without this,
         // "int i = 1; --i;" fails.)
-        if (iUseFlow && tree != null && TreeUtils.isExpressionTree(tree)) {
+        if (tree != null
+                && TreeUtils.isExpressionTree(tree)
+                && (iUseFlow || tree instanceof LiteralTree)) {
```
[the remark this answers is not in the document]
mernst: Without this, the interaction between the Initialized Fields Checker and the Index Checker fails, because the Initialized Fields Checker calls this (transitively) with iUseFlow set to false.

Literals don't depend on flow, so this change seems safe.

framework/src/main/java/org/checkerframework/framework/qual/PostconditionAnnotation.java
mernst: The whitespace was wrong before; this makes it line up in the generated HTML.  I could move it into a different pull request if you prefer.

framework/src/main/java/org/checkerframework/common/initializedfields/InitializedFieldsAnnotatedTypeFactory.java
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    /**
+     * Returns the fields that the constructor must initialize. These are the fields F declared in
+     * this class that satisfy all of the following conditions:
+     *
+     * <ul>
+     *   <li>F is a non-final field (if final, Java will issue a warning, so we don't need to).
+     *   <li>F's declaration has no initializer.
+     *   <li>No initialization block or static initialization block sets the field. (This seems to
```
smillst, line 169: I think these fields are still included in the returned array.  It's just that dataflow visits initialization block or static initialization block as part of the constructor.  So, I would delete this bullet point.

framework/src/main/java/org/checkerframework/common/initializedfields/qual/EnsuresInitializedFields.java
```diff
@@ -0,0 +1,33 @@
...
+@PostconditionAnnotation(qualifier = InitializedFields.class)
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD})
+public @interface EnsuresInitializedFields {
+    /**
+     * The object(s) whose fields this method initializes.
+     *
+     * @return object(s) whose fields are initialized
+     */
+    public String[] value() default {"this"};
```
smillst, line 24: What does `@EnsuresInitializedFields(value = {"this", "a"}, fields = { "c", "d"})` mean?  I think that `value` should just be a string.  Then `@EnsuresInitializedFields` should be repeatable.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1633,7 +1634,7 @@ protected final void addComputedTypeAnnotations(Tree tree, AnnotatedTypeMirror t
      */
     protected void addComputedTypeAnnotations(
             Tree tree, AnnotatedTypeMirror type, boolean iUseFlow) {
-        assert root != null
+        assert root != null || tree instanceof LiteralTree
```
smillst, line 1637: This seems wrong to me.  I'm assuming there's a problem because of the LiteralTrees created by the Initialized Fields Checker.  I'll see if I can find a better work around.

smillst (changes requested):

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
mernst: @smillst That is correct.  The synthetic LiteralTrees lack some information, such as a path in source code.  There are a few hard-coded checks for `LiteralTree` in this pull request.  Fixing that would be nice.

mernst:
@kelloggm Can you give more comments or approve the PR?  Thanks.

docs/manual/initialized-fields-checker.tex
```diff
@@ -0,0 +1,183 @@
...
+\includeimage{initializedfields}{4.5cm}
+\caption{The type qualifier hierarchy of the Initialized Fields Checker.
+\<@InitializedFieldsBottom> is rarely written by a programmer.}
+\label{fig-initialized-fields-hierarchy}
+\end{figure}
+
+Figure~\ref{fig-initialized-fields-hierarchy} shows the the subtyping
+relationships among the type qualifiers.
+
+% TODO: draw a diagram of the type hierarchy.
```
kelloggm, line 131: I think this is done, so this line can be removed.

kelloggm (approved):
Sorry, I didn't realize that this was ready for another review. It looks good to me now, once Suzanne's comments have been resolved.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
smillst: I was able to remove these.  See my last commit.

smillst (approved):
One last minor comment, then I think this is ready to merge.

PR #3891, opened 2020-11-13
Add AssignmentContexts for parameters of constructor calls

msridhar:
It seems these contexts should also be present when expressions are passed as a parameter to a constructor call (in addition to normal method invocations).  They would be useful for a checker we are working on.

@smillst I'm not sure how to write a test for this.  Maybe these additional contexts could be relevant to type inference?

### DIFF
--- a/dataflow/src/main/java/org/checkerframework/dataflow/cfg/node/ObjectCreationNode.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/cfg/node/ObjectCreationNode.java
@@ -5,6 +5,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
+import javax.lang.model.element.ExecutableElement;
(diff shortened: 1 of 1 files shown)


smillst (approved):
I don't see anyway to add a test for this, but the code looks good.

msridhar:
Thanks, @smillst!  Re: a test, I was wondering if there is any existing test where doing correct type inference relies on having a proper `AssignmentContext` for a method parameter.  If there is such a test, then maybe this change could be tested with similar code re-written to use a constructor parameter instead?  Just a thought, nothing critical.

smillst:
@msridhar I don't know of any place that we use the assignment context of method parameters from dataflow.  The assignment context is passed to the AnnotatedTypeFactory, but we only use it for the type of NewArrayTree and NewClassTree with diamond.  Code for both of those trees is a bit buggy, so I wouldn't want to write a test against them.

PR #3901, opened 2020-11-16
Prefer `getKind()` to `instanceof`

mernst:


smillst (approved):

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -1178,10 +1178,11 @@ public AnnotatedTypeMirror fromElement(Element elt) {
                 ElementAnnotationApplier.apply(type, elt, this);
             }
         } else if (decl instanceof ClassTree) {
+            // decl.getKind() might be CLASS, INTERFACE, etc.
```
wmdietl, line 1181: You could use TreeUtils.isClassTree https://github.com/typetools/checker-framework/blob/master/javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java#L925 if you also want to remove this instanceof.

mernst: Thanks for the tip.  I tried that change.  When the next line is a cast, using `instanceof` makes it obvious that the cast will succeed:

```java
        } else if (decl instanceof ClassTree) {
            type = fromClass((ClassTree) decl);
```

That fact isn't as obvious when using the utility method:

```java
        } else if (TreeUtils.isClassTree(decl)) {
            type = fromClass((ClassTree) decl);
```

I don't have a strong opinion, though.  I am willing to make the change if you prefer.

@wmdietl @smillst What do you think?  What should the policy be about when to use `instanceof` versus `getKind()`?

wmdietl: One could make the same argument against all the uses of `getKind()`. In

```java
        if (tree.getKind() == Tree.Kind.MEMBER_SELECT) {
            MemberSelectTree mstree = (MemberSelectTree) tree;
````

it also isn't as clear as with an instanceof/cast combination that the cast will succeed.
In this case I find the `isClassTree` pretty suggestive that a cast to `ClassTree` will succeed afterwards.

For `Element`s and `TypeMirror`s there is an explicit recommendation to prefer `getKind()` comparisons: https://github.com/typetools/jdk/blob/master/src/java.compiler/share/classes/javax/lang/model/element/Element.java#L48
https://github.com/openjdk/jdk/blob/master/src/java.compiler/share/classes/javax/lang/model/type/TypeMirror.java#L45

I didn't find an explicit statement for `Tree`s.

smillst: It sounds like using `getKind` isn't necessary for Trees, so we could use instanceof in cases like these.  I don't feel strongly either way. Note we do have a section in the manual about this, so if we change the policy for trees it should be updated, too. https://checkerframework.org/manual/#creating-using-the-apis

mernst: The manual says "Do not use instanceof to determine the class of the object, because you cannot necessarily predict the run-time type of the object that implements an interface."
The manual doesn't explicitly say whether using instanceof to test an interface is acceptable.  Do you think it should be acceptable?

smillst: I think it's acceptable for trees, because there isn't an explicit recommendation to use getKind instead.  For Elements and Types it's less clear.  Reading the documentation Werner linked to, seems to indicate instanceof should be not used with classes, but it's not clear about interfaces.  So, I don't know.  I would error on using getKind for Types and Elements.

wmdietl: TypeMirror and Element are interfaces and the problem exists more for interfaces, where you want to allow that one class implements multiple interfaces. The same problem doesn't exist with classes.
Tree is also an interface, so by the same logic we should use getKind.

PR #3903, opened 2020-11-17
Compute which processors were auto-discovered

mernst:


checker/src/main/java/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
```diff
@@ -712,4 +713,13 @@ protected boolean isNullableOrAlias(AnnotationMirror am) {
         }
         return AnnotationUtils.areSameByName(am, NULLABLE);
     }
+
+    @Override
+    public AnnotatedTypeMirror getDefaultValueAnnotatedType(TypeMirror typeMirror) {
+        AnnotatedTypeMirror result = super.getDefaultValueAnnotatedType(typeMirror);
+        if (getAnnotationByClass(result.getAnnotations(), Nullable.class) != null) {
```
smillst, line 720: Is this change supposed to be in this PR?

smillst (approved):

checker/src/main/java/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
mernst: It is related, in that tests using auto-discovery fail without it.  (Those tests are not yet in this repository.)

PR #3915, opened 2020-11-20
Print the underlying exception message.

smillst:


mernst (commented):
Suggestion:  I would pass `t` as a second argument to the BugInCF constructor.  That's strictly more information.

smillst:
`t` is passed as the second argument.  It's helpful to add `t`'s message to the new exception so I don't have to hunt down the message in the second stack trace.

mernst:
I see.  OK, then.  I usually assume information isn't repeated and might not think to look later in the stack trace if t.getMessage was repeated at the top.  But I don't object to this change.

PR #3916, opened 2020-11-21
Tweak recommendation about getKind(), and add links to Javadoc

mernst:


docs/manual/creating-a-checker.tex
```diff
@@ -2327,12 +2327,14 @@
...
-implements an interface.  Instead, use the \code{getKind()} method.  The
-method returns \refModelclass{type}{TypeKind},
-\refModelclass{element}{ElementKind}, and \refTreeclass{tree}{Tree.Kind}
-for the three interfaces, respectively.
+implements an interface.  Instead, use the
+\sunjavadoc{java.compiler/javax/lang/model/type/TypeMirror.html\#getKind()}{TypeMirror.getKind()}
+or
+\sunjavadoc{java.compiler/javax/lang/model/element/Element.html\#getKind()}{Element.getKind()}
+method.
+For \<Tree>s, you can use either \<Tree.getKind()> or \<instanceof>.
```
wmdietl, line 2337: Wouldn't it be more consistent to also recommend `Tree.getKind()` for `Trees`? Did you find any recommendation to use `instanceof` for `Trees`?
#3901 changes the code to use `Tree.getKind()` and if we had a clear recommendation, we could check for consistent use automatically.

smillst (approved):

docs/manual/creating-a-checker.tex
mernst: It would be possible to recommend using `Tree.getKind()`.

Pro:
 * more consistent with other model interfaces like `Element`.
Con:
 * requires a change to code, which currently uses `instanceof`.  (#3901 is the change)
 * code becomes less clear.  The current `instanceof` tests always guard a cast.  It is idiomatic to do `instanceof` then a cast, and such code is obviously correct.  Using `getKind()` requires a reader to understand the relationship between the kinds and the classes.  There is already a high overhead to someone reading the code to understand all the different interfaces, and using getKind() more would increase it just a little bit more.
 * not necessary, since there is only one implementation of `Tree`
 * not recommended anywhere:  I found no guidance to use `Tree.getKind()`.

wmdietl: #3901 is already done, so that doesn't seem like a con.
The second and third con equally apply to Element and TypeMirror, so following a consistent style would actually reduce the mental overhead, by applying a simple rule: if the interface provides a `getKind()` method, use it. Then one doesn't need to think about whether this is an Element/TypeMirror/Tree and has a consistent recommendation.

PR #3918, opened 2020-11-22
Change boolean to enum for readability; fixes #3390

### LINKED ISSUE #3390: Create an enum in Store.java to indicate the location of a store (before or after a node) that is required.
As @kelloggm pointed out in https://github.com/typetools/checker-framework/pull/3370:

```
    /**
     * Runs the analysis again within the block of {@code node} and returns the store at the
     * location of {@code node}. If {@code before} is true, then the store immediately before the
     * {@link Node} {@code node} is returned. Otherwise, the store immediately after {@code node} is
     * returned. If {@code analysisCaches} is not null, this method uses a cache. {@code
     * analysisCaches} is a map of a block of node to the cached analysis result. If the cache for
     * {@code transferInput} is not in {@code analysisCaches}, this method creates new cache and
     * stores it in {@code analysisCaches}. The cache is a map of nodes to the analysis results of
     * the nodes.
     *
     * @param node the node to analyze
     * @param before the boolean value to indicate which store to return (if it is true, return the
     *     store immediately before {@code node}; otherwise, the store after {@code node} is
     *     returned)
     * @param transferInput the transfer input of the block of this node
     * @param nodeValues abstract values of nodes
     * @param analysisCaches caches of analysis results
     * @return the store before or after {@code node} (depends on the value of {@code before}) after
     *     running the analysis
     */
    S runAnalysisFor(
            Node node,
            boolean before,
            TransferInput<V, S> transferInput,
            IdentityHashMap<Node, V> nodeValues,
            Map<TransferInput<V, S>, IdentityHashMap<Node, TransferResult<V, S>>> analysisCaches);
```

> I would also consider changing before to an enum, so that callsites say Store.BEFORE and Store.AFTER rather than just true and false. Distinguishing between true and false at callsites makes them harder to read.

mernst:
Fixes #3390

### DIFF
--- a/dataflow/src/main/java/org/checkerframework/dataflow/analysis/Analysis.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/analysis/Analysis.java
@@ -30,6 +30,17 @@ public interface Analysis<
     }
 
     /**
+     * In calls to {@code Analysis#runAnalysisFor}, whether to return the store before or after the
+     * given node.
+     */
+    public static enum PreOrPost {
+        /** Return the pre-store. */
+        BEFORE,
+        /** Return the post-store. */
+        AFTER
+    }
+
+    /**
      * Get the direction of this analysis.
      *
      * @return the direction of this analysis
@@ -68,9 +79,8 @@ public interface Analysis<
      * the nodes.
      *
      * @param node the node to analyze
-     * @param before the boolean value to indicate which store to return (if it is true, return the
-     *     store immediately before {@code node}; otherwise, the store after {@code node} is
-     *     returned)
+     * @param preOrPost indicates which store to return: the store immediately before {@code node}
+     *     or the store after {@code node}
      * @param blockTransferInput the transfer input of the block of this node
      * @param nodeValues abstract values of nodes
      * @param analysisCaches caches of analysis results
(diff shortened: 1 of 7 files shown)


dataflow/src/main/java/org/checkerframework/dataflow/analysis/Store.java
```diff
@@ -45,6 +45,17 @@
         BOTH_TO_ELSE,
     }
 
+    /**
+     * In calls to {@code Analysis#runAnalysisFor}, whether to return the store before or after the
+     * given node.
+     */
+    public static enum PreOrPost {
```
wmdietl, line 52: This doesn't seem to be used and is duplicating the new enum in `Analysis`.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/Analysis.java
```diff
@@ -29,6 +29,17 @@
         BACKWARD
     }
 
+    /**
+     * In calls to {@code Analysis#runAnalysisFor}, whether to return the store before or after the
+     * given node.
+     */
+    public static enum PreOrPost {
```
wmdietl, line 36: How about either calling this `BeforeOrAfter` or renaming the enum values to `PRE` and `POST`? Mixing the names is inconsistent.

smillst (approved):

mernst:
@wmdietl @smillst Thanks for the reviews.

PR #3924, opened 2020-11-23
Define method ElementUtils.closure

mernst:


smillst (approved):

javacutil/src/main/java/org/checkerframework/javacutil/ElementUtils.java
```diff
@@ -661,4 +664,20 @@ public static TypeElement getTypeElement(ProcessingEnvironment processingEnv, Cl
...
+
+    /**
+     * Get all the supertypes of a given type, including the type itself.
+     *
+     * @param type a type
+     * @param env the processing environment
+     * @return list including the type and all its supertypes, with a guarantee that supertypes
+     *     (i.e. those that appear in extends clauses) appear before indirect supertypes
+     */
+    public static List<TypeElement> closure(TypeElement type, ProcessingEnvironment env) {
```
wmdietl, line 676: `closure` is a very non-specific name for this method. What does it close over? All supertypes? Or all subtypes? Or something else?
Wasn't the previous `getAllSupertypes` much clearer?

mernst: Yes, I also prefer the name `getAllSupertypes`.
This method is just a wrapper around an existing method in `com.sun.tools.javac.code.Types`.  I want to make the correspondence clear, and I don't want to invent new terminology.

wmdietl: In many recent changes we preferred clear terminology based on the JLS over implementation terminology from javac.
This would seem like a clear instance where we should not feel bound to an obscure term that is used in an implementation class and instead strive for clarity and uniformity in our utility classes.

mernst: I think you are right.  I will change this.  Thanks for the suggestion.

PR #3922, opened 2020-11-23
Implement capture conversion.

smillst:
Merge with https://github.com/plume-lib/plume-util/pull/157.

This pull request implements capture conversion.  This required several related changes:
1. Re-implementing `isContainedBy` in `DefaultTypeHierarchy`.
2. Removing various workarounds.

To understand capture conversion, you need to understand the following sections of the JLS:
 * 5.1.10.
 * 4.5.1 containment
 * 4.10 subtyping (and note that "|T|" means "the erasure of T")

Reviewing order:
1. Capture conversion: `AnnotatedTypeFactory#applyCaptureConversion(AnnotatedTypeMirror)`
2. Containment: `DefaultTypeHierarchy#isContainedBy`
3. All other changes. (Note some of these are removing work arounds for not having capture conversion.)

Fixes #3013, fixes #3754, fixes #3791, fixes #3845, fixes #3970, and fixes #4523.

PR #3927, opened 2020-11-24
Captured types in enhanced for loops

smillst:
When Dataflow desugars enhanced for loops, it creates a variable declaration for the iterator type using the type of `expression.iterator()`, where `expression` is the iterator expression in the enhanced for loop.   If the type of `expression.iterator()` contains a captured type, the type tree of the variable declaration will be missing annotations.  I couldn't find a way to add those annotations to the type tree. 

Instead, the tree for `expression` is saved in the method innovation for `iter#num0.next()`, where `iter#num0` is the variable declaration created from the type of `expression.iterator()`.  Then the framework uses this tree to get the correct type of  `iter#num0.next()`.  

(This should be merged before #3922 )

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -35,7 +35,7 @@ Removed a formal parameter from methods in `BaseTypeVisitor`:
 In Analysis.runAnalysisFor(), changed boolean parameter to enum BeforeOrAfter.
 
 Removed org.checkerframework.framework.util.AnnotatedTypes#getIteratedType; use
-AnnotatedTypeFactory#getIteratorElementType(ExpressionTree) instead.
+AnnotatedTypeFactory#getIterableElementType(ExpressionTree) instead.
 
 ---------------------------------------------------------------------------
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -32,6 +32,8 @@ Removed a formal parameter from methods in `BaseTypeVisitor`:
  * `checkPostcondition`
  * `checkConditionalPostcondition`
 
+In Analysis.runAnalysisFor(), changed boolean parameter to enum BeforeOrAfter.
+
 ---------------------------------------------------------------------------
 
 Version 3.7.1, November 2, 2020
--- a/changelog.txt
+++ b/changelog.txt
@@ -35,7 +35,7 @@ Removed a formal parameter from methods in `BaseTypeVisitor`:
 In Analysis.runAnalysisFor(), changed boolean parameter to enum BeforeOrAfter.
 
 Removed org.checkerframework.framework.util.AnnotatedTypes#getIteratedType; use
-AnnotatedTypeFactory#getIteratedType(ExpressionTree) instead.
+AnnotatedTypeFactory#getIteratorElementType(ExpressionTree) instead.
 
 ---------------------------------------------------------------------------
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -34,6 +34,9 @@ Removed a formal parameter from methods in `BaseTypeVisitor`:
 
 In Analysis.runAnalysisFor(), changed boolean parameter to enum BeforeOrAfter.
 
+Removed org.checkerframework.framework.util.AnnotatedTypes#getIteratedType; use
+AnnotatedTypeFactory#getIteratedType(ExpressionTree) instead.
+
 ---------------------------------------------------------------------------
 
 Version 3.7.1, November 2, 2020
--- a/dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java
@@ -2435,8 +2435,7 @@ public class CFGTranslationPhaseOne extends TreePathScanner<Node, Void> {
 
             AssignmentNode assignNode =
                     translateAssignment(variable, new LocalVariableNode(variable), nextCall);
-
-            // Translate assignment scans variable and creates new nodes, so set the expression
+            // translateAssignment() scans variable and creates new nodes, so set the expression
             // there, too.
             ((MethodInvocationNode) assignNode.getExpression()).setIteratorExpression(expression);
 
--- a/dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java
@@ -2426,10 +2426,16 @@ public class CFGTranslationPhaseOne extends TreePathScanner<Node, Void> {
             MethodInvocationNode nextCallNode =
                     new MethodInvocationNode(
                             nextCall, nextAccessNode, Collections.emptyList(), getCurrentPath());
+            // If the type of iteratorVariable is a capture, its type tree may be missing
+            // annotations, so save the expression in the node so that the full type can be found
+            // later.
+            nextCallNode.setIterExpression(expression);
             nextCallNode.setInSource(false);
             extendWithNode(nextCallNode);
 
-            translateAssignment(variable, new LocalVariableNode(variable), nextCall);
+            AssignmentNode assignNode =
+                    translateAssignment(variable, new LocalVariableNode(variable), nextCall);
+            ((MethodInvocationNode) assignNode.getExpression()).setIterExpression(expression);
 
             assert statement != null;
             scan(statement, p);
--- a/dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java
@@ -2429,13 +2429,16 @@ public class CFGTranslationPhaseOne extends TreePathScanner<Node, Void> {
             // If the type of iteratorVariable is a capture, its type tree may be missing
             // annotations, so save the expression in the node so that the full type can be found
             // later.
-            nextCallNode.setIterExpression(expression);
+            nextCallNode.setIteratorExpression(expression);
             nextCallNode.setInSource(false);
             extendWithNode(nextCallNode);
 
             AssignmentNode assignNode =
                     translateAssignment(variable, new LocalVariableNode(variable), nextCall);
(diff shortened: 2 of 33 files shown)


dataflow/src/main/java/org/checkerframework/dataflow/cfg/node/MethodInvocationNode.java
```diff
@@ -76,6 +96,28 @@ public TreePath getTreePath() {
...
 
+    /**
+     * If this MethodInvocationNode is a node for an {@link Iterator#next()} desugared from an
+     * enhanced for loop, then return the expression in the for loop, e.g., {@code iter} in {@code
+     * for(Object o: iter}. Otherwise, return null.
+     *
+     * @return the iter expression, or null if this is not a {@link Iterator#next()} from an
+     *     enhanced for loop
+     */
+    public @Nullable ExpressionTree getIterExpression() {
```
wmdietl, line 107: I don't see an invocation of this method and no other read access of the field. I must be missing something.

wmdietl, line 107: I would expand the `Iter` part to make clear it stands for `Iterator`.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java
```diff
@@ -2426,10 +2426,16 @@ public Node visitEnhancedForLoop(EnhancedForLoopTree tree, Void p) {
...
+            // annotations, so save the expression in the node so that the full type can be found
+            // later.
+            nextCallNode.setIterExpression(expression);
             nextCallNode.setInSource(false);
             extendWithNode(nextCallNode);
 
-            translateAssignment(variable, new LocalVariableNode(variable), nextCall);
+            AssignmentNode assignNode =
+                    translateAssignment(variable, new LocalVariableNode(variable), nextCall);
+            ((MethodInvocationNode) assignNode.getExpression()).setIterExpression(expression);
```
wmdietl, line 2438: Can you add a comment why this second call is necessary? I would have expected the expression to be equal to `nextCallNode`, on which `setIterExpression` was already called.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/node/MethodInvocationNode.java
smillst: I failed to copy those files changes to this PR.  I've done so now, thanks for pointing this out!

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1393,10 +1393,7 @@ public Void visitAssignment(AssignmentTree node, Void p) {
     @Override
     public Void visitEnhancedForLoop(EnhancedForLoopTree node, Void p) {
         AnnotatedTypeMirror var = atypeFactory.getAnnotatedTypeLhs(node.getVariable());
-        AnnotatedTypeMirror iterableType = atypeFactory.getAnnotatedType(node.getExpression());
-        AnnotatedTypeMirror iteratedType =
-                AnnotatedTypes.getIteratedType(
```
mernst, line 1398: AnnotatedTypes.getIteratedType is not longer called.  The Checker Framework still compiles if I delete it.  I think it should be removed.

mernst (changes requested):

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -2061,6 +2061,75 @@ protected void adaptGetClassReturnTypeToReceiver(
...
+    /**
+     * Return the element type of the given iterator. This is usually the type of {@code
+     * expression.itertor().next()}. If {@code expression} is an array, it is the component type of
+     * the array.
+     *
+     * @param expression an expression whose type is an array or implements {@link Iterable}
+     * @return the type of {@code expression.itertor().next()} or if {@code expression} is an array,
+     *     the component type of the array.
+     */
+    public AnnotatedTypeMirror getIteratedType(ExpressionTree expression) {
```
mernst, line 2073: I realize the terminology existed before this pull request, but I find "iterated type" confusing because it's new, nonstandard jargon.  Consider renaming the methods, such as to `getIteratorElementType`.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -2061,6 +2061,75 @@ protected void adaptGetClassReturnTypeToReceiver(
...
+                return ((AnnotatedArrayType) iterableType).getComponentType();
+            case WILDCARD:
+                return getIteratedType(
+                        expression,
+                        ((AnnotatedWildcardType) iterableType).getExtendsBound().deepCopy());
+            case TYPEVAR:
+                return getIteratedType(
+                        expression, ((AnnotatedTypeVariable) iterableType).getUpperBound());
+            case DECLARED:
+                TypeElement iterableElement =
```
mernst, line 2100: It pains me to see this re-computed on every call.  I would save it in a field.  Can you do that with `iterableElmType` too?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -2061,6 +2061,75 @@ protected void adaptGetClassReturnTypeToReceiver(
...
+                AnnotatedDeclaredType dt =
+                        AnnotatedTypes.asSuper(this, iterableType, iterableElmType);
+                if (dt.getTypeArguments().isEmpty()) {
+                    TypeElement e = ElementUtils.getTypeElement(processingEnv, Object.class);
+                    return getAnnotatedType(e);
+                } else {
+                    return dt.getTypeArguments().get(0);
+                }
+
+                // TODO: Properly desugar Iterator.next(). The below doesn't work because
```
mernst, line 2112: Please add a test case and reference that test case from this comment.  Or, if the comment is obviated by other changes such as the addition of the `MethodInvocationNode.iteratorExpression` field, delete the comment.

mernst (changes requested):

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
smillst: This code is only need if an annotated JDK has annotations on Iterator#next. None of our current checkers have such an annotation, so I can't added a test case for any of the current checkers.  Do you want me to add a test checker that has such an annotation?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -2061,6 +2061,75 @@ protected void adaptGetClassReturnTypeToReceiver(
         classWildcardArg.setExtendsBound(newBound);
     }
 
+    /**
+     * Return the element type of the given iterator. This is usually the type of {@code
```
smillst, line 2065: @mernst I find the use of `iterator` here confusing.  If not an array, the expression must implement `Iterable` not `Iterator`.  I think you mean `iterator` as in an expression that can appear in a for each loop, but it's not clear.

mernst: You are right:  that should have been iterable, not iterator.  Can you change that and make any other documentation improvements that you find needful?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
mernst: Thanks for the clarification.  I think it's enough to add a comment explaining what you said here:

> This code is only need if an annotated JDK has annotations on Iterator#next.

mernst (approved):

changelog.txt
```diff
@@ -34,6 +34,9 @@ Removed a formal parameter from methods in `BaseTypeVisitor`:
 
 In Analysis.runAnalysisFor(), changed boolean parameter to enum BeforeOrAfter.
 
+Removed org.checkerframework.framework.util.AnnotatedTypes#getIteratedType; use
+AnnotatedTypeFactory#getIteratedType(ExpressionTree) instead.
```
wmdietl, line 38: Should this be `#getIteratorElementType...`?

smillst: Yes, thanks!

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -2061,6 +2069,75 @@ protected void adaptGetClassReturnTypeToReceiver(
...
+    /**
+     * Return the element type of {@code expression}. This is usually the type of {@code
+     * expression.itertor().next()}. If {@code expression} is an array, it is the component type of
+     * the array.
+     *
+     * @param expression an expression whose type is an array or implements {@link Iterable}
+     * @return the type of {@code expression.itertor().next()} or if {@code expression} is an array,
+     *     the component type of the array.
+     */
+    public AnnotatedTypeMirror getIteratorElementType(ExpressionTree expression) {
```
mernst, line 2081: Should this be `getIterableElementType`?  That indicates the type of the argument more precisely.  I'm OK with the current name, too.

smillst: Yes, I made that change.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/node/MethodInvocationNode.java
```diff
@@ -76,6 +96,28 @@ public TreePath getTreePath() {
...
+     *
+     * @return the iter expression, or null if this is not a {@link Iterator#next()} from an
+     *     enhanced for loop
+     */
+    public @Nullable ExpressionTree getIteratorExpression() {
+        return iteratorExpression;
+    }
+
+    /**
+     * Set the iter expression.
```
mernst, line 112: Can you make "iter expression" more descriptive?  Even changing it to a term that is used elsewhere in the code would be better, I think.

mernst (changes requested):

mernst (approved):

PR #3933, opened 2020-11-25
Add new annotation file format ajava files

jwaataja:
Adds support for the ajava file format, as described in [this](https://docs.google.com/document/d/1ZMIIEVr6C0_SuXw1BpY7yFCw_cChVYkCc0oJh12DhCA/edit?usp=sharing) document.

In short, this file format extends the capabilities of stub files by allowing annotations in anonymous classes and local classes to be read, but the ASTs of the ajava file and any source file they apply to must match up to annotations. In the future, this format could be extended to read annotations in more locations stub files wouldn't be able to support.

There's a new implementation of `WholeProgramInference` called `WholeProgramInferenceJavaParser` that behaves identically to `WholeProgramInferenceScenes` but uses ajava files and supports anonymous inner classes.

The main component is `JointJavacJavaParserVisitor`. This traverses a javac tree and JavaParser AST simultaneously (that represent the same Java file). This allows parsing ajava files with JavaParser and then relating its nodes to the javac tree of the same file.


mernst:
@jwaataja The tests are failing.  Could you please fix the problems?  Let me know if you need help with that.

jwaataja:
@mernst  They're fixed now.

Issue #1970, opened 2018-05-01
Add @Contravariant as a complement to @Covariant


mernst:
The Checker Framework supports a `@Covariant` annotation.  It should support a `@Contravariant` annotation as well.

smillst:
java.util.Comparator is an example of a class whose type parameter should be `@Contravariant`.

PR #3936, opened 2020-11-27
Make the visualization of stores deterministic

mernst:


javacutil/src/main/java/org/checkerframework/javacutil/ToStringComparator.java
```diff
@@ -0,0 +1,37 @@
+// TODO: Put this in plume-util and use it from there.
+package org.checkerframework.javacutil;
+
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Objects;
+
+/** A comparator that orders values based on the lexicographic ordering of their toString(). */
+public class ToStringComparator implements Comparator<Object> {
```
smillst, line 10: Should this be `Comparator<@Nullable Object>`?

smillst (approved):

javacutil/src/main/java/org/checkerframework/javacutil/ToStringComparator.java
mernst: It implements both `Comparator<Object>` and `Comparator<@Nullable Object>`.  I think I can only name one of those, and the two are incompatible.  Should I create a parameterized `ToStringComparator` instead?

smillst: The type argument of `Comparator` should be contravariant. That way we could make this `Comparator<@Nullable Object>` which would be a subtype of `Comparator<@NonNull Object>` (We don't have the functionality; #1970 is a request to implement it.)

So for now, I think it's enough to mention that ToStringComparator, handles `null` elements in that class Javadoc.

As an aside, the Nullness Checker line 34 should have issued a warning on line 34, I'll look into why it didn't.

mernst: @smillst Let me know if the new comment looks good.

smillst: Looks good thanks!

PR #3939, opened 2020-11-27
Remove a redundant method

mernst:


checker/src/main/java/org/checkerframework/checker/initialization/InitializationStore.java
```diff
@@ -236,10 +236,11 @@ protected String internalVisualize(CFGVisualizer<V, S, ?> viz) {
         }
     }
 
-    public Map<FieldAccess, V> getFieldValues() {
```
wmdietl, line 239: I didn't see a method that this is overiding.
However, this should better go to Store, where the field is declared.
Maybe the `getAnalysis` method below can be removed then, because it is only used to get the field values. (If the field values in the Analysis and Store are the same...)

mernst: Since it is neither used nor documented, I am in favor of removing it.  We can reintroduce it in the future if it is needed.

smillst (approved):

PR #3947, opened 2020-11-30
Define method getOverriddenMethods

mernst:


javacutil/src/main/java/org/checkerframework/javacutil/TypesUtils.java
```diff
@@ -879,4 +880,19 @@ public static TypeMirror substitute(
                 com.sun.tools.javac.util.List.from(newP),
                 com.sun.tools.javac.util.List.from(newT));
     }
+
+    // This logically belongs in ElementUtils because it maps from ExecutableElement to
```
smillst, line 884: I would put it in ElementUtils because that's where Checker Framework developers will look for it, but I don't feel strongly.

smillst (approved):

PR #3961, opened 2020-12-03
Rename StubParser to AnnotationFileParser

mernst:


jwaataja (changes requested):
Looks good, two quick things.

PR #3969, opened 2020-12-04
Fix CFG builder loosing catch blocks in try-with-resource

artempyanykh:
Summary:
This fix is for Phase 1 of the translation.
Consider the following block:

```
try (ZipFile f = openZipFile()) {
    return true;
} catch (IOException e) {
    // This node is missing in CFG
    log(TAG, "This `log` node is missing in CFG of the method");
}
```

where

```
private static ZipFile openZipFile() throws IOException {...}
```

This gets translated into the following CFG.
Note absence of `log(TAG, ...)`!
```
2 -> 3 EACH_TO_EACH
3 -> 4 EACH_TO_EACH
4 -> 5 EACH_TO_EACH
4 -> 1 java.io.IOException
4 -> 1 Throwable
5 -> 0 EACH_TO_EACH

2:
Process order: 1
<entry>

3:
Process order: 2
marker (start of try statement #0)   [ Marker ]
f   [ VariableDeclaration ]
Test   [ ClassName ]
Test.openZipFile   [ MethodAccess ]

4:
Process order: 3
Test.openZipFile()   [ MethodInvocation ]

5:
Process order: 4
f = Test.openZipFile()   [ Assignment ]
marker (start of try block #0)   [ Marker ]
true   [ BooleanLiteral ]
return true   [ Return ]

1:
Process order: 6
<exceptional-exit>

0:
Process order: 5
<exit>
```

The problem seems to be caused by too early translation of resources:
resources of a try block get translated **before** the new frame is pushed
onto the `tryStack`, therefore checked exceptions thrown by `openZipFile`
use toplevel `tryStack` and get attached to exceptional exit block,
instead of user-supplied `catch` block.

With this patch CFG now correctly translates try-with-resource block.

Test plan:
Added a test case for Nullness to check that flow-sensitive
type-refinements apply to code inside catch block and after try-catch.

### DIFF
--- a/checker/tests/nullness/TryWithResources.java
+++ b/checker/tests/nullness/TryWithResources.java
@@ -1,4 +1,6 @@
 import java.io.*;
+import java.util.zip.ZipFile;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 class TryWithResources {
     void m1(InputStream stream) {
@@ -14,5 +16,30 @@ class TryWithResources {
             in.toString();
         } catch (Exception e) {
         }
+    }
+
+    // Checking that catch blocks and code after try-catch is part of CFG (and flow-sensitive
+    // type-refinements work there)
+    boolean m3(@Nullable Object x) {
+        try (ZipFile f = openZipFile()) {
+            return true;
+        } catch (IOException e) {
+            if (x != null) {
+                // OK
+                x.toString();
+            }
+        }
+
+        if (x != null) {
+            // OK
+            return x.equals(x);
+        }
+
+        return false;
+    }
+
+    // Helper
+    private static ZipFile openZipFile() throws IOException {
+        throw new IOException("No zip-file for you!");
     }
 }
--- a/checker/tests/nullness/TryWithResources.java
+++ b/checker/tests/nullness/TryWithResources.java
@@ -18,8 +18,8 @@ class TryWithResources {
         }
     }
 
-    // Checking that catch blocks and code after try-catch is part of CFG (and flow-sensitive
-    // type-refinements work there)
+    // Check that catch blocks and code after try-catch are part of CFG (and flow-sensitive
+    // type-refinements work there).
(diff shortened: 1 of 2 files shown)


wmdietl:
Thanks for the fix!

The problem also causes missing type refinements, which causes false-positive errors:

```java
import java.io.IOException;
import java.util.zip.ZipFile;

class TryWithResources {
    public static boolean test(@Nullable Object x) {
        // try-with-resource is important!
        try (ZipFile f = openZipFile()) {
            return true;
        } catch (IOException e) {
            // This block is missing in CFG
            if (x != null) {
                // Unwanted error.
                x.toString();
            }
        }
        // This block is also missing in CFG
        if (x != null) {
            // Unwanted error.
            return x.equals(x);
        }
        return false;
    }

    private static ZipFile openZipFile() throws IOException {
        throw new IOException("No zip-file for you!");
    }
}
```

It might be easier to add this as test case instead of adding the CFG test.

artempyanykh:
Hi @wmdietl, thanks for checking!

> It might be easier to add this as test case instead of adding the CFG test.

Right, I noticed that `dataflow` package doesn't have a lot going on around CFG tests in particular, but wasn't sure whether adding an *expect* test specifically for CFG or an indirect test inside CF is better.

I can certainly remove the expect test and add a test case inside Nullness. Would this be all right?

artempyanykh:
@wmdietl I amended the commit w.r.t. test. Let me know if this looks good now.

wmdietl (approved):
Thanks! This change looks good to me.
I agree that we could add more direct CFG generation tests but should first improve the testing infrastructure for that. Adding this nullness test seems cleaner for now.

mernst:
@artempyanykh Thanks very much for this fix.  We appreciate it.

Sorry for the delay in merging -- I thought that @wmdietl was handling it.

artempyanykh:
No worries, thanks @mernst !

PR #3972, opened 2020-12-08
allow users to run the Called Methods Checker without the Returns Receiver Checker, if desired

### LINKED ISSUE #3971: Add flag to disable returns receiver for CalledMethodsChecker
Follow-up to https://github.com/typetools/checker-framework/pull/3743

We should add a command-line flag to the Called Methods Checker to turn off running the Returns Receiver Checker as a sub-checker.  Some use cases do not involve checking fluent APIs, and in those cases disabling the Returns Receiver Checker will make the Called Methods Checker run much faster.

kelloggm:
Fixes #3971

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsChecker.java
+++ b/checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsChecker.java
@@ -1,6 +1,7 @@
 package org.checkerframework.checker.calledmethods;
 
 import java.util.LinkedHashSet;
+import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
 import org.checkerframework.common.accumulation.AccumulationChecker;
 import org.checkerframework.common.basetype.BaseTypeChecker;
 import org.checkerframework.common.returnsreceiver.ReturnsReceiverChecker;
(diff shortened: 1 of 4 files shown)


checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsChecker.java
```diff
@@ -50,20 +51,55 @@
...
+    private @MonotonicNonNull Boolean rrDisabled = null;
+
+    /**
+     * Was the returns receiver checker disabled on the command line?
+     *
+     * @return whether the -AdisableReturnsReceiver option was specified on the command line.
+     */
+    private boolean isRRDisabled() {
+        if (rrDisabled == null) {
+            // BaseTypeChecker#hasOption calls this getImmediateSubcheckerClasses (so that all
```
msridhar, line 77:
```suggestion
            // BaseTypeChecker#hasOption calls getImmediateSubcheckerClasses (so that all
```

msridhar (approved):
A bit more tricky than one might have, but I see why it's needed.  Just a minor nit

mernst (approved):

PR #3976, opened 2020-12-09
Rename Receiver to JavaExpression

mernst:
Delete file SKIP-REQUIRE-JAVADOC before merging

### DIFF
--- a/SKIP-REQUIRE-JAVADOC
+++ b/SKIP-REQUIRE-JAVADOC
@@ -1,1 +1,1 @@
-Delete this file before merging the pull request
+Delete this file before merging the pull request.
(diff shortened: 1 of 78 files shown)


smillst (approved):
After this is merged (or in this pull request), the methods in org.checkerframework.dataflow.expression.JavaExpressions should be moved to `JavaExpressionParseUtil`.

mernst:
> the methods in org.checkerframework.dataflow.expression.JavaExpressions should be moved to `JavaExpressionParseUtil`.

JavaExpressionParseUtil is about parsing strings rather than converting one data structure to another, so the methods seem more appropriate to stay where they are.  Another reason to keep them in the `dataflow` project is that they don't depend on anything in the `framework` project.

What do you think?

PR #3977, opened 2020-12-10
Add new annotation file format ajava files

jwaataja:
New pull request since CI stopped on the old one.

Adds support for the ajava file format, as described in [this](https://docs.google.com/document/d/1ZMIIEVr6C0_SuXw1BpY7yFCw_cChVYkCc0oJh12DhCA/edit?usp=sharing) document.

In short, this file format extends the capabilities of stub files by allowing annotations in anonymous classes and local classes to be read, but the ASTs of the ajava file and any source file they apply to must match up to annotations. In the future, this format could be extended to read annotations in more locations stub files wouldn't be able to support.

There's a new implementation of `WholeProgramInference` called `WholeProgramInferenceJavaParser` that behaves identically to `WholeProgramInferenceScenes` but uses ajava files and supports anonymous inner classes.

The main component is `JointJavacJavaParserVisitor`. This traverses a javac tree and JavaParser AST simultaneously (that represent the same Java file). This allows parsing ajava files with JavaParser and then relating its nodes to the javac tree of the same file.


mernst:
@jwaataja Could you please pull master into your branch?  A [recent commit](https://github.com/typetools/checker-framework/commit/b6acf04747f2e0309cf9a8432793a2fbf83a4b5c) moved the WPI tests from the `checker` project to the `framework` project.

framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileParser.java
```diff
@@ -623,34 +705,46 @@ private void processPackage(PackageDeclaration packDecl) {
      * @param typeDecl the type declaration to process
      * @param outertypeName the name of the containing class, when processing a nested class;
      *     otherwise null
+     * @param classTree the tree corresponding to typeDecl if processing an ajava file, null
+     *     otherwise
+     * @return a list of types variables for {@code typeDecl}. These should be removed after
```
mernst, line 710: Removed from where?

Also, many clients ignore the return value.  Is that OK?  Is ignoring the return value OK exactly when the last argument is null?  Or maybe when isParsingStubFile is false:?

Should a similar note be added to `processCallableDeclaration`?

mernst (changes requested):
Could you please address the "TODO JWAATAJA" comments in WholeProgramInferenceJavaParser.java?

There are new prepare*ForWriting methods in WholeProgramInferenceScenes.
Could you please implement similar ones in WholeProgramInferenceJavaParser?

There is also an override of prepareMethodForWriting in FormatterAnnotatedTypeFactory.java.
That file will need to define a nested class FormatterWholeProgramInferenceJavaParser.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -682,6 +763,37 @@ public void setRoot(@Nullable CompilationUnitTree root) {
             // contents won't change between compilation units.
             // elementCache.clear();
         }
+
+        if (root != null && checker.hasOption("ajava")) {
+            // TODO: This relies on the toString implementations of MemberSelectTree and
+            // IdentifierTree. This could be turned into its own method that does this correctly.
```
mernst, line 769: What is "this" in "does this correctly"?  Can you add a brief description of what this code does?

framework/src/main/java/org/checkerframework/framework/ajava/AjavaUtils.java
```diff
@@ -0,0 +1,50 @@
...
+
+/** A utility class for working with javac and JavaParser representations of a source file. */
+public class AjavaUtils {
+    /**
+     * Returns the path to the source file containing {@code element}
+     *
+     * @param element the type element to look at
+     * @return path to the source file containing {@code element}
+     */
+    public static String getSourceFilePath(TypeElement element) {
```
mernst, line 20: This method does not seem specific to JavaParser.  I would put it in ElementUtils.

framework/src/main/java/org/checkerframework/framework/ajava/AjavaUtils.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Given the compilation unit node for a source file, returns the top level type definition with
+     * the given name.
+     *
+     * @param root compilation unit to search
+     * @param name name of a top level type declaration in {@code root}
+     * @return a top level type declaration in {@code root} named {@code name}
+     */
+    public static TypeDeclaration<?> getTypeDeclarationByName(CompilationUnit root, String name) {
```
mernst, line 32: For any method or class that is an operation just over JavaParser objects, I suggest putting them in a new `JavaParserUtil` class.
That would include this methad and also `ClearAnnotationVisitor` and `StringLiteralConcatenateVisitor`.
That would not include `AnnotationConversion` and `AnnotationTransferVisitor`, because they walk two data structures in tandem.

framework/src/main/java/org/checkerframework/framework/ajava/ExpectedTreesVisitor.java
```diff
@@ -0,0 +1,421 @@
...
+import org.checkerframework.javacutil.BugInCF;
+
+/**
+ * Visitor that stores each visited tree that should match with some JavaParser node if the same
+ * Java file was parsed with both. The primary purpose is to test the {@link
+ * JointJavacJavaParserVisitor} class when the -AcheckJavaParserVisitor flag is used. That classes
+ * traverses a javac tree and JavaParser AST simultaneously, so the trees this class stores can be
+ * used to test if the entirety of the javac tree was visited.
+ *
+ * <p>Adds these trees to a set during traversal. Many trees shouldn't be matched with a JavaParser
```
mernst, line 46: You don't need to list them here, but can you say how to find, in the source code below, all the trees that shouldn't be matched with a JavaParser node?  Is it just a search for the string "javac"?

framework/src/main/java/org/checkerframework/framework/ajava/StringLiteralConcatenateVisitor.java
```diff
@@ -0,0 +1,38 @@
...
+ * the two string literals are not added directly. For example, in the expression {@code 1 + "a" +
+ * "b"} we might imagine that {@code 1 + "a"} is evaluated and then {@code "b"} is added.
+ * Regardless, this is transformed into {@code 1 + "ab"}.
+ */
+public class StringLiteralConcatenateVisitor extends VoidVisitorAdapter<Void> {
+    @Override
+    public void visit(BinaryExpr node, Void p) {
+        super.visit(node, p);
+        if (node.getOperator() == BinaryExpr.Operator.PLUS
+                && node.getRight().isStringLiteralExpr()) {
```
mernst, line 19: By experimentation, I verified that javac returns the literal string "ab" for this code:
```java
    return ("a") + ("b");
```

Please make this implementation handle parenthesization, and add a test.

mernst (changes requested):

framework/src/main/java/org/checkerframework/framework/ajava/StringLiteralConcatenateVisitor.java
mernst: I observed that optimization in the generated bytecode.  But @jwaataja does not observe that in the AST.  So nothing needs to be done here (except maybe add a comment explaining why this code does not look within parentheses).

framework/src/main/java/org/checkerframework/framework/ajava/ExpectedTreesVisitor.java
jwaataja: I don't quite understand what you mean. The class as a whole contains the instructions for how to find trees that shouldn't be matched with a JavaParser node. These exceptions occur at each call to `trees.remove` and when some visit method is overriden and a child is not visited. In that case the entirety of the child's subtree tree shouldn't be matched to JavaParser nodes. As far as I'm aware, all such exceptions are accompanied by a comment.

framework/src/main/java/org/checkerframework/framework/ajava/ExpectedTreesVisitor.java
```diff
@@ -0,0 +1,421 @@
...
+            member.accept(this, p);
+        }
+
+        return null;
+    }
+
+    @Override
+    public Void visitLambdaExpression(LambdaExpressionTree tree, Void p) {
+        for (VariableTree parameter : tree.getParameters()) {
+            // Parameter types might not be specified for lambdas. When not specified, JavaParser
```
mernst, line 322: Make the actor clear in the first sentence:  it's the programmer.
Also, what does javac do in this case?  The second sentence mentions JavaParser.

framework/src/main/java/org/checkerframework/framework/ajava/ExpectedTreesVisitor.java
```diff
@@ -0,0 +1,421 @@
...
+            hasVarType = false;
+        }
+
+        @Override
+        public void visit(ClassOrInterfaceType node, Void p) {
+            if (node.getName().asString().equals("var")) {
+                hasVarType = true;
+            }
+
+            super.visit(node, p);
```
mernst, line 393: Once `hasVarType` is true, there is no need to do any more visiting.

(For efficiency, it would even be possible to shortcut, such as by throwing an exception that could be caught by the static `hasVarType` method that I added, and then you also wouldn't need the variable.  You could skip this optimization if you find it gross.)

framework/src/main/java/org/checkerframework/framework/ajava/JointJavacJavaParserVisitor.java
```diff
@@ -0,0 +1,2382 @@
...
+ * the visit methods in {@code com.sun.source.tree.TreeVisitor}, but for each javac tree type there
+ * may be multiple process methods for each possible node type it could be matched to.
+ *
+ * <p>The {@code process} methods are called in pre-order. That is, process methods for a parent are
+ * called before its children.
+ */
+public abstract class JointJavacJavaParserVisitor implements TreeVisitor<Void, Node> {
+    @Override
+    public Void visitAnnotation(AnnotationTree javacTree, Node javaParserNode) {
+        // javac stores annotation arguments as assignments, so @MyAnno("myArg") might be stored the
```
mernst, line 172: Should "might be" be "is"?  If not, under what circumstance is the statement about storage true?

framework/src/main/java/org/checkerframework/framework/ajava/JointJavacJavaParserVisitor.java
```diff
@@ -0,0 +1,2382 @@
...
+        // same as @MyAnno(value="myArg") which has a single element argument list with an
+        // assignment.
+        if (javaParserNode instanceof MarkerAnnotationExpr) {
+            processAnnotation(javacTree, (MarkerAnnotationExpr) javaParserNode);
+        } else if (javaParserNode instanceof SingleMemberAnnotationExpr) {
+            SingleMemberAnnotationExpr node = (SingleMemberAnnotationExpr) javaParserNode;
+            processAnnotation(javacTree, node);
+            assert javacTree.getArguments().size() == 1;
```
mernst, line 180: I suggest that you also assert that the element name is "value".

framework/src/main/java/org/checkerframework/framework/ajava/JointJavacJavaParserVisitor.java
```diff
@@ -0,0 +1,2382 @@
...
+     *
+     * @param javacStatements sequence of javac trees for statements
+     * @param javaParserStatements sequence of JavaParser statements representing the same block as
+     *     {@code javacStatements}
+     */
+    private void processStatements(
+            Iterable<? extends StatementTree> javacStatements,
+            Iterable<Statement> javaParserStatements) {
+        Iterator<? extends StatementTree> javacIter = javacStatements.iterator();
+        boolean hasNextJavac = javacIter.hasNext();
```
mernst, line 311: Here and elsewhere, you could drop the boolean and just test whether `javacStatement` is null.  That would be a bit shorter.  But you can leave this code as is if you prefer.

framework/src/main/java/org/checkerframework/framework/ajava/JointJavacJavaParserVisitor.java
```diff
@@ -0,0 +1,2382 @@
...
+        javacTree.getBlock().accept(this, node.getBody());
+        return null;
+    }
+
+    @Override
+    public Void visitClass(ClassTree javacTree, Node javaParserNode) {
+        if (javaParserNode instanceof ClassOrInterfaceDeclaration) {
+            ClassOrInterfaceDeclaration node = (ClassOrInterfaceDeclaration) javaParserNode;
+            processClass(javacTree, node);
+            if (javacTree.getTypeParameters().size() == node.getTypeParameters().size()) {
```
mernst, line 469: Should it be an error if these are not equal?

framework/src/main/java/org/checkerframework/framework/ajava/JointJavacJavaParserVisitor.java
```diff
@@ -0,0 +1,2382 @@
...
+     * <p>This method skips these synthetic constructors.
+     *
+     * @param javacBody body of an anonymous class body
+     * @param javaParserMembers list of members for the anonymous class body of an {@code
+     *     ObjectCreationExpr}
+     */
+    public void visitAnonymouClassBody(
+            ClassTree javacBody, List<BodyDeclaration<?>> javaParserMembers) {
+        List<Tree> members = new ArrayList<>(javacBody.getMembers());
+        while (!members.isEmpty()) {
```
mernst, line 611: Why is this a `while` loop rather than an `if` test?  I think there can be at most one initializer to remove.

framework/src/main/java/org/checkerframework/framework/ajava/JointJavacJavaParserVisitor.java
```diff
@@ -0,0 +1,2382 @@
...
+
+    @Override
+    public Void visitExpressionStatement(ExpressionStatementTree javacTree, Node javaParserNode) {
+        if (javaParserNode instanceof ExpressionStmt) {
+            ExpressionStmt node = (ExpressionStmt) javaParserNode;
+            processExpressionStatemen(javacTree, node);
+            javacTree.getExpression().accept(this, node.getExpression());
+        } else if (javaParserNode instanceof ExplicitConstructorInvocationStmt) {
+            // In this case the expression will be a MethodTree, which would be better to match with
+            // the statement than the expression statement itself.
```
mernst, line 779: I'm confused by this comment.  Should it be "the expression than the expression statement itself."?  In any event, please clarify.

framework/src/main/java/org/checkerframework/framework/ajava/JointJavacJavaParserVisitor.java
```diff
@@ -0,0 +1,2382 @@
...
+        if (javacTree.getCondition() != null) {
+            javacTree.getCondition().accept(this, node.getCompare().get());
+        }
+
+        // Javac stores a list of expression statements and JavaParser stores a list of statements,
+        // the javac statements must be unwrapped.
+        assert javacTree.getUpdate().size() == node.getUpdate().size();
+        Iterator<Expression> javaParserIter = node.getUpdate().iterator();
+        for (ExpressionStatementTree update : javacTree.getUpdate()) {
+            update.getExpression().accept(this, javaParserIter.next());
```
mernst, line 824: This seems to visit a javac expression and a JavaParser statement, which seems wrong to me because they are different kinds of AST nodes.  Perhaps clarifying the comment will resolve my confusion.

framework/src/main/java/org/checkerframework/framework/ajava/JointJavacJavaParserVisitor.java
```diff
@@ -0,0 +1,2382 @@
...
+        return null;
+    }
+
+    @Override
+    public Void visitLambdaExpression(LambdaExpressionTree javacTree, Node javaParserNode) {
+        if (!(javaParserNode instanceof LambdaExpr)) {
+            throwUnexpectedNodeType(javacTree, javaParserNode, LambdaExpr.class);
+        }
+
+        LambdaExpr node = (LambdaExpr) javaParserNode;
```
mernst, line 940: Minor:  I would find it easier to read the code if variables like `node` were prefixed by `javaParser` or `javac`.  In this case, a clearer name would be `javaParserLambdaExpr`.  (Do you prefer the generic `node` so all method bodies use the same variable name for the downcasted variable?)

framework/src/main/java/org/checkerframework/framework/ajava/JointJavacJavaParserVisitor.java
```diff
@@ -0,0 +1,2382 @@
...
+     */
+    private void visitMethodForMethodDeclaration(
+            MethodTree javacTree, MethodDeclaration javaParserNode) {
+        processMethod(javacTree, javaParserNode);
+        // TODO: Handle modifiers. In javac this is a ModifiersTree but in JavaParser it's a list of
+        // modifiers. This is a problem because a ModifiersTree has separate accessors to
+        // annotations and other modifiers, so the order doesn't match. It might be that for
+        // JavaParser, the annotations and other modifiers are also accessed separately.
+        javacTree.getReturnType().accept(this, javaParserNode.getType());
+        // Unlike other constructs, the list is non-null even if no type parameters are present.
```
mernst, line 1072: Does "the list" refer to a javac or JavaParser data structure?

framework/src/main/java/org/checkerframework/framework/ajava/JointJavacJavaParserVisitor.java
```diff
@@ -0,0 +1,2382 @@
...
+            javacTree.getDefaultValue().accept(this, javaParserNode.getDefaultValue().get());
+        }
+    }
+
+    @Override
+    public Void visitMethodInvocation(MethodInvocationTree javacTree, Node javaParserNode) {
+        if (javaParserNode instanceof MethodCallExpr) {
+            MethodCallExpr node = (MethodCallExpr) javaParserNode;
+            processMethodInvocation(javacTree, node);
+            // In javac, the type arguments will be empty even if no type arguments are specified,
```
mernst, line 1137: Can you remove "even" here?

mernst (changes requested):
Overall this looks great.  Here are a few comments and questions.

framework/src/main/java/org/checkerframework/framework/ajava/ExpectedTreesVisitor.java
jwaataja: This class is only used by the testing framework, and this would only be an optimization if the file actually uses `var`, which doesn't occur often in the tests as far as I know.

framework/src/main/java/org/checkerframework/framework/ajava/JointJavacJavaParserVisitor.java
jwaataja: Yes, I prefer that name because it can be used in every method, and the name `javaParserNode` is already used for the parameter type.

framework/src/main/java/org/checkerframework/framework/ajava/JointJavacJavaParserVisitor.java
```diff
@@ -0,0 +1,2386 @@
...
+        javacTree.getServiceName().accept(this, node.getName());
+        return null;
+    }
+
+    @Override
+    public Void visitVariable(VariableTree javacTree, Node javaParserNode) {
+        if (javaParserNode instanceof VariableDeclarator) {
+            VariableDeclarator node = (VariableDeclarator) javaParserNode;
+            processVariable(javacTree, node);
+            if (!node.getType().isVarType()
```
mernst, line 1512: Could you write a comment about what is happening here?

mernst (changes requested):
Here are comments on `JointJavacJavaParserVisitor.java`.

framework/src/main/java/org/checkerframework/framework/ajava/JointJavacJavaParserVisitor.java
```diff
@@ -0,0 +1,2386 @@
...
+            if (hasNextJavac
+                    && isDefaultSuperConstructorCall(javacStatement)
+                    && (!hasNextJavaParser
+                            || !isDefaultSuperConstructorCall(javaParserStatement))) {
+                hasNextJavac = javacIter.hasNext();
+                javacStatement = hasNextJavac ? javacIter.next() : null;
+                continue;
+            }
+
+            // In javac, a line like "int i = 0, j = 0" is expanded as two sibling VariableTree
```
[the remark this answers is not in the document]
jwaataja: The particular logic and methods called in each case are different, so abstracting out would probably complicate the code more than necessary. The code to advance the iterators was repetitive, so I've introduced a class `LookaheadIterator` to make it easier to read.

mernst (changes requested):
Oops, I hadn't posted these other comments.

framework/src/main/java/org/checkerframework/framework/ajava/ExpectedTreesVisitor.java
```diff
@@ -0,0 +1,419 @@
...
+     */
+    @Override
+    public void defaultAction(Tree tree) {
+        trees.add(tree);
+    }
+
+    @Override
+    public Void visitCompilationUnit(CompilationUnitTree tree, Void p) {
+        // When using the "var" keyword, javac replaces it with the correct type, meaning the
+        // inserted type has no corresponding javac tree. There is no way of knowing if the
```
mernst, line 75: Can you check whether it has a position?  For example:
```java
import com.sun.tools.javac.util.Position;
import com.sun.tools.javac.tree.JCTree.JCExpression;
public static boolean isVarDecl(VariableTree varTree) {
    return ((JCExpression)varTree.getType()).pos == Position.NOPOS;
}
```

mernst, line 75: Do you mean "no corresponding JavaParser tree"?

framework/src/main/java/org/checkerframework/framework/ajava/JointJavacJavaParserVisitor.java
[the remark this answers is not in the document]
mernst: Rather than defining your own `LookaheadIterator`, how about using Guava's `PeekingIterator` which I think provides the same functionality?

framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileParser.java
```diff
@@ -622,37 +696,55 @@ private void processPackage(PackageDeclaration packDecl) {
     /**
      * Process a type declaration: copy its annotations to {@code #annotationFileAnnos}.
      *
+     * <p>This method stores the declaration's type parameters in {@link #typeParameters}. When
+     * processing an ajava file, where traversal is handled externaly by a visitor, these type
```
mernst, line 702: Does "a visitor" specifically mean `AjavaParserVisitor`?  If possible, be specific.

framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileParser.java
```diff
@@ -2432,6 +2536,66 @@ private void stubDebug(String warning) {
         }
     }
 
+    /**
+     * Walks the javac tree for a file, processing the JavaParser node for each program element it
```
mernst, line 2542: From the documentation and the name of this class, I thought it was an abstract base class intended to be subclassed.  Can you be specific about what processing it does, and give it a better name?  This seems to do special processing for constructs that can declare type variables (classes, methods), but I'm not sure why it overrides `visitVariable`.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -233,6 +243,16 @@
     /** Parses stub files and stores annotations from stub files. */
     public final AnnotationFileElementTypes stubTypes;
```
mernst, line 244: Should the documentation of these two fields mention that they only store public elements?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -233,6 +243,16 @@
...
 
+    /** Parses ajava files and stores annotations from ajava files. */
+    public final AnnotationFileElementTypes ajavaTypes;
+
+    /**
+     * If processing a file, stores any annotations read from an ajava file for that class. Unlike
```
mernst, line 250: Should "processing a file" be "processing an ajava file"?  Or does this mean "if the Checker Framework is type-checking a .java file, stores annotations read from an ajava file for that class, if one exists"?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -480,13 +502,24 @@ public AnnotatedTypeFactory(BaseTypeChecker checker) {
...
+                    wpiOutputFormat = WholeProgramInference.OutputFormat.AJAVA;
+                    break;
                 default:
                     throw new UserError(
                             "Bad argument -Ainfer="
                                     + inferArg
                                     + " should be one of: -Ainfer=jaifs, -Ainfer=stubs");
             }
-            wholeProgramInference = new WholeProgramInferenceScenes(this);
+            if (checker.getOption("infer").equals("ajava")) {
```
mernst, line 514: Would it be cleaner to check the variable `wpiOutputFormat` here rather than calling `getOption` and `String.equals`?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -670,6 +703,58 @@ protected void initializeReflectionResolution() {
...
     public void setRoot(@Nullable CompilationUnitTree root) {
+        setRoot(root, /*shouldCheckVisitor=*/ true);
+    }
+
+    /**
+     * Set the CompilationUnitTree that should be used.
+     *
+     * @param root the new compilation unit to use
+     * @param shouldCheckVisitor if true, run the visitor that verifies that the javac tree can be
+     *     visited with its corresponding JavaParser AST. The check only occurs if
```
mernst, line 714: At first I thought this comment was explaining when `shouldCheckVisitor` is true.  But it is actually talking about a different boolean value.  What is the relationship between them, and why are both needed?  That is, why can't `shouldCheckeVisitor` be set according to `-AcheckJavaParserVisitor`?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -3562,6 +3703,34 @@ private AnnotationMirror getDeclAnnotation(
             }
         }
 
+        if (!ajavaTypes.isParsing()) {
+
+            // Retrieving annotations from ajava files.
```
mernst, line 3710: This code block appears identically 3 times.  Please abstract it out into a method.

mernst (changes requested):
Here are more comments.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
```diff
@@ -0,0 +1,1076 @@
...
+
+        // Only update the AnnotatedTypeMirror if there are no explicit annotations
+        if (curATM.getExplicitAnnotations().isEmpty() || !ignoreIfAnnotated) {
+            for (AnnotationMirror am : newATM.getAnnotations()) {
+                typeToUpdate.addAnnotation(am);
+            }
+        } else if (curATM.getKind() == TypeKind.TYPEVAR) {
+            // getExplicitAnnotations will be non-empty for type vars whose bounds are explicitly
+            // annotated.  So instead, only insert the annotation if there is not primary annotation
+            // of the same hierarchy.  #shouldIgnore prevent annotations that are subtypes of type
```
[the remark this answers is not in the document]
jwaataja: It's only relevant to the Scenes implementation, I've removed it here.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
```diff
@@ -0,0 +1,1076 @@
...
+    /// Reading in files
+    ///
+
+    @Override
+    public void preprocessClassTree(ClassTree classTree) {
+        addClassTree(classTree);
+    }
+
+    /**
+     * Reads in the source file containing {@code tree} and creates a wrapper around {@code tree}.
```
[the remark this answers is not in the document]
jwaataja: You're right in that it stores everything in `classToAnnos`.

> Please write the Javadoc in terms of the public abstractions.

This is a private method, so I think it's fine to mention the private fields here.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
```diff
@@ -0,0 +1,1076 @@
...
+                        // below call to TreeUtils.elementFromDeclaration causes a crash.
+                        if (TreeUtils.elementFromTree(javacTree) == null) {
+                            return;
+                        }
+
+                        VariableElement elt = TreeUtils.elementFromDeclaration(javacTree);
+                        if (!elt.getKind().isField()) {
+                            return;
+                        }
+
```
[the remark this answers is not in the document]
jwaataja: It does. There's a method `processClass` that takes an `EnumDeclaration` as its second argument. It's called `processClass` because javac stores the enum declaration as a `ClassTree`.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
```diff
@@ -0,0 +1,1076 @@
...
+     * Adds an explicit receiver type to a JavaParser method declaration.
+     *
+     * @param methodDeclaration declaration to add a receiver to
+     */
+    private static void addExplicitReceiver(MethodDeclaration methodDeclaration) {
+        if (methodDeclaration.getReceiverParameter().isPresent()) {
+            return;
+        }
+
+        if (!methodDeclaration.getParentNode().isPresent()) {
```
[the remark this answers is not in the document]
jwaataja: I'm not aware of any cases where the parent node wouldn't be present. I'll remove the check and see if it still works.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceStorage.java
```diff
@@ -0,0 +1,197 @@
...
+     * Indicates that inferred annotations for the file at {@code path} have changed since last
+     * written. This causes output files for {@code path} to be written out next time {@link
+     * #writeResultsToFile} is called.
+     *
+     * @param path path to the file with annotations that have been modified
+     */
+    public void setFileModified(String path);
+
+    /**
+     * Performs any preparation required for inference on Elements of a class. Should be called on
```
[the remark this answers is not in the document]
jwaataja: I think currently we only infer things for constructs with an `Element`, but that might change in the future so I'll remove the capitalization.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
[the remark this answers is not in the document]
jwaataja: We would only add enum constants if we did inference on them, which we're not doing for now.

jwaataja:
I've addressed everything. This should be ready for another review.

smillst (commented):
Here are comments on the manual.  I'm reviewing the code now.

smillst (commented):
Here's my review of the changes to the framework classes.  Please also see, https://github.com/jwaataja/checker-framework/pull/11.  I'm going to skim the reset of the changes next.

jwaataja:
@smillst I've merged your pull request and addressed the comments so far. Please see the new paragraph at the bottom of the `ajava-using` section of the manual in `annotating-libraries.tex` to make sure it looks correct.

docs/manual/annotating-libraries.tex
```diff
@@ -844,6 +844,69 @@
 in the version of the library you are using.
 
 
+\sectionAndLabel{Ajava Files}{ajava-files}
```
smillst, line 847: Should  https://checkerframework.org/manual/#whole-program-inference be updated to reference the new file format?

smillst (approved):
've skimmed the new files added and didn't see any needed changes.   The changes made in response to my other reviews look good, thanks! I

smillst:
@mernst should there be a change log entry for this?

docs/manual/annotating-libraries.tex
jwaataja: Currently, the `wpi.sh` and `wpi-many.sh` scripts only support stub files. It should be easy to update them to use ajava files, but that can be done separately.

mernst (approved):

mernst:
Thanks, Jason!

PR #3987, opened 2020-12-10
Add annotations that may be missing from inferred types.

smillst:


mernst (approved):
Is this a temporary or a permanent change?
Is there a test for this change?

PR #3992, opened 2020-12-11
Delete jaifs from WPI test before before testing WPI with stubs.

smillst:
WPI uses annotations in jaif files even if `-Ainfer=stubs` is supplied. (See javadoc of `WholeProgramInferenceScenesStorage`.)  So the jaifs have to be deleted before running the tests.

There's a failing test because two stub files are merged:
* `Generic-org.checkerframework.common.value.ValueChecker.astub`
* `Generic-org.checkerframework.checker.testchecker.wholeprograminference.WholeProgramInferenceTestChecker.astub`

The `WholeProgramInferenceTestChecker` stub has an annotation on a type variable, but the `ValueChecker` stub does not.  So the annotation is removed when the `ValueChecker` stub is read.  The correct fix here is to only clear the annotation if the stub file is annotated for the current checker.


mernst:
> There's a failing test

Does this mean that this pull request is not ready to be merged?  Should we open a different issue for that other bug, or is it your intention that it should be fixed in this pull request before it is merged?

smillst:
@mernst I added support for `@AnnotedFor` in stub files; see #3995.  But that isn't enough to fix this failing test.  WPI needs to output `@AnnotatedFor` in the stub files it produces.  So, I guess in the meantime, I should skip the test that fails in this pull request and merge it.

PR #3995, opened 2020-12-11
Add support for AnnotatedFor in stub files.

smillst:
They are still ignored in the annotated JDK because the purity annotations apply to every checker even if the code hasn't been annotated for a particular checker.


framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -4528,4 +4529,23 @@ public boolean containsSameByClass(
     protected WholeProgramInference createWholeProgramInference() {
         return new WholeProgramInferenceScenes(this);
     }
+
+    /**
+     * Does that {@link org.checkerframework.framework.qual.AnnotatedFor}, {@code anno} annotation
```
mernst, line 4534: I had to read this sentence a couple of times.  Can you remove "that `{@link ...}`", or change it to "Does an `{@link ...}` apply to this checker?"?

framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileParser.java
```diff
@@ -1745,6 +1750,33 @@ private PackageElement findPackage(String packageName) {
...
 
+    /**
+     * Returns true if one of the annotations is {@link AnnotatedFor} and its list of checkers does
+     * not contain this checker.
+     *
+     * @param annotations a list of JavaParser annotations
+     * @return true if one of the annotations is {@link AnnotatedFor} and its list of checkers does
+     *     not contain this checker
+     */
+    private boolean notAnnotatedForThisChecker(List<AnnotationExpr> annotations) {
```
mernst, line 1761: I find this name a bit confusing.  The name implies that the method returns true whenever there is no `@AnnotatedFor("ThisChecker")` annotation.  The name `doesAnnotatedForApplyToThisChecker` that this method uses is also clearer.  Could you choose a more descriptive name for this method?  Maybe `hasAnnotatedForWithoutThisChecker`, though you may be able to come up with a better name.

Also, please improve the Javadoc overview.   Add a first sentence that says this method returns true if the element is definitely not annotated for this checker, and returns false if it might or might not be annotated for this checker.  That explains the semantics of the operation and would have helped me understand the method.  (The current comment is useful, but it just recapitulates the definition.)  With this change, maybe the name change could be smaller, or maybe the current name could be retained.

framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileParser.java
```diff
@@ -1002,27 +1013,21 @@ private void processParameters(
...
-                                "in file %s at line %s removed existing annotations on type: %s",
-                                filename.substring(filename.lastIndexOf('/') + 1),
-                                typeDef.getBegin().get().line,
-                                atype.toString(true)));
-            }
-            */
-            // Clear existing annotations, which only makes a difference for
-            // type variables, but doesn't hurt in other cases.
-            atype.clearAnnotations();
+        // TODO: only produce output if the removed annotation isn't the top or default
```
mernst, line 1016: Does "produce output" refer to warning messages or something related to code generation?  Maybe you can change "produce output" to "warn".

mernst (approved):
There is a commit that was reverted.  What is the plan for it?  Will it be restored to this pull request, or will you open a new pull request that contains it?

https://github.com/typetools/checker-framework/pull/4025 adds `@AnnotatedFor` to stub files, and it will be merged shortly.

PR #4008, opened 2020-12-13
Pull request to illustrate false positive in Initialization Checker

mernst:
Do not merge, but this pull request illustrates a warning that looks like a false positive to me.
I was not able to reproduce it in a separate file that contains only the method that yields the type-checking warning.
Adding the `-AatfDoNotCache` command-line argument did affect the results (unless I wrote it in the wrong location in the Gradle build file). 

Also see #4009 where I added a `@SuppressWarnings` but there the Checker Framework reports that the `@SuppressWarnings` is redundant.


smillst:
To reproduce the failure, you must pass `-AuseConservativeDefaultsForUncheckedCode=source`.  I've fixed this bug in https://github.com/typetools/checker-framework/pull/4019, so I'm closing this.

PR #4013, opened 2020-12-13
TODOs about LUB computation in WPI

mernst:


framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenesStorage.java
```diff
@@ -257,6 +257,7 @@ private void updateAtmWithLub(AnnotatedTypeMirror sourceCodeATM, AnnotatedTypeMi
 
         switch (sourceCodeATM.getKind()) {
             case TYPEVAR:
+                // TODO: It seems wrong to LUB the type parameter bounds.
```
smillst, line 260: Can you expand on why it seems wrong?  Also, I think these are uses of type variables rather than type parameter declarations.  So they aren't invariant as you suggest in the comment below.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenesStorage.java
```diff
@@ -275,6 +276,7 @@ private void updateAtmWithLub(AnnotatedTypeMirror sourceCodeATM, AnnotatedTypeMi
                 //                              ((AnnotatedWildcardType) jaifATM).getSuperBound());
                 //            break;
             case ARRAY:
+                // TODO: Array components, like type parameters, are invariant.
```
smillst, line 279: I would delete "like type parameters".

smillst (changes requested):

PR #4014, opened 2020-12-13
Infer method contracts

mernst:
Review supertypes before subtypes, to see specifications for newly-added methods.

Merge with typetools/annotation-tools#337.


mernst:
@smillst This is unblocked for review.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
```diff
@@ -192,6 +210,112 @@ private void updateInferredExecutableParameterTypes(
...
+        // Process the store's fields.
+        for (Map.Entry<FieldAccess, ? extends CFAbstractValue<?>> entry :
+                store.getFieldValues().entrySet()) {
+            FieldAccess fa = entry.getKey();
+            CFAbstractValue<?> v = entry.getValue();
+
+            VariableElement fieldElement = fa.getField();
+            AnnotatedTypeMirror fieldType = atypeFactory.getAnnotatedType(fieldElement);
+
+            if (!ElementUtils.inSameClass(methodElt, fieldElement)) {
```
smillst, line 246: This checks if the field is declared in the current class, but it doesn't make sure that the receiver of the field access is `this`.  (For static fields, you don't have to check.)

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
```diff
@@ -192,6 +210,112 @@ private void updateInferredExecutableParameterTypes(
...
+        for (VariableElement fieldElement :
+                ElementFilter.fieldsIn(containingClass.getEnclosedElements())) {
+            FieldAccess fa =
+                    new FieldAccess(
+                            (ElementUtils.isStatic(fieldElement)
+                                    ? classNameReceiver
+                                    : thisReference),
+                            fieldElement.asType(),
+                            fieldElement);
+            if (store.getFieldValue(fa) == null) {
```
smillst, line 272: Instead of looping through all the field accesses in the store above, could you just add an else branch here that does the same thing as the code above?

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -2218,4 +2224,177 @@ public AnnotatedTypeMirror getDefaultValueAnnotatedType(TypeMirror typeMirror) {
...
+     */
+    protected Class<?> AnnotationMirrorToClass(AnnotationMirror am) {
+        try {
+            return Class.forName(AnnotationUtils.annotationBinaryName(am));
+        } catch (ClassNotFoundException e) {
+            throw new BugInCF(e);
+        }
+    }
+
+    // TODO: Implement all the below, probably by copying from the above or abstracting it.
```
smillst, line 2357: It's not clear what "all the below" is referring to.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -2218,4 +2224,177 @@ public AnnotatedTypeMirror getDefaultValueAnnotatedType(TypeMirror typeMirror) {
...
+    public List<AnnotationMirror> getPostconditionAnnotations(
+            AMethod m, List<AnnotationMirror> preconds) {
+        List<AnnotationMirror> result = new ArrayList<>();
+        for (Map.Entry<VariableElement, AField> entry : m.getPostconditions().entrySet()) {
+            result.addAll(getPostconditionAnnotation(entry.getKey(), entry.getValue(), preconds));
+        }
+        return result;
+    }
+
+    // TODO: Implement this here in this class, reading @PreconditionAnnotation meta-annotations.
```
smillst, line 2377: I think this is done? Or should precondition be postcondition?

javacutil/src/main/java/org/checkerframework/javacutil/TypesUtils.java
```diff
@@ -893,4 +894,24 @@ public static TypeMirror substitute(
...
     }
+
+    // This method is necessary because Type.stripMetadata does no work for primitive types.
+    /**
+     * Returns a copy of a type without annotations. May return tm itself if it has no annotations.
+     *
+     * @param tm a TypeMirror
+     * @return a copy of the given type without annotations
+     */
+    public static TypeMirror unannotatedType(TypeMirror tm) {
```
smillst, line 905: There's another version of this method `org.checkerframework.javacutil.TypeAnnotationUtils#unannotatedType`.  It makes more sense to have this method in TypesUtils.  Or at least that's where I always look for it first.

smillst (changes requested):

framework/src/main/java/org/checkerframework/common/wholeprograminference/SceneToStubWriter.java
```diff
@@ -706,7 +723,11 @@ private static void printClass(
             printWriter.println();
             for (Map.Entry<String, AMethod> methodEntry : aClass.getMethods().entrySet()) {
                 printMethodDeclaration(
-                        methodEntry.getValue(), innermostClassname, printWriter, indentLevel);
+                        methodEntry.getValue(),
+                        innermostClassname,
+                        printWriter,
+                        indentLevel,
+                        checker.getTypeFactory());
```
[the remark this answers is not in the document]
mernst: This does pass the AnnotatedTypeFactory.

Are you suggesting that the containing method printClass could have its formal parameter list changed, so it takes an AnnotatedTypeFactory rather than a checker?  The changes in #4025 require the checker rather than the type factory.

framework/src/main/java/org/checkerframework/common/wholeprograminference/scenelib/ASceneWrapper.java
```diff
@@ -144,9 +150,23 @@ public void writeToFile(
             try {
                 switch (outputFormat) {
                     case STUB:
-                        SceneToStubWriter.write(this, filepath);
+                        SceneToStubWriter.write(this, filepath, checker);
```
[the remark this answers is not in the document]
mernst: The checker is still needed for `@AnnotatedFor` annotations, but I can use the same code to compute the contracts in both places.

mernst:
@smillst This is ready for re-review.
There is a CI failure (which I can only sometimes reproduce locally) that I know the cause of.

framework/src/main/java/org/checkerframework/common/wholeprograminference/scenelib/ASceneWrapper.java
```diff
@@ -1,19 +1,24 @@
 package org.checkerframework.common.wholeprograminference.scenelib;
 
+import static org.plumelib.util.CollectionsPlume.mapList;
```
smillst, line 3: I think we avoid statically importing methods/fields.

framework/src/main/java/org/checkerframework/common/wholeprograminference/scenelib/ASceneWrapper.java
```diff
@@ -139,14 +144,48 @@ public void writeToFile(
...
+                        // For stub files, pass in the checker to compute contracts on the fly;
+                        // precomputing yields incorrect annotations, most likely due to nested
+                        // classes.
                         SceneToStubWriter.write(this, filepath, checker);
                         break;
                     case JAIF:
+                        // For .jaif files, precompute contracts because the Annotation File
+                        // Utilities knows nothing about (and cannot depend on) the Checker
+                        // Framework.
+                        for (Map.Entry<String, AClass> classEntry : scene.classes.entrySet()) {
```
smillst, line 175: This for loop is now duplicated above.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1,5 +1,9 @@
 package org.checkerframework.framework.type;
 
+import static org.checkerframework.dataflow.analysis.Analysis.BeforeOrAfter.AFTER;
```
smillst, line 3: More static imports.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -2218,4 +2228,204 @@ public AnnotatedTypeMirror getDefaultValueAnnotatedType(TypeMirror typeMirror) {
...
+     * created, because the qualifier has elements/arguments, which {@code @RequiresQualifier} does
+     * not support.
+     *
+     * <p>This is of the form {@code @RequiresQualifier(expression="this.elt",
+     * qualifier=MyQual.class)} when elt is declared as {@code @A} or {@code @Poly*} and f contains
+     * {@code @B} which is a sub-qualifier of {@code @A}.
+     *
+     * @param elt element for a field, which is declared in the same class as the method
+     * @param f AFU representation of a field's precondition annotations
+     * @return a precondition annotation for the element, or null if none is appropriate
```
smillst, line 2291: This says it returns `null`, but the implementation seems to return the empty list.  I'm not sure which was intended.  If clients should expect `null`, then please add `@Nullable` to the return type.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -2218,4 +2228,204 @@ public AnnotatedTypeMirror getDefaultValueAnnotatedType(TypeMirror typeMirror) {
...
+     *     postconditions
+     * @return postcondition annotations for the method
+     */
+    public List<AnnotationMirror> getPostconditionAnnotations(
+            AMethod m, List<AnnotationMirror> preconds) {
+        List<AnnotationMirror> result = new ArrayList<>();
+        for (Map.Entry<VariableElement, AField> entry : m.getPostconditions().entrySet()) {
+            result.addAll(getPostconditionAnnotation(entry.getKey(), entry.getValue(), preconds));
+        }
+        Collections.sort(result, Ordering.usingToString());
```
smillst, line 2276: This can be replaced with `result.sort(Ordering.usingToString());` if you like.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -2218,4 +2228,204 @@ public AnnotatedTypeMirror getDefaultValueAnnotatedType(TypeMirror typeMirror) {
...
+     * qualifier=MyQual.class)} when elt is declared as {@code @A} or {@code @Poly*} and f contains
+     * {@code @B} which is a sub-qualifier of {@code @A}.
+     *
+     * @param elt element for a field
+     * @param f AFU representation of a field's postcondition annotations
+     * @param preconds the precondition annotations for the method; used to suppress redundant
+     *     postconditions
+     * @return a postcondition annotation for the element, or null if none is appropriate
+     */
+    public List<AnnotationMirror> getPostconditionAnnotation(
```
smillst, line 2315: This says it returns `null`, but the implementation seems to return the empty list.  I'm not sure which was intended.  If clients should expect `null`, then please add `@Nullable` to the return type.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -2218,4 +2228,204 @@ public AnnotatedTypeMirror getDefaultValueAnnotatedType(TypeMirror typeMirror) {
...
+     *
+     * <p>This implementation makes no assumptions about preconditions suppressing postconditions,
+     * but subclasses may do so.
+     *
+     * @param elt element for a field
+     * @param f AFU representation of a field's postcondition annotations
+     * @param preOrPost whether to return preconditions or postconditions
+     * @param preconds the precondition annotations for the method; used to suppress redundant
+     *     postconditions; non-null exactly when {@code preOrPost} is {@code AFTER}
+     * @return a postcondition annotation for the element, or null if none is appropriate
```
smillst, line 2335: This says it returns `null`, but the implementation seems to return the empty list.  I'm not sure which was intended.  If clients should expect `null`, then please add `@Nullable` to the return type.

smillst (approved):
I just had a few minor comments.

framework/src/main/java/org/checkerframework/common/wholeprograminference/scenelib/ASceneWrapper.java
mernst: If this is not statically imported, then the call is little shorter than the verbose stream version.  I propose to make an exception for this method.

Issue #4005, opened 2020-12-13
`Optional.isEmpty()` not handled.


t-8ch:
It seems that `Optional.isEmpty()` is not handled by checker framework.
For that matter, no methods added to `java.util.Optional` in Java 9 or later are handled.

`.isEmpty()` is the exact inverse of the currently handled `.isPresent()`.
Switching the code to `!o.isPresent()` would be possible, but then another code checker complains that it should be rewritten as `o.isEmpty()`.

How are APIs that are not present in all JDK versions to be handled?
I could take a shot at implementing support for missing features for `j.u.Optional`.

## Commands

`javac -processor org.checkerframework.checker.optional.OptionalChecker OptionalDemonstration.java`

## Inputs

```java
import java.util.Optional;

public class OptionalDemonstration {
    public Object works() {
        Optional<String> o = makeOptional();
        
        if (o.isEmpty() || o.get().contains("foo")) {
            return 1;
        }
        return 2;
    }
    
    public Object doesNotWork() {
        Optional<String> o = makeOptional();

        if (!o.isPresent() || o.get().contains("foo")) {
            return 1;
        }
        return 2;
    }
    
    private static Optional<String> makeOptional() {
        return Optional.empty();
    }
}

```

## Output

```
OptionalDemonstration.java:9: error: [method.invocation.invalid] call to get() not allowed on the given receiver.
        if (o.isEmpty() || o.get().isEmpty()) {
                                ^
  found   : @MaybePresent Optional
  required: @Present Optional
1 error
```

## Expectations

No error should be reported. The same way `!Optional.isPresent()` works.

wmdietl:
Thanks for the report!
Method `isEmpty()` needs a post-condition annotation `@EnsuresQualifierIf(result = false, expression = "this", qualifier = Present.class)` that tells the Optional Checker that if `isEmpty` returns false, the value is `@Present`.

Could you prepare a PR for this file:
https://github.com/typetools/jdk/blob/master/src/java.base/share/classes/java/util/Optional.java#L196

t-8ch:
Will do.
Shouldn't there not also be support `OptionalVisitor`, for all features?

mernst:
Thanks for noticing the missing annotations, and for the pull requests.  We appreciate it.

> For that matter, no methods added to `java.util.Optional` in Java 9 or later are handled.
> How are APIs that are not present in all JDK versions to be handled?

[...]

PR #4015, opened 2020-12-13
j.u.Optional: test more methods

t-8ch:
Fixes https://github.com/typetools/checker-framework/issues/4005.
Merge with https://github.com/typetools/jdk/pull/96.

I'm not sure why `mos.orElseThrow(exceptionSupplier);` should be rejected.


mernst:
> I'm not sure why `mos.orElseThrow(exceptionSupplier);` should be rejected.

The reason is the `@Present` annotation on the receiver of `orElseThrow`.
I think you knew that, and you are wondering why the method has that annotation.
That is explained in the manual in [section "Methods that crash when passed certain values"](https://checkerframework.org/manual/#annotate-normal-behavior-always-crash) (see its last paragraph in particular), and in [section "What the Optional Checker guarantees"](https://checkerframework.org/manual/#optional-guarantees).  Please let us know if this is not clear, so that we can clarify the documentation.

t-8ch:
That makes sense.
I never used the checker framework before and have missed the `@Present` annotation on the receiver.
(And didn't read all of the manual..)
Thanks for explaining!

mernst (approved):

mernst:
It was confusing (and buggy!) that one `orElseThrow` override had the `@Present` annotation, but the other override lacked the `@Present` override.  I adjusted your pull request to also fix that inconsistency, which might have contributed to your confusion.  Thanks for bringing that to our attention.

smillst (approved):

smillst:
One of the typetools/jdk#96 jobs failed early this morning.  It looks like the fix in 35ed370, should make it pass, so I reran the failed job.

PR #4017, opened 2020-12-13
Improve AnnotationFileParser diagnostics

mernst:


framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileParser.java
```diff
@@ -1315,6 +1330,14 @@ private void recordDeclAnnotation(
                     // `annoMirror` is applicable to `elt`
                     annos.add(annoMirror);
                 }
+            } else {
+                // TODO: Warn about what file it was found in.
+                if (false) {
+                    stubWarnNotFound(
```
smillst, line 1336: I don't understand why this is disabled.

framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileParser.java
```diff
@@ -1284,7 +1298,8 @@ private void annotate(AnnotatedTypeMirror type, List<AnnotationExpr> annotations
             if (annoMirror != null) {
                 type.replaceAnnotation(annoMirror);
             } else {
-                stubWarnNotFound("Unknown annotation: " + annotation);
+                // TODO: Be more specific: In what file?
```
smillst, line 1301: The field `filename` can be used for this.  Maybe all AnnotationFileParser warnings should report the filename?

smillst (approved):

checker/jtreg/stubs/issue2059/AnnoNotFound.out
```diff
@@ -1,2 +1,2 @@
-- compiler.warn.proc.messager: AnnotationFileParser: Unknown annotation: @Nullable
+- compiler.warn.proc.messager: AnnotationFileParser: Unknown annotation @Nullable in /home/mernst/research/types/checker-framework-fork-mernst-branch-AnnotationFileParser-diagnostics/checker/jtreg/stubs/issue2059/AnnoNotFound.astub
```
mernst, line 1: @smillst Do you know offhand how to write this test so it will pass on all file systems?  (One way would be to add a flag, used only for debugging, that disables this output.)

smillst: Maybe if `-Anomsgtext` is passed, then don't output the filename?

smillst (approved):
The warnings are so much clearer now.  Thanks!

PR #4023, opened 2020-12-15
Rename `ThisLiteral` to `This`, because it's not a literal

mernst:


dataflow/src/main/java/org/checkerframework/dataflow/cfg/node/ImplicitThisKeywordNode.java
```diff
@@ -5,9 +5,9 @@
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 /** A node to model the implicit {@code this}, e.g., in a field access. */
-public class ImplicitThisLiteralNode extends ThisLiteralNode {
+public class ImplicitThisKeywordNode extends ThisKeywordNode {
```
smillst, line 8: I find this name a little confusing.  I don't know what an implicit key word is.  I think `ThisNode`, `ExplicitThisNode`, and `ImplicitThisNode` are a little less confusing.  (Although, `ThisNode` might be confusing.)  (I'm ok with `*KeywordNode` versions, too.)

smillst (approved):
FYI, ErrorProne will need to adapt to this change.

PR #4027, opened 2020-12-15
Stub files with no package and an annotation on a class fails to parse

smillst:
This adds a failing test case, so it shouldn't be merged.


mernst:
The test case was added in https://github.com/typetools/checker-framework/commit/62072f72c52ac82fb28f1298d8fc92a599e9d26a and it now passes.

PR #4056, opened 2020-12-21
Support fake overrides in stub files

mernst:
Currently, only changed return types have any effect.

I am blocked (and CI is failing) because of an error "AsSuperVisitor: type is not an erased subtype of supertype."  I tried to work around it, but you will be able to fix it or work around it much more quickly than I can.

Some of the changes here might need to be reverted or changed when that problem is resolved.


smillst (commented):
I fixed the crash: 98f4869bc52dc05992b4d6073278f5c25f726854.  Hopefully you can revert some of the union special casing now.

PR #4059, opened 2020-12-22
Clarify instructions in the custom checker wpi script

kelloggm:
Based on feedback from @t-rasmud


docs/examples/wpi-many/wpi-many-custom-checker-example.sh
```diff
@@ -24,7 +24,9 @@ checkername=no-literal
 repolist=securerandom.list
 workingdir=$(pwd)
 
-# the stub files for the checker being used
+# The stub files for the checker being used.
+# If you are not using stub files, you MUST comment this line out and modify the last command in
```
mernst, line 28: Rather than modifying the last command, I think this script would be easier to use if there is logic that checks whether the variable is set and provides the `--stubs` option only if the variable is set.  Similarly for `--libs`.

mernst (changes requested):

docs/examples/wpi-many/wpi-many-custom-checker-example.sh
kelloggm: I don't disagree that the change you're suggesting would make the script easier to use. However, this script isn't meant to be used - it is an example/template for a user to write their own script. For that reason, the script's logic should be as simple as possible - the "user" is expected to read and understand the entire script, modifying each line for their use-case as they go. No one should run this script without first understanding it and modifying it themselves.

mernst: The fewer modifications are necessary, the fewer errors users are likely to make.  I think it is better to make the script just a little bit more capable, rather than leaving landmines for less-than-perfect users to stumble over.

docs/examples/wpi-many/wpi-many-custom-checker-example.sh
```diff
@@ -71,6 +96,6 @@ bash wpi-many.sh -o "${workingdir}/${checkername}-${repolistbase}" \
      -t ${timeout} \
      -- \
      --checker "${checker}" \
-     --quals "${qual_classpath}" \
-     --lib "${checker_classpath}" \
-     --stubs "${custom_stubs}"
+     "${quals_arg}" \
```
mernst, line 99: I wouldn't think these variables should be quoted, because they represent multiple tokens.  If this works as is, then fine.

mernst (approved):

docs/examples/wpi-many/wpi-many-custom-checker-example.sh
kelloggm: My understanding has always been that variables that might be empty or contain whitespace should be quoted. This version also passes shellcheck as-is, though I didn't actually run the script (I don't have the environment to do so anymore).

mernst: If the variable is empty, your goal is not to pass a single empty token to the program; rather, your goal is to pass no tokens to the program.

Shellcheck assumes you want to pass a single token; a warning suppression is necessary in cases like this.

kelloggm: Thanks for the explanation. I looked through https://github.com/koalaman/shellcheck/wiki/SC2086 and found an alternative that avoids the shellcheck warning but also has the behavior we're looking for, I think, using alternate expansion (the discussion of this approach is near the bottom of that page).

docs/examples/wpi-many/wpi-many-custom-checker-example.sh
```diff
@@ -71,6 +94,6 @@ bash wpi-many.sh -o "${workingdir}/${checkername}-${repolistbase}" \
      -t ${timeout} \
      -- \
      --checker "${checker}" \
-     --quals "${qual_classpath}" \
-     --lib "${checker_classpath}" \
-     --stubs "${custom_stubs}"
+     ${quals_arg:+"--quals ${qual_classpath}"} \
```
mernst, line 97: This quoting is too broad.  It should be
```
${quals_arg:+--quals "${qual_classpath}"}
```
(See the very last example at https://github.com/koalaman/shellcheck/wiki/SC2086.)

kelloggm: Nice catch. I've fixed it.

mernst (approved):

t-rasmud (approved):

PR #4072, opened 2020-12-26
Test WPI on HEAD of plume-lib projects

mernst:
Includes the changes from https://github.com/typetools/checker-framework/pull/4069 (so review and merge it first).


checker/bin-devel/wpi-plumelib/README
```diff
@@ -0,0 +1,2 @@
+This directory contains the expected result of type-checking projects from
+https://github.com/plume-lib/, after whole-program inference.
```
kelloggm, line 2: This file needs to contain instructions for how to produce these files, because changes to plume-lib or to the checkers might cause these tests to fail only because these files need to be regenerated.

kelloggm (changes requested):
Thanks for building this!

My only concern is that these tests might be brittle - that's why `wpiManyTests` uses a frozen version of the plume-lib repositories. In particular, these tests will break if either:
* one of the `plume` repositories changes in a way that changes the output of the checkers (which will require us to create a new expected output file), or
* one of the false positives (i.e. expected errors) in one of these `plume` repositories is fixed by a change to one of the typecheckers

I suggest adding a note to `checker/bin-devel/wpi-plumelib/README` that mentions the two cases above that could cause failures, and also a process for regenerating the output files in those cases (which I think should just be copying the `typecheck.out` files produced by running WPI manually on the target repositories?).

mernst:
I agree with your comments and suggestions.  Thanks.

checker/build.gradle
```diff
@@ -617,3 +617,19 @@ task wpiManyTests(group: "Verification") {
...
+
+// This is run as part of the nonJunitTests task.
+task wpiPlumeLibTests(group: "Verification") {
+    description 'Tests whole-program inference on the plume-lib projects. Requires an Internet connection.'
+
+    // This test must always be re-run when requested.
+    outputs.upToDateWhen { false }
+
+    doLast {
+        // Run wpi-many.sh
```
kelloggm, line 629: This comment isn't accurate.

kelloggm (approved):
Minor nit on a typo in a comment, but otherwise this looks great. Thanks!

PR #4081, opened 2020-12-28
Misc. fixes to wpi-many scripts

kelloggm:
@mernst this fixes all of the problems that you reported to me via email, except the failures related to JDK 8 that #4067 addresses, which I haven't yet tried to debug. It also fixes one issue that @t-rasmud noticed and email me about, too.


checker/bin/wpi-many.sh
```diff
@@ -162,7 +173,7 @@ do
 
     if [ "${OWNER}" = "${GITHUB_USER}" ]; then
         ORIGIN=$(echo "${REPOHASH}" | awk '{print $3}')
-        git remote add unannotated "${ORIGIN}"
+        git remote add unannotated "${ORIGIN}" &> /dev/null
```
mernst, line 176: Rather than piping to /dev/null, it is usually better to use `-q` so that error messages are shown.

mernst (approved):

checker/bin/wpi-many.sh
kelloggm: My intention was actually to suppress error messages, which I found distracting and unhelpful.

In particular, running `git remote add` twice on the same repository always causes an error, but I don't know of an easy way to check if the remote already exists. Since the script re-uses a repository if its already present (to avoid downloading the repo again unnecessarily), running the script twice deterministically produces failures from this line.

mernst: Thanks for the explanation.  Could you put a note in the file to explain that this usage is acceptable (even though in general one should not use it)?

PR #4084, opened 2021-01-01
Warn the user if the JVM has too little memory

mernst:


javacutil/src/main/java/org/checkerframework/javacutil/GC.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     * A list of pairs of (timestamp, accumulated collection time). The timestamp is an epoch
+     * second, and the collection time is in milliseconds. New items are added to the front.
+     */
+    private static List<Pair<Long, Long>> gcHistory = new ArrayList<>();
+
+    /**
+     * Returns the percentage of time spent garbage collecting, in the past minute. Might return a
+     * value greater than 1 if multiple threads are spending all their time collecting. * Returns 0
```
smillst, line 21: Should there be a new line before `* Returns` or should `*` be deleted?

smillst (approved):

PR #4085, opened 2021-01-02
Rewrite remove-annotations program in Java

mernst:


framework/src/main/java/org/checkerframework/framework/stub/RemoveAnnotationsForInference.java
```diff
@@ -0,0 +1,515 @@
...
+        }
+    }
+
+    /**
+     * Rewrites the file in place, removing the given annotations from it.
+     *
+     * @param absolutePath the path to the file
+     * @param removals the annotations to remove
+     */
+    static void removeAnnotations(Path absolutePath, List<AnnotationExpr> removals) {
```
smillst, line 140: I was confused as to why you weren't using JavaParser to remove the annotations.  This comment:
```
    // An earlier implementation used ModifierVisitor.  However, JavaParser's unparser can change
    // the structure of the program. For example, it changes `protected @Nullable Object x;` to
    // `@Nullable protected Object x;` which yields a type.anno.before.modifier error.
```
This comment explains why.  It would be helpful if it was closer to this code.

smillst (approved):

PR #4094, opened 2021-01-05
Define two utility methods

mernst:


javacutil/src/main/java/org/checkerframework/javacutil/ElementUtils.java
```diff
@@ -339,7 +340,23 @@ public static boolean isElementFromByteCode(@Nullable Element elt) {
...
 
-    /** Returns the field of the class or {@code null} if not found. */
+    /**
+     * Returns the path to the source file containing {@code element}.
+     *
+     * @param element the type element to look at
+     * @return path to the source file containing {@code element}
+     */
+    public static String getSourceFilePath(TypeElement element) {
+        return ((ClassSymbol) element).sourcefile.toUri().getPath();
```
smillst, line 350: What happens if element wasn't defined in source?

smillst (changes requested):

PR #4099, opened 2021-01-06
Clarify code in DependentTypesHelper.

smillst:


framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -746,30 +746,26 @@ public Void visitTypeVariable(AnnotatedTypeMirror.AnnotatedTypeVariable type, Vo
...
-                // Work around this bug by remove all annotations of the same class.
-                if (type.removeAnnotation(anno)) {
+                if (newAnno != null) {
+                    // Standardized annotations are written into bytecode along with explicitly
+                    // written nonstandard annotations. (This is a bug.)
+                    // Remove the old annotation and add the new annotations. The new annotation for
+                    // both the nonstandard annotation and the standard annotation are equal with
+                    // respect to Object#equals, so only one new annotation will be added to the
+                    // type.
                     type.removeAnnotation(anno);
```
mernst, line 765: Could you also document that removeAnnotation removes any annotation of the given class/name, not just the argument?  (I think that is relevant to the correctness of this code.)

mernst (approved):

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
smillst: #4103 changes the behavior of type.removeAnnotation to actually remove the annotation passed, not just one matching the class.

PR #4130, opened 2021-01-11
Fix requires non null javadoc

ChristosToxiadis:
I fixed some small mistakes in  RequiresNonNull Javadoc. The Javadoc i am referring to is in the following link  https://checkerframework.org/api/org/checkerframework/checker/nullness/qual/RequiresNonNull.html


mernst (approved):
Thanks for fixing these mistakes!  We appreciate it.

PR #4131, opened 2021-01-11
Use the Maven publish Gradle plugin.

### LINKED ISSUE #3202: Use the "maven-publish" Gradle plugin to deploy artifacts to Maven Central
Currently, the Gradle scripts use `exec` to run the `mvn` command-line tool.
It would be nicer to use the "maven-publish" Gradle plugin to deploy artifacts to Maven Central.
That would reduce the amount of configuration code and would make it easier to create snapshot releases.

From @msridhar 
>FWIW, WALA now uses "maven-publish" for creating snapshot and release artifacts and pushing them to Maven Central. Here is the relevant Gradle code:

>https://github.com/wala/WALA/blob/master/gradle-mvn-push.gradle

smillst:
Fixes #3202.


docs/developer/release/release_push.py
```diff
@@ -149,11 +150,11 @@ def check_all_links(afu_website, checker_website, suffix, test_mode, checker_ver
         if not test_mode:
             release_option = " release"
         raise Exception("The link checker reported errors.  Please fix them by committing changes to the mainline\n" +
-                        "repository and pushing them to GitHub/Bitbucket, running \"python3 release_build.py all\" again\n" +
-                        "(in order to update the development site), and running \"python3 release_push" + release_option + "\" again.")
+                        "repository and pushing them to GitHub, running \"python release_build.py all\" again\n" +
```
mernst, line 153: I think this should be `python3` instead of `python` which might be Python 2 or Python 3 depending on the system.

mernst (approved):
Looks great, with a nice reduction of code size.

docs/developer/release/release_push.py
smillst: These scripts are not yet written in python3.

mernst: The [shebang line](https://github.com/typetools/checker-framework/blob/master/docs/developer/release/release_build.py#L1) contains `python3`.
If that is incorrect, the shebang line should be fixed.

smillst: I thought these scripts needed to be converted to Python 3 and that you haven't done that yet.  I see that you changed the shebang on Dec. 21 afd6923c, did you also convert the scripts?

build.gradle
```diff
@@ -1044,13 +1044,6 @@ final sharedPublicationConfiguration(publication) {
                 organization = 'University of Washington'
                 organizationUrl = 'https://www.cs.washington.edu/'
             }
-            developer {
-                id = 'wmdietl'
```
wmdietl: Strange correction.

mernst: You said last August that you no longer wanted to be a developer/maintainer.  This reflects that and takes you off the hook for maintenance that you do not want to do.

docs/developer/release/release_push.py
mernst: Good point -- just changing the shebang line is not enough.  I have opened https://github.com/typetools/checker-framework/pull/4134.

build.gradle
wmdietl: That's not how that happened, but OK.

PR #4135, opened 2021-01-12
Improve the documentation of standardization routines

mernst:


framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
```diff
@@ -606,32 +606,32 @@ protected boolean ignoreFieldInWPI(Element element, String fieldName) {
...
+     * @param methodDecl the declaration of the method whose return type may be updated
      * @param overriddenMethods the methods that the given method return overrides, each indexed by
      *     the annotated type of the class that defines it
      */
     @Override
     public void updateFromReturn(
             ReturnNode retNode,
             ClassSymbol classSymbol,
-            MethodTree methodTree,
+            MethodTree methodDecl,
```
smillst, line 629: I prefer `methodTree`.  A method declaration can also be represented by an element, so it's not clear whether a variable name `methodDecl` is a tree or an element..  But you can leave it if you feel strongly.

smillst (approved):

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceScenes.java
mernst: @smillst I found it confusing to know from the name `methodTree` whether it is a method *call* or method *declaration*.  How about `methodDeclTree`?

smillst: That works for me.

PR #4149, opened 2021-01-13
Deprecated `ElementUtils.enclosingClass`; use `ElementUtils.enclosingType`

mernst:


javacutil/src/main/java/org/checkerframework/javacutil/ElementUtils.java
```diff
@@ -55,10 +55,22 @@ private ElementUtils() {
...
+        return enclosingType(elem);
+    }
+
+    /**
+     * Returns the innermost type element enclosing the given element.
+     *
+     * @param elem the enclosed element of a class
+     * @return the innermost type element, or null if no type element encloses {@code elem}
+     */
+    public static @Nullable TypeElement enclosingType(final Element elem) {
```
smillst, line 71:
```
class Outer {
    static class Inner {  }
}
```
If the element is `Inner` this method returns `Outer`, which is not the enclosing type of  `Inner`.  Inner has no enclosing type.  (See `DeclaredType#getEnclosingType`.) I would name this method `enclosingTypeElement`.

smillst (approved):

PR #4150, opened 2021-01-13
changes to support new version of Annotation Utilities with no internal asmx

markro49:


markro49:
moving to my typetools fork

PR #4162, opened 2021-01-15
Add 'wpiPlumeLibTests' back to nonJunitTests.

smillst:


mernst:
Pulling master after #4166 is merged may resolve the CI failure.

PR #4166, opened 2021-01-17
Forbid `-Ainfer` without `-Awarns`

mernst:
Merge after ~https://github.com/kelloggm/do-like-javac/pull/14~; after merging that pull request, the failing tests for this pull request can be re-started.  (Or just pull master into this branch.)


kelloggm:
@mernst I've merged kelloggm/do-like-javac#14 and then merged master into this branch to re-run the tests. They are still failing, for a reason that appears to be unrelated to do-like-javac (the `wpiNullnessJaifs` tests). Can you take another look?

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -487,6 +488,11 @@ public AnnotatedTypeFactory(BaseTypeChecker checker) {
                                     + " should be one of: -Ainfer=jaifs, -Ainfer=stubs");
             }
             wholeProgramInference = new WholeProgramInferenceScenes(this);
+            if (!checker.hasOption("warns")) {
+                // Without -Awarns, the inference output may be incomplete, because javac halts
```
smillst, line 492: An alternative to issuing an error, would be to make it so that `-Ainfer` implies `-Awarns`.

mernst: Yes, that is a possibility. I think that behavior might surprise users (who hadn't read the manual carefully), so I wanted the `-Awarns` command-line argument to be explicit.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -487,6 +488,11 @@ public AnnotatedTypeFactory(BaseTypeChecker checker) {
...
+                // after issuing an error.
+                checker.message(Diagnostic.Kind.ERROR, "Do not supply -Ainfer without -Awarns");
```
kelloggm, line 494: What about:

```suggestion
                checker.message(Diagnostic.Kind.ERROR, "Do not supply -Ainfer without -Awarns. Without -Awarns, javac will halt after issuing errors, which may lead to incomplete inference results.");
```

kelloggm (approved):
Lgtm other than my minor suggestion about the text of the error message. Feel free to ignore my suggestion if you disagree with it.

smillst (approved):

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
mernst: It's a good inclination to provide information to the user.  I agree in general.  Thanks for the suggestion.

Two countervailing principles are:
 * keep messages short, which makes them easier to read and interpret; provide only the essential information.
 * avoid duplicating text, for example in the manual and in the warning message, unless necessary.

I think the rationale is easy to find in the manual if the user wants it, so I'm going to leave the message short for now.  If users get confused, I will admit I was wrong and we can make the message more lengthy.

PR #4167, opened 2021-01-17
Use appropriate branch of do-like-javac

mernst:


kelloggm:
@mernst, sorry for letting this PR fester for two months. In the meantime, several of the changes it proposed have become irrelevant or been subsumed by other PRs. What's left is a relatively simple change that makes the cloning of DLJC more defensive. When you get a chance, can you take a look at that and make sure that it seems reasonable, and then merge this? Thanks.

kelloggm (approved):

PR #4169, opened 2021-01-18
Exit if listpath file is empty

mernst:
@kelloggm Why does this change make the WPI tests fail?  Does that indicate a bug that has been in the scripts all along, a latent bug that this exposes, or a problem with this change?


mernst:
@kelloggm This pull request passes CI now.

kelloggm (approved):
LGTM

PR #4174, opened 2021-01-19
Fix formatting of array types

mernst:


javacutil/src/main/java/org/checkerframework/javacutil/TypesUtils.java
```diff
@@ -219,6 +219,21 @@ public static String simpleTypeName(TypeMirror type) {
         return null;
     }
 
+    // TODO: Is there a better name for this concept?
```
smillst, line 222: Maybe base component type or inner most component type?

smillst (approved):

PR #4176, opened 2021-01-20
[dataflow] Class literal nodes should have exceptional edges

artempyanykh:
Summary:
Absence of exceptional edges leads to missing CFG nodes in a code like
this:
```
try {
  var cls = MyClass.class;
  ...
} catch (NoClassDefFoundError e) {
  ...
}
```

According to JLS evaluating a class literal can fail in a number of
ways which are handled by this patch.

Test plan:
Added a test case under nullness to check if the types inside `catch`
get refined.

### DIFF
--- a/checker/tests/nullness/TryCatch.java
+++ b/checker/tests/nullness/TryCatch.java
@@ -27,4 +27,15 @@ public class TryCatch {
             t.toString();
         }
     }
+
+    void noClassDefFoundError(@Nullable Object x) {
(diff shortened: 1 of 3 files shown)


artempyanykh:
Re-rendered table in the manual:
Screenshot 2021-01-20 at 13 17 14

mernst (approved):

mernst:
Thanks for this improvement!

PR #4178, opened 2021-01-20
Move util classes from checker-qual to checker-util.

smillst:
Also, added `util` to each package name of the util classes.  That way the packages aren't split across jars.  (For example, `org.checkerframework.checker.nullness` is in `checker.jar` and `org.checkerframework.checker.nullness.util` is in `checker-util.jar`.

(This includes the changes in https://github.com/typetools/checker-framework/pull/4181.  You can merge that one first or just merge this one.)


framework/src/main/java/org/checkerframework/framework/util/CheckerMain.java
```diff
@@ -95,6 +99,12 @@ public static void main(String[] args) {
      */
     public static final String CHECKER_QUAL_PATH_OPT = "-checkerQualJar";
 
+    /**
+     * Option name for specifying an alternative checker-util.jar location. The accompanying value
+     * MUST be the path to the jar file (NOT the path to its encompassing directory)
+     */
+    public static final String CHECKER_UTIL_PATH_OPT = "-checkerQualJar";
```
mernst, line 106: Should this be `-checkerUtilJar`?

PR #4189, opened 2021-01-23
Common CFG for subcheckers

kelloggm:
This change allows a checker that uses dataflow to share the structure of its CFG (i.e. the `Node`s and `Block`s) with its subcheckers that also use dataflow. The primary motivation for this change is to allow a superchecker to use its own `Node`s to query the store of one of its subcheckers, but it should also give subcheckers a small performance boost by avoiding duplicate CFG construction. There ought to be no visible change in behavior before and after this change (though that's not quite true - see the small change to the documentation of the `@AssumeAssertion` warning suppression mechanism and item 2 in the list below.

The primary change is quite simple, but I had to work around two issues that sharing the CFG caused:
1. artificial nodes for foreach loops where the loop variable is a type variable were assigned the wrong type by dataflow - they were treated as `@Top T` rather than just `T`, which caused false positives when they were (psuedo-)assigned into other uses of the type variable (which ought to be case 3a of the type variable subtyping rules, but wasn't because of the extra inferred top annotation). An example of this is in the new test `TypeVarAndArrayRefinementSmall.java`, which is a minimized version of the existing `TypeVarAndArrayRefinement.java` test.
2. The warning suppression mechanism using assert statement with `@AssumeAssertion(checkername)` messages stopped working. The reason was that these warnings were suppressed during CFG construction, by pruning the assertion branch that could possibly fail if the checker for which the CFG was being constructed matched the string in the `@AssumeAssertion`. Because a CFG is now only constructed once by the first subchecker to run, `@AssumeAssertion` messages referring to supercheckers did not cause this pruning. The code here handles this by pruning if the message matches the checker or any of its ultimate parent checker's subcheckers (including the ultimate parent itself). This is a slight change to the semantics, but I don't think it could have an effect in practice - that would require both the subchecker and the superchecker to do refinement based on the same assertion, but for the assertion to only be respected by one of the two. I'm not even sure it's possible to find an example of the first condition (maybe in the Index Checker, related to integers?), but I'm quite confident that the second condition makes this safe - subcheckers shouldn't ever be assuming that assertions that their supercheckers respect will be false, and they definitely shouldn't rely on that for correctness.


kelloggm:
@mernst I just made a large change to this PR that fixes the issue we discussed on Monday in plume-lib/options, and, IMO, greatly simplifies the code. I also added some more tests, including one for that issue. There is only one unfortunate side-effect of this: it requires a small change in an annotation argument in Daikon for Daikon to typecheck: https://github.com/codespecs/daikon/pull/312

The change is required because the previous version of the code relied on the call into the AnnotatedTypeFactory to standardize annotations on the array's components. That's no longer possible, because the relevant type factory might not be available when the CFG is constructed. I don't think this is too serious - it's a rare case and it should be easy for a user to make a change like the one I made in the Daikon PR to fix it when it does come up.

If we're not willing to live with that minor annoyance, we could add logic in the type factory to force standardization for artificial trees. I tried to do so in `GenericAnnotatedTypeFactory#addComputedTypeAnnotations(Tree, AnnotatedTypeMirror)`, in the same way that `GenericAnnotatedTypeFactory#addComputedTypeAnnotations(Element, AnnotatedTypeMirror)` already does, but it didn't fix the problem.

checker/tests/nullness/DaikonEnhancedFor.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @SuppressWarnings("nullness")
+    Map<Object, Set<@KeyFor("cmap") Object>> cmap = null;
+
+    @SuppressWarnings("nullness")
+    Object[] getObjects() {
+        return null;
+    }
+
+    void process(@KeyFor("this.cmap") Object super_c) {
```
kelloggm, line 17: create another copy of this test case with expected warnings for the lack of "this"

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -268,6 +269,15 @@
...
             flowResultAnalysisCaches;
 
+    /**
+     * Subcheckers share the same ControlFlowGraph for each analyzed code statement. This maps from
+     * code statements to the shared control flow graphs. This map is null in all subcheckers (i.e.
+     * it is null exactly when {@code this.checker.parentChecker} is non-null, and vice-versa).
+     *
+     * <p>The initial capacity of the map is set by {@link #getCacheSize()}.
+     */
+    protected @Nullable Map<Tree, ControlFlowGraph> subcheckerSharedCFG;
```
mernst, line 279: I am concerned that this map is not reset by `setRoot`.  It seems to me that it will grow without bound and will prevent the garbage collection of its contents.  Both of these are likely to lead to memory problems.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -2434,6 +2444,61 @@ public AnnotatedTypeMirror getDefaultValueAnnotatedType(TypeMirror typeMirror) {
...
+        } else if (parentChecker != null) {
+            return false;
+        }
+        if (this.subcheckerSharedCFG == null) {
+            this.subcheckerSharedCFG = new HashMap<>(getCacheSize());
+        }
+        if (!this.subcheckerSharedCFG.containsKey(tree)) {
+            this.subcheckerSharedCFG.put(tree, cfg);
+        }
+        return true;
```
mernst, line 2471: Should there be an assertion that the map maps `tree` to `cfg`?

framework/src/main/java/org/checkerframework/framework/flow/CFCFGBuilder.java
```diff
@@ -65,19 +77,33 @@ public static ControlFlowGraph build(
...
+     *     subchecker of the given checker's ultimate parent checker
      */
     public static boolean assumeAssertionsActivatedForAssertTree(
-            SourceChecker checker, AssertTree tree) {
+            BaseTypeChecker checker, AssertTree tree) {
         ExpressionTree detail = tree.getDetail();
         if (detail != null) {
             String msg = detail.toString();
-            Collection<String> prefixes = checker.getSuppressWarningsPrefixes();
+            BaseTypeChecker ultimateParent = getUltimateParent(checker);
```
mernst, line 102: The computation in the next 4 lines is not terribly expensive, but it is a touch expensive, and it occurs for every assertion with a detail message.  Could its value be cached in the checker?  (This would require adding a new method to the checker.)

mernst (changes requested):
I have some questions about memory management.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
kelloggm: Your concern about memory leakage is reasonable.

Right now, this map is set when the first subchecker is run, in the ultimate parent checker (so that it can be shared by e.g. sibling checkers of the first checker to run). (Note that this map is always `null` in subcheckers that have a non-null parent.) If the map were to be reset by `setRoot`, then when that ultimate parent checker starts running (and `setRoot` is called on its ATF), the shared CFG would be cleared - defeating the point.

Ideally, this map should be cleared whenever:
* if the checker has no subcheckers, when `setRoot` is called (it would be easy to implement this case)
* if the checker does have subcheckers, when `setRoot` is called on *only the first* subchecker to run (i.e. the subchecker that is responsible for creating the CFG). This is the tricky case - I'm not sure how to determine in a given call to `setRoot` that the checker whose root is being set is the first subchecker to run (is the run order even defined, or is it arbitrary which subchecker runs first?).

It might be possible to add a flag (call it `okayToOverwriteCFG` inside the ultimate parent that is set to true whenever `setRoot` is called on that checker's ATF (i.e. when it itself is running, meaning all subcheckers have completed), and set it to false whenever the map is re-initialized. I need to think that through a bit more though to make sure it will work, and if you have better ideas I'd love to hear them.

kelloggm:
@mernst I think I've addressed all your concerns. The solution for the memory issues isn't particularly elegant, and I'd certainly appreciate any suggestions on how to improve it.

mernst (approved):
This looks good.  Thanks!

Should all uses of `subcheckerSharedCFG` be protected by a test of whether any subcheckers exist?  That would prevent the need to allocate and fill the cache, which seems like it would be worthwhile.
An efficient way to check for that would be to cache a boolean, or cache the ultimate parent checker.

When you are done, please assign to smillst.

framework/src/main/java/org/checkerframework/framework/flow/CFCFGBuilder.java
```diff
@@ -89,6 +115,13 @@ public static boolean assumeAssertionsActivatedForAssertTree(
         return false;
     }
 
+    /**
+     * A specialized phase-one CFG builder that is aware of Checker Framework annotated types. It is
```
smillst, line 119: Is 1 still true?  I thought you changed that.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -387,7 +387,7 @@ void checkRep(String aliasName) {
     /** Mapping from a Tree to its TreePath. Shared between all instances. */
```
smillst, line 387: In AnnotatedTypeFactory, there is code that copies the artificial trees from one subchecker to another.  I think you can remove this code now:
```
    public final void setEnclosingElementForArtificialTree(Tree tree, Element enclosing) {
        artificialTreeToEnclosingElementMap.put(tree, enclosing);
        for (BaseTypeChecker checker : checker.getSubcheckers()) {
            AnnotatedTypeFactory subFactory = checker.getTypeFactory();
            subFactory.artificialTreeToEnclosingElementMap.putAll(
                    artificialTreeToEnclosingElementMap);
        }
    }
```

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
smillst: What's the status of this comment?

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -391,6 +424,41 @@ public void setRoot(@Nullable CompilationUnitTree root) {
             this.flowResultAnalysisCaches.clear();
             this.initializerCache.clear();
             this.defaultQualifierForUseTypeAnnotator.clearCache();
+
+            if (this.checker.getParentChecker() == null) {
+                // This is an ultimate parent checker, so after it runs the shared CFG it is using
+                // will be dead.
```
smillst, line 430: What does it mean for the CFG to be dead?  You mean null?

smillst (approved):

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
kelloggm: I meant that it's no longer in use (like a variable that is dead in a liveness analysis). I'll update this comment to make it more clear.

framework/src/main/java/org/checkerframework/framework/flow/CFCFGBuilder.java
kelloggm: It's still true in the sense that this class doesn't just use the implementations in `CFGTranslationPhaseOne.java`, but it can definitely be more precise (and the reference to annotated types isn't really correct, either).

For loops using iterators, the only difference is the call to `handleArtificialTree`, so removing 1 would still be accurate.

For loops using arrays, though, there's still extra logic to make sure that local variables have the right declared type. I'll reword the comment to make it more clear what's happening.

kelloggm:
@smillst, for some reason GitHub won't let me reply to you comment about the comment on the declaration of `subcheckerSharedCFG` in `GenericAnnotatedTypeFactory`. This is my reply:

You asked:
> What's the status of this comment?

The comment is question is:
>Subcheckers share the same ControlFlowGraph for each analyzed code statement. This maps from code statements to the shared control flow graphs. This map is null in all subcheckers (i.e. it is null exactly when {`@code` this.checker.parentChecker} is non-null, and vice-versa). <p>The initial capacity of the map is set by {`@link` #getCacheSize()}.

Everything in this comment is still true. I could add a note here about e.g. when this is cleared and re-filled, but there are comments elsewhere and that seems orthogonal to describing what is in the map. Was there something in particular you wanted this comment to include (or not include) that would make what this map contains more clear?

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
smillst: This comment wasn't marked as resolved.  Did you make the changes you propose in this comment?  I think you did make them, but I wasn't sure.  (`subcheckerSharedCFG` is cleared in setRoot via a call to`clearSharedCFG`.)

smillst:
@kelloggm  Follow this link to see my last comment in context:
https://github.com/typetools/checker-framework/pull/4189/files/93bd89002ffb48c32d187cc227adc7ace7697045#r570659811

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
kelloggm: Sorry, I got confused by Github's interface before, and didn't realize that this comment chain is what's being referred to.

To answer your question, yes, I did make these changes. This comment should have been marked resolved.

kelloggm:
@smillst I think that I've addressed your comments. Is this okay to merge?

smillst:
You marked https://github.com/typetools/checker-framework/pull/4189#discussion_r575468827 resolved, but you didn't change the method.

kelloggm:
> You marked #4189 (comment) resolved, but you didn't change the method. Can the method not be changed?

I must have misunderstood your comment. I removed **my changes** to this method, restoring it to its state before this PR. It seems you intended for me to remove the method entirely; I've now done so (and locally it doesn't seem to break anything - we'll see about CI).

smillst:
@kelloggm The change you made wasn't what I was suggesting.  So, I reverted it and made the change I was talking about.  Assuming the tests pass, you can merge.

kelloggm:
@smillst thanks for fixing it the way you meant - sorry I was being dense.

PR #4195, opened 2021-01-23
Define UnaryOperation, parse BinaryOperation

mernst:


dataflow/src/main/java/org/checkerframework/dataflow/expression/ValueLiteral.java
```diff
@@ -37,6 +37,58 @@ public ValueLiteral(TypeMirror type, Object value) {
...
+            return Byte.valueOf((byte) -((Byte) value).byteValue());
+        }
+        if (value instanceof Short) {
+            return Short.valueOf((short) -((Short) value).shortValue());
+        }
+        if (value instanceof Integer) {
+            return Integer.valueOf(-((Integer) value).intValue());
+        }
+        if (value instanceof Long) {
+            return Long.valueOf(-((Long) value).longValue());
```
smillst, line 72: This doesn't work when parsing `-9223372036854775808L`, `Long.MIN_VALUE`.  I added a test case.

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -628,6 +635,129 @@ public JavaExpression visit(ArrayCreationExpr expr, JavaExpressionContext contex
...
+        }
+
+        @Override
+        public JavaExpression visit(BinaryExpr expr, JavaExpressionContext context) {
+            JavaExpression leftJe = expr.getLeft().accept(this, context);
+            JavaExpression rightJe = expr.getRight().accept(this, context);
+            TypeMirror leftType = leftJe.getType();
+            TypeMirror rightType = rightJe.getType();
+            TypeMirror type;
+            if (types.isSameType(leftType, rightType)) {
```
smillst, line 693: You can delete this case.  If the types are the same type, then `types.isSubtype` will return true.

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -628,6 +635,129 @@ public JavaExpression visit(ArrayCreationExpr expr, JavaExpressionContext contex
...
+                    return Tree.Kind.PREFIX_DECREMENT;
+                case PREFIX_INCREMENT:
+                    return Tree.Kind.PREFIX_INCREMENT;
+                default:
+                    throw new Error("unhandled " + op);
+            }
+        }
+
+        @Override
+        public JavaExpression visit(BinaryExpr expr, JavaExpressionContext context) {
```
smillst, line 687: Would you please add a few test cases for this?  You can do so in `framework/tests/flowexpression/`.

smillst (approved):

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
mernst: I was thinking that `isSameType` would be more efficient in the common case.  Is that not true?

smillst: The implementation of `isSubtype` checks for if the two are equal before the more complicated subtype check.

smillst (approved):

Issue #3475, opened 2020-07-14
Use Flow


smillst:
org.checkerframework.checker.index.OffsetDependentTypesHelper

PR #4197, opened 2021-01-23
Make OffsetDependentTypesHelper use standardizeString

mernst:
~Merge after #4195.  (Can review now, examining diffs without those in #4195.)~


framework/src/main/java/org/checkerframework/common/value/ValueCheckerUtils.java
```diff
@@ -389,4 +401,129 @@ public static int getMinLen(Tree tree, ValueAnnotatedTypeFactory valueAnnotatedT
         AnnotatedTypeMirror minLenType = valueAnnotatedTypeFactory.getAnnotatedType(tree);
         return valueAnnotatedTypeFactory.getMinLenValue(minLenType);
     }
+
+    /**
+     * Optimize the given JavaExpression, using Value Checker annotations.
```
kelloggm, line 406: `Optimize` is a vague name. I think I'm fine with it here as the name of this method - it conveys the right intuition for what this method does - but this Javadoc needs to define it. I think the specific meaning here is "constant fold any parts of the given expression that the factory has exact values for", based on reading the code.

kelloggm (changes requested):
I only reviewed the parts that aren't in #4195

framework/src/main/java/org/checkerframework/common/value/ValueCheckerUtils.java
```diff
@@ -389,4 +401,131 @@ public static int getMinLen(Tree tree, ValueAnnotatedTypeFactory valueAnnotatedT
...
+            JavaExpression optLeft = optimize(e.getLeft(), factory);
+            JavaExpression optRight = optimize(e.getRight(), factory);
+            if (e.getLeft() == optLeft && e.getRight() == optRight) {
+                return e;
+            } else {
+                return new BinaryOperation(e.getType(), e.getOperationKind(), optLeft, optRight);
+            }
+
+        } else if (je instanceof ClassName) {
+            ClassName e = (ClassName) je;
```
smillst, line 449: This cast and assignment aren't needed.

smillst (approved):
Does this fix https://github.com/typetools/checker-framework/issues/3475?

mernst:
> Does this fix #3475?

Not quite.  There is still one difference in `OffsetDependentTypesHelper`.  I think that code can be moved into the superclass in a future pull request.

PR #4199, opened 2021-01-25
Added -AenableRegexChecker option to run the Regex Checker as a subchecker of the Nullness Checker; fixes #4006

### LINKED ISSUE #4006: Utilize `RegexChecker`'s `group.count.invalid` check in `NullnessChecker`
Consider the following dummy code:
```sh
$ cat -n Dummy.java 
     1	import java.util.regex.Matcher;
     2	import java.util.regex.Pattern;
     3	import org.checkerframework.checker.nullness.qual.NonNull;
     4	
     5	final class Dummy {
     6	  private static final Pattern PATTERN = Pattern.compile("^.(.*).$");
     7	
     8	  String doSomethingOdd(String input) {
     9	    Matcher matcher = PATTERN.matcher(input);
    10	    if (!matcher.matches()) {
    11	      return "";
    12	    }
    13	
    14	    @NonNull String a = matcher.group(1);
    15	    @NonNull String b = matcher.group(2);
    16	    return a + b;
    17	  }
    18	}
```

Compiling this code using both the `NullnessChecker` and `RegexChecker` we get the following output:
```
$ javac \
    -J--add-opens=jdk.compiler/com.sun.tools.javac.comp=ALL-UNNAMED \
    -processorpath ~/.m2/repository/org/checkerframework/checker/3.8.0/checker-3.8.0.jar \
    -cp ~/.m2/repository/org/checkerframework/checker-qual
/3.8.0/checker-qual-3.8.0.jar \
    -Awarns \
    -processor org.checkerframework.checker.nullness.NullnessChecker,org.checkerframework.checker.regex.RegexChecker \
    Dummy.java
Dummy.java:14: warning: [assignment.type.incompatible] incompatible types in assignment.
    @NonNull String a = matcher.group(1);
                                     ^
  required: @UnknownInitialization @NonNull String
Dummy.java:15: warning: [assignment.type.incompatible] incompatible types in assignment.
    @NonNull String b = matcher.group(2);
                                     ^
  required: @UnknownInitialization @NonNull String
Dummy.java:15: warning: [group.count.invalid] invalid groups parameter 2. Only 1 groups are guaranteed to exist for matcher.
    @NonNull String b = matcher.group(2);
                                      ^
3 warnings
```

As can be seen, the `RegexChecker` is smart enough to realize that `matcher.group(1)` is safe, while `matcher.group(2)` is not. On the other hand, the `NullnessChecker` flags also `matcher.group(1)` as problematic, while this call will not yield `null`.

Which leads me to the feature request: would be be possible to "somehow" share the `RegexChecker` analysis such that the `NullnessChecker` does not produce this false positive?

UtR491:


mernst:
Thanks for this contribution.  It is well-written and clear, which shows the care you took with it.

Unfortunately, this optimization is not sound.  `Matcher.group(int)` can return null even if its argument is legal (is no more than the group count).

The [`Match.group(int)` documentation](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/regex/Matcher.html#group(int)) explains this:

> If the match was successful but the group specified failed to match any part of the input sequence, then `null` is returned.

Here is an example, which is a slight variation of the example you wrote for the manual:

```java
import java.util.regex.Matcher;
import java.util.regex.Pattern;
import java.util.Arrays;

public class Test {

    Pattern PATTERN = Pattern.compile("^.(x)?((x.*)|.*).$");
    Matcher matcher = PATTERN.matcher("hello");
    matcher.matches();
    System.out.printf("group count: %d%n", matcher.groupCount());
    for (int i = 0; i<=matcher.groupCount(); i++) {
      System.out.printf("%d: %s%n", i, matcher.group(i));
    }
  }

}
```

Its output is:

```
group count: 3
0: hello
1: null
2: ell
3: null
```

UtR491:
Thank you for the kind words.

> Unfortunately, this optimization is not sound. Matcher.group(int) can return null even if its argument is legal (is no more than the group count).

Yes I missed these false negatives. So we need to analyze the string we pass to `PATTERN.matcher(String)`. In most of the practical use cases, it will be a runtime variable, so will it be worth the effort to analyze hardcoded strings? From the manual I understand that Regex Checker warns when invalid arguments are passed to functions like `Matcher.group(int)`, so I am assuming the non-regex hardcoded strings are not analyzed.
If I am not mistaken, doesn't this invalidate the original issue (#4006) itself, because `@Nullable` is the right annotation even if a legal call is made?

mernst:
(Sorry for all the delays in responding.)

I added a [comment](https://github.com/typetools/checker-framework/issues/4006#issuecomment-777795734) to #4006.  I don't think that #4006 is invalidated.  In brief:
 * The fact that a string holds a value that is not a compile-time constant is not a problem, so long as the code does a check, such as calling the `isRegex` method.
 * A richer `@Regex` annotation can indicate which groups are definitely non-null, rather than just indicating the number of groups.

mernst:
@UtR491 I suggest that we close this pull request.  #4006 will remain open.  The comments there are much better now and they clarify the scope of the work to fix it.

UtR491:
Hi @mernst, sorry for the delay, I was only able to devote a couple of hours to the issue this week. I am closing this PR for now and will work on it over the weekend and reopen once I have come up with something reviewable. As an update, I did create an EnhancedRegex annotation (that holds NonNull groups) that kind of works for non-concatenation cases but it is still far from reviewable.

mernst:
That sounds great.  There is no rush.  Thanks for your contributions.

UtR491:
@mernst I have come up with an EnhancedRegex annotation that holds a list with guaranteed non-null groups and a total number of capturing groups as the last element. There are a couple of issues however,

1. In the `handleRegexUtil` function of RegexTransfer, I am not sure how to handle `isRegex(String, List)` cases, how to extract values from the node. I also came across the fact that for `isRegex(String, int)` case, `isRegex(s, 1)` passes but `isRegex(s, one)` where one is a final int does not. Was I expected to use something like ellipses?
2. To compute the guaranteed non-null groups, I have traversed the regular expression and analyzed it. I am not sure if it covers all the cases, but I have tried my best.

I will be adding relevant sections to the manual and writing stronger tests too, the one for now just checks the hierarchy is correctly specified.

checker/src/main/java/org/checkerframework/checker/regex/RegexAnnotatedTypeFactory.java
```diff
@@ -261,27 +418,121 @@ private int getRegexValue(AnnotationMirror anno) {
...
         return Pattern.compile(regexp).matcher("").groupCount();
     }
 
+    /**
+     * Returns the list of groups that are non-null in the given regex String.
+     *
+     * @param regexp The string to analyse.
+     * @return A list of non-null groups and the number of groups.
+     */
+    public static List<Integer> getNonNullGroups(@Regex String regexp) {
```
mernst, line 486: This method is duplicative with `RegexUitl.getNonNullGroups`.  Don't duplicate code; instead, call the method in RegexUtil.

mernst (changes requested):

checker/src/main/java/org/checkerframework/checker/nullness/NullnessTransfer.java
```diff
@@ -379,6 +393,27 @@ private boolean containsPolyNullNotAtTopLevel(AnnotatedTypeMirror t) {
             }
         }
 
+        // Refine result to @NonNull if n is an invocation of Matcher.group and the invocation is
+        // verified by the RegexChecker.
```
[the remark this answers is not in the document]
UtR491: Sorry for this. The comments were not fixed after I used the EnhancedRegex annotation for analysis. The same is the case with the above comment.

checker-util/src/main/java/org/checkerframework/checker/regex/util/RegexUtil.java
```diff
@@ -330,4 +360,59 @@ private static String regexErrorMessage(String s, int expectedGroups, int actual
...
+        for (int i = 0; i <= n; i++) {
+            nonNullGroups.add(i);
+        }
+        ArrayDeque<Integer> openingIndices = new ArrayDeque<>();
+        int group = 0;
+        boolean squareBracketOpen = false;
+        boolean escaped = false;
+        boolean notCapturing = false;
+        for (int i = 0; i < regexp.length(); i++) {
+            if (!escaped && !squareBracketOpen && regexp.charAt(i) == '(') {
```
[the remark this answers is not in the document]
UtR491: Thanks for this, turns out I was missing named capturing groups as well.

checker/src/main/java/org/checkerframework/checker/regex/RegexAnnotatedTypeFactory.java
UtR491: I did it because `RegexUtil.getGroupCount(Pattern)` was declared private, so I assumed plume-lib would not allow anything other than `isRegex` and `asRegex` functions to be visible outside the RegexUtil class.

UtR491:
Can we continue the discussion [here](https://github.com/UtR491/checker-framework/pull/6)?  I have added a test case that fails for the first point mentioned in the above [comment](https://github.com/typetools/checker-framework/pull/4199#issuecomment-783228665). The CI build would fail, so I thought I'd merge it when I have resolved that issue.

checker/src/main/java/org/checkerframework/checker/regex/RegexAnnotatedTypeFactory.java
mernst: Now that we have discovered a need for the method, I think it's fine to make the method public or protected.

mernst:
I'm closing this for now, since it depends on #4409 which is closed.  We can reopen it after #4409 is completed.

PR #4200, opened 2021-01-25
In Maven central, remove checker-qual from checker artifact. 

smillst:
Also, add a jar with classifier "all" to the checker artifact that contains all required dependencies.


checker/build.gradle
```diff
@@ -119,21 +116,53 @@ task allJavadocJar(type: Jar) {
...
+        attributes("Main-Class": "org.checkerframework.framework.util.CheckerMain")
+    }
+    exclude 'org/checkerframework/**/qual/*'
+    exclude 'org/checkerframework/checker/*/util/*'
+    relocators = shadowJar.getRelocators()
+}
+
+jar {
+    dependsOn(checkerJar)
+    // Never build the skinny jar.
```
mernst, line 142: What is the point of this `jar` block, given that it contains `onlyIf {false}`?

mernst (approved):

checker/build.gradle
smillst: `jar` is a task automatically added by the Java plugin.  Other tasks automatically added by the Java plugin use this task and assumes it builds the main jar for a project.  So, this block adds a dependency to the task that actually builds the main jar and prevents it from build the standard jar.

mernst: Could you write some of this in the `build.gradle` file?  I was asking for future me and other maintainers, who won't have this insight.  Thanks.

PR #4201, opened 2021-01-25
Relocate asm dependency.

smillst:
We rename all third party dependencies so they do not conflict with other versions on the classpath.


mernst:
Thanks for noticing and fixing this.

PR #4202, opened 2021-01-25
Better way to reference checker shadow jar.

smillst:


smillst:
@mernst Can you see if this fixes your javadoc problem?

PR #4204, opened 2021-01-26
Document methods for which a JDK 11 version exists

mernst:
Merge together with https://github.com/typetools/jdk/pull/97


smillst (approved):

smillst:
If you want to test the Java 11 methods, you can add a test class with a `\\ @below-java11-jdk-skip-test` comment. (I don't think this is necessary.)

PR #4213, opened 2021-01-29
Add checker-qual to javac script

kelloggm:
I think this is related to removing the checker-qual classes from checker.jar.


kelloggm:
I noticed this while testing a single file locally with the Index Checker - it issued an error about missing annotations:
```
error: AnnotationBuilder: error: fromClass can't load Class org.checkerframework.checker.index.qual.SubstringIndexUnknown
  ensure the class is on the compilation classpath
1 error
```

I verified locally that the error isn't issued anymore after this change.

smillst (approved):
Thanks!  C

PR #4221, opened 2021-02-02
Class literals require raw types

mernst:


dataflow/src/main/java/org/checkerframework/dataflow/expression/ClassName.java
```diff
@@ -33,6 +33,9 @@ public int hashCode() {
 
     @Override
     public String toString() {
+        if (typeString.endsWith(">")) {
```
smillst, line 36: I think it would be better to change `typeString` to be this substring instead.  That way `equals` would use the same string.
(You could store the erased type instead using javax.lang.model.util.Types#erasure, but you would have to pass the `Types` object and I don't think that's necessary.)

smillst (approved):

PR #4224, opened 2021-02-02
Tweak variable names and documentation

mernst:


javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -1222,7 +1222,12 @@ public static boolean isLocalVariable(Tree tree) {
      * Returns the type as a TypeMirror of {@code tree}. To obtain {@code tree}'s
      * AnnotatedTypeMirror, call {@code AnnotatedTypeFactory.getAnnotatedType()}.
      *
-     * @return the type as a TypeMirror of {@code tree}
+     * <p>This may return null if the tree occurs in a compilation unit earlier than the tree's type
```
smillst, line 1225: I'm a bit surprised by this.  The framework calls this method on most (all?) trees and assumes a non-null result.  Do you have a test case that show this?

mernst: Yes, I encountered null values (and crashes) that were a result of this problem.  It will take a bit of effort to dig them up, though.

mernst: I'm undoing this change from this pull request, to unblock this pull request.  I'll put it in a later pull request.

smillst (approved):

PR #4225, opened 2021-02-02
Add warning to documentation of TreeUtils.typeOf

mernst:


javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -1222,9 +1222,14 @@ public static boolean isLocalVariable(Tree tree) {
      * Returns the type as a TypeMirror of {@code tree}. To obtain {@code tree}'s
      * AnnotatedTypeMirror, call {@code AnnotatedTypeFactory.getAnnotatedType()}.
      *
-     * @return the type as a TypeMirror of {@code tree}
+     * <p>This may return null if the tree occurs in a compilation unit earlier than the tree's type
```
smillst, line 1225: I'm wondering if this is happens when the tree was accessed outside of its compilation unit rather than just earlier in the same compilation unit.  I've had problems with null fields in tree objects in this case.

PR #4229, opened 2021-02-02
Update URLs to use Java 11 API documentation

mernst:


build.gradle
```diff
@@ -158,11 +158,7 @@ allprojects {
                 options.taglets 'org.checkerframework.taglet.ManualTaglet'
                 options.tagletPath(project(':framework-test').sourceSets."${tagletVersion}".output.classesDirs.getFiles() as File[])
 
-                // We want to link to Java 9 documentation of the compiler classes since we use Java 9
-                // versions of those classes and Java 8 for everything else.  Because the compiler classes are not
-                // a part of the main documentation of Java 8, javadoc links to the Java 9 versions.
-                // TODO, this creates broken links to the com.sun.tools.javac package.
-                options.links = ['https://docs.oracle.com/javase/8/docs/api/', 'https://docs.oracle.com/javase/9/docs/api/']
+                options.links = ['https://docs.oracle.com/en/java/javase/11/docs/api/']
```
smillst, line 161: It seems like this change should be reverted.  The framework still uses Java 9 compiler classes not Java 11.

smillst (changes requested):

build.gradle
mernst: When running under JDK 8, you are right that the framework uses Java 9 compiler classes.
When running under JDK 11, I think the framework uses Java 11 everything.  Am I mistaken?

Also, is there any difference between Java 9 and Java 11 compiler classes?  There were significant changes between Java 8 and Java 9, I know.

smillst: Here are two concrete problems with this change:

1.  This change makes it so that there are not links to some compiler class.  For example, `TypeMirror`, compare https://checkerframework.org/api/org/checkerframework/javacutil/TypesUtils.html to the version built with this branch.
2. When I run `./gradlew allJavadoc` with JDK11 I get the following error:
```
javadoc: error - The code being documented uses packages in the unnamed module, but the packages define     d in https://docs.oracle.com/en/java/javase/11/docs/api/ are in named modules.
```

I don't remember what happens when the Checker Framework is run on JDK11.  The Checker Framework class files are always Java 8 class files, and I seem to remember that changes which compiler classes are used.

smillst (approved):

PR #4230, opened 2021-02-02
Move tests of `hasDependentAnnotations()` within DependentTypesHelper

mernst:


framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -370,9 +370,7 @@ public void processClassTree(ClassTree classTree) {
         }
 
         AnnotatedDeclaredType classType = atypeFactory.getAnnotatedType(classTree);
-        if (atypeFactory.getDependentTypesHelper() != null) {
-            atypeFactory.getDependentTypesHelper().checkClass(classTree, classType);
-        }
+        atypeFactory.getDependentTypesHelper().checkClass(classTree, classType);
```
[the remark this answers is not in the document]
mernst: Yes, part of the changes were missing.  Sorry about that.  I have now added them.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -1910,18 +1903,14 @@ public void addComputedTypeAnnotations(Element elt, AnnotatedTypeMirror type) {
...
     }
 
     @Override
     public ParameterizedExecutableType methodFromUse(MethodInvocationTree tree) {
         ParameterizedExecutableType mType = super.methodFromUse(tree);
         AnnotatedExecutableType method = mType.executableType;
-        if (dependentTypesHelper != null) {
-            dependentTypesHelper.viewpointAdaptMethod(tree, method);
-        }
+        dependentTypesHelper.viewpointAdaptMethod(tree, method);
```
smillst, line 1913: `#viewpointAdaptExecutable` doesn't call `#hasDependentAnnotations`  So now it is called even when the checker does not have any dependent type annotations.   Should `#viewpointAdaptExecutable` call  `#hasDependentAnnotations` ?  By making `dependentTypesHelper` non null for all checkers, there are several methods like this that are now called for every checker. 

How did you choose which methods have this check?  It seems to be most public methods, but not all.  Should it be all public methods?

smillst (changes requested):

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
mernst: It's all the methods where there is currently a check; I moved the checks but did not introduce new ones.  The main reason was to avoid behavior change.  The minor reason was to avoid unnecessary checks.  I can add the checks to all the public methods.

smillst: The previous design was that `dependentTypesHelper` is never used if the checker has no dependent type annotations.  The new design seems to be that methods in the `dependentTypesHelper` are called even if the checker has no dependent type, but the method immediately returns.  So it would make sense if all public methods do this check.  Otherwise, it would be easy to add a new call to a method that doesn't make this check.

smillst (approved):

PR #4237, opened 2021-02-04
Use TypeMirrors instead of AnnotatedTypeMirrors for constructing type trees for foreach loops

kelloggm:
This avoids the dependency in `CFCFGBuilder`'s construction of these trees on a particular annotated type factory.

Merge before #4189.


mernst (approved):

mernst:
@smillst Colud you take a quick look then merge or reassign?
The diffs are difficult to read because of variable renaming, but the underlying changes are small and look reasonable to me.

framework/src/main/java/org/checkerframework/framework/flow/CFCFGBuilder.java
```diff
@@ -159,27 +153,29 @@ protected VariableTree createEnhancedForLoopIteratorVariable(
...
+            // It is necessary to get the elt because just getting the type of expression
+            // directly (via TreeUtils.typeOf) doesn't include annotations on the declarations
+            // of local variables, for some reason.
+            Element elt = TreeUtils.elementFromTree(expression);
+            TypeMirror type = null;
+            if (elt != null) {
+                type = ElementUtils.getType(elt);
+            }
+
+            // But if the declaration is a generic, such as in tests/all-systems/Issue1775.java,
```
smillst, line 165: The correct path is `framework/tests/all-systems/java8inference/Issue1775.java`

framework/src/main/java/org/checkerframework/framework/flow/CFCFGBuilder.java
```diff
@@ -159,27 +153,29 @@ protected VariableTree createEnhancedForLoopIteratorVariable(
...
-                newArrayType.setComponentType(annotatedArrayType);
-                newArrayType.addAnnotations(annotatedArrayType.getEffectiveAnnotations());
-                annotatedArrayType = newArrayType;
+            // It is necessary to get the elt because just getting the type of expression
+            // directly (via TreeUtils.typeOf) doesn't include annotations on the declarations
+            // of local variables, for some reason.
+            Element elt = TreeUtils.elementFromTree(expression);
+            TypeMirror type = null;
+            if (elt != null) {
+                type = ElementUtils.getType(elt);
```
smillst, line 162: This isn't the correct type.  It hasn't been view-point adapted to the use of the element.  If it's only local variables whose annotations are missing, then you can check if expression is a use of a local variable and use the element type in that case.

smillst (changes requested):

framework/src/main/java/org/checkerframework/framework/flow/CFCFGBuilder.java
kelloggm: > It hasn't been view-point adapted to the use of the element.

Yes, I'm aware. We had to make a [small change in Daikon](https://github.com/codespecs/daikon/pull/312) to accommodate this change, but Mike and I decided that it was an acceptable price to pay. #4189 requires this code not to rely on a particular annotated type factory being available, which is why this can't be view-point adapated.

That said, I've changed this code so that this path is only taken for local variables.

framework/src/main/java/org/checkerframework/framework/flow/CFCFGBuilder.java
```diff
@@ -159,27 +153,33 @@ protected VariableTree createEnhancedForLoopIteratorVariable(
...
+                // It is necessary to get the elt because just getting the type of expression
+                // directly (via TreeUtils.typeOf) doesn't include annotations on the declarations
+                // of local variables, for some reason.
+                Element elt = TreeUtils.elementFromTree(expression);
+                if (elt != null) {
+                    type = ElementUtils.getType(elt);
+                }
+            }
+
+            // But if the declaration is a generic, such as in
```
smillst, line 168: This comment does make sense given the change above.

smillst (approved):

wmdietl:
@kelloggm This change makes the handling of foreach loops unsound.
The AnnotatedTypeFactory uses additional information to determine AnnotatedTypeMirrors. By removing that dependency, incorrect types are used.
A simple, self-contained example for the Nullness Checker is:

````java
import org.checkerframework.framework.qual.DefaultQualifier;

class YourClassNameHere {
  @DefaultQualifier(Nullable.class)
  Object @NonNull [] foo() { return new Object[] {null}; }

  void bar() {
    for (Object p : foo()) {
      p.toString();
    }
  }
}
````

This example uses defaults to make the component type of the return type of `foo` `@Nullable Object`.
By no longer using the AnnotatedTypeFactory, the above example no longer raises an error for the dereference of `p`.

I expect that similarly annotations from stub files are ignored, but didn't try that. Basically, if the AnnotatedTypeFactory could change the component type of an array or type argument of a container, and that type is used in an foreach loop, the behavior changed.

kelloggm:
@wmdietl thanks for noticing this problem that we had overlooked.

@smillst @mernst, should I create a PR to revert this change while we investigate the problem?

smillst:
@kelloggm I opened an issue for this https://github.com/typetools/checker-framework/issues/4248 with a possible fix.  I don't think you need to revert this if you can get the fix in soon.

kelloggm:
@smillst thanks for doing that, and thanks for the pointer to the existing fix for iterators. I'll try implementing the fix you suggested.

mernst:
@wmdietl Thanks for pointing out this problem.

PR #4238, opened 2021-02-04
Correct Javadoc of GenericAnnotatedTypeFactory#parseJavaExpressionString

smillst:


framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -861,10 +861,11 @@ public AnnotationMirror getAnnotationFromJavaExpression(
     }
 
     /**
-     * Produces the JavaExpression associated with expression on currentPath.
+     * Converts the given {@code expression} string to a {@link JavaExpression} by resolving
+     * identifiers as if they were written in Java code that {@code currentPath}.
```
mernst, line 865: This is not grammatical.  Should "that" be "at"?

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -861,10 +861,11 @@ public AnnotationMirror getAnnotationFromJavaExpression(
...
      *
      * @param expression a Java expression
-     * @param currentPath the path to an annotation containing {@code expression}
+     * @param currentPath the path to use to resolve identifiers
```
mernst, line 868: What does it mean to resolve an identifier?  It seems to be a different concept than the "resolve" in `resolveMethods`.

mernst (changes requested):

smillst:
I'm going to fix this Javadoc when removing `useLocalScope`, so I'm closing this.

PR #4239, opened 2021-02-06
Improve error messages relating to invalid overrides

mernst:
This pull request also:
 * makes variable names more consistent.
 * improves efficiency by computing strings lazily, if an error message must be produced, rather than eagerly.
 * checks preconditions before postconditions rather than in the midst of postconditions.


framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -3584,21 +3584,24 @@ private boolean checkMethodReferenceInference(
...
-        protected final String overriderMeth;
-        protected final String overriderTyp;
-        protected final String overriddenMeth;
-        protected final String overriddenTyp;
 
+        /** The declaration of an overriding method. */
         protected final Tree overriderTree;
-        protected final Boolean methodReference;
+        /** True if {@link #overriderTree} is a MEMBER_REFERENCE. */
+        protected final boolean isMethodReference;
```
smillst, line 3591: The renames of the protect fields should be in the change log.

smillst (approved):

PR #4243, opened 2021-02-07
Rename the `known.nonnull` error message key to `nulltest.redundant`

mernst:
Remove file `SKIP-REQUIRE-JAVADOC` after merging.


changelog.txt
```diff
@@ -21,6 +21,8 @@ by new message keys:
  * `expression.parameter.name.invalid`
  * `expression.parameter.name.shadows.field`
 
+Renamed the Nullness Checker's `known.nonnull` error message key to `nulltest.redundant`.
```
smillst, line 24: This is in the wrong section of the change log.

smillst (approved):

changelog.txt
mernst: Where should it be?  I put it in the "User-visible changes" section because error message keys are user-visible:  users may have to adjust `@SuppressWarnings` annotations.

smillst: I mean it should be in the change log for the next release not the last release.

PR #4246, opened 2021-02-08
Warnings about nullness annotations on receiver and exception parameters

mernst:


checker/tests/nullness/ExceptionParam.java
```diff
@@ -14,14 +14,15 @@ void exc1() {
...
+            // :: warning: (nullness.on.exception.parameter)
         } catch (@NonNull AssertionError e) {
             @NonNull Object o = e;
         }
     }
 
     void exc3() {
         try {
-            // :: error: (type.invalid.annotations.on.use)
+            // :: warning: (nullness.on.exception.parameter)
```
smillst, line 25: Should this be an error?  The other `nullness.on.*` messages are reported as errors.

smillst (approved):

checker/tests/nullness/ExceptionParam.java
mernst: This one is intentionally a warning, but the explanation deserves a comment.  I will add one.

PR #4247, opened 2021-02-08
Forbid nullness annotations on enum constant definitions

mernst:


checker/src/main/java/org/checkerframework/checker/nullness/NullnessVisitor.java
```diff
@@ -516,6 +519,26 @@ public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
         return null;
     }
 
+    @Override
+    public void processClassTree(ClassTree classTree) {
+        if (classTree.getKind() == Tree.Kind.ENUM) {
+            for (Tree member : classTree.getMembers()) {
+                if (member.getKind() == Tree.Kind.VARIABLE
+                        && (((JCVariableDecl) member).mods.flags & Flags.ENUM) != 0) {
```
smillst, line 527: Because this is using a compiler API that is not supported and many change, it should be a method in TreeUtils.  Alternatively, you could use `TreeUtils.elementFromDeclaration(...).getKind() == ElementKind.ENUM_CONSTANT`.

smillst (approved):

PR #4251, opened 2021-02-08
Treat array accesses from enhanced for loops the same way we treat iterators

### LINKED ISSUE #4248: Enhanced for loops: arrays as iterator expression.
```
import java.util.List;
import org.checkerframework.framework.qual.DefaultQualifier;

class YourClassNameHere {
    @DefaultQualifier(Nullable.class)
    Object @NonNull [] foo() {
        return new Object[] {null};
    }

    void bar() {
        for (Object p : foo()) {
            // :: error: (dereference.of.nullable)
            p.toString();
        }
    }

    @DefaultQualifier(Nullable.class)
    @NonNull List<Object> foo2() {
        throw new RuntimeException();
    }

    void bar2() {
        for (Object p : foo2()) {
            // :: error: (dereference.of.nullable)
            p.toString();
        }
    }
}
```

The first expect error is missing.  This is a regression added by https://github.com/typetools/checker-framework/pull/4237 and noticed by @wmdietl https://github.com/typetools/checker-framework/pull/4237#issuecomment-774745002.  The `List` version works because of these changes a20abff8756fc433787ffc1342b14b791b14ff9c.  A similar change for arrays should fix this bug.

kelloggm:
This fix should have been included in a20abff8756fc433787ffc1342b14b791b14ff9c.
Fixes #4248


smillst (approved):
Thanks!

kelloggm:
@smillst, CI was failing because of a crash while typechecking Daikon. I've fixed the crash and added tests that demonstrate the fix works. I'd appreciate a re-review. A description of the issue and fix follows.

The issue was that arrays of primitives can be converted to their boxed forms, which introduces a generated call to `$BoxedPrimitive#valueOf` between the array access and the assignment. My original code blindly casted the rhs of the assignment node to an array access, which caused the crash.

I'm surprised that to get the Index test here working, I had to use the method invocation node at the top level of the RHS. That kind of makes sense, especially since this didn't come up when you were testing the Iterator version of this fix - both calls to the iterator and these boxing conversations are method invocation nodes, so the same code worked for both.

smillst (approved):
I think this pull request is ok. For some checkers, conversions like boxing or widening change the annotations on the types.  Those changes will be missed for enhanced for loops.  But that was a problem before this pull request in the BaseTypeVisitor.  This pull request just makes the annotations used by dataflow  the same as the annotations used to check the enhanced for loop in BaseTypeVisitor.

Issue #4252, opened 2021-02-08
AnnotatedTypeScanners can use a lot of memory


smillst:
See #3851 for an example.  We need to:
- [ ] Document that AnnotatedTypeScanners should be stored in fields.
- [ ] Check for more AnnotatedTypeScanners that need to be moved to fields.

PR #4253, opened 2021-02-08
Add fields for Scanners in DependentTypesHelper

smillst:
Scanners have HashMaps that can use a lot of memory, so it's better to use fields rather than creating a new object for each use.   #3851  shows how a scanner caused problems else where.  (https://github.com/typetools/checker-framework/issues/4252 is a new issue to document this problem and find more scanners to move to fields.)

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
+++ b/framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
@@ -95,14 +95,14 @@ public class DependentTypesHelper {
 
     /**
      * A scanner that standardizes Java expression strings in dependent type annotations. Call
-     * {@link StandardizeTypeAnnotator#init(JavaExpressionContext, TreePath, boolean, boolean)}
-     * before using.
+     * {@code StandardizeTypeAnnotator#init(JavaExpressionContext, TreePath, boolean, boolean)}
+     * before each use.
      */
     private final StandardizeTypeAnnotator standardizeTypeAnnotator;
 
     /**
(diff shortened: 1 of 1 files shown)


framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -87,6 +87,30 @@
...
+    /**
+     * The {@code ExpressionErrorChecker} that scans an annotated type and returns a list of {@link
+     * DependentTypesError}.
+     */
+    private final ExpressionErrorChecker expressionErrorChecker;
+
+    /**
+     * A scanner that standardizes Java expression strings in dependent type annotations. Call
+     * {@link StandardizeTypeAnnotator#init(JavaExpressionContext, TreePath, boolean, boolean)}
+     * before using.
```
mernst, line 99: Please clarify whether this means to call `init` once, or to call `init` before each use.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -87,6 +87,30 @@
...
+     */
+    private final StandardizeTypeAnnotator standardizeTypeAnnotator;
+
+    /**
+     * Copies annotations that might have been viewpoint adapted from the visited type (the first
+     * formal parameter of the visit method) to the second formal parameter.
```
mernst, line 105: Is this the `visit` method of this class, or of the field?

mernst (approved):
Thanks for these efficiency improvements.  I have just two questions about documentation.

Issue #862, opened 2016-07-25
Flow Expressions:  Class names should be resolved based on the declaration scope


smillst:
```
package pkg1;
import org.checkerframework.checker.nullness.qual.*;
class Class1 {
  public static @NonNull Object field;
}

package pkg2;
import org.checkerframework.checker.nullness.qual.*;
import pkg1.Class1;
class Class2 {
  @RequiresNonNull("Class1.field")
  public void foo() {}

package use;
import pkg2.Class2;
class Use {
    void use(Class2 class2) {
      //:: error: (contracts.precondition.not.satisfied)
      class2.foo();
    }
}
```

The expected error in class Use is not issued; instead, the following error is issued:

```
Class3.java:4: error: [flowexpr.parse.error] cannot parse the expression 'Class1: identifier not found'
       class2.foo();
                 ^
```

Revert the part of this comment that suppresses warnings based on this issue:
https://github.com/codespecs/daikon/commit/437ca4a30c8fd8fa2489e26ae38253117640ad1c

smillst:
Testcase is in framework/tests/flow2/flowexpression-scope.  You can run it like so:

cd $CHECKERFRAMEWORK/framework/tests/flow2/flowexpression-scope
ant -e -find build.xml build-tests
javacheck -cp ../../build/ -processor tests.util.FlowTestChecker *.java

smillst:
This bug can be worked around by importing the class in the flow expression at the use.  In the example in https://github.com/typetools/checker-framework/issues/862#issue-167480443, import Class1 in Class3.java will cause the expected error to be issued.

mernst:
Maybe we can make the error message suggest importing (though we don't know whether an expression a.b is just a typo or is a reference to an un-imported class a).

jack17529:
@mernst anybody working on this , because I want to work.

mernst:
I'm not aware of anyone working on it.  If you want to work on a problem (especially an old one that doesn't have any recent comments), then it's best to make a pull request.  We generally don't "reserve" a bug report for people who don't have a record of contributions, because too often they have not followed through.  I hope that you will, though, and I look forward to your patch.

jack17529:
@mernst as you say sir.

wmdietlGC:
Let me expand on the "Class names should be resolved based on the declaration scope" to see whether we all have the same understanding of what the right fix is.
[...]

PR #4261, opened 2021-02-09
Refactor the JavaExpression parse method to clarify `localScope`

smillst:
Currently `localScope` is used both 
1. when parsing identifiers to local variables (if `useLocalScope` is true) and 
2. when parsing to things that are from elements and only need a TreePath because of the javac API the framework uses.  

I propose we split the two uses into two different variables named
1. `localPath` 
2. `somePath`
(I'm open to suggestions on other names, too)

In this pull request (that shall not be merged), I've made changes to `JavaExpressionParseUtil#parse` to show what the Javadoc will be.  @mernst, I wanted your feedback on this design before making similar changes at every use of `localScope` and `useLocalScope`.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
+++ b/framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
@@ -141,8 +141,8 @@ public class JavaExpressionParseUtil {
             String expression, JavaExpressionContext context, @Nullable TreePath localPath)
             throws JavaExpressionParseException {
         // The underlying javac API used to convert from Strings to Elements requires a tree path
-        // even when the information could be deduced from elements alone.  So use the
-        // path to the current CompilationUnit.
+        // even when the information could be deduced from elements alone.  So use the path to the
+        // current CompilationUnit.
         TreePath pathToCompilationUnit = context.checker.getPathToCompilationUnit();
         Expression expr;
         try {
--- a/framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
+++ b/framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
@@ -123,29 +123,27 @@ public class JavaExpressionParseUtil {
      * Return its representation as a {@link JavaExpression}, or throw a {@link
      * JavaExpressionParseException}.
      *
-     * <p>A {@link TreePath}, {@code somePath}, is required to use the underlying javac API to
-     * convert from Strings to {@link Element}s even when the information could be deduced from
-     * elements alone.
-     *
-     * <p>If {@code localPath} is nonnull, then identifiers may be parsed to local variables in
-     * scope at {@code localPath}. In this case, it is as if the identifier was written at the
-     * location of {@code localPath}. If {@code localPath} is null, then no identifier can be parsed
-     * to a local variable.
+     * <p>If {@code localPath} is nonnull, then identifiers are parsed as if the identifier was
+     * written at the location of {@code localPath}. This means identifiers will be parsed to local
+     * variables in scope at {@code localPath} when possible. If {@code localPath} is null, then no
+     * identifier can be parsed to a local variable. In either case, the parameter syntax, e.g. #1,
+     * is always parsed to the arguments in {@code context}. This is because a parameter of a lambda
+     * can refer both to local variables in scope at its declaration and to a parameter of the
+     * lambda.
      *
      * @param expression a Java expression to parse
      * @param context information about any receiver and arguments
-     * @param somePath path to use javac API
      * @param localPath if non-null, the location at which to parse identifiers to local variables
      * @return the JavaExpression for the given string
      * @throws JavaExpressionParseException if the string cannot be parsed
      */
     public static JavaExpression parse(
-            String expression,
-            JavaExpressionContext context,
-            TreePath somePath,
-            @Nullable TreePath localPath)
+            String expression, JavaExpressionContext context, @Nullable TreePath localPath)
             throws JavaExpressionParseException {
-
+        // A TreePath is required to use the underlying javac API to convert from Strings to
+        // Elements even when the information could be deduced from elements alone.  So use the the
+        // path to the current CompilationUnit.
+        TreePath pathToCompilationUnit = context.checker.getPathToCompilationUnit();
         Expression expr;
         try {
             expr = StaticJavaParser.parseExpression(replaceParameterSyntax(expression));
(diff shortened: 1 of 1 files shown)


framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -119,22 +119,31 @@
     private static final int PARAMETER_REPLACEMENT_LENGTH = PARAMETER_REPLACEMENT.length();
 
     /**
-     * Parse a string and return its representation as a {@link JavaExpression}, or throw a {@link
+     * Parse a string and view-point adapt it to the given {@code context} and {@code localPath}.
```
smillst, line 122: Another possible design, would be to make `localPath` a field of `JavaExpressionContext`.

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -119,22 +119,31 @@
...
+     * Return its representation as a {@link JavaExpression}, or throw a {@link
      * JavaExpressionParseException}.
      *
+     * <p>A {@link TreePath}, {@code somePath}, is required to use the underlying javac API to
```
mernst, line 126: What parts of `somePath` are used?  Do paths A and B always have the identical effect, so long as A and B are in the same compilation unit?  Does this path need to be in the same compilation unit as `localPath`?  If the compilation unit doesn't matter, then does this need to be passed in at all?  Could it be set as a field in `JavaExpressionParseUtil`?

I don't find `somePath` evocative.  If the compilation unit is what matters, maybe the name could reflect that, such as `pathWithinCompilationUnit`.  (I admit that is long and clunky; maybe it can be further improved.)

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -119,22 +119,31 @@
...
+     * convert from Strings to {@link Element}s even when the information could be deduced from
+     * elements alone.
+     *
+     * <p>If {@code localPath} is nonnull, then identifiers may be parsed to local variables in
```
mernst, line 130: I think that "may be" should be "are" (or "are ... whenever possible").
Actually, the next sentence about "as if written at that location" is even clearer and could come first, with this just as a clarifying comment.

If `localPath` is non-null, then is "#2" forbidden?

mernst (changes requested):
Thanks.  This is much clearer.  I still have a few questions.

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
smillst: My hypothesis is that when JavaExpressionParseUtil uses `somePath` rather than `localPath`, passing any path should have identical effect.  To test this, I passed `SourceChecker#currentRoot` when `useLocalScope` is false.  When I run the tests, the only errors are related to https://github.com/typetools/checker-framework/issues/862.  So, I think if I fix that bug by forcing class names to be fully-qualified, then `somePath` can just be `SourceChecker#currentRoot`.  (I'm working on that fix now.)

smillst: After #4266, is merged, `somePath` can be `SourceChecker#currentRoot`.  The last commit makes this change.

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
smillst: > If localPath is non-null, then is "#2" forbidden?
No, I explain why in the last commit.

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -119,24 +119,31 @@
...
+     * is always parsed to the arguments in {@code context}. This is because a parameter of a lambda
+     * can refer both to local variables in scope at its declaration and to a parameter of the
+     * lambda.
+     *
      * @param expression a Java expression to parse
      * @param context information about any receiver and arguments
-     * @param localScope a program element annotated with an annotation that contains {@code
-     *     expression}
-     * @param useLocalScope whether {@code annotatedConstruct} should be used to resolve identifiers
+     * @param localPath if non-null, the location at which to parse identifiers to local variables
```
mernst, line 136: I would add "the expression is parsed as if it were written at this location" either before, after, or in place of the current `@param` documentation.

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
mernst: Nit: I would spell it as "viewpoint-adapt" rather than "view-point adapt".  Viewpoint is one word, and viewpoint-adapt is a compound verb (people differ on whether to hyphenate them; I do).

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -119,24 +119,31 @@
     private static final int PARAMETER_REPLACEMENT_LENGTH = PARAMETER_REPLACEMENT.length();
 
     /**
-     * Parse a string and return its representation as a {@link JavaExpression}, or throw a {@link
+     * Parse a string and view-point adapt it to the given {@code context} and {@code localPath}.
+     * Return its representation as a {@link JavaExpression}, or throw a {@link
      * JavaExpressionParseException}.
      *
+     * <p>If {@code localPath} is nonnull, then identifiers are parsed as if the identifier was
```
mernst, line 126: Nit: "nonnull" => "non-null".

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -119,24 +119,31 @@
...
      */
     public static JavaExpression parse(
-            String expression,
-            JavaExpressionContext context,
-            TreePath localScope,
-            boolean useLocalScope)
+            String expression, JavaExpressionContext context, @Nullable TreePath localPath)
             throws JavaExpressionParseException {
-
+        // A TreePath is required to use the underlying javac API to convert from Strings to
```
mernst, line 143: Nit "required to use" leaves the actor ambiguous.  I would change the text to "The underlying javac API requires ...".

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -119,24 +119,31 @@
...
+        // Elements even when the information could be deduced from elements alone.  So use the the
```
mernst, line 144: Typo "the the"

mernst (approved):

smillst:
Thanks for the last round of feedback!  I made the changes you suggested.  I'm closing this pull request and will open a new when the code changes are complete.

PR #4266, opened 2021-02-10
JavaExpressionParseUtil: Parse unqualified class names correctly.

### LINKED ISSUE #862: Flow Expressions:  Class names should be resolved based on the declaration scope

smillst:
See manual change for explantation. I also moved the code that parses identifiers to fields to a new method, so you'll see those changes, too.
Fixes #862. 

Merge with https://github.com/codespecs/daikon/pull/316

### DIFF
--- a/javacutil/src/main/java/org/checkerframework/javacutil/Resolver.java
+++ b/javacutil/src/main/java/org/checkerframework/javacutil/Resolver.java
@@ -263,6 +263,37 @@ public class Resolver {
     }
 
     /**
+     * Finds the class with {@code name} in a given type.
+     *
+     * @param name the name of the class
+     * @param type the type
+     * @param path the tree path
+     * @return the {@code ClassSymbol} for the class if it is found, {@code null} otherwise
+     */
+    public @Nullable ClassSymbol findClassInType(String name, TypeMirror type, TreePath path) {
+        Log.DiagnosticHandler discardDiagnosticHandler = new Log.DiscardDiagnosticHandler(log);
+        try {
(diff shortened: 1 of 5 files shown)


javacutil/src/main/java/org/checkerframework/javacutil/Resolver.java
```diff
@@ -262,6 +262,36 @@ public Element findClass(String name, TreePath path) {
         }
     }
 
+    /**
+     * Finds the class with {@code name} in a given type.
```
mernst, line 266: Does "in a given type" mean this is looking for a nested class?  Saying that (and even adding "Nested" to the method name) would be clearer, I think.

mernst (approved):
Looks good, thanks.

I tweaked the documentation in the manual.  If what I wrote is not correct, can we adjust the code to conform to the manual?

PR #4268, opened 2021-02-11
Moved JavaExpressionParseUtil#fromVariableTree to JavaExpression.

smillst:
Also, create the JavaExpression without parsing.

### DIFF
--- a/dataflow/src/main/java/org/checkerframework/dataflow/expression/JavaExpression.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/expression/JavaExpression.java
@@ -531,6 +531,7 @@ public abstract class JavaExpression {
             case PARAMETER:
                 return new LocalVariable(ele);
             case FIELD:
+            case ENUM_CONSTANT:
                 // Implicit access expression, such as "this" or a class name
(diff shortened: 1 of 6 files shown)


dataflow/src/main/java/org/checkerframework/dataflow/expression/JavaExpression.java
```diff
@@ -531,6 +506,50 @@ public static JavaExpression fromTree(
...
+     *
+     * @param tree a variable tree
+     * @return a JavaExpression for {@code tree}
+     */
+    public static @Nullable JavaExpression fromVariableTree(VariableTree tree) {
+        return fromVariableElement(TreeUtils.typeOf(tree), TreeUtils.elementFromDeclaration(tree));
+    }
+
+    /**
+     * Returns the Java expression corresponding to the given variable element {@code ele}.
```
mernst, line 520: Under what circumstances does it return `null`?
Could you change the `return null` clause to throw an exception?

mernst (approved):
I wonder if you can eliminate the possibility of `fromVariableElement` returning null.

PR #4270, opened 2021-02-11
Add a delocalize method to DependentTypesHelper.

smillst:
Removed methods that were only called once or called with `removeErroneousExpressions` = `false`.

I also removed `standardizeForMethodSignature`.  We might want this method again in the future, but currently it's only called in two places and doesn't reduce code duplication.  (It's javadoc was also wrong; it wasn't called for contracts.)

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
+++ b/framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
@@ -420,10 +420,11 @@ public class DependentTypesHelper {
 
     /**
      * Viewpoint-adapt all dependent type annotations to the method declaration, {@code
-     * methodDeclTree}.
+     * methodDeclTree}. This changes occurrences of formal parameter names to "#2" syntax, and it
+     * removes expressions that contain other local variables.
      *
      * @param methodDeclTree the method declaration to which the annotations are viewpoint-adapted
-     * @param atm type to viewpoint-adapt
+     * @param atm type to viewpoint-adapt; is side-effected by this method
      */
     public void delocalize(MethodTree methodDeclTree, AnnotatedTypeMirror atm) {
         if (!hasDependentType(atm)) {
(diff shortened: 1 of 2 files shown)


framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -429,44 +412,42 @@ public void standardizeReturnType(
...
+        if (!hasDependentType(atm)) {
             return;
         }
 
-        TypeMirror enclosingType = ElementUtils.enclosingTypeElement(elt).asType();
+        TreePath pathToMethodDecl = factory.getPath(methodDeclTree);
+        if (pathToMethodDecl == null) {
+            return;
+        }
+        // TODO: 1.) parameter names need to be covert to the # index syntax.
```
mernst: Do these TODOs mean that the method is not fully implemented?  (Maybe that will be done in a future pull request?)

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -429,44 +412,43 @@ public void standardizeReturnType(
...
+        if (!hasDependentType(atm)) {
             return;
         }
 
-        TypeMirror enclosingType = ElementUtils.enclosingTypeElement(elt).asType();
+        TreePath pathToMethodDecl = factory.getPath(methodDeclTree);
+        if (pathToMethodDecl == null) {
+            return;
+        }
+        // TODO: 1.) parameter names need to be coverted to the # index syntax.
```
mernst, line 438: Do these TODOs mean that the method is not fully implemented?  (Maybe that will be done in a future pull request?)  Or should the TODO comments be removed and the documentation be updated?

mernst (approved):

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
smillst: This method is not fully implemented. It will be implemented after other `useLocalScope` is removed.  I had to add this method now to simplify the code and make it easier to remove `useLocalScope`.

PR #4275, opened 2021-02-12
Fake overrides in stub files

mernst:


docs/manual/annotating-libraries.tex
```diff
@@ -510,6 +510,37 @@
...
+\subsectionAndLabel{Stub methods in subclasses of the declaring class}{stub-fake-override}
+
+Sometimes, a method's return type is different in a subclass than in a
+superclass, even though the subtype does not define the method and just
+inherits the superclass's definition.
+
+You may write a ``fake override'' in a stub file.  This is a declaration of
+a method that is inherited.  The Java signature of the method must be
+identical to that in the superclass where it is actually defined, but you
+can change its annotations.
```
smillst, line 522: If the "fake override" doesn't have any explicit annotations, is it as if the "fake override" wasn't written?

For example,
```
class Super{
  @Nullable String method() {... }
}

class Sub extends Super {}

class Use {
  void use(Sub sub) {
    @NonNull String s = sub.method();  // Is this legal?
  }
}
```
```
// In a stub file:
class Sub extends Super {
   String method(); 
}
```

framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileParser.java
```diff
@@ -1526,6 +1578,161 @@ private void putNewElement(
...
+     * @param decl a fake override of the element
+     * @param fakeLocation where the fake override was defined
+     */
+    private void processFakeOverride(
+            ExecutableElement element, CallableDeclaration<?> decl, TypeElement fakeLocation) {
+        // This is a fresh type, which this code may side-effect.
+        AnnotatedExecutableType methodType = atypeFactory.getAnnotatedType(element);
+
+        // TODO: Walk the type and the declaration, copying annotations from the declaration to the
+        // element.  I think PR #3799 has a visitor that does that, which I should use afte it is
```
smillst, line 1713: I don't think you mean https://github.com/typetools/checker-framework/pull/3799 here.

smillst (changes requested):

framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileParser.java
```diff
@@ -1526,6 +1598,162 @@ private void putNewElement(
         }
     }
 
+    /**
+     * Given a method declaration that does not correspond to an element, returns all the methods
+     * that it would override. This includes transitively overridden methods.
```
smillst, line 1603: Why does this include transitively overridden methods?

framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileElementTypes.java
```diff
@@ -313,6 +321,81 @@ public AnnotatedTypeMirror getAnnotatedTypeMirror(Element e) {
...
+                applicable.add(fakeLocation);
+            }
+        }
+
+        if (applicable.isEmpty()) {
+            return null;
+        }
+        TypeMirror fakeReceiverType =
+                TypesUtils.mostSpecific(applicable, factory.getProcessingEnv());
+        if (fakeReceiverType == null) {
```
smillst, line 377: I don't understand how there can be no "most specific" fake override.  Can you explain how this case occurs?

smillst (changes requested):
I haven't reviewed all the code, but I'm stuck on how there can be no "most specific" fake override.  (See https://github.com/typetools/checker-framework/pull/4275#discussion_r577152032.)

framework/src/main/java/org/checkerframework/framework/util/AnnotatedTypes.java
```diff
@@ -471,14 +472,56 @@ private static AnnotatedTypeMirror asMemberOfImpl(
                 }
                 return result;
             case UNION:
-            case DECLARED:
                 return substituteTypeVariables(
                         types, atypeFactory, receiverType, member, memberType);
+            case DECLARED:
+                AnnotatedTypeMirror memberTypeWithOverrides =
```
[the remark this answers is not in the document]
mernst: Thanks!  That is indeed a better place.

framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileElementTypes.java
mernst: I thought it was possible to write
```java
class A { public int m() { return 0; } }
interface I { int m(); }
class B extends A implements I { }
```
but that is not legal Java.  I will change the `UserError` to `BugInCF`.

smillst: `javac` compiles that code.

framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileElementTypes.java
```diff
@@ -313,6 +321,82 @@ public AnnotatedTypeMirror getAnnotatedTypeMirror(Element e) {
...
+            StringJoiner message = new StringJoiner(System.lineSeparator());
+            message.add(
+                    String.format(
+                            "No most specific fake override found for %s with receiver %s.  These fake overrides are applicable:",
```
smillst, line 382: I've come up with an example that triggers this Exception.  I added it as a test case in `/Users/smillst/jsr308/checker-framework/checker/jtreg/stubs/fakeoverrides`.
```java
interface SuperInterface {
    default int m() {  return 0; }
}

class SuperClass implements SuperInterface {
    // fake override:
    // int m();
}

interface SubInterface extends SuperInterface {
    // fake override:
    // int m();
}
public class Use extends SuperClass implements SubInterface {
    void use(Use d) {
        d.m();  // crash here.
    }
}
```

I don't think this code should crash.  The fake override in `SuperClass` should be used and the fake override in `SubInterface` should be ignored.  In the future, we should issue an error at the declaration of `Use`.

smillst (changes requested):
I think I've reviewed all the code now.  I just have one more question.

framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileParser.java
```diff
@@ -1323,6 +1355,8 @@ private void annotate(
             if (annoMirror != null) {
                 type.replaceAnnotation(annoMirror);
             } else {
+                // TODO: Maybe always warn here.  It's so easy to forget an import statement and
```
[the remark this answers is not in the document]
mernst: They are valid TODO items, though not directly related to the changes.  I wasted too much time debugging a problem when a stub file was missing an `import` statement and so the annotation was being ignored.
I could put them in a different pull request if you prefer.

smillst (approved):

PR #4276, opened 2021-02-12
Remove `useLocalScope` when parsing JavaExpressions.

smillst:
`JavaExpressionParseUtil#parse` no longer takes the boolean `useLocalScope`. Instead, if the JavaExpression should be parsed as if it were written at a given TreePath, then a non-null `localPath` is passed.  If `localPath` is not needed, then no path should be passed. (Internally, the JavaExpression parser uses the path to the current compilation unit to when calling javac APIs.). The changes to `JavaExpressionParseUtil#parse` rippled out to many other methods; that's why this pull request is rather large.

Also, added `viewpointAdaptQualifierFromContract` to `DependTypesHelper`.

I recommend reviewing the changes in the following order: `JavaExpressionParseUtil`, `DependentTypesHelper`, everything else.


framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -198,10 +221,13 @@ private static String replaceParameterSyntax(String expression) {
...
 
         /**
-         * The Java program element that is annotated by an annotation that contains the expression
-         * that is being translated.
+         * The underlying javac API used to convert from Strings to Elements requires a tree path
+         * even when the information could be deduced from elements alone. So use the path to the
+         * current CompilationUnit.
          */
-        private final TreePath annotatedConstruct;
+        private final TreePath somePath;
```
mernst, line 228: Throughout, I think `pathToCompilationUnit` would be a better name (assuming it is accurate).

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -215,11 +241,15 @@ private static String replaceParameterSyntax(String expression) {
         /**
          * Create a new ExpressionToJavaExpressionVisitor.
          *
-         * @param annotatedConstruct path to the expression
+         * @param somePath required to use the underlying Javac API
+         * @param localPath if non-null, the expression is parsed as if it were written at this
```
mernst, line 245: Throughout, would `localVarPath` or `localVariablePath` be a clearer name?

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -632,60 +625,65 @@ public void standardizeVariable(AnnotatedTypeMirror type, Element elt) {
...
+     * @param localPath the expression is parsed as if it were written at this location
+     * @param type the type to viewpoint-adapt; is side-effected by this method
+     */
+    private void viewpointAdaptToLocalPathAndContext(
+            JavaExpressionContext context, TreePath localPath, AnnotatedTypeMirror type) {
+        standardizeAtm(context, localPath, type, /*removeErroneousExpressions=*/ false);
     }
 
     /**
+     * "Standardize" atm
```
mernst, line 673: Please document (even if by cross-reference to other documentation).

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -120,24 +120,44 @@
...
+     * @param context information about any receiver and arguments
+     * @return the JavaExpression for the given string
+     * @throws JavaExpressionParseException if the string cannot be parsed
+     */
+    public static JavaExpression parse(String expression, JavaExpressionContext context)
+            throws JavaExpressionParseException {
+        return parse(expression, context, null);
+    }
+    /**
+     * Parse a string and viewpoint-adapt it to the given {@code context} and {@code localPath}.
```
mernst, line 136: I don't think that viewpoint-adaptation is with respect to a given path.  Should this be "parse with respect to localPath and viewpoint-adapt with respect to context"?

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -120,24 +120,44 @@
     private static final int PARAMETER_REPLACEMENT_LENGTH = PARAMETER_REPLACEMENT.length();
 
     /**
-     * Parse a string and return its representation as a {@link JavaExpression}, or throw a {@link
+     * Parse a string and viewpoint-adapt it to the given {@code context}. Return its representation
+     * as a {@link JavaExpression}, or throw a {@link JavaExpressionParseException}.
+     *
+     * @param expression a Java expression to parse
+     * @param context information about any receiver and arguments
```
mernst, line 127: This contains more than just information about the receiver and arguments.  It is also used to obtain the processing environment and the path to the current compilation unit (and maybe other things?).

mernst (changes requested):

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -632,60 +625,66 @@ public void standardizeVariable(AnnotatedTypeMirror type, Element elt) {
...
 
     /**
-     * Standardize a type, setting useLocalScope to true.
+     * Parse the dependent types in {@code type} as if they were written at {@code localVarPath}.
      *
-     * @param context the context
-     * @param localScope the local scope
-     * @param type the type to standardize; is side-effected by this method
+     * @param localVarPath the expression is parsed as if it were written at this location
+     * @param type the type to viewpoint-adapt; is side-effected by this method
```
mernst, line 631: This is a parsing method, not a viewpoint-adaptation method.

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractTransfer.java
```diff
@@ -601,26 +597,13 @@ protected void addInformationFromPreconditions(
      *
      * @param annoFromContract a contract annotation that was written on a method declaration
```
mernst, line 598: It is a contract annotation, but it wasn't necessarily written explicitly.  For example, it might be `@NonNull` when the user wrote `@EnsuresNonNull`.  (This comment also applies to other occurrences of this Javadoc.)

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -415,7 +401,36 @@ public void standardizeReturnType(MethodTree methodDeclTree, AnnotatedTypeMirror
...
                         methodDeclTree, pathToMethodDecl, factory.getChecker());
-        standardizeDoNotUseLocalScope(context, pathToMethodDecl, atm);
+        viewpointAdaptToContext(context, atm);
+    }
+
+    /**
+     * Viewpoint-adapt a type qualifier obtained from a contract to {@code jeContext}.
+     *
+     * @param annoFromContract a qualifier that was written on a method declaration
+     * @param jeContext the context to use for standardization
```
mernst, line 411: I thought contexts were used for viewpoint-adaptation (and this method is a viewpoint-adaptation method).

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -415,7 +401,36 @@ public void standardizeReturnType(MethodTree methodDeclTree, AnnotatedTypeMirror
...
+     * @param errorTree if non-null, where to report any errors that occur will parsing the
```
mernst, line 412: "will" => "when"

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -415,7 +401,36 @@ public void standardizeReturnType(MethodTree methodDeclTree, AnnotatedTypeMirror
...
+    public AnnotationMirror viewpointAdaptQualifierFromContract(
+            AnnotationMirror annoFromContract,
+            JavaExpressionContext jeContext,
+            @Nullable Tree errorTree) {
+        if (!hasDependentAnnotations()) {
+            return annoFromContract;
+        }
+
+        AnnotationMirror standardized =
+                standardizeAnnotationIfDependentType(jeContext, null, annoFromContract, false);
```
mernst, line 426: The documentation of `standardizeAnnotationIfDependentType` doesn't say anything about viewpoint-adaptation, and the method of this body doesn't seem to have any other place that viewpoint-adaptation could be happening.

mernst (approved):
Parts of the documentation are still a bit inconsistent, and I have not carefully examined all the logic.  I'm OK with merging this now and doing a more thorough review after the next rounds of refactoring.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
smillst: I corrected the documentation of `standardizeAnnotationIfDependentType`; it does viewpoint-adapt.   Almost all of occurrences of "standardized" need to be changed to "viewpoint-adpat" in this class.  I'm going to defer this until JavaExpressions are viewpoint-adapted rather than strings.

PR #4285, opened 2021-02-14
Check determinism after, not during, parsing a JavaExpression

mernst:


dataflow/src/main/java/org/checkerframework/dataflow/expression/JavaExpression.java
```diff
@@ -205,17 +225,17 @@ public String toStringDebug() {
      * Returns the internal representation (as {@link FieldAccess}) of a {@link FieldAccessNode}.
      * The result may contain {@link Unknown} as receiver.
      *
+     * @param node the FieldAccessNode to convert to a JavaExpression
      * @return the internal representation (as {@link FieldAccess}) of a {@link FieldAccessNode}.
      *     Can contain {@link Unknown} as receiver.
      */
-    public static FieldAccess fromNodeFieldAccess(
-            AnnotationProvider provider, FieldAccessNode node) {
+    public static FieldAccess fromNodeFieldAccess(FieldAccessNode node) {
```
smillst, line 232: All the public method signatures that have changed in this file should be listed in the change log.

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractStore.java
```diff
@@ -323,22 +341,64 @@ public void insertValue(JavaExpression expr, AnnotationMirror a) {
...
+    /**
+     * Helper function for {@link #insertOrRefine(JavaExpression, AnnotationMirror)} and {@link
+     * #insertOrRefinePermitNondeterministic}.
+     *
+     * @param expr an expression
+     * @param newAnno the expression's annotation
+     * @param permitNondeterministic whether nondeterministic expressions may be inserted into the
+     *     store
+     */
+    public void insertOrRefine(
```
smillst, line 376: Should this be protected?

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractStore.java
```diff
@@ -385,6 +487,24 @@ public void insertValue(JavaExpression expr, @Nullable V value) {
...
+            // called.)
+            return;
+        }
+        insertValueImpl(expr, value);
+    }
+
+    /**
+     * Helper method for {@link #insertValue(JavaExpression, CFAbstractValue)} and {@link
+     * insertValuePermitNondeterministic}. This implementation does the real work. The expression
+     * will definitely be inserted.
```
smillst, line 502: `The expression will definitely be inserted.`

This isn't true. Only some classes of JavaExpressions are inserted and even then only if `sequentialSemantics` is true.

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractStore.java
```diff
@@ -374,8 +437,47 @@ public static boolean canInsertJavaExpression(JavaExpression expr) {
      * new and old value are taken (according to the lattice). Note that this happens per hierarchy,
      * and if the store already contains information about a hierarchy for which {@code value} does
      * not contain information, then that information is preserved.
+     *
+     * <p>This method does nothing if {@code expr} cannot be inserted into the store.
+     *
+     * @param expr the expression to insert in the store
+     * @param value the value of the expression
+     */
+    public final void insertValue(JavaExpression expr, @Nullable V value) {
```
smillst, line 446: You have changed the semantics of this method.  That should be mention in the change log.

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractStore.java
```diff
@@ -374,8 +437,47 @@ public static boolean canInsertJavaExpression(JavaExpression expr) {
      * new and old value are taken (according to the lattice). Note that this happens per hierarchy,
      * and if the store already contains information about a hierarchy for which {@code value} does
      * not contain information, then that information is preserved.
+     *
+     * <p>This method does nothing if {@code expr} cannot be inserted into the store.
```
smillst, line 441: The Javadoc doesn't say that this method will not add expressions that are nondeterministic.  Nondeterministic expression can be inserted in the store, so it's not enough to say `This method does nothing if {@code expr} cannot be inserted into the store.`.

checker/src/main/java/org/checkerframework/checker/lock/LockStore.java
```diff
@@ -226,17 +226,12 @@ boolean hasLockPossiblyHeld(CFValue value) {
     }
 
     @Override
-    public void insertValue(JavaExpression je, @Nullable CFValue value) {
-        if (value == null) {
-            // No need to insert a null abstract value because it represents
-            // top and top is also the default value.
-            return;
-        }
+    public void insertValueImpl(JavaExpression je, @Nullable CFValue value) {
```
smillst, line 229: It's weird that the "helper" method is the one that is overridden.

smillst (changes requested):

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractStore.java
mernst: `insertOrRefine` already existed and was public, as are the other `insert*` methods.  We could change them, but doing so probably belongs in a different pull request as it is orthogonal to this one.

checker/src/main/java/org/checkerframework/checker/lock/LockStore.java
mernst: Do you have a different suggestion?  I don't think this pattern is so uncommon.  Its purpose here is to avoid boilerplate that would otherwise appear at the beginning of every method.  I think it would be uglier and more error-prone to expect every overriding method to include that boilerplate.

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractStore.java
smillst: There are 3 insertOrRefine methods:
```
public final void insertOrRefine(JavaExpression expr, AnnotationMirror newAnno) {...}
public final void insertOrRefinePermitNondeterministic(
       JavaExpression expr, AnnotationMirror newAnno) {...}
public void insertOrRefine(
       JavaExpression expr, AnnotationMirror newAnno, boolean permitNondeterministic) {...}
```
Only the first method already exists. The last two were added by this pull request. I'm suggesting that the last `insertOrRefine` method should be protected because it should only be called by CFAbstractStore or its subclasses.

checker/src/main/java/org/checkerframework/checker/lock/LockStore.java
smillst: I would move the boilerplate code into a method, `shouldInsert`, that subclasses can call.  `shouldInsert` should also not be `final` in case subclasses wish to change the behavior.  (For example, a subclasses may wish to insert some nondeterministic expression, but not others.)

smillst (approved):
Thanks!

PR #4288, opened 2021-02-15
Model unchecked exception types more precisely in CFG construction

### LINKED ISSUE #4286: In CFG, unchecked exceptions from method calls should be modeled using RuntimeException | Error rather than Throwable
Currently, in CFG construction, method calls are modeled as possibly throwing a `java.lang.Throwable`, corresponding to the possibility of the call throwing an unchecked exception.  However, this is imprecise, as `Throwable` also includes the possibility of throwing a checked `Exception`.  This can lead to infeasible control flow, e.g.:

```java
public void foo(java.net.Socket s) {
    try {
        sock.isClosed();
        sock.close();
    } catch (java.io.IOException e) { }
}
```
Currently, the Checker Framework CFG adds an exceptional edge from the `isClosed()` call to the catch block, which is infeasible and can lead to false positives.

Instead, modeling of unchecked exceptions from calls should use two exceptional edges with types `RuntimeException` and `Error`, to more precisely capture the possibilities.  
This issue was previously discussed (indirectly) in a PR comment on a Checker Framework fork:

https://github.com/opprop/checker-framework/pull/134#issuecomment-647833037

msridhar:
Fixes #4286

### DIFF
--- a/checker/tests/calledmethods/Postconditions.java
+++ b/checker/tests/calledmethods/Postconditions.java
@@ -87,4 +87,12 @@ public class Postconditions {
             p.build();
         }
     }
+
+    @EnsuresCalledMethods(value = "#1", methods = "a")
+    static void callWithException(Postconditions p) {
+        try {
+            p.a();
+            throw new java.io.IOException();
+        } catch (java.io.IOException e) { }
+    }
 }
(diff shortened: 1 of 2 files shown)


kelloggm (approved):
LGTM

kelloggm (approved):

kelloggm:
@smillst can you take a look at this change?

smillst:
@kelloggm Mike knows this code better so he is going to review it.

mernst (approved):

PR #4291, opened 2021-02-17
Avoid calling getElementValue, for efficiency

mernst:
Merge after ~https://github.com/typetools/checker-framework/pull/4290~.

### DIFF
--- a/framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
@@ -380,32 +380,53 @@ public class ValueAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
     /**
      * Gets the from() element/field out of an IntRange annotation. The from() element/field must
      * exist. Clients should call {@link * #getFromValueFromIntRange} if it might not exist.
+     *
+     * @param anno an IntRange annotation
+     * @return its from() element/field
      */
     private long getIntRangeFromValue(AnnotationMirror anno) {
         return (long) anno.getElementValues().get(intRangeFromElement).getValue();
     }
 
     /**
-     * Gets the from() element/field out of an IntRange annotation. The from() element/field must
-     * exist. Clients should call {@link #getToValueFromIntRange} if it might not exist.
+     * Gets the to() element/field out of an IntRange annotation. The to() element/field must exist.
+     * Clients should call {@link #getToValueFromIntRange} if it might not exist.
+     *
+     * @param anno an IntRange annotation
+     * @return its to() element/field
      */
     private long getIntRangeToValue(AnnotationMirror anno) {
         return (long) anno.getElementValues().get(intRangeToElement).getValue();
     }
 
-    /** Gets the from() element/field out of an ArrayLenRange annotation. */
+    /**
+     * Gets the from() element/field out of an ArrayLenRange annotation.
+     *
+     * @param anno an ArrayLenRange annotation
+     * @return its from() element/field
+     */
     @SuppressWarnings("UnusedMethod")
     private int getArrayLenRangeFromValue(AnnotationMirror anno) {
         return (int) anno.getElementValues().get(arrayLenRangeFromElement).getValue();
     }
 
-    /** Gets the from() element/field out of an ArrayLenRange annotation. */
+    /**
+     * Gets the to() element/field out of an ArrayLenRange annotation.
+     *
+     * @param anno an ArrayLenRange annotation
+     * @return its to() element/field
+     */
     @SuppressWarnings("UnusedMethod")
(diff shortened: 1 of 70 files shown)


javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -684,9 +684,13 @@ public static boolean hasElementValue(AnnotationMirror anno, CharSequence elemen
      * Get the element with the name {@code elementName} of the annotation {@code anno}. The result
      * is expected to have type {@code expectedType}.
      *
-     * <p>For elements of array type, use {@code getElementValueArray} instead.
+     * <p>If the return type is an array, use {@link #getElementValueArray} instead.
      *
-     * <p>For elements of enum type, use {@code getElementValueEnum} instead.
+     * <p>If the return type is an enum, use {@link #getElementValueEnum} instead.
+     *
+     * <p>If possible, it is more efficient to use {@code
```
smillst, line 691: Just for my own understanding, this comment is true for all the `getElementValue*` method right?  Also, if the annotation only has one element, is it still more efficient to call `anno.getElementValues().get(someElement).getValue()`?

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -1006,12 +1081,15 @@ public AnnotationMirror convertIntValToIntRange(AnnotationMirror intValAnno) {
...
+     * Returns a {@link Range} bounded by the values specified in the given {@code @Range}
      * annotation. Also returns an appropriate range if an {@code @IntVal} annotation is passed.
      * Returns {@code null} if the annotation is null or if the annotation is not an {@code
      * IntRange}, {@code IntRangeFromPositive}, {@code IntVal}, or {@code ArrayLenRange}.
+     *
+     * @param rangeAnno a {@code @Range} annotation
+     * @return the {@link Range} that the annotation represents
      */
-    public static Range getRange(AnnotationMirror rangeAnno) {
+    public Range getRange(AnnotationMirror rangeAnno) {
```
smillst, line 1092: This method still has calls to `getElementValue`.  Should they be replaced with calls to `getIntRangeToValue`, `getArrayLenRangeToValue`, and other methods?  In general, should all the calls to `getElementValue` in the Value Checker be replace?

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -350,13 +370,68 @@ public long getFromValueFromIntRange(AnnotatedTypeMirror atm) {
...
+        return (long) anno.getElementValues().get(intRangeToElement).getValue();
+    }
+
+    /**
+     * Gets the from() element/field out of an ArrayLenRange annotation.
+     *
+     * @param anno an ArrayLenRange annotation
+     * @return its from() element/field
+     */
+    @SuppressWarnings("UnusedMethod")
```
smillst, line 408: I think this method should be used in getRange.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -350,13 +370,68 @@ public long getFromValueFromIntRange(AnnotatedTypeMirror atm) {
...
+    }
+
+    /**
+     * Gets the to() element/field out of an ArrayLenRange annotation.
+     *
+     * @param anno an ArrayLenRange annotation
+     * @return its to() element/field
+     */
+    @SuppressWarnings("UnusedMethod")
+    private int getArrayLenRangeToValue(AnnotationMirror anno) {
```
smillst, line 420: I think this method should be used in getRange.

framework/src/main/java/org/checkerframework/common/value/ValueAnnotatedTypeFactory.java
```diff
@@ -350,13 +370,68 @@ public long getFromValueFromIntRange(AnnotatedTypeMirror atm) {
...
+    }
+
+    /**
+     * Gets the value() element/field out of a MinLen annotation.
+     *
+     * @param anno a MinLen annotation
+     * @return its value() element/field
+     */
+    @SuppressWarnings("UnusedMethod")
+    private int getMinLenValueValue(AnnotationMirror anno) {
```
smillst, line 431: This method should be called in getSpecifiedMinLenValue.

smillst (approved):

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
mernst: Yes to both questions.  I updated the comment.

PR #4292, opened 2021-02-17
Use string literals instead of method calls

mernst:
An advantage is that the code is shorter.
A disadvantage may be that automated refactoring tools might not work as well.


smillst (approved):

smillst:
Most of the class names are java classes that are unlikely to change, so this change seems fine to me.

PR #4293, opened 2021-02-18
use areSameByName instead of areSameByClass in AccumulationAnnotatedTypeFactory

kelloggm:
@msridhar noticed that the calls to `areSameByClass` were taking an inordinate amount of wall clock time when running the CalledMethods Checker, because `areSameByClass` calls `getCanonicalName` under the hood. This semantics-preserving change modifies the checker to use `areSameByName` instead, which directly compares strings instead (and should therefore be faster).

### DIFF
--- a/framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
@@ -19,6 +19,7 @@ import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.util.Elements;
 import org.checkerframework.checker.nullness.qual.MonotonicNonNull;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.checkerframework.checker.signature.qual.CanonicalName;
 import org.checkerframework.common.accumulation.AccumulationChecker.AliasAnalysis;
 import org.checkerframework.common.basetype.BaseAnnotatedTypeFactory;
 import org.checkerframework.common.basetype.BaseTypeChecker;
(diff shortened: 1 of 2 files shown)


kelloggm:
@mernst the fields in `AccumulationAnnotatedTypeFactory` have dependencies between them, so their initializers cannot be re-ordered as you did in d90eb1b. I undid that change and instead changed the order of the declarations to match the order in which they are initialized, although I'm not thrilled that this splits `accumulator` and `accumulatorName` (with `top`).

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
```diff
@@ -663,6 +682,6 @@ protected String convertToPredicate(AnnotationMirror anno) {
      * @return true if anno is a predicate annotation
      */
     protected boolean isPredicate(AnnotationMirror anno) {
-        return predicate != null && AnnotationUtils.areSameByClass(anno, predicate);
+        return predicateName != null && AnnotationUtils.areSameByName(anno, predicateName);
```
smillst, line 685: You should use `AnnotatedTypeFactory#areSameByClass` instead.  It has a cache from Class to name that makes it much faster than AnnotationUtils.areSameByClass.  (You may want to open a new pull request with just that change and discard the other changes in this branch.)

smillst (changes requested):

framework/src/main/java/org/checkerframework/common/accumulation/AccumulationAnnotatedTypeFactory.java
smillst: @mernst  is going to add comments to `AnnotationUtils#areSameByClass` and related classes explaining that `AnnotatedTypeFactory#areSameByClass` or  `AnnotationUtils.areSameByName` are faster.

PR #4301, opened 2021-02-19
Default annotations based on variable names

mernst:


framework/src/main/java/org/checkerframework/framework/type/typeannotator/DefaultForTypeAnnotator.java
```diff
@@ -193,4 +212,151 @@ public DefaultForTypeAnnotator addStandardDefaults() {
...
+        }
+    }
+
+    @Override
+    public Void visitExecutable(AnnotatedExecutableType type, Void aVoid) {
+        ExecutableElement element = type.getElement();
+
+        Iterator<AnnotatedTypeMirror> paramTypes = type.getParameterTypes().iterator();
+        for (VariableElement paramElt : element.getParameters()) {
+            String paramName = paramElt.getSimpleName().toString();
```
smillst, line 242: Parameters from elements only have names if `-g` is used to compile.  See failing test case I added.

smillst (changes requested):

build.gradle
```diff
@@ -742,7 +742,8 @@ subprojects {
                                 "-workDir:${jtregOutput}/${name}/work",
                                 "-reportDir:${jtregOutput}/${name}/report",
                                 "-verbose:error,fail",
-                                "-javacoptions:-g",
+                                // Don't add debugging information
```
mernst, line 745: @smillst Is this change desirable?  Does it belong as part of this pull request?

framework/src/main/java/org/checkerframework/framework/type/typeannotator/DefaultForTypeAnnotator.java
mernst: I clarified when parameter names have an effect and expanded the test case.

build.gradle
smillst: If you revert this change, then the jtreg tests added by this pull request will fail.  So if we don't make this change here, then the test will have to be rewritten.

It took me a while to figure out that this was necessary to get the test case I added to fail.  (I could reproduce the problem locally, but not in a test case.)   In the future, we may have other test case the behave differently if `-g` is passed, so it may be worthwhile not to pass it.  But, I also don't know what the benefit of passing `-g` to the jtreg test was in the first place.

smillst (approved):

PR #4309, opened 2021-02-21
Permit format methods to have leading arguments

mernst:
This changes special-case code that permitted a `Locale` to general code that permits any extra arguments before the first `String`.


checker/src/main/java/org/checkerframework/checker/formatter/messages.properties
```diff
@@ -1,3 +1,4 @@
+format.method.invalid=method %s is annotated @FormatMethod but has no String formal parameter
```
smillst, line 1: Please add a test of this error.

smillst (approved):
Please update the Javadoc of `FormatMethod`.

PR #4317, opened 2021-02-22
In the toString of the JavaExpression Unknown, print more information

smillst:


framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1711,7 +1711,7 @@ protected void checkPreconditions(MethodInvocationTree tree, Set<Precondition> p
                 inferredAnno = hierarchy.findAnnotationInSameHierarchy(annos, anno);
             }
             if (!checkContract(exprJe, anno, inferredAnno, store)) {
-                if (exprJe != null && !exprJe.containsUnknown()) {
+                if (exprJe != null) {
```
smillst, line 1714: See  `checker/tests/nullness/Issue2470.java`.

dataflow/src/main/java/org/checkerframework/dataflow/expression/Unknown.java
```diff
@@ -31,7 +67,7 @@ public int hashCode() {
 
     @Override
     public String toString() {
-        return "?";
+        return originalExpression;
```
smillst, line 70: We could add strings to `originalExpression` to make it clear that the expression is unknown.  Something like `"Unknown(" + originalEI + ")"`. don't know if that's necessary.

mernst (approved):

PR #4318, opened 2021-02-22
Throw a parse exception if `this` is used in a context where it can't be.

smillst:


mernst (approved):
I added a test, plus tests that are commented out.

PR #4322, opened 2021-02-23
Don't simplify "value".length() in JavaExpressionParseUtil.

smillst:
Simplifying `"value".length()` in needs to be done after viewpoint-adaption.  Since that will soon happen after parsing, move the simplification code.


mernst (commented):
The title says this removes functionality, but the pull request only adds code.  Is this missing some changes?  Or should the title be changed?

smillst:
@mernst, yes I forgot to actually remove the code. When I remove the code, tests fail, so I'm going to close this pull request and open a new one later.  Thanks!

PR #4326, opened 2021-02-24
Simplify use of JavaExpressionContext in Subsequence

smillst:


checker/src/main/java/org/checkerframework/checker/index/Subsequence.java
```diff
@@ -160,11 +132,10 @@ public static JavaExpressionContext getContextFromJavaExpression(
      * applied.
      *
      * @param s a Java expression string
-     * @param context the parse context
      * @return the string, standardized and viewpoint-adapted
```
mernst, line 135: This comment (and text in the Javadoc description) seems no longer accurate.

mernst (approved):

PR #4330, opened 2021-02-25
Fix typo

ChristosToxiadis:
I think that those 3 annotations belong to the Initialization Checker.


mernst:
Your wording fix is definitely more clear.  Thanks.

ChristosToxiadis:
hahaha thanks!

mernst:
My guess is that the sentence was written with the implementation knowledge that the Nullness Checker and Initialization Checker are actually the same checker.  But the manual should be written from a user perspective.  So thanks again.

PR #4342, opened 2021-03-02
Always build all projects, AFU version is CF version, don't use hg

mernst:


smillst (approved):
I ran part of the build script , and found and fixed a few problems.  Other than that, I haven't reviewed these changes.

PR #4345, opened 2021-03-02
Add test for overloaded method in contract.

smillst:
Here's a code snippet taken from the test.
```
    String name(ClassA classA) {...}
    String name(ClassB classB) {...}

    @EnsuresKeyFor(value = "name(#1)", map = "map")
    void put(ClassA classA) {...}
    void test(ClassB classB) {
        put(classB);
     }
```

At the call to `put`, `name(classB)` should refer to `name(ClassA classA)` not `name(ClassB classB)`

The test is currently fails and is therefore skipped, but it will be fixed once string expression are not parsed and viewpoint-adapted at the same time.  ~~I also add new methods to TreeUtils that will be needed for the Nullness Checker to type check the dataflow subproject.~~

### DIFF
--- a/javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
+++ b/javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
@@ -408,6 +408,30 @@ public final class TreeUtils {
     }
 
     /**
+     * Determine whether the given MethodInvocationTree has an underlying element.
+     *
+     * @param node the MethodInvocationTree to test
+     * @return whether the tree refers to an identifier, member select, or method invocation
+     */
+    @EnsuresNonNullIf(result = true, expression = "elementFromUse(#1)")
+    @Pure
+    public static boolean isUseOfElement(MethodInvocationTree node) {
+        return isUseOfElement((ExpressionTree) node);
+    }
+
+    /**
+     * Determine whether the given NewClassTree has an underlying element.
+     *
+     * @param node the NewClassTree to test
+     * @return whether the tree refers to an identifier, member select, or method invocation
+     */
+    @EnsuresNonNullIf(result = true, expression = "elementFromUse(#1)")
+    @Pure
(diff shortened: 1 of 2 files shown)


javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -407,6 +407,30 @@ public static boolean isUseOfElement(ExpressionTree node) {
...
+     * @return true if the tree refers to an identifier, member select, or method invocation
+     */
+    @EnsuresNonNullIf(result = true, expression = "elementFromUse(#1)")
+    @Pure
+    public static boolean isUseOfElement(MethodInvocationTree node) {
+        return isUseOfElement((ExpressionTree) node);
+    }
+
+    /**
+     * Determine whether the given NewClassTree has an underlying element.
```
mernst, line 423: If this returns false, then is the NewClassTree definitely creating an anonymous class?  Are there any other possibilities?  I have a similar question about the MethodInvocationTree overload:  when is a method call not to a method with an element?  Does that happen when the method is non-public?  (I don't think so, but I'm wondering.)

mernst (approved):

javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
smillst: > when is a method call not to a method with an element? 

The only time a method call doesn't have an element is when javac issued an error for the class that contains the element.  In that case, the checker should not have continued type checking, but it currently does.  This a bug.  I'm going to open a pull request to fix that bug.

smillst: In another pull request, I'll change `elementFromUse(MethodInvocationTree node)` and `elementFromUse(NewClassTree node)` so that they never return null.  Then the two methods I added in this pull request won't be needed, so I'm going to delete them.

PR #4353, opened 2021-03-03
EC2 Async clients have both synchronous and asynchronous versions

kelloggm:
#4272 caused some tests in the Object Construction Checker to fail, because the EC2 Async client/interface support both async and sync requests, so we need stubs for both.


mernst (approved):
This looks good to me -- thanks for the fixes -- but please don't merge until the tests are passing.

PR #4354, opened 2021-03-03
NewClassTrees and MethodInvocation trees always have an element.

smillst:
Merge https://github.com/typetools/checker-framework/pull/4352 first.

### DIFF
--- a/checker/tests/nullness/java-unsound/Figure1.java
+++ b/checker/tests/nullness/java-unsound/Figure1.java
@@ -1,5 +1,6 @@
 // Unsound only in Java 8, Java 9+ already gives an error
 // @below-java9-jdk-skip-test
+// @skip-test noneed to test for the javac error.
 
 public class Figure1 {
     static class Constrain<A, B extends A> {}
--- a/checker/tests/nullness/java-unsound/Figure1.java
+++ b/checker/tests/nullness/java-unsound/Figure1.java
@@ -1,6 +1,6 @@
 // Unsound only in Java 8, Java 9+ already gives an error
 // @below-java9-jdk-skip-test
-// @skip-test noneed to test for the javac error.
+// @skip-test no need to test for the javac error.
 
 public class Figure1 {
     static class Constrain<A, B extends A> {}
--- a/docs/manual/introduction.tex
+++ b/docs/manual/introduction.tex
@@ -557,6 +557,8 @@ Unsound checking: ignore some errors
   which is a javac argument to halt compilation if a warning is issued.
 \item \<-AignoreInvalidAnnotationLocations>
   Ignore annotations in bytecode that have invalid annotation locations.
+\item \<-AnoWarnOnTypeCheckingHalt>
+  Don't issue the Don't issue the ``Javac errored; type checking halted.'' warning.
 \end{itemize}
 
 \label{unsound-by-default}
--- a/docs/manual/introduction.tex
+++ b/docs/manual/introduction.tex
@@ -558,7 +558,7 @@ Unsound checking: ignore some errors
 \item \<-AignoreInvalidAnnotationLocations>
   Ignore annotations in bytecode that have invalid annotation locations.
 \item \<-AnoWarnOnTypeCheckingHalt>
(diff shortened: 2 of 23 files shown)


checker/tests/nullness/java-unsound/Figure1.java
```diff
@@ -1,5 +1,6 @@
 // Unsound only in Java 8, Java 9+ already gives an error
 // @below-java9-jdk-skip-test
+// @skip-test no need to test for the javac error.
```
mernst, line 3: Should the file be removed (and likewise for the other newly skipped files)?

docs/manual/introduction.tex
```diff
@@ -557,6 +557,8 @@
   which is a javac argument to halt compilation if a warning is issued.
 \item \<-AignoreInvalidAnnotationLocations>
   Ignore annotations in bytecode that have invalid annotation locations.
+\item \<-AnoWarnOnTypeCheckingHalt>
+  Don't issue the ``Javac errored; type checking halted.'' warning.
```
mernst, line 561: Does type-checking still halt in this case?  (I assume so, but please make it explicit either way.)

When would a user want to pass this command-line argument?  If we don't have a compelling use-case, maybe this flag should not be added.

I see that two tests use it and they fail if it is removed, but I'm not sure why.
For instance, one fails this way:

```
org.checkerframework.framework.test.junit.InitializedFieldsValueTest > run[/home/mernst/research/types/checker-framework-fork-smillst-branch-no-null-element/framework/initialized-fields-value] FAILED
    java.lang.AssertionError: 3 out of 3 expected diagnostics were found.
    1 unexpected diagnostic was found:
    :0: warning: Javac errored; type checking halted.
    While type-checking /home/mernst/research/types/checker-framework-fork-smillst-branch-no-null-element/framework/tests/initialized-fields-value/ClassInitializer.java, /home/mernst/research/types/checker-framework-fork-smillst-branch-no-null-element/framework/tests/initialized-fields-value/ClassInitializer2.java, /home/mernst/research/types/checker-framework-fork-smillst-branch-no-null-element/framework/tests/initialized-fields-value/ClassInitializer2a.java, /home/mernst/research/types/checker-framework-fork-smillst-branch-no-null-element/framework/tests/initialized-fields-value/ClassInitializer3.java, /home/mernst/research/types/checker-framework-fork-smillst-branch-no-null-element/framework/tests/initialized-fields-value/ManualConstructor.java
```

but I am able to run `javac` on all of those files without problem.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
```diff
@@ -757,7 +762,9 @@ public void typeProcessingOver() {
         for (BaseTypeChecker checker : getSubcheckers()) {
             checker.typeProcessingOver();
         }
-
+        if (javacErrored && parentChecker == null && !hasOption("noWarnOnTypeCheckingHalt")) {
+            messager.printMessage(Kind.WARNING, "Javac errored; type checking halted.");
```
mernst, line 766: This message makes it sound like javac is buggy.  I think the point is that the program is not valid Java:  normal compilation without pluggable type-checking didn't succeed.  Please clarify the message.  Text like "javac issued an error" might be good, though from the user perspective javac also issues the pluggable type-checking errors, so the message should probably be even more explicit or detailed than that.  Another clarification is that *pluggable* type checking halted.

Also, please clarify what is the consequence to users, or why they care.  Might there be `.class` files generated that didn't get type-checked?  If javac erred, wouldn't users see other warning messages?

mernst (changes requested):
I'm a bit confused about the notion of a javac error and what the consequences are for users.

checker/tests/nullness/java-unsound/Figure1.java
smillst: See reply: https://github.com/typetools/checker-framework/pull/4352/files#r587872848

smillst:
Closed in favor of https://github.com/typetools/checker-framework/pull/4385.

Issue #4358, opened 2021-03-05
manual: building from source: `brew cask install java` doesn't work anymore


cushon:
The instructions for building from source in https://checkerframework.org/manual/#build-source include:

```
brew cask install java
```

* the `cask` subcommand isn't available in recent versions of homebrew. According to [this SO answer](https://stackoverflow.com/questions/30413621/homebrew-cask-option-not-recognized) it has been replaced with e.g. `brew install --cask java`
* the `java` cask doesn't exist anymore, one alternative is: https://github.com/AdoptOpenJDK/homebrew-openjdk

PR #4359, opened 2021-03-05
Update Mac OS prerequisites in 'Building from source'

cushon:
Fixes https://github.com/typetools/checker-framework/issues/4358

### DIFF
--- a/docs/manual/troubleshooting.tex
+++ b/docs/manual/troubleshooting.tex
@@ -697,8 +697,9 @@ pip3 install html5validator
 \begin{Verbatim}
 brew update
 brew install git ant hevea maven mercurial librsvg unzip make
-brew cask install java
-brew cask install mactex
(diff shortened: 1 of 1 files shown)


mernst:
@cushon Thanks for the fix, Liam.  I guess we haven't installed afresh on MacOS recently.

Your patch installs JDK 8.  Would it be better to recommend JDK 11 instead?  That is the latest stable version.

cushon:
Updated to use JDK 11, thanks! Suggesting whatever version most Checker Framework developers are using is probably best, if that's 11 at this point that SGTM.

mernst:
We use both JDK 8 and JDK 11, but only because so much of the world is still on JDK 8.  I hate to suggest JDK 8 to anyone these days.

mernst (approved):

PR #4366, opened 2021-03-06
Add groups to Gradle tasks

mernst:


checker/build.gradle
```diff
@@ -65,6 +65,7 @@ jar {
 }
 
 task copyJarsToDist(dependsOn: shadowJar, group: 'Build') {
+    group 'Build'
```
smillst, line 68: This tasks is already in a group; one line up.

smillst (approved):
The tasks that already have a group list the group on the same line as the task name and type:
```
task copyJarsToDist(dependsOn: shadowJar, group: 'Build') {
```

If it's not too much trouble, could you do the same for the groups added in this pull request?

smillst:
Thanks!

PR #4378, opened 2021-03-08
Parse Strings into JavaExpressions then viewpoint-adapt the JavaExpressions

smillst:
First set of files to review:
1. JavaExpressionParseUtil.java 
2. StringToJavaExpression.java
3. All changes in the `org.checkerframework.dataflow.expression` package
4. BaseTypeVisitor.java and CFAbstractTransfer.java (These changes are to parsing contract expressions.)

Second set of files to review:
1. DependentTypesHelper.java
2. Subclasses of DependentTypesHelper: OffsetDependentTypesHelper.java and anonymous class created in LockAnnotatedTypeFactory#createDependentTypesHelper

Third set:
All the other files.


framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -99,84 +86,64 @@
...
-     * JavaExpressionParseException}.
-     *
-     * <p>If {@code localPath} is non-null, then identifiers are parsed as if the expression was
-     * written at the location of {@code localPath}. This means identifiers will be parsed to local
-     * variables in scope at {@code localPath} when possible. If {@code localPath} is null, then no
-     * identifier can be parsed to a local variable. In either case, the parameter syntax, e.g. #1,
-     * is always parsed to the arguments in {@code context}. This is because a parameter of a lambda
-     * can refer both to local variables in scope at its declaration and to a parameter of the
-     * lambda.
+     * <p>Clients should call one of the static methods in {@link StringToJavaExpression} rather
```
mernst, line 104: If clients should not call this method, why is is public?  Who should call it?

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -99,84 +86,64 @@
...
-     *     checker
-     * @param localPath if non-null, the expression is parsed as if it were written at this location
-     * @return the JavaExpression for the given string
+     * @param expression the string expression to parse
+     * @param enclosingType type of the class that encloses the JavaExpression
+     * @param thisReference the JavaExpression to which to parse "this", or null if "this" should
+     *     not appear in the expression
+     * @param parameters list of JavaExpressions to which to parse formal parameter references such
+     *     as "#2", or null if formal parameter references should not appear in the expression
+     * @param localVarPath if non-null, the expression is parsed as if it were written at this
```
mernst, line 113: Is it an invariant that localVarPath is always within enclosingType?

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -216,47 +184,124 @@ private static String replaceParameterSyntax(String expression) {
      * A visitor class that converts a JavaParser {@link Expression} to a {@link JavaExpression}.
      */
     private static class ExpressionToJavaExpressionVisitor
```
mernst, line 186: This nested class is enormous, so we could consider moving it to its own file.  Leaving it here better encapsulates it, though.

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -216,47 +184,124 @@ private static String replaceParameterSyntax(String expression) {
...
+         * @param thisReference JavaExpression to which to parse "this", or null if "this" should
+         *     not appear in the expression
+         * @param parameters list of JavaExpressions to which to parse a formal parameter reference
+         *     such as "#2", or null if parameters should not appear in the expression
          * @param localVarPath if non-null, the expression is parsed as if it were written at this
          *     location
+         * @param pathToCompilationUnit required to use the underlying Javac API
          * @param env the processing environment
          */
         ExpressionToJavaExpressionVisitor(
```
mernst, line 239: I suggest making this private.  It is only called by `convert` immediately below.

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -348,22 +392,22 @@ public JavaExpression visit(ArrayAccessExpr expr, JavaExpressionContext context)
             }
             TypeMirror componentType = ((ArrayType) arrayType).getComponentType();
 
-            JavaExpression index = expr.getIndex().accept(this, context);
+            JavaExpression index = expr.getIndex().accept(this, null);
```
mernst, line 395: The code above throws an exception if the "a" part of "a[i]" is not an array.  Should it also check that the "i" part is an integer?

(There are probably other places that the type could also be checked.  I think that wrong types won't cause crashes, but they may be confusing to users, which would be a reason to warn during parsing, when it is easy to warn.)

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -461,7 +536,7 @@ public JavaExpression visit(NameExpr expr, JavaExpressionContext context) {
          *
          * @param type the type in which to search for {@code identifier}
          * @param identifier possible class name
-         * @return the {@code ClassName} for {@code identifier} or null if it is not a class name
+         * @return the {@code ClassName} for {@code identifier}, or null if it is not a class name
          */
         protected @Nullable ClassName getIdentifierAsUnqualifiedClassName(
                 TypeMirror type, String identifier) {
```
mernst, line 542: Should this method start with
```java
if (type.getKind() != TypeKind.DECLARED) {
  return null;
}
```

Then a later test on line 485/560 can be dropped.

mernst, line 542: Below, there is a test
```java
            Symbol.PackageSymbol packageSymbol =
                    resolver.findPackage("java.lang", pathToCompilationUnit);
            if (packageSymbol != null) {
```

How can `packageSymbol` be null?  Should an exception be thrown in that case?

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -526,9 +601,11 @@ public JavaExpression visit(NameExpr expr, JavaExpressionContext context) {
          * to using that field at the given {@code context}. If {@code identifier} is not a field
```
mernst, line 601: I don't understand "at the given context".  There is not a corresponding formal parameter.

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -526,9 +601,11 @@ public JavaExpression visit(NameExpr expr, JavaExpressionContext context) {
...
-         * @param receiverExpr the receiver of the field; the expression used to access the field
+         * @param receiverExpr the receiver of the field access; the expression used to access the
+         *     field
          * @param identifier possibly a field name
-         * @return a field access, or null if {@code identifier} is not a field
+         * @return a field access, or null if {@code identifier} is not a field that can be accessed
+         *     via {@code receiverExpr}
          */
         protected @Nullable FieldAccess getIdentifierAsField(
                 JavaExpression receiverExpr, String identifier) {
```
mernst, line 611: In this code below:
```java
            if (identifier.equals("length") && receiverType.getKind() == TypeKind.ARRAY) {
                fieldElem = resolver.findField(identifier, receiverType, pathToCompilationUnit);
            }
            if (fieldElem == null) {
```

1. Should there be an assertion that `fieldElem` is non-null after the assignment?
2. I think the `} if (fieldElem == null) {` can be changed to `} else {`.

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -569,30 +646,70 @@ public JavaExpression visit(NameExpr expr, JavaExpressionContext context) {
...
-                }
+            List<JavaExpression> arguments = new ArrayList<>(expr.getArguments().size());
+            for (Expression argument : expr.getArguments()) {
+                arguments.add(argument.accept(this, null));
             }
 
-            // Find the method element.
             ExecutableElement methodElement;
             try {
                 methodElement =
```
mernst, line 715: This call to `getMethodElement` does not seem to account for autoboxing.  (If it does, that should be documented in `getMethodElement`.  Therefore, I am concerned whether the (right) method will be found.

Also, why does `getMethodElement` return `Element` rather than `ExecutableElement`?  I suggest it be changed, which permits some local variable types to be changed and casts eliminated.  Within its implementation, how can the test `(element.getKind() == ElementKind.METHOD)` evaluate to false?  If it does somehow, should `element` be reset to `null`?

I suggest making these changes in a separate pull request because they are not directly related to the current pull request.

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -833,20 +952,20 @@ public JavaExpression visit(UnaryExpr expr, JavaExpressionContext context) {
...
-            // isSubtype() first does the cheaper test isSameType()
+            // isSubtype() first does the cheaper test isSameType(), so no need to do it here.
             if (types.isSubtype(leftType, rightType)) {
                 type = rightType;
             } else if (types.isSubtype(rightType, leftType)) {
                 type = leftType;
             } else if (expr.getOperator() == BinaryExpr.Operator.PLUS
                     && (types.isSameType(leftType, stringTypeMirror)
-                            || types.isSameType(rightType, stringTypeMirror))) {
+                            || TypesUtils.isString(rightType))) {
```
mernst, line 968: Why does one of the disjuncts use `isSameType` but the other uses `isString`?

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -907,19 +1026,17 @@ public JavaExpression visit(BinaryExpr expr, JavaExpressionContext context) {
         }
 
         /**
-         * Converts the JavaParser type to a TypeMirror. Returns null if {@code type}'s kind is not
+         * Converts the JavaParser type to a TypeMirror. Returns null if {@code type} is not
```
mernst, line 1029: When is the type not handled?  Can you give an example?
Also, there is one invocation of `convertTypeToTypeMirror` that is not tested for `null` before being used.  (The other two are tested.)

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -569,30 +646,70 @@ public JavaExpression visit(NameExpr expr, JavaExpressionContext context) {
             return null;
         }
 
+        /**
+         * Returns a JavaExpression for the given field.
+         *
+         * <p>{@code isOriginalReceiver} might be false when the user wrote "a.f" but {@code f} is
```
mernst, line 652: Is this explanation correct?
Please look over the diffs in my commit to ensure I didn't make any inadvertent mistakes elsewhere.

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -569,30 +646,70 @@ public JavaExpression visit(NameExpr expr, JavaExpressionContext context) {
...
+            if (ElementUtils.isStatic(fieldElem)) {
+                Element classElem = fieldElem.getEnclosingElement();
+                JavaExpression staticClassReceiver = new ClassName(ElementUtils.getType(classElem));
+                return new FieldAccess(staticClassReceiver, fieldType, fieldElem);
+            }
+            JavaExpression locationOfField;
+            if (isOriginalReceiver) {
+                locationOfField = receiverExpr;
+            } else {
+                locationOfField = new ThisReference(receiverType);
```
mernst, line 680: Is `ThisReference` always the right thing?  If the original expression was "a.f" but `f` is declared in a supertype, then might this produce "superthis.f" which could mean something different than "a.f" (because "this" could have its own f field, for example if "this" and "a" have the same type)?

mernst (changes requested):
Here are comments on JavaExpressionParseUtil.java.

dataflow/src/main/java/org/checkerframework/dataflow/expression/JavaExpression.java
```diff
@@ -660,4 +678,157 @@ public static JavaExpression getPseudoReceiver(TreePath path, TypeMirror enclosi
...
+     * varargs.
+     *
+     * @param tree invocation of the method or constructor
+     * @param argTrees the arguments to the method or constructor; subexpressions of {@code tree}
+     * @return the arguments, as JavaExpressions
+     */
+    private static List<JavaExpression> argumentTreesToJavaExpressions(
+            ExpressionTree tree, List<? extends ExpressionTree> argTrees) {
+
+        if (tree.getKind() == Kind.METHOD_INVOCATION) {
```
mernst, line 763: How can this be false?  The specification in the Javadoc says `tree` is an invocation.

dataflow/src/main/java/org/checkerframework/dataflow/expression/JavaExpression.java
```diff
@@ -660,4 +678,157 @@ public static JavaExpression getPseudoReceiver(TreePath path, TypeMirror enclosi
...
+            MethodInvocationTree methodInvoc = (MethodInvocationTree) tree;
+            assert TreeUtils.isUseOfElement(methodInvoc) : "@AssumeAssertion(nullness): tree kind";
```
mernst, line 765: Is this still needed?  I seem to recall some other changes in this diff removed code like this.

dataflow/src/main/java/org/checkerframework/dataflow/expression/JavaExpression.java
```diff
@@ -660,4 +677,159 @@ public static JavaExpression getPseudoReceiver(TreePath path, TypeMirror enclosi
...
+     * passed in an array.
+     *
+     * @param method the method or constructor
+     * @param args the arguments at the call site
+     * @return true if method is a varargs method and its varargs arguments are not passed in an
+     *     array
+     */
+    private static boolean isVarArgsInvocation(
+            ExecutableElement method, List<? extends ExpressionTree> args) {
+        if (method == null || !method.isVarArgs()) {
```
mernst, line 805: How can `method` be null?  If it is, I think there should be an exception thrown here or in the caller.

dataflow/src/main/java/org/checkerframework/dataflow/expression/JavaExpression.java
```diff
@@ -660,4 +677,159 @@ public static JavaExpression getPseudoReceiver(TreePath path, TypeMirror enclosi
...
+        return getArrayDepth(ElementUtils.getType(lastParamElt)) != getArrayDepth(lastArgType);
+    }
+
+    /**
+     * Returns the depth of an array type.
+     *
+     * @param arrayType an array type
+     * @return the depth of {@code arrayType}
+     */
+    private static int getArrayDepth(TypeMirror arrayType) {
```
mernst, line 826: I think this method belongs in `TypesUtils` rather than here.

mernst (changes requested):
Review of expression classes.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -993,32 +991,48 @@ private void checkContractsAtMethodDeclaration(
...
                     atypeFactory
                             .getDependentTypesHelper()
-                            .viewpointAdaptQualifierFromContract(annotation, jeContext, methodTree);
+                            .viewpointAdaptQualifierFromContract(
+                                    annotation, stringToJavaExpr, methodTree);
 
             JavaExpression exprJe;
             try {
-                exprJe = JavaExpressionParseUtil.parse(expressionString, jeContext);
+                // Parse the expressionString that was written on methodElement at the methodDecl.
```
mernst, line 1007: What is variable `methodDecl`?  Should this be "at the method declaration"?
Does "at" refer to "parse" or to "on methodElement"?  Please clarify.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -993,32 +991,48 @@ private void checkContractsAtMethodDeclaration(
...
+                exprJe = StringToJavaExpression.atMethodBody(expressionString, methodTree, checker);
             } catch (JavaExpressionParseException e) {
-                exprJe = null;
                 checker.report(methodTree, e.getDiagMessage());
+                if (formalParamNames != null && formalParamNames.contains(expressionString)) {
```
mernst, line 1011: I think there should be an error if the contract expression contains any formal parameter name as a subexpression, not just if the contract expression is a formal parameter name.

mernst (changes requested):
Part 4 of initial review.

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
smillst: As you point out, this code is buggy.  I rewrote it and fix the bugs in a separate pull request: https://github.com/typetools/checker-framework/pull/4388.  (That pull request is also now merged into this.)

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
smillst: I've done so here: https://github.com/typetools/checker-framework/pull/4394.

dataflow/src/main/java/org/checkerframework/dataflow/expression/JavaExpression.java
smillst: Tree can be a NewClassTree, which is a constructor invocation.

smillst: The Javadoc says `tree` is a method or constructor invocation.  If `tree` is a constructor invocation, then the tree kind is `NEW_CLASS`.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
smillst: See https://github.com/typetools/checker-framework/pull/4396.

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
smillst: After this PR is merged. We should make a new package "org.checkerframework.framework.util.expression" and move JavaExpressionParseUtil and StringToJavaExpression to that package.  Also, ExpressionToJavaExpressionVisitor will be made a top level class in that package.

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
smillst: Let's defer this, I opened an issue for it https://github.com/typetools/checker-framework/issues/4400.

framework/src/main/java/org/checkerframework/framework/util/StringToJavaExpression.java
```diff
@@ -0,0 +1,362 @@
...
+            String expression,
+            LambdaExpressionTree lambdaTree,
+            TreePath parentPath,
+            SourceChecker checker)
+            throws JavaExpressionParseException {
+
+        TypeMirror enclosingType = TreeUtils.typeOf(TreePathUtil.enclosingClass(parentPath));
+        JavaExpression receiver = JavaExpression.getPseudoReceiver(parentPath, enclosingType);
+        ThisReference thisReference =
+                receiver instanceof ClassName ? null : (ThisReference) receiver;
```
mernst, line 289: I'm confused by this line.  Should the instance of "ClassName" be "ThisReference"?

mernst (changes requested):
Is the getParameterIndices method useful?  Maybe it's just as good to do the substitution for all digits 0-9 and not have to call getParameterIndices, which can then be deleted.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -22,51 +22,51 @@
...
-import org.checkerframework.javacutil.ElementUtils;
-import org.checkerframework.javacutil.SystemUtil;
 import org.checkerframework.javacutil.TreePathUtil;
 import org.checkerframework.javacutil.TreeUtils;
+import org.checkerframework.javacutil.TypesUtils;
 import org.plumelib.util.StringsPlume;
 
 /**
  * A class that helps checkers use qualifiers that are represented by annotations with Java
  * expression strings. This class performs four main functions:
```
mernst, line 62: Only three are listed.

I think it makes sense to break this class into two (or more??) classes to make each one smaller and more focused.  What do you think?

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -76,15 +76,17 @@
...
+ * <p>Steps 2 and 3 are separated so that an error is issued only once per invalid expression string
  * rather than every time the expression string is parsed. (The expression string is parsed multiple
  * times because annotated types are created multiple times.)
  */
 public class DependentTypesHelper {
+
+    /** AnnotatedTypeFactory */
     protected final AnnotatedTypeFactory factory;
 
     /** A map of annotation classes to the names of their elements that are Java expressions. */
```
mernst, line 88: I think this should map to a list of ExecutableElement, not String.  That can be more efficient.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -76,15 +76,17 @@
...
-    private Map<Class<? extends Annotation>, List<String>> annoToElements;
+    private final Map<Class<? extends Annotation>, List<String>> annoToElements;
 
     /**
      * The {@code ExpressionErrorChecker} that scans an annotated type and returns a list of {@link
```
mernst, line 92: Should this be named `...Collector` rather than `...Checker`?  What it does seems to be collection.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -121,8 +128,16 @@ public DependentTypesHelper(AnnotatedTypeFactory factory) {
             }
         }
         this.expressionErrorChecker = new ExpressionErrorChecker();
```
mernst, line 130: Nit: I usually put initialization that doesn't depend on the constructor arguments in the field declaration.  Either way is OK, though.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -206,26 +233,21 @@ public void viewpointAdaptMethod(
...
-        List<? extends ExpressionTree> args = newClassTree.getArguments();
-        viewpointAdaptExecutable(newClassTree, constructorType, args);
+        atInvocation(newClassTree, constructorType);
     }
 
     /**
      * Viewpoint-adapts a method or constructor invocation.
      *
      * @param tree invocation of the method or constructor
      * @param methodType type of the method or constructor; is side-effected by this method
```
mernst, line 263: Throughout, please clarify whether this is the declared type.  I think it's a copy of the declared type.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -206,26 +233,21 @@ public void viewpointAdaptMethod(
...
-     * @param argTrees the arguments to the method or constructor; subexpressions of {@code tree}
      */
-    private void viewpointAdaptExecutable(
-            ExpressionTree tree,
-            AnnotatedExecutableType methodType,
-            List<? extends ExpressionTree> argTrees) {
+    private void atInvocation(ExpressionTree tree, AnnotatedExecutableType methodType) {
         assert hasDependentAnnotations();
         Element methodElt = TreeUtils.elementFromUse(tree);
         // The annotations on `viewpointAdaptedType` will be copied to `methodType`.
```
mernst, line 268: This variable name is confusing.  I think maybe it is not yet the viewpoint-adapted type, but by the end of the method it will be.  Please clarify.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -235,12 +257,6 @@ private void viewpointAdaptExecutable(
...
 
-        JavaExpression receiver = JavaExpression.getReceiver(tree);
-        List<JavaExpression> argsJe = argumentTreesToJavaExpressions(tree, methodType, argTrees);
-
-        JavaExpressionContext context =
-                new JavaExpressionContext(receiver, argsJe, factory.getChecker());
-
         // methodType cannot be viewpoint adapted directly because it is the type post type variable
         // substitution.  Dependent type annotations on type arguments cannot be
         // viewpoint adapted along with the dependent type annotations that are on the method
```
mernst, line 277: The comment below

> If the type of List.get is viewpoint adapted for the invocation "list.get(0)", then
> methodType would be @KeyFor("map") String get(int).

is written as if it is a counterfactual.  What is its relation to the actual situation?  Please clarify.

Also,

> use the type for the method (viewpointAdaptedType) and viewpoint adapt that
> type.

makes it sound like viewpoint adaptation is being done twice.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -256,198 +272,86 @@ private void viewpointAdaptExecutable(
         // Then copy annotations from the viewpoint adapted type to methodType, if that annotation
         // is not on a type that was substituted for a type variable.
 
-        viewpointAdaptToContext(context, viewpointAdaptedType);
+        StringToJavaExpression stringToJavaExpr;
+        if (tree instanceof MethodInvocationTree) {
+            stringToJavaExpr =
+                    stringExpr ->
+                            StringToJavaExpression.atMethodInvocation(
+                                    stringExpr, (MethodInvocationTree) tree, factory.getChecker());
```
mernst, line 295: Minor:  For some lambdas, you create a local variable with the casted value (maybe to avoid a cast on every execution of the lambda body?), and for some like this one you do not.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -256,198 +272,86 @@ private void viewpointAdaptExecutable(
...
-     * varargs.
+     * Viewpoint-adapts the Java expressions in annotations written on a field declaration for the
+     * use at {@code fieldAccess}.
      *
-     * @param tree invocation of the method or constructor
-     * @param methodType type of the method or constructor
-     * @param argTrees the arguments to the method or constructor; subexpressions of {@code tree}
-     * @return the arguments, as JavaExpressions
+     * @param fieldAccess a field access
+     * @param type its type; is side-effected by this method
```
mernst, line 313: This comment looks wrong.  `type` is not the type of the field access, but the declared type of the field.  Please review the formal parameter descriptions throughout.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -256,198 +272,86 @@ private void viewpointAdaptExecutable(
...
-                }
-                Element varargsElement =
-                        method.getParameters().get(method.getParameters().size() - 1);
-                TypeMirror tm = ElementUtils.getType(varargsElement);
-                result.add(new ArrayCreation(tm, Collections.emptyList(), varargArgs));
-
-                return result;
-            }
+    public void atFieldAccess(MemberSelectTree fieldAccess, AnnotatedTypeMirror type) {
+        if (!hasDependentType(type) || TreeUtils.isClassLiteral(fieldAccess)) {
```
mernst, line 316: Why is the special case needed?  (That is, can `hasDependentType` return true for a class literal?)
If it is needed, would it be more efficient to check it first?

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -256,198 +272,86 @@ private void viewpointAdaptExecutable(
...
-            }
-            AnnotatedTypeMirror lastArg = factory.getAnnotatedType(args.get(args.size() - 1));
-            List<AnnotatedTypeMirror> paramTypes = methodType.getParameterTypes();
-            AnnotatedArrayType lastParam =
-                    (AnnotatedArrayType) paramTypes.get(paramTypes.size() - 1);
-            return lastArg.getKind() != TypeKind.ARRAY
-                    || AnnotatedTypes.getArrayDepth(lastParam)
-                            != AnnotatedTypes.getArrayDepth((AnnotatedArrayType) lastArg);
+        Element ele = TreeUtils.elementFromUse(fieldAccess);
+        if (ele.getKind() != ElementKind.FIELD && ele.getKind() != ElementKind.ENUM_CONSTANT) {
```
mernst, line 321: What else might it be, other than these?  The documentation says it is a field access, but is it actually an arbitrary MemberSelectTree that might not be a field access?

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -256,198 +272,86 @@ private void viewpointAdaptExecutable(
...
+
+        convertAnnotatedTypeMirror(
+                stringExpr ->
+                        StringToJavaExpression.atFieldAccess(
+                                stringExpr, fieldAccess, factory.getChecker()),
+                type);
     }
 
     ///
     /// Standardization
```
mernst, line 333: This is a new term that isn't mentioned in the class documentation.  How is this related to the 3 functionalities that are mentioned in the class documentation?

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -256,198 +272,86 @@ private void viewpointAdaptExecutable(
...
-        }
-
-        TreePath path = factory.getPath(tree);
-        ;
-        parseToPath(path, type);
-    }
-
-    /**
-     * Standardize the Java expressions in annotations in a class declaration.
+     * Viewpoint-adapts the Java expressions in annotations to a class declaration.
```
mernst, line 337: The comment immediately above says this section is about standardization, but here there is more viewpoint-adaptation.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -256,198 +272,86 @@ private void viewpointAdaptExecutable(
...
      *
-     * @param node the class declaration
      * @param type the type of the class declaration; is side-effected by this method
      * @param classElt the element of the class declaration
      */
-    public void standardizeClass(ClassTree node, AnnotatedTypeMirror type, Element classElt) {
```
mernst, line 368: Nit: should "class" be "type" (here and in the method's Javadoc)?

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -566,7 +457,7 @@ public void standardizeExpression(ExpressionTree tree, AnnotatedTypeMirror annot
      * @param type the type to standardize
```
mernst, line 472: It's a bit pedantic, but please note whenever a formal parameter is side-effected by a method.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -593,195 +484,306 @@ public void standardizeVariable(AnnotatedTypeMirror type, Element elt) {
...
                 // example), so there is nothing to do.
                 break;
         }
     }
 
     /**
-     * Parse and standardize the expressions in dependent types in {@code type} as if they were
-     * written at {@code localVarPath}. For example, {@code @KeyFor("field") String} is changed to
-     * {@code @KeyFor("this.field")}
+     * Viewpoint-adapt a type qualifier obtained from a contract using {@code stringToJavaExpr}.
```
mernst, line 513: Why is "from a contract" important?  It does not seem to be used in the method body.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -593,195 +484,306 @@ public void standardizeVariable(AnnotatedTypeMirror type, Element elt) {
...
      *
-     * @param localVarPath the expression is parsed as if it were written at this location
-     * @param type the type to parse; is side-effected by this method
+     * <p>For example, if the contract is {@code @EnsuresKeyFor(value = "this.field", map =
```
mernst, line 515: Equally relevant is that in this case the argument corresponding to formal parameter `annoFromContract` is `@KeyFor("this.map")`, so I would mention that too.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -593,195 +484,306 @@ public void standardizeVariable(AnnotatedTypeMirror type, Element elt) {
...
-        standardizeAtm(localContext, localVarPath, type, /*removeErroneousExpressions=*/ false);
+    public AnnotationMirror viewpointAdaptQualifierFromContract(
+            AnnotationMirror annoFromContract,
+            StringToJavaExpression stringToJavaExpr,
+            @Nullable Tree errorTree) {
+        if (!hasDependentAnnotations()) {
+            return annoFromContract;
+        }
+
+        AnnotationMirror standardized = map(stringToJavaExpr, annoFromContract);
```
mernst, line 534: Methods `map` and `convertAnnotatedTypeMirror` are very similar.  It might be nice for their documentation to mention one another to clarify the relationship, and it might be nice for the method names to be more similar given the method similarities.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -593,195 +484,306 @@ public void standardizeVariable(AnnotatedTypeMirror type, Element elt) {
...
+     * it removes expressions that contain other local variables.
      *
-     * @param context the context
-     * @param type the type to viewpoint-adapt; is side-effected by this method
+     * @param methodDeclTree the method declaration to which the annotations are viewpoint-adapted
+     * @param atm type to viewpoint-adapt; is side-effected by this method
      */
-    private void viewpointAdaptToContext(JavaExpressionContext context, AnnotatedTypeMirror type) {
-        standardizeAtm(context, null, type, /*removeErroneousExpressions=*/ false);
+    public void delocalize(MethodTree methodDeclTree, AnnotatedTypeMirror atm) {
```
mernst, line 552: What is the relationship to `atMethodDeclaration`?  Why isn't this method named `atMethodDeclaration` too?

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -593,195 +484,306 @@ public void standardizeVariable(AnnotatedTypeMirror type, Element elt) {
...
-            boolean removeErroneousExpressions) {
-        this.standardizeTypeAnnotator.init(context, localVarPath, removeErroneousExpressions);
-        this.standardizeTypeAnnotator.visit(type);
+    protected @Nullable AnnotationMirror map(
+            StringToJavaExpression stringToJavaExpr, AnnotationMirror anno) {
+        if (!isExpressionAnno(anno)) {
+            return null;
+        }
+
+        Map<String, List<JavaExpression>> map = new HashMap<>();
```
mernst, line 644: Give `map` a more descriptive name, such as `newElements`.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -593,195 +484,306 @@ public void standardizeVariable(AnnotatedTypeMirror type, Element elt) {
...
+        for (String value : getListOfExpressionElements(anno)) {
```
mernst, line 645: Here is another place that I think `ExecutableElement` would be more appropriate (and more efficient) than `String`.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -593,195 +484,306 @@ public void standardizeVariable(AnnotatedTypeMirror type, Element elt) {
...
+     * viewpoint-adaption have occurred. {@code javaExpr} may be a {@link PassThroughExpression}
+     * that results from an {@link JavaExpressionParseException}.
      *
-     * @param expression a Java expression
-     * @param context the context
-     * @param localVarPath if non-null, the expression is parsed as if it were written at this
-     *     location
-     * @return the standardized version of the Java expression
+     * @param javaExpr a JavaExpression
+     * @return a transformed JavaExpression or {@code null} if no transformation exists
```
mernst, line 678: If null is returned, is the original used, or is the original discarded?  Or is that up to the caller?

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -593,195 +484,306 @@ public void standardizeVariable(AnnotatedTypeMirror type, Element elt) {
...
+        return !DependentTypesError.isExpressionError(expression);
+    }
 
+    /**
+     * Create a new annotation the same type as {@code originalAnno} using the provide {@code
+     * valueMap}.
+     *
+     * @param originalAnno the annotation passed to {@link #map(StringToJavaExpression,
+     *     AnnotationMirror)}
+     * @param valueMap a mapping of element names of {@code originalAnno} to {@code JavaExpression}s
```
mernst, line 706: This parameter would be better named `elementMap` or `elements` or the like.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -593,195 +484,306 @@ public void standardizeVariable(AnnotatedTypeMirror type, Element elt) {
...
-            this.localVarPath = localVarPath;
-            this.removeErroneousExpressions = removeErroneousExpressions;
+        @Override
+        public String toString() {
+            return string;
         }
+    }
+
+    /**
+     * Creates a allows an expression string to be converted to a JavaExpression and then to a
```
mernst, line 753: Grammar.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -793,43 +809,42 @@ public Void visitTypeVariable(AnnotatedTypeMirror.AnnotatedTypeVariable type, Vo
...
 
     /**
      * Checks all Java expressions in the given annotated type to see if the expression string is an
      * error string as specified by {@link DependentTypesError#isExpressionError}. If the annotated
-     * type has any errors, an expression.unparsable.type.invalid error is issued at the errorTree.
+     * type has any errors, an expression.unparsable.type.invalid error is issued at {@code
+     * errorTree}.
      *
      * @param atm annotated type to check for expression errors
      * @param errorTree the tree at which to report any found errors
```
mernst, line 850: This method doesn't do much checking of its own.  What do you think about renaming all the methods like `checkTypes` to names like `reportErrors` which seems more descriptive?

mernst, line 850: The documentation of `errorElements` below says it returns elements, but actually it returns errors.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -975,11 +990,13 @@ public void checkMethod(MethodTree methodDeclTree, AnnotatedExecutableType type)
...
             AnnotatedTypeMirror atm = methodType.getTypeVariables().get(i);
-            viewpointAdaptToContext(context, atm);
+            convertAnnotatedTypeMirror(
+                    stringExpr ->
+                            StringToJavaExpression.atMethodBody(
+                                    stringExpr, node, factory.getChecker()),
+                    atm);
             checkType(atm, node.getTypeParameters().get(i));
         }
     }
```
mernst, line 1002: `isExpressionAnno`'s documentation says "whose value" but I think it means, "if any element's declared type".

Also, can that `for` loop in it be made more efficient?

mernst, line 1002: In `ExpressionErrorChecker`, abstract out the list concatenation into a named method (or find one that already exists).  Also, for efficiency it should check each argument (`r1` and `r2`) with `.isEmpty()` and if so, return the other one without allocating a new list.  When it does need to concatenate, it should size the list appropriately.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -1053,7 +1070,7 @@ protected Void scan(AnnotatedTypeMirror type, AnnotatedTypeMirror p) {
             }
             p.replaceAnnotations(replacement);
             if (type.getKind() != p.getKind()) {
-                // if the underlying types don't match, then this type has be substituted for a
+                // If the underlying types don't match, then this type has be substituted for a
```
mernst, line 1073: Should "has be" be "has been" or "has to be"?

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -1053,7 +1070,7 @@ protected Void scan(AnnotatedTypeMirror type, AnnotatedTypeMirror p) {
...
                 // type variable, so don't recur. The primary annotation was copied because
                 // if the type variable might have had a primary annotation at a use.
```
mernst, line 1075: Grammar "because if".

mernst (changes requested):

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -458,14 +541,12 @@ public JavaExpression visit(NameExpr expr, JavaExpressionContext context) {
...
          *
-         * @param type the type in which to search for {@code identifier}
          * @param identifier possible class name
-         * @return the {@code ClassName} for {@code identifier} or null if it is not a class name
+         * @return the {@code ClassName} for {@code identifier}, or null if it is not a class name
          */
-        protected @Nullable ClassName getIdentifierAsUnqualifiedClassName(
-                TypeMirror type, String identifier) {
+        protected @Nullable ClassName getIdentifierAsUnqualifiedClassName(String identifier) {
             // Is identifier an inner class of this or of any enclosing class of this?
```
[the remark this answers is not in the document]
smillst: I clarified the comment.  There's no `receiverExpr` in this method, so I'm a little confused by your comment.

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -116,67 +107,51 @@
...
-        // The underlying javac API used to convert from Strings to Elements requires a tree path
-        // even when the information could be deduced from elements alone.  So use the path to the
-        // current CompilationUnit.
-        TreePath pathToCompilationUnit = context.checker.getPathToCompilationUnit();
+
         Expression expr;
         try {
             expr = StaticJavaParser.parseExpression(replaceParameterSyntax(expression));
         } catch (ParseProblemException e) {
             throw constructJavaExpressionParseError(expression, "is an invalid expression");
```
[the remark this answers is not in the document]
smillst: - [ ] Could information from e be usefully added to this message?
- [ ] The grammar of the second argument sometimes starts with a verb and sometimes does not. Please make them consistent and document in the constructJavaExpressionParseError method what the grammar should be.
- [X] Also in constructJavaExpressionParseError, should the Error be BugInCF?

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
smillst: The point was that this was the method to call to standardize contract annotations. I've moved this method to `Contract` class and renamed it.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
smillst: At call sites, I think the name `checkType` makes more sense.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
smillst: This comment is correct.  It's type of the field access, but the annotations are those written at the declaration of the field.  For example:
```
class MyClass<T> {
  Map map = ..;
  @KeyFor("this.map") List<T> field;
}

void method(MyClass<String> myClass) {
  myClass.field;
}
```
The declared type of `field` is `@KeyFor("this.map) List<T>`.  
For the call to `atFieldAccess`, the value of `type` is `@KeyFor("this.map") List<String>`.

I'm not sure how to make this clear in the Javadoc.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -22,51 +22,51 @@
...
-import org.checkerframework.javacutil.SystemUtil;
 import org.checkerframework.javacutil.TreePathUtil;
 import org.checkerframework.javacutil.TreeUtils;
+import org.checkerframework.javacutil.TypesUtils;
 import org.plumelib.util.StringsPlume;
 
 /**
  * A class that helps checkers use qualifiers that are represented by annotations with Java
- * expression strings. This class performs four main functions:
+ * expression strings. This class performs three main functions:
```
mernst, line 62: There is another functionality, standardization, which differs from viewpoint-adaptation and I think deserves its own section here.
Also, please add section headers in this file to show where each part of it is implemented.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -76,37 +76,45 @@
...
      */
-    private final ExpressionErrorChecker expressionErrorChecker;
+    private final ExpressionErrorCollector expressionErrorCollector =
+            new ExpressionErrorCollector();
 
     /**
-     * A scanner that standardizes Java expression strings in dependent type annotations. Call
-     * {@code StandardizeTypeAnnotator#init(JavaExpressionContext, TreePath, boolean, boolean)}
-     * before each use.
+     * A scanner that applies a function to each {@link AnnotationMirror} in a given {@code
```
mernst, line 96: Please make explicit whether this returns a new one or side-effects an existing one.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -116,13 +124,18 @@ public DependentTypesHelper(AnnotatedTypeFactory factory) {
...
                 annoToElements.put(expressionAnno, elementList);
             }
         }
-        this.expressionErrorChecker = new ExpressionErrorChecker();
-        this.standardizeTypeAnnotator = new StandardizeTypeAnnotator();
-        this.viewpointAdaptedCopier = new ViewpointAdaptedCopier();
+        this.hasDependentTypeScanner =
+                new SimpleAnnotatedTypeScanner<>(
+                        (type, unused) ->
+                                type.getAnnotations().stream().anyMatch(this::isExpressionAnno),
```
mernst, line 131: Streams are likely to be less efficient than a for loop.  That is a small thing, but this is likely to get called a lot.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -149,304 +162,209 @@ public boolean hasDependentAnnotations() {
...
-    /// Viewpoint adaptation
-    ///
+    /**
+     * Returns the elements of the annotation that are Java expressions.
+     *
+     * @param am AnnotationMirror
+     * @return the elements of the annotation that are Java expressions
+     */
+    private List<String> getListOfExpressionElements(AnnotationMirror am) {
+        for (Class<? extends Annotation> clazz : annoToElements.keySet()) {
```
mernst, line 178: Why isn't this `annoToElements.get(am.getClass())`?  `Class` objects are interned.  Is it because the AnnotationMirror might be built by javac or by the Checker Framework, or for some other reason?  Another possibility is to have two maps, one with Class as the key and one with String as the key.

There are other uses of this map that could also perhaps be optimized.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -149,304 +162,209 @@ public boolean hasDependentAnnotations() {
...
      *
      * @param classDecl class or interface declaration whose type variables should be viewpoint
      *     adapted
-     * @param bounds annotated types of the bounds of the type variables; side-effected by this
-     *     method
+     * @param bounds annotated types of the bounds of the type variables; its elements are
+     *     side-effected by this method (but the list itself is not side-effected)
      */
-    public void viewpointAdaptTypeVariableBounds(
+    public void atTypeVariableBounds(
```
mernst, line 204: This is an odd name because it viewpoint-adapts to a use, not to the bounds.  (It's the bounds that are being viewpoint-adapted.)  It calls `atTypeDecl` internally, and the mismatch between the method name and what it calls is also surprising.  Can you give it a more evocative name?

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -149,304 +162,209 @@ public boolean hasDependentAnnotations() {
...
+    }
 
     /**
      * Viewpoint-adapts the dependent type annotations on the bounds to the use of the type.
      *
      * @param classDecl class or interface declaration whose type variables should be viewpoint
      *     adapted
-     * @param bounds annotated types of the bounds of the type variables; side-effected by this
-     *     method
+     * @param bounds annotated types of the bounds of the type variables; its elements are
```
mernst, line 201: Try to use "type parameter" or "type argument" rather than "type variable", which is ambiguous between the two.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -149,304 +162,209 @@ public boolean hasDependentAnnotations() {
...
-            viewpointAdaptToContext(context, bound.getUpperBound());
-            viewpointAdaptToContext(context, bound.getLowerBound());
+            convertAnnotatedTypeMirror(stringToJavaExpr, bound.getUpperBound());
+            convertAnnotatedTypeMirror(stringToJavaExpr, bound.getLowerBound());
         }
     }
 
     /**
-     * Viewpoint-adapts the dependent type annotations in the methodDeclType based on the
+     * Viewpoint-adapts the dependent type annotations in the methodType based on the
```
mernst, line 221: The documentation uses the terms "based on", "for", and "to".  Please make them consistent.  I like "to" myself, as in "viewpoint-adapt T to its declaration" and that is already used in parts of the documentation, but I'm open to other possibilities if they are better.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -149,304 +162,209 @@ public boolean hasDependentAnnotations() {
...
             return;
         }
 
-        JavaExpression receiver = JavaExpression.getReceiver(tree);
-        List<JavaExpression> argsJe = argumentTreesToJavaExpressions(tree, methodType, argTrees);
-
-        JavaExpressionContext context =
-                new JavaExpressionContext(receiver, argsJe, factory.getChecker());
-
         // methodType cannot be viewpoint adapted directly because it is the type post type variable
```
mernst, line 276: I'm not sure what "direct" viewpoint-adaptation is; please clarify.

I am also confused about whether this comment is explaining the overall design (the "expect for dependent types" comment in the Javadoc above), the code below, or both.  Relatedly, I'm not sure of the relationship between the first and second sentences here.  Please clarify.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -149,304 +162,209 @@ public boolean hasDependentAnnotations() {
...
-    /**
-     * Standardizes the Java expressions in annotations on a constructor invocation.
-     *
-     * @param tree the constructor invocation
-     * @param type the type of the expression; is side-effected by this method
+     * @param fieldAccess a field access
+     * @param type its type; is side-effected by this method
      */
-    public void standardizeNewClassTree(NewClassTree tree, AnnotatedDeclaredType type) {
+    public void atFieldAccess(MemberSelectTree fieldAccess, AnnotatedTypeMirror type) {
```
mernst, line 318: This is a standardization routine.  I think than starting a viewpoint-adaptation routine's name with "at" is appropriate, but for a standardization routine it is misleading and does not make a distinction between the two types of functionality.  I suggest changing the standardization routine names; even adding "standardize" at the beginning of them would be clearer.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -457,10 +375,10 @@ public void delocalize(MethodTree methodDeclTree, AnnotatedTypeMirror atm) {
      * Standardize the Java expressions in annotations in a variable declaration.
      *
      * @param declarationTree the variable declaration
-     * @param type the type of the variable declaration
+     * @param type the type of the variable declaration; is side-effected by this method
      * @param variableElt the element of the variable declaration
      */
-    public void standardizeVariable(
+    public void atVariableDeclaration(
             Tree declarationTree, AnnotatedTypeMirror type, Element variableElt) {
```
mernst, line 378: Why is the formal parameter an Element rather than a VariableElement?

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -592,195 +497,285 @@ public void standardizeVariable(AnnotatedTypeMirror type, Element elt) {
...
+                            JavaExpression.getFormalParameters(methodElement);
+                    List<JavaExpression> paramsAsLocals =
+                            JavaExpression.getParametersAsLocalVariables(methodElement);
+                    JavaExpressionConverter jec =
+                            new JavaExpressionConverter() {
+                                @Override
+                                protected JavaExpression visitLocalVariable(
+                                        LocalVariable localVarExpr, Void unused) {
+                                    int index = paramsAsLocals.indexOf(localVarExpr);
+                                    if (index == -1) {
```
mernst, line 548: This method could throw an exception that is caught at the call to `jec.convert`.
That has the small advantages of not having to create a new object which will never be used for anything (other than being tested against), and if there is a problem it is signalled right away rather than continuing to visit the expression.

If you think that is too ugly, we can keep the code as is.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -592,195 +497,285 @@ public void standardizeVariable(AnnotatedTypeMirror type, Element elt) {
...
+                    return result.containsUnknown() ? null : result;
+                };
+        convertAnnotatedTypeMirror(stringToJavaExpr, atm);
     }
 
     /**
-     * Viewpont-adapt the dependent types in {@code type} using the {@code context} provided.
+     * Calls {@link #convertAnnotationMirror(StringToJavaExpression, AnnotationMirror)} on each
+     * annotation mirror on type with {@code stringToJavaExpr}. And replaces the annotation with the
+     * one created by {@code convertAnnotationMirror}, it it's not null. See {@link
```
mernst, line 568: "it it's" => "if it's".

Please document what happens if it *is* null.  I think the value is dropped, though on my reading of this Javadoc it led me to believe that the original value would be retained unchanged.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -592,195 +497,285 @@ public void standardizeVariable(AnnotatedTypeMirror type, Element elt) {
...
+     * returned annotation.
+     *
+     * <p>Instead of overriding this method, subclasses can override the following methods to change
+     * the behavior of this class:
      *
-     * @param context context to use
-     * @param localVarPath the expression is parsed as if it were written at this location
-     * @param type the type to viewpoint-adapt; is side-effected by this method
+     * <ul>
+     *   <li>{@link #shouldParseExpression(String)}: to control which expressions are skipped. If
```
mernst, line 592: This name is a bit too implementation-focused.  They key is that the expression string is included in the new annotation unchanged.  Please change the name to reflect that semantics.  Probably you will change its sense (true to false and false to true) as well.  You could use "isPassThroughExpression" or some other name.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -592,195 +497,285 @@ public void standardizeVariable(AnnotatedTypeMirror type, Element elt) {
...
-            boolean removeErroneousExpressions) {
-        this.standardizeTypeAnnotator.init(context, localVarPath, removeErroneousExpressions);
-        this.standardizeTypeAnnotator.visit(type);
+    protected @Nullable JavaExpression transform(JavaExpression javaExpr) {
+        return javaExpr;
     }
 
     /**
-     * Standardizes a Java expression.
+     * Whether or not {@code expression} should be parsed. The default implementation returns false
```
mernst, line 657: This Javadoc will need to be revised when the method is renamed.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -592,195 +497,285 @@ public void standardizeVariable(AnnotatedTypeMirror type, Element elt) {
...
-            JavaExpressionContext context,
-            @Nullable TreePath localVarPath,
-            AnnotationMirror anno,
-            boolean removeErroneousExpressions) {
-        if (!isExpressionAnno(anno)) {
-            return null;
-        }
-
+    protected AnnotationMirror buildAnnotation(
+            AnnotationMirror originalAnno, Map<String, List<JavaExpression>> elementMap) {
```
mernst, line 683: This implementation looks like it works if all of an annotation's elements are Java expressions.  But what if some elements are Java expressions and others are not?  The others seem to be dropped by this implementation.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -592,195 +497,285 @@ public void standardizeVariable(AnnotatedTypeMirror type, Element elt) {
...
+    }
+
+    /**
+     * Allows an expression string to be converted to a JavaExpression and then to a string without
+     * parsing the string.
+     *
+     * @param string some string
+     * @return a {@link PassThroughExpression}
+     */
+    protected PassThroughExpression createPassThroughString(String string) {
```
mernst, line 729: Why is this method needed?  That is, I think it could be made a constructor of `PassThroughExpression` and its single use be made a call to that constructor.  I don't see a need for a subclass to override it.

mernst (changes requested):

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -974,18 +960,20 @@ public void checkMethod(MethodTree methodDeclTree, AnnotatedExecutableType type)
...
     }
 
     /**
-     * Returns true if {@code am} is an expression annotation, that is an annotation whose value is
-     * a Java expression.
+     * Returns true if {@code am} is an expression annotation, that is an annotation whose element
+     * is a Java expression.
      *
      * @param am an annotation
      * @return true if {@code am} is an expression annotation
```
mernst, line 978: This method could be made much more efficient by looking up in the map rather than iterating over it.  (Or create another map or set of the string names and look up in that.)

mernst (changes requested):

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
smillst: `am.getClass()` returns `AnnotationMirror`.  `AnnotationMirror` doesn't have a field that stores the `Class` of the annotation it represents.  The key of this map should be the annotation's name (as a String) rather than a class.  We are also going to refactor this map so that the values are elements after this pull request is complete.  There's a todo at the declaration of this field.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -792,43 +779,42 @@ public Void visitTypeVariable(AnnotatedTypeMirror.AnnotatedTypeVariable type, Vo
...
             visitedNodes.put(type, r);
             return r;
         }
 
         @Override
-        protected Void scan(AnnotatedTypeMirror type, Void aVoid) {
+        protected Void scan(
+                AnnotatedTypeMirror type, Function<AnnotationMirror, AnnotationMirror> func) {
             for (AnnotationMirror anno :
                     AnnotationUtils.createAnnotationSet(type.getAnnotations())) {
```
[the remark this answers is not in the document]
smillst: The set returned by `type.getAnnotations()` is modified by ` type.removeAnnotation(anno);` .

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -838,7 +824,7 @@ public void checkType(AnnotatedTypeMirror atm, Tree errorTree) {
             return;
         }
 
-        List<DependentTypesError> errors = expressionErrorChecker.visit(atm);
+        List<DependentTypesError> errors = expressionErrorCollector.visit(atm);
         if (errors == null || errors.isEmpty()) {
             return;
         }
```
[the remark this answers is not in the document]
smillst: `errorTree` is never a `MethodTree`.  We could probably do something similar for `AnnotatedTypeTree` which are passed to this method, but I suggest we wait until after this pull request is merged.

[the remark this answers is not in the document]
smillst: This can be fixed when the type of `annoToElements` is changed. (See the TODO at the declaration of `annoToElements`.)

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
smillst: This can be fixed when the type of `annoToElements` is changed. (See the TODO at the declaration of `annoToElements`.)

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
smillst: `reportError` sounds like an error will definitely be reported, but most of the time an error isn't reported. The `checkType` methods are called in `BaseTypeVisitor`.   `BaseTypeVisitor` methods follow the same `checkSomething` such as `checkOverride` or `checkPostcondition`.   I can add `ForErrorExpressions` to the method names.

mernst: OK.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
smillst: This is not a standardization routine.  It viewpoint-adapts annotations from the declaration of the field to the field access.  For example, `type` is `@KeyFor("this.map") List<String>` and `fieldAccess` is `myClass.field`, then this method changes the KeyFor annotation to `@KeyFor("myClass.map")`.

I'm still thinking about renaming the methods that do not viewpoint-adapt.

smillst: >I think than starting a viewpoint-adaptation routine's name with "at" is appropriate, but for a standardization routine it is misleading and does not make a distinction between the two types of functionality. I suggest changing the standardization routine names; even adding "standardize" at the beginning of them would be clearer.

All the `at` methods preform standardization and most of them viewpoint-adapt.  Only `atTypeDecl` does no viewpoint adaption.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -1083,9 +1116,9 @@ protected Void combineRs(Void r1, Void r2) {
     }
```
mernst, line 1116: Just above in `combineRs`, I was trying to figure out the relationship between `p` and `R`.  I think they are unrelated.

In `AnnotatedTypeComparer`, could you document `R`?  I think it's the type of the value returned by the scanner.  In that case, `combineRs` would be better renamed to `combineResults` or the like.  Please do all that in a separate pull request from this one, and it will probably land before this one.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -1094,32 +1127,11 @@ private boolean hasDependentType(AnnotatedTypeMirror atm) {
...
-     */
-    private List<String> getListOfExpressionElements(AnnotationMirror am) {
-        for (Class<? extends Annotation> clazz : annoToElements.keySet()) {
-            if (factory.areSameByClass(am, clazz)) {
-                return annoToElements.get(clazz);
-            }
-        }
-        return Collections.emptyList();
+        // This is a test about this specific type.
+        return hasDependentTypeScanner.visit(atm);
```
mernst, line 1135: I would move the definition of `hasDependentTypeScanner` next to this method, since this is the only place it is used.

This implementation seems costly and it seems to be called a lot.  I suggest creating a cache and looking up values in the cache.  I guess the cache would map from class name as a string to boolean.

mernst (changes requested):

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -22,91 +22,125 @@
...
-import org.checkerframework.javacutil.ElementUtils;
 import org.checkerframework.javacutil.SystemUtil;
 import org.checkerframework.javacutil.TreePathUtil;
 import org.checkerframework.javacutil.TreeUtils;
+import org.checkerframework.javacutil.TypesUtils;
 import org.plumelib.util.StringsPlume;
 
 /**
  * A class that helps checkers use qualifiers that are represented by annotations with Java
  * expression strings. This class performs four main functions:
```
mernst, line 63: "four" is still inaccurate.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -22,91 +22,125 @@
...
- *       invocations.
- *   <li>Changes invalid expression strings to an error string that includes the reason why the
- *       expression is invalid. For example, {@code @KeyFor("m")} would be changed to
- *       {@code @KeyFor("[error for expression: m error: m: identifier not found]")} if m is not a
- *       valid identifier. This allows subtyping checks to assume that if two strings are equal and
- *       not errors, the reference the same valid Java expression.
+ *   <li>Converts the expressions strings in an {@link AnnotationMirror} {@code am}, by creating a
+ *       new annotation whose Java expression elements are the result of the converson. See {@link
+ *       #convertAnnotationMirror(StringToJavaExpression, AnnotationMirror)}. Subclasses can
+ *       specialize this process by overriding methods in this class. Conversion include:
```
mernst, line 69: Does "conversion include:" mean these are 3 types of conversion, or 3 steps of each conversion?  Please clarify, and fix the grammar in "conversion include".

mernst (changes requested):

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
smillst: > This implementation seems costly and it seems to be called a lot. I suggest creating a cache and looking up values in the cache. I guess the cache would map from class name as a string to boolean.

This is scanning an AnnotatedTypeMirror, so the key of such a map would have to be to be the AnnotatedTypeMirror itself.  Calling hashcode on an AnnotatedTypeMirror is just as inefficient as this annotated type scanner.

Or did you mean the call to `isExpressionAnno` in the scanner itself?  In that case, the implementation of is expression uses a map.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -149,319 +183,228 @@ public boolean hasDependentAnnotations() {
...
-     * Viewpoint-adapt all dependent type annotations to the method declaration, {@code
-     * methodDeclTree}. This changes occurrences of formal parameter names to "#2" syntax, and it
-     * removes expressions that contain other local variables.
-     *
-     * @param methodDeclTree the method declaration to which the annotations are viewpoint-adapted
-     * @param atm type to viewpoint-adapt; is side-effected by this method
-     */
-    public void delocalize(MethodTree methodDeclTree, AnnotatedTypeMirror atm) {
-        if (!hasDependentType(atm)) {
+    public void atTypeDecl(AnnotatedTypeMirror type, TypeElement typeElt) {
```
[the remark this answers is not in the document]
smillst: I agree.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -149,319 +183,228 @@ public boolean hasDependentAnnotations() {
...
-    public TreeAnnotator createDependentTypesTreeAnnotator() {
-        assert hasDependentAnnotations();
-        return new DependentTypesTreeAnnotator(factory, this);
-    }
+        // methodType is @KeyFor("this.map") String get(int)
+        // "this.map" must not be viewpoint-adapted to the invocation because it is not from
+        // the method declaration, but added during type variable substitution.
+        //
+        // So this implementation gets the declared type of the method, declaredMethodType,
+        // viewpoint-adapts all dependent type annotations in declaredMethodType to the call site,
```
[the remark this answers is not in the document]
smillst: Yes.  I'm not sure how to make this clearer in the comment.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -518,36 +469,13 @@ public void standardizeVariable(
...
-        }
-
-        JavaExpression receiver = JavaExpression.fromTree(node.getExpression());
-        JavaExpressionContext context = new JavaExpressionContext(receiver, factory.getChecker());
-        viewpointAdaptToContext(context, type);
-    }
-
-    /**
-     * Standardize the Java expressions in annotations in the type of an expression.
+     * Standardize the Java expressions in annotations in the type of an expression. Converts the
```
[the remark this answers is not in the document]
smillst: I've tried to clarify the comment here and also the class Javadoc.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -592,264 +523,346 @@ public void standardizeVariable(AnnotatedTypeMirror type, Element elt) {
...
-
-        JavaExpression receiver = JavaExpression.getPseudoReceiver(localVarPath, enclosingType);
-
-        JavaExpressionContext localContext =
-                new JavaExpressionContext(
-                        receiver,
-                        JavaExpression.getParametersOfEnclosingMethod(localVarPath),
-                        factory.getChecker());
-        standardizeAtm(localContext, localVarPath, type, /*removeErroneousExpressions=*/ false);
+    public void delocalize(MethodTree methodDeclTree, AnnotatedTypeMirror atm) {
```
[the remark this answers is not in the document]
smillst: >Why isn't this called "atMethodDeclaration"? Is it because it "removes expressions that contain other local variables" rather than turning them into errors?

Yes, also, `StringToJavaExpression#atMethodDecl` does not remove local variables, so rename `delocalize` to `atMethodDecl` would be confusing.

It seems different to me than the other `at` methods because it is removing expressions.  Maybe `fromMethodBodyToMethodDecl`?

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
smillst: I've taken a note to do this.

mernst:
Could you write the changelog entry?

mernst (approved):

PR #4383, opened 2021-03-08
add a note to the manual about how to use -A options with WPI

kelloggm:


docs/manual/inference.tex
```diff
@@ -188,7 +188,9 @@
   modification.  One argument is required: \<-\relax-checker>, which indicates
   what type-checker(s) to run (in the format described in Section~\ref{shorthand-for-checkers}).
   The \ahreforurl{https://github.com/kelloggm/do-like-javac}{documentation of do-like-javac}
-  describes the other commands that its WPI tool supports.
+  describes the other commands that its WPI tool supports. Notably, to pass checker-specific
+  arguments---such as any necessary \<-A> arguments---to invocations of \<javac>,
+  use the \<-\relax-extraJavacArgs> argument to \<dljc>.
```
mernst, line 193: Please give either a concrete example of how to use, or a pointer to where such an example appears.

mernst (changes requested):

docs/manual/inference.tex
kelloggm: I've added an example. Can you let me know if this is what you had in mind?

mernst (approved):
Thanks!  This looks good.

PR #4387, opened 2021-03-09
Reduce scope of try block

mernst:
This is an independent refactoring that reduces the size of the diffs in #4378.


smillst (approved):
Thanks!

PR #4388, opened 2021-03-09
Clean up parsing field access expressions.

smillst:
Also, fixes a bug when referencing a field declared in an outer type from an inner type expression.


framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -522,51 +522,80 @@ public JavaExpression visit(NameExpr expr, JavaExpressionContext context) {
...
-                    throw new ParseRuntimeException(
-                            constructJavaExpressionParseError(
-                                    identifier,
-                                    "a non-static field can't be referenced from a static inner class or enum"));
-                }
-                return fieldAccess;
+            if (fieldElem == null) {
+                // field not found.
+                return null;
             }
```
mernst, line 560: This can be moved up one line, inside the preceding "}".

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -522,51 +522,80 @@ public JavaExpression visit(NameExpr expr, JavaExpressionContext context) {
...
+                        constructJavaExpressionParseError(
+                                identifier,
+                                "field declared in an outer type cannot be accessed from an inner type"));
+            }
+            TypeElement receiverTypeElement = TypesUtils.getTypeElement(receiverExpr.getType());
+            if (receiverTypeElement == null || ElementUtils.isStatic(receiverTypeElement)) {
+                throw new ParseRuntimeException(
+                        constructJavaExpressionParseError(
+                                identifier,
+                                "a non-static field declared in an outer class cannot be referenced from a static inner class or enum"));
```
mernst, line 593: This message uses "outer class" whereas the one above used "outer type"; make them consistent.

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -522,51 +522,80 @@ public JavaExpression visit(NameExpr expr, JavaExpressionContext context) {
...
+            if (fieldDeclaredInReceiverType) {
+                // It's an instance field declared in the type of receiverExpr.
+                TypeMirror fieldType = ElementUtils.getType(fieldElem);
+                return new FieldAccess(receiverExpr, fieldType, fieldElem);
+            }
+            if (!(receiverExpr instanceof ThisReference)) {
+                throw new ParseRuntimeException(
+                        constructJavaExpressionParseError(
+                                identifier,
+                                "field declared in an outer type cannot be accessed from an inner type"));
```
mernst, line 586: How do we know this is a field declared in an outer type?
In particular, it seems to me that this code could be reached by
```java
class A {
  int f;
}
class B extends A {}

B myB;
... myB.f ...
```

because `f` is not declared in the receiver type and the receiver is not a `ThisReference`.  Am I missing something?

mernst (changes requested):

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
smillst: In your example, `enclosingTypeOfField == receiverExpr.getType()`, so this code isn't reached.  `resolver.findField(...)` search in supertypes. I added documentation to make this clearer.

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
```diff
@@ -522,51 +522,84 @@ public JavaExpression visit(NameExpr expr, JavaExpressionContext context) {
...
-                fieldElem = resolver.findField(identifier, receiverType, pathToCompilationUnit);
-            }
-            if (fieldElem == null) {
+            // Find the field element.
+            TypeMirror enclosingTypeOfField = receiverExpr.getType();
+            VariableElement fieldElem;
+            if (identifier.equals("length") && enclosingTypeOfField.getKind() == TypeKind.ARRAY) {
+                fieldElem =
+                        resolver.findField(identifier, enclosingTypeOfField, pathToCompilationUnit);
+                if (fieldElem == null) {
```
mernst, line 543: How can this happen?  Every array type should have a length field.  If this test is retained, I think it should throw BugInCF rather than returning null.

framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java
mernst: (Ah, the added documentation is for Resolver.findField.  Yes, that is helpful.)

mernst (changes requested):

mernst (approved):
Thanks!

PR #4389, opened 2021-03-10
Add option to wpi-many.sh to ignore repos that aren't buildable rather than deleting them

kelloggm:
@mernst I know I used the term "buildable" in the documentation, and that you hate that term, but I couldn't come up with a better way to express the concept. I did define it explicitly this time, but please do suggest an alternative phrasing if you can think of one.


checker/bin/wpi-many.sh
```diff
@@ -203,17 +209,24 @@ do
     RESULT_LOG="${OUTDIR}-results/${REPO_NAME_HASH}-wpi.log"
     touch "${RESULT_LOG}"
 
-    /bin/bash -x "${SCRIPTDIR}/wpi.sh" -d "${REPO_FULLPATH}" -t "${TIMEOUT}" -g "${GRADLECACHEDIR}" -- "$@" &> "${RESULT_LOG}" &> "${OUTDIR}-results/wpi-out" || cat "${OUTDIR}-results/wpi-out"
+    if [ -f "${REPO_FULLPATH}/.cannot-run-wpi" ] && [ "${SKIP_OR_DELETE_UNUSABLE}" = "skip" ]; then
+      echo "Skipping ${REPO_NAME_HASH} because it has a .cannot-run-wpi file present, indicating that an earlier run of WPI failed. To try again, delete the .cannot-run-wpi file and re-run the script."
+    else
+      /bin/bash -x "${SCRIPTDIR}/wpi.sh" -d "${REPO_FULLPATH}" -t "${TIMEOUT}" -g "${GRADLECACHEDIR}" -- "$@" &> "${RESULT_LOG}" &> "${OUTDIR}-results/wpi-out" || cat "${OUTDIR}-results/wpi-out"
```
mernst, line 215: If .cannot-run-wpi exists and SKIP_OR_DELETE_UNUSABLE=delete, this is run.
I think it should be executed only if .cannot-run-wpi does not exist.

docs/manual/inference.tex
```diff
@@ -316,6 +316,14 @@
   is motivated by
   \ahref{https://github.com/gradle/gradle/issues/1319}{Gradle issue \#1319}.
 
+\item[-s]
+  If this flag is present, then projects which are not buildable---for which
```
mernst, line 320: Nit: The rest of the manual puts a space around em-dashes, so this should do so as well, for consistency.

docs/manual/inference.tex
```diff
@@ -316,6 +316,14 @@
...
+  no supported build file is present or for whom running the standard build
```
mernst, line 321: "whom" => "which"

docs/manual/inference.tex
```diff
@@ -316,6 +316,14 @@
...
+  commands fail---are skipped on future runs but are NOT deleted immediately
+  (which is the default behavior for non-buildable projects). This flag is useful
```
mernst, line 323: Rather than "default behavior", say, "if this flag is not present".  That is more explicit.

mernst (changes requested):
I have one concern, and some small comments about the documentation.

PR #4396, opened 2021-03-10
Issue expression.unparsable.type.invalid rather than flowexpr.parse

smillst:
When a parameter is referenced anywhere in a contract expression.

### DIFF
--- a/checker/tests/nullness/ParameterExpression.java
+++ b/checker/tests/nullness/ParameterExpression.java
@@ -45,32 +45,32 @@ public class ParameterExpression {
     }
 
     @EnsuresNonNull("param")
-    // :: error: (expression.parameter.name.invalid)
+    // :: error: (flowexpr.parse.error)
     public void m6a(Object param) {
         param = new Object();
     }
 
     @EnsuresNonNull("param")
-    // :: error: (expression.parameter.name.invalid)
+    // :: error: (flowexpr.parse.error)
     public void m6b(Object param) {
         // :: error: (assignment.type.incompatible)
         param = null;
     }
 
     @EnsuresNonNull("param")
-    // :: error: (expression.parameter.name.invalid)
+    // :: error: (flowexpr.parse.error)
     public void m6c(@Nullable Object param) {
         param = new Object();
     }
 
     @EnsuresNonNull("param")
-    // :: error: (expression.parameter.name.invalid)
+    // :: error: (flowexpr.parse.error)
     public void m6d(@Nullable Object param) {
         param = null;
     }
 
     @EnsuresNonNull("param.toString()")
-    // :: error: (expression.parameter.name.invalid)
+    // :: error: (flowexpr.parse.error)
     public void m6e(@Nullable Object param) {
         param = null;
     }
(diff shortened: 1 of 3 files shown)


framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -974,6 +976,10 @@ private void reportPurityError(String msgKeyPrefix, Pair<Tree, String> r) {
         }
     }
 
+    /** Matches warnings about use of formal parameter name. */
```
mernst, line 979: It would be helpful to say what this is related to (I think it's `constructJavaExpressionParseError`).
Consider defining `destructJavaExpressionParseError` there that returns either a tuple of information or null.  That would keep related logic together.

Here is a different approach that I think would be better.  `constructJavaExpressionParseError` could take, as an argument, the message key.  Then it can use more specific key and can use different numbers of arguments rather than putting all information in a string that needs to be parsed later.
Currently, the only two message keys used with JavaExpressionParseException are "flowexpr.parse.index.too.big" and "flowexpr.parse.error", but there could be more.  That would permit them to be more specific and would lead to more readable error messages and better handling in catch blocks.
This pull request uses "expression.parameter.name.invalid", which could be used deeper in the code.

(I'm sorry that this leads to more refactoring, but I think it will be an improvement overall.)

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1010,15 +1016,36 @@ private void checkContractsAtMethodDeclaration(
             try {
                 exprJe = JavaExpressionParseUtil.parse(expressionString, jeContext);
             } catch (JavaExpressionParseException e) {
-                exprJe = null;
-                checker.report(methodTree, e.getDiagMessage());
+
+                Matcher m =
+                        FORMAL_PARAM_NAME_WARNING_PATTERN.matcher(
+                                e.getDiagMessage().getArgs()[0].toString());
```
mernst, line 1022: This assumes the array is never empty (there is always a 0th element).  That is true currently but might not be in the future.  Using a more specific error message key would solve this, or you could check the array size before doing this.

mernst (changes requested):

checker/tests/nullness/ParameterExpression.java
```diff
@@ -46,33 +46,35 @@ public void m5() {
...
 
     @EnsuresNonNull("param")
     // :: error: (flowexpr.parse.error)
-    // :: warning: (expression.parameter.name.invalid)
     public void m6d(@Nullable Object param) {
         param = null;
     }
 
+    @EnsuresNonNull("param.toString()")
+    // :: error: (flowexpr.parse.error)
```
smillst, line 73: The full error is:

ParameterExpression.java:74: error: [flowexpr.parse.error] cannot parse the expression 'param.toString()' in the postcondition EnsuresNonNull on the declaration of method 'm6e': Invalid 'param' because Use "#1" rather than "param"

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
smillst: I didn't take this approach, but I changed it so there isn't any string parsing.  I agree that it would be nice to issue different parse errors, but perhaps we can do that later.

mernst (approved):

PR #4403, opened 2021-03-11
Use Array#asList for calls to createCombinedSameLen.

smillst:


checker/src/main/java/org/checkerframework/checker/index/samelen/SameLenTransfer.java
```diff
@@ -99,11 +100,9 @@ private Node getLengthReceiver(Node lengthNode) {
                                     .getAnnotationInHierarchy(UNKNOWN);
 
                     AnnotationMirror combinedSameLen =
-                            // In Java 9, this can be:
-                            // aTypeFactory.createCombinedSameLen(
-                            //         List.of(targetRec, otherRec), List.of(lengthNodeAnnotation));
                             aTypeFactory.createCombinedSameLen(
-                                    targetRec, otherRec, UNKNOWN, lengthNodeAnnotation);
+                                    Arrays.asList(targetRec, otherRec),
```
mernst, line 104: This is slightly less efficient than the code that it replaces, because `Arrays.asList` creates an array and then a list, whereas the existing code only creates a list.

mernst (approved):
You can merge this if you want, but note that it contains a small inefficiency.

checker/src/main/java/org/checkerframework/checker/index/samelen/SameLenTransfer.java
smillst: Is it really less efficient?
* `new ArrayList(2)` creates an array list that creates an array of size 2
* `Arrays.asList( targetRec, otherRec)` creates an array of size 2, then creates a list backed by that array.

`List.of(targetRec, otherRec)` would be more efficient than either of the two because it doesn't create an array at all.

mernst: I think you are right (that is, the efficiency is the same but your version is much more concise).  I was looking at the implementation in `Arrays.java`, but that must not be the real one; the actual implementation must do magic.

smillst: The implementation in Arrays.java is:
```
    public static <T> List<T> asList(T... a) {
        return new ArrayList<>(a);
    }
```
Where ArrayList is *not* `java.util.ArrayList`, but rather `java.uti.Arrays.ArrayList`.

PR #4406, opened 2021-03-12
Make `removeOffset` more efficient: don't convert to and from sequences

mernst:
This pull request ended up being much larger than I anticipated to make `removeOffset` more efficient.

A question brought up by these changes is how OffsetEquation objects are treated by the implementation.  Are they ever changed after being put in a `UBQualifier`, or is it safe to share them between `UBQualifier`s?

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/index/upperbound/UBQualifier.java
+++ b/checker/src/main/java/org/checkerframework/checker/index/upperbound/UBQualifier.java
@@ -1,5 +1,6 @@
 package org.checkerframework.checker.index.upperbound;
 
+import java.lang.annotation.Annotation;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -22,6 +23,7 @@ import org.checkerframework.dataflow.qual.Pure;
 import org.checkerframework.framework.type.AnnotatedTypeMirror;
 import org.checkerframework.javacutil.AnnotationBuilder;
 import org.checkerframework.javacutil.AnnotationUtils;
+import org.checkerframework.javacutil.BugInCF;
 import org.checkerframework.javacutil.Pair;
 
 /**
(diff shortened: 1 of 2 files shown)


checker/src/main/java/org/checkerframework/checker/index/upperbound/UBQualifier.java
```diff
@@ -409,41 +533,24 @@ public AnnotationMirror convertToSubstringIndexAnnotation(ProcessingEnvironment
...
-                builder = new AnnotationBuilder(env, LTOMLengthOf.class);
+            } else if (annoClass == LTOMLengthOf.class) {
                 builder.setValue("value", sequences);
-            } else {
-                builder = new AnnotationBuilder(env, LTLengthOf.class);
+            } else if (annoClass == LTLengthOf.class) {
                 builder.setValue("value", sequences);
                 builder.setValue("offset", offsets);
-            }
+            } else throw new BugInCF("What annoClass? " + annoClass);
```
kelloggm, line 553: nit: this should be enclosed in `{}`

kelloggm (approved):
> A question brought up by these changes is how OffsetEquation objects are treated by the implementation. Are they ever changed after being put in a UBQualifier, or is it safe to share them between UBQualifiers?

We definitely re-use some `OffsetEquation` instances, such as the static fields `OffsetEquation.ZERO`, `OffsetEquation.ONE`, etc. I *think* in general it would be safe to share any offset equation - IIRC we designed the APIs to create a new instance any time a modification would occur (e.g. see the documentation for `plusOffset` and `minusOffset`, which is careful to say that the return value is **a copy of** the qualifier being operated on).

That said, I don't know if the implementation actually respects that design - I don't think there's anything right now that would break if it offset equations were modified in some circumstances (as long as it wasn't `ZERO` or its ilk that are actually re-used that were being modified).

@smillst should review this code - she knows it better than I do

smillst:
> A question brought up by these changes is how OffsetEquation objects are treated by the implementation. Are they ever changed after being put in a UBQualifier, or is it safe to share them between UBQualifiers?

OffsetEquations are not mutated once placed in an UBQualifier, so it should be safe to share them.

PR #4409, opened 2021-03-13
Implement RegexUtil.getNonNullGroups() to identify groups that will return @NonNull String on calls to group(int).

UtR491:
The misc tests fail because of some unrelated issue, `checker/src/main/java/org/checkerframework/checker/index/upperbound/OffsetEquation.java:35:5: missing documentation for OffsetEquation`


checker-util/src/test/java/org/checkerframework/checker/regex/util/RegexUtilTest.java
```diff
@@ -23,6 +26,10 @@ public void test_isRegex_and_asRegex() {
         RegexUtil.asRegex(s1, 1);
         Assert.assertFalse(RegexUtil.isRegex(s1, 2));
         Assert.assertThrows(Error.class, () -> RegexUtil.asRegex(s1, 2));
+        Assert.assertTrue(RegexUtil.isRegex(s1, 1, 1));
+        Assert.assertFalse(RegexUtil.isRegex(s1, 1, 1, 2));
```
mernst, line 30: There are no positive tests (that return `true`) of isRegex with more than 3 arguments.  Please add some.  Also add tests where the provided set of non-null regexes is a subset of what it could be, but the result is still true.

checker-util/src/test/java/org/checkerframework/checker/regex/util/RegexUtilTest.java
```diff
@@ -50,5 +63,30 @@ public void test_isRegex_and_asRegex() {
...
+    public void test_getNonNullGroups() {
+        String s1 = "\\(abc\\)?(123)";
+        String s2 = "([(abc)])*(xyz)\\?(abc)";
+        String s3 = "(\\Q()()\\E)";
+        String s4 = "([abc\\Qwww\\E])(abc)?";
+        String s5 = "[(abc]";
+
+        Assert.assertThrows(Error.class, () -> RegexUtil.getNonNullGroups(s1, 2));
+        Assert.assertEquals(
+                new ArrayList<>(Collections.singletonList(1)), RegexUtil.getNonNullGroups(s1, 1));
```
mernst, line 80: There is no need to use `new ArrayList` here and in the later lines of this file.

checker-util/src/main/java/org/checkerframework/checker/regex/util/RegexUtil.java
```diff
@@ -334,4 +427,159 @@ private static String regexErrorMessage(String s, int expectedGroups, int actual
...
+    }
+
+    /**
+     * Returns the index till which the regex can be skipped, when '\' is encountered.
+     *
+     * @param regexp the regular expression to analyse
+     * @param st the index of the '\' which causes the skip
+     * @return the index till which the traversal can be skipped
+     */
+    private static int resumeFromHere(String regexp, int st) {
```
mernst, line 578: This is not a very descriptive name.  (Readers don't know what is being resumed.)  Please rename the method to be more descriptive.

mernst (changes requested):
Here are a few small preliminary comments.

checker-util/src/main/java/org/checkerframework/checker/regex/util/RegexUtil.java
```diff
@@ -334,4 +427,159 @@ private static String regexErrorMessage(String s, int expectedGroups, int actual
...
+        // a capturing group (the top element is true) or a non-capturing group (the top element is
+        // false). One additional check is required. If '(' represented a capturing a group and was
+        // preceded by '|', it is an optional group.
+
+        // If you encounter ')', check the top of the lastUnclosedWasCapturingGroup deque. If it was
+        // false, do nothing otherwise remove the top element from the deque since it is now closed
+        // and check if it is followed by a '?', '*', '|' or '{0'. If it is, then it is an optional
+        // group otherwise not. If it is an optional group, remove it from the nonNullGroups list.
+
+        // If you encounter '[', traverse the regex till you find the closing '[', you may encounter
```
mernst, line 486: "closing ']'" should be "closing '['".  However, I think the code will be easier to read with comments like this inline rather than a huge comment at the beginning, which prevents readers from seeing which parts of the comment correspond to which parts of the code.

checker-util/src/main/java/org/checkerframework/checker/regex/util/RegexUtil.java
```diff
@@ -334,4 +427,159 @@ private static String regexErrorMessage(String s, int expectedGroups, int actual
...
+                boolean closesCapturingGroup = lastUnclosedWasCapturingGroup.pop();
+                if (closesCapturingGroup) {
+                    Integer closedGroupIndex = unclosedCapturingGroups.pop();
+                    if ((i < length - 1 && "?*|".contains(String.valueOf(regexp.charAt(i + 1))))
+                            || (i < length - 2 && regexp.startsWith("{0", i + 1))) {
+                        nonNullGroups.remove(closedGroupIndex);
+                    }
+                }
+            } else if (regexp.charAt(i) == '[') {
+                int balance = 1, j;
```
mernst, line 522: Give each variable its own declaration.  Also name variables descriptively.  "balance" could be "number of open character classes"; even "nesting level" would be more descriptive.

Why is j needed at all?  You could use a while loop that increments the main iteration variable `i`.

checker-util/src/main/java/org/checkerframework/checker/regex/util/RegexUtil.java
```diff
@@ -334,4 +427,159 @@ private static String regexErrorMessage(String s, int expectedGroups, int actual
...
+    }
+
+    /**
+     * Returns the index till which the regex can be skipped, when '\' is encountered.
+     *
+     * @param regexp the regular expression to analyse
+     * @param st the index of the '\' which causes the skip
+     * @return the index till which the traversal can be skipped
+     */
+    private static int resumeTraversalFromHere(String regexp, int st) {
```
mernst, line 554: This method name describes how it is used, but it does not describe what it does, which is more appropriate.
This method, given a backslash construct in a regular expression, returns the end of the backslash construct.

mernst (changes requested):
Thanks for the improvements.

For future reference, when you are done addressing all the comments in a code review, please so state (for example, ask for another review).  I saw you had pushed changes but did not know whether you were done and the code was ready for re-review.

checker-util/src/main/java/org/checkerframework/checker/regex/util/RegexUtil.java
UtR491: I'll inline the big wall of comment. It is taking some time. I'll push my changes by tomorrow.

checker-util/src/main/java/org/checkerframework/checker/regex/util/RegexUtil.java
```diff
@@ -334,4 +427,159 @@ private static String regexErrorMessage(String s, int expectedGroups, int actual
...
+        // need to be closed first.
+
+        // Indices of the groups that are currently not closed. It is a ArrayDeque which is used as
+        // a stack.
+        // An element is pushed to it when we encounter a '(' which opens a capturing group.
+        // An element popped from it when a ')' is encountered which is closing a capturing group.
+        // We know ')' closes a capturing group from our boolean stack.
+        ArrayDeque<Integer> unclosedCapturingGroups = new ArrayDeque<>();
+
+        // Stores whether the last occurrence of '(' in the regex marked a capturing group (true) or
```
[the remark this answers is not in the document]
UtR491: Yes. The "information" that the boolean stack provides is whether the last unclosed '(' was an opening of a capturing group or some other special construct. The Integer stack stores the serial numbers of the unclosed capturing groups. We only remove the top element from the Integer stack when the top of the Boolean stack is true.

checker-util/src/main/java/org/checkerframework/checker/regex/util/RegexUtil.java
UtR491: Is `getLastIndexOfEscapeConstruct` fine?

checker-util/src/main/java/org/checkerframework/checker/regex/util/RegexUtil.java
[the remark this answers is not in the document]
mernst: OK.  Please correct the documentation, which is currently wrong or at least misleading.  (I think I might have already made that fix, though.)

UtR491:
@mernst can you review this please?

checker-util/src/main/java/org/checkerframework/checker/regex/util/RegexUtil.java
```diff
@@ -334,4 +426,147 @@ private static String regexErrorMessage(String s, int expectedGroups, int actual
...
+        // Whenever we encounter any character, we know that it is not escaped. This is because,
+        // when we encountered a '\' which was used to escape something, we traverse till the end
+        // of the escape construct and then only resume normal traversal. This is why we are never
+        // checking if the previous character was a '\'.
+
+        // Optional group here onwards means the ith capturing group, which may not match any part
+        // of a text that matched the regular expression and thus may return null on calls to
+        // matcher.group(i).
+
+        // Special construct means either pure, non-capturing groups that do not capture text and do
```
mernst, line 476: What does "either" refer to?

checker-util/src/main/java/org/checkerframework/checker/regex/util/RegexUtil.java
```diff
@@ -334,4 +426,147 @@ private static String regexErrorMessage(String s, int expectedGroups, int actual
...
+                // If a '\' is encountered, it may escape a single character or may represent a
+                // quote. Traverse till the end of the escape construct.
+                i = resumeTraversalFromHere(regexp, i);
+            }
+        }
+        return nonNullGroups;
+    }
+
+    /**
+     * Returns the index till which the regex can be skipped, when '\' is encountered.
```
mernst, line 560: Update the documentation along with the method name.

checker-util/src/main/java/org/checkerframework/checker/regex/util/RegexUtil.java
```diff
@@ -334,4 +426,147 @@ private static String regexErrorMessage(String s, int expectedGroups, int actual
...
+        // not count towards the group total.
+
+        final int length = regexp.length();
+        for (int i = 0; i < length; i++) {
+            // Marks the opening of either a capturing a capturing group or some special
+            // construct.
+            if (regexp.charAt(i) == '(') {
+                boolean isThisCapturingGroup = false;
+                // If '(' is not followed by a '?' or is followed by '?<', it is a capturing group.
+                if ((i < length - 1)
```
mernst, line 486: I suggest abstracting this into a `startsCapturingGroup` method (even though it is called just once).

checker-util/src/main/java/org/checkerframework/checker/regex/util/RegexUtil.java
```diff
@@ -334,4 +426,147 @@ private static String regexErrorMessage(String s, int expectedGroups, int actual
...
+                    group += 1;
+                    // Push the group index onto a stack.
+                    unclosedCapturingGroups.push(group);
+                    // This opening bracket marked the opening of a capturing group.
+                    isThisCapturingGroup = true;
+                }
+                // Push the boolean on top of a stack. When we encounter a ')', we will use this
+                // stack to check whether it closes a capturing group or a special construct.
+                isCapturingGroup.push(isThisCapturingGroup);
+                if (isThisCapturingGroup) {
```
mernst, line 497: This is the same condition as the if statement above.  You can merge their bodies.

checker-util/src/main/java/org/checkerframework/checker/regex/util/RegexUtil.java
```diff
@@ -334,4 +426,147 @@ private static String regexErrorMessage(String s, int expectedGroups, int actual
...
+                    // If this opening was preceded by a '|', this group is optional.
+                    if (i > 0 && regexp.charAt(i - 1) == '|') {
+                        nonNullGroups.remove(Integer.valueOf(group));
+                    }
+                }
+            } else if (regexp.charAt(i) == ')') {
+                // This closes the last opened "thing" (either a capturing group or a special
```
mernst, line 504: You don't need a comment for every line of code.  Doing so makes the code much more verbose and harder to read.  The key to clarity is naming and documenting the variables well, once.  This entire comment is redundant with the documentation of `isCapturingGroup`.  However, the terminology is different.  The documentation of `isCapturingGroup` speaks of capturing groups and non-capturing groups, but this speaks of capturing groups and special constructs.  This terminology is better and should be used throughout instead of "non-capturing group".  But the most important thing is to use consistent terminology throughout.

Many of the other comments are also not needed, such as "Push the group index onto a stack" which is entirely redundant with the call to `push` on the following line.

checker-util/src/main/java/org/checkerframework/checker/regex/util/RegexUtil.java
```diff
@@ -334,4 +426,147 @@ private static String regexErrorMessage(String s, int expectedGroups, int actual
...
+                    Integer closedGroupIndex = unclosedCapturingGroups.pop();
+                    // If the ')' is followed by a '?', '*', '|' or '{0', then this capturing group
+                    // is optional and can be removed from the list of nonNullGroups.
+                    if ((i < length - 1 && "?*|".contains(String.valueOf(regexp.charAt(i + 1))))
+                            || (i < length - 2 && regexp.startsWith("{0", i + 1))) {
+                        nonNullGroups.remove(closedGroupIndex);
+                    }
+                }
+            } else if (regexp.charAt(i) == '[') {
+                // A character class has been opened. We will traverse till we close this class.
```
mernst, line 523: I suggest abstracting out this body into a separate method, much like `getLastIndexOfEscapeConstruct`.

checker-util/src/main/java/org/checkerframework/checker/regex/util/RegexUtil.java
```diff
@@ -334,4 +426,147 @@ private static String regexErrorMessage(String s, int expectedGroups, int actual
...
+                // stack.
+                // If the top is false, then it was a special construct.
+                // If the top is true, then it was a capturing group and we need to analyse it
+                // further.
+                boolean closesCapturingGroup = isCapturingGroup.pop();
+                if (closesCapturingGroup) {
+                    // The capturing group being closed is on top of the unclosedCapturingGroups
+                    // stack.
+                    Integer closedGroupIndex = unclosedCapturingGroups.pop();
+                    // If the ')' is followed by a '?', '*', '|' or '{0', then this capturing group
```
mernst, line 515: This is not a correct test.  A simple syntactic test of the nearby characters will not work, because the relevant `'|'` may be arbitrarily far after the `')'`, or may appear before it.  A single `'|'` character may also make multiple groups optional.  I added a couple test cases, but many more devious ones are possible.

mernst (changes requested):

checker-util/src/main/java/org/checkerframework/checker/regex/util/RegexUtil.java
mernst: You might have figured this out already, but one implementation approach would be to add a stack that keeps track of all opened parenthesis-constructs, recording whether that parenthesis-construct contains an alternation `|`.  When a parenthesis-construct is closed, mark any contained capturing groups as optional.  Likewise, when any capturing group is marked as optional because it is followed by `?` etc., mark any contained capturing group as optional.

UtR491: Yes, I completely missed this case. I am really sorry.
I don't see how keeping all opened parenthesis-constructs will help, I think we can do it with the information we already have. The main observation is that when a capturing group is enclosed within another, its group number will be greater than the group number of the enclosing group. So when we close a group with group number `i`, and we find it is immediately followed by one of `? * {0'`, we can mark all the groups with group numbers from `i+1` to `groupCount`. Handling `'|'` will be a little bit different, when we find a group containing `'|'`, everything inside the group will be optional.
Can you please find some bug in this approach and elaborate upon yours and why it might be a better one?
I am sorry for my late follow-ups, my class schedule has been really hectic for the past two weeks.

mernst:
I haven't heard from you recently.  Are you still interested in this pull request?

mernst:
@UtR491 Ping about whether you wish to continue working on this or close the pull request.

UtR491:
Hi @mernst. Sorry I will not be able to contribute till September. I am closing this for now. Thanks and sorry for the inconvenience.

mernst:
OK.  Thanks for the progress so far.  If you get back to this, we will be glad to have it.

Issue #4412, opened 2021-03-14
Algebraic Data Type Performance Problems


hborders:
Thanks so much for this project. It's the best way I've found to analyze nullability of Java!

When using an algebraic data type pattern, checker performance degrades as the number of cases within the algebraic data type increases. I haven't let checker run long enough to complete the `either5` or higher or `emptyeither6` or higher cases.

An example project is available at https://github.com/hborders/checker-perf

```
$ ./gradlew :emptyeither2:assemble -DrocksSkipCheckerFramework=false
Configuration on demand is an incubating feature.

BUILD SUCCESSFUL in 4s

$ ./gradlew :either2:assemble -DrocksSkipCheckerFramework=false
Configuration on demand is an incubating feature.

BUILD SUCCESSFUL in 5s

$ ./gradlew :emptyeither3:assemble -DrocksSkipCheckerFramework=false
Configuration on demand is an incubating feature.

BUILD SUCCESSFUL in 8s

$ ./gradlew :either3:assemble -DrocksSkipCheckerFramework=false
Configuration on demand is an incubating feature.

BUILD SUCCESSFUL in 16s

$ ./gradlew :emptyeither4:assemble -DrocksSkipCheckerFramework=false
Configuration on demand is an incubating feature.

BUILD SUCCESSFUL in 47s

$ ./gradlew :either4:assemble -DrocksSkipCheckerFramework=false
Configuration on demand is an incubating feature.

BUILD SUCCESSFUL in 2m 20s$ ./gradlew :emptyeither5:assemble -DrocksSkipCheckerFramework=false
Configuration on demand is an incubating feature.

BUILD SUCCESSFUL in 8m 17s
```

smillst:
Thanks for reporting this and the test case.  I've made some progress on speeding up this test case, #4416, but it's still too slow.

hborders:
Thanks. I'll test it on the next release.

PR #4416, opened 2021-03-16
Cache the hashcode of AnnotatedTypeMirror#underlyingType.

smillst:
This speeds up the test cases in #4412, but doesn't completely fix the performance problems.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeMirror.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeMirror.java
@@ -131,6 +131,9 @@ public abstract class AnnotatedTypeMirror {
     /** Actual type wrapped with this AnnotatedTypeMirror. */
     protected final TypeMirror underlyingType;
 
+    /** The hash code for {@code underlyingType.toString()}. Lazily initialized. */
+    protected int underlyingTypeHashcode;
+
     /** The annotations on this type. */
     // AnnotationMirror doesn't override Object.hashCode, .equals, so we use
     // the class name of Annotation instead.
@@ -154,6 +157,8 @@ public abstract class AnnotatedTypeMirror {
         this.underlyingType = underlyingType;
         assert atypeFactory != null;
         this.atypeFactory = atypeFactory;
+        this.underlyingTypeHashcode = -1;
+        ;
     }
 
     @Override
--- a/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeMirror.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeMirror.java
@@ -131,8 +131,11 @@ public abstract class AnnotatedTypeMirror {
     /** Actual type wrapped with this AnnotatedTypeMirror. */
     protected final TypeMirror underlyingType;
 
-    /** Used for generating the hash code for this. Lazily initialized. */
-    protected String underlyingTypeString = null;
+    /**
+     * Used for generating the hash code for this. Call {@link #getUnderlyingTypeString()} rather
+     * than using the field directly.
+     */
+    private String underlyingTypeString = null;
 
     /** The annotations on this type. */
     // AnnotationMirror doesn't override Object.hashCode, .equals, so we use
--- a/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeMirror.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeMirror.java
@@ -1304,12 +1304,14 @@ public abstract class AnnotatedTypeMirror {
             return type;
         }
 
+        /**
+         * Returns the erased types corresponding to the given types.
+         *
+         * @param lst annotated type mirrors
+         * @return erased annotated type mirrors
+         */
         private List<AnnotatedTypeMirror> erasureList(Iterable<? extends AnnotatedTypeMirror> lst) {
-            List<AnnotatedTypeMirror> erased = new ArrayList<>();
-            for (AnnotatedTypeMirror t : lst) {
-                erased.add(t.getErased());
-            }
-            return erased;
+            return SystemUtil.mapList(AnnotatedTypeMirror::getErased, lst);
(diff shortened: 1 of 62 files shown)


framework/src/main/java/org/checkerframework/framework/type/HashcodeAtmVisitor.java
```diff
@@ -31,8 +31,10 @@ protected Integer defaultAction(AnnotatedTypeMirror type, Void v) {
         if (type == null) {
             return 0;
         }
+        if (type.underlyingTypeHashcode == -1) {
+            type.underlyingTypeHashcode = type.underlyingType.toString().hashCode() * 13;
```
mernst, line 35: What is the point of this multiplication by 13?

framework/src/main/java/org/checkerframework/framework/type/HashcodeAtmVisitor.java
```diff
@@ -31,8 +31,10 @@ protected Integer defaultAction(AnnotatedTypeMirror type, Void v) {
...
+        }
 
-        return type.getAnnotations().toString().hashCode() * 17
-                + type.getUnderlyingType().toString().hashCode() * 13;
+        return type.getAnnotations().toString().hashCode() * 17 + type.underlyingTypeHashcode;
```
mernst, line 38: The previous code was in bad style because it attempted to reimplement hashing.  Please use `Objects.hash` instead.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeMirror.java
```diff
@@ -131,6 +131,9 @@ public static AnnotatedTypeMirror createType(
     /** Actual type wrapped with this AnnotatedTypeMirror. */
     protected final TypeMirror underlyingType;
 
+    /** The hash code for {@code underlyingType.toString()}. Lazily initialized. */
```
mernst, line 134: I would document that the hash code for this is not cached because ATM is mutable, but the underlying type never changes and is immutable, so its hash code does not change.

framework/src/main/java/org/checkerframework/framework/type/HashcodeAtmVisitor.java
mernst: I think the caching code would be cleaner and easier to understand if it was performed in the class that has the cache.  That is, create a method on ATM like "getUnderlyingTypeHashCode".

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeMirror.java
mernst: Nit: Java spells the method `hashCode` so I would capitalize the C here as well.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeMirror.java
```diff
@@ -154,6 +157,8 @@ private AnnotatedTypeMirror(TypeMirror underlyingType, AnnotatedTypeFactory atyp
         this.underlyingType = underlyingType;
         assert atypeFactory != null;
         this.atypeFactory = atypeFactory;
+        this.underlyingTypeHashcode = -1;
+        ;
```
mernst, line 161: Stray semicolon.

framework/src/main/java/org/checkerframework/framework/type/HashcodeAtmVisitor.java
```diff
@@ -31,8 +31,10 @@ protected Integer defaultAction(AnnotatedTypeMirror type, Void v) {
         if (type == null) {
```
mernst, line 31: Two ATMs have the same hash code if they have the same top-level annotations and the same underlying type.  This means that `Map<String, Object>` and `Map<@Nullable String, @Nullable Object>` have the same hash code.

The above fact is implied by the class documentation, but I suggest stating it explicitly there.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeMirror.java
```diff
@@ -154,6 +157,8 @@ private AnnotatedTypeMirror(TypeMirror underlyingType, AnnotatedTypeFactory atyp
         this.underlyingType = underlyingType;
         assert atypeFactory != null;
         this.atypeFactory = atypeFactory;
+        this.underlyingTypeHashcode = -1;
```
mernst, line 160: I would put this at the field declaration, making it explicit what the "uninit" value is.  (Currently that is buried in the constructor.)

mernst (changes requested):
Thanks for figuring out this problem!
I have a few suggestions to make the code cleaner.

framework/src/main/java/org/checkerframework/framework/type/HashcodeAtmVisitor.java
smillst: This isn't true.  `HashcodeAtmVisitor` scans the whole type and sums the hash code for each AnnotatedTypeMirror in the AnnotatedTypeMirror.  I updated the Class Javadoc to make this clearer.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeMirror.java
```diff
@@ -131,6 +131,12 @@ public static AnnotatedTypeMirror createType(
     /** Actual type wrapped with this AnnotatedTypeMirror. */
     protected final TypeMirror underlyingType;
 
+    /**
+     * Used for generating the hash code for this. Call {@link #getUnderlyingTypeString()} rather
+     * than using the field directly.
+     */
+    private String underlyingTypeString = null;
```
mernst, line 138: It would be more efficient to compute the hash code once and store it.  Storing the string, which forces the client to always recompute the string's hash code.  (You can pass an int to `Objects.hashCode`)
But, if you want it this way anyway, that's OK.

mernst (approved):

PR #4420, opened 2021-03-17
Make `refineGT` more efficient

mernst:
I'm not 100% positive this is correct.  If it is not, then the test suite is missing a test. :-)


kelloggm (approved):
The only way that this could lead to behavior changes is that the call to `store.insertValue` now doesn't occur on two paths:
* when `leftJe` is a float or double literal, and
* when `leftJe.toString()` is already a member of `lessThanExpressions`.

The latter case is uninteresting, since `lessThanExpressions` was derived from the type of `rightJe` in the store, so there's nothing to change.

The former case is also safe: no changes were made to `lessThanExpressions` based on float or double literals already. To be sure, I wrote the following test case, which already issues an error on `master`:
```

class LessThanFloatLiteral {
    void test(int x) {
        if (1.0 > x) {
            @LessThan("1") int y = x;
        }
    }
}
```

We could add this test case to the test suite to be sure that nothing continues to be inferred for float/double literals, but I don't really think it's necessary.

PR #4434, opened 2021-03-19
Add more efficient overload for `getElementValueEnum()`

mernst:


javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
```diff
@@ -1020,6 +1021,51 @@ public static long getElementValueLong(
...
+     * Get the element with the name {@code name} of the annotation {@code anno}. The result is an
+     * enum of type {@code T}.
+     *
+     * @param anno the annotation to disassemble
+     * @param element the element to access; it must be present in the annotation
+     * @param expectedType the type of the element and the return value, an enum
+     * @param <T> the class of the type
+     * @return the value of the element with the given name
+     */
+    public static <T extends Enum<T>> T getElementValueEnum(
```
smillst, line 1034: #4433 add TODOs about adding these methods.   Can the TODOs be removed now?

https://github.com/typetools/checker-framework/pull/4433/files#diff-37a231ba0f2d785f4475210f48906e3865d5c4204f93ed35855ed7f872d0a98dR737-R742

smillst (approved):

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationUtils.java
mernst: I thought I had done that, but I see I didn't.  Good catch!

mernst: I had done it in a different branch...

PR #4445, opened 2021-03-22
Improve documentation clarity

ChristosToxiadis:
In a SideEfectFree method If a parameter of type int or String change value there is no problem. But if a parameter int[] change value in some element then the code won't pass the checks. For example 

public void test1(int x, String y){
         x = 34;
         y = "change";
}
It is ok.

While 

public void test2(int[] x){
         x[0] = 34;
}
It is wrong.

So i think it is better to clarify that matter.

### DIFF
--- a/checker-qual/src/main/java/org/checkerframework/dataflow/qual/SideEffectFree.java
+++ b/checker-qual/src/main/java/org/checkerframework/dataflow/qual/SideEffectFree.java
@@ -29,7 +29,7 @@ import java.lang.annotation.Target;
  * of the following Java constructs:
  *
  * <ol>
- *   <li>Assignment to any expression, except for local variables and method parameters.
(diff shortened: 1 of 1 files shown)


mernst:
The expression `x[0] = 34` is *not* an assignment to a formal parameter.  It is an assignment to an element of `x`. That is a side effect to the value of `x`, but `x` remains assigned to exactly the same array object as before the statement.
So, I think the documentation as stated is correct, and the change is confusing.

You have pointed out a point of confusion, however:  a reader might not understand the concept of assignment in Java, and that `x = someArray` is an assignment to `x` whereas `x[0] = 34` is not.  I opened https://github.com/typetools/checker-framework/pull/4446 to explain that.

ChristosToxiadis:
I think that it is more clear now. I can't understand though, why my pull request didn't pass the test, since the only change i made was to add a word in javadoc.

PR #4441, opened 2021-03-20
Eliminate method `getElementValueArrayOrSingleton()`

mernst:
~~Merge after https://github.com/typetools/checker-framework/pull/4439 and https://github.com/typetools/checker-framework/pull/4440.~~

PR #4448, opened 2021-03-22
Avoid calling inefficient version of `getElementValueArray`

mernst:
Merge after #4441

I see some refactorings that I could split into separate pull requests if you prefer.


javacutil/build.gradle
```diff
@@ -12,6 +12,17 @@ dependencies {
     // External dependencies:
     // If you add an external dependency, you must shadow its packages.
     // See the comment in ../build.gradle in the shadowJar block.
+
+    /// TODO: The below is temporary until the `removeDuplicates` routine is moved into plume-lib.
```
smillst, line 16: I see that you also added `removeDuplicates` to SystemUtil.  Is the dependency on Guava still required?  Although, Guava is shadowed in `checker.jar`, it's not shadowed in the dataflow and javacutil artifacts we publish.  If we add a dependency to Guava here, it could cause a classpath conflict for users who uses those artifacts.

smillst (approved):
> I see some refactorings that I could split into separate pull requests if you prefer.

I skimmed all of the code and didn't see any problems.  It seems like the conversion is fairly straight forward.  Is it worth it for me to take a closer look?  Are there any changes in particular that I should review?  If you want me to review it closely, then it would be helpful if this pull request was split up.  Thanks!

javacutil/build.gradle
mernst: `removeDuplicates` depends on Guava.  I'll move the method into plume-util later this month, and then this dependency won't be needed any more.

smillst: So at that point javacutil will have a transitive dependency on Guava.  Are you going to shadow Guava in plume-util?

mernst: OK, I have removed the dependency on Guava.

mernst:
I removed the unrelated changes.  One concrete question is whether there is a better way in IndexChecker than passing around two type factories.

smillst:
>  One concrete question is whether there is a better way in IndexChecker than passing around two type factories.

Here's another way: https://github.com/mernst/checker-framework/pull/14

smillst (approved):

Issue #4452, opened 2021-03-23
BugInCF: Unrecognized class: class java.lang.OutOfMemoryError


cushon:
We recently encountered the following crash in some code that was using dataflow as a library:

```
  Caused by: org.checkerframework.shaded.javacutil.BugInCF: Unrecognized class: class java.lang.OutOfMemoryError
  	at org.checkerframework.shaded.javacutil.TypesUtils.typeFromClass(TypesUtils.java:74)
  	at org.checkerframework.shaded.dataflow.cfg.builder.CFGTranslationPhaseOne.getTypeMirror(CFGTranslationPhaseOne.java:3744)
  	at org.checkerframework.shaded.dataflow.cfg.builder.CFGTranslationPhaseOne.<init>(CFGTranslationPhaseOne.java:395)
  	at org.checkerframework.shaded.dataflow.cfg.builder.CFGBuilder.build(CFGBuilder.java:67)
```

The root cause was that `CFGTranslationPhaseOne` tries to load a `TypeMirror` for `java.lang.OutOfMemoryError` here:

https://github.com/typetools/checker-framework/blob/562105de109198ca6206b1aeb9a46fe9e9420bb0/dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java#L395

and the affected compilation was invoking javac using a custom `-bootclasspath` that did not contain `java.lang.OutOfMemoryError`. (The use-case is Java code that is translated to JavaScript using [j2cl](https://github.com/google/j2cl); the custom bootclasspath contains APIs that are emulated in JavaScript, so `OutOfMemoryError` is deliberately not present.)

I realize this is a kind of unusual use-case. Would you be willing to consider accepting a patch to make `CFGTranslationPhaseOne` handle the case where `OutOfMemoryError` isn't available?

mernst:
Liam, we would accept such a patch.  I agree it's weird, but I don't think it will make the code too much worse and we wish to help you with your use-case.

cushon:
Thanks! I mailed a proposal.

PR #4453, opened 2021-03-23
Make resolving some exception types optional in `CFGTranslationPhaseOne`

cushon:
Fixes https://github.com/typetools/checker-framework/issues/4452

### DIFF
--- a/dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java
@@ -320,16 +320,16 @@ public class CFGTranslationPhaseOne extends TreePathScanner<Node, Void> {
     final TypeMirror nullPointerExceptionType;
 
     /** The OutOfMemoryError type. */
-    final TypeMirror outOfMemoryErrorType;
+    final @Nullable TypeMirror outOfMemoryErrorType;
 
     /** The ClassCircularityError type. */
-    final TypeMirror classCircularityErrorType;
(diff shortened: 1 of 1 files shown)


mernst (changes requested):
Could you please write some documentation (probably on `maybeGetTypeMirror` to explain the reason for the method and the reason that the method might return null?  Otherwise someone might be confused by the code, and might simplify the code by removing the method.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java
```diff
@@ -3754,4 +3764,20 @@ public Node visitOther(Tree tree, Void p) {
...
+
+    /**
+     * Returns the TypeMirror for the given class, or {@code null}.
+     *
+     * @param clazz a class
+     * @return the TypeMirror for the class
+     */
+    private @Nullable TypeMirror maybeGetTypeMirror(Class<?> clazz) {
+        String name = clazz.getCanonicalName();
+        assert name != null : "@AssumeAssertion(nullness): assumption";
```
mernst, line 3776: The method only works for classes with a canonical name.  That is worth mentioning in its Javadoc, and then here "assumption" can be more specific.

cushon: Done

cushon (commented):
> Could you please write some documentation (probably on `maybeGetTypeMirror` to explain the reason for the method and the reason that the method might return null?

Done

mernst (approved):

mernst:
Thanks!

PR #4459, opened 2021-03-24
Handle unions via their tree, not their type; fixes #4381

### LINKED ISSUE #4381: regression with UnknownInitialization
This diagnostic is reported in 3.11.0 but not 3.10.0, is this a bug? Since `m` is a local variable I don't think there's anything here that couldn't be initialized, and when I was trying to minimize it, it seems like an interaction with the `try`/`catch`.

```java
import java.io.IOException;
import java.security.GeneralSecurityException;

abstract class T {

  public void t() {
    int m = 0;
    try {
      f();
    } catch (IllegalArgumentException | IOException e) {
    } catch (GeneralSecurityException e) {
      g(m);
    }
  }

  abstract void g(int x);

  abstract void f() throws IllegalArgumentException, IOException, GeneralSecurityException;
}
```

```
$ ./checker-framework-3.11.0/checker/bin/javac -processor Nullness T.java
...
T.java:12: error: [argument.type.incompatible] incompatible argument for parameter x of g.
      g(m);
        ^
  found   : @UnknownInitialization @NonNull int
  required: @Initialized @NonNull int
1 error
```

mernst:


smillst:
I think the problem is actually using` instanceof DeclaredType`, see my alternative fix https://github.com/typetools/checker-framework/pull/4463.

smillst:
We merged  #4463 instead.

PR #4464, opened 2021-03-24
Create annotation insertion script for ajava files

jwaataja:
Adds a program `InsertAjavaAnnotations.java` that inserts annotations from ajava files to java files, similarly to the `insert-annotations-to-source` script. This allows adding annotations inferred from whole program inference to the original program they were inferred from.


framework/src/main/java/org/checkerframework/framework/ajava/TypeAnnotationMover.java
```diff
@@ -0,0 +1,225 @@
...
+                            Collections.singletonList(annoTypeElt)));
+        }
+
+        return annoTypeElt;
+    }
+
+    /**
+     * Returns if {@code annotation} could be declaration annotation for {@code declarationType}.
+     * This would be the case if the annotation isn't recognized at all, or if it was recognized and
+     * has {@code declarationType} as one of its targets.
```
mernst, line 164: If an annotation has no `@Target` meta-annotation, then it is applicable to declaration locations.  Please handle this case too.

framework/src/main/java/org/checkerframework/framework/ajava/TypeAnnotationMover.java
```diff
@@ -0,0 +1,226 @@
...
+     * Returns whether the annotation represented by {@code annotationDeclaration} might be a
+     * declaration annotation for {@code declarationType}. This holds if {@code declarationType} is
+     * a target of the annotation, or if {@code ElementType.TYPE_USE} is not a target of the
+     * annotation.
+     *
+     * @param annotationDeclaration declaration for an annotation
+     * @param declarationType the declaration type to check if the annotation might be a declaration
+     *     annotation for
+     * @return true if {@code annotationDeclaration} contains {@code declarationType} as a target or
+     *     doesn't contain {@code ElementType.TYPE_USE} as a target
```
mernst, line 192: I don't see why the second clause (after "or") is necessary.  Maybe the code should issue an error if the targets include neither `declarationType` nor `TYPE_USE`, though, as a sanity check.

framework/src/main/java/org/checkerframework/framework/ajava/TypeAnnotationMover.java
```diff
@@ -0,0 +1,226 @@
...
+     */
+    private boolean isPossiblyDeclarationAnnotation(
```
mernst, line 194: This method doesn't need "possibly" in its name, I think.  It can definitively check.

framework/src/main/java/org/checkerframework/framework/ajava/TypeAnnotationMover.java
```diff
@@ -0,0 +1,226 @@
...
+                return true;
+            }
+        }
+
+        return !hasTypeUse;
+    }
+
+    /**
+     * Returns whether {@code type} has a name containing multiple parts separated by dots, e.g.
+     * "java.lang.String".
```
mernst, line 217: I would also mention "Outer.Inner", because readers might not think of that case.  Does the code handle it correctly?

framework/src/main/java/org/checkerframework/framework/ajava/InsertAjavaAnnotations.java
```diff
@@ -0,0 +1,580 @@
...
+    /**
+     * Gets an instance of {@code Elements} from the current Java compiler.
+     *
+     * @return an instance of {@code Elements}
+     */
+    private static Elements createElements() {
+        JavaCompiler compiler = javax.tools.ToolProvider.getSystemJavaCompiler();
+        if (compiler == null) {
+            System.err.println("Could not get compiler instance");
+            System.exit(0);
```
mernst, line 66: If there is a problem, the exit code should be non-zero.

framework/src/main/java/org/checkerframework/framework/ajava/InsertAjavaAnnotations.java
```diff
@@ -0,0 +1,580 @@
...
+
+        @Override
+        public String toString() {
+            return "Insertion [contents=" + contents + ", position=" + position + "]";
+        }
+    }
+
+    /**
+     * Given two JavaParser ASTs representing the same Java file but with differing annotations,
+     * stores a list of {@link Insertion}s for all annotations in the first AST into the second AST.
```
mernst, line 140: It would be helpful to say that after a client calls some method (which one?), the client should read the `insertions` field.  Is that what this does?  Or does this just transfer annotations from one AST into the second, in which case the `insertions` field, and even the mention of `Insertion`, is an implementation detail that does not belong in the class Javadoc.

framework/src/main/java/org/checkerframework/framework/ajava/InsertAjavaAnnotations.java
```diff
@@ -0,0 +1,580 @@
...
+         * representation of the AST to insertion annotations to. When visiting a node pair, the
+         * second node must always be from an AST generated from this String.
+         *
+         * @param destFileContents the String the second vistide AST was parsed from
+         */
+        public BuildInsertionsVisitor(String destFileContents) {
+            allAnnotations = null;
+            insertions = new ArrayList<>();
+            printer = new PrettyPrinter();
+            String[] lines = destFileContents.split(System.lineSeparator());
```
mernst, line 177: This looks suspicious to me.  I occasionally process "foreign" files -- created under Windows or on a Mac -- and it would be best if the program handle those.  If you want to defer that to a later pull request, at least note the limitation.

framework/src/main/java/org/checkerframework/framework/ajava/InsertAjavaAnnotations.java
```diff
@@ -0,0 +1,580 @@
...
+                        node2ArrayTypes.b.get(i - 1).getTokenRange().get().toRange().get().end;
+                addAnnotations(position, node1ArrayTypes.b.get(i).getAnnotations(), 1, true);
+            }
+
+            // Visit the component type.
+            node1ArrayTypes.a.accept(this, node2ArrayTypes.a);
+        }
+
+        @Override
+        public void visit(final CompilationUnit node1, final Node other) {
```
mernst, line 240: Throughout this file, the variables `node1` and `node2` make the code unnecessarily hard to read.  Please change them to something like `dest` and `source`, which better indicates what they are and how they are used.

mernst, line 240: Minor point:  I find `final` to be clutter on these formal parameters.  It is not present on other formal parameters in the file, nor on local variables in this method.

framework/src/main/java/org/checkerframework/framework/ajava/InsertAjavaAnnotations.java
```diff
@@ -0,0 +1,580 @@
...
+            boolean ownLine = true;
+            for (int i = 0; i < insertionColumn; i++) {
+                if (line.charAt(i) != ' ' && line.charAt(i) != '\t') {
+                    ownLine = false;
+                    break;
+                }
+            }
+
+            if (ownLine) {
+                StringBuilder insertionContent = new StringBuilder();
```
mernst, line 318: Use StringJoiner instead of StringBuilder, and the body of the `for` loop can be 1 line.

framework/src/main/java/org/checkerframework/framework/ajava/InsertAjavaAnnotations.java
```diff
@@ -0,0 +1,580 @@
...
+                    // "String"), but @Pure should always appear first.
+                    if (cmp == 0 && (insertion1.ownLine != insertion2.ownLine)) {
+                        if (insertion1.ownLine) {
+                            cmp = -1;
+                        } else {
+                            cmp = 1;
+                        }
+                    }
+
+                    return -cmp;
```
mernst, line 475: I suggest abstracting out this function so that you can give it a name and simplify this code.

framework/src/main/java/org/checkerframework/framework/ajava/InsertAjavaAnnotations.java
```diff
@@ -0,0 +1,580 @@
...
+     * @param args command line arguments, the first element should be a path to ajava files and the
+     *     second should be the directory containing Java files to insert into
+     */
+    public static void main(String[] args) {
+        if (args.length != 2) {
+            System.out.println(
+                    "Usage: java InsertAjavaAnnotations <ajava-directory> <java-file-directory");
+            System.exit(1);
+        }
+
```
mernst, line 531: I suggest assigning `args[0]` and `args[1]` into local variables here; their names will be clearer in the following code than the array accesses are.

framework/src/main/java/org/checkerframework/framework/ajava/InsertAjavaAnnotations.java
```diff
@@ -0,0 +1,580 @@
...
+                        Set<String> annotationFilesForRoot = new LinkedHashSet<>();
+                        for (TypeDeclaration<?> type : root.getTypes()) {
+                            String name = type.getNameAsString();
+                            if (root.getPackageDeclaration().isPresent()) {
+                                name =
+                                        root.getPackageDeclaration().get().getNameAsString()
+                                                + "."
+                                                + name;
+                            }
```
mernst, line 559: I suggest abstracting this logic into a method (especially since I think it occurs at least twice).
It could go in file `JavaParserUtils.java`.

framework/src/main/java/org/checkerframework/framework/ajava/AnnotationEqualityVisitor.java
```diff
@@ -0,0 +1,85 @@
...
+     * AST where annotations differed, or null otherwise.
+     *
+     * @return the node from the second AST with differing annotations or null
+     */
+    public @Nullable NodeWithAnnotations<?> getMismatchedNode2() {
+        return mismatchedNode2;
+    }
+
+    @Override
+    public void defaultAction(Node node1, Node node2) {
```
mernst, line 61: Does it make sense to declare the method as
```java
<T extends Node> defaultAction(T node1, T node2) { ... }
```
to emphasize that the two arguments have the same type?

framework/src/main/java/org/checkerframework/framework/ajava/AnnotationEqualityVisitor.java
```diff
@@ -0,0 +1,85 @@
...
+
+        Node node1Copy = node1.clone();
+        Node node2Copy = node2.clone();
+        for (Comment comment : node1Copy.getAllContainedComments()) {
+            comment.remove();
+        }
+
+        for (Comment comment : node2Copy.getAllContainedComments()) {
+            comment.remove();
+        }
```
mernst, line 75: Why is this code necessary?  Please write a comment.

framework/src/main/java/org/checkerframework/framework/ajava/AnnotationEqualityVisitor.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @Override
+    public void defaultAction(Node node1, Node node2) {
+        if (!(node1 instanceof NodeWithAnnotations<?>)
+                || !(node2 instanceof NodeWithAnnotations<?>)) {
+            return;
+        }
+
+        Node node1Copy = node1.clone();
```
mernst, line 67: The type of `node1Copy` and `node2Copy` should be `NodeWithAnnotations`, not just `Node`.

framework/src/main/java/org/checkerframework/framework/ajava/DoubleJavaParserVisitor.java
```diff
@@ -0,0 +1,866 @@
...
+
+/**
+ * A visitor that visits two JavaParser ASTs simultaneously that almost match. When calling a visit
+ * method, the secondary argument must be another JavaParser {@code Node} that represents an AST
+ * identical to the first argument except for the following exceptions:
+ *
+ * <ol>
+ *   <li>The annotations may differ
+ *   <li>They may differ in whether a method has an explicit receiver
+ * </ol>
```
mernst, line 110: What about import statements? implicit bounds being made explicit? casts?
Maybe cross-reference the manual section about ajava files that explains what differences are possible?

mernst (changes requested):
This looks good overall.  Thank you!
Here are some suggestions -- mostly small ones -- to improve it somewhat.

framework/src/main/java/org/checkerframework/framework/ajava/AnnotationEqualityVisitor.java
jwaataja: The `NodeWithAnnotations` interface doesn't have the method `getAllContainedComments`, so we have to store these as `Node`s.

framework/src/main/java/org/checkerframework/framework/ajava/InsertAjavaAnnotations.java
```diff
@@ -0,0 +1,583 @@
...
+
+        String whitespaceCopy = line.substring(0, position.column - 1);
+        int filePosition = getFilePosition(position);
+        insertions.add(
+            new Insertion(
+                filePosition,
+                insertionContent.toString() + System.lineSeparator() + whitespaceCopy,
+                true));
+      } else {
+        addAnnotations(position, annotations, 0, false);
```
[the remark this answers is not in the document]
jwaataja: I think the way it is now makes the most sense since we usually don't want to add extra space before an annotation. For example, in a method declaration `void myMethod(@Tainted String s1, @Untainted String s2)`, each annotation has a space before it if and only if there was a space before the original type it was attached to, that is, there is no extra space before the annotations. Are you saying we should treat the annotation differently since it was requested to have its own line?

framework/src/main/java/org/checkerframework/framework/ajava/InsertAjavaAnnotations.java
```diff
@@ -0,0 +1,589 @@
...
+    }
+  }
+
+  /**
+   * Inserts annotations from ajava files into Java files in place.
+   *
+   * <p>The first argument is an ajava file or a directory containing ajava files.
+   *
+   * <p>The second argument is a Java file or a directory containing Java files to insert
+   * annotations into. The files must use the same line separator as the host system.
```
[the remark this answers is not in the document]
jwaataja: Previously, it was important for correctness because how the lines were broken up affected whether we decided to put an annotation on its own line. But it's moot now since I switched it to infer the line separator, as you suggested.

framework/src/main/java/org/checkerframework/framework/ajava/InsertAjavaAnnotations.java
```diff
@@ -0,0 +1,581 @@
...
+    /**
+     * Stores the offsets of the lines in the string representation of the second AST. At index i,
+     * stores the number of characters from the start of the file to the beginning of the ith line.
+     */
+    private List<Integer> cumulativeLineSizes;
+
+    /**
+     * Constructs a {@code BuildInsertionsVisitor} where {@code destFileContents} is the String
+     * representation of the AST to insertion annotations to. When visiting a node pair, the second
+     * node must always be from an AST generated from this String.
```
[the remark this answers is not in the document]
jwaataja: The original wording was my intention. The idea is that the user has a file, generates an AST from it, and passes the file contents to the visitor before using it on the AST. Is there a wording that would make that more clear?

mernst (approved):

mernst:
Thanks for this insertion program.
Please merge when you are satisfied, or let me know if you don't have permission to do so.

framework/src/main/java/org/checkerframework/framework/ajava/InsertAjavaAnnotations.java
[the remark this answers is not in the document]
jwaataja: The `addAnnotations` method on this line already adds a space after the insertion content. The whitespace that precedes an inserted annotation should already be present before its insertion. Setting the last parameter to `true`, I think, is only necessary in very niche edge cases in array types.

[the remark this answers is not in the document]
mernst: OK.  Thanks for the explanation!  And thanks again for this program.

PR #4465, opened 2021-03-24
Add an example of why to compare using TypeKind.

smillst:


docs/manual/creating-a-checker.tex
```diff
@@ -2269,13 +2269,19 @@
 help you to create your checker.
 
 \emph{Type-checking}:
-Do not use \code{instanceof} to determine the class of a \<TypeMirror> or \<Element>,
+Do not use \code{instanceof} to determine the kind of a \<TypeMirror> or \<Element>,
```
mernst, line 2272: `instanceof` tests classes, not kinds, so I think this change is not desirable.

mernst (approved):
Thanks for this example.  I find the design atrocious, but we have to live with it.

docs/manual/creating-a-checker.tex
smillst: I think using `class` is (also) confusing because `instanceof` does correctly determine a class or interface.  I rewrote this sentence to be clearer.

smillst:
@mernst I rewrote the first sentence.  Would you please review again?

mernst (approved):

PR #4470, opened 2021-03-26
Use ExecutableElements in DenpendentTypesHelper#annoToElements.

smillst:


framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -103,9 +104,8 @@
     /** AnnotatedTypeFactory */
     protected final AnnotatedTypeFactory factory;
 
-    // TODO: Using strings is inefficient.  This should probably map to ExecutableElement instead.
-    /** Maps from an annotation name to the names of its elements that are Java expressions. */
-    private final Map<String, List<String>> annoToElements;
+    /** Maps from an annotation name to its elements that are Java expressions. */
```
mernst, line 107: Is this a simple name or a fully-qualified name?

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -136,7 +136,8 @@ public DependentTypesHelper(AnnotatedTypeFactory factory) {
 
         this.annoToElements = new HashMap<>();
         for (Class<? extends Annotation> expressionAnno : factory.getSupportedTypeQualifiers()) {
-            List<String> elementList = getExpressionElementNames(expressionAnno);
+            List<ExecutableElement> elementList =
+                    getExpressionElementNames(expressionAnno, factory.getProcessingEnv());
```
mernst, line 140: This is an odd name for the method if it returns ExecutableElements rather than names.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -156,24 +157,27 @@ public boolean hasDependentAnnotations() {
...
+     * expressions, namely those annotated with {@code @}{@link JavaExpression}.
      *
      * @param clazz annotation class
-     * @return a list of the names of elements in the annotation class that should be interpreted as
-     *     Java expressions
+     * @return a list of the elements in the annotation class that should be interpreted as Java
+     *     expressions
      */
-    private static List<String> getExpressionElementNames(Class<? extends Annotation> clazz) {
+    private static List<ExecutableElement> getExpressionElementNames(
```
mernst, line 167: Please change method name.

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -286,7 +287,14 @@ public AnnotationMirror build() {
     public void copyElementValuesFromAnnotation(
             AnnotationMirror valueHolder, String... ignorableElements) {
         Set<String> ignorableElementsSet = new HashSet<>(Arrays.asList(ignorableElements));
-        copyElementValuesFromAnnotation(valueHolder, ignorableElementsSet);
```
mernst, line 289: This method was changed.  I'm curious why the call didn't remain but was replaced by a `for` loop.

mernst (changes requested):
Disregard the misc CI failures for now.

javacutil/src/main/java/org/checkerframework/javacutil/AnnotationBuilder.java
```diff
@@ -274,32 +276,39 @@ public AnnotationMirror build() {
...
+   * @param ignorableElements the name of the elements of {@code other} that can be safely dropped
    */
-  public void copyElementValuesFromAnnotation(
-      AnnotationMirror valueHolder, String... ignorableElements) {
-    Set<String> ignorableElementsSet = new HashSet<>(Arrays.asList(ignorableElements));
-    copyElementValuesFromAnnotation(valueHolder, ignorableElementsSet);
+  public void copyElementValuesFromAnnotation(AnnotationMirror other, String... ignorableElements) {
+    for (Map.Entry<? extends ExecutableElement, ? extends AnnotationValue> eltValToCopy :
+        other.getElementValues().entrySet()) {
+      Set<String> ignorableElementsSet = new HashSet<>(Arrays.asList(ignorableElements));
```
mernst, line 286: It seems wasteful to re-create this same set on every iteration through the loop.
It is only used to test containment, so I think you could just do a containment check on the original array or on the `Arrays.asList()` result, without the extra cost of making a set.

mernst (approved):

PR #4491, opened 2021-03-30
Reflow comments

mernst:
Are changes like these worthwhile?  It makes more code fit within a screenfull.


smillst (approved):
I can see the benefit of these changes, but I don't feel strongly either way.

PR #4498, opened 2021-03-31
Warn if polymorphic annotations are used on a class declaration.

### LINKED ISSUE #1376: Better warning if a polymorphic qualifier is used on classes and fields

### LINKED ISSUE #4480: Issue warning when a polymorphic qualifier is misused
* Err if a polymorphic qualifier is written on a type variable declaration.
 * Err if a polymorphic qualifier is written in a class declaration, not in the body.

smillst:
Fixes  #4480 and fixes #1376.


framework/src/main/java/org/checkerframework/common/basetype/messages.properties
```diff
@@ -98,6 +98,7 @@ contracts.conditional.postcondition.false.methodref.invalid=Subclass postconditi
 lambda.unimplemented=This version of the Checker Framework does not type-check lambda expressions.
 methodref.inference.unimplemented=This version of the Checker Framework does not type-check method references with implicit type arguments.
 
+invalid.polymorphic.qualifier=Cannot use polymorphic qualifier %s %s
```
mernst, line 101: It seems that the second string is always "in a class declaration".  Should it be inline here, or should some of the strings be changed at call sites in the source code?

mernst (changes requested):
I added a test case for the first half of #4480, which this does not yet address.

framework/src/main/java/org/checkerframework/common/basetype/messages.properties
smillst: Now theres a call with a different string.

mernst (approved):

PR #4501, opened 2021-04-01
Merge JavaParserUtils into JavaParserUtil

mernst:


smillst (approved):
The changes look good.  I don't know why that test is failing.

kelloggm:
The test failure here is caused by the WPI tests, which failed because of kelloggm/do-like-javac#33 being merged earlier today, which caused a breaking change in do-like-javac. I've reverted it and am investigating now. Restarting that test should cause it to pass, but I don't seem to have permission to the pipeline, which I think I'd need to restart it myself. Sorry for the inconvenience.

PR #4502, opened 2021-04-01
Compute upper bound annotations for integer literals

mernst:
This is superseded by #4514.

--------

Merge #4509 first.

This PR also removes old code that was unsound for certain values of the `offset` element.


mernst:
@smillst This is massively inefficient, but more relevantly, it seems to be leaking lots of memory.  Do you have any ideas where that might be happening?

mernst:
Superseded by #4514

PR #4524, opened 2021-04-04
Don't format ajava files

jwaataja:
When I run `./gradlew wholeProgramInferenceTestCheckerAjavaTests` and then `./gradlew reformat`, it attempts to format the generated ajava files, filling the terminal with messages about attempting to format a non-Java file. This is because it attempts to format all files ending in `java`, which includes ajava files. It should instead collect files ending with `.java`.


mernst (approved):
Thanks!

PR #4530, opened 2021-04-05
Don't infer from recursive calls

mernst:


framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceImplementation.java
```diff
@@ -152,11 +156,60 @@ public void updateFromMethodInvocation(
...
+    // a completely unrelated method of the same name.
+
+    MethodTree enclosingMethod = TreePathUtil.enclosingMethod(methodInvNode.getTreePath());
+    if (enclosingMethod == null) {
+      return false;
+    }
+
+    Tree invoked = methodInvNode.getTree().getMethodSelect();
+    Name invokedMethodName;
+    switch (invoked.getKind()) {
```
smillst, line 197: This can be replaced with `org.checkerframework.javacutil.TreeUtils#getMethodName`.

smillst (approved):

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceImplementation.java
```diff
@@ -152,11 +156,60 @@ public void updateFromMethodInvocation(
...
   }
 
+  /**
+   * Returns true if the given call is a recursive call.
+   *
+   * @param methodInvNode a method invocation
+   * @return true if the given call is a recursive call
+   */
+  private boolean isRecursiveCall(MethodInvocationNode methodInvNode) {
+    // TODO: This tests only the name.  It should also ensure that the call isn't to an overload or
```
mernst, line 187: @smillst What is a way to test whether the call is actually a recursive call, as opposed to a call to a method of the same name?

smillst: I changed the method to do this.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceImplementation.java
```diff
@@ -152,11 +155,51 @@ public void updateFromMethodInvocation(
...
+    if (enclosingMethod == null) {
+      return false;
+    }
+    ExecutableElement methodInvocEle = TreeUtils.elementFromUse(methodInvNode.getTree());
+    ExecutableElement methodDeclEle = TreeUtils.elementFromDeclaration(enclosingMethod);
+    if (!methodDeclEle.equals(methodInvocEle)) {
+      return false;
+    }
+
+    if (ElementUtils.isStatic(methodDeclEle)) {
```
mernst, line 196: @smillst I don't think this is necessary (or desirable).  The above code has already established that the method is the same.  I think that inference should not be done for recursive calls, regardless of the value of the arguments (including the receiver).  I would change the above to `return methodDeclEle.equals(methodInvocEle);`.  Is there a reason this code is desirable?

smillst: I agree  that this isn't necessary.  I was think of calls to the super method, but in that case the elements would not be equal. So, please make that change.

PR #4532, opened 2021-04-05
Define `getAnnotationsFromJavaExpression()`

mernst:
Is this method worth defining?  We have no clients of it now, and it is inefficient.  But it performs a well-defined and useful purpose.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
+++ b/framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
@@ -899,7 +899,7 @@ public abstract class GenericAnnotatedTypeFactory<
   }
 
   /**
-   * Returns the primary annotation on an expression.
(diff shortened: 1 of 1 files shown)


framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -935,6 +935,51 @@ public AnnotationMirror getAnnotationFromJavaExpression(
...
   }
 
+  /**
+   * Returns the primary annotations on an expression, at a particular location.
+   *
+   * @param expr the expression for which the annotation is returned
+   * @param tree current tree
+   * @return the annotation on expression or null if one does not exist
+   */
+  public Set<AnnotationMirror> getAnnotationsFromJavaExpression(JavaExpression expr, Tree tree) {
```
smillst, line 945: I think this is a good change.  Could you change the overload to call this method?

smillst (approved):

PR #4536, opened 2021-04-05
Disable what isn't needed in `JavaParserConfiguration`.

smillst:


mernst (approved):
Could a ParserConfiguration be stored and re-used rather than a new one created for each invocation of the methods?  Or, is a new one essential because its state is side-effected?
Put another way, is it `JavaParser` or `ParserConfiguration` or both that accumulates data until memory is exhausted?

Could you write a comment about this?  Thanks.

PR #4542, opened 2021-04-06
Hack to work around lack of a tree

mernst:
This can be replaced by #4545.


smillst:
I think there is a tree now: https://github.com/typetools/checker-framework/pull/1390/files.

PR #4544, opened 2021-04-07
Rename `DependentTypesHelper.atReturnType` => `atMethodBody`

mernst:


framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -344,7 +344,7 @@ public void atFieldAccess(AnnotatedTypeMirror type, MemberSelectTree fieldAccess
    * @param atm the method return type; is side-effected by this method
```
smillst, line 344: The JavaDoc needs to be updated, too.

PR #4545, opened 2021-04-07
Remove test for null tree

mernst:
I can't find a test case for which the tree is null, but removing the code will help us find one.


smillst (approved):
It looks like dataflow was changed to add a tree for varargs: https://github.com/typetools/checker-framework/pull/1390/files.
So, I don't think this null check is required any more.

PR #4548, opened 2021-04-08
Don't blow up on BINDING_VARIABLE elements; fixes #4410 

### LINKED ISSUE #4410: dataflow support for Java 15: bad element kind BINDING_VARIABLE
This [switch in `CFGTranslationPhaseOne`](https://github.com/typetools/checker-framework/blob/f501266c6a78ddfe264586ec6ae6fece6133b8a6/dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java#L2769-L2774) doesn't handle [ElementKind.BINDING_VARIABLE](https://github.com/openjdk/jdk/blob/d2f9e31ae4b82e031401c3efcbb499c59efa37d7/src/java.compiler/share/classes/javax/lang/model/element/ElementKind.java#L123-L127), which was introduced in Java 15 as part of support for pattern matching, which leads to crashes like the following:

```
[ERROR]   Caused by: org.checkerframework.javacutil.BugInCF: bad element kind BINDING_VARIABLE
[ERROR]         at org.checkerframework.dataflow.cfg.CFGBuilder$CFGTranslationPhaseOne.visitIdentifier(CFGBuilder.java:4038)
[ERROR]         at org.checkerframework.dataflow.cfg.CFGBuilder$CFGTranslationPhaseOne.visitIdentifier(CFGBuilder.java:1526)
[ERROR]         at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCIdent.accept(JCTree.java:2427)
```

(If there's an existing bug tracking Java 15 support please feel free to dupe this one to it instead of tracking it separately, I'm filing this mostly to record [this crash report](https://github.com/google/error-prone/issues/2230) for posterity.)

craig-day:
I _think_ this fixes #4410 , or at least will prevent it from blowing up when using some newer features of Java. 

I would like to write a test case for this, but that would require the tests run with java 14 or java 15 and have preview features enabled.


mernst (approved):
Thanks for this code!

PR #4551, opened 2021-04-09
make equals between ThisReference more accurate

### LINKED ISSUE #4550: ThisReference JavaExpressions are always equal, even when they refer to different types
The `equals` and `syntacticEquals` methods of [ThisReference](https://github.com/typetools/checker-framework/blob/master/dataflow/src/main/java/org/checkerframework/dataflow/expression/ThisReference.java) return true whenever the other Object/JavaExpression is another `ThisReference`. However, not all `ThisReference`s are equal: for example, in a static inner class, one could have a JavaExpression for e.g `Outer.this` or for `Outer.Inner.this`. 

This problem arose for us during development of a custom checker: https://github.com/kelloggm/object-construction-checker/issues/368. The solution for that bug (https://github.com/kelloggm/object-construction-checker/pull/376) is more complex, but I'd like to direct your attention to the [representSame](https://github.com/kelloggm/object-construction-checker/blob/945b469f6b892ea7e18c2b1323318ef263a1a8ab/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/MustCallInvokedChecker.java#L344) method, which I needed to write to work around this problem. I'll open a PR that implements that fix in `ThisReference` as soon as I finish writing up this issue, though I don't have a good test case for this that uses any of the built-in checkers.

kelloggm:
Fixes #4550

### DIFF
--- a/dataflow/src/main/java/org/checkerframework/dataflow/expression/ThisReference.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/expression/ThisReference.java
@@ -13,7 +13,8 @@ public class ThisReference extends JavaExpression {
 
   @Override
   public boolean equals(@Nullable Object obj) {
-    return obj instanceof ThisReference;
+    return obj instanceof ThisReference
+        && type.toString().equals(((ThisReference) obj).type.toString());
   }
 
   @Override
@@ -48,7 +49,8 @@ public class ThisReference extends JavaExpression {
 
   @Override
   public boolean syntacticEquals(JavaExpression je) {
-    return je instanceof ThisReference;
+    return je instanceof ThisReference
+        && type.toString().equals(((ThisReference) je).type.toString());
   }
 
   @Override


dataflow/src/main/java/org/checkerframework/dataflow/expression/ThisReference.java
```diff
@@ -13,7 +13,8 @@ public ThisReference(TypeMirror type) {
 
   @Override
   public boolean equals(@Nullable Object obj) {
-    return obj instanceof ThisReference;
+    return obj instanceof ThisReference
+        && type.toString().equals(((ThisReference) obj).type.toString());
```
mernst, line 17: It is inefficient to call toString just for the purpose of comparison.  Instead, use a comparison method in `javax.lang.model.util.Types`.

Also, if you redefine `equals`, please define `hashCode` as well.

mernst (changes requested):

kelloggm:
It seems that the handling of pre-condition and post-condition contracts relies on the behavior that this PR was intended to change. In particular, [BaseTypeVisitor](https://github.com/typetools/checker-framework/blob/19f8c30b496b8c9c8ec4a4ffc877b88c4243c84f/framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java#L4161) relies on `ThisReference`s in a subclass being `equals` to `ThisReference`s in a superclass, which causes this PR to fail the tests.

I think this must therefore be intended behavior, so I'm going to close this PR.

Issue #3039, opened 2020-01-07
Make StubGenerator suggest Java 9+ if it fails with NoClassDefFoundError


cpovirk:
I tried the stub generator and encountered this error:

```
$ $JAVA8/bin/java -cp checker/dist/checker.jar org.checkerframework.framework.stub.StubGenerator java.lang.String
Exception in thread "main" java.lang.NoClassDefFoundError: com/sun/tools/javac/util/Context
  at org.checkerframework.framework.stub.StubGenerator.main(StubGenerator.java:417)
Caused by: java.lang.ClassNotFoundException: com.sun.tools.javac.util.Context
  at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
  at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
  at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:350)
  at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
  ... 1 more
```

I figured it was a problem with the version of Java I was using, and sure enough, it worked with 9, 10, and 11:

```
$ $JAVA9/bin/java -cp checker/dist/checker.jar org.checkerframework.framework.stub.StubGenerator java.lang.String
package java.lang;

class String implements Serializable, Comparable<String>, CharSequence {
...
```

I was originally going to suggest [documenting](https://checkerframework.org/manual/#stub-creating-without-source) the version requirement, but maybe it would be nicer for the stub generator to catch the error and output some specific advice then?

Then again, probably not a lot of people use the stub generator, and probably a lot of them are using Java 9+, and probably the remainder of them will figure this out relatively quickly :) So this is not likely to be a very important issue to take action on.

SwingGuy1024:
I've run into this, too. Personally, I would be happy if the checker framework manual simply pointed out that the stub generator required java 9 or higher. That way I wouldn't have wasted time trying to figure out what I was doing wrong using java 8. 

I will point out that most of the places I've worked or interviewed are still using Java 8. I believe Oracle's statistics say the same thing, although I don't have a reference for that. It seems like very few people have upgraded to newer versions. I'm looking for a new job right now, and every description that mentions what version of java they want has said Java 8, except for some that actually say Java 7! That's why I think a small change to the manual is important.

mernst:
Fixed by https://github.com/typetools/checker-framework/commit/c60a5ef8673c425a4238f3425f7ee663cf0827c3

PR #4554, opened 2021-04-11
Remove StubGenerator section from manual

mernst:
StubGenerator only works with JDK 8.
The manual section is only commented out, and the code is retained, because one day we might reimplement StubGenerator to work with Java 9+.


smillst:
I think this fixes:
https://github.com/typetools/checker-framework/issues/3039

smillst (approved):

PR #4555, opened 2021-04-12
Add note about putting quals/ directory in its own project

mernst:
Does the new text use the correct terminology?

### DIFF
--- a/docs/manual/creating-a-checker.tex
+++ b/docs/manual/creating-a-checker.tex
@@ -144,13 +144,16 @@ To illustrate, a checker for MyProp must be laid out as follows:
 \begin{Verbatim}
 myPackage/
   | qual/                               type qualifiers
-  | MyPropChecker.java                  [optional] interface to the compiler
+  | MyPropChecker.java                  interface to the compiler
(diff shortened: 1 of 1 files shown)


docs/manual/creating-a-checker.tex
```diff
@@ -144,13 +144,16 @@
...
+  | MyPropChecker.java                  interface to the compiler
   | MyPropVisitor.java                  [optional] type rules
   | MyPropAnnotatedTypeFactory.java     [optional] type introduction and dataflow rules
 \end{Verbatim}
 %
-Note that \<MyPropChecker.java> is required unless you are building on the
-Subtyping Checker.
+\<MyPropChecker.java> is occasionally optional, such as if you are building
+on the Subtyping Checker.  You may wish to put the \<qual/> directory in a
+separate Maven/Gradle project, if you want to create an artifacts
```
smillst, line 154: `an artifacts` => `an artifact`.  

The Maven term is "module" and the Gradle term is "subproject", but I think what you have written here makes sense.

smillst (approved):

PR #4557, opened 2021-04-13
Add stub file for JUnit's `Assertions` class

mernst:


checker/src/main/java/org/checkerframework/checker/nullness/junit-assertions.astub
```diff
@@ -0,0 +1,3453 @@
+/*
+ * Copyright 2015-2021 the original author or authors.
+ *
+ * All rights reserved. This program and the accompanying materials are
+ * made available under the terms of the Eclipse Public License v2.0 which
+ * accompanies this distribution and is available at
+ *
+ * https://www.eclipse.org/legal/epl-v20.html
```
smillst, line 8: Do we need to update our licenses file and the licenses of the Maven artifact to include this?

checker/src/main/java/org/checkerframework/checker/nullness/NullnessChecker.java
```diff
@@ -31,6 +32,7 @@
   NullnessChecker.LINT_TRUSTARRAYLENZERO,
   NullnessChecker.LINT_PERMITCLEARPROPERTY
 })
+@StubFiles({"junit-assertions.astub"})
```
smillst, line 35: This causes the stub file to be parsed every time the checker is used.  We should probably put it somewhere to be passed only if using JUnit.

mernst: @smillst I'm concerned that it will not be very discoverable by users if they have to supply a special command line option explicitly.
Would this be acceptable if the stub file were pared down to be minimal?
Or should we document `@StubFiles` as to be used only for JDK annotations and not for any other libraries?

checker/src/main/java/org/checkerframework/checker/nullness/junit-assertions.astub
mernst: We should probably recreate the file from scratch, minimally, using the public API.  Then licensing is not an issue.

PR #4560, opened 2021-04-14
Don't make purity suggestions about synthetic methods

mernst:
An example is the implicit constructor of a class, when the superclass's no-argument constructor is pure.

Merge with https://github.com/typetools/annotation-tools/pull/388 and https://github.com/typetools/jdk/pull/103


smillst (approved):

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -986,7 +986,9 @@ protected void checkPurity(MethodTree node) {
       reportPurityErrors(r, node, kinds);
     }
 
-    if (suggestPureMethods) {
+    // This may make a suggestion about an implicit default constructor, which is not considered
+    // synthetic.
+    if (suggestPureMethods && !TreeUtils.isSynthetic(node)) {
```
smillst, line 991: Should the documentation of `isSynthetic` be update?

The implementation looks like it handles implicit default constructors.  But maybe a "generated" constructor is not an implicit default constructor?

smillst (approved):

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
mernst: You are right: `Flags.GENERATEDCONSTR` is exactly for the default constructor.

PR #4565, opened 2021-04-15
Set map capacity, to avoid allocating extra memory

mernst:


framework/src/main/java/org/checkerframework/framework/flow/CFAbstractStore.java
```diff
@@ -120,8 +121,15 @@ public long getUid() {
...
 
+  /**
+   * Creates a new CFAbstractStore.
+   *
+   * @param analysis the analysis class this store belongs to
+   * @param sequentialSemantics should the analysis use sequential Java semantics?
+   */
   protected CFAbstractStore(CFAbstractAnalysis<V, S, ?> analysis, boolean sequentialSemantics) {
     this.analysis = analysis;
+    // TODO: Can we save space by starting these maps smaller than the default of 16?
```
smillst, line 132: I don't have answer to do this todo.  Should we just remove the todo?

smillst (approved):

PR #4573, opened 2021-04-20
Add `-AassumeKeyFor` command-line argument

mernst:


framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
```diff
@@ -814,13 +817,47 @@ public void typeProcessingOver() {
...
+    return this.options;
+  }
+
+  /**
+   * Like {@link #getOptions}, but only includes options provided to this checker. Does not include
+   * those passed to subcheckers.
+   *
+   * @return the the active options for this checker, not including those passed to subcheckers
+   */
+  public Map<String, String> getOptionsNoSubcheckers() {
```
smillst, line 838: Is this related to the other changes in this pull request? It's fine if it's not, I'm just wondering if I'm missing something.

smillst (approved):

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeChecker.java
mernst: This is used in `hasOptionNoSubcheckers`, which is used in `NullnessChecker.java`.  Using `hasOption` there led to an infinite loop.  Is there a better way?

smillst: Oh, I see it now.  Thanks!  I don't see a better way to do this.  It looks like other checkers have had to work around this, too and should now be changed to call this method.  I create a pull request with the other places I've found that should now call `hasOptionNoSubcheckers`.

PR #4589, opened 2021-04-28
Abstract out method `isPrivateInJdk`

mernst:
Maybe private members should always be ignored, not just for the JDK.  https://github.com/typetools/checker-framework/pull/4592 does that.


framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileParser.java
```diff
@@ -703,6 +709,19 @@ private void processPackage(PackageDeclaration packDecl) {
     // TODO: Handle atypes???
   }
 
+  /**
+   * Returns true if the given program construct is in the annotated JDK and is private.
+   *
+   * <p>If so, it is skipped. Private consturcts can't be referenced outside of the JDK and might
```
smillst, line 715: `consturcts` => `constructs`.

smillst (approved):

smillst:
I would close this pull request and merge https://github.com/typetools/checker-framework/pull/4592 instead.

PR #4592, opened 2021-04-28
Method `skipNode()` indicates whether to skip a stub file construct

mernst:


smillst (approved):
I think this change makes sense.  I would merge this pull request rather than https://github.com/typetools/checker-framework/pull/4589.

Issue #4614, opened 2021-05-02
Parser error for valid Java code


jpschewe:
I'm using checker 3.12.0 with OpenJDK 11.

This code
```java
        final String versionInfo = Version.getAllVersionInformation().entrySet().stream() //
                                          .map(e -> String.format("%s:%s", e.getKey(), e.getValue())) //
                                          .collect(Collectors.joining("\n"));
```

Results in the parser error
```
warning: [expression.unparsable.type.invalid] Expression invalid in dependent type annotation: [error for expression: ()java.util.Map<java.lang.String,java.lang.String>.getAllVersionInformation(); error: Invalid '()java.util.Map<java.lang.String,java.lang.String>.getAllVersionInformation()' because the expression did not parse. Error message: Encountered unexpected token: "java" <IDENTIFIER>]
/home/jpschewe/projects/fll-sw/working-dir/src/main/java/fll/web/GatherBugReport.java:91: warning: [lambda.param.type.incompatible] incompatible parameter types for parameter 0 in lambda expression.
                                          .map(e -> String.format("%s:%s", e.getKey(), e.getValue())) //
                                               ^
  found   : Entry<@KeyFor("()java.util.Map<java.lang.String,java.lang.String>.getAllVersionInformation()") String, String>
  required: Entry<@KeyFor("[error for expression: ()java.util.Map<java.lang.String,java.lang.String>.getAllVersionInformation(); error: Invalid '()java.util.Map<java.lang.String,java.lang.String>.getAllVersionInformation()' because the expression did not parse. Error message: Encountered unexpected token: "java" <IDENTIFIER>]") String, String>
```

jpschewe:
Stand-alone example.

```java
package net.mtu.eggplant.checker.parser_error;

import java.io.IOException;
import java.io.InputStream;
import java.util.Collections;
import java.util.HashMap;
import java.util.Map;
import java.util.Properties;
import java.util.stream.Collectors;

/**
 * Get version information about the FLL software.
 *
 * @author jpschewe
 */
public final class Version {

  private static final Map<String, String> VERSION_INFORMATION;

  private static final String VERSION;
  static {
    final InputStream stream = Version.class.getResourceAsStream("/fll/resources/git.properties");
    if (null == stream) {
      System.out.println("Unable to find version.properties!");
      VERSION = "NO-PROPERTIES-FILE";
      VERSION_INFORMATION = Collections.emptyMap();
    } else {
      final Properties versionProps = new Properties();
[...]

PR #4628, opened 2021-05-04
Enable tracing where a buggy lambda came from

mernst:
This helped diagnosing #4614.  Is there a better way?


smillst (approved):
I don't quite understand how you used this debugging output.   I guess you output the reference id of the lambda else where in the code and then compared the id to the output here?  I use the IntelliJ debugger to walk through the code, so I wouldn't use this logging.  But it's ok with me if you want to commit it.

PR #4630, opened 2021-05-04
Assertion about argument to ClassName

mernst:


dataflow/src/main/java/org/checkerframework/dataflow/expression/ClassName.java
```diff
@@ -17,10 +20,28 @@
...
     super(type);
+    TypeMirror baseType;
+    TypeKind baseKind = type.getKind();
+    if (baseKind != TypeKind.ARRAY) {
+      baseType = type;
+    } else {
+      baseType = TypesUtils.getInnermostComponentType((ArrayType) type);
+      baseKind = baseType.getKind();
+    }
+    // Primitives, arrays, and void are permitted so FieldAccess can represent class literals.
```
smillst, line 36: ` FieldAccess can represent class literals.` I thought that `ClassName` represents class literals?

dataflow/src/main/java/org/checkerframework/dataflow/expression/FieldAccess.java
```diff
@@ -11,14 +11,30 @@
 import org.checkerframework.javacutil.ElementUtils;
 import org.checkerframework.javacutil.TypesUtils;
 
+/**
+ * A FieldAccess represents a field access or the occurrence of {@code .class} in source code, e.g.,
+ * SomeClass.class, SomeClass[].class, int.class, or int[].class.
```
smillst, line 16: I thought `ClassName` represents class literals?

smillst (changes requested):

dataflow/src/main/java/org/checkerframework/dataflow/expression/ClassName.java
mernst: ClassName represents class names, such as `int` or `String`.
The JLS uses the term "class literal" to mean a class name or array, followed by `.class`, such as `int.class` or `String.class` or `int[].class`.
Maybe you were thinking of "class literal" differently, eg as a class name?  Or did you think that ClassName represents `int.class` and `String.class`?

smillst: > Or did you think that ClassName represents int.class and String.class?

Yes, that's what I think.  You also changed the Javadoc of `ClassName` to say that it represents class literals.  Also, the `toString`  method appends `".class"' to `typeString`.

smillst: Oh I see that Javadoc says `"as part of ... a class literal"`.

smillst: Currently `ClassName` are used for class literals, the whole class literal not just the "class name" part.  See:
https://github.com/typetools/checker-framework/blob/d6de8b7a0f66f35e70fa285b5b3f1def0969cc41/framework/src/main/java/org/checkerframework/framework/util/JavaExpressionParseUtil.java#L825-L835

smillst: Also here:
https://github.com/typetools/checker-framework/blob/d6de8b7a0f66f35e70fa285b5b3f1def0969cc41/dataflow/src/main/java/org/checkerframework/dataflow/expression/JavaExpression.java#L276-L281

mernst: In that case, there are several other bugs (and missing/misleading documentation) that led me astray.
I opened https://github.com/typetools/checker-framework/pull/4636 to address it.

dataflow/src/main/java/org/checkerframework/dataflow/expression/FieldAccess.java
mernst: In the current code, it sometimes does and sometimes does not.  I guessed wrong about how to rectify the inconsistency...

mernst:
@smillst ping

smillst (approved):

PR #4635, opened 2021-05-04
Stub files can override JDK declaration annotations

mernst:


docs/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ Version 3.?.? (June 1, 2021)
 
 **User-visible changes:**
 
+Stub files can now override declaration annotations in the annotated JDK.
+The could always override type annotations in the annotated JDK.
```
smillst, line 7: `They could`

framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileParser.java
```diff
@@ -2615,10 +2616,27 @@ private void builderSetValue(AnnotationBuilder builder, String name, Object valu
...
+        // JDK annotations should not replace any annotation of the same type.
+        List<DeclaredType> storedAnnoTypes =
+            CollectionsPlume.mapList(AnnotationMirror::getAnnotationType, stored);
+        Types typeUtils = processingEnv.getTypeUtils();
+        for (AnnotationMirror anno : annos) {
+          DeclaredType annoType = anno.getAnnotationType();
+          boolean existing =
+              storedAnnoTypes.stream()
+                  .anyMatch(storedAnnoType -> typeUtils.isSameType(annoType, storedAnnoType));
+          if (!existing) {
```
smillst, line 2635: Instead of computing existing, you could just use `AnnotationUtils.containsSameByName(stored, anno)`.  That should be faster than comparing types.

smillst (approved):

PR #4636, opened 2021-05-05
Make fromNodeFieldAccess return a ClassName for class literals

mernst:


javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -968,19 +968,23 @@ public static boolean isClassLiteral(Tree tree) {
...
+   * @param tree a tree that might be a field access
    * @return true iff if tree is a field access expression (implicit or explicit)
    */
   public static boolean isFieldAccess(Tree tree) {
     if (tree.getKind() == Tree.Kind.MEMBER_SELECT) {
-      // explicit field access
+      // explicit member access (or a class literal)
       MemberSelectTree memberSelect = (MemberSelectTree) tree;
+      if (memberSelect.getIdentifier().contentEquals("class")) {
+        return false;
```
smillst, line 986: I wouldn't make this change.  Most places in the Checker Framework, we expect class literals to be field accesses; it's just JavaExpressions that treat them differently.  (It would probably make more sense to change JavaExpressions so the class literals are FieldAccess.  This would also simplify some code.)

smillst (approved):

javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
mernst: Where does the Checker Framework expect class literals to be field accesses (that this isn't forced on us by other tools)?
I think a better design would be to treat class literals differently than field accesses and than class names, since they aren't really either one of those.  I didn't want to do such a bit refactoring in this pull request, however.

smillst: > Where does the Checker Framework expect class literals to be field accesses (that this isn't forced on us by other tools)?

At all calls to this method.  Also, dataflow converts class literals to FieldAccessNodes. Making this method return false is changing that behavior. 

> I think a better design would be to treat class literals differently than field accesses and than class names, since they aren't really either one of those. I didn't want to do such a bit refactoring in this pull request, however.

I'm fine with that design and it makes sense not to make that change now.  But since you aren't doing the full change now, I would hold off on making this method return false for class literals.  Or if you want to make the change, then the Javadoc should make it explicit that class literals are not field accesses even though javac treats them as such.

PR #4639, opened 2021-05-05
Add the version number to the shaded dataflow package names.

smillst:


mernst (approved):
Thanks!

dataflow/build.gradle
```diff
@@ -4,16 +4,24 @@ dependencies {
...
 shadowJar {
     archiveFileName = "dataflow-shaded.jar"
     // Without this line, the Maven artifact will have the classifier "all".
     archiveClassifier = ''
-    relocate 'org.checkerframework.dataflow', 'org.checkerframework.shaded.dataflow'
-    relocate 'org.checkerframework.javacutil', 'org.checkerframework.shaded.javacutil'
+    archiveName('dataflow-shaded')
+    // The version number includes both "." and possibly "-SNAPSHOT", so convert it to a legal
+    // package name.
+    String shadedPkg = "shaded_${project.version}".replace('-', '_').replace('.','_')
```
msridhar, line 19: Interesting!  Putting the version number in the package name does make upgrading slightly more difficult for users (a search and replace on relevanat import statements should work).  But, it does allow for multiple different versions of `dataflow-shaded` to be in the classpath without conflict.  This could enable NullAway to switch to `dataflow-shaded` in the future and no longer conflict with unshaded `dataflow` or the `dataflow-shaded` used by Error Prone.  Nice!

msridhar: On further thought, I am less sure that having the version number in the package name is a good idea.  The problem is that typical build systems like Gradle will only allow at most one version of an artifact on either the classpath or processor path.  By default it chooses the most recent version requested via a dependency.  If the classes in different versions of the `dataflow-shaded` artifact have different fully-qualified names, this will almost certainly lead to a conflict.  E.g., if Error Prone uses `dataflow-shaded:1.1` and NullAway uses `dataflow-shaded:1.2`, Error Prone will break, since only 1.2 will be present in the processor path and all the classes from 1.1 will appear to me "missing."  This could be solved by also renaming the _artifact_ in each version (`dataflow-shaded-1_1`, `dataflow-shaded-1_2`, etc.), but I'm not sure that is easy or desirable.  @smillst any thoughts?

smillst: Thanks for pointing this out! I agree this still doesn't fix the problem and probably makes things worse.  Here are two possible solutions:

1. Include the version number in the artifact id so that each new version is a new artifact.
2. Include all versions for the dataflow-shaded packages in the same jar.  So, dataflow-shaded version 3.15 would include `org.checkerfamework.shaded_3.13.1`,`org.checkerfamework.shaded_3.14.0`, and `org.checkerfamework.shaded_3.15.0`.  This would still cause problems with the different versions of checker-qual, but I still need to look closely on how dataflow uses checker-qual.

msridhar: Of these two possible solutions, I think 1 is better.  (2 will lead to unbounded bloat of the `dataflow-shaded` artifact and that might cause issues in the future.)  But for 1, I am not sure about the implications of having a differently-named artifact in each Checker Framework release; I haven't seen other projects do something like this.

I personally would be ok with a more incremental approach.  We could first shade plume-util (using a fixed naming scheme), which mostly solves the highest-priority issue of potential conflicts between `dataflow` and `dataflow-shaded`.  We could then think more about how common an issue other conflicts are, and do a bit more digging to make sure there won't be unexpected impacts from 1 before going for it.

PR #4644, opened 2021-05-08
Test the javac scripts

mernst:
Some recent change (most likely shading) has broken the `checker/bin-devel/javac` script.

PR #4653, opened 2021-05-11
Correct `checker/bin-devel/javac-debug` and `checker/bin-devel/javac`

smillst:
This includes the changes in #4644.

### DIFF
--- a/checker/bin-devel/javac
+++ b/checker/bin-devel/javac
@@ -32,7 +32,7 @@ resources="build/resources/main"
 
 buildDirs="${cfDir}"/dataflow/"${classes}":"${cfDir}"/javacutil/"${classes}":"${cfDir}"/framework/"${classes}":"${cfDir}"/framework/"${testclasses}":"${cfDir}"/framework/"${resources}":"${cfDir}"/checker/"${classes}":"${cfDir}"/checker/"${resources}":"${cfDir}"/checker-qual/"${classes}":"${stubparserDir}"/javaparser-core/stubparser.jar:$( (cd "${cfDir}" && ./gradlew -q :checker:printPlumeUtilJarPath) )
 
-echo $buildDirs
+echo "${buildDirs}"
 
 ## Preserve quoting and spaces in arguments, which would otherwise be lost
 ## due to being passed through the shell twice.
--- a/checker/bin-devel/javac
+++ b/checker/bin-devel/javac
@@ -29,7 +29,7 @@ classes="build/classes/java/main"
 testclasses="build/classes/java/test"
 resources="build/resources/main"
 
-buildDirs="${cfDir}"/dataflow/"${classes}":"${cfDir}"/javacutil/"${classes}":"${cfDir}"/framework/"${classes}":"${cfDir}"/framework/"${testclasses}":"${cfDir}"/framework/"${resources}":"${cfDir}"/checker/"${classes}":"${cfDir}"/checker/"${resources}":"${cfDir}"/checker-qual/"${classes}":"${stubparserDir}"/javaparser-core/stubparser.jar:$( (cd "${cfDir}" && ./gradlew -q :checker:printPlumeUtilJarPath) )
+buildDirs="${cfDir}"/dataflow/"${classes}":"${cfDir}"/javacutil/"${classes}":"${cfDir}"/framework/"${classes}":"${cfDir}"/framework/"${testclasses}":"${cfDir}"/framework/"${resources}":"${cfDir}"/checker/"${classes}":"${cfDir}"/checker/"${resources}":"${cfDir}"/checker-qual/"${classes}":"${stubparserDir}"/javaparser-core/stubparser.jar:$( (cd "${cfDir}" && ./gradlew -q :checker:printClasspath) )
 
 ## Preserve quoting and spaces in arguments, which would otherwise be lost
 ## due to being passed through the shell twice.
--- a/checker/bin-devel/javac
+++ b/checker/bin-devel/javac
@@ -24,7 +24,7 @@ binaryDir="${mydir}"/../dist
 javacJar=${binaryDir}/javac.jar
 
 cfDir="${mydir}"/../..
-annoToolsDir="${cfDir}"/../annotation-tools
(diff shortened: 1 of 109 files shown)


checker/bin-devel/javac
```diff
@@ -24,13 +24,12 @@ binaryDir="${mydir}"/../dist
 javacJar=${binaryDir}/javac.jar
 
 cfDir="${mydir}"/../..
-annoToolsDir="${cfDir}"/../annotation-tools
```
mernst, line 27: I think this change means that running the `compileJava` in AFU and CF doesn't work; it is necessary to do `./gradlew assemble` if AFU might have changed.  Is that correct?

checker/build.gradle
```diff
@@ -98,9 +98,9 @@ task copyJarsToDist(dependsOn: shadowJar, group: 'Build') {
 
 assemble.dependsOn copyJarsToDist
 
-task printPlumeUtilJarPath {
-    description "Print the path to plume-util.jar"
-    doFirst { println project.configurations.compile.find { it.name.startsWith("plume-util") } }
+task printClasspath {
+    description "Print the classpath for compileJava"
+        doFirst { println compileJava.classpath.asPath }
```
mernst, line 103: Stray spacing.

checker/bin-devel/javac-debug
```diff
@@ -24,12 +24,11 @@ binaryDir="${mydir}"/../dist
 javacJar=${binaryDir}/javac.jar
 
 cfDir="${mydir}"/../..
-annoToolsDir="${cfDir}"/../annotation-tools
 stubparserDir="${cfDir}"/../stubparser
 classes="build/classes/java/main"
 resources="build/resources/main"
 
-buildDirs="${cfDir}"/dataflow/"${classes}":"${cfDir}"/javacutil/"${classes}":"${cfDir}"/framework/"${classes}":"${cfDir}"/framework/"${resources}":"${cfDir}"/checker/"${classes}":"${cfDir}"/checker/"${resources}":"${cfDir}"/checker-qual/"${classes}":"${stubparserDir}"/javaparser-core/stubparser.jar:"${annoToolsDir}"/scene-lib/bin:"${annoToolsDir}"/annotation-file-utilities/annotation-file-utilities-all.jar:$( (cd "${cfDir}" && ./gradlew -q :checker:printPlumeUtilJarPath) )
+buildDirs="${cfDir}"/dataflow/"${classes}":"${cfDir}"/javacutil/"${classes}":"${cfDir}"/framework/"${classes}":"${cfDir}"/framework/"${resources}":"${cfDir}"/checker/"${classes}":"${cfDir}"/checker/"${resources}":"${cfDir}"/checker-qual/"${classes}":"${stubparserDir}"/javaparser-core/stubparser.jar:$( (cd "${cfDir}" && ./gradlew -q :checker:printClasspath) )
```
mernst, line 31: Is it not possible to use the jar file because of shading, or for some other reason?

mernst (changes requested):
I have a couple questions.

If it is too error-prone or confusing to support use of `compileJava`, we could drop that as a feature of `bin-devel/javac`.

checker/bin-devel/javac
smillst: Oh, I can put that back.  It only included the class files from scene-lib before and used the wrong location:
```
"${annoToolsDir}"/scene-lib/bin
```

I can fix it so that it is the correct location.

smillst: I fixed this.

mernst:
> > If it is too error-prone or confusing to support use of compileJava, we could drop that as a feature of bin-devel/javac.
> 
> I think it's ok now, because all of checker's classpath will be printed by `printClasspath`.

But perhaps there is little point to the script now?  I believe its purpose is to reduce the edit-compile-test cycle by permitting running `./gradlew compileJava` rather than `./gradlew assemble` before running the Checker Framework.  If the embedded gradle command is expensive, maybe it is best to reduce complexity and our maintenance burden by deleting the bin-devel/javac script entirely, and tell developers to alwas run `.gradlew assemble`.

Also, do you use the bin-devel/javac-debug script?  If we retain it, this pull request should update it to keep it in sync with bin-devel/javac.

mernst (approved):

smillst:
I run `./gradlew copyJarsToDist` then use `checker/bin/javac`.  If I want to debug, I pass `-J-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005` to `checker/bin/java`.  So I don't use either script in `checker/bin-devel`.   

`./gradlew copyJarsToDist`  is just as fast as `./gradlew compileJava`, so I don't think theres any point  to `checker/bin-devel/javac`.

>EDIT: Or, at least to remove that feature of it. Maybe there are other differences between CheckerDevelMain and CheckerMain, but they should be documented.

I'll check on this.  There seems to be a few difference. One example is `CheckerDevelMain` adds `-J-ea`.

> Also, do you use the bin-devel/javac-debug script? If we retain it, this pull request should update it to keep it in sync with bin-devel/javac.

I don't use it.  This pull request does update it.

mernst:
> I run `./gradlew copyJarsToDist` then use `checker/bin/javac`. If I want to debug, I pass `-J-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005` to `checker/bin/java`. So I don't use either script in `checker/bin-devel`.
> 
> `./gradlew copyJarsToDist` is just as fast as `./gradlew compileJava`, so I don't think theres any point to `checker/bin-devel/javac`.

OK.  I generally run `./gradlew assemble`, which looks like it is the same as `./gradlew copyJarsToDist`.

> > Also, do you use the bin-devel/javac-debug script? If we retain it, this pull request should update it to keep it in sync with bin-devel/javac.
> 
> I don't use it. This pull request does update it.

I see that the pull request updates it (thanks!), but it has gotten out of sync due to some unrelated change.  This may be another reason to eliminate it.

smillst:
We decided to delete the scripts and CheckerDevelMain.

smillst:
Closed in favor of https://github.com/typetools/checker-framework/pull/4664.

PR #4660, opened 2021-05-12
Add a NullAway version of the dataflow-shaded artifact.

smillst:


mernst (approved):
LGTM

smillst:
@msridhar Do this work for you?

dataflow/build.gradle
```diff
@@ -9,23 +9,43 @@ dependencies {
...
+    }
 }
 
-artifacts {
-    archives shadowJar
-}
+// Creates a new shaded datflow artifact.  To add a new one, add a new method call below, and add
+// the new jar to publishing and signing blocks.
+createDataflowShaded('shaded')
+createDataflowShaded('NullAway')
```
msridhar, line 48: While the official name of the tool is NullAway, in retrospect I wish we had named the github repo just uber/nullaway instead of retaining the capitalization there.  Here, for package naming, I think `nullaway` would be better than `NullAway`

dataflow/build.gradle
```diff
@@ -124,9 +147,17 @@ publishing {
...
+            dataflowShadedPom(it, 'shaded')
+
+            artifact getTasksByName('datflowshadedJar',false).iterator().next()
+            artifact sourcesJar
+            artifact javadocJar
+        }
 
-            artifact shadowJar
+        datflowShadedNullAway(MavenPublication) {
+            dataflowShadedPom(it, 'NullAway')
```
msridhar, line 158: Possibly the same capitalization issue here as above

msridhar (commented):
This looks great based on what I read!  It will take me a little time to locally test if NullAway works using this new shaded artifact; I will report back when that is done.

msridhar (changes requested):
@smillst when I run `./gradlew publishToMavenLocal` and look at the resulting `dataflow-nullaway-3.13.1-SNAPSHOT.jar` file, it seems the `javacutil` and `plume-util` classes are missing.  E.g., `TreeUtils` is not present: running `unzip -l dataflow-nullaway-3.13.1-SNAPSHOT.jar | grep TreeUtils` does not yield any results.  The same issue exists for `dataflow-shaded-3.13.1-SNAPSHOT.jar`.  So I think something is not quite right here.

smillst:
@msridhar Thanks for catching that!  I fixed the jar to actually include all the classes files.

msridhar (approved):
I tested locally with NullAway and this latest version works!

PR #4661, opened 2021-05-12
Make nullness handling of `Arrays.copyOf` more precise

mernst:
~~Merge after #4658~~

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
+++ b/checker/src/main/java/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
@@ -16,6 +16,7 @@ import com.sun.source.tree.TypeCastTree;
 import com.sun.source.tree.UnaryTree;
 import com.sun.source.tree.VariableTree;
 import com.sun.source.util.TreePath;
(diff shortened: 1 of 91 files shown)


checker/src/main/java/org/checkerframework/checker/nullness/NullnessAnnotatedTypeFactory.java
```diff
@@ -534,6 +540,28 @@ public Void visitNewArray(NewArrayTree node, AnnotatedTypeMirror type) {
...
       return null;
     }
+
+    @Override
+    public Void visitMethodInvocation(MethodInvocationTree tree, AnnotatedTypeMirror type) {
+      if (TreeUtils.isMethodInvocation(tree, copyOfMethods, processingEnv)) {
+        List<? extends ExpressionTree> args = tree.getArguments();
+        ExpressionTree lengthArg = args.get(1);
+        if (TreeUtils.isArrayLengthAccess(lengthArg)) {
+          // TODO: This syntactic test may not be not correct if the array expression has a side
```
smillst, line 550: It seems unlikely that the side effect would impact the length of the array, so I'm fine leaving this TODO.

smillst (approved):

PR #4662, opened 2021-05-13
Improve debuggability of WPI tests and scripts

kelloggm:
Also some notes on requirements (especially for running on a Mac) and a few other minor bug fixes.

Full list of changes:
* `wpi.sh` and `wpi-many.sh` now use `set -eo pipefail`, meaning that they fail if any of their constituent commands fail
* `|| true` added to commands that should be allowed to fail, in conjunction with the last change
* trapping of errors to print a message with the line number in the offending script, along with context
* added `--` to `rm` commands, to avoid failing on directories and files whose names start with `-`
* removed double `&>` that prevented output from `wpi.sh` from being logged. Also, don't delete that log, and print it if no `*-typecheck.out` files are generated while running tests.
* if the `$DLJC` environment variable is set, use that for the `dljc` executable rather than `git-clone-related`. This has two benefits: (1) it makes debugging locally easier, and (2) it will permit do-like-javac's CI to avoid re-cloning itself, which I believe is a source of errors in the do-like-javac integration test that runs the CF's WPI tests when a user other than me makes a PR
* add the cause of WPI not running to the (previously-empty) `.cannot-run-wpi` file, to ease debugging local failures
* run `wpi-many.sh` during the `wpiManyTests` gradle task in skip rather than delete mode, so that (1) when running locally, you can inspect failing tests, and (2) when running in CI, the task can print relevant logs when it fails
*  related, before starting the main part of the `wpiManyTests` task, delete any `.cannot-run-wpi` files in the build directories, so that tests are actually re-run rather than skipped, if they have already failed locally
* when the `wpiManyTests` task fails, print the standard out/err of the last run of DLJC and the contents of any `.log` files in the result directories

### DIFF
--- a/checker/bin/wpi-many.sh
+++ b/checker/bin/wpi-many.sh
@@ -38,8 +38,8 @@ shift $(( OPTIND - 1 ))
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 SCRIPTPATH="${SCRIPTDIR}/wpi-many.sh"
 
-# report line numbers when the script fails, for debugging in CI, from
-# https://unix.stackexchange.com/a/522815
+# Report line numbers when the script fails for debugging. Adapted from
+# https://unix.stackexchange.com/a/522815.
 trap 'echo >&2 "Error - exited with status $? at line $LINENO of wpi-many.sh:";
          pr -tn ${SCRIPTPATH} | tail -n+$((LINENO - 3)) | head -n7' ERR
 
--- a/checker/bin/wpi-many.sh
+++ b/checker/bin/wpi-many.sh
@@ -43,7 +43,7 @@ SCRIPTPATH="${SCRIPTDIR}/wpi-many.sh"
 trap 'echo >&2 "Error - exited with status $? at line $LINENO of wpi-many.sh:";
          pr -tn ${SCRIPTPATH} | tail -n+$((LINENO - 3)) | head -n7' ERR
 
-echo "Starting wpi-many.sh. The output of this script is purely informational."
+echo "Starting wpi-many.sh."
 
 # check required arguments and environment variables:
 
@@ -307,4 +307,4 @@ fi
 
 export JAVA_HOME="${JAVA_HOME_BACKUP}"
 
-echo "Exiting wpi-many.sh. The output of this script was purely informational. Results were placed in ${OUTDIR}-results/."
+echo "Exiting wpi-many.sh. Results were placed in ${OUTDIR}-results/."
--- a/checker/bin/wpi-many.sh
+++ b/checker/bin/wpi-many.sh
@@ -6,6 +6,9 @@
 # section of the Checker Framework manual:
 # https://checkerframework.org/manual/#whole-program-inference
 
+set -eo pipefail
+# not set -u, because this script checks variables directly
+
 DEBUG=0
 # To enable debugging, uncomment the following line.
 # DEBUG=1
@@ -31,6 +34,14 @@ done
 
 # Make $@ be the arguments that should be passed to dljc.
 shift $(( OPTIND - 1 ))
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+SCRIPTPATH="${SCRIPTDIR}/wpi-many.sh"
+
+# report line numbers when the script fails, for debugging in CI, from
+# https://unix.stackexchange.com/a/522815
+trap 'echo >&2 "Error - exited with status $? at line $LINENO of wpi-many.sh:";
+         pr -tn ${SCRIPTPATH} | tail -n+$((LINENO - 3)) | head -n7' ERR
 
 echo "Starting wpi-many.sh. The output of this script is purely informational."
 
@@ -124,8 +135,6 @@ export JAVA_HOME="${JAVA11_HOME}"
 
 echo "Finished configuring wpi-many.sh. Results will be placed in ${OUTDIR}-results/."
 
-SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
-
 export PATH="${JAVA_HOME}/bin:${PATH}"
 
 mkdir -p "${OUTDIR}"
@@ -180,7 +189,7 @@ do
            continue
         fi
     else
-        rm -rf "${REPO_NAME}/dljc-out"
+        rm -rf -- "${REPO_NAME}/dljc-out"
     fi
 
     cd "./${REPO_NAME}" || (echo "command failed in $(pwd): cd ./${REPO_NAME}" && exit 5)
@@ -199,7 +208,7 @@ do
         # whose owner is the github user always causes an error on this line,
         # because the `unannotated` remote is already set. The output is piped
         # to /dev/null to suppress that error.
-        git remote add unannotated "${ORIGIN}" &> /dev/null
+        git remote add unannotated "${ORIGIN}" || true &> /dev/null
     fi
 
     REPO_FULLPATH=$(pwd)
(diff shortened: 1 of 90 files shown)


docs/manual/inference.tex
```diff
@@ -214,6 +214,14 @@
 \item If set, the \verb|JAVA8_HOME| environment variable must point to a Java 8 JDK.
 \item If set, the \verb|JAVA11_HOME| environment variable must point to a Java 11 JDK.
 \item \<CHECKERFRAMEWORK> environment variable must point to a built copy of the Checker Framework.
+\item If set, the \verb|DLJC| environment variable must point to a copy of the \<dljc> script
+from \ahref{https://github.com/kelloggm/do-like-javac}{do-like-javac}. (If this variable is not
+set, the WPI scripts will download this dependency automatically.)
+% This next item is true because of the use of ${*@Q}: @Q is a Bash 4.4 feature and causes "bad substitution"
```
mernst, line 220: I suggest placing a run-time check in the script, so that the manifestation of the problem is less obscure.  This text could be moved there.

mernst (changes requested):
`./gradlew inferenceTests` does not pass on CSE machines.  I can't interpret the error.
Could you please fix that?

I don't understand the "is purely informational" line that is output.  The opposite would be disinformational, but a user would assume that a script would not be intentionally disinformational.  Could you clarify that?

checker/bin/wpi-many.sh
```diff
@@ -32,6 +35,14 @@ done
 # Make $@ be the arguments that should be passed to dljc.
 shift $(( OPTIND - 1 ))
 
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+SCRIPTPATH="${SCRIPTDIR}/wpi-many.sh"
+
+# report line numbers when the script fails, for debugging in CI, from
```
mernst, line 41: I'm not sure how useful this is in CI.  In CI, you have the commit ID that is being used and so the line number is sufficient.  I don't object to this per se, but every bit of added complexity is an opportunity for problems or unexpected interactions.  I think it would be better to focus effort on making the error messages easier to understand.

kelloggm: The comment about CI was perhaps misleading. I found this useful while debugging locally, and I think it will also be useful when we encounter failures in CI (but I don't have any evidence of that). I changed the comment to just say "useful for debugging" without specifically mentioning CI.

I don't understand this comment:
> In CI, you have the commit ID that is being used and so the line number is sufficient

The code here (or something else that uses `$LINENO`) is necessary to get the line number, so I don't see how the above fact (which is true) is relevant.

> I think it would be better to focus effort on making the error messages easier to understand.

This trap is part of that effort. Before this PR, the script ignored failures. Adding `set -eo pipefail` makes errors stop the script, but doesn't tell us where the error occurred. So, for example, we might get an error like "bad substitution" (this example from the `${*@Q}` issue we discussed briefly this morning), but the error message wouldn't include a line number. I added this trap code specifically so that errors like that also include a line number, so that they're easier to debug.

> every bit of added complexity is an opportunity for problems or unexpected interactions

That's fair, and I'll admit that this is a moderately complex debugging feature relative to the rest of the script. I found it very useful, though, since when a bash script fails the line number isn't printed automatically.

kelloggm:
@mernst, the tests are now succeeding on my Mac (with bash 3.2.57) and on one of the CSE research servers. Can you take another look at this?

In reply to your comment:
> I don't understand the "is purely informational" line that is output

I think you suggested this wording; I've never liked it and personally think it's unnecessary. As I recall, we included it because the script issues output even when it is succeeding, which goes against the usual unix philosophy of "quiet = success, noise = failure". Given the script's purpose and long-running nature, though, I've found it much more useful when it provides periodic updates, so that as a user I can tell that it's making progress.

Another option could be to add "[INFO]: " or some other similar prefix to all of the expected outputs from the script, and delete the "is purely informational" message. The downside of that is that we'd have to remember to include all of the script's messaging statements, including new ones we add in the future. Or, with a bit of added complexity, we could re-define `echo` to include that automatically - but your point about added complexity above suggests to me that that's a bad idea. What do you think?

mernst:
@kelloggm I have always hated the "is purely informational" text.  I don't recall arguing for that wording.  Maybe it would be helpful to express its meaning in plain English.
Before, I recall that the script didn't output any (or maybe just not enough) information about where to find its output.  Since I think it now does so, maybe it no longer needs a disclaimer, that I think was just there to tell users that they needed to look elsewhere to find the actual output.
The script could alternately print a message indicating whether it succeeded or failed, though I'm not sure that is necessary.

checker/bin/wpi-many.sh
mernst: OK, I'm fine with keeping it.  I agree it isn't very much complexity, especially if you have found it helpful while debugging.

> Adding set -eo pipefail makes errors stop the script, but doesn't tell us where the error occurred.

This is the aspect of the scripts that I was proposing should be improved.  That could still be done in addition to the trap.  (And I wasn't sure whether the trap might have interfered with other output or other behavior.)

checker/bin/wpi-many.sh
```diff
@@ -191,15 +200,11 @@ do
...
-        # the `unannotated` remote is not added - it's just a convenience
-        # for data analysis. But, running this script twice in a row on projects
-        # whose owner is the github user always causes an error on this line,
-        # because the `unannotated` remote is already set. The output is piped
-        # to /dev/null to suppress that error.
-        git remote add unannotated "${ORIGIN}" &> /dev/null
+        # The `unannotated` remote is just a convenience for data analysis.
+        # Running this script twice in a row on projects whose owner
+        # is the github user always causes a harmless error on this
+        # line, because the `unannotated` remote is already set.
```
mernst, line 206: Could you use `git ls-remote --exit-code ...` to test this?  That would be shorter than the comment explaining it...

checker/build.gradle
```diff
@@ -647,25 +647,56 @@ task wpiManyTests(group: "Verification") {
...
+                        '-i', "${project.projectDir}/tests/wpi-many/testin.txt",
+                        '-o', "${project.projectDir}/build/wpi-many-tests",
+                        '-s',
+                        '--', '--checker', 'nullness,interning,lock,regex,signature'
+            }
+        } catch (Exception e) {
+            println("Failure: Running wpi-many.sh failed with a non-zero exit code.")
+            File wpiOut = new File("${typecheckFilesDir}/wpi-out")
+            if (wpiOut.exists()) {
+                println("========= Output from last run of wpi.sh: ========")
```
mernst, line 673: I think it would be helpful to also state the file that contains this output (likewise for the similar code below).

mernst (approved):

mernst:

checker/bin/wpi-many.sh
kelloggm: The advantage of the comment is that it is not executable, so it doesn't increase the script's complexity. I don't mind having a long comment in the script (which most users will never read). Using `|| true` here to capture the failure case seems fine to me, since it doesn't matter if this fails.

If you think this is serious issue, I'm happy to change it to a test using `git ls-remote` as you suggested, but I personally prefer it as-is.

checker/bin/wpi-many.sh
kelloggm: > This is the aspect of the scripts that I was proposing should be improved. That could still be done in addition to the trap

I'm not sure what the antecedent of "this" and "that" are in this sentence. This PR both adds `set -eo pipefail` and adds the trap. These two things have different functions:
* `set -eo pipefail` stops the script when an error occurs
* the trap intercepts the error from the above, prints a line number and some other context, and then proceeds to stop the script

mernst: The failure I was seeing came from an invocation of `wpi.sh` by `wpi-many.sh`.  I didn't see any useful diagnostics from `wpi.sh` in that run.  I'm pointing out that it would have been better for `wpi.sh` to report a clear error message.  The logic in `wpi-many.sh` is useful as a stopgap when `wpi.sh` fails in an obscure way.

checker/bin/wpi-many.sh
mernst: It's not a *serious* issue, but it's an issue.

The comment *does* increase the script's complexity:  it makes the script longer and harder to read and understand.

Using `|| true` is a code smell and might, in the future, obscure some other problem.  That is my concern.

What is the data analysis step?  Could this be moved to that script?

checker/bin/wpi-many.sh
kelloggm: Thanks, I understand now.

I think the changes to the `build.gradle` script that print the `wpi-out` file (which contains the stdout/stderr from `wpi.sh`) should make it easier to understand that kind of failure in the future.

checker/bin/wpi-many.sh
kelloggm: > What is the data analysis step? Could this be moved to that script?

This feature was intended to allow the user to 1) run `wpi-many.sh` once on a set of repositories, 2) triage the results and write any necessary annotations in forks of the repositories, and 3) re-run the script using the newly-annotated forks while still being able to compute the correct lines of code and other statistics on the unannotated code.

I'm inclined to remove the feature entirely. It's complicated, and I've only ever used it for one experiment (the `DescribeImagesRequest` experiment in our ICSE 2020 paper on builders) - in more recent experiments, we haven't written any annotations by hand at all, and even if we did I'm not convinced this mechanism would be all that useful.

kelloggm: That said, let's do that as a separate PR, since I'd like to merge this one now so that I can fix do-like-javac's CI.

checker/bin/wpi-many.sh
mernst: Great, thanks.

PR #4663, opened 2021-05-13
Use String instead of Name

mernst:


smillst (approved):
The changes look good to me.  I'm curious why you made these changes?

PR #4665, opened 2021-05-14
remove the -u option from wpi-many.sh

kelloggm:
We didn't test it, it adds complexity, its implementation was buggy, and as far as we're aware no-one is using it.

See also the comment chain here: https://github.com/typetools/checker-framework/pull/4662#discussion_r632745038


mernst (approved):
Thanks for doing this!
The misc job seems to indicate a missing edit.

PR #4675, opened 2021-05-17
Small fixes to wpi-many.sh

kelloggm:
These minor fixes are required for do-like-javac's CI jobs to succeed: https://github.com/kelloggm/do-like-javac/pull/40. I also added a few bits of debugging info, and cleaned up the failure state when no projects succeed.


mernst (approved):
Looks good.  Thanks for the fixes!

PR #4676, opened 2021-05-18
Fix build and future compatibility

twistedsquare:
Three commits -- I can split into three pull requests if you'd prefer as per your guidelines, but they're all one line each:

One commit (missing exec) clears up issues in the build file which I suspect I uncovered because I started a fresh build and the devs never hit that branch.

One commit fixes a duplicate comma; I don't know why Gradle accepts this at the moment?

One commit fixes a heading level issue in Javadoc, which is an error in more recent versions of the Javadoc tool.

Backstory: I have been experimenting with building checker framework under Gradle 7 and JDK 16.  These are the subset of those changes that are currently compatible with building on JDK 8 and JDK 11, but which will be needed if you upgrade Gradle and/or JDK in future.  If you're interested, there were only about 10 more changes needed to get it to at least build and run on JDK 16, albeit without the new language features properly supported.

There are no explicit tests for this pull request because it's about the build working/not working, but they are simple to look over.


mernst (approved):
Thanks very much for these changes!

We would also be interested in the other changes to get the Checker Framework working on JDK 16.  We could put them on a branch even if they are currently incompatible with Java 8 and/or Java 11.  We would need to resolve the incompatibilities before merging the branch.

We do plan for support for LTS versions of Java, including Java 17.  Currently we are working on fixing some issues with capture conversion and generic type inference, and we plan to work on Java 17 as soon as we are done with those tasks.  But we would gratefully accept your assistance in the meanwhile, which could let us work on both tasks simultaneously.

By the way, CI will fail because your name is not in `contributors.tex`.  ~Could you add it?  (Or, I can do so.)~ I added it.

Thanks again!

Issue #2373, opened 2019-03-19
Dataflow support for switch expressions


michaelhixson:
I am trying to use [switch expressions](http://openjdk.java.net/jeps/325), which is a new [preview feature](http://openjdk.java.net/jeps/12) in Java 12.  My application depends on [NullAway](https://github.com/uber/NullAway), which is a compiler plugin that depends on the Checker Framework's dataflow library.  When I try to compile code that includes a switch expression, NullAway propagates an exception thrown by the dataflow library, causing compilation to fail.

The problematic section of the source code being compiled looks like this:
```java
public static void main(String[] args) {
  String s = switch (args.length) {
    case 0 -> null;
    case 1 -> args[0];
    default -> { throw new IllegalArgumentException("invalid args"); }
  };

  System.out.println(s);
}
```

The root cause of the error that is thrown is this, from the stack trace that NullAway prints:
```
[ERROR]   Caused by: java.lang.AssertionError: case visitor is implemented in SwitchBuilder
[ERROR]         at shadow.checkerframework.dataflow.cfg.CFGBuilder$CFGTranslationPhaseOne.visitCase(CFGBuilder.java:3381)
[ERROR]         at shadow.checkerframework.dataflow.cfg.CFGBuilder$CFGTranslationPhaseOne.visitCase(CFGBuilder.java:1416)
[ERROR]         at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCCase.accept(JCTree.java:1293)
[ERROR]         at jdk.compiler/com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:82)
[ERROR]         at jdk.compiler/com.sun.source.util.TreeScanner.scan(TreeScanner.java:106)
[ERROR]         at jdk.compiler/com.sun.source.util.TreeScanner.scanAndReduce(TreeScanner.java:114)
[ERROR]         at jdk.compiler/com.sun.source.util.TreeScanner.visitSwitchExpression(TreeScanner.java:354)
[ERROR]         at jdk.compiler/com.sun.tools.javac.tree.JCTree$JCSwitchExpression.accept(JCTree.java:1325)
[ERROR]         at jdk.compiler/com.sun.source.util.TreePathScanner.scan(TreePathScanner.java:82)
[ERROR]         at shadow.checkerframework.dataflow.cfg.CFGBuilder$CFGTranslationPhaseOne.translateAssignment(CFGBuilder.java:2687)
[...]

PR #4677, opened 2021-05-18
JDK 16 compatibility

twistedsquare:
This is paired with the pull request for annotation-tools https://github.com/typetools/annotation-tools/pull/411 .
Linked to other pull requests (that cannot be merged until after a new release of the Checker Framework is made):
https://github.com/plume-lib/bcel-util/pull/116
https://github.com/plume-lib/bibtex-clean/pull/104
https://github.com/plume-lib/html-pretty-print/pull/103
https://github.com/plume-lib/icalavailable/pull/91
https://github.com/plume-lib/lookup/pull/100
https://github.com/plume-lib/multi-version-control/pull/108
https://github.com/plume-lib/options/pull/119
https://github.com/plume-lib/plume-util/pull/160
https://github.com/plume-lib/reflection-util/pull/105
https://github.com/plume-lib/require-javadoc/pull/93

This gets the framework compiling with JDK 16 and Gradle 7.0.  The first few commits should be fine, although look carefully at the UpperBoundTransfer changes.  The last commit is temporary; it just fills in the visitor for the new Java features with a blank implementation for now, but it at least compiles then.

I had to commit without code formatting checks as google-java-format doesn't have a release with the latest error-prone version that checker-framework just had to upgrade to, so it won't run on JDK 16 as it stands.

Needless to say several of these changes are not backwards compatible with JDK 11, so you'll want to put it in a branch if anything for now.


dataflow/src/main/java/org/checkerframework/dataflow/cfg/node/CaseNode.java
```diff
@@ -2,9 +2,8 @@
 
 import com.sun.source.tree.CaseTree;
 import com.sun.source.tree.Tree.Kind;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Objects;
+import java.util.*;
```
mernst, line 5: Our policy is not to use wildcard imports.  Love it or hate it, could you please revert this change?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/node/CaseNode.java
```diff
@@ -24,30 +23,34 @@
...
 
   public Node getSwitchOperand() {
     return switchExpr;
   }
 
-  public Node getCaseOperand() {
-    return caseExpr;
+  /**
+   * Gets the nodes corresponding to the case expressions. There can be multiple expressions since
+   * Java 12.
```
mernst, line 50: The formatting rules will fail unless you also add a `@return` tag to this Javadoc comment.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/node/CaseNode.java
```diff
@@ -62,7 +65,9 @@ public CaseTree getTree() {
 
   @Override
   public String toString() {
-    return "case " + getCaseOperand() + ":";
+    return "case "
+        + getCaseOperands().stream().map(Object::toString).collect(Collectors.joining(", "))
```
mernst, line 69: I would use [`StringsPlume.join`](https://plumelib.org/plume-util/api/org/plumelib/util/StringsPlume.html#join-java.lang.CharSequence-java.lang.Iterable-) which is more concise and slightly more efficient.

mernst (changes requested):
Thanks for these changes!
I'm sorry for my delay in responding; I was traveling for 2 weeks with limited Internet access.

I prefer not to put these changes on a branch, because there is the possibility of merge conflicts and because the improvements do not benefit everyone.  In fact, I don't think a branch is necessary.

One incompatibility between JDK versions is `CaseTree.getExpression()` vs `CaseTree.getExpressions()`.  Could you please create a method `caseTreeGetExpressions()` in `org.checkerframework.javacutil.TreeUtils` that calls one of the two methods reflectively, depending on which one is available?  That should both compile and run under all JDK versions.  If there any other methods require the same treatment, do that for them as well.

Since this pull request makes a change to the `dataflow` and `javacutil` packages, I'll need your confirmation that you approve of relicensing your contributions under a more lenient open source license such as the MIT License.

Thanks again!

twistedsquare:
I confirm I'm happy with the other licence.  I can fix CaseTree as requested, and a similar change in Resolve, to look up methods dynamically.

The JointJavacJavaParserVisitor class is not as obvious though.  For JDK 16+ it needs implementations of three new methods that use new types not present in JDK 8 and JDK 11, e.g.

```
  @Override
  public Void visitBindingPattern(BindingPatternTree tree, Node node) {
    return null;
  }
```

uses the new BindingPatternTree class, so dropping the Override tag (my first inclination) is not enough.  Do you have a suggestion on how best to handle this?

mernst:
Thanks for pointing out the problem with the visitor.

In places where the Checker Framework defines the data structures, such as in `CFGTranslationPhaseOne.java`, it can define nodes corresponding to the JDK 16 syntax as needed.  That doesn't work for the visitor, though.  And I'll understand if you don't want to add that support in this pull request.

Here are two ideas.

1. Add to the visitor a method `visitJava17Syntax` that takes an `Object` as its argument.  Implementations of this will need to reflectively determine the run-time type and reflectively perform operations on it.

2. Create a subclass of the Visitor that contains methods `visitBindingPattern`, etc.  Compile this subclass using Java 16.  The code does not refer directly to this class.  At run time, test the JDK version; if it is 16, then reflectively instantiate the visitor subclass and use it in preference to an object of the regular Visitor class.  This would need to be done for every visitor subclass.

The first alternative keeps the build system and requirements simple, but the code is ugly.
The second alternative requires Java 16 to be present in order to compiler the Checker Framework, but the code is cleaner.

I lean toward the second alternative.  Can someone improve these suggestions or find a third way that is better?

cc @smillst

mernst (changes requested):

twistedsquare:
I've pushed up everything except the visitor, which is proving fiddly.  It is subclassed two levels further (excuse using an image):

image

Which means if the visitor has two variants (11 and 16), so must each of its two direct subclasses, and then so must each of the three subclasses beneath that.  At first I thought this can be cut down a little bit by using more of a trait-like approach where the three lowest-level subclasses choose whether to pull in default implementations using a trait-like interface that just implements the JDK 16 bits -- but you can't dynamically pull in an interface like this reflectively, only at compile-time.

smillst:
> Here are two ideas.
> 
> 1. Add to the visitor a method `visitJava17Syntax` that takes an `Object` as its argument.  Implementations of this will need to reflectively determine the run-time type and reflectively perform operations on it.
> 2. Create a subclass of the Visitor that contains methods `visitBindingPattern`, etc.  Compile this subclass using Java 16.  The code does not refer directly to this class.  At run time, test the JDK version; if it is 16, then reflectively instantiate the visitor subclass and use it in preference to an object of the regular Visitor class.  This would need to be done for every visitor subclass.
> 

A third alternative is to create a subclass of `com.sun.source.util.SimpleTreeVisitor`.  SimpleTreeVisitor implements all the methods of `TreeVisitor` by calling `defaultAction` and exists all Java versions greater than 5.   We could override defaultAction to check if the tree is one of the three new trees. (This can be done via `tree.getKind().name().equals("YIELD")`). If it is a new tree, use reflection to get the subtrees and call a new method with those sub trees. For example, `visitYield(Tree yieldTree, ExpressionTree expression, P p);`

mernst:
Thanks for the improvements!

You might already know this, but [this test](https://dev.azure.com/checkerframework/checkerframework/_build/results?buildId=6781&view=logs&jobId=4b06b90d-5821-5e7c-238e-1a35bafe92f9&j=4b06b90d-5821-5e7c-238e-1a35bafe92f9&t=9ea81916-7299-5d55-7e8f-ffc1e68b12e3) is failing due to a type error in Gradle code:  comparison of string to integer.

twistedsquare:
Thanks -- I think I've just fixed the Gradle error.   I think with the new PR https://github.com/typetools/annotation-tools/pull/407 it's getting close.  I believe that one is mergeable as it stands, and if maven is updated for the JDK 16 issue (see comments on that PR), we will start to see proper output from the CI for this PR and see if it's working correctly to merge.

twistedsquare:
I know I've made many PRs now to different projects, and the commits here are a bit messy (happy to redo a simpler set if you'd like, now that I've iterated to a reasonable state), but here's the state now.  The JDK 8 and JDK 11 CI items all pass.  The JDK 16 CI has 6 failures in my CI: https://dev.azure.com/twistedsquare/checkerframework/_build/results?buildId=46&view=results that uses the annotation-tools version currently being worked on at https://github.com/typetools/annotation-tools/pull/411 :

plume_lib_jdk16 and inference_tests_jdk16 are both failing for the same reason: bcel-util doesn't support JDK 16.  I believe this will be fixed by PR https://github.com/plume-lib/bcel-util/pull/112 that upgrades the Gradle version.

daikon_jdk16 fails because Daikon doesn't support JDK 16.  I've only glanced at this but I think it's because of the Javadoc package rename out of com.sun into jdk.

guava_jdk16 fails due to an unknown flag "donotexist"; I haven't yet figured out where that comes from.

nonjunit_tests_jdk16 has 3 test failures.  I think they are all "legitimate" in that they are caused by a change in compiler behaviour in JDK 16, rather than missing configuration.

junit_tests_jdk16 has a few test failures.  I think one is again the Javadoc package move, and the others are again legitimate JDK 16 behaviour changes.

I don't know if you want to keep this PR open until all the above are fixed, or disable those CI and merge in before looking at some those "legitimate" failures.

mernst:
Neil, thanks so much for all your changes!  There is no need to apologize for messy commit history -- it will all be clean after we squash-and-merge, and in the meanwhile the diffs show all the changes.

I have disabled the daikon_jdk16 test for now; we can come back to it later.  I would like to get the other tests passing if possible, however.  (And maybe that will be enough to get daikon_jdk16 passing.)

Could you please help me with the plume_lib_jdk16 changes?
For each of these 10 projects:

 * https://github.com/plume-lib/bcel-util
 * https://github.com/plume-lib/bibtex-clean
 * https://github.com/plume-lib/html-pretty-print
 * https://github.com/plume-lib/icalavailable
 * https://github.com/plume-lib/lookup
 * https://github.com/plume-lib/multi-version-control
 * https://github.com/plume-lib/options
 * https://github.com/plume-lib/plume-util
 * https://github.com/plume-lib/reflection-util
 * https://github.com/plume-lib/require-javadoc

please do the following:

1. Make a fork
2. In the fork create a branch named exactly `jdk16`
3. Make the following edit in file `.github/workflows/gradle.yml`:

```diff
     strategy:
       matrix:
-        java: [ '8', '11' ]
+        java: [ '8', '11', '16' ]
 
     steps:
```

4. Open a pull request.
5. Link the pull requests from the beginning of your [overview comment](https://github.com/typetools/checker-framework/pull/4677#issue-647103208) at the beginning of this pull request.

This will help in investigating and fixing problems.

If steps 3 and 5 are too much trouble for you, I can do them.  But I don't have permissions to do the other steps.

Thanks again!

twistedsquare:
Done!  You sure have a lot of repositories.

twistedsquare:
I've just filed PR https://github.com/typetools/guava/pull/51 which fixes the odd "doesnotexist" compiler error seen on the JDK 16 CI, which I think is actually due to a Maven version difference between the CI machines, not the JDK itself.  With that, the new errors for building Guava are about missing opens that are needed for JDK 16 with the compiler checker framework.  I'll have a look at those next and work out if it needs to change in the guava fork or in the checker-framework fork.

mernst:
@twistedsquare Thanks for diagnosing that tricky issue.  I enabled CI for you on that branch.

twistedsquare:
Thanks -- although, I've pushed a few more commits to https://github.com/typetools/guava/pull/51 to fix issues highlighted by the CI but the new run says it requires a maintainer's approval, so I think maybe it was only enabled for that one specific run and not the branch?

mernst:
In typetools/guava#51 I don't see any remark about need for a maintainer's approval.  (I've only ever seen the notification about the first run for a given committer.)  I do see that jobs seem to be running there.  Please let me know if I am missing something!

smillst:
I've been approving the runs in typetools/guava#51.  I was trying to figure out if there was a way to enable them once and for all.  It seems like the only way is to have a pull request merged: (https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)
>After a contributor has at least one pull request merged into a project's repository, any future pull requests from that contributor's fork will automatically run workflows.

twistedsquare:
Ah, thank you, and sorry to have triggered so many!  I was trying to track down the out of memory error in the CI but I haven't yet figured out which of the many JVMs involved is actually running out of memory.  It seems from the stack trace like it is not actually the tests themselves, but rather the process responsible for reading in all the output.

mernst:
Neil, could you please create a pull request against this repository, in a branch *not* named `jdk16`, that contains only the two  `combine.children="append"` lines?

That will be independently useful, even to people not using JDK 16, so it is appropriate in a different PR.  And, merging of that PR will give this PR permission to run CI without maintainer approval.

Thanks.

twistedsquare:
Done: https://github.com/typetools/guava/pull/52  Although I suspect the CI will fail for the same reasons the jdk16 branch is failing: an outdated snapshots link (fixed in the jdk16 branch, so I could pull that commit in too) and possibly the out of memory error as well.

mernst:
Thank you.  You are right about the snapshots link; it does need to be pulled in.

I was hopeful that the out-of-memory problem would not arise with just the `combine.children` issue.  We will see if my optimism was misplaced.  If the out-of-memory problem still arises in typetools/guava#52, then you could make a pull request with some trivial change to file `README-typetools.md`.

Thanks again.

twistedsquare:
Status update - the CI here shows four failures:
 - The guava_jdk16 failure is solved by changes I'm trying to merge in via https://github.com/typetools/guava/pull/52 but I haven't got Guava's own CI to pass there yet, so I'm continuing to investigate.
 - The inference_tests_jdk16 failure is because it is running WPI tests on kelloggm's forks of bcel-util etc, as listed in checker/tests/wpi-many/testin.txt, and those forks don't compile under JDK 16.  The two possibilities are to bring those forks up to date with JDK 16 (I've tried this for bcel-util as a proof of concept) or to take the latest JDK16-compatible versions of the libraries and redo the wpi-many-tests forks from there.  I don't know if you have a preference?
 - The nonjunit_tests_jdk16 failure seems to be a difference in JDK 16 over how many annotations it returns and whether it annotates the method return.  I haven't looked into this yet.
 - The plume_lib_jdk16 failure is because options uses the old com.sun.javadoc package not the new JDK 9+ equivalent.  I've experimented with excluding the problematic OptionsDoclet class from compilation on JDK 16 just to check that it passes otherwise; I imagine you would probably like a reflection-based solution for this issue?

mernst:
Thanks!

 * I fiddled with the Maven 3.8 compatibility pull request over the weekend, but I don't understand Maven (and in particular, how Guava uses it) well enough to find a fix.

 * Regarding the wpi-many tests, I agree with your approach that it is more straightforward to change kelloggm's forks.

 * I have it as an action item to look at the difference in annotation locations across JDK versions.  The annotations appear in different locations in both the compiler AST and the generated bytecode.  I think this is probably a regression (that is, a new bug in javac), but probably a regression that we will have to find a way to work around.  We do report bugs in javac and the JLS on a regular basis, but we aren't paying for support and those bug reports often languish.

 * I am working on upgrading the plume-lib/options package to use the new Javadoc interface.  Then there will be two versions of the options package:  v1 that works on Java 8 and 11, and v2 that works on Java 9 and above.

mernst:
There are some conflicts in this pull request.  It's probably easier to address those now rather than waiting until CI succeeds.  Do you want to do that, or should I?

mernst:
I am also working on a pull request to convert uses of `new Integer` to `Intger.valueOf`, which is delayed because it has revealed the need for a new feature in the Lock Checker.

twistedsquare:
I can address the conflicts here.  Then I'll look at fixing the kelloggm forks, and if that works, I'll issue PRs for those projects.

I'll also make a comment on the Guava PR in a sec - there's something you are likely to know more about than me.

twistedsquare:
Update: getting closer!  The inference_tests now pass on JDK 16 so technically there's only one remaining test failure, although that is misleading as I'll explain.  I think the remaining steps are now as follows:

 - Getting the inference tests to pass involved temporarily pointing wpi-many-tests-bcel-util to my own fork; I've just made https://github.com/kelloggm/wpi-many-tests-bcel-util/pull/1 which will pull my wpi-many-tests-bcel-util changes into kelloggm's repository.  Once that's merged, I'll point back to his repository.
 - The plume test relies on my jdk16 branch of options which disables the Javadoc support, so it will need the proposed change to the options library first (unless you'd like to merge in my disabling of OptionsDoclet on JDK 16 in the mean time).
 - The guava test relies on my Guava jdk16 branch which has an outstanding issue around Nullness annotations in its own CI: https://github.com/typetools/guava/pull/52
 - The nonjunit_tests are failing for reasons given above around annotations appearing on methods that mernst suggestions above is a JDK regression.
 - The annotation-tools is using my jdk16 branch which also has an outstanding issue in its own CI, this time around annotation targets: https://github.com/typetools/annotation-tools/pull/411

I think that's everything, as it stands.  I don't want to duplicate work but @mernst if I can help by looking into any of the remaining issues you suggested you'd look into, I'm happy for you to assign me whichever one I'm most likely to be able to help with.

mernst:
Regarding Guava: The problem is that the `@Nullable` annotation you added was a type annotation only.  Although Android now supports type annotaitons, Guava contains code that does not look for type annotations (in particular the `isNullable` method in the `android/` part of Guava).  That is a bug, but it interacts with other bugs.  For example, changing the `@Target` meta-annotation on the `@Nullable` annotation is not sufficient.  I am now trying just disabling the NullPointerTests under Android, which is the Guava developers' long-term plan.

mernst:
@twistedsquare It looks to me like your branches at https://github.com/twistedsquare/guava/branches all have no diffs and can be deleted.

smillst (approved):

mernst (approved):

mernst:
@twistedsquare Thanks again for your work on JDK 16 compatibility.  We really appreciate your hard work and the comprehensive diffs!  Sorry it took so long to get the PRs merged -- it revealed other issues that were good to fix also.  This opens the way to type-checking of JDK 16 constructs, which will need to happen in a future pull request.

twistedsquare:
It was certainly a lot of work all-in-all!  I think you ended up doing most of it.  I do have some initial work on record checking that I'll issue a PR for shortly -- I was waiting to rebase it on to this work as otherwise things were going to get messy with the merges.

mernst:
@smillst did quite a bit of work too.  All in all, a great team effort, with you at the point.

Support for records would be terrific.  I'm happy to collaborate on that as well.

twistedsquare:
Yes, sorry -- that was a you-plural, meaning the checker-framework team.  I've shifted my records commits across, I'm just getting the CI to pass then I'll submit a PR.  Famous last words but I don't think records are too bad to get basic functionality working because with the way javac compiles them, it's primarily the constructors that differ from standard classes.

msridhar:
@twistedsquare this is awesome work!  FYI, we've had #2373 open for a while on getting Checker Framework support for switch expressions, at least in the dataflow library.  I would be happy to collaborate on that feature, if it's a part of your plan

PR #4679, opened 2021-05-20
add @EnsuresCalledMethodsVarArgs annotation to Called Methods Checker

kelloggm:


checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsTransfer.java
```diff
@@ -130,4 +139,90 @@ public void accumulate(Node node, TransferResult<CFValue, CFStore> result, Strin
...
+    if (annot == null) {
+      return;
+    }
+    String[] ensuredMethodNames =
+        AnnotationUtils.getElementValueArray(
+                annot,
+                ((CalledMethodsAnnotatedTypeFactory) atypeFactory)
+                    .ensuresCalledMethodsVarArgsValueElement,
+                String.class)
+            .toArray(new String[0]);
```
mernst, line 164: Why is converting to an array necessary?  I think this can be eliminated.

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsTransfer.java
```diff
@@ -130,4 +139,90 @@ public void accumulate(Node node, TransferResult<CFValue, CFStore> result, Strin
...
+  /**
+   * Extract the current called-methods type from {@code currentType}, and then add each element of
+   * {@code methodNames} to it, and return the result. This method is similar to GLB, but should be
+   * used when the new methods come from a source other than an {@code CalledMethods} annotation.
+   *
+   * @param currentType the current type in the called-methods hierarchy
+   * @param methodNames the names of the new methods to add to the type
+   * @return the new annotation, to be added to the type
+   */
+  private AnnotationMirror getUpdatedCalledMethodsType(
```
mernst, line 198: This can return null.  Add `@Nullable` and document what returning null means.

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsTransfer.java
```diff
@@ -130,4 +139,90 @@ public void accumulate(Node node, TransferResult<CFValue, CFStore> result, Strin
...
+    }
+
+    if (AnnotationUtils.areSame(type, atypeFactory.bottom)) {
+      return null;
+    }
+
+    CalledMethodsAnnotatedTypeFactory cmAtf = (CalledMethodsAnnotatedTypeFactory) atypeFactory;
+
+    List<String> currentMethods =
+        AnnotationUtils.getElementValueArray(type, cmAtf.calledMethodsValueElement, String.class);
```
mernst, line 221: I would put `calledMethodsValueElement` in a field of this class to avoid computing every time (mostly to clarify the code rather than for the efficiency benefit), but either way is OK.

mernst (approved):

PR #4681, opened 2021-05-20
Update the dataflow manual section about shading.

smillst:


mernst (approved):
I edited the text to put the directive first and the motivation afterward.
Could you double-check it and merge when you are satisfied?

Thanks!

Issue #4699, opened 2021-06-04
EnsuresCalledMethodsIf doesn't permit verification of an EnsuresCalledMethods method


kelloggm:
Consider the following test case, which is a proxy for a use of the `SocketChannel` class in the JDK:
```
import org.checkerframework.checker.calledmethods.qual.EnsuresCalledMethods;
import org.checkerframework.checker.calledmethods.qual.EnsuresCalledMethodsIf;

import java.io.IOException;

class EnsuresCalledMethodsIfTest {

    @EnsuresCalledMethods(value = "#1", methods = "close")
    public static void closeSock(EnsuresCalledMethodsIfTest sock) {
        if (!sock.isOpen()) {
            return;
        }
        try {
            sock.close();
        } catch (Exception e) {
        }
    }

    void close() throws IOException { }

    @SuppressWarnings("calledmethods") // this is a proxy for the isOpen methods in the JDK, to make this test case self-contained
    @EnsuresCalledMethodsIf(expression="this", result=false, methods={"close"})
    boolean isOpen() {
	return true;
    }
}
```

I expect this code to verify with no errors. Instead, I see:
```
EnsuresCalledMethodsIfTest.java:9: error: [contracts.postcondition] postcondition of closeSock is not satisfied.
    public static void closeSock(EnsuresCalledMethodsIfTest sock) {
                       ^
  found   : sock is @CalledMethods
  required: sock is @CalledMethods("close")
1 error
```

msridhar:
@kelloggm did you test this on CF master?  I just realized it might require your change to treat methods as called even if they throw an expection in order to pass.

kelloggm:
@msridhar I did test it on CF master.

mernst:
This behavior looks right to me.  Suppose `sock` is null.  Then `sock.close()` throws a NullPointerException before `sock.close()` has a chance to be called.  The exception is caught and control exits the method without `close()` being called.

If you change `catch (Exception e)` to `catch (IOException e)` then this code verifies.

mernst:
Perhaps we want the Called Methods Checker to treat null pointer exceptions specially.  Or perhaps in general the Checker Framework should refine the type of the receiver to the type of null, along NullPointerException edges in the control flow graph.  However, neither of those things is done now.

msridhar:
[...]

PR #4700, opened 2021-06-04
Test case for issue #4699

mernst:


kelloggm (approved):
Thanks for the explanation, this makes sense. This should close #4699.

PR #4702, opened 2021-06-08
Always use version 1.7 of Google Java Format.

smillst:
The newer versions don't work with Java8.  Alternatively, we can require Java 11 to format and not test the format on 8.


mernst:
> The newer versions don't work with Java8.

Thanks for figuring out the problem.

>  Alternatively, we can require Java 11 to format and not test the format on 8.

I prefer this solution, if it is not too much trouble to implement.  Thanks!

PR #4709, opened 2021-06-09
How to make changes to related repositories

mernst:


docs/developer/developer-manual.html
```diff
@@ -283,6 +283,18 @@ <h3 id="pull-requests-branches">Branches</h3>
   with all pull requests.
 </p>
 
+<p>
```
smillst, line 286: I would reference (or maybe move?) the `Related repository` section here.  It has the same information plus a bit more.  (Its important that a pull request is made for each related repository and that the pull requests mention one another.)

smillst (changes requested):

docs/developer/developer-manual.html
mernst: Thanks.  I looked for that information by searching for "branches" and didn't find it, which surprised me.  I have changed the text, so now it's a cross-reference.

smillst: Thanks for the change!  I searched "branch name".

smillst (approved):

PR #4687, opened 2021-05-24
Resource Leak Checker

kelloggm:
Merge after https://github.com/typetools/jdk/pull/120.

This PR adds the new Resource Leak Checker, along with associated tests and documentation.

This PR includes the changes from several other PRs, which should be merged before this one:
* ~~#4678~~
* ~~#4679~~
* ~~#4682~~
* ~~#4683~~
* ~~#4686~~

Some review of parts of this PR have already happened on draft PRs in my fork:
* kelloggm#243 (documentation; all comments have already been addressed)
* kelloggm#245 (an earlier version of this PR; high-level comments only)

This PR is a port of the `always-call-checker` branch of [kelloggm/object-construction-checker](github.com/kelloggm/object-construction-checker).

PR #4710, opened 2021-06-09
rename CreatesObligation to CreatesMustCallFor

kelloggm:
merge ~~after~~ with typetools/jdk#120

This PR also includes a fix to the documentation of the annotation: one of the paragraphs was just incorrect and said exactly the opposite of the true thing. I assume that a mistake was made when editing at some point and we didn't notice. The target of a `@CreatesMustCallFor` annotation must **not** have a declared CalledMethods type that is non-empty. For example, suppose that some `Socket` has a declared type of `@CalledMethods("close")`. Then even calling `bind()` on that socket (`bind()` has an `@CreatesMustCallFor` annotation in the JDK) would not actually reset the obligation, because the socket would always be considered `@CalledMethods("close")`.

(that PR uses a different branch name, because it's also needed for #4687. I created another branch with the name `createsmustcallfor` so that CI should be able to pass for this PR, too. If CI is failing, you might need to update that branch in my fork of `jdk`)


mernst:
@kelloggm 
> merge after [typetools/jdk#120](https://github.com/typetools/jdk/pull/120)
Why is this "merge after" rather than "merge with"?  If it should be "merge with", then I think the two PRs should be against branches of the same name.

kelloggm:
> Why is this "merge after" rather than "merge with"? If it should be "merge with", then I think the two PRs should be against branches of the same name.

Because typetools/jdk#120 was already open, and is also required for #4687's tests to pass. Creating two duplicate PRs in typetools/jdk seemed like a bad idea to me, so I just created an appropriately-named branch so that CI on this PR will pass.

mernst (approved):

PR #4717, opened 2021-06-11
Show fully-qualified names as `-processor` arguments

mernst:


kelloggm (commented):
There are other places in the manual where a checker shorthand in used. I suggest merging the `longform-checker-names` branch from my fork into this branch before you merge this, which corrects some other instances that I noticed.

mernst:
I only saw two other places.  Your pull request removes mentions of the short form rather than giving both.

smillst (approved):

PR #4721, opened 2021-06-14
Handle return expressions in LiveVarTransfer

ethantkoenig:
Variables which appear in the return expression should be live before
the return node.


mernst:
Thanks for this fix to the example!  I appreciate it.

Do you approve of licensing this contribution under a more liberal license such as the MIT License?

ethantkoenig:
re license: Yes, I approve.

mernstcheckerframework:
Thanks for this fix!

The pull request inspired me to add tests and to make some other changes, which I pulled into this.  The PR history is messy, but the commit is only 3 lines bigger than yours.

(For future reference, you might find it easier to make pull requests from a branch rather than from `master`.)

Thanks again.

PR #4723, opened 2021-06-14
Fix dependencies and `.gitignore` for dataflow tests

mernst:
Running `./gradlew :dataflow:liveVariableTest` from a clean checkout does not work; this pull request fixes that.

### DIFF
--- a/dataflow/build.gradle
+++ b/dataflow/build.gradle
@@ -54,7 +54,7 @@ createDataflowShaded('shaded')
 createDataflowShaded('nullaway')
 createDataflowShaded('errorprone')
 
-task liveVariableTest(dependsOn: ':checker:assemble', group: 'Verification') {
+task liveVariableTest(dependsOn: [':checker:assemble', compileTestJava], group: 'Verification') {
     description 'Test the live variable analysis test for dataflow framework.'
     inputs.file('tests/live-variable/Expected.txt')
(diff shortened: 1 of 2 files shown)


dataflow/build.gradle
```diff
@@ -54,7 +54,7 @@ createDataflowShaded('shaded')
 createDataflowShaded('nullaway')
 createDataflowShaded('errorprone')
 
-task liveVariableTest(dependsOn: compileTestJava, group: 'Verification') {
+task liveVariableTest(dependsOn: ':checker:assemble', group: 'Verification') {
```
smillst, line 57: This task must depend on `compileTestJava`.  You can add multiple dependOn in the body on the task.  But I don't think it needs `checker:assemble`; I don't see where checker.jar is used.  Why did you make this change?

smillst (changes requested):

dataflow/build.gradle
mernst: Thanks, I independently figured that out too.

I made the change because, as noted at the top of this pull request, running `./gradlew :dataflow:liveVariableTest` from a clean checkout does not work.  The failure is ClassNotFoundException for BugInCF.  Maybe some smaller dependence would also work; I did not try to minimize the dependency, but I agree it would be better to minimize it.

smillst (approved):

PR #4727, opened 2021-06-15
Rename wasRaw to isUnderlyingTypeRaw.

smillst:


mernst (approved):
Please review my changes.

How about renaming the method to indicate that the underlying type is raw, or that the programmer wrote a raw type in source code?  That would be better than `wasRaw`.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeMirror.java
```diff
@@ -973,8 +973,9 @@ public void setTypeArguments(List<? extends AnnotatedTypeMirror> ts) {
     }
 
     /**
-     * Returns true if the type was raw, that is, type arguments were not provided but instead
-     * inferred.
+     * Returns true if the underlying type is raw -- that is, the programmer wrote a raw type in
```
smillst, line 976: I don't think `-- that is, the programmer wrote a raw type in source code.` is correct.
Can't raw types be in byte code, too or are raw types always converted to non-raw in byte code?
Also, this AnnotatedTypeMirror could be the type of an expression, in which case its type is not written.  Though I suppose somewhere in the code the programmer wrote a raw type which lead to the expression having a raw type.

What should the new name be? Maybe `isUnderlyingTypeRaw`?

mernst: `isUnderlyingTypeRaw` seems most descriptive.

For an expression type, I think of the programmer-written type as the type of the field being accessed or the method being called.  You are right that byte code can contain raw types, but only if the programmer wrote raw types in source code.  Nonetheless, I see that my wording was misleading, so feel free to change it.  Thanks!

smillst:
@mernst I've re-assigned you in case you wish to review the method renaming.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeMirror.java
```diff
@@ -973,21 +974,44 @@ public void setTypeArguments(List<? extends AnnotatedTypeMirror> ts) {
...
+    }
+
+    /**
+     * Returns true if the underlying type is raw. The receiver of this method is not raw, however;
+     * its annotated type arguments have been inferred.
+     *
+     * @return true iff the type was raw
+     * @deprecated Use {@link #isUnderlyingTypeRaw()} instead
+     */
+    @Deprecated
```
mernst: Please add a date to each `@Deprecated` annotation to help us know when to remove the deprecated methods.  (I did so for this pull request.)

mernst (approved):
Looks great, thanks.  My only changes are about non-executable, non-Javadoc code, so I will merge now.

PR #4738, opened 2021-06-21
Add my papers to the list of papers in the CF manual

kelloggm:


mernst:
This pull request didn't add citations for some checkers, such as the Must Call Checker and the Returns Receiver Checker.  I went ahead and did a bigger reorganization in #4743.

PR #4742, opened 2021-06-23
Changed BaseTypeVisitor to set the language level for parsing...

twistedsquare:
… to match the current source level of the compilation.

This seems wise generally, and will be necessary for Java 16 support to function correctly, as otherwise Java 16 features won't be accepted by the parser side.  This commit should work as-is in older JDKs (CI will check that!) and doesn't depend on the open JDK 16 pull request.

The other uses of the parser in the source code currently use Java 8 only.  This may also need to change (what if you want a stub involving a record?) but I have not changed those in this commit/PR.


smillst:
Thanks for these changes.  Is there any reason not to set the ParserLevel to JAVA_16 for all uses?  Or maybe to JAVA_11 for now and then to JAVA_16 in the jdk16 pull request?

twistedsquare:
There's not a strong reason.  My thinking for user code was that the parser should match the source level they are telling to javac to target in case of any source incompatibilities.  It's a bit of a corner case but in theory if someone had Java 8 code that was using var as a type, if the source level is 8 then the parser should parse it as a type, not as a keyword like Java 11 and 16 would count it.  But that's probably quite obscure so I'm happy to switch to always using Java 16 if you think that will make life simpler.

Although it's not in this PR, for the stubs and annotation processing and expression parsing and so on I could see an argument for moving all parsing up to JAVA_16 as that is more under the source framework's control.  And for the expression parsing in particular, it's either going to need a lot of code to thread though the source language level correctly, or to just bump the language level up to the latest available one, which seems much easier.

smillst:
@twistedsquare that's good point about the key word `var`.  I'm going to think through each parser call to decide which level makes sense.

smillst:
I've set the language level to the current source version for JavaExpressions, but used Java 11 for stub files and ajava files.

framework/src/main/java/org/checkerframework/framework/util/JavaParserUtil.java
```diff
@@ -28,6 +30,12 @@
  */
 public class JavaParserUtil {
 
+  /**
+   * The Language Level to use when parsing if a specific level isn't applied. This should be the
+   * highest version of Java that the Checker Framework can process. Currently, Java 11.
+   */
+  public static LanguageLevel DEFAULT_LANGUAGE_LEVEL = LanguageLevel.JAVA_11;
```
mernst, line 37: Should this also be used in `getCurrentSourceVersion()`?

framework/src/main/java/org/checkerframework/framework/util/JavaParserUtil.java
```diff
@@ -276,4 +308,39 @@ public void visit(BinaryExpr node, Void p) {
...
   }
+
+  /**
+   * Returns the {@link com.github.javaparser.ParserConfiguration.LanguageLevel} corresponding to
+   * the current source version.
+   *
+   * @param env processing environment used to get source version
+   * @return the current source version
+   */
+  public static ParserConfiguration.LanguageLevel getCurrentSourceVersion(
```
mernst, line 319: Can the result of this method be cached?
It is far from our biggest performance problem, but the recomputation does seem needless.

mernst (approved):

PR #4746, opened 2021-06-24
In getFakeOverride, remove code that isn't needed.

smillst:
This method is only called when `receiverType` is nonnull.  So, therefore the method must have a receiver.

### DIFF
--- a/framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileElementTypes.java
+++ b/framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileElementTypes.java
@@ -32,7 +32,6 @@ import javax.lang.model.element.Element;
 import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.TypeElement;
-import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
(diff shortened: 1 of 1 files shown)


framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileElementTypes.java
```diff
@@ -398,15 +397,10 @@ public AnnotatedTypeMirror getAnnotatedTypeMirror(Element e) {
 
     ExecutableElement method = (ExecutableElement) elt;
 
-    TypeMirror methodReceiverType = method.getReceiverType();
```
smillst, line 401: In Java 11 `method.getReceiverType()` returns `null` in some cases when the method does have a receiver.  So, the condition in the if below is false. In Java 16, it returns a type with Kind.NONE, so the condition is true.

This happens in `checker/tests/stubparser-nullness/NoExplicitAnnotations.java `.   
```
    nble = sub1.method1();
```
Where the type of `sub1` is `NoExplicitAnnotationsSub1` and `method1` is defined in `NoExplicitAnnotationsSub1`.   

I think the problem is that `NoExplicitAnnotationsSub1` is defined in the same file as `nble = sub1.method1();`.  So it may be that the element isn't fully initialized.

mernst (approved):
Thanks for fixing this subtle problem!

PR #4747, opened 2021-06-24
Update the manual to account for changes to how the wpi.sh and wpi-many.sh scripts work

kelloggm:
Also, add a little bit of debugging info when the WPI tests fail that I found useful when debugging https://github.com/kelloggm/do-like-javac/pull/25.


docs/manual/inference.tex
```diff
@@ -505,18 +505,18 @@
 
 Each script repeatedly runs the checker with an \<-Ainfer=> command-line option to infer
 types for fields and method signatures.  The output of this step
-is a \<.jaif> (for \<infer-and-annotate.sh>) or stub (for \<wpi.sh>) file that records the inferred types.
+is a \<.jaif> (for \<infer-and-annotate.sh>) or \<.ajava> (for \<wpi.sh>) file that records the inferred types.
 Each script adds the inferred annotation to the next run, so that the checker takes them into
 account (and checks them). \<wpi.sh> does this using the \<-AmergeStubsWithSource> command-line
```
jwaataja, line 510: This mentions `-AmergeStubsWithSource`. It's no longer necessary to specify this with ajava files. They behave similarly to the `-AmergeStubsWithSource` automatically.

jwaataja (commented):
Could you also update references to stub files in this test script?
<https://github.com/kelloggm/checker-framework/blob/9572cd9bfca479f3875b868fa5f67aae673bc581/checker/bin-devel/wpi-plumelib/test-wpi-plumelib.sh#L84>
This could also be done in another PR, but it might be easier to just do it here.

docs/manual/inference.tex
kelloggm: I've removed the reference to it. I also opened a PR to actually remove it from the wpi tool in DLJC (which we should have done during the PR I merged earlier today, but I didn't realize this!): https://github.com/kelloggm/do-like-javac/pull/49.

jwaataja (approved):
Looks good, thanks!

mernst (approved):

PR #4748, opened 2021-06-24
Improve performance of CFGTranslationPhaseOne.visitReturn

msridhar:
The call to `TreePath.getPath()` was showing up as taking measurable time in some NullAway profiles.  This change should avoid it.


msridhar:
@smillst @mernst if this change looks ok and could be merged before the next Checker Framework release, we'd love to be able to update and get the fix before our next NullAway release.  Thanks!

mernst (approved):
Thanks for this fix!  I wonder how many other places we could avoid calling the expensive `Tree.getPath`.

PR #4749, opened 2021-06-25
Infer purity annotations

mernst:
Remove file SKIP-REQUIRE-JAVADOC after merging.

Has a few related refactorings that I hope are not too distracting.  (I can abstract them into a different pull request if you like.)

The `inferPurity` variable is a bit of a hack, but it does work and maybe the Lock Checker's treatment of `isSideEffectFree` is the actual hack.


checker/src/main/java/org/checkerframework/checker/lock/LockVisitor.java
```diff
@@ -79,6 +79,14 @@
 
   public LockVisitor(BaseTypeChecker checker) {
     super(checker);
+    for (String checkerName : atypeFactory.getCheckerNames()) {
+      if (!(checkerName.equals("lock")
+          || checkerName.equals("LockChecker")
+          || checkerName.equals("org.checkerframework.checker.lock.LockChecker"))) {
+        inferPurity = false;
```
smillst, line 86: Why can't the Lock Checker inferPurity annotations?

smillst (approved):
Should there be tests for this?

mernst:
> Should there be tests for this?

Perhaps!  This was missing a CHANGELOG entry, so I added:

> The `-Ainfer` command-line argument now outputs purity annotations even if
> neither `-AsuggestPureMethods` nor `-AcheckPurityAnnotations` is present.

The existing tests of `-Ainfer` are `./gradlew wpiManyTests wpiPlumeLibTests`.  That does fail without the `inferPurity` variable, and passes with it (it passes in this PR, that is).  I agree with you that it would be good to add unit tests for `-Ainfer`, but I suggest not putting that in this pull request.

smillst:
>  I agree with you that it would be good to add unit tests for -Ainfer, but I suggest not putting that in this pull reques

Sounds good.

kelloggm:
> The existing tests of -Ainfer are ./gradlew wpiManyTests wpiPlumeLibTests

There are also five smaller tests (that don't use `wpi.sh` or `wpi-many.sh`, and so are closer to unit tests than the two you mentioned). From `./gradlew tasks`:
```
wholeProgramInferenceNullnessJaifTests - Run tests for whole-program inference using .jaif files
wholeProgramInferenceTestCheckerAjavaTests - Run tests for whole-program inference using ajava files
wholeProgramInferenceTestCheckerJaifTests - Run tests for whole-program inference using .jaif files
wholeProgramInferenceTestCheckerStubTests - Run tests for whole-program inference using stub files
wholeProgramInferenceTests - Run tests for all whole program inference modes.
```

You could probably add a test to the test suite that these tests use (which is in `checker/tests/wpi-testchecker/non-annotated`) that depends on purity annotations being inferred to get close to a unit test for this feature.

PR #4752, opened 2021-06-25
Review of MustCallConsistencyAnalyzer.java 

smillst:
I've started to review MustCallConsistencyAnalyzer.java . I've made some small changes and added some question/comments to the code. @kelloggm Would you please address the comments and then assign this back to me so I can continue to review this code. 

This is seemed to be the easiest way to add review comments, but let me know if you would like to do it some other way.


checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -1495,6 +1529,7 @@ private void incrementNumMustCall(Element elt) {
    *
    * @param type the type of the object that has a must call obligation
    */
+  // TODO: Should this be removed?  It doesn't seem like some thing a user needs.
```
kelloggm, line 1532: We've left it in as an undocumented feature on the theory that if we want to re-run the experiments from the RLC paper it will be useful. I have a slight preference for keeping it, but I'm also willing to remove it if you think maintaining it will be more effort than it's worth.

kelloggm:
@smillst This format is fine with me. I've addressed (and deleted) most of your comments, and I think this is ready for another review. I took your advice to introduce a class for an obligation, which makes the relationship between an obligation and a set of resource aliases clear. (A set of resource aliases is an obligation, iff satisfying the must-call obligations of one of the aliases satisfies them all.)

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
smillst: It's fine to keep it.

smillst (commented):
I added more comments.  Add the Obligation and ResourceAlias classes made the code easier to understand.  Thanks!

smillst:
@kelloggm Sorry I broke the code!  (I was running the wrong tests locally.) I can work on fixing my refactoring if you would like.

kelloggm:
> Sorry I broke the code! (I was running the wrong tests locally.) I can work on fixing my refactoring if you would like. I reverted one of my refactorings

@smillst No worries, thanks for reverting. I can take it from here, I think.

kelloggm:
@smillst, I think I've addressed your comments, so this can be reviewed again.

Given the extent of these changes, I think @msridhar should also have the chance to review this PR, so I'm tagging him here to make him aware of it. Manu, I'd welcome your feedback on these changes anytime.

mernst:
@kelloggm I agree. @smillst and I had discussed that @msridhar should review the code before it is merged.  Doing it now or later is fine with me.

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -726,48 +737,53 @@ private Node removeCasts(Node node) {
    *     temporary variable (via a call to {@link
    *     ResourceLeakAnnotatedTypeFactory#getTempVarForNode(Node)})
    */
-  private void doGenKillForPseudoAssignment(
-      Node node,
-      Set<ImmutableSet<LocalVarWithTree>> obligations,
-      LocalVariableNode lhsVar,
-      Node rhs) {
-    // Replacements to eventually perform in obligations.  We keep this map to avoid a
+  private void updateObligationsForPseduoAssignment(
```
msridhar, line 740:
```suggestion
  private void updateObligationsForPseudoAssignment(
```

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -1664,35 +1712,112 @@ public int hashCode() {
...
-   * formal parameter. We keep the tree for error-reporting purposes (so we can report an error per
-   * assignment to a local, pinpointing the expression whose MustCall may not be satisfied).
+   * An Obligation is a set of resource aliases whose must-call obligations can all be fulfilled by
+   * calling the required method(s) on any of the resource aliases. {@link
+   * MustCallConsistencyAnalyzer} is a dataflow analysis whose dataflow facts are Obligations.
+   */
+  private static class Obligation {
+
+    /** The set of resource aliases that can satisfy this obligation. */
+    public final Set<ResourceAlias> resourceAliases;
```
msridhar, line 1722: I would prefer that `resourceAliases` be an `ImmutableSet<ResourceAlias>` to enforce that `Obligation` is deeply immutable.  If some code were to accidentally mutate a `resourceAliases` set in the future it could be really nasty to debug.

msridhar (changes requested):
This looks amazing!  Thanks so much for the additional review and fixes, @smillst.  I have one substantive comment and one typo comment.  Also, it looks like `checker/tests/resourceleak/TryWithResourcesSimple.class` is still committed.

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
kelloggm: That's a good point. I made the change because there were several places where we constructed a mutable set of aliases and then used a copy operation, which seemed wasteful. But your point about correctness is well-taken, and I'll change it back. Thanks for the feedback!

msridhar (approved):

smillst:
@kelloggm I reviewed your improvements and they all look good.  I've add some more comments to code I hadn't closely reviewed.

kelloggm:
@smillst thanks for the comments on `@CreatesMustCallFor`. You're right that the "target" phrasing wasn't defined or consistently used. I've removed it everywhere now, I think, and addressed the comments you left.

kelloggm (approved):

smillst:
@kelloggm Thanks for all these improvements!

PR #4753, opened 2021-06-25
the absence of -AmergeStubsWithSource should not impact the results of parsing Ajava files, even when parsing them AS stub files

kelloggm:
This problem was discovered when DLJC's CI failed in [this PR](https://github.com/kelloggm/do-like-javac/pull/49), when we tried to remove the `-AmergeStubsWithSource` option from WPI after switching WPI to use `.ajava` files exclusively. This PR must be merged before that one, and only if that one's tests are passing.


jwaataja (approved):
Seems to fix CI for <https://github.com/kelloggm/do-like-javac/pull/49> so I think it's good to go. Thanks!

mernst (approved):

PR #4757, opened 2021-06-26
Don't import inner class Tree.Kind

mernst:
I think this is our style preference, but no one had made the code consistent with it.


smillst (approved):
Yes, this is our preferred style.  Thanks for making it consistent!

PR #4763, opened 2021-06-29
Add more external tools that use the Checker Framework

mernst:


docs/manual/external-checkers.tex
```diff
@@ -50,6 +50,14 @@
 Nullness Checker.
 
 
+\sectionAndLabel{NullAway}{nullaway-checker}
```
smillst, line 53: This seems to be just a duplicate for the above.

docs/manual/external-checkers.tex
```diff
@@ -193,6 +201,17 @@
 Error Prone and the Nullness Checker.
 
 
+\sectionAndLabel{Error Prone}{error-prone-checker}
+
+% Has used the Dataflow Framework since summer 2014.
```
smillst, line 206: Other than this line, this is duplicating the above.

smillst (changes requested):
I think you already merged this branch or a branch that contained these changes.  See comments.

mernst:
> I think you already merged this branch or a branch that contained these changes. See comments.

You are right.  There was a merge conflict somewhere, it seems.

PR #4764, opened 2021-06-29
remove documentation of a removed feature

kelloggm:
should have been removed as part of https://github.com/typetools/checker-framework/pull/4665

### DIFF
--- a/docs/manual/inference.tex
+++ b/docs/manual/inference.tex
@@ -273,14 +273,6 @@ you should manually examine the log files for the projects that appear in the
 that the script was able to successfully run WPI on.  (This does not mean
 that the project type-checks without errors afterward.)
 
-\item (Optional) Fork repositories and make changes (e.g., add annotations or fix bugs).
-Modify the input file for \<wpi-many.sh> to remove the line for the original repository,
-but add a new line that indicates the location of both your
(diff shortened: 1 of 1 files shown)


mernst:
@kelloggm What is the best approach for a user who wants to make edits?

@t-rasmud reports that no `typecheck.out` file is produced when a repository in the `.list` file contains annotations.  Is that a bug or expected behavior?  If a bug, could the two of you hash out a fix?  Thanks!

kelloggm:
@mernst the best approach for a user that wants to make edits is to fork the repository, make edits, and then change the `.list` file to point at their forked copy. We could make that explicit in the documentation here if you think it's necessary.

> no typecheck.out file is produced when a repository in the .list file contains annotations. Is that a bug or expected behavior? If a bug, could the two of you hash out a fix? Thanks!

This seems like a bug: the script shouldn't care whether annotations are already present in the project or not. @t-rasmud, can you give more details? Generally, no `typecheck.out` files are produced when the checker couldn't be run or the project isn't buildable on its own. One possibility that comes to mind is that if annotations are added to a project, you'd need to modify the target build system to depend on `checker-qual.jar` so that the original project continues to compile without the scripts.

PR #4776, opened 2021-07-02
New Lock Checker type `@NewObject`, for the result of a constructor invocation

mernst:
Merge together with: https://github.com/typetools/jdk/pull/122
Once this is merged, deprecations in https://github.com/typetools/checker-framework/pull/4677 can be removed.

This enables (for example) `Integer.valueOf()` to be used in place of `new Integer()`.

~This is ready for review, but not for merging until plume-lib projects have been updated to eliminate checker warnings.~


checker/tests/lock/FlowExpressionsTest.java
```diff
@@ -18,7 +18,6 @@
   public void method() {
     // :: error: (lock.not.held)
     getm().field = new Object();
-    // :: error: (lock.not.held)
```
smillst, line 21: I think this warning should be issued. The type of `m` is `@GuardedBy("<self>")` and the lock `m` isn't held.  I think the problem is that `m` is being refined it `@NewObject` because it's a final field.

smillst (changes requested):
The changes look good, but there's a bug.

mernst:
@smillst Could you take another look at this?  I think nothing substantive has changed since your last review, but I'm not positive.
The diffs are much larger than they ought to be, but I'm not sure why.  You can see the diffs here instead, which don't have the spurious parts: https://github.com/typetools/checker-framework/compare/master...mernst:new-integer-4?expand=1

smillst (approved):

PR #4780, opened 2021-07-05
Add TreeUtils.isVarArgs() method

mernst:


smillst (approved):
I added a new test case and fixed two bugs.  You should review my changes.

mernst:
Thanks for handling the `null` case.  The changes look good.

PR #4783, opened 2021-07-05
Use "parameters" for formals, instead of "arguments"

mernst:


framework/src/main/java/org/checkerframework/framework/util/AnnotatedTypes.java
```diff
@@ -810,7 +810,15 @@ public static AnnotatedTypeMirror leastUpperBound(
...
-  public static List<AnnotatedTypeMirror> expandVarArgsFromTypes(
+  /**
+   * Returns the method parameters for the invoked method, with the same number of formal parameters
+   * as the arguments in the given list.
+   *
+   * @param method the method's type
+   * @param args the types of the arguments at the call site
+   * @return the method parameters, with varargs replaced by instances of its component type
+   */
+  public static List<AnnotatedTypeMirror> expandVarArgsParametersFromTypes(
```
smillst, line 821: This should be in the change log.

smillst (approved):

PR #4787, opened 2021-07-07
Enable inference of @EnsuresCalledMethods

kelloggm:


mernst (approved):
Does this infer `@EnsuresCalledMethods` for `this`, or only for fields?
I suspect it will be important to infer it for `this`, but you would know the answer to that better than I do.

kelloggm:
I believe it currently only handles fields of `this`. I agree that it will be important to extend it to `this` and to method parameters (`#1`, etc.), but that is currently a limitation of the framework. See https://github.com/typetools/checker-framework/blob/cddf46e3f5716648cc13a8cb1bc5ab90edd0c0d9/framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceImplementation.java#L237, which is where I think those changes should be made; they'll be part of future PRs.

PR #4790, opened 2021-07-07
MustCallConsistencyAnalyzer code review

mernst:


checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -78,24 +78,31 @@
...
- * type {@link Obligation}, each representing a set of resource aliases for some value with a
- * non-empty {@code @MustCall} obligation. These obligations can be resolved either via ownership
- * transfer (e.g. by being assigned into an owning field) or via their must-call obligations being
- * contained in their called-methods type when the last reference in a set goes out of scope.
+ * <p>TODO: The use of the term "obligation" is ambiguous throughout this class. Sometimes it means
+ * "an object of type Obligation", and sometimes it means "a must-call obligation". These are
+ * different things, and therefore the naming is confusing. Please rename the Obligation class
+ * (suggestions: ResourceAliasSets, ResourceAliases, AliasSets, Aliases). Then, ensure that,
+ * throughout, the word "obligation" is not used in documentation, nor as a variable name, for
+ * instances of that class.
```
kelloggm, line 99: I strongly object to this suggestion. A previous version of this class (before Suzanne's review) did not use the term "obligation" for this concept, and only referred to it as "a set of resource aliases" (and its type was `Set<ResourceAlias>` - there was not yet a wrapper class). In that version of the code, the documentation was much more confusing, because there wasn't a clear link between the dataflow facts actually being tracked mechanically by the analysis ("sets of resource aliases") and the logical dataflow facts ("must-call obligations").

I also disagree with your claim that "an object of type Obligation" and "a must-call obligation" are different enough for there to be confusion. An `Obligation` object does represent the must-call obligations of a single resource. The confusion comes only because the abstraction (a must-call obligation) doesn't quite match the implementation (a set of resource aliases); that's only the case, though, because we can query the Must Call checker at any time to get the "what methods need to be called" part of a must-call obligation using just the resource alias set. Storing those values alongside the resource alias sets as part of the `Obligation` class might be technically possible and might make the name somewhat more intuitive, but it would complicate the logic (in my opinion) unnecessarily.

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -326,11 +495,15 @@ private boolean isValidInvocation(
...
+          // It seems that this predicate could be made into a method in the `Obligation` class,
+          // which would make the code slightly shorter and -- more importantly -- clearer.
           for (ResourceAlias alias : obligation.resourceAliases) {
             if (expression.equals(alias.reference)) {
               // This satisfies case 2 above. Remove all its aliases, then return below.
               if (toRemove != null) {
-                throw new BugInCF(
+                // TODO: I changed this from BugInCF to TypeSystemError.  Please look over the other
+                // uses of BugInCF and use TypeSystemError instead where appropriate.
+                throw new TypeSystemError(
```
kelloggm, line 506: @mernst I don't think the use of `TypeSystemError` is appropriate. It's documentation says that it should only be thrown by the `framework` subproject, and that it indicates that a type system mis-used the framework.

That said, `BugInCF`'s documentation directly contradicts that. In my experience, we've always used `BugInCF` to indicate an error in a checker that's part of the framework. Maybe you're trying to change that, but if so can you please clarify the documentation on the two errors so that it is clear to type system implementers which to use? Until then, I will continue to use `BugInCF`.

mernst: `TypeSystemError` is relatively new (just over a year old).  We are trying to distinguish between bugs in the framework and bugs in checkers built on the framework, and it is a way to do that.  There may be uses of `BugInCF` that still need to be changed to `TypeSystemError`.  I removed the remark about "only thrown by the `framework` subproject in #4810.  Let me know if it is clearer, or it should be enhanced.

kelloggm:
@mernst I think this is ready for another review.

mernst (commented):
"obligation" is ambiguous.  Change it to "Obligation" or "must-call obligation" throughout.

Please see the TODO comments added in my most recent commit.

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -537,7 +557,7 @@ private boolean isValidCreatesMustCallForExpression(
       Element elt = ((LocalVariable) expression).getElement();
       if (!checker.hasOption(MustCallChecker.NO_LIGHTWEIGHT_OWNERSHIP)
           && typeFactory.getDeclAnnotation(elt, Owning.class) != null) {
-        // The expression is an Owning param.  This satisfies case 1.
+        // The expression is an Owning formal parameter or local variable.  This satisfies case 1.
```
kelloggm, line 560: This comment is now incorrect: `@Owning` annotations cannot be written on local variable declarations (which are always treated as owning), so this case can only be reached for owning formal parameters.

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -260,13 +264,17 @@ private boolean canBeSatisfiedThrough(LocalVariableNode localVariableNode) {
...
+            // TODO: Why use TOP?  I think it would be better to throw an exception, if this is
+            // something that should not happen.
             // That shouldn't happen, but if it does fall back to a safe default (i.e. top).
             mcAnno = mustCallAnnotatedTypeFactory.TOP;
           } else {
+            // TODO: Do we have an answer for the below question?
             // TODO: Why does this happen sometimes?
             if (typeElt.asType().getKind() == TypeKind.VOID) {
+              // TODO: Why does this return immediately rather than using `continue` or setting
+              // `mcAnno` to BOTTOM?
```
kelloggm, line 277: If the one of the Obligation's resource aliases is `VOID`, then that implies that there is no actual must-call obligation, because methods cannot be called on void types. (Recall that there is no guarantee that every Obligation will correspond to an actual resource, instead the analysis guarantees that every real resource will have one or more Obligations. So, it's possible to have spurious Obligations that don't actually need to be fulfilled, and this code can be viewed as removing some of them.)

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -260,13 +264,17 @@ private boolean canBeSatisfiedThrough(LocalVariableNode localVariableNode) {
...
             // typeElt is null if reference.getType() was not a class, interface, annotation type,
-            // or
-            // enum---that is, was not an annotatable type.
+            // or enum -- that is, was not an annotatable type.
+            // TODO: Why use TOP?  I think it would be better to throw an exception, if this is
+            // something that should not happen.
             // That shouldn't happen, but if it does fall back to a safe default (i.e. top).
             mcAnno = mustCallAnnotatedTypeFactory.TOP;
           } else {
+            // TODO: Do we have an answer for the below question?
```
kelloggm, line 273: I'll look into it more deeply, but in general there is no guarantee that every Obligation corresponds to a real resource. This code (and the code above that sets `mcAnno` to top when `typeElt` is null) can be viewed as ways to avoid false positives while staying conservative and preserving the guarantees of the analysis.

kelloggm:
@mernst this should be ready for you again.

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -429,61 +679,70 @@ private void trackInvocationResult(Set<Obligation> obligations, Node node) {
...
-   * because it is annotated with {@link NotOwning}.
+   * <p>Specifically, an invocation result does NOT need to be tracked if any of the following is
+   * true:
    *
-   * <p>This method can also side-effect obligations, if node is a super or this constructor call
-   * with MustCallAlias annotations, by removing that obligation.
+   * <ul>
+   *   <li>The invocation is a call to a {@code this()} or {@code super()} constructor.
+   *   <li>The method's return type is annotated with MustCallAlias and the argument passed in this
+   *       invocation in the corresponding position is an owning field.
```
mernst, line 734: Why does this have to be an owning field as opposed to some other expression that is owning?

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -492,21 +751,32 @@ private boolean shouldTrackInvocationResult(Set<Obligation> obligations, Node no
...
     return true;
   }
 
   /**
    * Returns true if this node represents a method invocation of a must-call-alias method, where the
-   * other must call alias is untrackable: an owning field or a pointer that is guaranteed to be
-   * non-owning, such as {@code "this"} or a non-owning field. Owning fields are handled by the rest
-   * of the checker, not by this algorithm, so they are "untrackable". Non-owning fields and this
-   * nodes are guaranteed to be non-owning, and therefore do not need to be tracked, either.
+   * argument in the must-call-alias position is untrackable: an owning field or a pointer that is
```
mernst, line 776: This gives a good operational definition of "untrackable", but please also state its meaning.  Does it mean that the expression is definitely non-owning, or something else?

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -846,7 +1121,7 @@ private void checkReassignmentToField(AssignmentNode node, Set<Obligation> oblig
     // extend beyond the method's body (and which therefore could not be targeted by an annotation
```
mernst, line 1128: A few lines above, at "static initializers don't need to be considered here": What about instance initializer blocks?

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
kelloggm: Because for other owning expressions, such as local variables or method parameters, the MustCallAlias result is added to the resource alias set in the Obligation for the owning expression.

Owning fields are treated specially because they are handled separately. If the result of a method invocation is MustCallAlias with an owning field, then there is no need for this algorithm to track it, because it is guaranteed to be closed via the owning field (that is the definition of an owning field, after all).

kelloggm: Do you think this logic needs to be laid out in the source code?

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
kelloggm: It means what it says: that the algorithm isn't going to track the given expression. Non-owning expressions are untrackable, but as this says so are e.g. owning fields, because they're handled separately.

It's clear that "untrackable" isn't sufficient to get across the point that there are certain kinds of expressions that this algorithm (i.e. the consistency checker) is not responsible for, and that this method's job is to filter those expressions out. Do you have a better suggestion? I'm concerned that just saying "untrackable" again in this comment in different words will only make the comment more verbose without adding any more content.

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
kelloggm: The checker was unsound. I added a new test in this PR (`InstanceInitializers.java`) and some code here that defaults to issuing an error if an owning field with a non-empty must-call obligation is overwritten anywhere except:
* in a method, where we can use the existing procedure to check that the field has actually had its must-call obligations resolved first
* at the field's declaration (this is only sound-y in that it ignores the possibility that a superclass checks its own type and casts to a subclass in one of its constructors to assign the field, but that involves breaking the type system in other ways too so I think it's an acceptable trade-off)

Thanks for noticing this.

kelloggm:
@mernst I think this can be reviewed again. I've addressed all your comments, and resolved the unsoundness you noticed with respect to instance initialization blocks (should that go in a separate PR?).

PR #4802, opened 2021-07-13
Initial support for records

twistedsquare:
This PR adds basic support for records in the checker framework.

It supports parsing them and handling them, and dealing with annotations on the fields.  It supports the canonical constructor form where the parameters are implicitly declared.  There are three new tests which pass, although I suspect more tests will be needed to be confident that more advanced features work (e.g. generic records, records implementing interfaces, etc).

There is one basic thing that I've struggled to test.  Given the record:

```
public record MyRecord(@Foo String myVar)
{
    public MyRecord
    {
        myVar = "x";
        // Implicitly defined assignment at the end of the body: this.myVar = myVar;
    }
}
```

The assignment in the body is actually to the implicitly-declared constructor parameter myVar, not the field myVar.  As it stands with this PR, the checker framework thinks it's an assignment to the field.  But since the parameter and the field have the same annotations by definition, I found it hard to think of an assignment to the parameter that is invalid, but that is valid to assign to the field.  That is, I can't think of a test which would actually show up the difference between assigning to the parameter and assigning to the field.  Maybe you can think of a checker and annotation (Foo) and assignment value ("x") which will show up the problem.


mernst (changes requested):
I have added a test case that fails.
I think that many or all occurrences of `CLASS` (e.g., `case CLASS:` and `Type.Kind.CLASS` and `ElementKind.CLASS`) in the code should be expanded to handle `RECORD`, presumably via a reflective test.

twistedsquare:
Ok, I've fixed that specific case in my local branch.  For the rest, do you want me to first figure out a failing test for each point where I need to modify Kind.CLASS, or happy for me to just take my best guess at each case and then you'll review?

mernst:
I do think we need a few more test cases.  However, I don't feel it is necessary to write a test case for every line you change.  It's fine to take your best guess at the changes that are necessary, and I'll review.  Thanks!

twistedsquare:
For info, I think I may have found a small bug in the existing checker framework in the process of adding more tests.  Given this class:
```
enum StandardQualEnum {
  DUMMY;
  // :: error: assignment
  public static String s = null;
  // :: error: initialization.static.field.uninitialized
  public static String u;
}
```
That second error isn't being issued, but unless I'm mistaken, it should be.  (For the equivalent class, it is.)  I can see why -- a check for Kind.CLASS that should include Kind.ENUM on InitializationVisitor:288.  Would you like me to file an issue for this bug?

mernst:
> ... That second error isn't being issued, but unless I'm mistaken, it should be. (For the equivalent class, it is.) I can see why -- a check for Kind.CLASS that should include Kind.ENUM on InitializationVisitor:288. Would you like me to file an issue for this bug?

You are right.  Thanks for noticing this bug!  In general, yes, please do file issues.
In this case, I went ahead and fixed it to avoid the delay of waiting for an issue:  https://github.com/typetools/checker-framework/pull/4811 .

twistedsquare:
I've reached a problem with the stubs.  It seems reasonable that a user may want to supply a stub for a record that changes the annotation on a record component.  The JEP for records says that these annotations should be automatically applied to the corresponding field and accessor, so it seems like the same should happen for the stubs.  But that accessor is automatically generated, so it probably won't occur in the stubs, and it might not occur in the code either.  But we somehow need to annotate the return type of this implicitly-existing method.

I tried an approach where I kept track in the stub parser of an extra annotation to apply later to the method return type, but since the method isn't observed in the stubs or code, that didn't have an effect.  I then tried an approach where I faked an annotated method but I ran into problems because there's no corresponding Element in the source code, and various exceptions popped up as the fake didn't hold together.  Do you have any suggestions as to how to handle this case?

mernst:
Thanks for pointing out the problem with stubs.  I presume you have a test case; could you share that, to save a bit of time and to ensure that any fix addresses your problem?  Thanks!

twistedsquare:
I've pushed up the failing test case.  Note that this relies on https://github.com/typetools/stubparser/pull/70 which adds support for top-level record declarations to the stub parser.

mernst:
@twistedsquare Thanks for the test case.  That helped me think through the issues.

> I tried an approach where I kept track in the stub parser of an extra annotation to apply later to the method return type, but since the method isn't observed in the stubs or code, that didn't have an effect.

I would try to do something similar to your first approach, but applying the types at *uses* of the method rather than at the (non-existent) declaration of the method.

["Fake overrides"](https://checkerframework.org/manual/#stub-fake-override) are similar, and an implementation for records could be similar to fake overrides.
 * `AnnotationFileParser` records the annotations on record components (that's the JLS's name for fields), similarly to the existing `fakeOverrides` field.
 * Define a method `AnnotationFileElementTypes.getRecordMethodType` that is similar to the existing `getFakeOverride()` method.
 * `AnnotatedTypeFactory.methodFromUse` currently calls `applyFakeOverrides()`.  It would also need to call `applyRecordMethodTypes()`.

Overall, the code will be simpler than that for fake overrides, which has to look up methods across classes.

I can try to implement this, or you could have a go if you prefer.  Either way is fine with me.

twistedsquare:
@mernst Thanks, I'll have a go at implementing your suggestion.

twistedsquare:
I've now added (admittedly sparse) tests and support for type annotations in record stubs, local records, and generic records.  This involved one tiny PR to stubparser: https://github.com/typetools/stubparser/pull/71

Apart from the Daikon test (which gives warnings about excessive GC during compilation which then trigger the -Werror flag) everything in the CI is passing.

The increasingly obscure areas where I'm aware we may still be deficient is:
 1. I haven't looked at method annotations on compact constructors in code and in stubs (does checker framework deal with method annotations on constructors in stubs?)
 2. data flow in compact constructors.  As mentioned at the start of this PR I think the checker framework is confusing the implicit constructor parameters with fields, but since they share annotations and the parameter is assigned to the field at the end of the constructor, I still haven't found a good test case.
 3. Annotations on record parameters in stubs that only apply to methods/fields/parameters.  The [records JEP](https://openjdk.java.net/jeps/395) has a section on how these should transfer; is that something that the stubs should support?  I'm presuming when not in stubs, javac should handle this for us.  My feeling was maybe stubs should only transfer the type parameters to types?

mernst:
Minor comment: there is some missing Javadoc that is causing CI to fail.

mernst (changes requested):
Overall, this looks great!  Thanks very much for this feature, and for the care you took in doing it.
I have left a few comments in this pull request; could you take care of them?  Thanks.

framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileParser.java
```diff
@@ -850,7 +945,11 @@ private boolean skipNode(NodeWithAccessModifiers<?> node) {
           processField((FieldDeclaration) decl, (VariableElement) elt);
           break;
         case ENUM_CONSTANT:
-          processEnumConstant((EnumConstantDeclaration) decl, (VariableElement) elt);
+          if (decl instanceof FieldDeclaration) {
```
mernst, line 948: Under what circumstances is the kind `ENUM_CONSTANT` but the run-time type is `FieldDeclaration`?

framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileParser.java
```diff
@@ -850,7 +945,11 @@ private boolean skipNode(NodeWithAccessModifiers<?> node) {
...
+            processField((FieldDeclaration) decl, (VariableElement) elt);
+          } else if (decl instanceof EnumConstantDeclaration) {
+            processEnumConstant((EnumConstantDeclaration) decl, (VariableElement) elt);
+          }
```
mernst, line 952: I think it would be best to add a default/fallthrough `else` that throws an exception, rather than silently doing nothing.

framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileElementTypes.java
```diff
@@ -392,6 +394,71 @@ public AnnotatedTypeMirror getAnnotatedTypeMirror(Element e) {
...
+      if (((ExecutableElement) elt).getParameters().isEmpty()) {
+        String recordName = ElementUtils.getQualifiedName(elt.getEnclosingElement());
+        AnnotationFileParser.RecordStub recordComponentType =
+            annotationFileAnnos.records.get(recordName);
+        if (recordComponentType != null) {
+          // If the record component has an annotation in the stub, the component annotation
+          // replaces any from the same hierarchy on the accessor method, unless there is a specific
+          // annotation on the accessor in the stubs file.
+          // We thus only replace the method annotation with the component annotation
+          // if there is no accessor in the stubs file:
```
mernst, line 422: These comments are inconsistent: "there is a specific annotation" vs "there is an accessor is the stubs file".

mernst (changes requested):
Here are just a couple final comments.  Thanks again.

mernst:
CI is now passing.

> The increasingly obscure areas where I'm aware we may still be deficient is:
> 
> 1. I haven't looked at method annotations on compact constructors in code and in stubs (does checker framework deal with method annotations on constructors in stubs?)

Yes, stub file may contain both declaration annotations and type annotations, and both types of annotations are respected/utilized.

> 2. data flow in compact constructors.  As mentioned at the start of this PR I think the checker framework is confusing the implicit constructor parameters with fields, but since they share annotations and the parameter is assigned to the field at the end of the constructor, I still haven't found a good test case.

I don't think any test case is necessary.  (In part because I don't see how to write a test case or how this can go wrong.)  I think it is fine as written.

> 3. Annotations on record parameters in stubs that only apply to methods/fields/parameters.  The [records JEP](https://openjdk.java.net/jeps/395) has a section on how these should transfer; is that something that the stubs should support?  I'm presuming when not in stubs, javac should handle this for us.  My feeling was maybe stubs should only transfer the type parameters to types?

This is something that should be supported, because stub files can contain both declaration annotations and type annotations.

Although there are some small missing features related to declaration annotations, I am OK with merging this pull request without them, leaving them to a later pull request.  The feature is good enough now, and I hope we can merge this before the release on Monday, August 2.

Thanks again!

twistedsquare:
I've addressed all the comments you made, so assuming the CI completes successfully I think it's ready to merge.

mernst (approved):

PR #4808, opened 2021-07-16
Allow for multiple method parameters to be @MustCallAlias in Resource Leak Checker; fixes #4785

### LINKED ISSUE #4785: Allow for multiple method parameters to be `@MustCallAlias` in Resource Leak Checker
(moving from https://github.com/kelloggm/object-construction-checker/issues/409)

Right now, the Resource Leak Checker only supports writing `@MustCallAlias` on one formal parameter of a method, but we should generalize this.  Consider `SequenceInputStream`:

https://docs.oracle.com/javase/7/docs/api/java/io/SequenceInputStream.html

For the Resource Leak Checker, I believe the correct qualifiers for this class are:
* `@MustCall({})` for the class
* `@MustCallAlias` for _both_ formal parameters of the `SequenceInputStream(InputStream s1, InputStream s2)` constructor

I believe the semantics are clear; when that constructor is invoked, the returned `SequenceInputStream` is a resource alias for _both_ `s1` and `s2`, and closing the `SequenceInputStream` will close both `s1` and `s2`.  I do not believe it will be terribly difficult to add support for this pattern to the checker.

Nargeshdb:
Fixes [#4785](https://github.com/typetools/checker-framework/issues/4785).

In this PR, methods and constructors are allowed to have Multiple @MustCallAlias arguments like:

`public @MustCallAlias SequenceInputStream(@MustCallAlias InputStream s1, @MustCallAlias InputStream s2) {...}`

If a call has @MustCallAlias annotation, then for each argument passed in the @MustCallAlias position, it creates a set containing the argument and it's resource aliases and add it to the set of obligations. Consider the following example:

`SequenceInputStream s3 = new SequenceInputStream(s1, s2)`

It creates two sets {s1, s3} and {s2, s3}, and then adds them to the set of obligations.

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
+++ b/checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
@@ -429,8 +429,7 @@ class MustCallConsistencyAnalyzer {
     }
 
     // `mustCallAliases` is a set of the argument(s) passed in the MustCallAlias position(s), if any
-    // exists,
-    // otherwise is an empty List.
+    // exists, otherwise is an empty List.
     List<Node> mustCallAliases = getMustCallAliasArgumentNode(node);
     // If mustCallAlias is an empty List and call returns @This, add the receiver to
     // mustCallAliases.
--- a/checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
+++ b/checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
@@ -521,12 +521,12 @@ class MustCallConsistencyAnalyzer {
    */
   private boolean returnTypeIsMustCallAliasWithUntrackable(MethodInvocationNode node) {
     List<Node> mustCallAliasArguments = getMustCallAliasArgumentNodes(node);
-    return !mustCallAliasArguments.isEmpty()
-        && mustCallAliasArguments.stream()
-            .allMatch(
-                mustCallAliasArgument ->
-                    mustCallAliasArgument instanceof FieldAccessNode
-                        || mustCallAliasArgument instanceof ThisNode);
+    for (Node mustCallAliasArg : mustCallAliasArguments) {
+      if (!(mustCallAliasArg instanceof FieldAccessNode || mustCallAliasArg instanceof ThisNode)) {
+        return false;
+      }
+    }
+    return !mustCallAliasArguments.isEmpty();
   }
 
   /**
--- a/checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
+++ b/checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
@@ -68,7 +68,14 @@ import org.checkerframework.framework.flow.CFStore;
 import org.checkerframework.framework.flow.CFValue;
 import org.checkerframework.framework.util.JavaExpressionParseUtil.JavaExpressionParseException;
 import org.checkerframework.framework.util.StringToJavaExpression;
+import org.checkerframework.javacutil.AnnotationUtils;
+import org.checkerframework.javacutil.BugInCF;
+import org.checkerframework.javacutil.ElementUtils;
+import org.checkerframework.javacutil.Pair;
+import org.checkerframework.javacutil.TreePathUtil;
+import org.checkerframework.javacutil.TreeUtils;
 import org.checkerframework.javacutil.TypeSystemError;
+import org.checkerframework.javacutil.TypesUtils;
 import org.plumelib.util.StringsPlume;
 
 /**
--- a/checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
+++ b/checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
@@ -238,19 +238,21 @@ class MustCallConsistencyAnalyzer {
 
   /**
    * If node is an invocation of a this or super constructor that has a MustCallAlias return type
-   * and a MustCallAlias parameter, check if any variable in the current set of obligations is being
-   * passed to the other constructor. If so, remove it from the obligations.
+   * and some MustCallAlias parameter(s), check if any variable in the current set of obligations is
+   * being passed to the other constructor. If so, remove it from the obligations.
    *
    * @param obligations current obligations
    * @param node a super or this constructor invocation
    */
   private void handleThisOrSuperConstructorMustCallAlias(Set<Obligation> obligations, Node node) {
-    Node mustCallAliasArgument = getMustCallAliasArgumentNode(node);
+    List<Node> mustCallAliasArguments = getMustCallAliasArgumentNode(node);
     // If the MustCallAlias argument is also in the set of obligations, then remove it -- its
     // obligation has been fulfilled by being passed on to the MustCallAlias constructor (because a
     // this/super constructor call can only occur in the body of another constructor).
-    if (mustCallAliasArgument instanceof LocalVariableNode) {
-      removeObligationsContainingVar(obligations, (LocalVariableNode) mustCallAliasArgument);
+    for (Node mustCallAliasArgument : mustCallAliasArguments) {
+      if (mustCallAliasArgument instanceof LocalVariableNode) {
+        removeObligationsContainingVar(obligations, (LocalVariableNode) mustCallAliasArgument);
+      }
     }
   }
 
@@ -406,8 +408,8 @@ class MustCallConsistencyAnalyzer {
   /**
    * Given a node representing a method or constructor call, checks that if the result of the call
    * has a non-empty {@code @MustCall} type, then the result is pseudo-assigned to some location
-   * that can take ownership of the result. Searches for the set of same resources in {@code
-   * obligations} and adds the new resource alias to it if one exists. Otherwise creates a new set.
+   * that can take ownership of the result. Searches for the set(s) of same resources in {@code
+   * obligations} and adds the new resource alias to it if any exists. Otherwise creates a new set.
    *
    * @param obligations the currently-tracked obligations. This is always side-effected: an
    *     obligation is either modified (by removing it from the obligation set and adding a new one)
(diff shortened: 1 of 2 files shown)


checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -986,38 +997,38 @@ private String receiverAsString(FieldAccessNode fieldAccessNode) {
...
    * @return if {@code callNode} invokes a method with a {@code @MustCallAlias} annotation on some
-   *     formal parameter (or the receiver), returns the result of calling {@link
-   *     #removeCastsAndGetTmpVarIfPresent(Node)} on the argument passed in that position.
-   *     Otherwise, returns {@code null}.
+   *     formal parameter(s) (or the receiver), returns the result of calling {@link
+   *     #removeCastsAndGetTmpVarIfPresent(Node)} on the argument(s) passed in corresponding
+   *     position(s). Otherwise, returns an empty List.
    */
-  private @Nullable Node getMustCallAliasArgumentNode(Node callNode) {
+  private List<Node> getMustCallAliasArgumentNode(Node callNode) {
```
msridhar, line 1009: Rename to `getMustCallAliasArgumentNodes`

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -426,39 +428,44 @@ private void trackInvocationResult(Set<Obligation> obligations, Node node) {
       return;
     }
 
-    // `mustCallAlias` is the argument passed in the MustCallAlias position, if any exists,
-    // otherwise null.
-    Node mustCallAlias = getMustCallAliasArgumentNode(node);
-    // If mustCallAlias is null and call returns @This, set mustCallAlias to the receiver.
-    if (mustCallAlias == null
+    // `mustCallAliases` is a set of the argument(s) passed in the MustCallAlias position(s), if any
```
msridhar, line 431: You can just write, `// mustCallAliases is a (possibly-empty) list of arguments passed in a MustCallAlias position`

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -426,39 +428,44 @@ private void trackInvocationResult(Set<Obligation> obligations, Node node) {
...
+    // exists, otherwise is an empty List.
+    List<Node> mustCallAliases = getMustCallAliasArgumentNode(node);
+    // If mustCallAlias is an empty List and call returns @This, add the receiver to
+    // mustCallAliases.
+    if (mustCallAliases.isEmpty()
```
msridhar, line 436: Why do we need this `isEmpty()` check?

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -426,39 +428,44 @@ private void trackInvocationResult(Set<Obligation> obligations, Node node) {
...
-            FluentIterable.from(obligationContainingMustCallAlias.resourceAliases)
-                .append(tmpVarAsResourceAlias)
-                .toSet();
-        obligations.remove(obligationContainingMustCallAlias);
-        obligations.add(new Obligation(newResourceAliasSet));
+    for (Node mustCallAlias : mustCallAliases) {
+      if (mustCallAlias instanceof FieldAccessNode) {
+        // Do not track the call result if the MustCallAlias argument is a field.  Handling of
+        // @Owning fields is a completely separate check, and there is never a need to track an
+        // alias
```
msridhar, line 449: Fix the line breaks in the comments here

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -426,39 +428,44 @@ private void trackInvocationResult(Set<Obligation> obligations, Node node) {
...
+                  .toSet();
+          obligations.remove(obligationContainingMustCallAlias);
+          obligations.add(new Obligation(newResourceAliasSet));
+        }
       }
-    } else {
-      // If mustCallAlias is neither a field nor a local already in the set of obligations,
-      // add it to a new set.
+    }
+    if (mustCallAliases.isEmpty()) {
```
msridhar, line 467: I don't think this logic is equivalent to what was present before.  Before, if `mustCallAlias` was neither a `FieldAccessNode` nor a `LocalVariableNode`, then `tmpVarAsResourceAlias` would be added to a new obligation set.  Now, if `mustCallAliases` is a singleton list where the contained node is neither a `FieldAccessNode` nor a `LocalVariableNode`, nothing happens.  Can we figure out which logic is correct, and add a test case to expose the difference?

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -986,38 +997,38 @@ private String receiverAsString(FieldAccessNode fieldAccessNode) {
   }
 
   /**
-   * Finds the argument passed in the {@code @MustCallAlias} position for a call.
+   * Finds the argument(s) passed in the {@code @MustCallAlias} position(s) for a call.
```
msridhar, line 1000: No need to write "argument(s)" and "position(s)", you can just write "arguments" and "positions."  It should be clear that if there are no such arguments / positions, the method returns an empty list.

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -986,38 +997,38 @@ private String receiverAsString(FieldAccessNode fieldAccessNode) {
...
-   *     formal parameter (or the receiver), returns the result of calling {@link
-   *     #removeCastsAndGetTmpVarIfPresent(Node)} on the argument passed in that position.
-   *     Otherwise, returns {@code null}.
+   *     formal parameter(s) (or the receiver), returns the result of calling {@link
+   *     #removeCastsAndGetTmpVarIfPresent(Node)} on the argument(s) passed in corresponding
+   *     position(s). Otherwise, returns an empty List.
    */
-  private @Nullable Node getMustCallAliasArgumentNode(Node callNode) {
+  private List<Node> getMustCallAliasArgumentNode(Node callNode) {
+    List<Node> result = new ArrayList<>();
```
msridhar, line 1010: Put this after the precondition check

checker/tests/resourceleak/MultipleMethodParamsMustCallAliasTest.java
```diff
@@ -0,0 +1,62 @@
...
+    private final @Owning InputStream in2;
+
+    public @MustCallAlias ReplicaInputStreams(
+        @MustCallAlias InputStream i1, @MustCallAlias InputStream i2) {
+      this.in1 = i1;
+      this.in2 = i2;
+    }
+
+    @Override
+    @EnsuresCalledMethods(
```
msridhar, line 54: This is interesting.  I think this `@EnsuresCalledMethods` annotation gets verified because if the normal exit of `close()` is reached, then both `in1.close()` and `in2.close()` will be invoked.  But, if `in1.close()` throws an `IOException`, then `in2.close()` never gets called and `in2` might leak!  @kelloggm is this a possible soundness hole in the type system?

msridhar (changes requested):
Mostly looks good!  I have a few comments

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
Nargeshdb: In the previous logic `mustCallAlias` could be null(it's equivalent to empty list now) and else branch handled these cases. I can't think of a test case that `mustCallAlias`  is neither a `FieldAccessNode` nor a `LocalVariableNode` after line 437 where we replace it with its tempVar.

checker/tests/resourceleak/MultipleMethodParamsMustCallAliasTest.java
kelloggm: This might be a soundness bug. `@EnsuresCalledMethods` is defined to mean that the given methods are called on the given expressions *if the method terminates successfully* (emphasis mine). That is, an `@EnsuresCalledMethods` annotation doesn't make any guarantees about exceptional exits.

However, we may have relied on a stronger contract when we decided how to handle owning fields. In particular, I think we use `@EnsuresCalledMethods` to mean "the given methods are either called or are the cause of any exceptional exit" - because close() can throw an IOException, we need that last clause to verify even a simple method that attempts to close a single resource and allows any exceptions to propagate. (A different question, in light of this test case, is whether we ought to verify that at all...)

msridhar: We'll fix this separately, #4838

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
msridhar: I think you're right.  I'll add another comment on how to make this more clear.

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -426,39 +428,39 @@ private void trackInvocationResult(Set<Obligation> obligations, Node node) {
...
-                .append(tmpVarAsResourceAlias)
-                .toSet();
-        obligations.remove(obligationContainingMustCallAlias);
-        obligations.add(new Obligation(newResourceAliasSet));
+    for (Node mustCallAlias : mustCallAliases) {
+      if (mustCallAlias instanceof FieldAccessNode) {
+        // Do not track the call result if the MustCallAlias argument is a field.  Handling of
+        // @Owning fields is a completely separate check, and there is never a need to track an
+        // alias of a non-@Owning field, as by definition such a field does not have obligations!
+      } else if (mustCallAlias instanceof LocalVariableNode) {
```
msridhar, line 447: Rather than doing `else if` here, just do an `else`, and throw a `TypeSystemError` at the beginning of the `else` block if `!(mustCallAlias instanceof LocalVariableNode)`.

msridhar (changes requested):
Looks mostly good to me; just one more comment.  Please re-assign to me once it is addressed.

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
Nargeshdb: https://github.com/typetools/checker-framework/pull/4808/commits/7de173ecf89a3877e76f3e2e5502dbd3c07e1e7c

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -426,39 +428,42 @@ private void trackInvocationResult(Set<Obligation> obligations, Node node) {
...
-        obligations.remove(obligationContainingMustCallAlias);
-        obligations.add(new Obligation(newResourceAliasSet));
+    for (Node mustCallAlias : mustCallAliases) {
+      if (mustCallAlias instanceof FieldAccessNode) {
+        // Do not track the call result if the MustCallAlias argument is a field.  Handling of
+        // @Owning fields is a completely separate check, and there is never a need to track an
+        // alias of a non-@Owning field, as by definition such a field does not have obligations!
+      } else {
+        if (!(mustCallAlias instanceof LocalVariableNode)) {
+          throw new BugInCF("unexpected node type for mustCallAlias: " + mustCallAlias.getClass());
```
msridhar, line 449: Make it a `TypeSystemError`?

msridhar (changes requested):

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
Nargeshdb: https://github.com/typetools/checker-framework/pull/4808/commits/0cb9dba497391462af6854928ee425a63f6e58f6

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -68,13 +68,7 @@
...
 import org.checkerframework.framework.util.JavaExpressionParseUtil.JavaExpressionParseException;
 import org.checkerframework.framework.util.StringToJavaExpression;
-import org.checkerframework.javacutil.AnnotationUtils;
-import org.checkerframework.javacutil.BugInCF;
-import org.checkerframework.javacutil.ElementUtils;
-import org.checkerframework.javacutil.Pair;
-import org.checkerframework.javacutil.TreePathUtil;
-import org.checkerframework.javacutil.TreeUtils;
-import org.checkerframework.javacutil.TypesUtils;
+import org.checkerframework.javacutil.*;
```
msridhar, line 71: something weird happened; we shouldn't have this `*` import here.  can you fix?

Nargeshdb: https://github.com/typetools/checker-framework/pull/4808/commits/70644b02e2f28feb3475d3f245840fdeb0f9a11f

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
msridhar: Sorry, one last comment here (that I think I wrote before but somehow got lost).  Can we move this `if` check at line 466 above the previous `for (Node mustCallAlias : mustCallAliases) {` loop, and then put the loop in an `else` block for the `if`?  That will make the logic clearer.

msridhar (changes requested):
Just one more comment

msridhar (approved):
Went ahead and addressed my final comment.  Now looks good to go

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -513,9 +520,13 @@ private boolean shouldTrackInvocationResult(Set<Obligation> obligations, Node no
...
-    Node mustCallAliasArgument = getMustCallAliasArgumentNode(node);
-    return mustCallAliasArgument instanceof FieldAccessNode
-        || mustCallAliasArgument instanceof ThisNode;
+    List<Node> mustCallAliasArguments = getMustCallAliasArgumentNodes(node);
+    return !mustCallAliasArguments.isEmpty()
+        && mustCallAliasArguments.stream()
+            .allMatch(
+                mustCallAliasArgument ->
+                    mustCallAliasArgument instanceof FieldAccessNode
+                        || mustCallAliasArgument instanceof ThisNode);
```
kelloggm, line 529: In the Checker Framework, the preferred style is to use a loop rather than a stream. Can you rewrite this to use a loop rather than a stream?

checker/tests/resourceleak/MultipleMethodParamsMustCallAliasTest.java
```diff
@@ -0,0 +1,62 @@
...
+  void testMultiMethodParamsCorrect(@Owning InputStream in1, @Owning InputStream in2)
+      throws IOException {
+
+    ReplicaInputStreams r = new ReplicaInputStreams(in1, in2);
+
+    r.close();
+  }
+
+  // :: error: required.method.not.called
+  void testMultiMethodParamsWrong1(@Owning InputStream in1, @Owning InputStream in2)
```
kelloggm, line 18: Can you add a variant of this test that calls both `in1.close()` and `in2.close()` on all paths? I think that ought to be verifiable and not issue any errors.

kelloggm (changes requested):
Overall, this looks good to me, modulo a few relatively minor comments.

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
Nargeshdb: https://github.com/typetools/checker-framework/pull/4808/commits/cf5ae953043084ab03b3f20aeec23136e636707b

checker/tests/resourceleak/MultipleMethodParamsMustCallAliasTest.java
Nargeshdb: I've added two more tests to call close on in1 and in2 on all paths.

checker/tests/resourceleak/MultipleMethodParamsMustCallAliasTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+    try {
+      in1.close();
+    } catch (IOException e) {
+    } finally {
+      in2.close();
+    }
+  }
+
+  // Is this a bug in checker framework?
```
msridhar, line 30: I think this has to be an imprecision somewhere.  The code clearly closes `in1` and `in2` on all paths.  We should open an issue to investigate as a follow up

kelloggm: The only difference between this and the previous passing test case is the `catch` block. That definitely shouldn't make a difference, since it's empty! So I suspect there's some imprecision in CFG construction when there is no catch block.

Nargeshdb: I agree there's some imprecision in CFG construction. I remember another [example](https://github.com/Nargeshdb/checker-framework/blob/multipleParamsWithMCA/checker/tests/resourceleak/HBaseReport1.java) that we randomly reported a warning.

Nargeshdb: I opened an issue for this bug https://github.com/typetools/checker-framework/issues/4843

msridhar:
@kelloggm do you have any further comments on this one?

kelloggm (approved):

PR #4809, opened 2021-07-17
Support @EnsuresQualifier inference on "this" and params

kelloggm:
Merge with https://github.com/typetools/annotation-tools/pull/430.

Only postconditions are inferred; preconditions referencing receivers and parameters don't make sense (the receiver/parameter would have its type inferred directly rather than requiring a precondition).

My biggest concern with this PR is the need to change `AnnotationEqualityVisitor`. Without that change, `:checker:ainferTestCheckerValidateJaifs` (i.e. jaif-based WPI) fails because a purity annotation and a `StringVal` annotation are re-ordered in `DefaultsTest.java`. It's possible that that's the result of a bug in the AFU (e.g. in IndexFileWriter), or because of something more serious. (The same failure does not occur in the other Ainfer tests, which pass even without that change.) The change seems okay to me based on the stated purpose of the equality visitor - annotation order in general shouldn't matter - but the need for it makes me nervous.


mernst:
@kelloggm Please write a changelog entry.  The minor version number needs to be incremented because framework methods such as `requiresOrEnsuresQualifierAnno` and `getPostconditionAnnotation` have been removed.

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsAnnotatedTypeFactory.java
```diff
@@ -381,37 +380,38 @@ public Void visitExecutable(AnnotatedTypeMirror.AnnotatedExecutableType t, Void
...
 
   @Override
-  public List<AnnotationMirror> getPostconditionAnnotation(
-      VariableElement elt, AnnotatedTypeMirror fieldAnnos, List<AnnotationMirror> preconds) {
-    AnnotationMirror cmAnno =
-        AnnotationUtils.getAnnotationByName(
-            fieldAnnos.getAnnotations(),
-            "org.checkerframework.checker.calledmethods.qual.CalledMethods");
-    if (cmAnno != null) {
+  protected @Nullable AnnotationMirror requiresOrEnsuresQualifierAnnoForExpression(
```
mernst, line 383: Why is the `ForExpression` part of this name necessary?  There isn't any other `requiresOrEnsuresQualifierAnno` method that needs to be distinguished from this one.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -2568,34 +2540,50 @@ public AnnotatedTypeMirror getDefaultValueAnnotatedType(TypeMirror typeMirror) {
...
+  protected @Nullable AnnotationMirror requiresOrEnsuresQualifierAnnoForExpression(
+      String expression,
+      AnnotationMirror qualifier,
+      AnnotatedTypeMirror declaredType,
+      Analysis.BeforeOrAfter preOrPost,
+      @Nullable List<AnnotationMirror> preconds) {
+
+    // Do not generate RequiresQualifier annotations for "this" or parameter expressions.
+    if (preOrPost == BeforeOrAfter.BEFORE
+        && ("this".equals(expression) || expression.startsWith("#"))) {
```
mernst, line 2573: `expression.startsWith("#")` is too lenient because it also matches, for example, `#1.f`.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -2444,77 +2468,46 @@ public AnnotatedTypeMirror getDefaultValueAnnotatedType(TypeMirror typeMirror) {
...
-   *
-   * <p>This is of the form {@code @RequiresQualifier(expression="this.elt",
-   * qualifier=MyQual.class)} when elt is declared as {@code @A} or {@code @Poly*} and f contains
-   * {@code @B} which is a sub-qualifier of {@code @A}.
-   *
-   * @param elt element for a field, which is declared in the same class as the method
-   * @param inferredType the type of the field, on method entry
-   * @return precondition annotations for the element (possibly an empty list)
-   */
-  public List<AnnotationMirror> getPreconditionAnnotation(
```
mernst, line 2468: This pull request removes the convenience methods `getPreconditionAnnotation` and `getPostconditionAnnotation`.  What is the reason for that?  I think it will make this pull request much easier to understand and review if that change is made in a separate pull request.

mernst (changes requested):

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
kelloggm: There wasn't any reason to remove them, and I'm not really sure why I did, looking back. I've restored them.

kelloggm:
@mernst I think I've addressed your comments, including adding a changelog entry. I think it's ready for another review.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -2468,36 +2486,42 @@ public AnnotatedTypeMirror getDefaultValueAnnotatedType(TypeMirror typeMirror) {
...
+   * restriction).
    *
    * <p>This is of the form {@code @RequiresQualifier(expression="this.elt",
    * qualifier=MyQual.class)} when elt is declared as {@code @A} or {@code @Poly*} and f contains
    * {@code @B} which is a sub-qualifier of {@code @A}.
    *
-   * @param elt element for a field, which is declared in the same class as the method
-   * @param inferredType the type of the field, on method entry
+   * @param expression an expression
+   * @param inferredType the type of the expression, on method exit
```
mernst: I think this should be on method entry

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -2591,40 +2603,60 @@ public AnnotatedTypeMirror getDefaultValueAnnotatedType(TypeMirror typeMirror) {
...
+   *
+   * @param expression the expression to which the qualifier applies
    * @param qualifier the qualifier that must be present
+   * @param declaredType the declared type of the expression, which is used to avoid inferring
+   *     redundant pre- or postcondition annotations
    * @param preOrPost whether to return a precondition or postcondition annotation
-   * @param preconds the precondition annotations for the method; used to suppress redundant
-   *     postconditions; non-null exactly when {@code preOrPost} is {@code AFTER}
+   * @param preconds the list of precondition annotations; used to suppress redundant - *
+   *     postconditions; always null if {@code preOrPost} is {@code BeforeOrAfter.BEFORE}
```
mernst, line 2635: Why was this changed from "non-null exactly when" to "always null if"?  The latter is a weaker condition.  Can the list be null when `preOrPost` is `AFTER`?

Another occurrencef of "non-null exactly when" in the code was *not* changed like this one was.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -2468,36 +2486,42 @@ public AnnotatedTypeMirror getDefaultValueAnnotatedType(TypeMirror typeMirror) {
...
     }
     Collections.sort(result, Ordering.usingToString());
     return result;
   }
 
   /**
-   * Return a list of precondition annotations for the given field. Returns an empty list if none
-   * can be created, because the qualifier has elements/arguments, which {@code @RequiresQualifier}
-   * does not support.
+   * Returns an {@code @RequiresQualifier} annotation for the given expression. Returns an empty
```
mernst, line 2500: This comment says "returns an" but the `@return` clause is plural and the return type is List.

mernst (changes requested):

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceImplementation.java
```diff
@@ -254,15 +262,84 @@ public void updateContracts(
...
-        inferredType = atypeFactory.getAnnotatedType(fieldElement);
+        // This field is not in the store. Use the declared type.
+        inferredType = fieldDeclType;
       }
       T preOrPostConditionAnnos =
-          storage.getPreOrPostconditionsForField(preOrPost, methodElt, fieldElement, atypeFactory);
+          storage.getPreOrPostconditions(
+              preOrPost, methodElt, fa.toString(), fieldDeclType, atypeFactory);
       String file = storage.getFileForElement(methodElt);
       updateAnnotationSet(
```
mernst, line 272: `updateAnnotationSet` is called on the output of `getPreOrPostconditions`.

1. Why does that type-check?  `updateAnnotationSet`'s first type parameter is a `T`, but `getPreOrPostconditions` returns a `@Nullable T`.
2. I think `updateAnnotationSet`, and anywhere else that the result of `updateAnnotationSet` may flow, should have an early exit if the argument is null.

kelloggm: > 1. Why does that type-check? 

I don't think it should. Is the Nullness Checker running on this code? I don't know if we run the Nullness Checker on all parts of the CF's codebase.

> 2. I think updateAnnotationSet, and anywhere else that the result of updateAnnotationSet may flow, should have an early exit if the argument is null

It seems to me that it would make more sense here to do a null-check on the result of `getPreOrPostConditions` and then skip the call to `updateAnnotationSet` if that result was null, because in that case there is nothing to update.

kelloggm: On that second point, the code that I added for parameters and the receiver does in fact do a null check before calling `updateAnnotationSet`, so it should be safe. It's only the code for fields that needs to be changed. I added the null-check at this point, but if you feel strongly that we should change the spec of `updateAnnotationSet` instead I'm fine with doing that, instead.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -2468,77 +2486,90 @@ public AnnotatedTypeMirror getDefaultValueAnnotatedType(TypeMirror typeMirror) {
...
    * subclasses may do so.
    *
-   * @param elt element for a field
-   * @param inferredType the type of the field, on method entry or exit (depending on the value of
-   *     {@code preOrPost})
+   * @param expression an expression whose type annotations to return
+   * @param inferredType the type of the expression, on method entry or exit (depending on the value
+   *     of {@code preOrPost})
+   * @param declaredType the declared type of the expression, which is used to determine whether the
+   *     inferredType is redundant
```
mernst, line 2563: I think you mean that the inferred type supplies no additional information beyond the declared type.  It might be better to say that.  I think "redundant" is not the right word; maybe "vacuous"?  Or just use the definition rather than a word.

framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java
```diff
@@ -2591,40 +2601,62 @@ public AnnotatedTypeMirror getDefaultValueAnnotatedType(TypeMirror typeMirror) {
...
+   */
+  protected static final Pattern formalParameterPattern = Pattern.compile("^#[0-9]+$");
+
+  /**
+   * Creates a {@code RequiresQualifier("...")} or {@code EnsuresQualifier("...")} annotation for
+   * the given expression.
    *
    * <p>This is of the form {@code @RequiresQualifier(expression="this.elt",
    * qualifier=MyQual.class)} or {@code @EnsuresQualifier(expression="this.elt",
    * qualifier=MyQual.class)} when elt is declared as {@code @A} or {@code @Poly*} and f contains
```
mernst, line 2614: What is `f`?  (The pull request edits other text like this, so it should probably edit this one too.)

mernst (approved):

PR #4811, opened 2021-07-17
Check enum fields just like class fields

mernst:


checker/src/main/java/org/checkerframework/checker/initialization/InitializationVisitor.java
```diff
@@ -285,7 +285,7 @@ public void processClassTree(ClassTree node) {
     super.processClassTree(node);
 
     // Warn about uninitialized static fields.
-    if (node.getKind() == Tree.Kind.CLASS) {
+    if (node.getKind() == Tree.Kind.CLASS || node.getKind() == Tree.Kind.ENUM) {
```
smillst, line 288: Should this be `TreeUtils.isClassTree(...)` instead?

smillst (approved):

PR #4814, opened 2021-07-20
better handling of JAVA_HOME in wpi.sh and wpi-many.sh

kelloggm:
This PR cleans up the handling of the `JAVA_HOME` environment variable in `wpi.sh` and `wpi-many.sh`, and fixes some associated bugs:
* `wpi-many.sh` no longer changes `JAVA_HOME` at all - only `wpi.sh` is responsible for setting and unsetting the variable, which simplifies the script
* the "re-run if the project can't build under JDK X" logic has been fixed. Previously, it would have tried to run under Java 8 twice, and never under Java 11. I also added comments to make the logic more clear.
* if `JAVA_HOME` is unset when `wpi.sh` starts, it will be unset when `wpi.sh` finishes instead of being set to an empty string

I considered writing a test for the "re-run if the project can't build under JDK X" functionality to prevent another regression. I don't think I could add such a test to `wpiManyTests`, because our CI runs with only one JDK available in `JAVA_HOME` at a time, I think. @mernst, can you think of a way to test this functionality given our CI setup, which you know much better than I do?

### DIFF
--- a/checker/bin/wpi.sh
+++ b/checker/bin/wpi.sh
@@ -305,9 +305,10 @@ if [ "${WPI_RESULTS_AVAILABLE}" != "yes" ] && [ "${has_java11}" = "yes" ]; then
   fi
 fi
 
-# If results still aren't available and Java 16 is available, then Java 16 has not yet
-# been tried if either Java 8 or Java 11 are available, because those have higher
-# priority. If that's the case, then re-try using Java 16.
+# If results still aren't available, then re-run with Java 16 if it is available
+# and the first run used Java 8 or Java 11 (since Java 16 has the lowest priority,
+# the first run using Java 8 or Java 11 is equivalent to either of these being
+# available).
 if [ "${WPI_RESULTS_AVAILABLE}" != "yes" ] && [ "${has_java16}" = "yes" ]; then
   if [ "${has_java11}" = "yes" ] || [ "${has_java8}" = "yes" ]; then
     export JAVA_HOME="${JAVA16_HOME}"
--- a/checker/bin/wpi.sh
+++ b/checker/bin/wpi.sh
@@ -282,6 +282,8 @@ rm -f -- "${DIR}/.cannot-run-wpi"
 cd "${DIR}" || exit 5
 
 JAVA_HOME_BACKUP="${JAVA_HOME}"
+# For the first run, use the Java versions in ascending order: 8 if its available, then
+# 11, then 16.
 if [ "${has_java8}" = "yes" ]; then
   export JAVA_HOME="${JAVA8_HOME}"
 elif [ "${has_java11}" = "yes" ]; then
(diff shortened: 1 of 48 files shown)


checker/bin/wpi.sh
```diff
@@ -291,18 +293,17 @@ elif [ "${has_java16}" = "yes" ]; then
...
+    configure_and_exec_dljc "$@"
 fi
 
-if [ "${has_java11}" = "yes" ] && [ "${has_java8}" = "yes" ] && [ "${WPI_RESULTS_AVAILABLE}" != "yes" ]; then
-    # If running under Java 11 and Java 8 fails, try Java 16.
-    if [ "${has_java16}" = "yes" ]; then
+# If running under Java 8 and/or 11 failed and Java 16 is available, then try Java 16.
+if [ "${has_java16}" = "yes" ] && [ "${WPI_RESULTS_AVAILABLE}" != "yes" ]; then
+  # If only Java 16 was available, then the first run would have used Java 16, so there's no reason to run again.
+  if [ "${has_java11}" = "yes" ] || [ "${has_java8}" = "yes" ]; then
```
mernst, line 306: This block uses two `if` statements, but the logic for Java 11 uses one.  I think the code will be easier to read and understand if both blocks use logic of the same shape.

mernst (approved):
Thanks for this improvement.

checker/bin/wpi.sh
```diff
@@ -291,22 +293,26 @@ elif [ "${has_java16}" = "yes" ]; then
 fi
 configure_and_exec_dljc "$@"
 
+# If running under Java 8 failed and Java 11 is also available, then try running again under Java 11.
```
mernst, line 296: The logic is right, but I had to read it several times to convince myself of that.
This sentence here lays out the logic, but the `if` statements take a different approach than explained in the sentence.
How about making the `if` statements closer to the English, such as grouping
`"${has_java8}" = "yes" && "${WPI_RESULTS_AVAILABLE}" != "yes"`
together since it expresses "running under Java 8 failed"?

mernst (approved):

checker/bin/wpi.sh
kelloggm: I both re-wrote the comments and re-ordered the tests, so that the two match up exactly. Can you take another look?

checker/bin/wpi.sh
```diff
@@ -291,22 +294,26 @@ elif [ "${has_java16}" = "yes" ]; then
...
 fi
 
-if [ "${has_java11}" = "yes" ] && [ "${has_java8}" = "yes" ] && [ "${WPI_RESULTS_AVAILABLE}" != "yes" ]; then
-    # If running under Java 11 and Java 8 fails, try Java 16.
-    if [ "${has_java16}" = "yes" ]; then
-      export JAVA_HOME="${JAVA16_HOME}"
-      echo "couldn't build using Java 11 or Java 8; trying Java 16"
-      configure_and_exec_dljc "$@"
-    fi
+# If results still aren't available and Java 16 is available, then Java 16 has not yet
```
mernst, line 308: The text of this comment block and the one above has gotten out of sync.

mernst (approved):

checker/bin/wpi.sh
kelloggm: They did! Sorry about that. I've fixed them both to be more like the first one.

kelloggm:
@mernst the only test still failing on this is the daikon job you mentioned was broken on another PR. Does that mean this PR can be merged?

Issue #3387, opened 2020-06-17
Line numbers for delomboked code


mernst:
Code that Lombok processes is not necessarily legal Java code.  (Lombok does various transformations and writes out legal Java bytecodes.)  To run the Checker Framework on code that Lombok processes, it is necessary to [run the Checker Framework on the output of the `delombok` tool](https://checkerframework.org/manual/#lombok).  Then, the Checker Framework error messages refer not to the user-supplied code, but to temporary directories with modified versions of what the programmer wrote.  It is a hassle for programmers to determine where to fix the problems in their code.  This hassle is a deal-breaker for some programmers, making them not want to use the Checker Framework.

A solution to this problem would be useful for all static analysis tools, not just the Checker Framework.

A possible solution is:
 * Create a mapping of line numbers from the delomboked code to the developer's source code.
 * Write a tool that post-processes static analysis output:  it changes errors/warnings that refer to the delomboked code into errors/warnings that refer to the programmer-written code.  Perhaps the postprocessed output would still contain the original warning in the delomboked source code, so the programmer can use it if necessary.

Here are two approaches to create the mapping:
 * Modify the delombok tool.  It already has methods `endPos()` and `lineEndPos()` that might have some of the needed information.  Currently they seem to be used primarily to make sure that comments are copied to the delomboked file.
 * Do textual processing of the original source code and/or the debug information in the .class file (which has correct line numbers), plus of the delomboked source code.

PawelLipski:
@mernst wrt. to _To run the Checker Framework on code that Lombok processes, it is necessary to run the Checker Framework on the output of the delombok tool._... actually, I observe in my project (http://github.com/VirtusLab/git-machete-intellij-plugin/) that it's NOT necessary to run `delombok` tool for CF to see the output of Lombok annotation processor, including getters/setters, c'tors etc... or am I missing something?

PawelLipski:
Please take a look at https://github.com/VirtusLab/git-machete-intellij-plugin/blob/develop/build.gradle#L157, this is basically the differences that I observed b/w running CF on delomboked code vs just on the compilation trees produced by Lombok's annotation processor.

I'm not sure though how it's even possible that CF's annotation processor always seems to run after Lombok's annotation processor in our project... is it guaranteed somehow, or is it just by chance in or project (but possibly not in others)?

PawelLipski:
[...]

PR #4826, opened 2021-07-23
Test that demonstrates the unsoundness that can occur if Lombok and the Checker Framework are run together

kelloggm:
This PR is meant to illustrate a point for the discussion on https://github.com/typetools/checker-framework/issues/3387#issuecomment-883283755. I think we might want to merge this test (with the expected error commented out) so that we continue to keep track of the fact that Lombok and the CF don't always play nicely (which necessitates the use of delombok), as well as the why of it.

For now, I expect CI to fail on this PR.


kelloggm:
There has been no further activity on #3387 since I made this PR, so I've modified it to test that the current behavior (unsoundness when the CF and Lombok are run together) is preserved, with notes about how it would be better if the error were issued. That way, if someone is ever interested in trying to figure out why this unsoundness occurs, we'll have test cases to start from.

mernst (approved):

mernst:
Thanks for this reproduction.

PR #4830, opened 2021-07-23
Store both the initializer value and the declared value of fields. 

smillst:


framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -930,7 +930,7 @@ public void typeProcess(TypeElement e, TreePath p) {
     }
 
     if (!warnedAboutGarbageCollection) {
-      String gcUsageMessage = SystemPlume.gcUsageMessage(.30, 60);
+      String gcUsageMessage = SystemPlume.gcUsageMessage(.20, 60);
```
smillst, line 933: This should probably be reverted.

checker/src/main/java/org/checkerframework/checker/initialization/InitializationVisitor.java
```diff
@@ -273,8 +273,10 @@ public void processClassTree(ClassTree node) {
         Store store = atypeFactory.getRegularExitStore(block);
 
         // Add field values for fields with an initializer.
-        for (Pair<VariableElement, Value> t : store.getAnalysis().getFieldValues()) {
-          store.addInitializedField(t.first);
+        for (FieldValues<Value> fieldValues : store.getAnalysis().getFieldValues()) {
+          if (fieldValues.initializer != null) {
+            store.addInitializedField(fieldValues.field.getField());
```
mernst, line 278: This construct `field.getField()` is a bit wierd.  If it's already a field, why is `getField()` needed?
Hovering over the `field` field access indicates its value, but renaming the field from `field` to `fieldAccess` would eliminate the need to do so.
However, if you strongly prefer the name `field`, then the code can stand as is.

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractAnalysis.java
```diff
@@ -60,8 +60,38 @@
   /** A checker that contains command-line arguments and other information. */
   protected final SourceChecker checker;
 
+  /**
+   * A triple of field, value corresponding to the annotations on its declared type, value of its
+   * initializer. The value of the initializer is {@code null} if the field does not have one.
+   *
+   * @param <V> type of value
+   */
+  public static class FieldValues<V extends CFAbstractValue<V>> {
```
mernst, line 69: I think this class represents only initial values, in which case I would rename it to indicate that, such as to `FieldIinitialValue`.
Since it's for only one field, I would make the name singular, to clarify some use sites (see comments elsewhere).

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractTransfer.java
```diff
@@ -407,71 +406,47 @@ public S initialStore(UnderlyingAST underlyingAST, @Nullable List<LocalVariableN
...
+   * <p>The initializer value is inserted into {@code info} if the field is private and final.
+   *
+   * <p>If {@code methodTree} is a constructor, then the declared value is inserted into {@code
+   * info} if the field has an initializer. (Use the declaration type rather than the initializer
+   * because an initialization block might have re-set it.) If {@code methodTree} is not a
+   * constructor, then the declared value is inserted into {@code info} if the receiver of the
+   * method is fully initialized as determined by {@link
+   * #isNotFullyInitializedReceiver(MethodTree)}.
+   *
+   * @param info initial store into which field values are inserted
```
mernst, line 421: Is this store empty on entry to the method?  Or might some work already have been done before this method is called?

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractTransfer.java
```diff
@@ -407,71 +406,47 @@ public S initialStore(UnderlyingAST underlyingAST, @Nullable List<LocalVariableN
...
+   * @param classTree the class that contains {@code methodTree}
+   * @param methodTree the method or constructor tree
+   */
+  private void addFieldValues(S info, ClassTree classTree, MethodTree methodTree) {
```
mernst, line 425: This only adds initial values (I think), so I would reflect that in the method name.

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractTransfer.java
```diff
@@ -407,71 +406,47 @@ public S initialStore(UnderlyingAST underlyingAST, @Nullable List<LocalVariableN
...
+    boolean isConstructor = TreeUtils.isConstructor(methodTree);
+    List<FieldValues<V>> fields = analysis.getFieldValues();
+    TypeElement classEle = TreeUtils.elementFromDeclaration(classTree);
+    for (FieldValues<V> fieldValues : fields) {
```
mernst, line 429: I think it would be clearer to name the first variable `fieldValues` and the last variable `fieldValue`.  That is a motivation for changing the class name `FieldValues` to a singular one.

framework/src/main/java/org/checkerframework/framework/flow/CFAbstractTransfer.java
```diff
@@ -407,71 +406,47 @@ public S initialStore(UnderlyingAST underlyingAST, @Nullable List<LocalVariableN
...
-              found = true;
-              break;
-            }
-          }
-          if (!found) {
-            // no initializer found, cannot use static type
-            continue;
-          }
+      // Maybe insert the declared type:
+      // If it's not a constructor, use the declared type if the rreceiver of the method is fully
```
mernst, line 440: Minor: These comments are in a different order ("not a constructor" then "is a constructor") than the comments in the method Javadoc.  It's easier to verify that they match if they are in the same order.  If this one is reordered, the `if` test should probably be restructured so it still matches the comment.

mernst (changes requested):

mernst (approved):

mernst:
Thanks for this performance improvement!

PR #4831, opened 2021-07-23
Make @RequiresNonNull repeatable

kelloggm:
This isn't strictly necessary (since `@RequiresNonNull` can already take multiple arguments), but it makes the annotation more consistent with `@RequiresQualifier`, which does permit both multiple arguments and multiple annotations per declaration.

This PR will be useful if we start using `-Ainfer` with the Nullness Checker, because right now it emits each precondition as its own annotation (which works with `@RequiresQualifier`, but would not work with `@RequiresNonNull` without this PR).

I considered adding a special test case for `-Ainfer`'s interaction with this PR, but decided that was probably not necessary. If you think it is, I can write one (or you can).


mernst (approved):
Thanks!

PR #4836, opened 2021-07-26
Update manual's discussion of conflict with NullAway

msridhar:
The conflict no longer exists as of NullAway 0.9.2, which was released today.


mernst (approved):
Thanks!  And thank you for releasing the NullAway version that doesn't have a conflict with the Checker Framework.

PR #4844, opened 2021-07-28
Validate return types correctly.

smillst:


framework/src/main/java/org/checkerframework/common/basetype/BaseTypeValidator.java
```diff
@@ -471,6 +472,14 @@ protected void visitTypeParameterBounds(
       case TYPE_PARAMETER:
```
mernst, line 472: The Javadoc for the containing method twice says "type parameter tree" and once says "parameter type tree".  Please make them consistent.

mernst (approved):

PR #4854, opened 2021-08-02
Use the same `--add-exports`  and `--add-opens` everywhere.

smillst:
Also, don't use `--illegal-access` option.


docs/manual/external-tools.tex
```diff
@@ -601,16 +601,19 @@
 \subsubsectionAndLabel{Non-modularized code}{javac-jdk11-non-modularized}
 
 To compile non-modularized code:
-
```
mernst, line 604: Minor: I would leave the blank line to make the comment more prominent.

build.gradle
```diff
@@ -121,6 +121,27 @@ allprojects {
         allProjects subprojects
     }
 
+  ext {
+    // A list of add-export and add-open arguments to be used when running the Checker Framework.
+    // Keep this list in sync with the lists in CheckerMain#getExecArguments and
+    // the section with label "javac-jdk11-non-modularized" in the manual.
```
mernst, line 127: What about the Maven and SBT sections of the manual?

mernst (approved):

build.gradle
mernst: Also, what about the Checker Fraemwork Gradle plugin?

PR #4857, opened 2021-08-03
Issue4855 attempt to refactor CreateAnnotationForStringConcatenation()

smonteiro:
It has been a while since I did anything in github. Coding was easy, stumbling on github was not.

### DIFF
--- a/framework/src/main/java/org/checkerframework/common/value/ValueTransfer.java
+++ b/framework/src/main/java/org/checkerframework/common/value/ValueTransfer.java
@@ -660,8 +660,8 @@ public class ValueTransfer extends CFTransfer {
     List<String> rightValues = getStringValues(rightOperand, p);
     if (leftValues != null && rightValues != null) {
       // Both operands have known string values, compute set of results
-      leftValues = appendNull(nonNullStringConcat, leftOperand, leftValues);
-      rightValues = appendNull(nonNullStringConcat, rightOperand, rightValues);
+      leftValues = appendNullToStrings(nonNullStringConcat, leftOperand, leftValues);
+      rightValues = appendNullToStrings(nonNullStringConcat, rightOperand, rightValues);
       List<String> concatValues = new ArrayList<>(leftValues.size() * rightValues.size());
       for (String left : leftValues) {
         for (String right : rightValues) {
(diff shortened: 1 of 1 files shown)


framework/src/main/java/org/checkerframework/common/value/ValueTransfer.java
```diff
@@ -1668,4 +1619,63 @@ protected void processConditionalPostconditions(
         resultValues,
         transferResult.getResultValue().getUnderlyingType());
   }
+
+  /** convenience function used by createAnnotationForStringConcatenation(). */
```
mernst, line 1623: Could you please document exactly what the method does, beyond just saying that it is a convenience function?
That will enable readers of the code to understand the method without reverse-engineering it from its implementation.
That will also make continuous integration pass (it is currently failing).

mernst (changes requested):
Thank you for this refactoring! Could you please add documentation, to permit it to be reviewed?  Thanks.

framework/src/main/java/org/checkerframework/common/value/ValueTransfer.java
smonteiro: Hi Michael

I think not and I intend to present a valid argument for that.

I am well aware of checker-framework/docs/developer/developer-manual.html 
**Code style** 
"Every class, method, and field (even private ones) must have a descriptive Javadoc comment."

Java does not have function-local functions (defined in the preamble of an encompassing function) which is essentially what these newly refactor-extracted functions are.

IMHO, writing a detailed Javadoc for a private function extracted from what was a copy-paste-edit function body **paragraph** (that is key) which has no relevance elsewhere and is not meant to be re-used elsewhere sends the wrong message.

We disagree on that point.

The source function createAnnotationForStringConcatenation(Node, Node, TransferInput) itself only has a high level brief comment as its Javadoc
`/** Creates an annotation for a result of string concatenation. */`

^ No @ elements, no details, just high level... and somehow does not get warnings.

So,  in the absence of function-local functions and keeping with the spirit of the minimalist comment of createAnnotationForStringConcatenation() itself,  I arrived at private functions in the same file, using  a brief comment about where it is used.
(sure, it could be better with "used by `{@link #createAnnotationForStringConcatenation(Node, Node, TransferInput)`}")

I tried writing the Javadoc for the new private functions and I gave up as it bloated the final source file with more lines than the original, didn't add value, and it turned out a lot of noise. If one is looking at the private function body there is no shorter English description to '?:' and if statements.

The appendNull*() functions definitely do not have use anywhere else.

The getString*() functions follow the signature of the existing underlying getString*()  functions.

There are other source files in the whole effort that spew warnings about lack of Javadoc elements but they are flagged with the skip flag.

P.S. My original development box from previous contributions was destroyed and I am doing source editing in notepad without even a compiler, have pity.

PR #4863, opened 2021-08-04
Fix handling of exception edges in backward analysis

### LINKED ISSUE #4861: Incorrect logic in `BackwardAnalysisImpl` for exceptional successors
I was looking at the following code:

https://github.com/typetools/checker-framework/blob/b256527a82816115c3df6c120f0f3b0a1f7238a9/dataflow/src/main/java/org/checkerframework/dataflow/analysis/BackwardAnalysisImpl.java#L256-L263

It seems to be trying to check if `block` is an exceptional successor of `pred`.  But the check is wrong, as `((ExceptionBlock) pred).getSuccessor()` will get the non-exceptional successor (`getExceptionalSuccessors()` should be used instead).  Reading the surrounding code, I think the impact of this bug is that in many cases, stores will not get propagated at all along exceptional control-flow edges in the backwards analysis.  Not sure if there are tests for this.

msridhar:
Fixes #4861

### DIFF
--- a/dataflow/tests/live-variable/Expected.txt
+++ b/dataflow/tests/live-variable/Expected.txt
@@ -4,7 +4,12 @@
 4 -> 10 ELSE_TO_BOTH
 8 -> 11 EACH_TO_EACH
 10 -> 11 EACH_TO_EACH
-11 -> 0 EACH_TO_EACH
+11 -> 12 EACH_TO_EACH
+12 -> 13 EACH_TO_EACH
+12 -> 15 ArithmeticException
+13 -> 17 EACH_TO_EACH
+15 -> 17 EACH_TO_EACH
+17 -> 0 EACH_TO_EACH
 
 2:
 Process order: 1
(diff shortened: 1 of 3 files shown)


dataflow/tests/live-variable/Expected.txt
```diff
@@ -66,23 +71,75 @@ b   [ LocalVariable ]
...
+0   [ IntegerLiteral ]
+b = 0   [ Assignment ]
+marker (start of try statement #0)   [ Marker ]
+marker (start of try block #0)   [ Marker ]
+f   [ LocalVariable ]
+1   [ IntegerLiteral ]
+a   [ LocalVariable ]
+~~~~~~~~~
+TransferInput#21
+After:   live variables = a, b
```
msridhar, line 93: Before the logic change in `BackwardAnalysisImpl`, the live variable set here would incorrectly be missing `b`

kelloggm:
This looks reasonable to me, but I'm not the right person to review it, nor am I sure who is. @mernst, can you either review this change yourself or assign someone else who you think is an appropriate reviewer?

kelloggm (approved):
LGTM

mernst (approved):
Thanks!

PR #4867, opened 2021-08-05
Ignore control flow due to NullPointerExceptions in Called Methods Checker

### LINKED ISSUE #4843: False positive in resource leak checker when there is no catch block before finally block
The following example shows a false positive in resource leak checker. The code closes `in1` and `in2` on all paths in both tests but the checker reports a warning for `test2`. The only difference between this two tests is an empty catch block that shouldn't make any difference.

```java
import java.io.*;
import java.net.*;
import org.checkerframework.common.returnsreceiver.qual.*;

class FalsePositiveInRLC {

  void test1(@Owning InputStream in1, @Owning InputStream in2)
      throws IOException {
    try {
      in1.close();
    } catch (IOException e) {
    } finally {
      in2.close();
    }
  }

  // :: error: required.method.not.called
  void test2(@Owning InputStream in1, @Owning InputStream in2)
      throws IOException {
    try {
      in1.close();
    } finally {
      in2.close();
    }
  }
}
```

msridhar:
Fixes #4843

The Called Methods Checker assumes that the checked code is free of NullPointerExceptions (this property should be verified by the Nullness Checker).  This PR enhances the Called Methods Checker's type inference to ignore control flow due to NullPointerExceptions, enabling it to verify more code.  See https://github.com/typetools/checker-framework/issues/4843#issuecomment-889263475 for more context.

The change is implemented by enhancing the dataflow framework to allow for ignoring exceptional control flow for particular exception types.

We may want to ignore more exception types in the future; we can add that in a follow-up PR.

### DIFF
--- a/checker/tests/calledmethods/EnsuresCalledMethodsIfTest.java
+++ b/checker/tests/calledmethods/EnsuresCalledMethodsIfTest.java
@@ -7,14 +7,11 @@ import org.checkerframework.checker.calledmethods.qual.EnsuresCalledMethodsIf;
 class EnsuresCalledMethodsIfTest {
 
   @EnsuresCalledMethods(value = "#1", methods = "close")
-  // The contract is not satisfied.  Suppose `sock` is null.  Then `sock.close()` throws a
-  // NullPointerException before `sock.close()` has a chance to be called.  The exception is caught
-  // and control exits the method without `close()` being called.
-  // :: error: (contracts.postcondition)
+  // If `sock` is null, `sock.close()` will not be called, and the method will exit normally, as the
+  // NullPointerException is caught.  But, the Called Methods Checker
+  // assumes the program is free of NullPointerExceptions, delegating verification of that
+  // property to the Nullness Checker.  So, the postcondition is verified.
   public static void closeSock(EnsuresCalledMethodsIfTest sock) throws Exception {
-    if (!sock.isOpen()) {
-      return;
-    }
     try {
       sock.close();
     } catch (Exception e) {
--- a/checker/tests/calledmethods/EnsuresCalledMethodsIfTest.java
+++ b/checker/tests/calledmethods/EnsuresCalledMethodsIfTest.java
@@ -12,6 +12,9 @@ class EnsuresCalledMethodsIfTest {
   // assumes the program is free of NullPointerExceptions, delegating verification of that
   // property to the Nullness Checker.  So, the postcondition is verified.
   public static void closeSock(EnsuresCalledMethodsIfTest sock) throws Exception {
+    if (!sock.isOpen()) {
+      return;
+    }
     try {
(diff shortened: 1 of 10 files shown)


checker/tests/calledmethods/EnsuresCalledMethodsIfTest.java
```diff
@@ -7,14 +7,11 @@
...
-  // and control exits the method without `close()` being called.
-  // :: error: (contracts.postcondition)
+  // If `sock` is null, `sock.close()` will not be called, and the method will exit normally, as the
+  // NullPointerException is caught.  But, the Called Methods Checker
+  // assumes the program is free of NullPointerExceptions, delegating verification of that
+  // property to the Nullness Checker.  So, the postcondition is verified.
   public static void closeSock(EnsuresCalledMethodsIfTest sock) throws Exception {
-    if (!sock.isOpen()) {
-      return;
-    }
```
kelloggm, line 17: Why remove this part of the test? I think part of this test's goal was originally to make sure that the `@EnsuresCalledMethodsIf` annotation on `isOpen` works correctly.

kelloggm (commented):
Thanks for this work!

I'm not sure whether we should override the new method in the Called Methods Checker or in the Resource Leak Checker (which is a Called Methods Checker subclass). I would be inclined to override it in the RLC rather than in the CMC as you've done here, since I don't see a pressing need for the CMC to make this assumption (but I think making the assumption in the CMC is sound, so it shouldn't make much of a difference).

I pulled this branch into the branch I've been using to try to resolve the RLC soundness bug related to exceptional exits from destructors, and this dramatically improves the code there. There are a few remaining new FPs that I can't currently explain, but given the error messages look non-sensical I suspect there is a bug in my new code. I can confirm that this change gets rid of most of the new FPs of the kind we discussed before.

Someone else with more knowledge of the dataflow framework should review this PR before it is merged. Maybe @mernst or @smillst?

checker/tests/calledmethods/EnsuresCalledMethodsIfTest.java
msridhar: You're right, I shouldn't have removed these lines; I got confused.  Will fix

msridhar:
> I'm not sure whether we should override the new method in the Called Methods Checker or in the Resource Leak Checker (which is a Called Methods Checker subclass). I would be inclined to override it in the RLC rather than in the CMC as you've done here, since I don't see a pressing need for the CMC to make this assumption (but I think making the assumption in the CMC is sound, so it shouldn't make much of a difference).

Fair enough.  The main motivation for this change is the RLC, so we can make the change there if desired, to narrow the scope.  I'm ok either way, so let's leave it for now and get an opinion from another reviewer.

mernst (approved):

PR #4869, opened 2021-08-06
Fix unsoundess in Resource Leak Checker related to owning fields and EnsuresCalledMethods

### LINKED ISSUE #4838: Resource Leak Checker's rules for owning fields rely on a stronger contract for @EnsuresCalledMethods than the actual contract

kelloggm:
Fixes #4838.

Merge after #4867.

This is a replacement for #4839, based on the comments on it from @msridhar. It relies on #4867 to resolve the false positives described in the comment chain on #4839.


msridhar:
This looks pretty clean!  The change looks good to me.  Do we have a test where the finalizer does a `try-finally` and we are able to still verify it?

kelloggm:
> Do we have a test where the finalizer does a try-finally and we are able to still verify it?

We did not, but I just added one as `ReplicaInputStreams2.java`.

checker/src/main/java/org/checkerframework/checker/resourceleak/ResourceLeakVisitor.java
```diff
@@ -80,6 +86,66 @@ public Void visitMethod(MethodTree node, Void p) {
...
+            .getAnnotatedType(containingClass)
+            .getAnnotationInHierarchy(mustCallAnnotatedTypeFactory.TOP);
+    List<String> mcValues =
+        AnnotationUtils.getElementValueArray(
+            mcAnno, mustCallAnnotatedTypeFactory.getMustCallValueElement(), String.class);
+    String methodName = elt.getSimpleName().toString();
+    if (!mcValues.contains(methodName)) {
+      // Not a destructor, just a method with an ECM annotation. No further checking to do.
+      return;
+    }
```
msridhar, line 120: Can we extract this code to a new method that returns a boolean?  Something like `isDestructorForEnclosingClass`?

msridhar (changes requested):
Mostly LGTM; just one style comment

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsAnalysis.java
```diff
@@ -20,21 +22,31 @@
   protected CalledMethodsAnalysis(
       BaseTypeChecker checker, CalledMethodsAnnotatedTypeFactory factory) {
     super(checker, factory);
+    // Use the Nullness Checker.
+    ignoredExceptionTypes.add("java.lang.NullPointerException");
+    // Ignore run-time errors, which cannot be predicted statically. Doing so is unsound
+    // in the sense that they could still occur - e.g., the program could run out of memory -
+    // but if they did, the checker's results would be useless anyway.
+    ignoredExceptionTypes.add("java.lang.Error");
+    ignoredExceptionTypes.add("java.lang.RuntimeException");
```
msridhar, line 31: Why are we adding these exception types in this PR?  Do the tests not pass without it?

msridhar (commented):
Sorry, one more question I forgot to ask before

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsAnalysis.java
kelloggm: The tests do not pass without it.

In particular, we get these three false positives:
```
    3 unexpected diagnostics were found
    CheckFields.java:68: error: (destructor.exceptional.postcondition)
    CreatesMustCallForOverride2.java:38: error: (destructor.exceptional.postcondition)
    CreatesMustCallForOverride2.java:56: error: (destructor.exceptional.postcondition)
```

For reference, here's the relevant lines of `CheckFields`:
```
    @EnsuresCalledMethods(
        value = {"this.finalOwningFoo", "this.owningFoo"},
        methods = {"a"})
    void b() {
      this.finalOwningFoo.a();
      this.finalOwningFoo.c();
      this.owningFoo.a();
    }
  }
```

msridhar: After this lands we may want to think about another manual update like #4892

msridhar (approved):

checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsAnalysis.java
kelloggm: I'm not sure that's really necessary - that run-time errors don't invalidate CalledMethods types follows from the idea that annotations (and therefore types) specify *normal behavior*, and a run-time exception is definitely not normal behavior.

PR #4870, opened 2021-08-06
Improve the error message for required.method.not.called...

kelloggm:
...by adding the simple name of the variable that did not have its obligations fulfilled. Also improve its formatting to better match other CF error messages.


msridhar (approved):

msridhar:
@kelloggm not sure if this has dependencies on other PRs so I'll let you merge

PR #4892, opened 2021-08-16
Notes on what the Called Methods Checker is actually checking, in the manual

kelloggm:
also removed a duplicate line that slipped into the manual somehow

### DIFF
--- a/docs/manual/called-methods-checker.tex
+++ b/docs/manual/called-methods-checker.tex
@@ -7,11 +7,9 @@ of the form ``call method A before method B''. For the purpose of this
 checker, a method has ``definitely been called'' if it is \emph{invoked}:
 a method that might never return or that might throw an exception has
 definitely been called on every path after the call, including exceptional paths.
-The checker also assumes that the program is free of null-pointer dereferences:
-for example, a method that is called on an object that is \<null> at run time
-is considered to have been invoked.
-(You can verify that all pointer dereferences are safe
-by running the Nullness Checker (Section~\ref{nullness-checker}).)
+The checker also assumes that the program is free of null-pointer dereferences.
+You can verify that all pointer dereferences are safe
+by running the Nullness Checker (Section~\ref{nullness-checker}).
 
 The Called Methods Checker provides built-in support for one such property:
 that clients of the builder pattern for object
--- a/docs/manual/called-methods-checker.tex
+++ b/docs/manual/called-methods-checker.tex
@@ -6,8 +6,13 @@ been called on an object. This checker is useful for checking any property
 of the form ``call method A before method B''. For the purpose of this
 checker, a method has ``definitely been called'' if it is \emph{invoked}:
 a method that might never return or that might throw an exception has
-a method that might never return or that might throw an exception has
 definitely been called on every path after the call, including exceptional paths.
+The checker also ignores the possibility that an exception might occur as part
+of the call (even if the called method does not throw the exception):
+that is, it assumes that the receiver of every method invocation
+is non-null at each call-site, and that run-time errors (such as \<OutOfMemoryError>
+and the like) do not occur. (You can verify that every receiver is non-null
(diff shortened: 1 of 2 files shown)


docs/manual/called-methods-checker.tex
```diff
@@ -6,8 +6,13 @@
 of the form ``call method A before method B''. For the purpose of this
 checker, a method has ``definitely been called'' if it is \emph{invoked}:
 a method that might never return or that might throw an exception has
-a method that might never return or that might throw an exception has
 definitely been called on every path after the call, including exceptional paths.
+The checker also ignores the possibility that an exception might occur as part
+of the call (even if the called method does not throw the exception):
+that is, it assumes that the receiver of every method invocation
+is non-null at each call-site, and that run-time errors (such as \<OutOfMemoryError>
```
msridhar, line 13: The assumption is stronger than this.  The checker now assumes that _all_ dereferences (field accesses or method calls) will succeed without throwing an NPE.  Can you clarify?

msridhar (changes requested):

docs/manual/called-methods-checker.tex
kelloggm: You're right. I think I've made the wording more clear; can you take another look?

docs/manual/called-methods-checker.tex
```diff
@@ -6,8 +6,12 @@
 of the form ``call method A before method B''. For the purpose of this
 checker, a method has ``definitely been called'' if it is \emph{invoked}:
 a method that might never return or that might throw an exception has
-a method that might never return or that might throw an exception has
 definitely been called on every path after the call, including exceptional paths.
+The checker also ignores the possibility that a run-time exception might occur:
```
msridhar, line 10: As of right now, the assumptions do not extend to all built-in run-time exceptions.  E.g., `ClassCastException` and `ArrayIndexOutOfBoundsException` are not completely ignored.  Are you thinking we will ignore such exceptions in a future change?  If not, maybe we should be specific here about which exact exception types are ignored.

kelloggm: https://github.com/typetools/checker-framework/pull/4869 will add `Error` and `RuntimeException`, which was what I was thinking about when I wrote this. Maybe this text shouldn't mention that, though, and instead only focus on nullness.

kelloggm: I rewrote the text again to only mention nullness. I think this version is a bit better, but I'm still not that happy with it.

docs/manual/called-methods-checker.tex
```diff
@@ -6,8 +6,12 @@
 of the form ``call method A before method B''. For the purpose of this
 checker, a method has ``definitely been called'' if it is \emph{invoked}:
 a method that might never return or that might throw an exception has
-a method that might never return or that might throw an exception has
 definitely been called on every path after the call, including exceptional paths.
+The checker also assumes that the program is free of null-pointer dereferences:
+for example, a method that is called on an object that is \<null> at run time
+is considered to have been invoked.
+(You can verify that all pointer dereferences are safe
+by running the Nullness Checker (Section~\ref{nullness-checker}).)
```
msridhar, line 14:
```suggestion
The checker also assumes that the program is free of null-pointer dereferences.
You can verify that all pointer dereferences are safe
by running the Nullness Checker (Section~\ref{nullness-checker}).
```

How about just this?  Maybe we want @mernst's feedback here?

kelloggm: That's definitely shorter and sweeter. I'm okay with this version.

I agree getting @mernst's feedback on this would be useful.

kelloggm:
@msridhar I'm assuming you're okay with this version of the change (since you proposed it) and assigning it to @mernst. If you want to think about it some more, please feel free to take this PR back.

msridhar (approved):

mernst (approved):

PR #4894, opened 2021-08-17
Make RLC's ignored exception types consistent

kelloggm:


msridhar (commented):
LGTM, but do we want to take the opportunity to update the manual for the RLC in terms of what kinds of exceptional behaviors are ignored?

kelloggm:
That's a good idea. The current manual section doesn't mention it at all, but I've added a new section at the end of the manual chapter that's dedicated entirely to exceptions. It is not comprehensive, but I think it touches on the most important points.

msridhar (approved):
LGTM

kelloggm:
@mernst, can you take a look at this PR, especially at the new manual text? If you think it looks good, you can go ahead and merge it.

mernst (approved):

Issue #1630, opened 2017-11-03
type inference limitation with streams


cushon:
If the explicit type argument to `map` is removed from the following example it is rejected by the nullness checker. It would be nice if the checker's type inference handled this case without the explicit type.

```java
import java.util.Collections;
import java.util.List;
import java.util.Objects;
import java.util.stream.Collectors;
import org.checkerframework.checker.nullness.qual.Nullable;

class T {
  @Nullable
  static String toString(Object o) {
    return null;
  }

  public static List<String> f(List<Integer> xs) {
    return xs != null
        ? xs.stream()
            .<@Nullable String>map(T::toString)
            .filter(Objects::nonNull)
            .collect(Collectors.toList())
        : Collections.emptyList();
  }
}
```

If `<@Nullable String>` is removed:

```
T.java:16: error: [methodref.return.invalid] Incompatible return type
            .map(T::toString)
                 ^
  Method
    @Initialized @Nullable String toString(@Initialized @NonNull Object p0) in T
  is not a valid method reference for
    @Initialized @NonNull String apply(@Initialized @NonNull Function<@Initialized @NonNull Integer, @Initialized @NonNull String> this, @Initialized @NonNull Integer p0) in java.util.function.Function
  found   : @Initialized @Nullable String
  required: @Initialized @NonNull String
1 error
```

mernst:
I suspect this may be an instance of issue #979, which @smillst  is actively working on.
(But it's still great to have the test case, thanks!)

wmdietl:
I think this is also related to #1345.

cushon:
#979 and #1345 are still open, but this example seems to work now. I sent a PR to enable the test.

PR #4904, opened 2021-09-09
Enable now-passing test for #1630

cushon:


mernst:
I'm closing this because the test is not passing (see the CI run).
@smillst is actively working on issue #979 (which is a big change).

PR #4906, opened 2021-09-09
Must Call Checker must respect the "mustcall" warning suppression prefix, even when the -AnoCreatesMustCallFor option is passed and the MustCallNoCreatesMustCallForChecker is used instead

### LINKED ISSUE #4903: When the -AnoCreatesMustCallFor option is passed to the Resource Leak Checker, warning suppressions for "mustcall" no longer work
Consider the following program:

```
// A test case that the RLC's -AnoCreatesMustCallFor argument doesn't change
// warning suppression behavior for the Must Call Checker.

import org.checkerframework.checker.mustcall.qual.MustCall;

@SuppressWarnings("required.method.not.called")
class DifferentSWKeys {
    void test(@Owning @MustCall("foo") Object obj) {
	// :: warning: unneeded.suppression
	@SuppressWarnings("mustcall")
	@MustCall("foo") Object bar = obj; 
    }

    void test2(@Owning @MustCall("foo") Object obj) {
	// actually needed suppression
	@SuppressWarnings("mustcall")
        @MustCall({}) Object bar = obj; 
    }
}
```

This test case gives different results depending on whether the `-AnoCreatesMustCallFor` option is supplied to the Resource Leak Checker:

```> checker/bin/javac -processor resourceleak -AwarnUnneededSuppressions  DifferentSWKeys.java 
DifferentSWKeys.java:11: warning: [unneeded.suppression] warning suppression "mustcall" is not used by ResourceLeakChecker
	@SuppressWarnings("mustcall")
	^
1 warning
```

```
> checker/bin/javac -processor resourceleak -AwarnUnneededSuppressions -AnoCreatesMustCallFor DifferentSWKeys.java 
DifferentSWKeys.java:18: error: [assignment] incompatible types in assignment.
        @MustCall({}) Object bar = obj; 
                                   ^
  found   : @MustCall("foo") Object
  required: @MustCall Object
1 error
```

I expect the behavior to be the same, regardless of the option.

kelloggm:
fixes #4903

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/mustcall/MustCallNoCreatesMustCallForChecker.java
+++ b/checker/src/main/java/org/checkerframework/checker/mustcall/MustCallNoCreatesMustCallForChecker.java
@@ -2,6 +2,7 @@ package org.checkerframework.checker.mustcall;
 
 import org.checkerframework.framework.qual.StubFiles;
 import org.checkerframework.framework.source.SupportedOptions;
+import org.checkerframework.framework.source.SuppressWarningsPrefix;
 
 /**
  * This copy of the Must Call Checker is identical, except that it does not load the stub files that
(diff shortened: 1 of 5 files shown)


msridhar (approved):
LGTM

PR #4910, opened 2021-09-23
Update FAQ about RUNTIME vs CLASS retention

rbygrave:
Add a note around use of checker-qual-android dependency instead of checker-qual such
that CLASS retention is used.


mernst:
Thanks for this enhancement to the documentation.  We appreciate it!

PR #4926, opened 2021-10-13
Inferring @Owning annotation for final owning fields

Nargeshdb:
This PR is for inferring `@Owning` annotations for final owning fields. First it collects all possible owning fields and then searches for their finalizers. If there is a finalizer for field `f`, then it adds `@Owning` annotation for `f`. A finalizer for a field is a method that satisfies the field's `@MustCall` obligation along some path in its body.

*update

To avoid building CFG for each method again in the postProccessClassTree, I changed the implementation to use the cfg in postAnalyze method. If a method satisfies must call obligation of a field in some path to the regular exit point, then it adds owning annotation for that field and store the method as a finalizer for the field.


checker/src/main/java/org/checkerframework/checker/resourceleak/ResourceLeakAnnotatedTypeFactory.java
```diff
@@ -80,6 +97,84 @@ public ResourceLeakAnnotatedTypeFactory(final BaseTypeChecker checker) {
     this.postInit();
   }
 
+  /**
+   * Collects all the possible owning fields enclosed by the {@link ClassTree} and initializes the
+   * fieldsToFinalizers map.
```
kelloggm, line 102: This method's documentation should probably also include a sentence or two describing the purpose of the `fieldsToFinalizers` map and in what contexts it makes sense to call the method.

checker/src/main/java/org/checkerframework/checker/resourceleak/ResourceLeakAnnotatedTypeFactory.java
```diff
@@ -80,6 +97,84 @@ public ResourceLeakAnnotatedTypeFactory(final BaseTypeChecker checker) {
...
+   * @param element a element
+   * @return true iff the given element is a final field with non-empty @MustCall obligation
+   */
+  private boolean isElementPossibleOwningField(Element element) {
+    return (element.getKind().isField()
+        && ElementUtils.isFinal(element)
+        && !getMustCallValue(element).isEmpty());
+  }
+
+  @Override
```
kelloggm, line 131: Why is this overridden?

checker/src/main/java/org/checkerframework/checker/resourceleak/ResourceLeakAnnotatedTypeFactory.java
```diff
@@ -80,6 +97,84 @@ public ResourceLeakAnnotatedTypeFactory(final BaseTypeChecker checker) {
...
+    super.setRoot(root);
+  }
+
+  /**
+   * Getter method to get the fieldToFinalizers.
+   *
+   * @return the fieldToFinalizers field
+   */
+  public Map<Element, Set<Element>> getFieldToFinalizers() {
+    return fieldToFinalizers;
```
kelloggm, line 142: If this method just returns a direct pointer to the map, why not just make it a public field? Alternatively, this method could return an immutable copy.

checker/src/main/java/org/checkerframework/checker/resourceleak/ResourceLeakAnnotatedTypeFactory.java
```diff
@@ -80,6 +97,84 @@ public ResourceLeakAnnotatedTypeFactory(final BaseTypeChecker checker) {
...
+  @Override
+  public void preProcessClassTree(ClassTree classTree) {
+    fieldToFinalizers.clear();
+    this.fieldToFinalizers = CollectFields(classTree);
+    super.preProcessClassTree(classTree);
+  }
+
+  @Override
+  public void postProcessClassTree(ClassTree tree) {
+    if (checker.hasOption("infer") && !fieldToFinalizers.isEmpty()) {
```
kelloggm, line 154: Rather than checking this directly, you can check whether `getWholeProgramInference() == null` (if it is null, WPI is disabled).

kelloggm (changes requested):
Overall, this looks like solid work. Nice job! I don't have any major concerns, so on my next pass I'll read all of the code carefully rather than just the documentation.

checker/src/main/java/org/checkerframework/checker/resourceleak/ResourceLeakAnnotatedTypeFactory.java
```diff
@@ -80,6 +97,84 @@ public ResourceLeakAnnotatedTypeFactory(final BaseTypeChecker checker) {
...
+          MethodTree mt = (MethodTree) t;
+          Set<Modifier> flags = mt.getModifiers().getFlags();
+          if (flags.contains(Modifier.ABSTRACT) || flags.contains(Modifier.NATIVE)) {
+            break;
+          }
+          if (mt.getBody() == null) {
+            break;
+          }
+          CFGMethod met = new CFGMethod(mt, tree);
+          ControlFlowGraph cfg = CFCFGBuilder.build(this.root, met, checker, this, processingEnv);
```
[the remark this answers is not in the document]
Nargeshdb: https://github.com/typetools/checker-framework/pull/4926/commits/b0e0b8fccbe0024141a5a944ab6dbb875900866a

checker/src/main/java/org/checkerframework/checker/resourceleak/ResourceLeakAnnotatedTypeFactory.java
```diff
@@ -70,6 +78,13 @@
    */
   private final BiMap<LocalVariableNode, Tree> tempVarToTree = HashBiMap.create();
 
+  /**
+   * Map to store a set of FieldToFinalizers for classes. Keys are class trees; values are sets of
+   * FieldToFinalizers which represents a set of method Elements that satisfy @MustCall obligation
+   * of FieldToFinalizers's field along some path to their regular exit points
+   */
+  Map<ClassTree, Set<FieldToFinalizers>> classToFieldToFinalizers = new HashMap<>();
```
msridhar, line 86: As far as I can tell this mapping is never used in this PR, only updated.  It might be useful for other inference tasks later, though.  Is that correct?  If so, maybe we should move its creation and relevant code to a later PR, where it is actually used?

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallInferenceLogic.java
```diff
@@ -0,0 +1,284 @@
...
+  }
+
+  /**
+   * It tracks called methods for fields with non-empty @MustCall obligation of type {@link
+   * FieldWithCMVs} along all paths to the regular exit point in the method body. This is the main
+   * function in the MustCallInferenceLogic.
+   */
+  void inference() {
+
+    Set<BlockWithFieldsFacts> visited = new HashSet<>();
```
msridhar, line 71: Document that the only reason we need a worklist here is to handle cases where `@MustCall` has multiple methods.  It's still not clear to me that supporting that scenario is worth it at this point.

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallInferenceLogic.java
```diff
@@ -0,0 +1,284 @@
...
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(block, fieldsWithCMV);
+    }
+  }
+
+  /**
+   * A pair of a field {@link Element} and a set of method names that were called on the field
+   * somewhere in a method body.
```
msridhar, line 246: Don't they need to be called from blocks on a single control-flow path?

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallInferenceLogic.java
```diff
@@ -0,0 +1,284 @@
...
+   */
+  private static class FieldWithCMVs {
```
msridhar, line 248: Can we just name it `FieldAndCalledMethods`?  I got confused wondering what a CMV was.

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallInferenceLogic.java
```diff
@@ -0,0 +1,284 @@
...
+   * represents a set of fields with a least upper bound of methods that were called somewhere in
+   * the previous blocks.
+   */
+  private static class BlockWithFieldsFacts {
+
+    /** The block. */
+    public final Block block;
+
+    /** The set of FieldsWithCMV. */
+    public final ImmutableSet<FieldWithCMVs> fieldsWithCMV;
```
msridhar, line 213: Could this be a Map from each field to a set of called methods instead?  Representing as a set of pairs seems a bit convoluted and could lead to weird bugs (e.g., since you don't want to allow mutiple entries in the set for the same field)

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallInferenceLogic.java
```diff
@@ -0,0 +1,284 @@
...
+
+    for (Block b : successors) {
+      if (b.getType() == Block.BlockType.SPECIAL_BLOCK) {
+
+        if (((SpecialBlock) b).getSpecialType() == SpecialBlock.SpecialBlockType.EXCEPTIONAL_EXIT) {
+          // Do nothing if the successor block is an exceptional exit point
+          continue;
+        }
+
+        findOwningFields(fieldsWithCMVs, cfgMethod.getMethod());
```
msridhar, line 169: It is strange to see this call here.  From the method name, I would have thought that all this method does is propagate to successors.  Is the call here because you only want it to run when we reach the regular exit point?  If so, please add comments to that effect

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallInferenceLogic.java
```diff
@@ -0,0 +1,284 @@
...
+  /**
+   * Checks all element in the fieldsWithCMVs and adds an {@code @Owning} annotation on a field if
+   * the called methods set of that field in {@link FieldWithCMVs} includes all the methods in
+   * the @MustCall type of the field. It also updates classToFieldToFinalizers with the new detected
+   * finalizer.
+   *
+   * @param fieldsWithCMVs the set of FieldWithCMVs
+   * @param methodTree the method tree
+   */
+  private void findOwningFields(Set<FieldWithCMVs> fieldsWithCMVs, Tree methodTree) {
```
msridhar, line 188: We need a better name for this method.  It actually goes in and mutates the `WholeProgramInference` object, which I wouldn't expect from the name.  Maybe `updateOwningFieldsInWPI`?

msridhar (changes requested):
I am getting the high-level idea; left a few comments.  Overall, I'm not sure if it's worth supporting types with multiple `@MustCall` methods right now, as it adds a lot of complexity

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallInferenceLogic.java
```diff
@@ -0,0 +1,131 @@
...
+    Element receiverEl = TreeUtils.elementFromTree(receiver.getTree());
+
+    if (typeFactory.isElementPossibleOwningField(receiverEl)) {
+      Element method = TreeUtils.elementFromTree(mNode.getTree());
+      List<String> mustCallValues = typeFactory.getMustCallValue(receiverEl);
+
+      // Because we assumed that any must call annotation has at most one method, the following
+      // check is enough to decide whether the receiver is an owning field
+      if (mustCallValues.contains(method.getSimpleName().toString())) {
+        WholeProgramInference wpi = typeFactory.getWholeProgramInference();
```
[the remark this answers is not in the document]
Nargeshdb: We don't go over exceptional successors, so we're always on paths to the regular exit point but what you suggested is clearer and I'll update it.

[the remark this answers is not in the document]
msridhar: Ah interesting. Then why do you need to check if the successor is the exceptional exit in the propagate method?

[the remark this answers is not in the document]
Nargeshdb: you're right! We don't need to check that! 

While I was updating this part and testing the case you mentioned in the next comment, I found a bug in adding successors in the propagation method! `block.getSuccessors()` returns both regular and exceptional successors. So instead of using that I added back `getSuccessors(Block cur)` method.

checker/tests/resourceleak-inference/CheckFields.java
```diff
@@ -0,0 +1,49 @@
+import org.checkerframework.checker.calledmethods.qual.EnsuresCalledMethods;
+import org.checkerframework.checker.mustcall.qual.MustCall;
+import org.checkerframework.checker.mustcall.qual.Owning;
+
+@MustCall("finalizer") class CheckFields {
```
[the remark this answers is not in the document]
msridhar: Ok, we definitely need an automated test somewhere, like a system test

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallInferenceLogic.java
```diff
@@ -0,0 +1,164 @@
...
+        }
+      }
+
+      propagateRegPaths(current, visited, worklist);
+    }
+  }
+
+  /**
+   * If the receiver of {@code mNode} is a possible owning field and the method invocation satisfies
+   * the field's must call obligation, then adds owning annotation for that field.
```
msridhar, line 85: This documentation is out of date.  The method actually marks the field as owning; the annotation is added later

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallInferenceLogic.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+  }
+
+  /**
+   * Returns the non-exceptional successors of the current block.
+   *
+   * @param cur the current block
+   * @return the successors of this current block
+   */
+  private List<Block> getSuccessors(Block cur) {
```
msridhar, line 142:
```suggestion
  private List<Block> getNormalSuccessors(Block cur) {
```

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallInferenceLogic.java
```diff
@@ -0,0 +1,164 @@
...
+   * @param curBlock the current block
+   * @param visited set of blocks already on the worklist
+   * @param worklist current worklist
+   */
+  private void propagateRegPaths(Block curBlock, Set<Block> visited, Deque<Block> worklist) {
+
+    List<Block> successors = getSuccessors(curBlock);
+
+    for (Block b : successors) {
+      if (b.getType() == Block.BlockType.SPECIAL_BLOCK) {
```
msridhar, line 122: Add a comment that the block must be the regular exit, since we do not propagate to exceptional successors

checker/src/test/java/org/checkerframework/checker/test/junit/ResourceLeakInferenceTest.java
```diff
@@ -0,0 +1,26 @@
+package org.checkerframework.checker.test.junit;
+
+import java.io.File;
+import java.util.List;
+import org.checkerframework.checker.resourceleak.ResourceLeakChecker;
+import org.checkerframework.framework.test.CheckerFrameworkPerDirectoryTest;
+import org.junit.runners.Parameterized.Parameters;
+
+/** Tests for the Resource Leak Checker. */
+public class ResourceLeakInferenceTest extends CheckerFrameworkPerDirectoryTest {
```
msridhar, line 10: I am still unclear on how automated testing should work; can you clarify?  I think you said there is a system test to make sure the inference is working.  If so, we should delete this class and `CheckFields.java`, and then update the relevant WPI testing scripts so that the new system test is always run.  There should be some change in this PR that adds a new system test

msridhar (changes requested):
Looking almost there!  A few more comments

checker/src/test/java/org/checkerframework/checker/test/junit/ResourceLeakInferenceTest.java
Nargeshdb: https://github.com/typetools/checker-framework/pull/4926/commits/9449840ffca5cf19c1c5ce46a9263310f5956537

msridhar:
Looks like CI is failing?

https://dev.azure.com/checkerframework/checkerframework/_build/results?buildId=7917&view=logs&j=f885938b-06ab-53d6-b00c-1db87d55a4ce&t=2fa64c6b-278b-5da5-32c2-d8fe18424fd0&l=1497

@kelloggm any idea what is going on?

msridhar (dismissed):
Changes now look good to me, though we need CI to pass

msridhar:
@kelloggm I think this is ready for you to review, while we figure out the CI issue

msridhar:
@Nargeshdb and I are a bit stumped on how to debug the CI failure, which neither of us can reproduce.  One issue is that it seems some output is being swallowed in the CI log.  We see lines like these in the CI log:
```
'jdk.compiler/com.sun.source.util.TreePathScanner.scan(TreePathScanner.java' 'jdk.compiler/com.sun.tools.javac.Main.compile(Main.java' 'jdk.compiler/com.sun.tools.javac.Main.main(Main.java' 'jdk.compiler/com.sun.tools.javac.api.ClientCodeWrapper$WrappedTaskListener.finished(ClientCodeWrapper.java' 'jdk.compiler/com.sun.tools.javac.api.MultiTaskListener.finished(MultiTaskListener.java' 'jdk.compiler/com.sun.tools.javac.main.JavaCompiler.compile(JavaCompiler.java' 'jdk.compiler/com.sun.tools.javac.main.JavaCompiler.flow(JavaCompiler.java' 'jdk.compiler/com.sun.tools.javac.main.Main.compile(Main.java' 'jdk.compiler/com.sun.tools.javac.tree.JCTree$JCClassDecl.accept(JCTree.java' 
```

This seems to indicate an exception was thrown.  But, we do not see the actual exception type and message, the lines are not sorted in the order of the stack trace, and the line numbers are missing.

@kelloggm do you agree that some output is being lost somewhere?  Any suggestions on how to proceed?  My only thoughts are (1) try to set up the same Docker container being used on Azure locally, so we can run WPI manually within the container, or (2) give up on adding a system test and try to go for a unit test instead.

msridhar (dismissed):

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallInferenceLogic.java
```diff
@@ -0,0 +1,168 @@
...
+import org.checkerframework.dataflow.cfg.block.ConditionalBlock;
+import org.checkerframework.dataflow.cfg.block.SingleSuccessorBlock;
+import org.checkerframework.dataflow.cfg.node.MethodInvocationNode;
+import org.checkerframework.dataflow.cfg.node.Node;
+import org.checkerframework.javacutil.AnnotationBuilder;
+import org.checkerframework.javacutil.BugInCF;
+import org.checkerframework.javacutil.TreeUtils;
+
+/**
+ * This class contains Resource Leak Checker annotation inference algorithm. For now, it just
```
kelloggm, line 24: Can we drop the "for now" here? It seems likely to become out of date and it doesn't add much.

checker/src/main/java/org/checkerframework/checker/resourceleak/ResourceLeakAnnotatedTypeFactory.java
```diff
@@ -101,6 +115,15 @@ public void postAnalyze(ControlFlowGraph cfg) {
     MustCallConsistencyAnalyzer mustCallConsistencyAnalyzer =
         new MustCallConsistencyAnalyzer(this, this.analysis);
     mustCallConsistencyAnalyzer.analyze(cfg);
+
+    // Inferring owning annotations for final owning fields
+    if (getWholeProgramInference() != null && cfg != null) {
```
kelloggm, line 120: When can `cfg` be null here? That seems like it might be a bug?

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
```diff
@@ -1128,6 +1155,34 @@ public AnnotatedTypeMirror getType(AnnotatedTypeMirror type, AnnotatedTypeFactor
...
+      if (declarationAnnotations == null) {
+        declarationAnnotations = new HashSet<>();
+      }
+
+      return declarationAnnotations.add(annotation);
+    }
+
+    /**
+     * Returns the inferred declaration annotations on this field, or null if there are no
+     * annotations.
```
kelloggm, line 1175: This method seems to return the empty set if there are no declaration annotations, not null. I think that the code is right and this specification is wrong, but you should check to be sure, and then make the appropriate fix.

kelloggm (changes requested):
I also am about to push some documentation/naming changes, mostly in `MustCallInferenceLogic`. Please check them over and make sure I haven't made any mistakes.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
Nargeshdb: I updated this to be "an empty set"

checker/src/main/java/org/checkerframework/checker/resourceleak/ResourceLeakAnnotatedTypeFactory.java
Nargeshdb: I couldn't find an example in which cfg be null. I think I added this by mistake.

msridhar (approved):

kelloggm (approved):
LGTM

PR #4928, opened 2021-10-14
Remove uses of VisitorState from GuiEffectVisitor

smillst:
These are the only uses of the VisitorState classType and receiverTypes in the framework.


checker/src/main/java/org/checkerframework/checker/guieffect/GuiEffectVisitor.java
```diff
@@ -437,6 +446,7 @@ public Void visitMethod(MethodTree node, Void p) {
     Void ret = super.visitMethod(node, p);
     currentMethods.removeFirst();
     effStack.removeFirst();
+    receiverType = previousReceiverType;
```
mernst, line 449: Should this be in a finally block, or would an exception be fatal enough that there is no point?

mernst (approved):
I have just one comment.  Thanks!

PR #4959, opened 2021-11-12
Fix false positive in Resource Leak Checker related to try-with-resources

Nargeshdb:
RLC reported a false positive for this [line](https://github.com/kelloggm/wpi-many-tests-bibtex-clean/blob/703c0190ab7d7436a5be6e160667f194da44fdee/src/main/java/org/plumelib/bibtex/BibtexClean.java#L60-L61). That was because assignments to resource variables are handled in Must Call Checker which doesn't have any information about resource aliases. This PR removes the set of rhs obligations if lhs is a resource variable and must call type of rhs only contains `close()` method.

Reported error:
`try (PrintWriter out = new PrintWriter(UtilPlume.bufferedFileWriter(outFile.toString()));                                                                       
The type of object is: java.io.BufferedWriter.
Reason for going out of scope: possible exceptional exit due to new EntryReader(filename) with exception type java.io.IOException`

### DIFF
--- a/checker/tests/resourceleak/TryWithResourcesMultiResources.java
+++ b/checker/tests/resourceleak/TryWithResourcesMultiResources.java
@@ -1,0 +1,41 @@
+import java.io.IOException;
+import java.net.Socket;
+import org.checkerframework.checker.calledmethods.qual.*;
+import org.checkerframework.checker.mustcall.qual.*;
+
+public class TryWithResourcesMultiResources {
+
+  class OuterResource implements java.io.Closeable {
+    private final @Owning Socket socket;
+
+    public @MustCallAlias OuterResource(@MustCallAlias Socket sock) throws IOException {
+      this.socket = sock;
+    }
+
+    @Override
+    @EnsuresCalledMethods(
+        value = {"this.socket"},
+        methods = {"close"})
+    public void close() throws IOException {
+      this.socket.close();
+    }
+  }
+
+  public void multiResourcesWrong(String address, int port) {
+    // :: error: required.method.not.called
+    try (OuterResource outer = new OuterResource(new Socket(address, port))) {
+
+    } catch (Exception e) {
+
+    }
+  }
+
+  public void multiResourcesCorrect(String address, int port) {
+    try (Socket s = new Socket(address, port);
+        OuterResource outer = new OuterResource(s)) {
+
+    } catch (Exception e) {
+
+    }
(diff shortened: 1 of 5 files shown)


msridhar (approved):
LGTM assuming CI passes, though it may not; it may complain about missing Javadoc.  Also, I'm surprised the `TryWithResourcesFP` test passes, since it depends on plumelib; but maybe that is just available on the classpath.

msridhar:
@kelloggm can you also take a quick look at this one?

checker/tests/resourceleak/TryWithResourcesMultiResources.java
```diff
@@ -0,0 +1,41 @@
...
+        value = {"this.socket"},
+        methods = {"close"})
+    public void close() throws IOException {
+      this.socket.close();
+    }
+  }
+
+  public void multiResourcesWrong(String address, int port) {
+    // :: error: required.method.not.called
+    try (OuterResource outer = new OuterResource(new Socket(address, port))) {
```
kelloggm, line 26: Can you add a comment here explaining exactly what's wrong with this version of the test? I don't understand the difference between this version and the version that passes below.

kelloggm (approved):
LGTM modulo one thing I'm confused about in the tests. Thanks for the fix.

checker/tests/resourceleak/TryWithResourcesMultiResources.java
msridhar: Yes we should add a comment.  It's a subtle difference.  try-with-resources can only close resources that have been assigned to a variable.  Here, the call to the `OuterResource` constructor can throw an `IOException`, and if that exception is thrown, the `Socket` will not be closed.  In the other version, the `Socket` is assigned to a separate try-with-resources variable, so it _will_ be closed even if the `OuterResource` constructor throws an exception.

Nargeshdb: The difference is that if `new OuterResource(...)` throws an exception, the reference to the socket goes out of scope without being closed.

Nargeshdb: https://github.com/typetools/checker-framework/pull/4959/commits/7690901a26ac163da9c6877b9a4743c99b10022e

msridhar (approved):

PR #4976, opened 2021-12-03
Don't crash on switch expressions or switch statements with arrow cases.

### LINKED ISSUE #4911: Guide how to add Checker to existing huge code base.
I am trying to plug-in Checker to existing **massive** project. Tons of Java source files etc, so it is physically impossible to annotate whole code base at once. So my first idea was to enable Checker in Gradle, then annotate all the soucre the way that would make Checker transparent. So I added

```kotlin
configure<CheckerFrameworkExtension> {
  checkers = listOf(
    "org.checkerframework.checker.nullness.NullnessChecker",
  )
  extraJavacArgs = listOf(
    "-AsuppressWarnings=initialization",
  )
  excludeTests = true
}
```

and then I wanted to make my sources compilable despite not being annotated properly. I read the docs and found that while by default `@NonNull` applies, I can override this with `@DefaultQualifier` annotation, so I annotated all the classes with:

```
@DefaultQualifier(value = Nullable.class, locations = TypeUseLocation.OTHERWISE)
public class XXX {
....
```

but this is still not making Checker transparent as I am still facing errors like:

```
error: [dereference.of.nullable] dereference of possibly-null reference hex
HexModel model = hex.getModel();
```

So at this point I'd appreciate any assistance that would help me make my code compile w/o detailed annotations as apparently I am missing something here in my actions.

smillst:
I added an issue to track the remaining switch expressions todos: https://github.com/typetools/checker-framework/issues/4977.

Fixes #4911.


dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java
```diff
@@ -469,6 +469,46 @@ public PhaseOneResult process(CompilationUnitTree root, UnderlyingAST underlying
...
+  /**
+   * Visit a SwitchExpressionTree
+   *
+   * @param switchExpressionTree a SwitchExpressionTree, typed as Tree to be backward-compatible
+   * @param p parameter
+   * @return the result of visiting the switch expression tree
+   */
+  public Node visitSwitchExpression17(Tree switchExpressionTree, Void p) {
+    // TODO: Analyze switch expressions properly.
+    return new MarkerNode(
```
smillst, line 506: I decided not to add a SwitchExpressionNode, because I wasn't sure what fields it would need when we properly implement switch expressions.

msridhar:
This change prevents crashes not only for `switch` expressions, but also for `switch` statements that use the new kind of "arrow" `case` labels, right?

smillst:
@msridhar 
> This change prevents crashes not only for `switch` expressions, but also for `switch` statements that use the new kind of "arrow" `case` labels, right?

Yes.  I'll update the change log, PR title to make this clear.

PR #4981, opened 2021-12-10
Fix discrepancy between checking of MustCallAlias in the paper and in the implementation

kelloggm:
This closes a soundness bug that @msridhar noticed.


checker/tests/resourceleak/MustCallAliasSubstitution.java
```diff
@@ -0,0 +1,16 @@
+// A test case that checks that resolving the obligations of one object and then
+// substituting a fresh object is not counted as an alias, for the purpose of MustCallAlias
+// verification.
+
+import java.io.*;
+import java.net.Socket;
+import org.checkerframework.checker.mustcall.qual.*;
+
+class MustCallAliasSubstitution {
```
msridhar, line 9: For illustrative purposes, it might be nice to extend this test case with a call to `example()`, where `close()` is invoked on the value returned from `example()`.  That will show how without reporting the error, the tool could allow a leak.

msridhar (commented):
Can you give a brief description of your strategy for fixing this?  From what I can tell, it looks like there is a new case analysis at every point where an `Obligation` may get resolved, where the algorithm decides whether an `Obligation` derived from a `@MustCallAlias` parameter can be resolved in that particular way.  Does that sound right?

kelloggm:
> Can you give a brief description of your strategy for fixing this? From what I can tell, it looks like there is a new case analysis at every point where an Obligation may get resolved, where the algorithm decides whether an Obligation derived from a `@MustCallAlias` parameter can be resolved in that particular way. Does that sound right?

@msridhar that sounds about right. Basically, my strategy was to add a bit to each `ResourceAlias` that records whether it is a `@MustCallAlias` parameter. Then, any time an `Obligation` gets resolved, we check that bit if the way it was resolved was some way that a `@MustCallAlias` parameter cannot get resolved.

Admittedly, this strategy isn't as clean or elegant as I would have liked. But, it was straightforward to implement given the existing code.

checker/tests/resourceleak/MustCallAliasSubstitution.java
kelloggm: b40c5ec38fb46385d34261762a2fc5ba602e3aa6

msridhar (approved):
The changes here all look correct to me.  It's hard to tell for sure from the diff if we've checked all the possible cases of obligation removal that need to be special cased, but it looks like we covered them to me

PR #4982, opened 2021-12-10
Add support for switch expressions in dataflow

### LINKED ISSUE #2373: Dataflow support for switch expressions

### LINKED ISSUE #4979: Make nodes for switch expressions appear on their own in the containing CFG block
I'm working on getting NullAway to not crash in the presence of switch expressions, now that Checker dataflow does not crash for them.  One small issue is that the marker node for a switch expression may not appear as a top-level node in the containing CFG block, so it never gets directly visited by the dataflow analysis.  If I have a statement `Object o = null;`, I get four nodes in the corresponding CFG basic block: a `VariableDeclarationNode`, a `NullLiteralNode`, an `AssignmentNode`, and a `LocalVariableNode`.  But, if I write `Object o = switch (i) { ... };`, I just get three nodes in the corresponding block:  a `VariableDeclarationNode`, an `AssignmentNode`, and a `LocalVariableNode`; the marker node for the switch expression is missing.

I imagine the representation of `switch` expressions in CFGs is very temporary, and as it gets handled it will get its own node in the containing block for cases like these.  And I can add a special case to work around this for now.  Just wanted to note the issue.

smillst:
They are analyzed the same as switch statements, except that dataflow adds adds an assignment between each result expression and a dataflow-created variable at each yield statement or expression that is a result expression.  The value of that dataflow-created variable is the value of the switch expression.

~~This includes https://github.com/typetools/checker-framework/pull/4978, merge that one first.~~

Delete SKIP-REQUIRE-JAVADOC after merging!

Fixes #2373 and fixes #4979.


SKIP-REQUIRE-JAVADOC
```diff
@@ -0,0 +1 @@
+Delete after merge.
```
mernst, line 1: Why is this file necessary?   I don't see wide-ranging uninteresting edits in this pull request.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/node/SwitchExpressionNode.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Objects;
+import javax.lang.model.type.TypeMirror;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/** A node for a switch expression. */
+public class SwitchExpressionNode extends Node {
+
+  /** The switch expression tree corresponding to this node. */
```
mernst, line 13: Does this have a specific run-time type?  If so, document that.  Probably the constructor should throw an exception if an object of the wrong type is passed in.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java
```diff
@@ -2128,28 +2137,63 @@ public Node visitSwitch(SwitchTree tree, Void p) {
    */
   private class SwitchBuilder {
     /** The switch tree. */
```
mernst, line 2139: Javadoc above says this nested class is only for switch statements; I think that should be updated, and this Javadoc should say "It might be a switch statement or a switch expression."

dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java
```diff
@@ -2128,28 +2137,63 @@ public Node visitSwitch(SwitchTree tree, Void p) {
...
+      this.caseBodyLabels = new Label[caseTrees.size() + 1];
     }
 
-    /** Build up the CFG for the switchTree. */
-    public void build() {
+    /**
+     * Build up the CFG for the switchTree.
+     *
+     * @return if the switch is a switch expression, then a {@link SwitchExpressionNode}; otherwise,
+     *     null
```
mernst, line 2192: Why is that the right thing?  I guess probably because elsewhere builds all the CFG nodes for the case bodies and none is needed explicitly for the switch statement itself.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java
```diff
@@ -2160,16 +2204,20 @@ public void build() {
...
 
-      // Build CFG for the cases.
-      extendWithNode(
-          new MarkerNode(
-              switchTree,
-              "start of switch statement #" + TreeUtils.treeUids.get(switchTree),
-              env.getTypeUtils()));
+      buildSwitchExpressionVar();
+
+      if (switchTree.getKind() == Kind.SWITCH) {
```
mernst, line 2209: After this line, I would add a comment "// It's a switch statement, not a switch expression.".
Likewise elsewhere.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java
```diff
@@ -2180,17 +2228,35 @@ public void build() {
...
+                env.getTypeUtils()));
+      }
 
-      extendWithNode(
-          new MarkerNode(
-              switchTree,
-              "end of switch statement #" + TreeUtils.treeUids.get(switchTree),
-              env.getTypeUtils()));
+      switchBuilder = oldSwitchBuilder;
+      if (switchExprVarTree != null) {
```
mernst, line 2245: This seems to be the same test as `switchTree.getKind() == Kind.SWITCH` elsewhere.  Would it make the code easier to read & understand if all the tests were consistent?

dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java
```diff
@@ -2256,11 +2341,51 @@ private void buildCase(CaseTree tree, int index) {
           scan(stmt, null);
         }
       } else {
-        scan(TreeUtils.caseTreeGetBody(tree), null);
+        Tree bodyTree = TreeUtils.caseTreeGetBody(tree);
+        if (bodyTree instanceof ExpressionTree) {
```
mernst, line 2345: This might be a third different test for the "switch statement vs switch expression" concept.  I think consistency would make the code easier to read.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java
```diff
@@ -2256,11 +2341,51 @@ private void buildCase(CaseTree tree, int index) {
...
+        } else {
+          scan(bodyTree, null);
+        }
       }
       extendWithExtendedNode(new UnconditionalJump(nextBodyL));
       addLabelForNextNode(nextCaseL);
     }
+
+    /**
+     * Does the following for each result expression of a switch expression:
```
mernst, line 2356: This comment seems wrong.  The client has the logic that this sentence described.  This method only handles one result expression at a time.

mernst (changes requested):

SKIP-REQUIRE-JAVADOC
smillst: The problem is that NodeVisitor has 80 undocumented methods.  I would have to document them all to get the Javadoc test to pass. See https://github.com/typetools/checker-framework/runs/4509004521.

dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java
smillst: That's right. A switch statement does not have a value so it doesn't need to have a corresponding node.

PR #4985, opened 2021-12-14
Fix multiple case constants.

smillst:


framework/tests/value/java17/MultiCaseConst.java
```diff
@@ -0,0 +1,31 @@
+// @below-java17-jdk-skip-test
+import org.checkerframework.common.value.qual.IntVal;
+
+public class MultiCaseConst {
```
msridhar, line 4: Based on reading the code it should work fine, but it might be nice to have a test using the "arrow" case labels as well

smillst: I've got such a test case, but it won't pass until #4984 is merged.  I'll open a pull request or add it to this pull request when that happens.  Thanks!

Issue #4977, opened 2021-12-03
Fully support switch expressions


smillst:
https://github.com/typetools/checker-framework/pull/4976 makes it so that the Checker Framework doesn't not crash on switch expressions, but it does not handle them as precisely as possible.  Here's a list of TODOs to finish supporting switch expressions:
- [x] `TypeFromExpressionVisitor.java`: Correctly compute the type of switch expressions. (https://github.com/typetools/checker-framework/pull/4978)
- [x]  `BaseTypeVisitor.java`: Check that the type of the result expressions of the switch expression are subtypes of the type of the switch expression. (https://github.com/typetools/checker-framework/pull/4978)
- [x] `CFGTranslationPhaseOne.java` : analyze switch expressions. (https://github.com/typetools/checker-framework/pull/4982) 
 - [x] `BaseTypeVisitor.java`: `JointJavacJavaParserVisitor` needs to be updated to handle switch expressions. (https://github.com/typetools/checker-framework/pull/4992)

msridhar:
@Gellert5225 has started looking into handling arrow case labels in switch _statements_, but he is not looking at switch _expressions_.  When I spoke to @mernst we decided the work for statements and expressions could be done somewhat in parallel, but if you think closer coordination is warranted, let us know and we can discuss further.

smillst:
`CFGTranslationPhaseOne.SwitchBuilder` needs to be update to handle the arrow case labels.  I think support for switch expressions will use the updated switch builder.  (I say "think", because Mike and I haven't settled on an implementation strategy yet.). However, we can probably work on support for switch expressions before `SwitchBuilder` is updated by just using test cases that use `case L  : <block>` or  `case : -> <expression>` varieties.

msridhar:
Ok, sounds good, @Gellert5225 and I will communicate / open PRs as we make progress.

smillst:
All the tasks have been fixed, so I'm closing this issue.

PR #4988, opened 2021-12-15
JointJavacJavaParserVisitor now handles switch expressions

smillst:
JavaParsers crashes on some yields and also parses some to objects other than YieldStmts. ( https://github.com/javaparser/javaparser/issues/3364)

This pull request doesn't rely on any other, but it does duplicate TreeUtil methods.  So, it should probably be merged after all the others.

Fixes the last bullet point in https://github.com/typetools/checker-framework/issues/4977.


smillst:
I merged the wrong branch and messed up the history.  I've opened a new PR #4992 .

PR #4989, opened 2021-12-15
add note to documentation of Owning about why it is a declaration annotation, and fix and reenable a test for java 17

### LINKED ISSUE #4934: `@Owning` should be a type annotation
If a fix for #4932 is merged before a fix for this issue is merged, then some tests will need to be re-enabled.

`org.checkerframework.checker.mustcall.qual.Owning` is currently defined as a declaration annotation (not a type annotation):

```java
@Target({ElementType.METHOD, ElementType.PARAMETER, ElementType.FIELD})
public @interface Owning {}
```

This ensures that `@Owning` is only written on methods, parameters, and fields.  However, it is problematic for two reasons.  First, its documentation indicates that it is about types rather than parameters.  Second, in Java 17 a declaration annotation cannot be written on the `this` type parameter.  Under Java 17, running the tests yields

```
org.checkerframework.checker.test.junit.MustCallTest > run[/home/mernst/research/types/checker-framework-fork-mernst-branch-jdk17/checker/mustcall] FAILED
    java.lang.AssertionError: 0 out of 40 expected diagnostics were found.
    1 unexpected diagnostic was found
    OwningParams.java:8: error: annotation @org.checkerframework.checker.mustcall.qual.Owning not applicable in this type context
```

kelloggm:
fixes #4934

### DIFF
--- a/checker-qual/src/main/java/org/checkerframework/checker/mustcall/qual/NotOwning.java
+++ b/checker-qual/src/main/java/org/checkerframework/checker/mustcall/qual/NotOwning.java
@@ -7,7 +7,8 @@ import java.lang.annotation.Target;
 
 /**
  * Annotation indicating that ownership should not be transferred to the annotated parameter, field,
- * or (when this is written on a method) return value, for the purposes of Must Call checking.
+ * or method's call sites, for the purposes of Must Call checking. For a full description of the
+ * semantics, see the documentation of {@link Owning}.
  *
  * <p>Parameters and fields are treated as if they have this annotation by default unless they have
  * {@link Owning}.
--- a/checker-qual/src/main/java/org/checkerframework/checker/mustcall/qual/NotOwning.java
+++ b/checker-qual/src/main/java/org/checkerframework/checker/mustcall/qual/NotOwning.java
@@ -7,7 +7,7 @@ import java.lang.annotation.Target;
 
 /**
  * Annotation indicating that ownership should not be transferred to the annotated parameter, field,
- * or (when this is written on a method) return type, for the purposes of Must Call checking.
+ * or (when this is written on a method) return value, for the purposes of Must Call checking.
  *
  * <p>Parameters and fields are treated as if they have this annotation by default unless they have
  * {@link Owning}.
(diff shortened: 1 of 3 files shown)


checker-qual/src/main/java/org/checkerframework/checker/mustcall/qual/NotOwning.java
```diff
@@ -7,7 +7,7 @@
 
 /**
  * Annotation indicating that ownership should not be transferred to the annotated parameter, field,
- * or (when this is written on a method) return type, for the purposes of Must Call checking.
+ * or (when this is written on a method) return value, for the purposes of Must Call checking.
```
mernst, line 10: I would not make this change.  The return value is a run-time notion.  If it's used for compile-time, then it would refer to the expression in a `return` statement.

For a declaration annotation, you are right that it's not really on the return type.  Personally, I would leave it.  But if you want to be a bit more precise, you can say that it is written on the method declaration itself, but interpreted as being on the return type.

checker-qual/src/main/java/org/checkerframework/checker/mustcall/qual/Owning.java
```diff
@@ -7,11 +7,20 @@
 
 /**
  * Annotation indicating that ownership should be transferred to the annotated parameter, field, or
- * (when written on a method) return type, for the purposes of Must Call checking. Static fields
+ * (when written on a method) return value, for the purposes of Must Call checking. Static fields
```
mernst, line 10: Same comment as above.

checker-qual/src/main/java/org/checkerframework/checker/mustcall/qual/Owning.java
```diff
@@ -7,11 +7,20 @@
...
  * cannot be owning.
  *
- * <p>Method return types are treated as if they have this annotation by default unless their method
- * is annotated as {@link NotOwning}.
+ * <p>This annotation is a declaration annotation rather than a type annotation, because it does not
+ * logically belong to any type hierarchy. Logically, it is a directive to the Resource Leak Checker
+ * that informs it whether checking that a value's must-call obligations have been satisfied is
+ * necessary, or not. In that way, it can be viewed as an annotation expressing an aliasing
```
mernst, line 16: Hard to parse.  Try "whether it is necessary to check ...".

mernst (dismissed):
Thanks for these improvements.  I have a few comments on the wording.

PR #5000, opened 2022-01-01
Add synthetic variables to CFG for ternary expressions

### LINKED ISSUE #4994: Proposal: during CFG construction, create a synthetic variable for the value of each ternary expression
As in the recent handling of `switch` expressions (see #4982), it would be useful if there were a synthetic variable created for the value of each ternary expression during CFG construction, with an assignment of the "then" and "else" sub-expressions to the variable in the relevant CFG blocks.  I can think of two advantages:

1. Transfer functions in dataflow analysis could be simplified.  If there is already a transfer function for assignments, then it is likely that the abstract value for a ternary expression will just be the abstract value for its synthetic variable.
2. For analyses that track results for each control-flow path separately, like `MustCallConsistencyAnalyzer`, reasoning about ternary expressions will become much simpler.  Right now, `MustCallConsistencyAnalyzer` has some tricky special-case logic for ternary expressions.  With a synthetic variable for the ternary expressions, correct handling would fall out of handling the corresponding assignments.

@smillst @mernst thoughts?

msridhar:
Fixes #4994.  See that issue for the proposal.

The changes to dataflow were straightforward, and mirror the logic used for switch expressions.  These changes made the tests for the Resource Leak Checker fail, since it had a bunch of hacky CFG-pattern-matching logic for dealing with ternary expressions.  That code is now removed or greatly simplified.  Also simplify `CFAbstractTransfer.visitTernaryExpression()`.


msridhar:
The CI test failure should go away once #5001 lands

checker/src/main/java/org/checkerframework/checker/mustcall/MustCallTransfer.java
```diff
@@ -178,7 +178,11 @@ private void lubWithStoreValue(CFStore store, JavaExpression expr, AnnotationMir
   public TransferResult<CFValue, CFStore> visitTernaryExpression(
       TernaryExpressionNode node, TransferInput<CFValue, CFStore> input) {
     TransferResult<CFValue, CFStore> result = super.visitTernaryExpression(node, input);
-    updateStoreWithTempVar(result, node);
+    if (!TypesUtils.isPrimitiveOrBoxed(node.getType())) {
+      // Add the synthetic variable created during CFG construction to the temporary
```
msridhar, line 182: We don't need to update the stores here anymore, since the synthetic variable is already in the CFG and tracked

checker/src/main/java/org/checkerframework/checker/resourceleak/ResourceLeakTransfer.java
```diff
@@ -42,7 +42,11 @@ public ResourceLeakTransfer(final ResourceLeakAnalysis analysis) {
   public TransferResult<CFValue, CFStore> visitTernaryExpression(
       TernaryExpressionNode node, TransferInput<CFValue, CFStore> input) {
     TransferResult<CFValue, CFStore> result = super.visitTernaryExpression(node, input);
-    updateStoreWithTempVar(result, node);
+    if (!TypesUtils.isPrimitiveOrBoxed(node.getType())) {
+      // Add the synthetic variable created during CFG construction to the temporary
```
msridhar, line 46: We don't need to update the stores here anymore, since the synthetic variable is already in the CFG and tracked

checker/tests/nullness/KeyForLub.java
```diff
@@ -27,7 +27,7 @@ void method(
   }
 
   @PolyKeyFor String poly1(@KeyFor("map1") String key1, @PolyKeyFor String poly) {
-    // :: error: (return)
+    // :: error: (conditional)
```
msridhar, line 30: The error message changed due to the change in `CFAbstractTransfer

checker/tests/resourceleak/ACRegularExitPointTest.java
```diff
@@ -299,102 +299,4 @@ void testSubFoo2() {
     // :: error: required.method.not.called
     SubFoo f = new SubFoo();
   }
-
```
msridhar, line 302: I moved these test inputs to a new class `TernaryExpressions` to ease debugging test failures

dataflow/src/main/java/org/checkerframework/dataflow/cfg/node/TernaryExpressionNode.java
```diff
@@ -17,33 +17,87 @@
  */
 public class TernaryExpressionNode extends Node {
 
+  /** The {@code ConditionalExpressionTree} corresponding to this node */
```
msridhar, line 20: Had to add a bunch of Javadoc here to satisfy CI

kelloggm (approved):
All these changes look good to me - thanks for doing this!

I'm not an expert on the CFG construction code, so someone who is ought to review that code again before this is merged. Wrt to the changes to the RLC, this looks great.

smillst (approved):
Looks good.  Thanks!

PR #4999, opened 2021-12-30
Update BCEL line number

mernst:

PR #5001, opened 2022-01-02
Update expected inference output for bcel-util

msridhar:
Required to make inference tests pass,  due to https://github.com/plume-lib/bcel-util/commit/1fefa60c329236486f74891e0b44567d622c7c7c


msridhar:
Whoops, duplicate of #4999

PR #5003, opened 2022-01-05
Capture enclosing types.

### LINKED ISSUE #4996: Wildcards aren't captured for inner classes that use type parameters from the outer class
```java
abstract class CaptureOuter<T> {
  abstract T get();

  abstract class Inner {
    abstract T get();
  }
}

class Client {
  Object getFrom(CaptureOuter<?> o) {
    return o.get();
  }

  Object getFrom(CaptureOuter<?>.Inner o) {
    return o.get();
  }
}
```

```
$ $HOME/checker-framework-3.21.0/checker/bin/javac -processor nullness CaptureOuter.java
CaptureOuter.java:11: error: [return] incompatible types in return.
    return o.get();
                ^
  type of expression: capture#84 extends @Initialized @Nullable Object
  method return type: @Initialized @NonNull Object
CaptureOuter.java:15: error: [return] incompatible types in return.
    return o.get();
                ^
  type of expression: @Initialized ? extends @Initialized @Nullable Object
  method return type: @Initialized @NonNull Object
2 errors
```
 
The call to CaptureOuter.get() has a type of `capture#84 extends @Initialized @Nullable Object`, as expected.
 
But the call to CaptureOuter.Inner.get() has a type of `@Initialized ? extends @Initialized @Nullable Object`.
 
Contrast this to what javac does if the return types of the getFrom methods is changed to String: In both cases, it identifies the call as having a type of `CAP#1`.

smillst:
Fixes #4996.

### DIFF
--- a/checker/tests/nullness/NullnessIssue4996.java
+++ b/checker/tests/nullness/NullnessIssue4996.java
@@ -9,10 +9,12 @@ class NullnessIssue4996 {
 
   class Client {
     Object getFrom(CaptureOuter<?> o) {
+      // :: error: (return)
       return o.get();
     }
 
     Object getFrom(CaptureOuter<?>.Inner o) {
+      // :: error: (return)
       return o.get();
     }
   }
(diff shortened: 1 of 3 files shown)


mernst (approved):
Elegant one-line fix!

PR #5004, opened 2022-01-07
Add note to update Daikon.

smillst:
Also, remove note about Randoop, since dependabot should update that now.


mernst (approved):
Thanks!

PR #5005, opened 2022-01-09
Add support for switch expressions to Resource Leak Checker

### LINKED ISSUE #4993: Support switch expressions in the Resource Leak Checker
This will require some (hopefully modest) modifications to the `MustCallConsistencyAnalyzer`.  Fortunately, there is already a synthetic variable for the value of each `switch` expression created during CFG construction.  So I think all we will have to do is treat that variable as the "temporary" representing the `switch` expression, and everything else should work out.

msridhar:
The only change required was to update temporary variable mappings, like we now do for ternary expressions.

Fixes #4993


checker/tests/resourceleak/SwitchExpressions.java
```diff
@@ -0,0 +1,180 @@
+// @below-java17-jdk-skip-test
```
msridhar, line 1: These tests are exactly the same as those in `TernaryExpressions`, but using switch expressions instead

msridhar:
FYI @Nargeshdb

kelloggm (dismissed):
LGTM

msridhar:
Sorry @kelloggm your review got dismissed because I just moved a file to a different directory

kelloggm (approved):

PR #5009, opened 2022-01-14
Issue an error when super bounded wildcards are collapsed 

### LINKED ISSUE #5006: Nullness checker false negative related to capture conversion
The nullness checker doesn't report any diagnostics for the following program, which fails at runtime with an NPE:

```java

interface C<T> {
  T get();
}

interface X {
  C<? extends Object> get();
}

interface Y extends X {
  @Override
  C<? super Object> get(); // should be an error
}

class A {
    C<@Nullable Object> c = () -> null;
    Y y = () -> c;
    X x = y;
    x.get().get().toString(); // NPE
  }
}
```

```
$ ./checker-framework-3.21.1/checker/bin/javac -processor org.checkerframework.checker.nullness.NullnessChecker A.java
$ java A
        at A.main(A.java:21)
```

smillst:
and the annotations on the super and extends bounds are the same.

(This is a javac bug, see https://bugs.openjdk.java.net/browse/JDK-8054309)

Fixes #5006.


framework/src/main/java/org/checkerframework/common/basetype/BaseTypeValidator.java
```diff
@@ -580,27 +580,45 @@ protected Void visitParameterizedType(AnnotatedDeclaredType type, ParameterizedT
...
+              checker.reportError(
+                  tree.getTypeArguments().get(i),
+                  "type.argument",
+                  element.getTypeParameters().get(i),
+                  element.getSimpleName(),
+                  wildcard.getExtendsBound(),
+                  capturedTypeVar.getUpperBound());
+            }
+          } else if (AnnotatedTypes.isExplicitlySuperBounded(wildcard)) {
+            if (!(atypeFactory
```
mernst, line 606: I understand what is going on here thanks to your verbal explanation, but without that I would be lost.  I think a comment here would be helpful.

mernst (dismissed):

mernst (approved):

PR #5010, opened 2022-01-14
Elaborate on how to choose Optional vs nullness

kennknowles:
In some cases with generics, Optional is actually the right choice. Adding that wording to the manual.

I have to give this explanation to collaborators and coworkers pretty frequently.

### DIFF
--- a/docs/manual/nullness-checker.tex
+++ b/docs/manual/nullness-checker.tex
@@ -1251,6 +1251,18 @@ but written for regular possibly-null Java references.
 To use the \refclass{checker/nullness/util}{Opt} class, the
 \<checker-util.jar> file must be on the classpath at run time.
 
+It is true that \<Optional> can express types that cannot be expressed via
+nullness annotations. Nullability cannot be composed, while \<Optional>
+can. This is most importantly in code that uses generics. When a generic
+method with type paramter \code{T} accepts an optional argument of type
+\code{@Nullable T}, it is impossible to distinguish whether the argument
+is null because the user chose not to provide it, or because the underlying
+type \code{T} also allows null. These two situations often have completely
+different meanings. A simple example of such a situation is when
+a command line parameter can be absent, empty, or non-empty. In summary,
+for a concrete class it is often simpler to use nullness annotations,
+while applying \code{@Nullable} to a type variable always risks
+conflating two semantically distinct occurrences of null.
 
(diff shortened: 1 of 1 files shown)


docs/manual/nullness-checker.tex
```diff
@@ -1251,6 +1251,18 @@
 To use the \refclass{checker/nullness/util}{Opt} class, the
 \<checker-util.jar> file must be on the classpath at run time.
 
+It is true that \<Optional> can express types that cannot be expressed via
```
mernst, line 1254: Kenn, thanks for his addition.  I would like to include information to address this issue, but I would like to clarify the description first.

Could you please include an example?  The current description is a bit vague, making it harder to understand than necessary.  It also uses undefined terms like "composed"; please define that, or replace it, or give an example.

For a command-line parameter, an invalid or default value can often represent a missing argument.  (For example, `""` for a string or -1 for an Integer.)  You might argue that is bad style, or make an argument that it is not always possible, but please make such an argument explicitly, or else the writing will not be as persuasive as it should be.

The compositionality advocated by this patch violates the [first rule about `Optional`](https://checkerframework.org/manual/#optional-checker): "Never, ever, use null for an Optional variable or return value.".  It may be of value to sometimes violate a rule, but please do address this explicitly and justify exactly when and why you think the rule should be violated.

I wonder if this text belongs in Section 3.7.3  "Relationship to `Optional<T>`" or Chapter 5 "Optional Checker for possibly-present data" rather than at this location.  There could be a cross-reference to it from the current location, of course.

Finally, CI is failing.  Could you please add your name to `docs/manual/contributors.tex` as described in the [CI failure message](https://dev.azure.com/checkerframework/checkerframework/_build/results?buildId=8145&view=logs&jobId=4b19ccca-7d7d-5749-26de-b87349fba2b4&j=4b19ccca-7d7d-5749-26de-b87349fba2b4&t=4dba1f6c-8e7b-5172-04db-4dd554c55b12)?

Thanks again!

mernst (changes requested):

kennknowles:
Oh, here's a goodie: https://issues.apache.org/jira/browse/BEAM-13709

mernst:
I'm closing this because the original poster has not responded to 3 requests for clarification over 3 months.  If someone else wants to wordsmith this, we welcome the contribution.

PR #5015, opened 2022-01-21
Fix for issue 5013: handle binding variables when when reading java expression from string 

kelloggm:


kelloggm:
I've investigated the test failure here and I'm not sure how to resolve it. @smillst do you have any idea how to handle the issue described below?

The tests are failing due to a `BugInCF` error issued by the method `testJointJavacJavaParserVisitor` in `BaseTypeVisitor`. This issue does not occur when I run the index checker from the command line on its own; only during the tests. I think it has something to do with the processing of `.ajava` files.

The problem appears to be that the `JointJavacJavaParserVisitor` does not have any notion of what to do with a binding variable. There is a method in that class `visitBindingPattern17` that I think is where such an action should take place, but it's empty and returns `null`. I don't know this part of the code base, so I'm not exactly certain what it is supposed to be doing - from what I can gather, it's just testing that every Javac Tree has a matching node in the JavaParser parse? But not sure how that is achieved.

Here are some possible paths forward:
* figure out what needs to go in `JointJavacJavaParserVisitor#visitBindingPattern17` to make the test succeed
* disable the test in the Index Checker tests, somehow.

There are two things I'm unsure of:
* what the specification of `JointJavacJavaParserVisitor#visitBindingPattern17` should be
* what the consequences of disabling this test for the Index Checker's test suite would be

smillst (approved):

mernst (approved):

PR #5017, opened 2022-01-22
Make result of LambdaResultExpressionNode non-null

msridhar:
I think this code was modeled on that in `ReturnNode`, where the result can be `null` (for a statement `return;` with no value).  But if a lambda has an expression body, it cannot be empty; so here the result should be `@NonNull`.


mernst (approved):
Thanks for this simplification.

PR #5018, opened 2022-01-22
Make CFGLambda.method @Nullable

msridhar:
A lambda can be written outside of a method (e.g., in a field declaration), so `CFGLambda.method` should be `@Nullable`


mernst (approved):
Thanks for this fix!

PR #5030, opened 2022-01-27
Add option to wpi.sh that allows custom build commands

kelloggm:
@Nargeshdb can you give this version of `wpi.sh` a try the next time you're running on a multi-module project, and let me know if the instructions make sense. We shouldn't merge this until that happens, so that we know the instructions are good enough for a real user to follow.

### DIFF
--- a/checker/bin/wpi.sh
+++ b/checker/bin/wpi.sh
@@ -9,7 +9,7 @@
 set -eo pipefail
 # not set -u, because this script checks variables directly
 
-while getopts "d:t:b:g:" opt; do
+while getopts "d:t:b:g:c:" opt; do
   case $opt in
     d) DIR="$OPTARG"
        ;;
@@ -18,6 +18,8 @@ while getopts "d:t:b:g:" opt; do
     b) EXTRA_BUILD_ARGS="$OPTARG"
        ;;
     g) GRADLECACHEDIR="$OPTARG"
+       ;;
+    c) COMPILE_CMD="$OPTARG"
        ;;
     \?) # echo "Invalid option -$OPTARG" >&2
        ;;
@@ -135,6 +137,9 @@ fi
 function configure_and_exec_dljc {
 
   if [ -f build.gradle ]; then
+      if [ "${COMPILE_CMD}" = "" ]; then
+        COMPILE_CMD="compileJava"
+      fi
       if [ -f gradlew ]; then
         chmod +x gradlew
         GRADLE_EXEC="./gradlew"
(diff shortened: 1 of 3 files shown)


docs/manual/inference.tex
```diff
@@ -164,7 +164,15 @@
   or Maven buildfile. The default is the current working directory.
 
 \item[-t TIMEOUT]
-  The timeout for running the checker, in seconds
+  The timeout for running the checker, in seconds.
+
+\item[-c COMPILE\_CMD]
```
mernst, line 169: I find "compile_cmd" and "the command to use" a bit misleading.
It's actually a build system target (plus additional command-line arguments to the build system? I'm not sure, but I recall the example you gave to Narges including those rather than putting them in `-b`.)
I think the naming add documentation should be clarified.

Also, please add a concrete example of what a user might pass on the command line, to further clarify.
This might help with the text "depending on the setup of the target project", which is currently vague and doesn't give actionable advice.

mernst (changes requested):
Thanks for this feature.

The logic looks good.  I have a comment on the naming and documentation.

mernst:
Please mention the new command-line option in the changelog.

docs/manual/inference.tex
kelloggm: I added an example. I still find it somewhat unsatisfying, but I'm not sure that it's possible to give advice that's broadly applicable here, since what the user will need to do varies project-by-project, and these options exist to allow that kind of customization. Let me know if you have any better ideas than my latest set of changes.

checker/bin/wpi.sh
```diff
@@ -19,6 +19,8 @@ while getopts "d:t:b:g:" opt; do
        ;;
     g) GRADLECACHEDIR="$OPTARG"
        ;;
+    c) BUILD_TARGET="$OPTARG"
```
mernst, line 22: "c" is not a great option name, but "b" and "t" are already taken, so OK.

docs/manual/inference.tex
```diff
@@ -149,10 +149,21 @@
...
+The \<wpi.sh> script is most useful when analyzing projects that follow the standard
+conventions of their build system for single-module projects. When analyzing a project
+that requires non-standard build system commands, use the \<-c> and \<-b> options to
+override the defaults used by \<wpi.sh>. For example, suppose that you wanted to use
+\<wpi.sh> to infer annotations for the Resource Leak Checker on
+\ahref{https://github.com/apache/zookeeper}{Apache Zookeeper's}
+\<zookeeper-server> module only. Without \<wpi.sh>, one might use the following command:
+\<mvn -B --projects zookeeper-server --also-make install -DskipTests>. To use this command via
+\<wpi.sh>, use the following:
+\<sh wpi.sh -b "-B --projects zookeeper-server --also-make" -c "install -DskipTests" -- resourceleak>.
```
mernst, line 161: Would this have worked if `-DskipTests` were in the `-b` command-line argument rather than the `-c` command-line argument?

mernst (changes requested):
This looks good, thanks.  I have just a couple small documentation requests.

checker/bin/wpi.sh
kelloggm: I agree! I considered some other names, but didn't like any of them. And I don't want to change any of the existing options, for backwards-compatibility.

docs/manual/inference.tex
kelloggm: Possibly? Skipping tests on a `clean` command doesn't make any sense, but I don't think Maven would throw an error because of it.

kelloggm: I put it in the `-c` argument both because it makes more sense only for `install` and because I wanted to show that the user can put command-line arguments there, too, if they're only for the compilation phase.

mernst (approved):
Thanks!

PR #5032, opened 2022-01-28
Type check and refine binding patterns.

### LINKED ISSUE #5021: Add proper type checking and type refinement for binding variables.
```
// @below-java17-jdk-skip-test

public class TaintingBindingVariable {

  void bar1(@Untainted Object o) {
    if (o instanceof String s) {
      @Untainted String f = s;  // legal
    }
  }
  void bar2( Object o) {
    if (o instanceof @Untainted String s) { // error or warning?
      @Untainted String f = s;
    }
  }
}
```

There should be no errors in the `bar1` method, but there should be an error or waring in the `bar2` method.

smillst:
Fix #5021.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -55,6 +55,9 @@ switch (JavaVersion.current()) {
     case JavaVersion.VERSION_1_9:
     case JavaVersion.VERSION_1_10:
     case JavaVersion.VERSION_12:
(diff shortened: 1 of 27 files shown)


mernst (approved):
Thanks!

PR #5045, opened 2022-02-08
Fix `CFGTranslationPhaseOne#findOwner()` for lambdas

### LINKED ISSUE #5042: 3.21.2 infers wrong type for locals assigned to by conditional expression
After upgrading from 3.20.0 to 3.21.1, I discovered the following issue in some of our source files, which also reproduces in 3.21.2.

# Command

```
./checker-framework-3.21.2/checker/bin/javac \
  -processor org.checkerframework.checker.nullness.NullnessChecker \
  T.java
```

# Inputs

`./T.java`:

```

class T {
  interface PromptViewModel {
    boolean isPending();
    @Nullable PromptButtonViewModel getConfirmationButton();
  }
  interface PromptButtonViewModel {
    @Nullable ConfirmationPopupViewModel getConfirmationPopup();
  }
  interface ConfirmationPopupViewModel {
    boolean isShowingConfirmation();
  }

  static final Function<PromptViewModel, Boolean> IS_PENDING_OR_SHOWING_CONFIRMATION =
      (viewModel) -> {
        PromptButtonViewModel prompt = viewModel.getConfirmationButton();
        ConfirmationPopupViewModel popup = prompt != null ? prompt.getConfirmationPopup() : null;
        return viewModel.isPending() || (popup != null && popup.isShowingConfirmation());
      };

  boolean f(PromptViewModel viewModel) {
    PromptButtonViewModel prompt = viewModel.getConfirmationButton();
    ConfirmationPopupViewModel popup = prompt != null ? prompt.getConfirmationPopup() : null;
    return viewModel.isPending() || (popup != null && popup.isShowingConfirmation());
  }
}
```

# Output

```
T.java:19: error: [conditional] incompatible types in conditional expression.
        ConfirmationPopupViewModel popup = prompt != null ? prompt.getConfirmationPopup() : null;
                                                                                       ^
  found   : @Initialized @Nullable ConfirmationPopupViewModel
  required: @Initialized @NonNull ConfirmationPopupViewModel
T.java:19: error: [conditional] incompatible types in conditional expression.
        ConfirmationPopupViewModel popup = prompt != null ? prompt.getConfirmationPopup() : null;
                                                                                            ^
  found   : null (NullType)
  required: @Initialized @NonNull ConfirmationPopupViewModel
2 errors
```

Note that the errors are both in the definition of `IS_PENDING_OR_SHOWING_CONFIRMATION`; none of them are in the definition of `f`, which should have the same nullness implications.

# Expectation

I expected no error. The checker framework should infer that, since popup may be assigned a null value, its type is `@Nullable`. This is fine, since it is checked before use. More generally, I would expect the same behavior for the method `f` as for the lambda with the same body.

We speculated that the checker might somehow be confused about lambdas, since the erroneous case occurs in a lambda. But wrapping the body of `f` with a lambda does not reproduce the issue. My current guess is that, because a field is initialized with a lambda, the checker is tricked into thinking that the lambda's local variables are actually fields, for which it does not wish to infer nullability. But I wasn't able to narrow the reproduction case down sufficiently to prove this.

smillst:
Fixes #5042.

(I didn't change `CFAbstractTransfer#initialStore` because I think that method is doing something different with initializer  blocks.)

### DIFF
--- a/checker/tests/nullness/Issue5042.java
+++ b/checker/tests/nullness/Issue5042.java
@@ -1,0 +1,67 @@
+import java.util.function.Function;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+public class Issue5042 {
+  interface PromptViewModel {
+    boolean isPending();
+
+    @Nullable PromptButtonViewModel getConfirmationButton();
+  }
+
+  interface PromptButtonViewModel {
+    @Nullable ConfirmationPopupViewModel getConfirmationPopup();
+  }
+
+  interface ConfirmationPopupViewModel {
+    boolean isShowingConfirmation();
+  }
+
+  boolean f(PromptViewModel viewModel) {
+    PromptButtonViewModel prompt = viewModel.getConfirmationButton();
+    ConfirmationPopupViewModel popup = prompt != null ? prompt.getConfirmationPopup() : null;
+    return viewModel.isPending() || (popup != null && popup.isShowingConfirmation());
+  }
+
+  static final Function<PromptViewModel, Boolean> IS_PENDING_OR_SHOWING_CONFIRMATION =
+      (viewModel) -> {
+        @Nullable PromptButtonViewModel promptLambda = viewModel.getConfirmationButton();
+        @Nullable ConfirmationPopupViewModel popup =
+            promptLambda != null ? promptLambda.getConfirmationPopup() : null;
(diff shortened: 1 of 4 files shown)


mernst (approved):
This looks great.  The comments are very informative, thanks.

PR #5046, opened 2022-02-09
Prevent WPI crash related to ClassCastException

kelloggm:
Encountered by @Nargeshdb. The test that failed before the change is `ArrayAndTypevar.java`.

I included the other test (`AnonymousOverride.java`) in the PR because my initial hypothesis was that that was the code pattern causing the crash, and WPI did not have a test for that sort of code pattern. I don't see any harm in having it as a regression test.


mernst (dismissed):

mernst:
@kelloggm This fix causes test failures.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceImplementation.java
```diff
@@ -662,6 +662,19 @@ protected void updateAnnotationSet(
     if (rhsATM instanceof AnnotatedNullType && ignoreNullAssignments) {
       return;
     }
+
+    // If the rhsATM and the lhsATM have different kinds (which can happen e.g. when
+    // an array type is substituted for a type parameter), do not attempt to update
+    // the inferred type, because this method is written with the assumption
+    // that rhsATM and lhsATM are the same kind.
+    if (rhsATM.getKind() != lhsATM.getKind()) {
+      // The one difference in kinds situation that this method can account for is the RHS being
```
mernst, line 671: One possibility is to forbid all cases where the kinds differ, and add a few exceptions.
The other possibility is to only forbid certain cases.  The latter is less likely to lead to difficult-to-debug problems.  It is also more likely to lead to future crashes that we must address, but at least we'll know about them rather than them being silent failures.  Therefore, the second possibility seems better to me.

mernst (changes requested):
I have a comment on the exclusion strategy.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceImplementation.java
kelloggm: I considered the approach you suggested, and chose this approach for two reasons:
1. simplicity: there are more cases that we want to forbid than cases that we want to permit, so rejecting anything that isn't a specifically-listed exception results in simpler code, and
2. WPI's specification generally permits it to fail to make a true inference, except in the cases we know to work (and test). So, I preferred to fail to make (some) inferences rather than crashing, which prevents WPI from being used at all and therefore fails to make any inferences.

I'm open to changing to the other approach. Do either of the reasons above change your feeling that the allow-by-default approach is better than the current deny-by-default approach I've proposed?

kelloggm: @mernst, do you have an opinion on this? I'm assigning this PR to you, because I'd like to discuss what the right behavior for WPI in this case actually is.

kelloggm:
@mernst, as we discussed yesterday, I've added a warning to this PR that will trigger whenever an unexpected pair of type kinds are encountered. The change was more invasive than I'd expected (owing to the need to thread through an appropriate place for the error). I also find the allow-list for combinations that do not trigger a warning to seem kind of silly once I actually wrote the code for it - type kinds are a rather course way to determine whether an update ought to be permitted or not. Let me know what you think.

kelloggm:
@mernst as we discussed yesterday, I've added a debug option to WPI that prints a simple message when inference fails. This version is very simple: it doesn't print any other debugging information like the location where the annotation should have been inferred or anything like that. The reason for this is that the place that the issue can be detected is quite deep in the WPI logic, in code that only has access to the two `AnnotatedTypeMirror`s representing the declared and the inferred types. That means I'm not particularly confident how useful it'll be for debugging.

I think it's an open question whether it is worthwhile to complicate this code further by threading through enough information to add a line number to the error message. What do you think?

(Also, let's get this PR (and therefore the machinery for this new debugging option) merged before we try to make any more progress on #5057 or #5059.)

PR #5048, opened 2022-02-09
Widening from an integral type to an integral type

smillst:
should not change the annotation.

From the [JLS](https://docs.oracle.com/javase/specs/jls/se17/html/jls-5.html#jls-5.1.2)

> A widening conversion of a signed integer value to an integral type T simply sign-extends the two's-complement representation of the integer value to fill the wider format.

```
@Unsigned short s = -2;
@Signed int i = s; // this should be an error.
```

@mernst  is this the right or am I misunderstanding something?


mernst:
@smillst The existing code was incorrect; I overthought it.
@markro Could you look at the test failures and determine how to update the tests (or find more errors in the checker)?
Thanks!

mernst (approved):

PR #5051, opened 2022-02-10
Add an error for String concats with an unsigned value.

smillst:
This includes the changes in #5054, so merge that one first.


checker/src/main/java/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -119,6 +120,17 @@ public Void visitBinary(BinaryTree node, Void p) {
         }
         break;
 
+      case PLUS:
+        if (TreeUtils.isStringConcatenation(node)) {
+          if (leftOpType.hasAnnotation(Unsigned.class)) {
```
smillst, line 125: This isn't right because it allows the following:

```
 @UnknownSignedness int unknown = ...;
String s = "" +unknown; // this should be an error.
```

However the following should not be an error:

```
 @UnknownSignedness Date unknown = ...;
String s = "" +unknown; // this should be ok.
```

checker/tests/signedness/LocalVarDefaults.java
```diff
@@ -8,7 +8,7 @@ void methodInt(@Unsigned int unsignedInt, @Signed int signedInt) {
     int local2 = signedInt;
   }
 
-  // :: error: (annotations.on.use) :: error:  (anno.on.irrelevant)
```
mernst, line 11: It would be nice to retain this error if possible, only for floating-point types.
A user might think that `@Unsigned` and `@Signed` are applicable to floating-point types, but they are not, and the error would prevent a possible user confusion.

mernst (changes requested):
This looks good (thanks!).  I have one request to retain an error.

checker/src/main/java/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -237,4 +278,18 @@ public Void visitCompoundAssignment(CompoundAssignmentTree node, Void p) {
...
+  @Override
+  protected Set<? extends AnnotationMirror> getExceptionParameterLowerBoundAnnotations() {
+    return Collections.singleton(atypeFactory.SIGNED);
+  }
+
+  @Override
+  protected void checkConstructorResult(
+      AnnotatedExecutableType constructorType, ExecutableElement constructorElement) {}
+
+  @Override
```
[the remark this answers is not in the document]
smillst: It's needed to get the `anno.on.irrelevant` without a `@RelevantTypes` annotation.

PR #5055, opened 2022-02-14
Remove manual section that isn't true.

smillst:


docs/manual/signedness-checker.tex
```diff
@@ -126,24 +126,6 @@
...
-
-\subsectionAndLabel{Widening conversions}{signedness-checker-widening-conversions}
-
-Figure~\ref{fig-signedness-hierarchy} shows the type qualifier hierarchy.
-When upcasting among integral types, the expression has type
-\<@SignedPositive> because the value is guaranteed to be within the signed
-positive range.  For example:
-
-\begin{Verbatim}
-@Signed int sint;
```
mernst, line 138: I think the facts about the signed values are true, even if the ones about the unsigned values are not true.

Rather than removing this, I think it would be more useful to slightly expand it:
 * signed narrow types (as currently)
 * unsigned narrow types that are correctly widened (no errors, type is @SignedPositive)
 * unsigned narrow types that are not correctly widened (same code as now, but indicate an error occurs)

Is that proposal correct?

mernst (changes requested):

docs/manual/signedness-checker.tex
smillst: The facts about the signed values are not true.
```
@Signed int sint = ... some negative int...;
@SignedPositive long splong = sint;  // error
```
The assignment is not legal because the most significant bit of `slong` is set.

mernst (approved):

PR #5056, opened 2022-02-17
Optimize LocalVariable#hashCode/equals

artempyanykh:
**Summary**:
I've been profiling my checker that uses Dataflow, and
LocalVariable.hashCode came out **very hot** both on CPU and memory
allocation profiles.

`equals/hashCode` methods build a bunch of strings which is very 
expensive when you e.g. put your LocalVariables into a HashMap.

The new implementation of equals compares `Name name` and `Symbol owner`
using their intrinsic equality. Which should be fine with Names since
these are all interned. I'm not 100% sure about Symbols but it looks
like those are also properly unique.

The implementation of hashCode now follows that of equal. Note that I
dropped `TypeAnnotationUtils.unannotatedType(vs.type)` from `hashCode`
as it seems to be redundant and produces a new type that may be
structurally equivalent but not referentially equal to some existing
type.

**Test plan:**
1. CF's `./gradlew test` passes with the new implementation. 
2. Our internal checker tests also pass. 
3. Moreover, there were basically no difference in checker
    reports when ran against our internal code before and after the change
    which is a good indicator that the change is safe to do.

### DIFF
--- a/dataflow/src/main/java/org/checkerframework/dataflow/expression/LocalVariable.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/expression/LocalVariable.java
@@ -7,7 +7,6 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 import org.checkerframework.dataflow.cfg.node.LocalVariableNode;
 import org.checkerframework.javacutil.AnnotationProvider;
 import org.checkerframework.javacutil.ElementUtils;
-import org.checkerframework.javacutil.TypeAnnotationUtils;
 import org.checkerframework.javacutil.TypesUtils;
 
 /**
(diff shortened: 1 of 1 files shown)


mernst (dismissed):
Thank you for this improvement!  I appreciate it, and your careful testing.

We must have been concerned about interning, but I guess it is not (currently?) an issue.
And the previous discrepancy between `hashCode` and `equals` seems wrong.

Let's merge if @smillst also approves.

(For future reference, please do not force-push to pull requests.  We always squash-and-merge, and force-pushing can remove comments or dissociate them from the code -- which wasn't a concern for this pull request.  Thanks!)

Thanks again.

smillst (approved):
>The new implementation of equals compares `Name name` and `Symbol owner`
using their intrinsic equality. Which should be fine with Names since
these are all interned. I'm not 100% sure about Symbols but it looks
like those are also properly unique.

In the past, the Checker Framework would create different CFGs for the main checker and  each of its subcheckers.  That would cause a different synthetic variable with a different element and name to be created for what should be the same variable.  The Checker Framework now creates a single CFG for use by the main and subcheckers, so I think the names should be interned.   This also means that the implementation of `sameElement` could be changed to `vs1.equals(vs2)` and `hashCode` could be change to `vs.hashCode()`.

smillst (dismissed):
I was wrong different elements are created by `JavaExpressionParseUtil#parse`, so I reverted @mernst's changes and added a comment.

mernst (dismissed):
Thanks!

smillst (approved):

PR #5057, opened 2022-02-18
Fix varargs-related IndexOutOfBoundsException during WPI

kelloggm:
This problem was discovered by running stub-based WPI on the all-systems tests. There are a few remaining issues there other than this one, but they all seem pretty minor, and all seem to be related to purity. (They seem especially minor compared to the problems we encountered when we tried the same thing with ajava-based WPI.)


mernst (approved):
Thanks for the fix!

PR #5059, opened 2022-02-23
WPI: do not crash when a lambda parameter is re-assinged

kelloggm:
Based on a crash encountered by @Nargeshdb. The previous logic was too stringent: there is not a guarantee that the existence of a formal parameter Element implies the existence of an enclosing ExecutableElement for which that Element is a parameter, because lambda expressions are not represented by Elements, but their parameters are.


mernst (approved):

kelloggm:
@mernst before merging this, we should finish our discussion of #5046 - this and #5057 both assume that it's acceptable for WPI to give up on inference rather than crashing.

PR #5078, opened 2022-03-09
Improve build speed by reusing computations from previous builds

jiushiyaojinqu:
[Gradle daemon](https://docs.gradle.org/current/userguide/gradle_daemon.html#header). The Daemon is a long-lived process that help to avoid the cost of JVM startup for every build. Since Gradle 3.0, Gradle daemon is enabled by default. For an older version, you should enable it by setting `org.gradle.daemon=true`.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.


checker/bin-devel/build.sh
```diff
@@ -81,6 +81,6 @@ fi
 (./gradlew help || sleep 10) > /dev/null 2>&1
 
 echo "running \"./gradlew assemble\" for checker-framework"
-./gradlew assemble --console=plain --warning-mode=all -s --no-daemon -Dorg.gradle.internal.http.socketTimeout=60000 -Dorg.gradle.internal.http.connectionTimeout=60000
+./gradlew assemble --console=plain --warning-mode=all -s --daemon -Dorg.gradle.internal.http.socketTimeout=60000 -Dorg.gradle.internal.http.connectionTimeout=60000
```
mernst, line 84: Since running the daemon is the default, it is better to remove `--no-daemon` rather than replacing it with `--daemon`.

mernst (changes requested):
Thanks for this change.  We wrote the scripts at a time that the Gradle developers recommended not to use the daemon on CI.  They have since changed their advice.

jiushiyaojinqu:
@mernst  Thanks for your reply!

I have remove the `--no-daemon`

mernst:
Your pull request left in 17 occurrences of `--no-daemon` in the Checker Framework.  I have removed them, making your pull request comprehensive.

There is still a CI failure, which you should have noticed.  I have pasted it below.  Could you please correct that?  Then, we can merge this pull request!

Thanks again.

```
diff --git a/docs/manual/contributors.tex b/docs/manual/contributors.tex
index ed108fe82..0874876ad 100644
--- a/docs/manual/contributors.tex
+++ b/docs/manual/contributors.tex
@@ -122,4 +122,5 @@ Vladimir Sitnikov,
 Vlastimil Dort,
 Weitian Xing,
 Werner Dietl,
-Zhiping Cai.
+Zhiping Cai,
+jiushiyaojinqu.
+ set +x
docs/manual/contributors.tex is not up to date.
If the above suggestion is appropriate, run: make -C docs/manual contributors.tex
If the suggestion contains a username rather than a human name, then do all the following:
  * Update your git configuration by running:  git config --global user.name "YOURFULLNAME"
  * Add your name to your GitHub account profile at https://github.com/settings/profile
  * Make a pull request to add your GitHub ID to
    https://github.com/plume-lib/plume-scripts/blob/master/git-authors.sed
    and remake contributors.tex after that pull request is merged.
```

PR #5103, opened 2022-04-05
Correct TreeUtil#constructor

smillst:


smillst:
I think `constructorFromUse` is going to require more extensive refactoring, which will change the refactors in this pull request.  So, I'm closing this and will open a new pull request when all the refactoring is done.

PR #5106, opened 2022-04-12
Improve performance of TransferResult.getRegularStore

Nargeshdb:
The call to TransferResult.getRegularStore() for conditional blocks was expensive because it computed the least upper bound of else and then store for each call. This PR should fix this performance issue.

### DIFF
--- a/dataflow/src/main/java/org/checkerframework/dataflow/analysis/ConditionalTransferResult.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/analysis/ConditionalTransferResult.java
@@ -28,6 +28,9 @@ public class ConditionalTransferResult<V extends AbstractValue<V>, S extends Sto
   /** The 'else' result store. */
   protected final S elseStore;
 
+  /** The regular result store. */
+  protected final S store;
+
   /**
    * Create a new {@link #ConditionalTransferResult(AbstractValue, Store, Store, Map, boolean)},
    * using {@code null} for {@link #exceptionalStores}.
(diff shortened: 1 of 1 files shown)


msridhar:
@smillst I think you might be the best person to review this change, if you have time?

dataflow/src/main/java/org/checkerframework/dataflow/analysis/ConditionalTransferResult.java
```diff
@@ -113,12 +116,13 @@ public ConditionalTransferResult(
     this.thenStore = thenStore;
     this.elseStore = elseStore;
     this.storeChanged = storeChanged;
+    this.store = thenStore.leastUpperBound(elseStore);
```
smillst, line 119: The then and else stores can change after the ConditionalTransferResult is created, so I don't think this will work.  I suspect that's why the tests aren't failing.  If you can get the tests to pass, I can take a closer look.

Nargeshdb: Thanks for your comment. I will work on getting the tests to pass.

Nargeshdb: @smillst There are a lot of places that store can change and I am not sure whether tracking those changes is more efficient. I closed this PR and opened a new [PR](https://github.com/typetools/checker-framework/pull/5107) that cache the result of getStoreAfter in MustCallConsistencyAnalyzer where stores can't change.

PR #5108, opened 2022-04-14
Improve constructorFromUse for anonymous classes

smillst:
Review in this order: 
1. framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
2. framework/src/main/java/org/checkerframework/framework/util/AnnotatedTypes.java
3. Everything else

### DIFF
--- a/javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
+++ b/javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
@@ -354,25 +354,20 @@ public final class TreeUtils {
   }
 
   /**
-   * Returns the super constructor invoked by {@code newClassTree}.
-   *
-   * <p>If {@code newClassTree} is an anonymous class, then this method returns the super
-   * constructor invoked in the body of the anonymous constructor.
-   *
-   * <p>If {@code newClassTree} is not an anonymous class, then this method returns {@link
-   * #constructor(NewClassTree)}.
+   * Returns the constructor invoked by {@code newClassTree} unless {@code newClassTree} is creating
+   * an anonymous class. In which case, the super constructor is returned.
    *
    * @param newClassTree the constructor invocation
    * @return the super constructor invoked in the body of the anonymous constructor; or {@link
-   *     #constructor(NewClassTree)} if {@code newClassTree} is not an anonymous class
+   *     #constructor(NewClassTree)} if {@code newClassTree} is not creating an anonymous class
    */
   public static ExecutableElement getSuperConstructor(NewClassTree newClassTree) {
     if (newClassTree.getClassBody() == null) {
       return constructor(newClassTree);
     }
     JCNewClass jcNewClass = (JCNewClass) newClassTree;
-    // Anonymous constructor bodies should contain exactly one statement
-    // in the form:
+    // Anonymous constructor bodies, which are always synthetic, contain exactly one statement in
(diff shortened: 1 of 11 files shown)


javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -355,45 +354,53 @@ public static ExecutableElement elementFromUse(NewClassTree node) {
   }
 
   /**
-   * Determines the symbol for a constructor given an invocation via {@code new}.
+   * Returns the super constructor invoked by {@code newClassTree}.
+   *
+   * <p>If {@code newClassTree} is an anonymous class, then this method returns the super
```
mernst, line 359: This is a bit imprecise: `newClassTree` is always `new ...`.  It might be creating an anonymous class, however.
Same comment applies below for "not an anonymous class".

javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -355,45 +354,53 @@ public static ExecutableElement elementFromUse(NewClassTree node) {
...
+   * @param newClassTree the constructor invocation
+   * @return the super constructor invoked in the body of the anonymous constructor; or {@link
+   *     #constructor(NewClassTree)} if {@code newClassTree} is not an anonymous class
+   */
+  public static ExecutableElement getSuperConstructor(NewClassTree newClassTree) {
+    if (newClassTree.getClassBody() == null) {
+      return constructor(newClassTree);
+    }
+    JCNewClass jcNewClass = (JCNewClass) newClassTree;
+    // anonymous constructor bodies should contain exactly one statement
```
mernst: Are such bodies always synthetic?

javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -355,45 +354,53 @@ public static ExecutableElement elementFromUse(NewClassTree node) {
   }
 
   /**
-   * Determines the symbol for a constructor given an invocation via {@code new}.
+   * Returns the super constructor invoked by {@code newClassTree}.
```
mernst, line 357: I think this returns the constructor directly invoked, unless an anonymous class is being constructed, in which case the super constructor is returned.  Is that accurate?

javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -355,45 +354,53 @@ public static ExecutableElement elementFromUse(NewClassTree node) {
...
    *
    * @see #elementFromUse(NewClassTree)
    * @param tree the constructor invocation
    * @return the {@link ExecutableElement} corresponding to the constructor call in {@code tree}
    */
   public static ExecutableElement constructor(NewClassTree tree) {
-
-    if (!(tree instanceof JCTree.JCNewClass)) {
+    if (!(tree instanceof JCNewClass)) {
       throw new BugInCF("TreeUtils.constructor: not a javac internal tree");
```
mernst, line 401: Is this necessary?  The next line does a cast, so the error message will be essentially the same with or without this `if` and `throw`.

mernst (approved):

mernst (approved):

PR #5119, opened 2022-04-18
Consolidate JDK 8 Maven instructions

mernst:


docs/manual/external-tools.tex
```diff
@@ -990,7 +980,26 @@
   Error Prone \<javac.jar>.  They were declared as Maven dependencies above.
```
smillst, line 980: `They were declared as Maven dependencies above.`. That's no longer true.  Maybe just delete this?

smillst (dismissed):

smillst (approved):

PR #5120, opened 2022-04-19
Fix fully-qualified name of CFGVisualizeLauncher in manual

msridhar:
The fully-qualified name changed to ` org.checkerframework.dataflow.cfg.visualize.CFGVisualizeLauncher` but the manual was not updated.


mernst (approved):
Thanks for this fix!

PR #5124, opened 2022-04-25
Check `equals()` like `==`

mernst:
~Merge after #5123.~
Merge with https://github.com/typetools/jdk/pull/127.

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/signedness/SignednessVisitor.java
+++ b/checker/src/main/java/org/checkerframework/checker/signedness/SignednessVisitor.java
@@ -3,12 +3,16 @@ package org.checkerframework.checker.signedness;
 import com.sun.source.tree.BinaryTree;
 import com.sun.source.tree.CompoundAssignmentTree;
 import com.sun.source.tree.ExpressionTree;
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.source.tree.MethodTree;
 import com.sun.source.tree.Tree;
 import java.util.Collections;
 import java.util.Set;
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.type.TypeKind;
+import org.checkerframework.checker.interning.InterningVisitor;
+import org.checkerframework.checker.interning.qual.EqualsMethod;
 import org.checkerframework.checker.signedness.qual.PolySigned;
 import org.checkerframework.checker.signedness.qual.Signed;
 import org.checkerframework.checker.signedness.qual.Unsigned;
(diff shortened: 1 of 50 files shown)


checker/src/main/java/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -158,6 +166,63 @@ public Void visitBinary(BinaryTree node, Void p) {
...
+        } else if (params == 2) {
+          leftOpType = atypeFactory.getAnnotatedType(node.getArguments().get(0));
+          rightOpType = atypeFactory.getAnnotatedType(node.getArguments().get(1));
+        } else {
+          throw new BugInCF("Checked that params is 1 or 2");
+        }
+        if (!atypeFactory.maybeIntegral(leftOpType) || !atypeFactory.maybeIntegral(rightOpType)) {
+          // nothing to do
+        } else if (leftOpType.hasAnnotation(Unsigned.class)
+            && rightOpType.hasAnnotation(Signed.class)) {
```
smillst, line 214: If the rightOp is `@SignednessGlb` or any other subtype of `@Signed` then an error won't be issued.

smillst (changes requested):
You may want to add a few test case for this change.

checker/src/main/java/org/checkerframework/checker/signedness/SignednessVisitor.java
mernst: Every subtype of `@Signed` is also a subtype of `@Unsigned`, so I think this is OK.

smillst (dismissed):

checker/src/main/java/org/checkerframework/checker/signedness/SignednessVisitor.java
```diff
@@ -161,6 +166,62 @@ public Void visitBinary(BinaryTree node, Void p) {
...
+  public Void visitMethodInvocation(MethodInvocationTree node, Void p) {
+    ExecutableElement methElt = TreeUtils.elementFromUse(node);
+    boolean hasEqualsMethodAnno =
+        atypeFactory.getDeclAnnotation(methElt, EqualsMethod.class) != null;
+    if (hasEqualsMethodAnno || InterningVisitor.isInvocationOfEquals(node)) {
+      int params = methElt.getParameters().size();
+      if (!(params == 1 || params == 2)) {
+        checker.reportError(
+            node, "invalid.method.annotation", "@EqualsMethod", "1 or 2", methElt, params);
+        // fallthrough
```
smillst, line 195: I think this is a copy paste error?

smillst (approved):

PR #5127, opened 2022-04-25
Restrict instantiations of `@PolySigned`

mernst:


docs/manual/creating-a-checker.tex
```diff
@@ -569,25 +569,73 @@
...
+
+\begin{Verbatim}
+  void foo(@UnknownSignedness Object a, @UnknownSignedness Object b) { ... }
+\end{Verbatim}
+
+\noindent
+which is consistent with any two arguments, regardless of their signedness
+type.
+
+It is possible to permit only method calls whose arguments have exactly the
```
smillst, line 626: Should we say something about how this still allows upcasts to top?  Also if a type system designer is considering this, then the type system may need [class qualifier polymorphism](https://checkerframework.org/manual/#class-qualifier-polymorphism).

checker/src/main/java/org/checkerframework/checker/signedness/SignednessAnnotatedTypeFactory.java
```diff
@@ -690,4 +693,43 @@ public boolean isRelevant(TypeMirror tm) {
     }
     return true;
   }
+
+  /**
+   * Requires that, when two formal parameter types are annotated with {@code @PolySigned}, the two
```
smillst, line 698: The Javadoc of PolySigned should be updated, and probably its section of the manual.

smillst (changes requested):

smillst (approved):

PR #5128, opened 2022-04-26
`@SignedPositive` corresponds to `@NonNegative`

mernst:


kelloggm:
Should the annotations be aliased, then? (That is, should the Signedness Checker treat `@NonNegative` as `@SignedPositive`, and the Index Checker treat `@SignedPositive` as `@NonNegative`? Implementing that would be pretty trivial.)

smillst (dismissed):

checker/src/main/java/org/checkerframework/checker/signedness/SignednessAnnotatedTypeFactory.java
```diff
@@ -98,6 +99,8 @@ public class SignednessAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
   public SignednessAnnotatedTypeFactory(BaseTypeChecker checker) {
     super(checker);
 
+    addAliasedTypeAnnotation(NonNegative.class, SIGNEDNESS_GLB);
```
smillst, line 102: You shouldn't need this because the Signedness Checker uses the Value Checker, which reads `@NonNegative`.  Also, I think this change is causing the test failure.

Issue #5136, opened 2022-05-10
Resource Leak Checker crash with IllegalArgumentException


msridhar:
Input:
```java
  class Handler extends Thread {
      boolean running;
      static class Call { boolean isResponseDeferred() { return true; } }
    @Override
    public void run() {
      while (running) {
        Call call = null;
        try {
          if (running) {
            continue;
          }
        } catch (Exception e) {
        } finally {
            String s = call.isResponseDeferred() ? ", deferred" : "";
        }
      }
    }
  }
```
Output:
```
$ java -jar checker/dist/checker.jar -processor resourceleak /tmp/Handler.java
error: SourceChecker.typeProcess: unexpected Throwable (IllegalArgumentException) while processing /tmp/Handler.java; message: value already present: call.isResponseDeferred() ? ", deferred" : ""
  ; The Checker Framework crashed.  Please report the crash.
  Compilation unit: /tmp/Handler.java
  Last visited tree at line 1 column 3:
    class Handler extends Thread {
  Exception: java.lang.IllegalArgumentException: value already present: call.isResponseDeferred() ? ", deferred" : ""; java.lang.IllegalArgumentException: value already present: call.isResponseDeferred() ? ", deferred" : ""
  	at org.checkerframework.com.google.common.collect.HashBiMap.put(HashBiMap.java:300)
  	at org.checkerframework.com.google.common.collect.HashBiMap.put(HashBiMap.java:281)
  	at org.checkerframework.checker.resourceleak.ResourceLeakAnnotatedTypeFactory.addTempVar(ResourceLeakAnnotatedTypeFactory.java:234)
  	at org.checkerframework.checker.resourceleak.ResourceLeakTransfer.visitTernaryExpression(ResourceLeakTransfer.java:49)
  	at org.checkerframework.checker.resourceleak.ResourceLeakTransfer.visitTernaryExpression(ResourceLeakTransfer.java:23)
  	at org.checkerframework.dataflow.cfg.node.TernaryExpressionNode.accept(TernaryExpressionNode.java:108)
  	at org.checkerframework.dataflow.analysis.AbstractAnalysis.callTransferFunction(AbstractAnalysis.java:349)
  	at org.checkerframework.dataflow.analysis.ForwardAnalysisImpl.callTransferFunction(ForwardAnalysisImpl.java:372)
  	at org.checkerframework.dataflow.analysis.ForwardAnalysisImpl.performAnalysisBlock(ForwardAnalysisImpl.java:128)
[...]

PR #5137, opened 2022-05-10
Resource Leak Checker: Add a check to avoid mapping multiple keys to a ternary expression for tempVarToTree

Nargeshdb:
Fixes [#5136](https://github.com/typetools/checker-framework/issues/5136)

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/resourceleak/ResourceLeakAnnotatedTypeFactory.java
+++ b/checker/src/main/java/org/checkerframework/checker/resourceleak/ResourceLeakAnnotatedTypeFactory.java
@@ -231,7 +231,9 @@ public class ResourceLeakAnnotatedTypeFactory extends CalledMethodsAnnotatedType
    * @param tree the tree of the expression the tempvar represents
    */
   /* package-private */ void addTempVar(LocalVariableNode tmpVar, Tree tree) {
-    tempVarToTree.put(tmpVar, tree);
+    if (!tempVarToTree.containsValue(tree)) {
+      tempVarToTree.put(tmpVar, tree);
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


checker/src/main/java/org/checkerframework/checker/resourceleak/ResourceLeakAnnotatedTypeFactory.java
```diff
@@ -231,7 +231,9 @@ protected ResourceLeakAnalysis createFlowAnalysis() {
    * @param tree the tree of the expression the tempvar represents
    */
   /* package-private */ void addTempVar(LocalVariableNode tmpVar, Tree tree) {
-    tempVarToTree.put(tmpVar, tree);
+    if (!tempVarToTree.containsValue(tree)) {
```
msridhar, line 234: I'm a little concerned about this change.  If `addTempVar()` is being called, the caller has already created a new `tmpVar` for the `tree`.  It seems problematic if `tmpVar` gets used elsewhere but doesn't get added to the `tempVarToTree` mapping.  What do you think of this concern?

Nargeshdb: I don't think it makes any problem because each node has to have a unique tmpVar which is defined in MustCallAnnotatedTypeFactory. But the problem here is that in ResourceLeakTransfer in visitTernaryExpression we don't check if there exist a tmpVar for the node like [this](https://github.com/Nargeshdb/checker-framework/blob/fix-5136/checker/src/main/java/org/checkerframework/checker/resourceleak/ResourceLeakTransfer.java#L140).

msridhar: Ok, assuming the tests pass, I am ok with this change for now.  I'll finish reviewing once CI passes (I think there is a formatting error)

msridhar (approved):

PR #5140, opened 2022-05-12
Handle `@MustCall()` like `@MustCall({})`

mernst:


kelloggm (approved):
LGTM. Thanks for the fix!

PR #5162, opened 2022-06-15
Add parameter to control warning about memory constraints

### LINKED ISSUE #5159: -Werror interacts poorly with "Memory constraints are impeding performance; please increase max heap size"
My apologies if this is a duplicate report. I recall discussing it before but have failed to find the prior issue or mailing list thread.

We use `-Werror` for code hygiene, to give careful control over what warnings we allow and which we do not.

Checkerframework sometimes issues a warning "Memory constraints are impeding performance; please increase max heap size." which is upgraded to an error and causes the build to fail.

It is not desirable, from my point of view, for this warning to have the same effect on a build as an illegal use of rawtypes, etc.

The issue apache/beam#20819 tracks this now (migrated from https://issues.apache.org/jira/browse/BEAM-11837 which has additional context). In the past we have fixed this by giving more memory. But realistically we would prefer to have the build succeed even if it is slow for a bit.

Some ideas:

 - a flag to downgrade it to an INFO log (it is still good to have the info in case one is debugging a super slow build)
 - increase the amount of GC thrashing needed to issue the warning
 - some combination of the two, where the INFO log becomes a WARNING log after some number of occurrences

kennknowles:
Please take a look and see what you think about this approach. I did it pretty quick (in `vim` not even opening an IDE or compiling) so I'm leaning on CI and reviewer a bit here. Just wanted to get things started to make my suggestion very concrete.

Fixes #5159


mernst:
@kennknowles Do you prefer your name to be listed as "Kenn Knowles" or "Kenneth Knowles"?

mernst:
I would like to find a better name than `-AnoWarnMemoryConstraints` if possible.

kennknowles:
Yea, fair point. Thanks for just jumping in and fixing it up!

kennknowles:
`-AgcWarningLogLevel=NOTE` could be more legible, but it seems like more of a knob than you need

kennknowles:
Opened plume-lib/plume-scripts#20 in preparation for adjusting contributors list.

kennknowles:
What do you think about these various options, or some other blend of choice of name for the parameter and choice of values (vs flag that does not accept a value)?

 - `-AmemoryPressureLogLevel=<level>`
 - `-AdowngradeMemoryPressureLogLevel`
 - `-AwarnOnMemoryPressure=<boolean>`

kennknowles:
Any thoughts on your preferred naming for the parameter?

smillst (approved):
The code changes look good to me. @mernst you wanted to find a better name.  I haven't thought of any, should we just go with`noWarnMemoryConstraints`?

mernst:
I don't have a better suggestion for a flag name, so let's go with this.
Thanks, @kennknowles , and thanks for your patience also.

kennknowles:
Fab! Thank you!

PR #5163, opened 2022-06-15
Add `-AstubNoWarnIfNotFound`, and stub file `junit-assertions.astub` which is read by default

mernst:


kelloggm (dismissed):
LGTM, except the weirdness in the CHANGELOG. If that's just an error, no need for me to re-review.

Issue #5166, opened 2022-06-17
False positive when comparing two char with the Signedness checker


txiang61:
Signedness checker complains about comparison between two `char`. 
```
private static boolean isWhitespace(char c) {
    return c <= '\u0020';
}
```
Gives the following error:
```
error: [comparison.unsignedlhs] comparison has an unsigned LHS
    return c <= '\u0020';
           ^
```
From JLS https://docs.oracle.com/javase/specs/jls/se9/html/jls-5.html#jls-5.1.2
> A widening conversion of a char to an integral type T zero-extends the representation of the char value to fill the wider format.

Comparison between two `char` should be safe.

mernst:
I agree that the code you gave should pass type-checking.  It seems to do so; see pull request #5167.

What version of the Checker Framework are you using?

txiang61:
Thanks good to know it's already fixed. I'm on a very old version. Will update.

PR #5167, opened 2022-06-17
Add test for issue #5166

mernst:


txiang61:
Thanks good to know it's already fixed. I'm on a very old version. Will update.

PR #5169, opened 2022-06-19
Don't crash on static field accesses

mernst:


msridhar (approved):
This change looks fine to me.  I'm not sure if we've handled all cases; e.g. I think one can write `super.f` (though maybe that is quite rare)

msridhar (approved):

PR #5171, opened 2022-06-21
`-ApermitStaticOwning` suppresses Resource Leak Checker warnings related to static owning fields

mernst:


javacutil/src/main/java/org/checkerframework/javacutil/TreePathUtil.java
```diff
@@ -341,6 +342,51 @@ public static boolean isTreeInStaticScope(TreePath path) {
...
 
+  /**
+   * Returns true if the path is to a top-level (not within a loop) assignment within an initializer
+   * block. Will return true for a re-assignment even if there is another initialization (within
+   * this initializer block, another initializer block, or the variable declaration).
+   *
+   * @param path the path to test
+   * @return true if the path is to an initialization within an initializer block
+   */
+  public static boolean isInitializerInBlock(TreePath path) {
```
msridhar, line 353: Maybe `isTopLevelAssignmentInInitializerBlock` would be a better name?

msridhar (dismissed):
One optional naming comment

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -1248,6 +1255,10 @@ private void checkReassignmentToField(Set<Obligation> obligations, AssignmentNod
       // also a declaration must be a field initializer.
       if (node.getTree().getKind() == Tree.Kind.VARIABLE) {
         return;
+      } else if (TreePathUtil.isInitializerInBlock(currentPath)) {
+        // This is likely not reassignment; if reassignment, the number of assignments that
+        // were not warned about is limited to other initializations (is not unbounded).
```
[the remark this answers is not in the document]
kelloggm: After re-reading the new method in `TreePathUtil` being used here, it doesn't seem to distinguish static and non-static initializer blocks, so I think this is a valid concern.

checker/src/main/java/org/checkerframework/checker/resourceleak/ResourceLeakVisitor.java
```diff
@@ -33,6 +34,9 @@
  */
 public class ResourceLeakVisitor extends CalledMethodsVisitor {
 
+  /** True if errors related to static owning fields should be suppressed. */
+  boolean permitStaticOwning;
```
kelloggm, line 38: When I saw that the similarly-named field in `MustCallConsistencyAnalyzer` was package-private, I assumed that the reason was that this class and that one would share a single field value. That doesn't appear to be the case, so can these two fields both be made `private` or `protected`?

kelloggm (changes requested):

mernst:
@kelloggm This is ready for re-review.  Thanks!

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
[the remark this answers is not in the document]
mernst: You are right.  This was a thing I wanted feedback on:  whether this unsoundness is acceptable.  I have put it under a different command-line argument.

docs/manual/resource-leak-checker.tex
```diff
@@ -381,6 +407,7 @@
 
 % TODO: should this section also include text about unconnected sockets, or is what's here sufficient?
 
+
```
kelloggm, line 410: Unnecessary whitespace change.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -339,10 +339,13 @@ public void setRoot(CompilationUnitTree root) {
 
   @Override
   public Void scan(@Nullable Tree tree, Void p) {
-    if (tree != null && getCurrentPath() != null) {
+    if (tree == null) {
```
kelloggm, line 342: What is the purpose of this change? I don't see any obvious relation to the purpose of this PR.

docs/manual/resource-leak-checker.tex
```diff
@@ -50,8 +50,16 @@
 \end{Verbatim}
 
 The Resource Leak Checker supports all the command-line arguments
- listed in Section~\ref{called-methods-run-checker} for
-the Called Methods Checker.
+listed in Section~\ref{called-methods-run-checker} for
+the Called Methods Checker, plus two others:
```
kelloggm, line 54: One of these is commented out, but I think this is okay as long as we merge #5173 soon after we merge this (but, the PR description should have a note about that).

dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java
```diff
@@ -394,7 +394,7 @@ public S getStoreAfter(Block block) {
    */
   protected @Nullable S runAnalysisFor(Node node, Analysis.BeforeOrAfter preOrPost) {
     Block block = node.getBlock();
-    assert block != null : "@AssumeAssertion(nullness): invariant";
+    assert block != null : "@AssumeAssertion(nullness): null block for node " + node;
```
kelloggm, line 397: This seems like an unrelated (but fine?) change.

kelloggm (changes requested):

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
mernst: There are three statements in the body.
Two of them check for `tree` being null, but the third does not.  However, it (the call to `super.scan`) also crashes when `tree` is null.  This prevents that crash.

I'm not sure when or why `tree` can be null, but this does prevent the crash.

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -1022,6 +1026,9 @@ private void updateObligationsForAssignment(
       Set<Obligation> obligations, AssignmentNode assignmentNode) {
     Node lhs = assignmentNode.getTarget();
     Element lhsElement = TreeUtils.elementFromTree(lhs.getTree());
+    if (lhsElement == null) {
```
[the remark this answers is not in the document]
mernst: The tests pass without this, so I have removed it.  It must have been triggered by an incorrect intermediate code change.  Thanks for pushing me on this.

kelloggm (approved):

PR #5173, opened 2022-06-24
Add `-ApermitInitializationLeak` command-line argument

mernst:


checker/src/main/java/org/checkerframework/checker/resourceleak/ResourceLeakChecker.java
```diff
@@ -16,6 +16,7 @@
  * via the {@link org.checkerframework.checker.mustcall.MustCallChecker} have been fulfilled.
  */
 @SupportedOptions({
+  "permitInitializationLeak",
```
kelloggm, line 19: For consistency, I'd also make this is a static final constant.

checker/tests/resourceleak-permitinitializationleak/InstanceInitializer.java
```diff
@@ -0,0 +1,47 @@
+// A test that the checker is sound in the presence of instance initializer blocks.
```
kelloggm, line 1: A comment in this test should call out exactly where and how the unsoundness occurs.

dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java
```diff
@@ -393,8 +393,9 @@ public S getStoreAfter(Block block) {
    *     running the analysis
    */
   protected @Nullable S runAnalysisFor(Node node, Analysis.BeforeOrAfter preOrPost) {
+    // block is null if node is a formal parameter of a method, or is a field access thereof
     Block block = node.getBlock();
-    assert block != null : "@AssumeAssertion(nullness): invariant";
+    assert block != null : "@AssumeAssertion(nullness): null block for node " + node;
```
kelloggm, line 398: This change seems unrelated to this PR.

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -462,6 +465,7 @@ public int hashCode() {
     this.typeFactory = typeFactory;
     this.checker = (ResourceLeakChecker) typeFactory.getChecker();
     this.analysis = analysis;
+    this.permitInitializationLeak = checker.hasOption("permitInitializationLeak");
```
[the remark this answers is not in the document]
mernst: What is the point of using a static final constant?  From my point of view, it:
 * adds indirection
 * adds clutter
 * makes it harder to search for all uses of the flag in the code (because the static constant name is spelled differently from the command-line switch itself)
 * gives no help if we ever want to change the name.
    * it is trivial to do a global search
    * we must do a global search anyway, to find occurrences in comments and in the manual

I don't see any benefits to using a static final constant.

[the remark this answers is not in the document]
kelloggm: > makes it harder to search for all uses of the flag in the code

If you're using an IDE, it should actually make it easier - a global search for uses of the static final constant will return only the places where it is used in the code.

Regardless, this is a matter of style/preference. Since you wrote this code, I'm fine with it.

kelloggm (approved):

PR #5177, opened 2022-06-25
Don't crash when a left-hand side contains a cast

mernst:


javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -218,6 +218,23 @@ public static boolean isSelfAccess(final ExpressionTree tree) {
     return t;
   }
 
+  /**
+   * If the given tree is a parenthesized tree or cast tree, return the enclosed non-parenthesized,
```
mernst, line 222: Perhaps, rather than introducing this new method, `withoutParens` should also strip away casts.

kelloggm (approved):

PR #5178, opened 2022-06-27
Copy maps lazily in `AnalysisResult`

### LINKED ISSUE #5172: Creating `AnalysisResult` objects is expensive due to `IdentityHashMap` copies
I've been looking into profiling NullAway's dataflow analysis more carefully with a micro-benchmark to find unnecessary overheads.  I created the following synthetic benchmark (may be useful for Nullness Checker as well), and I profiled repeatedly running NullAway on the benchmark in a loop:

https://github.com/msridhar/NullAway/blob/manu-dflow-bench-test/nullaway/src/test/resources/com/uber/nullaway/testdata/DFlowBench.java

I observed that on this benchmark, nearly half the execution time is spent creating `IdentityHashMap` objects, as part of constructing the final `AnalysisResult` after dataflow analysis completes.  There are several `IdentityHashMap` fields inside `AnalysisResult`.  I see that the contents of the maps get mutated inside the `AnalysisResult.combine()` method, but not elsewhere that I can see (it's possible subclasses are allowed to mutate the contents).  Depending on the intended invariants, I am wondering if we could rewrite this code to lazily create copies of `IdentityHashMap`s (like only when combining results) rather than doing it eagerly, maybe using `Collections.unmodifiableMap()` to catch any unintended mutation without copying entire maps.  I'm happy to put up a PR if we can agree on an approach.

msridhar:
Fixes #5172

This introduces lazy copies for certain maps stored in `AnalysisResult`; such copies can be expensive for analysis of large methods.  I have confirmed that this change enables NullAway to process the micro-benchmark linked from #5172 much more quickly (about 2X faster!).  I'm unsure if the change will make much difference for the Checker Framework checkers, as it looks like the `AnalysisResult.combine()` method is invoked as part of analyzing each class (see [here](https://github.com/typetools/checker-framework/blob/23efed29e7612dfe45cc20a4445d0a7c417d3cdf/framework/src/main/java/org/checkerframework/framework/type/GenericAnnotatedTypeFactory.java#L1511)).

For safety, before the maps are copied, they are wrapped in a new `UnmodifiableIdentityHashMap` class, to introduce runtime failures for unexpected copies.

I deliberately made this change "minimal," only removing the costly copies.  There are maps in `AnalysisResult` that were not copied at all before, and I did not make any changes to such code, as I'm not sure of all the invariants involved.  I'm open to making more changes here if this is a good opportunity to do so.

### DIFF
--- a/dataflow/src/main/java/org/checkerframework/dataflow/util/UnmodifiableIdentityHashMap.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/util/UnmodifiableIdentityHashMap.java
@@ -25,6 +25,11 @@ public class UnmodifiableIdentityHashMap<K, V> extends IdentityHashMap<K, V> {
   /** the wrapped map */
   private final IdentityHashMap<K, V> map;
 
+  /**
+   * Create an UnmodifiableIdentityHashMap
+   *
+   * @param map the map to wrap
+   */
   private UnmodifiableIdentityHashMap(IdentityHashMap<K, V> map) {
     this.map = map;
   }
--- a/dataflow/src/main/java/org/checkerframework/dataflow/util/UnmodifiableIdentityHashMap.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/util/UnmodifiableIdentityHashMap.java
@@ -129,6 +129,7 @@ public class UnmodifiableIdentityHashMap<K, V> extends IdentityHashMap<K, V> {
     return Collections.unmodifiableSet(map.entrySet());
   }
 
+  // `action` has no side effects on the map, because it is only passed keys and values.
   @Override
   public void forEach(BiConsumer<? super K, ? super V> action) {
     map.forEach(action);
--- a/dataflow/src/main/java/org/checkerframework/dataflow/util/UnmodifiableIdentityHashMap.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/util/UnmodifiableIdentityHashMap.java
@@ -23,6 +23,7 @@ import org.checkerframework.checker.nullness.qual.PolyNull;
 @SuppressWarnings("keyfor") // Keys for `this` are also keys for `this.map`
 public class UnmodifiableIdentityHashMap<K, V> extends IdentityHashMap<K, V> {
 
+  /** The serial version UID. */
   private static final long serialVersionUID = -5147442142854693854L;
 
   /** The wrapped map. */
--- a/dataflow/src/main/java/org/checkerframework/dataflow/util/UnmodifiableIdentityHashMap.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/util/UnmodifiableIdentityHashMap.java
@@ -38,7 +38,8 @@ public class UnmodifiableIdentityHashMap<K, V> extends IdentityHashMap<K, V> {
   }
 
   /**
-   * Create an {@link UnmodifiableIdentityHashMap} wrapper for a map.
+   * Create an {@link UnmodifiableIdentityHashMap} wrapper for a map. Returns the argument if it is
+   * already an {@link UnmodifiableIdentityHashMap}.
    *
    * @param map the map to wrap
    * @return the wrapper
--- a/dataflow/src/main/java/org/checkerframework/dataflow/util/UnmodifiableIdentityHashMap.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/util/UnmodifiableIdentityHashMap.java
@@ -111,11 +111,6 @@ public class UnmodifiableIdentityHashMap<K, V> extends IdentityHashMap<K, V> {
   }
 
   @Override
-  public Object clone() {
-    throw new UnsupportedOperationException();
-  }
-
-  @Override
   public Set<K> keySet() {
     return Collections.unmodifiableSet(map.keySet());
   }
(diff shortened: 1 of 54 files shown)


dataflow/src/main/java/org/checkerframework/dataflow/util/UnmodifiableIdentityHashMap.java
```diff
@@ -0,0 +1,191 @@
...
+/**
+ * A wrapper around an {@link IdentityHashMap} that makes it unmodifiable. All mutating operations
+ * throw {@link UnsupportedOperationException}, and all other operations delegating to the
+ * underlying map.
+ *
+ * <p>This class extends {@link IdentityHashMap} only so it is assignable to variables / fields of
+ * static type {@link IdentityHashMap}. All valid operations are delegated to the wrapped map, and
+ * any inherited state from the superclass is unused.
+ */
+@SuppressWarnings("nullness")
```
msridhar, line 22: I tried to add nullness annotations to this type but got confused.  Here are a couple of the CI errors I got:
```
/__w/1/s/dataflow/src/main/java/org/checkerframework/dataflow/util/UnmodifiableIdentityHashMap.java:71: error: [keyfor:contracts.conditional.postcondition] conditional postcondition is not satisfied when containsKey returns true.
    return map.containsKey(key);
    ^
  found   : key is @KeyFor("this.map")
  required: key is @KeyFor("this")
/__w/1/s/dataflow/src/main/java/org/checkerframework/dataflow/util/UnmodifiableIdentityHashMap.java:101: error: [nullness:override.receiver] Incompatible receiver type
  public boolean equals(@Nullable Object o) {
                 ^
  found   : @Initialized @NonNull UnmodifiableIdentityHashMap<K extends @Initialized @Nullable Object, V extends @Initialized @Nullable Object>
  required: @Initialized @Nullable IdentityHashMap<K extends @Initialized @Nullable Object, V extends @Initialized @Nullable Object>
  Consequence: method in @Initialized @NonNull UnmodifiableIdentityHashMap<K extends @Initialized @Nullable Object, V extends @Initialized @Nullable Object>
    @Initialized @NonNull boolean equals(@Initialized @NonNull UnmodifiableIdentityHashMap<K extends @Initialized @Nullable Object, V extends @Initialized @Nullable Object> this, @Initialized @Nullable Object p0)
  cannot override method in @Initialized @NonNull IdentityHashMap<K extends @Initialized @Nullable Object, V extends @Initialized @Nullable Object>
    @Initialized @NonNull boolean equals(@Initialized @Nullable IdentityHashMap<K extends @Initialized @Nullable Object, V extends @Initialized @Nullable Object> this, @Initialized @NonNull Object p0)
/__w/1/s/dataflow/src/main/java/org/checkerframework/dataflow/util/UnmodifiableIdentityHashMap.java:102: error: [nullness:argument] incompatible argument for parameter o of equals.
```
Here is the build:
https://dev.azure.com/checkerframework/checkerframework/_build/results?buildId=8760&view=logs&jobId=794427f4-26d8-57f0-fc22-31159cf0c9b8&j=794427f4-26d8-57f0-fc22-31159cf0c9b8&t=83f7d39a-b612-5a99-9aa3-cd2517324f7f

If we want to fully annotate this wrapper type for nullness, I could use some help.  Otherwise, we can add a comment to the suppression.

dataflow/src/main/java/org/checkerframework/dataflow/util/UnmodifiableIdentityHashMap.java
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  @Override
+  public Set<Map.Entry<K, V>> entrySet() {
+    return Collections.unmodifiableSet(map.entrySet());
+  }
+
+  @Override
+  public void forEach(BiConsumer<? super K, ? super V> action) {
+    map.forEach(action);
```
mernst, line 134: What if `action` performs side effects?  Is it up to the client to ensure that it does not?

mernst (changes requested):

dataflow/src/main/java/org/checkerframework/dataflow/util/UnmodifiableIdentityHashMap.java
msridhar: I think `action` is only passed the keys and values contained in the map, but not a direct reference to the wrapped map; so it should not be able to perform unwanted side effects on the map.  Did I miss something?

mernst: You are right.  Thanks for clarifying.

dataflow/src/main/java/org/checkerframework/dataflow/util/UnmodifiableIdentityHashMap.java
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    return map.hashCode();
+  }
+
+  @Override
+  public Object clone() {
+    throw new UnsupportedOperationException();
```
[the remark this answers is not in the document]
msridhar: I think what I will do here is just not override `clone()`.  With the inherited behavior, the cloned object will wrap the same underlying map, but I don't think that will allow any unexpected mutation.  Good catch!

mernst (approved):
Thanks!

mernst (approved):

PR #5179, opened 2022-06-27
permit calling @CreatesMustCallFor(this) methods in constructors

### LINKED ISSUE #5175: Resource Leak Checker: calling `@CreatesMustCallFor("this")` method from constructor
Consider the following code, which appears in `checker/tests/resourceleak/LemmaStack.java`:

```java
// The Resource Leak Checker issues the following error:
// LemmaStack.java:37: error: [reset.not.owning] Calling method startProver resets the must-call
// obligations of the expression this, which is non-owning. Either annotate its declaration with an
// @Owning annotation, extract it into a local variable, or write a corresponding
// @CreatesMustCallFor annotation on the method that encloses this statement.
//     startProver();
//                ^
// 1 error

import java.io.Closeable;
import java.io.IOException;
import java.io.PrintWriter;
import org.checkerframework.checker.mustcall.qual.CreatesMustCallFor;
import org.checkerframework.checker.mustcall.qual.MustCall;
import org.checkerframework.checker.nullness.qual.EnsuresNonNull;

@MustCall("close") public class LemmaStack implements Closeable {

  private @Owning @MustCall("close") PrintWriter session;

  @CreatesMustCallFor("this")
  @EnsuresNonNull("session")
  private void startProver() {
    try {
      if (session != null) {
        session.close();
      }
      session = new PrintWriter("filename.txt");
    } catch (IOException e) {
      throw new Error(e);
    }
  }

  public LemmaStack() {
    startProver();
  }

  @EnsuresCalledMethods(value = "session", methods = "close")
  @Override
  public void close(LemmaStack this) {
    session.close();
  }
}
```

None of the three suggestions in the error messages works.
 * annotate its declaration with an `@Owning` annotation:  there is no declaration of `this` to annotate
 * extract it into a local variable :  `@Owning` cannot be written on a local variable
 * write a corresponding `@CreatesMustCallFor` annotation on the method that encloses this statement:  that does not work for the constructor.

The message should be improved to only make suggestions that are possible.
Furthermore, this false negative warning should be eliminated.

kelloggm:
fixes #5175

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
+++ b/checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
@@ -562,8 +562,12 @@ class MustCallConsistencyAnalyzer {
    * Checks that an invocation of a CreatesMustCallFor method is valid.
    *
    * <p>Such an invocation is valid if any of the conditions in {@link
-   * #isValidCreatesMustCallForExpression(Set, JavaExpression, TreePath)} is true. If none of these
-   * conditions are true, this method issues a reset.not.owning error.
+   * #isValidCreatesMustCallForExpression(Set, JavaExpression, TreePath)} is true for each
+   * expression in the argument to the CreatesMustCallFor annotation. As a special case, the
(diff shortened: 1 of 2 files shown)


msridhar (approved):
LGTM!

mernst (approved):
Thanks!

PR #5180, opened 2022-06-27
If running Resource Leak Checker, no need to run Must Call Checker

mernst:


kelloggm (approved):
I added one word for clarity. LGTM

PR #5182, opened 2022-06-27
WPI: avoid crash when inferring facts about fields defined in other compilation units

kelloggm:
Only occurs for ajava-based WPI. Similar to a problem we encountered when enabling the all-systems tests for WPI, but in that case we only had a test for methods. None of the tests in this PR elicit the bug, but they cover cases we didn't have in the ainfer-testchecker suite before so I figured it would be fine to include them; I'm not sure what, exactly, causes an element to "be in a different compilation unit", so I struggled to write a test for this. However, the change in this PR does prevent the crash @Nargeshdb encountered when running the RLC with ajava-based WPI on HBase.


mernst (dismissed):
LGTM

PR #5183, opened 2022-06-28
add error when writing an explicit @MustCall annotation on a non-final class declaration (@InheritableMustCall should be used instead)

### LINKED ISSUE #5181: Resource Leak Checker: warn of `@MustCall` on class declaration
Usually, writing `@MustCall` on a class declaration is the wrong thing.  Instead, one should write `@InteritableMustCall`.  However, it is an easy error to make.

I suggest that writing `@MustCall` on a class declaration should result in a warning or errror, but one that can be suppressed.

kelloggm:
fixes #5181


mernst (dismissed):
Thanks!

kelloggm:
@mernst, I had to make some small changes to get CI passing, so your approval has been marked stale. If you re-approve, it should merge automatically.

mernst (approved):

PR #5188, opened 2022-06-30
Ajava-based WPI should not print invisible qualifiers

kelloggm:
Also adds `@InvisibleQualifier` annotations to the top annotations from some hierarchies that didn't have it (I only checked the ones used by ICalAvailable, which was the program I was testing on).

The goal is to try to get ajava-based WPI closer to the stated specification of WPI wrt invisible qualifiers and defaults, which says:
```
 * When outputting a file, WPI ignores an inferred type if:
 *
 * <ol>
 *   <li>The @Target annotation does not permit the annotation to be written at this location.
 *   <li>The @RelevantJavaTypes annotation does not permit the annotation to be written at this
 *       location.
 *   <li>The inferred annotation has the @InvisibleQualifier meta-annotation.
 *   <li>The inferred annotation would be the same annotation applied via defaulting &mdash; that
 *       is, if omitting it has the same effect as writing it.
 * </ol>
```
(from `WholeProgramInferenceImplementation.java`)

In practice, WPI doesn't reach this goal; astub/jaif-based WPI make an attempt via `WholeProgramInferenceScenesStorage#shouldIgnore`, but that method has its own problems and has a note about how it shouldn't be edited, but rather ought to be replaced.

I considered attempting a general solution along the lines suggested there, but couldn't get one working quickly. This PR is a bit of a kludge to add (some of) the same functionality to ajava-based WPI - it prevents annotations with `@InvisibleQualifier` annotations from being printed at the output level. It's worth having a discussion about how this problem ought to be solved long-term, however.


mernst (commented):
Thanks for doing this.  I have one suggestion, then you can merge the pull request.

mernst (approved):

kelloggm:
@mernst your previous review was marked stale because I factored out the collection of the names of the invisible qualifiers, as you suggested, into a new static method. If you approve this again, it'll be merged automatically. Thanks!

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
```diff
@@ -105,6 +114,23 @@
...
+   *
+   * @param atypeFactory a type factory
+   * @return the names of every invisible qualifier supported by {@code atypeFactory}
+   */
+  public static Set<String> getInvisibleQualifierNames(AnnotatedTypeFactory atypeFactory) {
+    return atypeFactory.getSupportedTypeQualifiers().stream()
+        .filter(
+            qual ->
+                Arrays.stream(qual.getAnnotations())
+                    .anyMatch(anno -> anno.annotationType() == InvisibleQualifier.class))
```
mernst, line 129: My suggestion was to abstract out `isInvisible`, which might be useful in another context.
But, this is OK if you prefer it.

mernst (approved):

PR #5191, opened 2022-07-06
Add failing test case for subclassing lombok builders

Meijuh:
This test case shows that with checker framework's builder checker you can not easily extend generated builders without running  into a `finalizer.invocation` error.
To be fair, I have not run this test case, because I am unsure how to do this exactly, but it should be quite close to something that is reproducible. The error, specifically, is of the form
`[finalizer.invocation] This finalizer cannot be invoked, because the following methods have not been called: attribute()
      final LombokBuilderSubclassExample result = super.build();`

Maybe the solution (for this example) is to annotate the generated method `LBSEBuilder#build()` with `@CalledMethods("attribute")`, but I am not entirely sure.

### DIFF
--- a/checker/tests/calledmethods-lombok/LombokBuilderSubclassExample.java
+++ b/checker/tests/calledmethods-lombok/LombokBuilderSubclassExample.java
@@ -1,0 +1,29 @@
+import lombok.Builder;
+import lombok.NonNull;
+import lombok.Value;
+
+@Builder(builderClassName = "LBSEBuilder")
+@Value public class LombokBuilderSubclassExample {
+
+  @NonNull Integer attribute;
+
+  public static LombokBuilderSubclassExample builder() {
+    return new LombokBuilderSubclassExampleBuilder();
+  }
+
+  public static class LombokBuilderSubclassExampleBuilder extends LBSEBuilder {
+
+    @Override
+    public LombokBuilderSubclassExample build() {
+      final LombokBuilderSubclassExample result = super.build();
+      // here result.getAttribute() is guaranteed to be non null, so we do not have to check this
+      // ourselves
+
+      if (result.getAttribute() < 0) {
+        throw new IllegalArgumentException("attribute must be >= 0");
+      }
+
+      return result;
+    }
+  }
+}


kelloggm:
> Maybe the solution (for this example) is to annotate the generated method `LBSEBuilder#build()` with `@CalledMethods("attribute")`, but I am not entirely sure.

In fact, the warning is caused because an annotation is inferred for that generated method: its receiver has that annotation automatically. That's why you get a `finalizer.invocation` error at the call to `super.build()`: calling the `LBSEBuilder#build()` requires the receiver (in this case, `this`) to have had `attribute` called on it already.

In this case I think the solution you want is to add a similar annotation to your subclass's `build` method - effectively, require that `attribute` be called before a call to the subclass's `build` method, too:

```
    @Override
    public LombokBuilderSubclassExample build(
        @CalledMethods("attribute") LombokBuilderSubclassExampleBuilder this) {
      final LombokBuilderSubclassExample result = super.build();
```

I modified the test case that you provided on [this branch](https://github.com/kelloggm/checker-framework/tree/Meijuh/master) as I described above and confirmed that the test case passes.

Is there any way we could have improved the documentation in order to make it more clear how to handle this?

Meijuh:
Thanks so much. That seems to work, although I had to add an extra overridden method to make the subclassed builder useful. What do you think of this addition?
I now see it is indeed in the documentation at https://checkerframework.org/manual/#called-methods-spec. It uses a fancy language feature I have never seen. I could have read it better. However, maybe the documentation can be complemented to include the extension in https://github.com/typetools/checker-framework/pull/5191/commits/2e44bb20b2400fe65c04ab3870c0f449d42c87af?
Naturally, this PR now also serves as documentation for lucky search engine users.

kelloggm:
> That seems to work, although I had to add an extra overridden method to make the subclassed builder useful. What do you think of this addition?

I'm not surprised that was necessary to make it actually useful: those are the equivalents of the annotations that the framework automatically infers for the generated method in the builder that is being extended here.

> I now see it is indeed in the documentation at https://checkerframework.org/manual/#called-methods-spec. It uses a fancy language feature I have never seen. I could have read it better

Sorry it was confusing! By "fancy language feature", are you referring to the use of an explicit receiver parameter (i.e., the explicit `this` as a method argument)? If so, unfortunately, I'm not sure how that can be avoided, since that is unfortunately where the annotation needs to be placed.

> However, maybe the documentation can be complemented to include the extension in https://github.com/typetools/checker-framework/commit/2e44bb20b2400fe65c04ab3870c0f449d42c87af?

That's a good idea! You could add a suggestion in this PR by editing the file `docs/manual/called-methods-checker.tex`, or I can write it up later.

I'd like to merge this PR with the test case (as you said, so it can help serve as documentation for users who run into this problem in the future). However, it looks like CI is failing for two reasons:
* there is a missing symbol in the new test case (a missing import?): `LombokBuilderSubclassExample.java:17: error: cannot find symbol`
* you haven't yet contributed to the Checker Framework, and we'd like to give you appropriate credit:
   ```
@@ -45,6 +45,7 @@ Jason Waataja,
 Javier Thaine,
 Jeff Luo,
 Jenny Xiang,
+Jeroen Meijer,
 Jianchu Li,
 Jiangqi Zhang,
 Jiasen (Jason) Xu,
docs/manual/contributors.tex is not up to date.
If the above suggestion is appropriate, run: make -C docs/manual contributors.tex
```

If you follow that suggestion (and fix that missing symbol), CI should pass.

Thanks again!

mernst:
@Meijuh A ping about this pull request.

Meijuh:
I added the suggested edits. I have to say I now also understand the subtyping relationship described in `@CalledMethods` (which I did not before and was the original reason I asked the question.

PR #5210, opened 2022-07-14
Add more map copying to AnalysisResult

msridhar:
This fixes a crash in NullAway discovered when trying to update to 3.23.0; see https://github.com/uber/NullAway/issues/623.  It is a follow-up to https://github.com/typetools/checker-framework/pull/5178.  In that change I missed a case where the maps inside `AnalysisResult` need to be copied (see code comments for details).

I am open to suggestions on how to add a test for this.  I think the Checker Framework tests passed before because Checker Framework checkers almost always end up calling `AnalysisResult.combine()`, which ends up copying the maps anyway; NullAway does not call this method.

### DIFF
--- a/dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java
@@ -151,7 +151,7 @@ public class AnalysisResult<V extends AbstractValue<V>, S extends Store<S>> impl
     finalLocalValues.putAll(other.finalLocalValues);
   }
 
-  /** Make copies of certain internal IdentityHashMaps, if they have not been copied already */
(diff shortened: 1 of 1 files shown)


msridhar:
Actually, there may be a better fix for this, if we change the code here:

https://github.com/typetools/checker-framework/blob/f93103e5f82737720280d0e3c0e65f753ad2ab52/dataflow/src/main/java/org/checkerframework/dataflow/analysis/AbstractAnalysis.java#L217-L221

This is the code that clears the internal `nodeValues` map, which causes problems in the `AnalysisResult`.  I wonder if instead of calling `clear()` on the `nodeValues` map, this code could just do `nodeValues = new IdentityHashMap<>(in)`?  This seems a bit cleaner to me than my initial approach, and it also does less copying.  The downside would be that `nodeValues` could no longer be final, but that is only providing shallow immutability anyway.  Thoughts @smillst @mernst?

mernst (approved):

PR #5215, opened 2022-07-17
Add a dataflow test for constant propagation

msridhar:
Currently we only have dataflow tests for backward analysis; constant propagation covers the forward analysis case.  The code is based heavily on the existing dataflow tests.

This was originally written in the hopes of having a test exposing the issue addressed by #5210.  It doesn't do that, unfortunately (that failure relies on more complex transfer functions), but I think it's still useful.

### DIFF
--- a/dataflow/src/main/java/org/checkerframework/dataflow/constantpropagation/ConstantPropagationStore.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/constantpropagation/ConstantPropagationStore.java
@@ -168,6 +168,6 @@ public class ConstantPropagationStore implements Store<ConstantPropagationStore>
   @Override
   @SuppressWarnings("nullness")
   public String visualize(CFGVisualizer<?, ConstantPropagationStore, ?> viz) {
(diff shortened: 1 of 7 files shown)


dataflow/src/main/java/org/checkerframework/dataflow/constantpropagation/ConstantPropagationStore.java
```diff
@@ -168,6 +168,6 @@ public boolean canAlias(JavaExpression a, JavaExpression b) {
   @Override
   @SuppressWarnings("nullness")
   public String visualize(CFGVisualizer<?, ConstantPropagationStore, ?> viz) {
-    return viz.visualizeStoreKeyVal("constant propagation", null);
+    return viz.visualizeStoreKeyVal("constant propagation", toString());
```
msridhar, line 171: Before, this method would not give any useful information about the constant propagation store

mernst:
Thanks!

msridhar:
@mernst I think you just need to review this and approve it for it to be merged

mernst (approved):

PR #5218, opened 2022-07-22
HashSets are usually small

mernst:


kelloggm (approved):
This obviously will improve performance. Thanks for finding it!

PR #5221, opened 2022-07-24
Make reflection more efficient

mernst:


javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -109,6 +110,42 @@ private TreeUtils() {
...
+        Class<?> bindingPatternClass = Class.forName("com.sun.source.tree.BindingPatternTree");
+        bindingPatternGetVariable = bindingPatternClass.getMethod("getVariable");
+        instanceOfGetPattern = InstanceOfTree.class.getMethod("getPattern");
+        Class<?> switchExpressionClass = Class.forName("com.sun.source.tree.SwitchExpressionTree");
+        switchExpressionGetExpression = switchExpressionClass.getMethod("getExpression");
+        switchExpressionGetCases = switchExpressionClass.getMethod("getCases");
+        Class<?> yieldTreeClass = Class.forName("com.sun.source.tree.YieldTree");
+        yieldGetValue = yieldTreeClass.getMethod("getValue");
+      } catch (ClassNotFoundException | NoSuchMethodException e) {
+        throw new BugInCF("JDK 17+ reflection problem", e);
```
kelloggm, line 144: This exception message is confusing, because (given the test above against JDK version 12), a user could get this message when running on a JDK whose version is in the range 12-16. I know we don't **technically** support those versions, but I think the error message should reflect the actual conditions under which this error might occur.

kelloggm (approved):
Once the change I suggested is made, this can be merged without another review.

PR #5231, opened 2022-07-29
Handle nested variable declarations

mernst:


framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -2057,38 +2057,48 @@ public boolean shouldSuppressWarnings(@Nullable TreePath path, String errKey) {
...
-      if (shouldSuppressWarnings(elt, errKey)) {
-        return true;
-      }
+          if (isAnnotatedForThisCheckerOrUpstreamChecker(elt)) {
+            // Return false immediately. Do NOT check for AnnotatedFor in the enclosing elements,
+            // because they may not have an @AnnotatedFor.
+            return false;
+          }
+          break;
+        default:
```
kelloggm, line 2089: I think the reason that the `class` case is in `default` rather than in an explicit `case` is that there are multiple kinds of "class" `Tree`s (i.e., the elements of `TreePathUtil#classTreeKinds`). I think we're going to end up with bad style here no matter what we do:
* if we leave it as-is, the use of `default` will lead to any new or unexpected kinds of declarations (for example, that are added in future versions of Java) being treated as classes, which might cause a difficult-to-debug error years down the line
* if we make the class cases explicit, we'll need to duplicate the elements of `TreePathUtil#classTreeKinds` in the source code here. That would be undesirable, for a similar reason to the above: if the list of "class-like" kinds ever changes, it'll be a pain to debug.

What about using a set of `if`/`else` statements instead? It would be a bit verbose, but then we could directly test membership in `TreePathUtil#classTreeKinds` and throw a `BugInCF` if we encounter something unexpected.

kelloggm (approved):

PR #5253, opened 2022-08-09
Improve variable names and comments

mernst:
Merge after #5252.

PR #5254, opened 2022-08-09
Make `mustcall.not.inheritable` a warning rather than an error

mernst:


msridhar:
I think the code changes here were also included in #5253?  Should they be removed from that other PR?

mernst:
I'll close this.  It was in case you didn't want it in that other PR.

PR #5262, opened 2022-08-13
Augment documentation of `@Ensurescalledmethods`

mernst:


msridhar (dismissed):
This is definitely good to document

checker-qual/src/main/java/org/checkerframework/checker/calledmethods/qual/EnsuresCalledMethods.java
```diff
@@ -18,6 +18,14 @@
  *
  * <p>This method guarantees that {@code t.m()} is always called before the method returns.
  *
+ * <p>If a class has any {@code @}{@link org.checkerframework.checker.mustcall.qual.Owning Owning}
+ * fields, then its {@code close()} method should be annotated to close them:
```
kelloggm, line 22: Rather than referring to `close()` specifically, this should refer to the class' `@MustCall` methods. `close()` is the most common example, but an owning filed might also have its must-call obligations satisfied by some other method named in the class' `@MustCall` annotation - for example, I've seen classes whose `@MustCall` method is named `shutdown()` or `stop()` or the like. How about the following phrasing instead:
```
 * <p>If a class has any {@code @}{@link org.checkerframework.checker.mustcall.qual.Owning Owning}
 * fields, then its one or more of its must-call methods (i.e., the methods named by the {@code @}{@link org.checkerframework.checker.mustcall.qual.MustCall} or {@code @}{@link org.checkerframework.checker.mustcall.qual.InheritableMustCall} annotation on the class declaration, such as {@code close()}) should be annotated to indicate that the must-call obligations are satisfied:
```

kelloggm (changes requested):

kelloggm (approved):

PR #5263, opened 2022-08-14
Fix issue with multiple identical Nodes in MustCallConsistencyAnalyzer

msridhar:
The caches in `MustCallConsistencyAnalyzer` must be `IdentityHashMap`s to avoid caching the wrong result for a `Node`.

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
+++ b/checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
@@ -16,6 +16,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Deque;
 import java.util.HashSet;
+import java.util.IdentityHashMap;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


msridhar:
CI is failing because of this suppression in Daikon, which now can be removed:

https://github.com/codespecs/daikon/blob/725f24d4b2f2ed7e38c110c20e1919a468662c9d/java/daikon/chicory/Instrument.java#L194-L196

I'm not sure of the exact set of steps to get CI passing.  @mernst Do we patch Daikon first?

msridhar:
Opened https://github.com/codespecs/daikon/pull/416 to remove the Daikon suppression

kelloggm (approved):
This was nasty! Thanks for figuring it out.

I'm concerned we might have a similar bug lurking in our tempvar construction process. In particular, ResourceLeakAnnotatedTypeFactory has a map keyed on LocalVariableNode, but I think that might be okay, because those nodes are all artificial and therefore different? The other possible point of concern is the main tempvar map in MustCallAnnotatedTypeFactory, which is keyed on Tree. Do we think Tree might have a similar failure mode if a method has two identical tempvar-creating expressions?

msridhar:
> I'm concerned we might have a similar bug lurking in our tempvar construction process. In particular, ResourceLeakAnnotatedTypeFactory has a map keyed on LocalVariableNode, but I think that might be okay, because those nodes are all artificial and therefore different?

I _think_ this case is ok, because as you note the nodes are artificial.  But this also assumes that it is ok to compare `Tree`s based on `.equals()` since the map is bi-directional (see below).

> The other possible point of concern is the main tempvar map in MustCallAnnotatedTypeFactory, which is keyed on Tree. Do we think Tree might have a similar failure mode if a method has two identical tempvar-creating expressions?

From the javac implementation, it seems that `JCTree` implementations do not override `equals()` and therefore are compared based on `==`.  So I think this should be safe, assuming we don't have our own artificial `Tree` implementations that override `equals()` and do something different.  I did notice there is an `IdentityHashMap` in CF code where the keys are `Tree`s:

https://github.com/typetools/checker-framework/blob/f1ace3aef2e0db7c71c7a088c47bec8494636fd5/dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java#L48

Maybe this was just defensive coding that is in fact unnecessary; not sure.

msridhar:
@mernst assigning to you to advise on how to deal with the CI failure on Daikon

mernst:
Regarding CI, you have two choices:
 * merge and let me deal with it.
 * Make a fork of Daikon in msridhar, create a branch named exactly resource-leak-identical-returns, and make changes in that branch.  Then re-run the failing CI jobs.

mernst:
@msridhar Thanks for fixing this!  That is great.

msridhar:
> Regarding CI, you have two choices:
> 
> * merge and let me deal with it.
> * Make a fork of Daikon in msridhar, create a branch named exactly resource-leak-identical-returns, and make changes in that branch.  Then re-run the failing CI jobs.

I'm trying the second option now

msridhar:
@mernst I set this to auto-merge, but once it does, https://github.com/codespecs/daikon/pull/417 should be landed as well

PR #5265, opened 2022-08-19
Defensive programming and documentation to avoid another problem like #5263

kelloggm:
This is a follow-up to [my comment](https://github.com/typetools/checker-framework/pull/5263#pullrequestreview-1072138414) on #5263.

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/mustcall/MustCallTransfer.java
+++ b/checker/src/main/java/org/checkerframework/checker/mustcall/MustCallTransfer.java
@@ -270,11 +270,14 @@ public class MustCallTransfer extends CFTransfer {
   }
 
   /** A unique identifier counter for node names. */
-  protected long uid = 0;
+  protected static long uid = 0;
 
   /**
-   * Creates a unique, abitrary string that can be used as a name for a temporary variable, using
+   * Creates a unique, arbitrary string that can be used as a name for a temporary variable, using
    * the given prefix. Can be used up to Long.MAX_VALUE times.
+   *
+   * <p>Note that the correctness of the Resource Leak Checker depends on these names actually being
+   * unique, because {@code LocalVariableNode}s derived from them are used as keys in a map.
    *
    * @param prefix the prefix for the name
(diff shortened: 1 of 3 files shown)


checker/src/main/java/org/checkerframework/checker/mustcall/MustCallTransfer.java
```diff
@@ -270,12 +270,15 @@ public void updateStoreWithTempVar(TransferResult<CFValue, CFStore> result, Node
   }
 
   /** A unique identifier counter for node names. */
-  protected long uid = 0;
+  protected static long uid = 0;
```
msridhar, line 273: This opens us up to data races.  Unlikely to cause a bug, but if it ever did that would be a nightmare to debug.  Can we use an `AtomicInteger` and `getAndIncrement()`?

msridhar (changes requested):

checker/src/main/java/org/checkerframework/checker/mustcall/MustCallTransfer.java
msridhar: I'll also say that I would prefer this counter somehow be stored in an object scoped to the current compilation, so that concurrent builds aren't impacting each other's temporary variable numbers.  But I don't immediately know of an easy way to do that.  If we have access to the javac `Context` object somehow, we could stick the state there.  But if that's not easy, we can go with the current approach.

kelloggm: I don't think there's an easy, safe way to get to the `Context` (but maybe you know one): I can get the `ProcessingEnvironment` from the checker, but I'd need to cast that to `JavacProcessingEnvironment` and then call `getContext()`, which has a specific comment in its javadoc telling you not to do that.

msridhar: Ok, that's fine.  I don't have another suggestion.

msridhar (approved):

PR #5272, opened 2022-08-30
Set environment variable before invoking `wpi-many.sh`

mernst:
Maybe a better approach would be for the `wpi*` scripts to default CHECKERFRAMEWORK to scriptsdirectory/../.. .


kelloggm (approved):
I think this is fine. `wpi-many.sh` is not yet, in my opinion, stable enough to use by non-developer users, so requiring a built copy of the CF is fine for now. Eventually, I agree that it would be better if it had a useful default.

PR #5274, opened 2022-08-30
Changed -Apropfiles to use File.pathSeparator

### LINKED ISSUE #5273: -Apropfiles separates file paths by colon, which prevents absolute paths on Windows
The documentation for the property file checker (and by extension, the I18n checker and the compiler message checker) in section 16.1 of the manual states:

```
Multiple files are separated by colons ’:’.
```

There seems to be no particular technical reason for this being colon on all operating systems besides an oversight in assuming it works on all of them.  However, on Windows this prevents the use of absolute paths for property files.  So if you pass:

```
-Apropfiles=C:\mydir\myprops.props:C:\mydir\myprops2.props
```

Then this is parsed as four files, the first and third being named "C".  This is annoying with some build tools as making a relative path is not always easy (or even possible, should the properties live on another drive).

I think the obvious solution is to change the separator on Windows to be a semi-colon (or more generally, to be Java's File.pathSeparator).  This does generate a backward incompatibility for users who are on Windows and use colons already (presumably with relative paths) with Apropfiles.  However I think (a) this is still the right solution and (b) I can't see a way round this without instead introducing an extra property to specify the path separator, which seems overkill when the convention is already clear and justified on each OS (colon on Linux, semicolon on Windows).

(I'm happy to make the merge request if you agree on my solution of using File.pathSeparator on all OSes and updating the manual to match.)

twistedsquare:
As suggested in #5273, changed the property file checker (and its subcheckers, the internationalization checker and the compiler message checker) to use File.pathSeparator to separate property file paths in -Apropfiles, rather than ':' (as that prevented use of absolute paths like C:\myfile.props on Windows).  Updated all the documentation and usages to match.

Fixes #5273


mernst (dismissed):
Looks great.  Thanks very much for this fix.

mernst (dismissed):

mernst (dismissed):

mernst (approved):

PR #5279, opened 2022-09-07
Adapt to JDK 18's different specification of `WildCardType.isUnbound()`

mernst:
This is the start of edits to handle Java 18.  More are to follow.


checker/tests/nullness/RawTypesNullness18.java
```diff
@@ -0,0 +1,32 @@
...
+class TestBounded18 {
+  BoundedGeneric18 rawReturn() {
+    return new BoundedGeneric18<>();
+  }
+
+  BoundedGeneric18 rawField = new BoundedGeneric18();
+
+  void useBoundedWildCard() {
+    BoundedGeneric18 rawLocal = new BoundedGeneric18<String>();
+    BoundedGeneric18<? extends Object> generic1 = rawReturn();
```
smillst, line 28: So the Checker Framework is going to have different defaulting rules for  `? extends Object` wildcards depending on the Java version? Or is this something we can fix later?

mernst: It should not.  Which of these is correct?

smillst: Oh I just noticed that the warnings in RawTypesNullness17.java are issued by `javac` not by the Checker Framework work.  You could add a `@SuppressWarnings("unchecked")` to this file and it would pass with all version of Java.  The Checker Framework passed both versions.

smillst (approved):
The code changes look good. I would consider merging the test cases into one and suppressing the javac warning.

PR #5283, opened 2022-09-08
Add copyJarToDist to developer manual.

smillst:
`assemble` is a standard Gradle task that builds all of the jars of a project including Javadoc jars.  If you just want to build enough of the jars to use checker/bin/javac, run `./gradlew copyJarsToDist`.


mernst (approved):

mernst:
Thanks.  I had recently figured that out.  I was just so used to using the `assemble` task in other projects.

PR #5289, opened 2022-09-09
Handle NPE in `getReceiverType()`

mernst:


framework/src/main/java/org/checkerframework/framework/type/DefaultAnnotatedTypeFormatter.java
```diff
@@ -270,10 +270,16 @@ public String visitExecutable(AnnotatedExecutableType type, Set<AnnotatedTypeMir
         sb.append("METHOD");
       }
       sb.append('(');
-      AnnotatedDeclaredType rcv = type.getReceiverType();
-      if (rcv != null) {
-        sb.append(visit(rcv, visiting));
-        sb.append(" this");
+      AnnotatedDeclaredType rcv;
+      try {
+        rcv = type.getReceiverType();
```
smillst, line 275: Is the NPE in `type.getReceiverType()` or the call to `visit` below?

smillst (commented):
Is there a test case for this? Or maybe you weren't able to minimize  one?

framework/src/main/java/org/checkerframework/framework/type/DefaultAnnotatedTypeFormatter.java
mernst: A stack trace indicates it is in `getReceiverType()`.  This only happens under JDK 18.

PR #5299, opened 2022-09-12
Use the spotless plugin.

smillst:
Note that `spotlessCheck` rejects `framework-test/src/main/java/org/checkerframework/framework/test/CheckerFrameworkWPIPerDirectoryTest.java`, but reformatting it `spotlessApply` causes the git hook formatting check to fail.


mernst (approved):
(I made changes to address my comments.)

Thanks!

PR #5301, opened 2022-09-13
Clean up uses of `elementFrom*` methods

mernst:


javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -272,72 +272,271 @@ public static boolean isSelfAccess(final ExpressionTree tree) {
...
+  @Deprecated // not for removal; retain to prevent calls to this overload
+  @Pure
+  public static @Nullable Element elementFromTree(ExpressionTree tree) {
+    return TreeUtils.elementFromUse(tree);
+  }
+
+  /**
+   * Returns the element corresponding to the given use.
+   *
+   * @param tree the tree corresponding to a use of an element
```
smillst, line 350: Not all ExpressionTrees are uses of an element.  For example, BinaryTrees are ExpressionTrees, but they are not uses of an element.  I would deprecate this method and use  `Element elementFromTree(ExpressionTree tree)` or `Element elementFromTree(Tree tree)` instead.

javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -272,72 +272,271 @@ public static boolean isSelfAccess(final ExpressionTree tree) {
...
+   */
+  @Pure
+  public static @Nullable Element elementFromUse(ExpressionTree tree) {
+    return TreeUtils.elementFromTree((Tree) tree);
+  }
+
+  /**
+   * Returns the VariableElement corresponding to the given use.
+   *
+   * @param tree the tree corresponding to a use of an element
```
smillst, line 361: This isn't a use of an element, it's a use of a variable element.  If tree is not a use of a variable, then this method will throw a ClassCastException.

javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -272,72 +272,271 @@ public static boolean isSelfAccess(final ExpressionTree tree) {
...
   /**
-   * Gets the {@link Element} for the given Tree API node.
+   * Returns the ExecutableElement for the given method declaration.
+   *
+   * <p>The result can be null, when `tree` is a method in an anonymous class.
+   *
+   * @param tree a method declaration
+   * @return the element for the given method
+   */
+  public static @Nullable ExecutableElement elementFromDeclaration(MethodTree tree) {
```
smillst, line 422: I don't think this should be `@Nullable`. I ` TreeInfo.symbolFor((JCTree) tree)` returns null, this method should throw a BugInCF like in `elementFromUse(MethodInvocationTree)`

mernst: The Javadoc explains:

```
 <p>The result can be null, when `tree` is a method in an anonymous class.
```

I have encountered this in practice.

javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
smillst: Did you intentional not mark this as deprecate or was it an oversight?  The Javadoc here is still incorrect. `tree` is an expression which may or may not be the use of an element.

javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
```diff
@@ -272,87 +272,335 @@ public static boolean isSelfAccess(final ExpressionTree tree) {
...
   }
 
   /**
-   * Gets the {@link Element} for the given Tree API node.
+   * Returns the ExecutableElement for the given method declaration.
+   *
+   * <p>The result can be null, when {@code tree} is a method in an anonymous class.
+   *
+   * @param tree a method declaration
+   * @return the element for the given method
```
smillst, line 463: It would be helpful if this said that `null` can be return for anonymous classes because this line makes it seem like the return is non-null.

smillst (approved):

javacutil/src/main/java/org/checkerframework/javacutil/TreeUtils.java
mernst: Done.

smillst (approved):

PR #5304, opened 2022-09-13
Remove method that isn't called anymore.

smillst:
Calls to it were removed in d7149190dc06ab63c71af0bb7a07cf13db18b6b4 and replaced with calls to ` AnnotatedTypes#overriddenMethods`.


mernst (approved):
Thanks!

mernst (approved):

PR #5307, opened 2022-09-14
Fix crash in Index Checker when reading an ajava file during WPI

kelloggm:
Also includes infrastructure to run a WPI unit test using the Index Checker, and some associated changes in the WPI testing infrastructure to remove the assumption that all WPI tests use the AinferTestChecker.


kelloggm:
Updated to fix a similar problem that could occur in the Interning Checker, which had a similar "call the Value Checker" kind of code pattern.

kelloggm:
I've checked every other checker that overrrides `GenericAnnotatedTypeFactory#addComputedTypeAnnotations` and no others have this issue.

mernst (approved):
Thanks!

PR #5308, opened 2022-09-14
Don't run Java formatting under JDK 8

mernst:
~I added the changes in https://github.com/typetools/checker-framework/pull/5325 to this PR to test that they fix the bug exposed by this PR.  Merge https://github.com/typetools/checker-framework/pull/5325  first.~

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -90,6 +90,32 @@ task installGitHooks(type: Copy, dependsOn: 'setLocalRepo') {
     into localRepo + "/hooks"
 }
 
+spotless {
+  // Resolve the spottless plugin dependencies from the buildscript repositories rather than the
+  // project repositories.  That way the spotless plugin does not use the locally built version of
+  // checker-qual as a dependency. Without this errors like the follow are issued when running
+  // a spotless task without a local built version of checker-qual.jar:
+  // Could not determine the dependencies of task ':checker-qual:spotlessCheck'.
+  //> Could not create task ':checker-qual:spotlessJavaCheck'.
+  //   > Could not create task ':checker-qual:spotlessJava'.
+  //      > File signature can only be created for existing regular files, given: /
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -169,6 +170,7 @@ allprojects {
       googleJavaFormat()
       formatAnnotations()
     }
+  }
   }
```
mernst, line 175: Alternately, we could handle this [the way I did in Randoop](https://github.com/randoop/randoop/pull/1072/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R280-R285).  Which is better, or is there an even better solution?

smillst: It's better to disable it, because it's not defined what happens if no spotless configuration is specified.  (Or at least a quick skim for the documentation doesn't say what happends.

Why can't spotless be used with Java 8?

smillst (changes requested):

build.gradle
mernst: > Why can't spotless be used with Java 8?

Spotless uses the latest version of google-java-format that is compatible with the currently running JVM.  Under Java 8, that is an older version that produces different formatting than more recent versions of google-java-format that run under Java 11+.

smillst (dismissed):

mernst:
@smillst Can you look at the CI failures?

smillst:
I can reproduce this on master with the following command:
```
./gradlew clean
./gradlew spotlessApply
```
This seems to be a general problem with the way the spotless plugin works for subprojects.  https://github.com/diffplug/spotless/issues/715 is somewhat related.  I'm trying to figure out a work around.

smillst:
The problem is that `checker-qual` is a dependency of `google-java-format` and Gradle is trying to use the built from source version.   I'm still working on a fix, but it's a Checker Framework specific problem.

smillst (approved):

PR #5317, opened 2022-09-16
Test under JDK 18

mernst:
Merge together with
https://github.com/typetools/annotation-tools/pull/503
https://github.com/typetools/jdk/pull/146
https://github.com/typetools/jdk17u/pull/3


build.gradle
```diff
@@ -261,7 +263,7 @@ allprojects {
             options.deprecation = true
             options.compilerArgs += [
                     '-g',
-                    '-Werror',
+                    // '-Werror',
```
smillst, line 266: Should this change be reverted?

build.gradle
```diff
@@ -310,7 +312,7 @@ allprojects {
                         // the errors get fixed.  On the downside, Error Prone (or maybe the compiler?)
                         // stops as soon as it issues one warning, rather than outputting them all.
                         // https://github.com/google/error-prone/issues/436
-                        '-Werror',
+                        // '-Werror',
```
smillst, line 315: Should this change be reverted?

checker/tests/fenum/UpperBoundsInByteCode.java
```diff
@@ -1,3 +1,5 @@
+// @above-java17-jdk-skip-test TODO: reinstate on JDK 18, false positives may be due to issue #979
```
smillst, line 1: Should I look into these warnings?

smillst (dismissed):
I had two minor question, but otherwise looks good.

build.gradle
mernst: Yes.  Thanks for catching this.

checker/tests/fenum/UpperBoundsInByteCode.java
mernst: Yes, please look into them as part of your work on #979, when you return to that task.

I don't have an opinion whether you should look into these problems first, or do other parts of the task that you know are essential before seeing whether these get fixed as a side effect.

PR #5318, opened 2022-09-16
Fix ajava-based WPI generating malformed annotation with String constants containing escapable characters

kelloggm:
This PR actually (partially) addresses three related bugs, all of which are triggered by the new test case I added (ContainsString.java):
* ajava-based WPI generated malformed annotations when an annotation value contained a string literal with escapable characters such as ". The specific cause of this was `@SameLen(""Hamburg"")` being generated; that should have been `@SameLen("\"Hamburg\"")`.
* The SameLen Checker shouldn't have been inferring annotations with manifest string literals in the first place. I fixed that here, too, and also modified the test to force the Value Checker to elicit the above bug without the fix in this PR.
* The test still fails without a warning suppression, because WPI isn't correctly viewpoint adapting to the method when inferring parameter types. I didn't fix that here, since I think it will be a more invasive change that should be done separately, but I did leave some TODOs where I think those changes should go. @smillst, this is what I emailed you about a few minutes ago.


mernst (approved):
Thanks for the fixes.

PR #5321, opened 2022-09-16
Viewpoint adapt during WPI at method call sites

kelloggm:
Based on the suggestion from @smillst.

I'm concerned that this PR computes a `TreePath` sometimes when it doesn't really need one, which I think I remember might be relatively expensive, but I'm not sure how to check whether the `TreePath` will be necessary earlier than in `DependentTypesHelper`. I'd appreciate any suggestions (or a reassurance that it's fine, if it is fine).

Also fixes a bug in GLB in the SameLen checker that I encountered while testing this fix: the `unionIfNotDisjoint` helper method for SameLen's GLB just did not do what its specification said, leading to infuriating test failures caused by the GLB of `@SameLen("#1")` and `@SameLen("#1")` being `@SameLenBottom`. If you'd like, I could pull that out into a separate PR, but I kept it here since I don't have a separate test case.

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
+++ b/checker/src/main/java/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
@@ -23,6 +23,7 @@ import org.checkerframework.checker.index.qual.PolySameLen;
 import org.checkerframework.checker.index.qual.SameLen;
 import org.checkerframework.checker.index.qual.SameLenBottom;
 import org.checkerframework.checker.index.qual.SameLenUnknown;
+import org.checkerframework.checker.nullness.qual.Nullable;
 import org.checkerframework.common.basetype.BaseAnnotatedTypeFactory;
 import org.checkerframework.common.basetype.BaseTypeChecker;
 import org.checkerframework.dataflow.expression.ArrayCreation;
@@ -177,14 +178,19 @@ public class SameLenAnnotatedTypeFactory extends BaseAnnotatedTypeFactory {
      * If the collections are disjoint, returns null. Otherwise, returns their union. The
      * collections must not contain duplicates.
      */
-    private Set<String> unionIfNotDisjoint(Collection<String> c1, Collection<String> c2) {
+    private @Nullable Set<String> unionIfNotDisjoint(Collection<String> c1, Collection<String> c2) {
       Set<String> result = new TreeSet<>(c1);
+      boolean disjoint = true;
       for (String s : c2) {
         if (!result.add(s)) {
-          return null;
+          disjoint = false;
         }
       }
-      return result;
+      if (!disjoint) {
+        return result;
+      } else {
+        return null;
+      }
     }
 
     // The GLB of two SameLen annotations is the union of the two sets of arrays, or is bottom
--- a/checker/tests/ainfer-index/non-annotated/ContainsSubstring.java
+++ b/checker/tests/ainfer-index/non-annotated/ContainsSubstring.java
@@ -4,9 +4,6 @@
 
 public class ContainsSubstring {
 
-  @SuppressWarnings("samelen") // TODO: caused by a bug related to viewpoint adaptation. Make WPI
-  // actually viewpoint-adapt the annotations that it infers, then
-  // remove this warning suppression.
   public static void run() {
     String word1 = "\"Hamburg\"";
     String word2 = "burg";
--- a/framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
+++ b/framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
@@ -15,14 +15,9 @@ import com.sun.source.util.TreePath;
 import com.sun.tools.javac.tree.JCTree;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.EnumSet;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 import javax.annotation.processing.ProcessingEnvironment;
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.Element;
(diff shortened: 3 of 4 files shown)


smillst (changes requested):

kelloggm:
@smillst Your comments made me realize that, because I do actually only want the `#1`, `#2`, etc., syntax, I actually don't need a separate `delocalizeAtCallsite` method at all: the semantics are identical to `delocalize`, and I can get a `MethodTree` from the element using the `AnnotatedTypeFactory`. This is a much simpler version of this change that 1) doesn't duplicate code, and 2) includes the test you suggested (i.e., an inferred dependent type annotation with a local variable).

checker/tests/ainfer-index/non-annotated/ContainsSubstring.java
```diff
@@ -4,12 +4,14 @@
...
-  // remove this warning suppression.
   public static void run() {
     String word1 = "\"Hamburg\"";
     String word2 = "burg";
+    // The existence of word3 here forces the inferred @SameLen
+    // annotation to include a local variable that isn't a parameter
+    // to compute(), to test that such local variables are viewpoint-adapted
+    // correctly.
+    String word3 = word1;
     System.out.println(compute(word1, word2));
```
smillst, line 15: You need a test case where the arguments don't have the same names as the method parameters.  `delocalize` is resolving the names base on the declaration of `compute` rather than the callsite.

smillst (changes requested):
I added a comment separate from this review.

checker/tests/ainfer-index/non-annotated/ContainsSubstring.java
kelloggm: Ah, right you are. It was a nice thought while it lasted :)

I will revert back to the previous version and address your previous review directly. Sorry!

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -654,6 +656,61 @@ protected JavaExpression visitLocalVariable(
...
+      ExecutableElement methodElt) {
+
+    if (!hasDependentType(atm)) {
+      return;
+    }
+
+    // For use in stringToJavaExpr below, to avoid re-computation. Especially
+    // important for the TreePath, which is expensive to compute.
+    List<JavaExpression> argsAsExprs =
+        arguments.stream().map(LocalVariable::fromNode).collect(Collectors.toList());
```
smillst, line 683: Use org.plumelib.util.CollectionsPlume#mapList(Function<? super FROM,? extends TO>, Iterable<FROM>)

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -654,6 +656,61 @@ protected JavaExpression visitLocalVariable(
...
+        stringExpr -> {
+          JavaExpression expr =
+              StringToJavaExpression.atPath(stringExpr, path, factory.getChecker());
+          JavaExpressionConverter jec =
+              new JavaExpressionConverter() {
+                @Override
+                public JavaExpression convert(JavaExpression javaExpr) {
+                  // if javaExpr is an argument to the method,
+                  // then return formal parameter expression.
+                  int index = argsAsExprs.indexOf(javaExpr);
```
smillst, line 696: What about converting the receiver expression at the call site to `this`? Should `visitThisReference` also return null (unless the receiver expression is at the call site is `this`)?

smillst (changes requested):
I had one minor

kelloggm:
This PR is currently blocked by #5328. Once that PR is merged, I can proceed with this one. (`@SameLen` is not currently permitted on user-defined sequence classes, so I cannot currently test receiver parameters with `@SameLen` annotations, since the only classes `@SameLen` is currently permitted on are either final (like `String`) or arrays.)

kelloggm:
@smillst this is ready for another review when you get a chance

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
smillst: This hasn't been implemented.  For example, if the call sit is `expr.method(param)` and the type of `param` is  `@SameLen("expr")`, then it can be viewpoint-adapted to `@SameLen("this")`.  Maybe you don't want to implement that right now?  If so, you might to leave a TODO.

smillst (approved):
If you want to merge this without addressing this last comment that's fine.

checker/src/main/java/org/checkerframework/checker/index/samelen/SameLenAnnotatedTypeFactory.java
```diff
@@ -174,17 +175,34 @@ public AnnotationMirror getTopAnnotation(AnnotationMirror start) {
...
+     * union. The collections must not contain duplicates.
+     *
+     * @param c1 a collection of Strings (intended to be the value argument of a SameLen annotation)
+     * @param c2 another collection of Strings
+     * @return if the two inputs are disjoint (i.e., have no elements in common) and both are
+     *     non-empty, returns null. Otherwise, returns the union of the two collections (which, if
+     *     one collection is empty, is just the other collection).
      */
-    private Set<String> unionIfNotDisjoint(Collection<String> c1, Collection<String> c2) {
+    private @Nullable Collection<String> unionIfNotDisjoint(
```
smillst, line 187: You can put this is a separate PR without a test case.  I'm also fine if you want to leave it here.

kelloggm: I'd prefer to keep it here so that it is connected to the test case in this PR.

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
kelloggm: Ah, I misunderstood what you were suggesting. WPI doesn't actually have support for inferring types on the receivers of methods from call sites, which is a TODO for WPI. I've added that TODO explicitly in the appropriate place in the WPI code with a note about the need to update `DependentTypesHelper` as well when that's implemented, and maybe I'll try implementing it later this week if I get a chance.

I also wrote a test for the situation you're describing, but commented it out (since it's still TODO).

kelloggm:
@smillst I think this is ready to be merged, but your review is stale because I added a (skipped) test and some todos

smillst (approved):

smillst (approved):

PR #5322, opened 2022-09-17
Avoid tracking useless must-call obligations in MustCallConsistencyAnalyzer

msridhar:
Previously, `MustCallConsistencyAnalyzer` would track obligations for expressions whose type had no `@MustCall` obligations, which was wasteful.  This PR detects such cases and avoids any obligation tracking for those expressions.

For this change to be safe, we must disallow creation of obligations (via `@CreatesMustCallFor`) on expressions whose declared type has no `@MustCall` obligation.  This PR also includes an implementation of this new check, inside `ResourceLeakVisitor`.  Implementing this check required refactoring the code in `CreatesMustCallForElementSupplier`.


msridhar:
@kelloggm I made a best effort to update existing tests in a sensible way, as several failed the newly-implemented check.  In some cases, I added a non-empty `@MustCall` to the relevant type, and in other cases I left the test alone and made the error expected.

checker/src/main/java/org/checkerframework/checker/mustcall/CreatesMustCallForToJavaExpression.java
```diff
@@ -0,0 +1,198 @@
...
+import org.checkerframework.framework.util.JavaExpressionParseUtil;
+import org.checkerframework.framework.util.StringToJavaExpression;
+import org.checkerframework.javacutil.AnnotationUtils;
+import org.checkerframework.javacutil.TreeUtils;
+
+/**
+ * Utility methods to convert targets {@code @CreatesMustCallFor} annotations to {@link
+ * org.checkerframework.dataflow.expression.JavaExpression}s.
+ */
+public class CreatesMustCallForToJavaExpression {
```
msridhar, line 25: Extracted this class to enable making some of the static methods private

checker/src/main/java/org/checkerframework/checker/mustcall/CreatesMustCallForToJavaExpression.java
```diff
@@ -0,0 +1,201 @@
...
+        atypeFactory.getDeclAnnotation(method, CreatesMustCallFor.class);
+    if (createsMustCallFor != null) {
+      result.add(createsMustCallFor);
+    }
+    return result;
+  }
+
+  /**
+   * Parses a single CreatesMustCallFor annotation. Clients should use {@link
+   * #getCreatesMustCallForExpressionsAtInvocation(MethodInvocationNode,
```
kelloggm, line 140: I don't think this suggestion is relevant or even necessarily true anymore:
* it's not relevant, since this method is now private rather than package-private, so no clients have access to it (this is a good change you orchestrated by moving these methods out of that interface!)
* it's not longer really true, since the interface of this method vs `getCreatesMustCallForExpressionsAtInvocation` is now different, and that method is no longer the only alternative (`getCreatesMustCallForExpressionsAtDeclaration` now exists, too)

I think we can just remove this suggestion, since it's clear that clients of this class should use the `public` methods.

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
```diff
@@ -1565,22 +1570,37 @@ private List<? extends VariableElement> getParametersOfInvocation(Node node) {
...
+    }
+    TypeMirror type = ElementUtils.getType(executableElement);
+    // void or primitive-returning methods are "not owning" by construction
+    if (type.getKind() == TypeKind.VOID || type.getKind().isPrimitive()) {
+      return false;
+    }
+    TypeElement typeElt = TypesUtils.getTypeElement(type);
+    // no need to track if type has no possible @MustCall obligation
+    if (typeElt != null && typeFactory.getMustCallValue(typeElt).isEmpty()) {
+      return false;
```
kelloggm, line 1600: This is unsound, because it doesn't take into account potential `@MustCall` annotations on types that don't have default `@MustCall` annotations on their declarations. I wrote a failing test here: https://github.com/msridhar/checker-framework/pull/74

checker/tests/resourceleak/CreatesMustCallForTargets.java
```diff
@@ -10,6 +10,7 @@ class CreatesMustCallForTargets {
   @Owning InputStream is1;
 
   @CreatesMustCallFor
+  // :: error: createsmustcallfor.target.unparseable
```
kelloggm, line 13: This error looks correct, but it worries me that it only started showing up in this PR - "this" was never a valid target for a static method.

kelloggm (changes requested):

checker/tests/resourceleak/CreatesMustCallForTargets.java
msridhar: Before, we would only try to parse the annotation targets in the invocation context, not the method declaration context.  Further, this method is never invoked.  So we didn't get the unparseable error before.  Now we parse the targets in both contexts, so we get an error at this declaration.

checker/src/main/java/org/checkerframework/checker/mustcall/CreatesMustCallForToJavaExpression.java
msridhar: 3fc29f3b6a626edf9b37c734b773d062dcdc52d0

checker/src/main/java/org/checkerframework/checker/resourceleak/MustCallConsistencyAnalyzer.java
msridhar: Good catch!  Fixed in f77cda6de94b88ebed79a452f7d4962fe734734f, and I also added a test case for constructors.

checker/tests/resourceleak/CreatesMustCallForTargets.java
kelloggm: That explanation makes sense, thanks. I'm glad it doesn't indicate a deeper problem.

kelloggm (approved):
LGTM

kelloggm:
We might still want to do some profiling on bigger benchmarks, but I seriously doubt this will make perf **worse**, so I'll merge it now if tests pass.

msridhar:
Sorry @kelloggm I added one word to a comment and it dismissed your review 😅

kelloggm (approved):

PR #5325, opened 2022-09-19
Fix problem with configuring the spotless plugin.

smillst:
This fixed the problem with configuring the spotless plugin that was exposed by https://github.com/typetools/checker-framework/pull/5308.


mernst (approved):
Thanks for figuring this out.

PR #5331, opened 2022-09-21
Rename a confusingly-named method

kelloggm:


mernst (approved):
Thanks for the improvement.

PR #5334, opened 2022-09-22
Treat resource variables the same as local variables.

smillst:
Fixes issue in #5332.


mernst (approved):
Thanks for the quick fix.

PR #5338, opened 2022-09-27
Rename `getPurityKinds()` to `getPurityAnnotations()`

mernst:
This change makes clearer where the information comes from, and is more consistent with clients such as `hasPurityAnnotation()`.

### DIFF
--- a/dataflow/src/main/java/org/checkerframework/dataflow/util/PurityUtils.java
+++ b/dataflow/src/main/java/org/checkerframework/dataflow/util/PurityUtils.java
@@ -31,7 +31,7 @@ public class PurityUtils {
    * @return whether the method has any purity annotations
    */
   public static boolean hasPurityAnnotation(AnnotationProvider provider, MethodTree methodTree) {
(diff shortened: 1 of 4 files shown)


dataflow/src/main/java/org/checkerframework/dataflow/util/PurityUtils.java
```diff
@@ -98,36 +98,64 @@ public static boolean isSideEffectFree(AnnotationProvider provider, MethodTree m
...
+  }
+
+  /**
+   * Returns the types of purity of the method {@code methodElement}.
+   *
+   * @param provider how to get annotations
+   * @param methodElement a method to test
+   * @return the types of purity of the method {@code methodElement}
+   */
+  public static EnumSet<Pure.Kind> getPurityAnnotations(
```
smillst, line 156: Given this method name, I would expect it to return a set (or collection) of `AnnotationMirror`s.  The previous name  reflected that set of `Pure.Kind` is returned.  So, I wouldn't rename this method.

PR #5343, opened 2022-09-28
Infer purity when using the Lock Checker

mernst:
This ought to fix the nondeterministic problem in CI inference jobs.


javacutil/src/main/java/org/checkerframework/javacutil/BasicAnnotationProvider.java
```diff
@@ -41,4 +42,26 @@ public class BasicAnnotationProvider implements AnnotationProvider {
...
+   * {@inheritDoc}
+   *
+   * <p>This implementation returns true if the {@code @SideEffectFree} annotation is present on the
+   * given method.
+   */
+  @Override
+  public boolean isSideEffectFree(ExecutableElement methodElement) {
+    List<? extends AnnotationMirror> annotationMirrors = methodElement.getAnnotationMirrors();
+
+    // Then look at the real annotations.
```
smillst, line 56: I don't understand this comment.

smillst (approved):
I just had one minor comment, but otherwise this is fine to merge.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -230,10 +230,14 @@
    * command line.
    */
   private final boolean checkPurity;
+  // TODO: Given that setting this to false is problematic, should it be removed entirely?
```
mernst, line 233: @smillst I am inclined to remove the variable.  What do you think?

mernst: I removed it.  I think the Lock Checker always set it.

Why was `getCheckerNames()` nondeterministic?  Was it because WPI was creating file `META-INF/services/javax.annotation.processing.Processor` nondeterministically?

smillst (approved):

PR #5354, opened 2022-10-03
WPI new feature: infer receiver types for method declarations based on the types of receivers at callsites

kelloggm:
Fixes a weakness in WPI that we discovered while working on #5321, so this is sort of a follow up PR.


framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceImplementation.java
```diff
@@ -222,21 +233,35 @@ private boolean isRecursiveCall(MethodInvocationNode methodInvNode) {
...
-    // 3. DependentTypesHelper#delocalizeAtCallsite should be updated to handle dependent types
-    //    that refer to the receiver (there is a (commented-out) test for this in
-    //    tests/ainfer-index/non-annotated/DependentTypesViewpointAdapationTest.java).
+      ExecutableElement methodElt,
+      List<Node> arguments,
+      @Nullable Node receiver,
+      ExpressionTree invocationTree) {
 
     String file = storage.getFileForElement(methodElt);
+    // TODO: why can a constructor possibly have a non-null receiver? That makes no sense, because
```
mernst, line 247: A constructor for a nested non-static class (that is, an inner class) has a receiver that is a reference to the outer class instance that encloses the class.

mernst (approved):
This looks good; thanks for implementing it.

Tests are failing because `@NonNull` and `@Nullable` should not be inferred on receivers.

mernst (approved):

PR #5355, opened 2022-10-03
Clone JSpecify, now that its build works again across JDKs.

netdpb:
This reverts commit 803bc76c71944cf53956771b669c570dc2ea3517.


mernst:
Hi, David.  CI doesn't run the JSpecify tests, so this change would only add time to the CI build. (It would also test that JSpecify didn't break its build again, but I think that is better done in the JSpecify repository.)  Is there some other reason to make this change?  Thanks!

netdpb:
I agree, and I'm only doing this because we've made that change. (The problem was that JSpecify's CI only tested JDK 11, and even then didn't generate Javadoc.)

If CF's CI doesn't need to run the JSpecify tests, can we delete this entire stanza rather than commenting it out and implying that that's a problem?

mernst:
That's a good idea.  The comment wasn't meant to throw any shade on JSpecify, and was accurate at the time.

netdpb:
Does the build still rely on this script cloning the JSpecify repo? Can I delete everything, or should I keep the git clone?

The fact that the build jumps around between Gradle and shell-scripting makes it hard to understand what's required.

mernst:
@netdpb I cannot find the `jspecify-ci` branch in your fork any more.

netdpb:
Looks like it got cleaned up for being stale.

PR #5356, opened 2022-10-03
Exclude AinferNullnessAjavaTest from junit tests.

smillst:


mernst:
What is the reason for this exclusion?  @kelloggm might be a better reviewer than I am.

checker/build.gradle
```diff
@@ -374,6 +374,7 @@ test {
         excludeCategories 'org.checkerframework.checker.test.junit.ainferrunners.AinferTestCheckerStubsTest'
         excludeCategories 'org.checkerframework.checker.test.junit.ainferrunners.AinferTestCheckerAjavaTest'
         excludeCategories 'org.checkerframework.checker.test.junit.ainferrunners.AinferNullnessJaifsTest'
+        excludeCategories 'org.checkerframework.checker.test.junit.ainferrunners.AinferNullnessAjavaTest'
```
kelloggm, line 377: This should also be added to the list of `ainferTest` tasks around line 840, so that it does get run.

kelloggm (changes requested):
Thanks for catching my mistake: I should have added this exclusion when I added the test.

checker/build.gradle
smillst: Is that in addition to this change?

kelloggm (approved):
LGTM

PR #5358, opened 2022-10-04
`git diff-index` exit code is 0 if there are no diffs

smillst:


docs/developer/release/release_utils.py
```diff
@@ -269,7 +269,7 @@ def commit_tag_and_push(version, path, tag_prefix):
     """Commit the changes made for this release, add a tag for this release, and
     push these changes."""
     # Do nothing (instead of erring) if there is nothing to commit.
-    if execute("git diff-index --quiet HEAD", False, False, working_dir=path) == 0:
+    if execute("git diff-index --quiet HEAD", False, False, working_dir=path) != 0:
```
mernst, line 272: The return codes of `git diff-index` are confusing (because they are different than `diff`'s. I am not wild about its choices, but consistency is important).  There is also `--exit-code` which makes `git diff-index` have exit codes similar to `diff`'s.  I often use that, but really either way is fine with me.

mernst (approved):

PR #5368, opened 2022-10-12
Fix WPI crash when delocalizing deterministic method calls

kelloggm:
Fixes a regression introduced in #5321, based on a crash discovered by @dd482IT when running WPI on require-javadoc.


mernst (approved):
Thanks for the fix!

PR #5374, opened 2022-10-13
Fix typo for Resource Leak Checker

Thomsch:
This PR fixes a typo in the Checker Framework manual for the Resource Leak Checker.


mernst:
Thank you!

CI is going to fail because you need to add yourself to the list of contributors (that is, you need to edit one more file).  I have prioritized that CI job, so you should see the failure within 15 minutes.

Thomsch:
So sorry about that! Thank you for adding me!

PR #5392, opened 2022-10-26
Make @EnsuresCalledMethods repeatable

kelloggm:
The motivation for this is two-fold:
* other `@EnsuresX` annotations (e.g., `@EnsuresNonNull`, `@EnsuresQualifier`) are repeatable
* WPI sometimes produces more than one `@EnsuresCalledMethods` annotation for the same method. It would be possible to merge them at the WPI stage, but that would be more complicated


mernst (approved):
Thanks!

PR #5397, opened 2022-11-01
Update CF Gradle plugin version numbe

smillst:


kelloggm (approved):

kelloggm (changes requested):
The current version of the plugin (i.e., the one I just released) is 0.6.19, not 0.6.18. I'm not sure how they got out of sync, but we should correct it.

Issue #5402, opened 2022-11-10
@This does not work with @RequiresCalledMethods when chaining


tmarback:
I'm observing a behavior where it seems that `@This` (i.e. the Returns Receiver Checker) does not properly forward the `@CalledMethods` qualifier of the object when chaining the result to a method that uses `@RequiresCalledMethods` to define requirements for the receiver.

Chaining the method call works when the requirements are defined with `@CalledMethods` on the explicit receiver parameter:

```java
import org.checkerframework.checker.calledmethods.qual.CalledMethods;
import org.checkerframework.checker.calledmethods.qual.RequiresCalledMethods;
import org.checkerframework.common.returnsreceiver.qual.This;

public class Foo {

    public @This Foo bar() {
        return this;
    }

    public void baz(@CalledMethods("bar") Foo this) {}

    public static void test() {
        final Foo foo = new Foo();
        foo.bar().baz(); // No error
    }

}
```

Storing the intermediate result, then calling a `@RequiresCalledMethods`-annotated method also works:

```java
import org.checkerframework.checker.calledmethods.qual.RequiresCalledMethods;
import org.checkerframework.common.returnsreceiver.qual.This;

public class Foo {

    public @This Foo bar() {
        return this;
    }

    @RequiresCalledMethods( value = "this", methods = "bar" )
    public void baz() {}

    public static void test() {
        final Foo foo = new Foo();
        final Foo foo1 = foo.bar();
        foo1.baz(); // No error
    }

}
```

But if I chain the call to the `@RequiresCalledMethods`-annotated method:

```java
import org.checkerframework.checker.calledmethods.qual.RequiresCalledMethods;
import org.checkerframework.common.returnsreceiver.qual.This;

public class Foo {

    public @This Foo bar() {
        return this;
    }

    @RequiresCalledMethods( value = "this", methods = "bar" )
    public void baz() {}

    public static void test() {
        final Foo foo = new Foo();
        foo.bar().baz(); // Error
    }

}
```

I get the error

```
> .\bin\javac.bat -cp . -processor calledmethods Foo.java
Foo.java:42: error: [contracts.precondition] precondition of foo.bar().baz is not satisfied.
        foo.bar().baz();
                     ^
  found   : foo.bar() is @CalledMethods
  required: foo.bar() is @CalledMethods("bar")
1 error
```

Which indicates that the qualifier isn't getting forwarded.

[...]

PR #5403, opened 2022-11-10
Test case for issue #5402

mernst:
~Hmm, this seems to contain an unrelated change.~

@kelloggm You might already have a test case; this is in case you didn't create it yet.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -164,6 +164,7 @@ jobs:
     fetchDepth: 25
   - bash: ./checker/bin-devel/test-cftests-inference.sh
     displayName: test-cftests-inference.sh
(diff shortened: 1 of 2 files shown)


mernst:
I removed the unrelated change.

kelloggm:
I added the example that uses `@Deterministic` from [my comment](https://github.com/typetools/checker-framework/issues/5402#issuecomment-1310456779) on the issue. I suspect we won't actually be able to solve this false positive, unfortunately - I looked into whether I could get the Called Methods Checker to automatically treat all `@This` methods as deterministic, but the code that handles that is all in `PurityUtils` and so can't be customized per-checker.

kelloggm (approved):

PR #5407, opened 2022-11-15
Run the Resource Leak Checker

mernst:


checker/src/main/java/org/checkerframework/checker/mustcall/JdkCompiler.astub
```diff
@@ -0,0 +1,6 @@
+import org.checkerframework.checker.mustcall.qual.*;
+
+package com.sun.tools.javac.processing;
+
+@MustCall({})
+public class JavacProcessingEnvironment implements ProcessingEnvironment, Closeable {}
```
msridhar, line 6: If we are adding this stub file, why do we need the `-AskipUses` on `ProcessingEnvironment` in `build.gradle`?

framework-test/src/main/java/org/checkerframework/framework/test/diagnostics/JavaDiagnosticReader.java
```diff
@@ -156,7 +159,7 @@ private interface StringToTestDiagnosticLine {
   private final String filename;
 
   /** The reader for the file. */
-  private LineNumberReader reader;
+  private final @Owning @MustCall("close") LineNumberReader reader;
```
msridhar, line 162: Why do we need to write `@MustCall("close")` here?

msridhar (commented):
Mostly LGTM, just a couple of small questions

msridhar (approved):

PR #5432, opened 2022-12-02
Add interning annotations

mernst:


mernst:
Per @smillst:

> All Enums should be defaulted to `@Interned`, so the fact that you had to add this annotation actually indicates there is a bug in the Interning Checker.  (It looks like Enums that are type parameter bounds are not defaulted correctly.)

PR #5444, opened 2022-12-13
Fix "RemoveAnnotationsForInference.java" for multiline annotations 

MuyeedAhmed:
This script failed to remove the last line of a multiline annotation properly. 
For example, for the following 3 lines, the RemoveAnnotationsForInference fails to remove the last line.
```
  @InlineMe(
      replacement = "BcelUtil.binaryNameToType(classname)",
      imports = "org.plumelib.bcelutil.BcelUtil")
```
The string newLastLine should be everything after the annotation. But instead, it is 0 till the end of the annotation (Removing everything after the annotation).


mernst (approved):
Thanks for the fix.  That is an embarrassing mistake.

PR #5459, opened 2022-12-20
Add `static` and `final` modifiers

mernst:


checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsAnnotatedTypeFactory.java
```diff
@@ -112,7 +112,7 @@ public CalledMethodsAnnotatedTypeFactory(BaseTypeChecker checker) {
    *
    * @param disabledFrameworks the disabled builder frameworks
    */
-  private void enableFrameworks(String[] disabledFrameworks) {
+  private final void enableFrameworks(String[] disabledFrameworks) {
```
smillst, line 115: `final` on a `private` method is redundant.  Is there a reason you like to add it?

framework/src/main/java/org/checkerframework/framework/util/dependenttypes/DependentTypesHelper.java
```diff
@@ -212,6 +212,7 @@ public TreeAnnotator createDependentTypesTreeAnnotator() {
   ///
 
   /** If true, log information about where lambdas are created. */
+  // This variable is only set here; edit the source code to modify it.
   private static boolean debugStringToJavaExpression = false;
```
smillst, line 216: Should this be final?

smillst (approved):

PR #5461, opened 2022-12-20
Resolve RLC unsoundness with two (or more) @Owning fields + @MustCallAlias on a constructor

### LINKED ISSUE #5453: Soundness bug in Resource Leak Checker for type wrapping multiple resources
Example:
```java
@InheritableMustCall({"finish1", "finish2"})
 class TwoOwningMCATest {
   @Owning private final Foo f1 = new Foo();
   @Owning private final Foo f2;
   @MustCallAlias
   TwoOwningMCATest(@MustCallAlias Foo g) {
     this.f2 = g;
   }
   @EnsuresCalledMethods(value="this.f1", methods="a")
   void finish1() {
     this.f1.a();
   }
   @EnsuresCalledMethods(value="this.f2", methods="a")
   void finish2() {
     this.f2.a();
   }
   @InheritableMustCall("a")
   static class Foo {
     void a() {}
   }
 }
```
The Resource Leak Checker reports no error for the above.  But the `@MustCallAlias` annotations on the constructor are invalid, as satisfying the must-call obligation of the parameter is _not_ sufficient to satisfy the obligations of the newly-created `TwoOwningMCATest` object.

/cc @kelloggm @Nargeshdb

kelloggm:
Fixes #5453


msridhar (commented):
The changes look good to me.  Could we add a test case like `TwoSocketContainer` where the constructor parameters are annotated `@Owning` rather than MCA?  If we have a client that calls `TwoSocketContainer.close1()` _and_ `TwoSocketContainer.close2()` that should get verified by the checker.

kelloggm:
> The changes look good to me. Could we add a test case like `TwoSocketContainer` where the constructor parameters are annotated `@Owning` rather than MCA? If we have a client that calls `TwoSocketContainer.close1()` _and_ `TwoSocketContainer.close2()` that should get verified by the checker.

089a6d3f7

msridhar (approved):
LGTM!

PR #5467, opened 2022-12-22
Throw more informative error when element is null

mernst:


framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
```diff
@@ -446,6 +446,10 @@ public void preprocessClassTree(ClassTree classTree) {
    */
   private void addClassTree(ClassTree tree) {
     TypeElement element = TreeUtils.elementFromDeclaration(tree);
+    if (element == null) {
+      throw new BugInCF("element is null for ClassTree: " + tree);
```
kelloggm, line 450: I think it would be better to `return` here rather than throw an exception: WPI is allowed to fail to make inferences, and we'd prefer that WPI fail to make an inference rather than crash. We could add debug logging here (using `atypeFacory.getWholeProgramInference().showFailedInferences` or an accessor of that field to guard it).

The consequences of a `return` here would be no entry for the class represented by `tree` in the `classToAnnos` map. There are three important uses of that map in `WholeProgramInferenceJavaParserStorage`: in `getMethodAnnos` (where the result is already checked for `null`, indicating that there might be keys without values already) and in `getFieldAnnos` and `getFieldAnnotations` (neither of which check for null before dereferencing the value).

The correct fix is probably to modify both `getFieldAnnos` and `getFieldAnnotations` so that they check for a `null` value rather than to throw an exception here.

kelloggm (changes requested):

kelloggm:
I added the null guards I mentioned in my comment in my previous review. I think this is ready to merge, but you should double-check my changes.

kelloggm (approved):

PR #5470, opened 2022-12-26
Prevent a WPI crash; note another one

mernst:


framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
```diff
@@ -649,11 +649,12 @@ public void processVariable(
             enclosingClass.enumConstants.add(fieldName);
 
             // Ensure that if an enum constant defines a class, that class gets registered properly.
-            // See e.g. https://docs.oracle.com/javase/specs/jls/se7/html/jls-8.html#jls-8.9.1 for
+            // See e.g. https://docs.oracle.com/javase/specs/jls/se17/html/jls-8.html#jls-8.9.1 for
             // the specification of an enum constant, which does permit it to define an anonymous
             // class.
             NewClassTree constructor = (NewClassTree) javacTree.getInitializer();
             if (constructor.getClassBody() != null) {
+              // addClass assumes there is an element for its argument, but that is not always true!
```
kelloggm, line 657: Does this still cause a crash? If so, do we know what sort of code triggers it (from the context, I would assume something like an enum constant with a method that somehow doesn't have a body, but I'm not sure how that could happen).

kelloggm (approved):

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
mernst: Yes, it does still crash.  The method does have a body, but there is no element for it.

PR #5474, opened 2022-12-27
Fix Issue 5471

### LINKED ISSUE #5471: Index Checker give a false positive
I am having a problem with ```org.checkerframework.checker.index.IndexChecker```. Here is my sample program, which checks if a line contains two consecutive question marks at the beginning.

```
public class SimpleTestCase
{
  private static boolean atTheBeginning(@IndexFor("#2") int index, String line) {
    return (index==0);
  }

  private static boolean hasDoubleQuestionMarkAtTheBeginning (String line) {
    int i = line.indexOf("??");
    if (i!=-1) {
      return (atTheBeginning(i, line));
    }
    return false;
  }

    String x = "Hello?World, this is our new program";
    if (hasDoubleQuestionMarkAtTheBeginning(x))
      System.out.println("TRUE");
  }
}
```

and here is the warning that I get

```
 return (atTheBeginning(i, line));
                             ^
  found   : @LTEqLengthOf("line") int
  required: @LTLengthOf("line") int
```

As ```i``` is declared as the index of ```??``` in ```line```, I don't see how ```i``` could be equal to the length of ```line```. Is it a bug in the Checker Framework or am I missing something?

kelloggm:
Fixes #5471. Sorry about the extremely ugly formatting in `IndexMethodIdentifier`: I'm not sure how I could re-arrange that to avoid Spotless mangling it.

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/index/IndexMethodIdentifier.java
+++ b/checker/src/main/java/org/checkerframework/checker/index/IndexMethodIdentifier.java
@@ -77,13 +77,22 @@ public class IndexMethodIdentifier {
   /**
    * Returns true iff the argument is an invocation of String#indexOf or String#lastIndexOf that
    * takes a string parameter.
+   *
+   * @param methodTree the method invocation tree to be tested
+   * @return true iff the argument is an invocation of one of String's indexOf or lastIndexOf
+   *     methods that takes another string as a parameter.
    */
   public boolean isIndexOfString(Tree methodTree) {
     ProcessingEnvironment processingEnv = factory.getProcessingEnv();
     return TreeUtils.isMethodInvocation(methodTree, indexOfStringMethods, processingEnv);
   }
 
-  /** Returns true iff the argument is an invocation of Math.min. */
+  /**
+   * Returns true iff the argument is an invocation of Math.min.
+   *
+   * @param methodTree the method invocation tree to be tested
+   * @return true iff the argument is an invocation of Math.min()
+   */
   public boolean isMathMin(Tree methodTree) {
     ProcessingEnvironment processingEnv = factory.getProcessingEnv();
     return TreeUtils.isMethodInvocation(methodTree, mathMinMethods, processingEnv);
--- a/checker/src/main/java/org/checkerframework/checker/index/IndexMethodIdentifier.java
+++ b/checker/src/main/java/org/checkerframework/checker/index/IndexMethodIdentifier.java
@@ -38,8 +38,11 @@ public class IndexMethodIdentifier {
   /** The LengthOf.value argument/element. */
   private final ExecutableElement lengthOfValueElement;
 
-  /** The {@code java.lang.String#indexOf} methods that take a string argument. */
-  private final List<ExecutableElement> stringIndexOfMethods;
+  /**
+   * The {@code java.lang.String#indexOf} and {@code #lastIndexOf} methods that take a string as a
+   * non-receiver parameter.
+   */
+  private final List<ExecutableElement> indexOfStringMethods;
 
   /** The type factory. */
   private final AnnotatedTypeFactory factory;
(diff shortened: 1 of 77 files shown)


checker/src/main/java/org/checkerframework/checker/index/IndexMethodIdentifier.java
```diff
@@ -37,6 +38,9 @@ public class IndexMethodIdentifier {
   /** The LengthOf.value argument/element. */
   private final ExecutableElement lengthOfValueElement;
 
+  /** The {@code java.lang.String#indexOf} methods that take a string argument. */
```
mernst, line 41: I would add "as a non-receiver argument" because I think all of these also take a string as the receiver.

Also, please add "and String#lastIndexOf".

checker/src/main/java/org/checkerframework/checker/index/IndexMethodIdentifier.java
```diff
@@ -52,9 +56,30 @@ public IndexMethodIdentifier(AnnotatedTypeFactory factory) {
...
+            "java.lang.String", "lastIndexOf", processingEnv, "java.lang.String", "int"));
+
     lengthOfValueElement = TreeUtils.getMethod(LengthOf.class, "value", 0, processingEnv);
   }
 
+  /**
+   * Returns true iff the argument is an invocation of String#indexOf or String#lastIndexOf that
+   * takes a string parameter.
+   */
+  public boolean isStringIndexOf(Tree methodTree) {
```
mernst, line 78: I would name this "isIndexOfString" (and likewise for the private variable) to emphasize that these are not just "String.indexOf" methods, but they are "indexOf(String)" methods.

checker/src/main/java/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -492,9 +493,66 @@ public Void visitMethodInvocation(MethodInvocationTree tree, AnnotatedTypeMirror
...
       }
+      if (imf.isStringIndexOf(tree)) {
+        // String#indexOf(String) and its variants that also take a String technically return
+        // (and are annotated as) @LTEqLengthOf the receiver. However, the result is always
+        // @LTLengthOf the receiver unless both the receiver and the target string are
+        // the empty string: "".indexOf("") returns 0, which isn't an index into "". So, this
+        // special case modifies the return type of these methods if either the parameter or
+        // the receiver is known (by the Value Checker) to not be the empty string. There are
+        // three ways the Value Checker might have that information: either string could have a
+        // @StringVal annotation whose value isn't "", either could have an @ArrayLen annotation
```
mernst, line 504: "isn't" => "doesn't include"

checker/src/main/java/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -492,9 +493,66 @@ public Void visitMethodInvocation(MethodInvocationTree tree, AnnotatedTypeMirror
...
+        // whose value isn't zero, or either could have an @ArrayLenRange annotation whose from
```
mernst, line 505: also "doesn't contain".

checker/src/main/java/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -492,9 +493,66 @@ public Void visitMethodInvocation(MethodInvocationTree tree, AnnotatedTypeMirror
...
     }
 
+    /**
+     * Returns true if the given Value Checker annotations guarantee that the annotated element is
+     * not the empty string.
+     *
+     * @param atm an annotated type from the Value Checker
+     * @param vatf the Value Annotated Type Factory
+     * @return true iff atm contains a {@code StringVal} annotation whose value isn't "", an {@code
+     *     ArrayLen} annotation whose value isn't 0, or an {@code ArrayLenRange} annotation whose
```
mernst, line 532: Here, too, discuss containment (of "" or 0) rather than equality with that value.  The code gets this right.

checker/src/main/java/org/checkerframework/checker/index/upperbound/UpperBoundAnnotatedTypeFactory.java
```diff
@@ -492,9 +493,66 @@ public Void visitMethodInvocation(MethodInvocationTree tree, AnnotatedTypeMirror
...
+     *     from value is greater than 0
+     */
+    private boolean definitelyIsNotTheEmptyString(
+        AnnotatedTypeMirror atm, ValueAnnotatedTypeFactory vatf) {
+      Set<AnnotationMirror> annos = atm.getAnnotations();
+      for (AnnotationMirror anno : annos) {
+        switch (AnnotationUtils.annotationName(anno)) {
+          case ValueAnnotatedTypeFactory.STRINGVAL_NAME:
+            List<String> strings = vatf.getStringValues(anno);
+            return strings != null && !strings.contains("");
```
mernst, line 542: Unconditional return seems wrong here.  If the boolean is true, return it.  Otherwise, continue the `for` loop because one of the other annotations might be present.

checker/tests/index/Issue5471.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class Issue5471 {
+  private static boolean atTheBeginning(@IndexFor("#2") int index, String line) {
+    return (index==0);
+  }
+
+  private static boolean hasDoubleQuestionMarkAtTheBeginning(String line) {
+    int i = line.indexOf("??");
+    if (i != -1) {
+      // Unexpected warning on this line: i is definitely an index for line.
```
mernst, line 13: This comment refers to before the fix.  I would clarify that, or (more likely) just remove the comment.

mernst (dismissed):

mernst (dismissed):

kelloggm:
@mernst this didn't merge because there was required Javadoc in `IndexMethodIdentifier` that was missing. I've added that now, so this should merge as soon as you approve (and the tests finish).

mernst (approved):
Thanks!

Issue #5472, opened 2022-12-27
JavaEE stub not working for Resource Leak Checker


msridhar:
Input:
```java
class Test {
    static void foo(javax.servlet.ServletResponse s) throws java.io.IOException {
	s.getWriter();
    }
}
```

In `JavaEE.astub` we have marked the `getWriter()` method as `@NotOwning`:

https://github.com/typetools/checker-framework/blob/master/checker/src/main/java/org/checkerframework/checker/mustcall/JavaEE.astub#L17-L17

And yet, compiling the above gives an error with 3.28.0:

```
$ ./checker-framework-3.28.0/checker/bin/javac -processor resourceleak -cp /path/to/javax.servlet-api-3.1.0.jar:. Test.java
Test.java:4: error: [required.method.not.called] @MustCall method close may not have been invoked on temp-var-1 or any of its aliases.
	s.getWriter();
	           ^
  The type of object is: java.io.PrintWriter.
  Reason for going out of scope: regular method exit
1 error
```

I used the following `javax.servlet-api-3.1.0.jar` in the classpath:

https://repo1.maven.org/maven2/javax/servlet/javax.servlet-api/3.1.0/javax.servlet-api-3.1.0.jar

kelloggm:
There is a serious problem in `AnnotatedTypeFactory#getDeclAnnotation`, which (when called from `MustCallConsistencyAnalyzer`) is not producing the `@NotOwning` annotation from the stub file, contrary to its specification (checked via print-line debugging). I don't know what the cause of that problem is, but we should investigate further.

In the mean time, I'll open a PR that works around this problem in the short term by changing the `@NotOwning` annotations to `@MustCall({})`, delegating the work to the Must Call Checker.

msridhar:
> There is a serious problem in `AnnotatedTypeFactory#getDeclAnnotation`, which (when called from `MustCallConsistencyAnalyzer`) is not producing the `@NotOwning` annotation from the stub file, contrary to its specification (checked via print-line debugging). I don't know what the cause of that problem is, but we should investigate further.
> 
> In the mean time, I'll open a PR that works around this problem in the short term by changing the `@NotOwning` annotations to `@MustCall({})`, delegating the work to the Must Call Checker.

[...]

PR #5475, opened 2022-12-27
Workaround for bug related to missing declaration annotations

kelloggm:
See #5472.

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/mustcall/JavaEE.astub
+++ b/checker/src/main/java/org/checkerframework/checker/mustcall/JavaEE.astub
@@ -13,6 +13,8 @@ package javax.servlet;
 interface ServletResponse {
     // The link below justifies why these annotations are correct
     // https://stackoverflow.com/questions/1159168/should-one-call-close-on-httpservletresponse-getoutputstream-getwriter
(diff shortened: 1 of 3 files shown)


checker/src/main/java/org/checkerframework/checker/mustcall/JavaEE.astub
```diff
@@ -13,6 +13,6 @@ package javax.servlet;
 interface ServletResponse {
     // The link below justifies why these annotations are correct
     // https://stackoverflow.com/questions/1159168/should-one-call-close-on-httpservletresponse-getoutputstream-getwriter
-    @NotOwning ServletOutputStream getOutputStream() throws IOException;
-    @NotOwning PrintWriter getWriter() throws IOException;
+    @MustCall({}) ServletOutputStream getOutputStream() throws IOException;
```
msridhar, line 16: Can we add comments here saying that we should switch back to `@NotOwning` once #5472 is fixed?  `@MustCall({})` works for the checker but it doesn't accurately capture what's going on here IMO.

msridhar (changes requested):
I think we should add a regression test as well; I'm pretty sure the bug itself is a regression, given we wrote this stub file before and it worked

msridhar (approved):

kelloggm:
Closing in favor of #5476.

PR #5476, opened 2022-12-27
Fix #5472

### LINKED ISSUE #5472: JavaEE stub not working for Resource Leak Checker

kelloggm:
The underlying problem is that `@NotOwning` and `@Owning` are owned by the Must Call Checker, not the Resource Leak Checker. So, only the Must Call Checker has access to such annotations if (and only if) they were written in stub files. I fixed both the problem described in #5472 and all other uses of `getDeclAnnotation(elt)` involving `Owning` or `NotOwning` in the consistency analyzer.

Supersedes #5475.

### DIFF
--- a/checker/build.gradle
+++ b/checker/build.gradle
@@ -62,6 +62,8 @@ dependencies {
     testImplementation 'com.amazonaws:aws-java-sdk-kms'
     // The AWS SDK is used for testing the Called Methods Checker.
     testImplementation platform('com.amazonaws:aws-java-sdk-bom:1.12.293')
+    // For the Resource Leak Checker's support for JavaEE.
+    testImplementation 'javax.servlet:javax.servlet-api:3.1.0'
 
     testImplementation group: 'junit', name: 'junit', version: '4.13.2'
     testImplementation project(':framework-test')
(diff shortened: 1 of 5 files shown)


msridhar (approved):
Nice find and fix!

PR #5478, opened 2022-12-28
RLC and Must Call Checker stub files for Apache Commons `IOUtils` class

msridhar:
I took the stub file from our old one [here](https://github.com/kelloggm/object-construction-checker/blob/9ce37327f745944e919ef3293effe69f22970e61/object-construction-checker/src/main/java/org/checkerframework/checker/objectconstruction/IOUtils.astub) and just updated an import.  However, my test case is not passing and still reporting an unexpected warning.  @kelloggm any idea how I can debug?

**Update**: This is now debugged.


kelloggm:
I looked into this briefly and I can't see anything wrong with the stub file (other than some unannotated methods being present and some non-checker-framework annotations, but those are just extraneous, not a problem). Running with `-AstubDebug` would be my first thought, but I tried that and the stub file is loaded properly with no errors. I don't see any reason that this test case wouldn't pass.

My next debugging action here would be to instrument either the consistency analyzer (to see why it's reporting the error) or the Called Methods Checker (to see if it can see the stub file). Somewhere along the chain, those `@EnsuresCalledMethods` annotations are getting lost, but I'm not sure where. Unfortunately, I don't have time to look into this today, but I might be able to tomorrow.

mernst:
This is a long shot, but could you try using explicit rather than wildcard imports?

msridhar:
> This is a long shot, but could you try using explicit rather than wildcard imports?

I gave it a shot, but it didn't seem to work.

smillst:
@msridhar The stub file was not being read by the Resource Leak Checker since it was added to the MustCall Checker.  I've fixed this.

msridhar:
> @msridhar The stub file was not being read by the Resource Leak Checker since it was added to the MustCall Checker. I've fixed this.

Thanks for the fix!

msridhar:
@smillst when is the next CF release planned?  If we could get this merged before the release is cut that would be helpful to us.

smillst:
> @smillst when is the next CF release planned? If we could get this merged before the release is cut that would be helpful to us.

I'm doing the release tomorrow, so as long as this is merged by 8am PST tomorrow, it will be in the release.

kelloggm (dismissed):
Thanks @smillst for figuring this out! I should have realized what the problem was myself.

mernst (approved):

smillst:
@mernst @msridhar @kelloggm I'll wait to do the release today until this is merged.

PR #5488, opened 2023-01-03
Add APIs to WPI for inferring 1) annotations on class declarations and 2) formal parameter annotations

kelloggm:
This PR is based on work by @Nargeshdb in [this branch](https://github.com/Nargeshdb/checker-framework/tree/inference). I copied over the code implementing the APIs themselves and have added inference logic and test cases to the `AinferTestChecker` to test the two new APIs.

The goal of these APIs is to support inference of `@MustCall` annotations on class declarations and `@Owning` annotations on method parameters, respectively. @Nargeshdb will follow this PR with implementations of those features.

Getting these APIs to pass the tests required two small bug fixes, both related to stub-based WPI:
* `SceneToStubWriter` printed class declaration annotations after the `class`, which is the wrong place: the annotations should be before the `class` keyword. Previously, we didn't have a test for `SceneToStubWriter` where a class-level annotation was inferred.
* `WholeProgramInferenceScenesStorage` could not locate the annotation file associated with an element whose type was `CLASS`.

### DIFF
--- a/framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
+++ b/framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
@@ -374,7 +374,7 @@ public class WholeProgramInferenceJavaParserStorage
   public boolean addDeclarationAnnotationToFormalParameter(
       ExecutableElement methodElt, int index, AnnotationMirror anno) {
     CallableDeclarationAnnos methodAnnos = getMethodAnnos(methodElt);
-    boolean isNewAnnotation = methodAnnos.addDeclarationAnnotationForParams(anno, index);
+    boolean isNewAnnotation = methodAnnos.addDeclarationAnnotationToFormalParameter(anno, index);
     if (isNewAnnotation) {
       modifiedFiles.add(getFileForElement(methodElt));
     }
@@ -633,6 +633,8 @@ public class WholeProgramInferenceJavaParserStorage
            *
            * @param tree tree to add. Its corresponding element is used as the key for {@code
            *     classToAnnos}.
+           * @param javaParserNode the node corresponding to the declaration, which is used to place
+           *     annotations on the class itself. Can be null, e.g. for an anonymous class.
            */
           private void addClass(ClassTree tree, @Nullable TypeDeclaration<?> javaParserNode) {
             addClass(tree, null, javaParserNode);
@@ -1027,13 +1029,13 @@ public class WholeProgramInferenceJavaParserStorage
      * Annotations on the declaration of the class (note that despite the name, these can also be
      * type annotations).
      */
-    private @MonotonicNonNull Set<AnnotationMirror> declarationAnnotations = null;
+    private @MonotonicNonNull Set<AnnotationMirror> classAnnotations = null;
 
     /**
      * The Java Parser TypeDeclaration representing the class's declaration. Used for placing
      * annotations inferred on the class declaration itself.
      */
-    private @MonotonicNonNull TypeDeclaration<?> declaration = null;
+    private @MonotonicNonNull TypeDeclaration<?> classDeclaration;
 
     /**
      * Create a new ClassOrInterfaceAnnos.
@@ -1042,9 +1044,7 @@ public class WholeProgramInferenceJavaParserStorage
      *     used for placing annotations on the class declaration
      */
     public ClassOrInterfaceAnnos(@Nullable TypeDeclaration<?> javaParserNode) {
-      if (javaParserNode != null) {
-        declaration = javaParserNode;
-      }
+      classDeclaration = javaParserNode;
     }
 
     /**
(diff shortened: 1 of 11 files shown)


framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
```diff
@@ -997,6 +1024,44 @@ private static class ClassOrInterfaceAnnos {
...
+     * Annotations on the declaration of the class (note that despite the name, these can also be
+     * type annotations).
+     */
+    private @MonotonicNonNull Set<AnnotationMirror> declarationAnnotations = null;
+
+    /**
+     * The Java Parser TypeDeclaration representing the class's declaration. Used for placing
+     * annotations inferred on the class declaration itself.
+     */
+    private @MonotonicNonNull TypeDeclaration<?> declaration = null;
```
mernst, line 1037: I suggest naming this `classDeclaration`.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
```diff
@@ -607,8 +634,8 @@ private void addClass(ClassTree tree, @Nullable @BinaryName String classNameKey)
            * @param tree tree to add. Its corresponding element is used as the key for {@code
            *     classToAnnos}.
            */
-          private void addClass(ClassTree tree) {
-            addClass(tree, null);
+          private void addClass(ClassTree tree, @Nullable TypeDeclaration<?> javaParserNode) {
```
mernst, line 637: Could you add `@param` Javadoc for this?  (I'm surprised that require-javadoc didn't require this.)

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
```diff
@@ -997,6 +1024,44 @@ private static class ClassOrInterfaceAnnos {
...
+    private @MonotonicNonNull TypeDeclaration<?> declaration = null;
+
+    /**
+     * Create a new ClassOrInterfaceAnnos
+     *
+     * @param javaParserNode the java parser node corresponding to the class declaration, which is
+     *     used for placing annotations on the class declaration
+     */
+    public ClassOrInterfaceAnnos(@Nullable TypeDeclaration<?> javaParserNode) {
+      if (javaParserNode != null) {
```
mernst, line 1046: What is the point of this test?  I would drop it, and the `= null` where `declaration` is declared.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
```diff
@@ -997,6 +1024,44 @@ private static class ClassOrInterfaceAnnos {
...
+
+    /**
+     * Adds {@code annotation} to the set of annotations on the declaration of this class.
+     *
+     * @param annotation an annotation (can be declaration or type)
+     * @return true if this is a new annotation for this class
+     */
+    public boolean addAnnotationToClassDeclaration(AnnotationMirror annotation) {
+      if (declarationAnnotations == null) {
+        declarationAnnotations = new HashSet<>();
```
mernst, line 1059: I am concerned that use of a HashSet might lead to nondeterministic WPI output.  The meaning will be the same, but it could affect diffs.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
```diff
@@ -1146,6 +1222,21 @@ public boolean addDeclarationAnnotation(AnnotationMirror annotation) {
...
     }
 
+    /**
+     * Adds a declaration annotation to this parameter and returns whether it was a new annotation.
+     *
+     * @param annotation declaration annotation to add
+     * @param index index of the parameter
+     * @return true if {@code annotation} wasn't previously stored for this parameter
+     */
+    public boolean addDeclarationAnnotationForParams(AnnotationMirror annotation, int index) {
```
mernst, line 1232: I would be inclined to drop the trailing `s` because this adds an annotation to exactly one parameter.

Or, should the names in this file be more consistent with those in `WholeProgramInferenceScenesStorage.java`?  That could improve naming and understandability.

mernst (changes requested):
Overall, this looks good.  Thanks!
I have a few small suggestions.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
kelloggm: This class already uses `HashSet` and `HashMap` to store other things whose order impacts diffs, e.g. declaration annotations on methods, the methods in a class, the fields in a class, etc. I think this is permitted, and they'll be sorted later (when the class is printed). 

(But, maybe all of those other uses are errors, too; if so, it might be better to fix them all at once separately.)

Nargeshdb:
LGTM. Thanks!

Nargeshdb (dismissed):

mernst (approved):

mernst:
@kelloggm CI is failing, for type-checking the Checker Framework.

kelloggm:
> @kelloggm CI is failing, for type-checking the Checker Framework.

The Signature Checker was warning about an error in the implementation. It should be fixed now.

mernst (approved):

PR #5490, opened 2023-01-03
Annotation declarations are declarations

kelloggm:
Fixes a crash that occurs when running WPI on plume-lib's options project, which had [this stacktrace](https://pastebin.com/raw/v4w9B2WR) recorded by @dd482IT.


mernst:
Superseded by #5491

PR #5497, opened 2023-01-05
Changes to wpi.sh and wpi-many.sh to adapt to the most recent versions of DLJC

kelloggm:
Merge with https://github.com/kelloggm/do-like-javac/pull/57.

This is in support of https://github.com/SRI-CSL/do-like-javac/pull/31.


mernst (approved):
Do not merge until Checker Framework 3.29.0 is released.

PR #5517, opened 2023-01-10
Translate 'this' to the method receiver when checking preconditions.

### LINKED ISSUE #5402: @This does not work with @RequiresCalledMethods when chaining

smillst:
Fixes #5402.

### DIFF
--- a/checker/tests/calledmethods/Issue5402.java
+++ b/checker/tests/calledmethods/Issue5402.java
@@ -1,26 +1,70 @@
+// Test case for https://github.com/typetools/checker-framework/issues/5402
+
+import org.checkerframework.checker.calledmethods.qual.CalledMethods;
 import org.checkerframework.checker.calledmethods.qual.RequiresCalledMethods;
 import org.checkerframework.common.returnsreceiver.qual.This;
-import org.checkerframework.checker.calledmethods.qual.CalledMethods;
+import org.checkerframework.dataflow.qual.Deterministic;
 
-public class Issue5402 {
+public class Issue5402 {}
 
-  public @This Issue5402 bar() {
+class Issue5402_Ok1 {
+
+  public @This Issue5402_Ok1 bar() {
     return this;
   }
-  public @This Issue5402 bar(String s) {
-    return this;
-  }
 
-  @RequiresCalledMethods( value = "this", methods = "bar" )
-  public void baz() {}
-
+  public void baz(@CalledMethods("bar") Issue5402_Ok1 this) {}
 
   public static void test() {
-    final Issue5402 foo = new Issue5402();
-    foo.bar().baz();
-    new Issue5402().bar().baz();
-     final Issue5402 foo3 = new Issue5402();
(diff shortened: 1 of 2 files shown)


mernst (approved):
Thanks for the fix.

PR #5520, opened 2023-01-10
In Ajava files, print 'l' after longs.

### LINKED ISSUE #5519: WPI fails to print "L" after long values and therefore generates invalid .ajava files
There is a failing test on [this branch](https://github.com/kelloggm/checker-framework/tree/number-format-2), which you can run with `./gradlew ainferIndexAjavaTest`. The test crashes with a `NumberFormatException`, but I expect it to pass.

Testing by hand, I discovered that if an "L" is appended after the actual long values in the ajava file, the ajava file is parsed correctly. I'm not sure why we don't produce the "L": the `LongLiteralExpr` class in Java Parser looks to have been written assuming that an "L" is appended to the (string) value that is passed into it, so maybe we're not doing that when we convert long-valued annotation values to string to construct those nodes?

@smillst would you mind taking a look at this and seeing if there is an easy fix?

smillst:
Fixes #5519.

### DIFF
--- a/checker/tests/ainfer-index/non-annotated/LongMathTest.java
+++ b/checker/tests/ainfer-index/non-annotated/LongMathTest.java
@@ -1,0 +1,5 @@
+public class LongMathTest {
+    public static long mod(long x, long y) {
+        return x % y;
+    }
+}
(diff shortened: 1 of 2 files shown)


kelloggm (approved):

mernst (approved):
Thanks for the fix.

PR #5521, opened 2023-01-11
Add intersection case to simpleTypeName method

smillst:
This should fix the crash in https://github.com/mernst/checker-framework/tree/wpi-multi-version-control-onlyone.


mernst (approved):
Thanks!

PR #5525, opened 2023-01-12
Add test infrastructure for the RLC and WPI

Nargeshdb:
Add the infrastructure to run WPI and RLC on a simple test case.


checker/src/test/java/org/checkerframework/checker/test/junit/ainferrunners/AinferResourceLeakAjavaTest.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.List;
+import org.checkerframework.checker.resourceleak.ResourceLeakChecker;
+import org.checkerframework.framework.test.AinferGeneratePerDirectoryTest;
+import org.junit.experimental.categories.Category;
+import org.junit.runners.Parameterized.Parameters;
+
+/**
+ * Tests whole-program inference with the aid of ajava files. This test is the first pass on the
+ * test data, which generates the ajava files. This specific test suite is designed to elicit
+ * problems with ajava parsing that only occur when an aggregate checker is in use.
```
kelloggm, line 13: I think this comment isn't true of this test suite (it's true of the Index Checker test suite). Although, the RLC is also an aggregate checker, so maybe it's okay. I'd say the purpose of this test suite is to allow us to test RLC-specific inference features.

kelloggm (approved):

checker/src/test/java/org/checkerframework/checker/test/junit/ainferrunners/AinferResourceLeakAjavaTest.java
Nargeshdb: I updated the comment.

kelloggm (approved):

PR #5527, opened 2023-01-16
Fix crash in WPI when receiver type is null

Nargeshdb:
The crash occurs in `updateInferredExecutableParameterTypes` when it tries to get the annotations for the receiver type of static method calls.


framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceImplementation.java
```diff
@@ -256,6 +256,10 @@ private void updateInferredExecutableParameterTypes(
       AnnotatedTypeMirror receiverArgATM = atypeFactory.getReceiverType(invocationTree);
       AnnotatedExecutableType methodDeclType = atypeFactory.getAnnotatedType(methodElt);
       AnnotatedTypeMirror receiverParamATM = methodDeclType.getReceiverType();
+      if (receiverParamATM == null) {
+        // receiver type is null for static methods and constructors
+        return;
```
kelloggm, line 261: I don't think returning here is the right choice, because that will cause inference to skip inference for the rest of the parameters, too. I think this should just skip the rest of this block, instead.

kelloggm (changes requested):

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceImplementation.java
Nargeshdb: I replaced return statement with the rest of this block.

kelloggm (approved):

PR #5539, opened 2023-01-20
Increase WPI diagnostics, for debugging CI jobs

mernst:


kelloggm (approved):

kelloggm:
@mernst the only reason I didn't merge this is because is also increases CI job times, which in the past you've put into separate PRs. If that was your intention, feel free to merge. If that's by accident, my approval applies to the changes to the WPI scripts.

PR #5542, opened 2023-01-20
Avoid WPI crash when processing records

kelloggm:


checker/src/test/java/org/checkerframework/checker/test/junit/ainferrunners/AinferTestCheckerJaifsTest.java
```diff
@@ -26,6 +26,7 @@ public AinferTestCheckerJaifsTest(List<File> testFiles) {
         "ainfer-testchecker/non-annotated",
         "-Anomsgtext",
         "-Ainfer=jaifs",
+        "-AskipDefs=SimpleRecord",
```
mernst, line 29: What is the reason for `-AskipDefs` here and below?  It seems that would avoid processing the new test case that you added.

mernst (changes requested):

checker/src/test/java/org/checkerframework/checker/test/junit/ainferrunners/AinferTestCheckerJaifsTest.java
kelloggm: The AFU's jaif reading/writing code doesn't support Java 17, and so executing jaif-based WPI on this test case causes a crash in the AFU.

kelloggm: I'll add a comment to that effect here.

mernst (approved):
Thanks!

PR #5545, opened 2023-01-20
Use count rather than boolean for parsing

mernst:


framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileElementTypes.java
```diff
@@ -225,7 +222,8 @@ public void parseAjavaFiles() {
 
       parseAnnotationFiles(ajavaFiles, AnnotationFileType.AJAVA);
     } finally {
-      parsing = false;
+      --parsingCount;
+      assert parsingCount == 0;
```
smillst, line 226: Maybe remove the assert?  Or what happens when the finally block is reached in the exceptional case and then an assertion error is thrown?

smillst (approved):

framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileElementTypes.java
mernst: I don't know what happens in that case.  I'm inclined to leave in the assertion, so that we learn what happens.  Does that sound reasonable?

PR #5548, opened 2023-01-20
Fix crash in WPI caused by mixing array and wildcard types

kelloggm:
test case provided by @Nargeshdb


Nargeshdb (approved):
LGTM. Thanks.

PR #5549, opened 2023-01-20
Adjust command that builds the Checker Framework

mernst:
@smillst is this an improvement?

### DIFF
--- a/docs/manual/troubleshooting.tex
+++ b/docs/manual/troubleshooting.tex
@@ -825,11 +825,12 @@ You might want to add an \<export CHECKERFRAMEWORK=...> line to your
 \begin{enumerate}
 
 \item
-Run \code{./gradlew assemble} to build the Checker Framework:
+Run \code{./gradlew copyJarsToDist} to build the Checker Framework (ignore
+the Javadoc errors):
 
 \begin{Verbatim}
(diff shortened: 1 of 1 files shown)


docs/manual/troubleshooting.tex
```diff
@@ -825,11 +825,12 @@
 \begin{enumerate}
 
 \item
-Run \code{./gradlew assemble} to build the Checker Framework:
+Run \code{./gradlew copyJarsToDist} to build the Checker Framework (ignore
```
smillst, line 828: This section shows how to build the Checker Framework for use the Javac wrapper.  I think that should be made clear.  Then perhaps also explain how to publish artifacts to Maven local.

This pull request is related to https://github.com/typetools/checker-framework/pull/5563. I do think this is an improvement, but I think we should add a task that has a better name.  (See comments on https://github.com/typetools/checker-framework/pull/5563).

smillst (changes requested):

docs/manual/troubleshooting.tex
smillst: Rename to `assembleForJavac`

smillst:
Closing in favor of https://github.com/typetools/checker-framework/pull/5596.

PR #5552, opened 2023-01-21
Use `getAnnotationInHierarchy()`

mernst:


smillst (changes requested):
The MustCall Checker only has a single hierarchy, so the`getAnnotation` method will always return the annotation in the MustCall hierarchy.

PR #5562, opened 2023-01-22
Capture conversion tweaks

mernst:


framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeFactory.java
```diff
@@ -2307,6 +2307,7 @@ public ParameterizedExecutableType methodFromUse(
       adaptGetClassReturnTypeToReceiver(methodType, receiverType, tree);
     }
 
+    methodType.setReturnType(applyCaptureConversion(methodType.getReturnType()));
```
smillst, line 2310: I'm not sure if this is the right place to applyCaptureConversion.  I need to try this change in my 979 branch.

PR #5563, opened 2023-01-23
Add `fastAssemble` task

mernst:


build.gradle
```diff
@@ -1184,6 +1184,20 @@ task checkBasicStyle(group: 'Format') {
 
 assemble.mustRunAfter(clean)
 
+/**
+ * Execute the `assemble` task without generating any Javadoc or sources.jar files.
+ */
+task fastAssemble(group: 'Build') {
```
smillst, line 1190: I don't think this is working as intended.  It's also not building `framework-test/build/libs/framework-test-3.29.1-SNAPSHOT.jar`.  I think disabling the Javadoc tasks is disabling any task that depends on a Javadoc, too.  So in the end only `copyJarsToDist` is running, which doesn't depend on any Javadoc tasks.

docs/manual/troubleshooting.tex
```diff
@@ -873,6 +873,11 @@
 
   \end{itemize}
 
+\item To quickly rebuild the Checker Framework for local development,
```
smillst, line 876: Depending on what is being developed, this task may or may not be artifacts required. For example, it doesn't publish to Maven local.  I think you want to speed up testing use the javac wrapper script.  In that case, `copyJarsToDist` is the task that builds only what is needed for the javac script to be updated.  The name of that task doesn't convey this, so maybe we should add an additional task that only calls `copyJarsToDist` that has a better name. Maybe `buildJavacWrapperJars`?

smillst (changes requested):

PR #5572, opened 2023-01-24
Correct how inferred annotations on type variables are stored in AJava format

### LINKED ISSUE #5524: WPI + Nullness: no annotation inferred for field with typevar type
Consider the following code, which is derived from `AbstractMostlySingletonSet` in plume-util:
```
import java.util.Set;

public abstract class TwoCtorGenericAbstract<T extends Object> implements Set<T> {
  protected T value;

  protected TwoCtorGenericAbstract() {
    this.value = null;
  }

  protected TwoCtorGenericAbstract(T v) {
    this.value = v;
  }
}
```

WPI + the nullness checker should pretty clearly infer a type annotation for `value` (either `@Nullable` or `@MonotonicNonNull`. However, nothing is inferred instead.

I think this is contributing to WPI not terminating on `plume-util` right now: the `value` field flip-flops between being annotated and not being annotated in each iteration. (I ran WPI on plume-util for ~80 iterations last night...)

@smillst @mernst do either of you have any idea what might be going on here? I investigated a bit and concluded that WPI was properly attempting to infer `@Nullable`/`@MonotonicNonNull` in `WholeProgramInferenceImplementation`, but that there didn't seem to be any evidence of that inferred annotation in storage when the field assignment in the second constructor is analyzed. I suspect that there is a bad interaction between `AnnotatedTypeVar` and `AnnotatedNullType` in here somewhere, but I haven't been able to locate it myself.

There is a test case on this branch: https://github.com/kelloggm/checker-framework/tree/nullness-wpi-loop

You can run the test case with `./gradlew ainferNullnessAjavaTest`: a warning continues to be issued at the `null` assignment in the second WPI round (validation), because no annotation is inferred for `value`.

smillst:
This fixes #5524 for AJava files, but does Jaif files have the same problem.  I've not been able to figure out how to fix them.  It seems like it would still be good to get this fix committed. However the test is failing for jaifs.  Should I just remove the test?


kelloggm:
@smillst I agree that merging the fix for ajava files is worth doing, even if jaif-based WPI fails the test. I don't know of anyone using JAIF-based WPI right now, so fixing issues in it is low priority.

I think skipping this test when testing Jaifs is fine. I'd suggest either 1) deleting the test from the `annotated` folder before running validation for jaif-based wpi tests (in `:checker`'s `build.gradle` file - there should be examples there for other tests), or 2) adding a `-skipDefs` to the arguments to the jaif-based WPI tests, as I did in #5542.

mernst (approved):

mernst (dismissed):

mernst:
@smillst @kelloggm CI is failing.

kelloggm (approved):

kelloggm:
The CI failures look like they might indicate actual problems. @smillst can you take a look?

I did merge from master in case that resolves the failures, but I don't expect it to.

smillst:
I'm going to re-open a new pull request.

PR #5581, opened 2023-01-26
Record which classes are being processed

mernst:


framework/src/main/java/org/checkerframework/framework/stub/AnnotationFileElementTypes.java
```diff
@@ -87,6 +88,13 @@ public class AnnotationFileElementTypes {
   /** Parse all JDK files at startup rather than as needed. */
   private final boolean parseAllJdkFiles;
 
+  /**
+   * Stores the fully qualified name of top-level classes (from any type of stub file) that are
+   * currently being parsed. This can stop recursively parsing an annotated JDK class that is
+   * currently being processed, which prevents conflicts of definition and infinite loops.
+   */
+  private final Set<String> processingClasses = new HashSet<>();
```
smillst, line 96: This set isn't updated.

smillst (changes requested):

PR #5582, opened 2023-01-26
Add `@SideEffectsOnly` annotation

mernst:


smillst:
@mernst The tests are failing:

```
org.checkerframework.framework.test.junit.SideEffectsOnlyTest > run[/__w/1/s/framework/sideeffectsonly] FAILED
    java.lang.AssertionError: 16 out of 16 expected diagnostics were found.
      CheckSideEffectsOnly.java:12: error: (purity.incorrect.sideeffectsonly)
      CheckSideEffectsOnly.java:17: error: (purity.incorrect.sideeffectsonly)
```

framework/src/main/java/org/checkerframework/common/basetype/messages.properties
```diff
@@ -80,6 +80,8 @@ purity.not.sideeffectfree.call=call to side-effecting %s not allowed in side-eff
 purity.more.deterministic=the method %s could be declared as @Deterministic
 purity.more.pure=the method %s could be declared as @Pure
 purity.more.sideeffectfree=the method %s could be declared as @SideEffectFree
+purity.incorrect.sideeffectsonly=the method side-effects %s
```
jyoo980, line 83: @mernst I feel like this error message might be confusing/misleading in cases like
```java
@SideEffectsOnly("#1")
void foo(List<String> strs) {
    strs.add("...");
    baz();
}

baz() { ... }
```
The error message displayed will be:
```
baz()
    ^
the method side effects strs
```
Which might be true. Could we change this to "the method might side-effect ..."?

mernst: Yes, that is a good change.  Thanks!

jyoo980:
FYI this commit `d4230fe7e043425aa366da2981ca473eef632e7f` is the latest good commit before my merge to make refinement happen for local variables.

jyoo980:
Failing test case is fixed with a JDK annotation (not merged).

PR #5590, opened 2023-01-27
Allow inference of primary annotations of type vars as types of fields

### LINKED ISSUE #5524: WPI + Nullness: no annotation inferred for field with typevar type

smillst:
@mernst @kelloggm Previously, primary annotations on fields was only allowed if the inferred annotations were a super type of the upper bound annotations.  This pull request changes this for type variables that are types of fields.  For types a fields, any primary annotation can be inferred.  Is this a good change?

Fixes #5524 for ajava, but not jaif files.


mernst (commented):
This is OK with me if it's OK with @kelloggm.

kelloggm:
> Previously, primary annotations on fields was only allowed if the inferred annotations were a super type of the upper bound annotations. This pull request changes this for type variables that are types of fields. For types a fields, any primary annotation can be inferred. Is this a good change?

I think this is a good change. In particular, the possible negative effects from this change aren't that bad: the worst thing I think that could happen because of this PR is that a field whose type is a type variable has a primary annotation inferred that's not compatible with the declared upper bound, and we get a warning because of that (because WPI doesn't infer upper bounds, yet). That warning could be fixed by inferring that the upper bound should be changed, so this change moves us closer to having effective inference for bounds.

kelloggm (approved):

PR #5595, opened 2023-01-29
Clarify that ajava files are for use by tools, not people

mernst:


kelloggm (approved):
This is a good change: humans definitely shouldn't be writing ajava files.

PR #5596, opened 2023-01-30
Improve the build from source manual section.

smillst:


mernst (approved):
Thanks!

docs/manual/troubleshooting.tex
```diff
@@ -857,6 +857,16 @@
 Putting the Checker Framework's \<javac> earlier in your \<PATH> will
 ensure that the Checker Framework's version is used.
 
+\item
```
mernst, line 860: Why is this step necessary?  It's not strictly necessary, and these instructions seem duplicative with section maven-locally-built.  It also assumes use of Maven, which the user may not be using.

mernst (changes requested):

docs/manual/troubleshooting.tex
smillst: The first step assumes use of `javac` which the user may also not be using.  I updated the test here to indicate that this step is if you are using Gradle or Maven.

mernst (approved):

PR #5602, opened 2023-02-01
Clarify comments about Javadoc

mernst:
Could you please address the TODO comments as well?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -307,11 +307,14 @@ allprojects {
         options.taglets 'org.checkerframework.taglet.ManualTaglet'
         options.tagletPath(project(':framework-test').sourceSets."${tagletVersion}".output.classesDirs.getFiles() as File[])
 
-        // We want to link to Java 9 documentation of the compiler classes since we use Java 9
-        // versions of those classes and Java 8 for everything else.  Because the compiler classes are not
+        // Link to Java 8 documentation because the compilation target is Java 8.
+        // We use Java 9 compiler classes, so use Java 9 documentation for them.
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -307,11 +307,14 @@ allprojects {
         options.taglets 'org.checkerframework.taglet.ManualTaglet'
         options.tagletPath(project(':framework-test').sourceSets."${tagletVersion}".output.classesDirs.getFiles() as File[])
 
-        // We want to link to Java 9 documentation of the compiler classes since we use Java 9
-        // versions of those classes and Java 8 for everything else.  Because the compiler classes are not
+        // Link to Java 8 documentation because the compilation target is Java 8.
+        // We use Java 9 compiler classes, so use Java 9 documentation for them.
+        // Because the compiler classes are not
         // a part of the main documentation of Java 8, javadoc links to the Java 9 versions.
+        // TODO: The following comment and code seems wrong.  Don't we want Java 8
```
smillst, line 314: When the Checker Framework is run with a JDK 8 JVM, the compiler used is a Java 9 compiler.  So, the Checker Framework uses Java 9 compiler classes.  So, this code was supposed to make Javadoc find the correct classes. However, I can't find any Java 9 links anymore.  Also, when the CF Javadoc is built with a JDK 17 JVM this doesn't work anymore and all the links are to Java 8 classes.  So, I removed this code in https://github.com/typetools/checker-framework/pull/5603.

smillst (changes requested):
This should be closed, but I'm leaving it open with a comment to explain why.

PR #5604, opened 2023-02-01
WPI: use AnnotationSet to store declaration annotations

kelloggm:
@Nargeshdb encountered duplicate declaration annotations while working on RLC inference that was caused by the use of `HashSet` rather than `AnnotationSet` here. I tested these changes on the example she provided, but that example requires a fair amount of code in `MustCallInferenceLogic` that's not yet ready to be PR'd, so this will be untested for now. @Nargeshdb will include that test in a future PR, once its dependencies are merged.


mernst:
Superseded by #5607

PR #5607, opened 2023-02-02
Use `AnnotationMirrorSet` instead of `Set`

mernst:
Perhaps a better solution than https://github.com/typetools/checker-framework/pull/5604


smillst (dismissed):
I think this is a good idea.  Could you add a changelog entry?

docs/CHANGELOG.md
```diff
@@ -1,15 +1,17 @@
 Version 3.30.1 (March 1, 2023)
---------------------------------
+------------------------------
 
 **User-visible changes:**
 
 **Implementation details:**
 
+Changed uses of `Set<AnnotationMirror>` to `AnnotationMirrorSet` including in APIs.
```
smillst, line 8: You also deleted some methods in AnnotationUtils and moved AnnotationMirrorSet and AnnotatinMirrorMap.

smillst (commented):
The title of the pull request suggests that uses of Map<AnnotationMirror, *> were also changed, but I don't see that in the code.

smillst (approved):

PR #5612, opened 2023-02-03
Don't refine the type of cast expressions

mernst:


smillst (approved):
Should there be a change log entry for this?

smillst (approved):

PR #5618, opened 2023-02-08
WPI: decision on whether to infer a type for a type variable should be made based on explicit annotations

kelloggm:
If the effective annotations are used, types from previous rounds of inference are considered, leading to flip-flopping behavior in subsequent inference rounds (e.g., in round 1 the type is inferred; in round 2 it is not inferred because it is already present; in round 3 it is no longer present and therefore inferred; etc.).


smillst (approved):

kelloggm:
This includes the changes in #5620 now, so we probably should wait to merge this until after that PR is merged.

kelloggm:
@smillst I think this PR got overlooked. I'm the original author, so I can't approve it, but the changes you made look good to me. I've set it to auto-merge, so if you approve it and the tests pass, we should be good.

smillst (approved):

PR #5619, opened 2023-02-08
Passing Resource Leak Checker test case

mernst:


kelloggm (dismissed):

checker/tests/resourceleak/RlcThisTest.java
```diff
@@ -0,0 +1,40 @@
...
+  void test() {
+    Foo f1 = new Foo();
+    f1.b1(); // RLC reports a FP at this line
+    f1.a();
+
+    Foo f2 = new Foo();
+    f2.b2(); // RLC reports a FP at this line
+    f2.a();
+
+    f1 = new Foo();
```
msridhar, line 32: For clarity I'd prefer this to assign a new variable `f3`.  Also don't we get a false positive warning for this resource?  If so we could maybe add a comment

checker/tests/resourceleak/RlcThisTest.java
```diff
@@ -0,0 +1,40 @@
...
+    Foo ff1 = f1.b1();
+    ff1.a();
+
+    f2 = new Foo();
```
msridhar, line 36: Similarly I think this should assign a new variable `f4`, and add a comment if we are getting a false positive warning for this resource

msridhar (changes requested):
Couple small changes if possible

checker/tests/resourceleak/RlcThisTest.java
mernst: The error only occurs if there is a reassignment.  I have added versions of the test with and without the reassignment to show the difference in behavior.

msridhar (approved):

PR #5624, opened 2023-02-09
Reduce number of run-time lookups for options

mernst:


mernst:
@smillst Some of these changes lead to test failures.  Under what circumstances (if any?) are these optimization permissible?

framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java
```diff
@@ -543,6 +543,17 @@ public abstract class SourceChecker extends AbstractTypeProcessor implements Opt
...
+  private boolean warns;
+  /** True if the -AshowSuppressWarningsStrings command-line argument was passed. */
+  private boolean showSuppressWarningsStrings;
+  /** True if the -ArequirePrefixInWarningSuppressions command-line argument was passed. */
+  private boolean requirePrefixInWarningSuppressions;
+  /** True if the -AwarnUnneededSuppressions command-line argument was passed. */
+  private boolean warnUnneededSuppressions;
+
   @Override
   public final synchronized void init(ProcessingEnvironment env) {
```
smillst, line 558: After a lot of digging, I figured out that `init` is not called on subcheckers. So all the logic in this method that should happen for all checkers needs to be moved to `initChecker`.

smillst (approved):

PR #5632, opened 2023-02-10
Add Javadoc

mernst:


framework/src/main/java/org/checkerframework/framework/util/defaults/QualifierDefaults.java
```diff
@@ -302,7 +307,23 @@ public void addCheckedCodeDefaults(
...
-  /** Sets the default annotations for a certain Element. */
+  /**
+   * Sets the default annotations for a certain Element.
+   *
+   * @param elem the scope to set the default within
+   * @param elementDefaultAnno the default to set
+   * @param location the location to apply the default to
+   */
+  /*
+   * TODO(cpovirk): This method looks dangerous for a type system to call early: If it "adds" a
```
smillst, line 318: I would rather not add this TODO.

smillst (changes requested):

smillst (approved):

PR #5634, opened 2023-02-10
Make `Node` field types more specific

mernst:


smillst (approved):

SKIP-REQUIRE-JAVADOC
```diff
@@ -0,0 +1 @@
+Whoever merges this pull request should remove this file when the pull request is merged.
```
smillst, line 1: I'm not auto merging because this needs to be removed after merging, so it must be manually merged.

PR #5636, opened 2023-02-10
Logic fixes

mernst:


dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java
```diff
@@ -627,7 +627,7 @@ protected void addToLookupMap(Node node) {
         Set<Node> newSet = new IdentityArraySet<>(1);
         newSet.add(node);
         treeToCfgNodes.put(enclosingParens, newSet);
-      } else if (!existing.contains(node)) {
+      } else if (!exp.contains(node)) {
```
smillst, line 630: This check isn't needed at all.

smillst (changes requested):

smillst (approved):

PR #5637, opened 2023-02-10
Reduce computation

mernst:


dataflow/src/main/java/org/checkerframework/dataflow/cfg/builder/CFGTranslationPhaseOne.java
```diff
@@ -1183,32 +1183,29 @@ protected Node commonConvert(Node node, TypeMirror varType, boolean contextAllow
...
-        nodeType = node.getType();
       }
     } else if (isRightPrimitive && isLeftPrimitive) {
       if (contextAllowsNarrowing && conversionRequiresNarrowing(varType, node)) {
         node = narrow(node, varType);
-        nodeType = node.getType();
       }
     }
+    // node might have been re-assigned; if nodeType is needed, set it again
+    // nodeType = node.getType();
```
smillst, line 1208: Why is this commented out?

smillst (changes requested):

smillst (approved):

PR #5639, opened 2023-02-10
Handle multiple Java versions differently

mernst:


javacutil/src/main/java/org/checkerframework/javacutil/ElementUtils.java
```diff
@@ -885,6 +890,23 @@ public static boolean isRecordAccessor(ExecutableElement methodElement) {
...
 
+  /**
+   * Returns true if the given {@link Element} is part of a record that has been automatically
+   * generated by the compiler. This can be a field that is derived from the record's header field
+   * list, or an automatically generated canonical constructor.
+   *
+   * @param e the {@link Element} for a member of a record
+   * @return true if the given element is generated by the compiler
+   */
+  public static boolean isAutoGeneratedRecordMember(Element e) {
```
smillst, line 901: This was moved and should be mentioned in the change log.

smillst (changes requested):

PR #5649, opened 2023-02-10
Add new method `getAnnotationOrTypeDeclarationBound`

mernst:


smillst:
No need to add an unused method.

PR #5653, opened 2023-02-10
RLC: @MustCallAlias contract only needs to be fulfilled on regular method exits

### LINKED ISSUE #5597: Only verify `@MustCallAlias` at normal exits
Consider the following (contrived and rough) example:
```java
@InheritableMustCall("close")
class Test {
  @Owning InputStream is;
  @MustCallAlias Test(@MustCallAlias InputStream p, boolean b) throws SomeException {
    if (b) { throw new SomeException(); }
    this.is = p;
  }
  @EnsuresCalledMethods("this.is", "close")
  public void close() {
    this.is.close();
  }
}
```
Right now, the Resource Leak Checker will report a `mustcallalias.out.of.scope` warning for the `Test` constructor, since when the method throws an exception, `p` is not written into the owning field `this.is`.  This warning is unexpected, as most users would expect the `@MustCallAlias` relationship to hold only when the constructor exits normally, not on both normal and exceptional exits.

We should enhance the Resource Leak Checker so that it only verifies `@MustCallAlias` relationships at normal exits, so no warning would be issued for the above example.  We'll also need to enhance the `MustCallConsistencyAnalyzer` to only assume `@MustCallAlias` relationships on normal (not exceptional) successors of relevant calls.

/cc @kelloggm @mernst @Nargeshdb

kelloggm:
fixes #5597 

@msridhar this fix seems suspiciously simple - much simpler than our discussions about this problem had led me to expect. I tried to write a test case that demonstrated an unsoundness in this implementation, but I wasn't able to find one. It would be great if you could take a few minutes to try to figure out a way to take advantage of the fact that I haven't actually changed how call-sites of methods annotated with `@MustCallAlias` annotations are treated.

### DIFF
--- a/checker/tests/resourceleak/MustCallAliasNormalExit.java
+++ b/checker/tests/resourceleak/MustCallAliasNormalExit.java
@@ -1,0 +1,73 @@
+// Test case for https://github.com/typetools/checker-framework/issues/5597
+
+import java.io.*;
+import org.checkerframework.checker.calledmethods.qual.*;
+import org.checkerframework.checker.mustcall.qual.*;
+
+@InheritableMustCall("close")
+public class MustCallAliasNormalExit {
+
+  final @Owning InputStream is;
+
+  @MustCallAlias MustCallAliasNormalExit(@MustCallAlias InputStream p, boolean b) throws Exception {
+    if (b) {
+      throw new Exception("an exception!");
+    }
+    this.is = p;
+  }
+
+  @EnsuresCalledMethods(value = "this.is", methods = "close")
+  public void close() throws IOException {
+    this.is.close();
+  }
+
+  public static @MustCallAlias MustCallAliasNormalExit mcaneFactory(InputStream is)
+      throws Exception {
+    return new MustCallAliasNormalExit(is, false);
+  }
+
+  // :: error: required.method.not.called
+  public static void testUse1(@Owning InputStream inputStream) throws IOException {
+    MustCallAliasNormalExit mcane = null;
+    try {
+      mcane = new MustCallAliasNormalExit(inputStream, true); // at run time, this WILL throw
+    } catch (Exception e) {
+      // At run time would fail (NPE), but for illustrative purposes this is fine.
+      // This absolutely must not cause inputStream to be considered closed because of the
+      // MustCallAlias
+      // relationship.
+      mcane.close();
+    }
+  }
+
+  // :: error: required.method.not.called
+  public static void testUse2(@Owning InputStream inputStream) throws IOException {
+    MustCallAliasNormalExit mcane = null;
+    try {
+      mcane = mcaneFactory(inputStream);
+    } catch (Exception e) {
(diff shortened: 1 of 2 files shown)


msridhar (approved):
This is great!  I think the change is sufficient for the following reason.  Currently, whenever a pair of `@MustCallAlias` annotations are written on a method `m`, we require one of the annotations to be on the return type.  This means that at a call site, to make use of the `@MustCallAlias` information, the result of the call to `m` must be assigned to some location.  But, in the case where `m` throws an exception, the corresponding exceptional edge at the call site is from a _predecessor_ of the CFG node performing the assignment (some kind of call node).  So, the assignment does not even appear on the control-flow path to a catch block for the exception, which means the consistency analyzer could not have added the assigned variable to the resource-alias set for the relevant resource.  Hence, closing the assigned variable in the exception block will have no effect on satisfying the relevant obligation, from the point of view of the checker.

The above discussion ignores constructors and pseudo-assignments, but I believe the same core argument holds.  So, I think this change should do the trick!  Of course @kelloggm if you see a flaw in the argument above please let us know 🙂

checker/tests/resourceleak/MustCallAliasNormalExit.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  // :: error: required.method.not.called
+  public static void testUse3(@Owning InputStream inputStream) throws Exception {
+    // if mcaneFactory throws, then inputStream goes out of scope w/o being closed
+    MustCallAliasNormalExit mcane = mcaneFactory(inputStream);
+    mcane.close();
+  }
+
+  // TODO: this appears to be a false positive, but the RLC doesn't handle it correctly because
```
msridhar, line 60: Should we open a separate issue for this one?  Looks weird.

kelloggm: Up to you. I'm not actually that surprised that we issue a false positive on this code, because the chain of reasoning is a bit confusing:
* if `mcaneFactory` does throw an exception, then `close()` is directly called on `inputStream` and no must-call-alias relationship is established
* if `mcaneFactory` doesn't throw an exception, then `mcane` and `inputStream` are resource aliases, and their obligations are fulfilled via `mcane.close()` on line 71

I think the problem for the consistency checker here is that the alias sets are different on the different paths, but I'm not sure. So, maybe we should open an issue.

kelloggm: Regardless, I'm going to merge this PR.

msridhar: > I think the problem for the consistency checker here is that the alias sets are different on the different paths, but I'm not sure. So, maybe we should open an issue.

As I understand it, the consistency checker does not do any kind of merge / least-upper bound computation at control-flow merges.  It should handle each of the paths separately.  So I am surprised by this behavior.  I can open an issue.

PR #5656, opened 2023-02-12
Make Value Checker aware of the `@Unsigned` annotation

mernst:
It is a bit unfortunate how intertwined this makes the Value Checker and the Signedness Checker.
Should the Signedness Checker be moved into the `common` package?

It would reduce the duplication of methods to always require the boolean `isUnsigned` argument.
That would also reduce the chances of accidentally calling the version without the argument.
Should I make that change?


framework/src/main/java/org/checkerframework/common/value/util/Range.java
```diff
@@ -344,6 +365,26 @@ public Range intRange() {
...
+   * range of the Short class, return SHORT_EVERYTHING.
+   *
+   * <p>If {@link #ignoreOverflow} is false and the bounds of this range are not representable as
+   * 16-bit integers, convert the bounds to Short type in accordance with Java twos-complement
+   * overflow rules, e.g., Short.MAX_VALUE + 1 is converted to Short.MIN_VALUE.
+   *
+   * @param isUnsigned if true, return a range of unsigned shorts
+   * @return this range, converted to a 16-bit short range
+   */
+  public Range shortRange(boolean isUnsigned) {
```
smillst, line 387: This `isUnsigned` parameter is not used.

framework/src/main/java/org/checkerframework/common/value/util/NumberUtils.java
```diff
@@ -32,24 +46,81 @@ public static List<? extends Number> castNumbers(
...
 
+  /**
+   * Returns the given number, casted to unsigned byte.
+   *
+   * @param n a number
+   * @return the given number, casted to unsigned byte
+   */
+  private static Short byteValueUnsigned(Number n) {
+    short result = n.byteValue();
+    if (result > Byte.MAX_VALUE) {
```
smillst, line 90: `result > Byte.MAX_VALUE` is never true.  (Same comment for the other *ValueUnsigned methods.

framework/src/main/java/org/checkerframework/common/value/ValueTreeAnnotator.java
```diff
@@ -257,6 +258,21 @@ public Void visitTypeCast(TypeCastTree tree, AnnotatedTypeMirror atm) {
...
         return null;
       }
+
+      boolean isUnsigned = false;
+      // I would like to call ((AnnotatedTypeTree) castTree).hasAnnotation(Unsigned.class), but
+      // `Unsigned` is in the checker package and this code is in the common package.
+      List<? extends AnnotationTree> annoTrees =
+          TreeUtils.getExplicitAnnotationTrees(null, tree.getType());
+      List<AnnotationMirror> annos = TreeUtils.annotationsFromTypeAnnotationTrees(annoTrees);
+      for (AnnotationMirror am : annos) {
```
smillst, line 268: This for loop can be replaced with:
be `isUnsigned = AnnotationUtils#containsSameByName(annos, "org.checkerframework.checker.signedness.qual.Unsigned")`.

smillst (changes requested):
I a bit confused how the Value Checker is using `@Unsigned`.   There's no explanation about how the Value Checker treats `@Unsigned` differently. 
> It is a bit unfortunate how intertwined this makes the Value Checker and the Signedness Checker.
> Should the Signedness Checker be moved into the common package?

The Value Checker also uses `@NonNegative` in the same way, but the Index Checker isn't in the  

> It would reduce the duplication of methods to always require the boolean isUnsigned argument.
> That would also reduce the chances of accidentally calling the version without the argument.
> Should I make that change?

Right now those parameters aren't being used, so I'm not sure how to answer this question.

framework/src/main/java/org/checkerframework/common/value/ValueVisitor.java
```diff
@@ -324,6 +333,154 @@ public Void visitTypeCast(TypeCastTree tree, Void p) {
...
+          }
+        } else {
+          switch (castTypeKind) {
+            case BYTE:
+              {
+                TreeSet<Byte> castValuesTree =
+                    new TreeSet<Byte>(CollectionsPlume.mapList(Number::byteValue, castValues));
+                TreeSet<Byte> exprValuesTree =
+                    new TreeSet<Byte>(CollectionsPlume.mapList(Number::byteValue, exprValues));
+                return sortedSetContainsAll(castValuesTree, exprValuesTree);
```
smillst, line 382: These sets are no more than 10 elements long, so it's not really necessary to use this special method.

smillst (approved):
Just one minor comment that you can take or leave.  Thanks!

framework/src/main/java/org/checkerframework/common/value/ValueVisitor.java
mernst: For sets of size 4, the method saves over 10% of run time.  That said, I do agree this isn't a bottleneck for the checker.

PR #5668, opened 2023-02-20
add ability to exclude annotations from RemoveAnnotationsForInference on a per-project basis

kelloggm:
requested by @MuyeedAhmed based on his experience running WPI on a project that relied on project-specific annotations to build. @MuyeedAhmed can you try this feature out and make sure it works for your use case?


framework/src/main/java/org/checkerframework/framework/stub/RemoveAnnotationsForInference.java
```diff
@@ -51,6 +56,12 @@
  *
  * <p>Does not remove trusted annotations: those that the checker trusts rather than verifies.
  *
+ * <p>Also does not remove "excluded" annotations. Provide excluded annotations with the {@code
+ * -exclusionFile} command line argument, which must be the first argument to this program if it is
+ * present. The exclusion file should be a list of newline-separated annotation names (without
+ * {@literal @} symbols). Both the simple and fully-qualified name of each annotation must be
+ * included in the exclusion file. TODO: remove this restriction?
```
mernst, line 63: "must be included": presumably only if both forms appear in the source code and neither should be removed.

framework/src/main/java/org/checkerframework/framework/stub/RemoveAnnotationsForInference.java
```diff
@@ -65,6 +76,13 @@
  */
 public class RemoveAnnotationsForInference {
 
+  /**
+   * A list of annotations not to remove. Used to exclude project-specific annotations that must
+   * remain for the project to build. (It would be burdensome to add all project-specific
+   * annotations to the global list in {@link #isTrustedAnnotation(String)}).
```
mernst, line 82: Minor: period goes inside parentheses.

framework/src/main/java/org/checkerframework/framework/stub/RemoveAnnotationsForInference.java
```diff
@@ -76,6 +94,26 @@ public static void main(String[] args) {
       System.err.println("Usage: provide one or more directory names to process");
       System.exit(1);
     }
+    // TODO: using plume-lib's options here would be better, but would add a dependency
+    // to the whole Checker Framework, which is undesirable. Move this program elsewhere
+    // (e.g., to a plume-lib project)?
+    if (args[0].contentEquals("-exclusionFile")) {
+      if (args.length < 3) {
+        System.err.println(
+            "Usage: -exclusionFile requires at least two more arguments: the path to "
```
mernst, line 103: Those are not arguments of `-exclusionFile`; that is, `-exclusionFile` does not take two arguments.  Instead, have this remove the first two elements of `args` (oh, I see you do this), and move the test for a directory name to process below this.

framework/src/main/java/org/checkerframework/framework/stub/RemoveAnnotationsForInference.java
```diff
@@ -350,6 +392,17 @@ static boolean isTrustedAnnotation(String name) {
         || name.equals("org.checkerframework.common.aliasing.qual.LeakedToResult");
   }
 
+  /**
+   * Returns true iff the annotation is present in the user-supplied exclusion file (via the
+   * -exclusionFile command-line option).
```
mernst, line 397: Minor: mark up `-exclusionFile` with `{@code ... }`.

mernst (changes requested):

MuyeedAhmed:
I've tested the changes in automatalib and it worked. I got the same output as the one I previously did by adding 5 trusted annotations in the "isTrustedAnnotation". @kelloggm

framework/src/main/java/org/checkerframework/framework/stub/RemoveAnnotationsForInference.java
kelloggm: Sorry if this test was unclear; I've rewritten the error-message to be more specific.

However, I don't think I can move this test: its purpose is to guard the access to `args[1]` a few lines below, which requires an `@MinLen(2)` array. (Yes, I know we're not actually running the Index Checker on this code, but it's better to follow its rules, both for safety and in case we ever do want to run it here.)

framework/src/main/java/org/checkerframework/framework/stub/RemoveAnnotationsForInference.java
```diff
@@ -65,6 +78,20 @@
...
+  private RemoveAnnotationsForInference() {
+    throw new Error("Do not instantiate RemoveAnnotationsForInference.");
+  }
+
+  /**
+   * A list of annotations not to remove. Used to exclude project-specific annotations that must
+   * remain for the project to build. (It would be burdensome to add all project-specific
+   * annotations to the global list in {@link #isTrustedAnnotation(String)}.)
+   */
+  private static @MonotonicNonNull List<String> excludedAnnotations = null;
```
mernst, line 93: `exclude` and `remove` have similar connotations.  Maybe this should be `nonRemovedAnnotations`.

framework/src/main/java/org/checkerframework/framework/stub/RemoveAnnotationsForInference.java
```diff
@@ -76,6 +103,28 @@ public static void main(String[] args) {
       System.err.println("Usage: provide one or more directory names to process");
       System.exit(1);
     }
+    // TODO: using plume-lib's options here would be better, but would add a dependency
+    // to the whole Checker Framework, which is undesirable. Move this program elsewhere
+    // (e.g., to a plume-lib project)?
+    if (args[0].contentEquals("-exclusionFile")) {
+      if (args.length < 3) {
```
mernst, line 110: I would have tested for length < 2 here, and then on line 118 set args and ended the `if "-exclusionFile"`.  Then move the `if args.length < 1` test to just after the end of the `if "-exclusionFile"` then clause.  I think that satisfies the Index Checker and does each test just once.  But I you prefer it this way, that's OK.

framework/src/main/java/org/checkerframework/framework/stub/RemoveAnnotationsForInference.java
```diff
@@ -350,6 +403,17 @@ static boolean isTrustedAnnotation(String name) {
...
   }
 
+  /**
+   * Returns true iff the annotation is present in the user-supplied exclusion file (via the {@code
+   * -exclusionFile} command-line option).
+   *
+   * @param name the annotation's name (simple or fully-qualified)
+   * @return true if the annotation was excluded by the user
+   */
+  static boolean isExcluded(String name) {
```
mernst, line 413: Similar comment to above about naming.

mernst (dismissed):
Thanks.

framework/src/main/java/org/checkerframework/framework/stub/RemoveAnnotationsForInference.java
kelloggm: Ah, I understand what you mean! Sorry for the confusion before - I misunderstood your suggestion. I agree that that is simpler, so I've implemented your suggestion.

framework/src/main/java/org/checkerframework/framework/stub/RemoveAnnotationsForInference.java
kelloggm: Good point. I'd prefer to use a positive verb, though. I've change it to `keep` throughout, but I'm open to other suggestions if you have a better one.

kelloggm:
@mernst I made enough changes that your review is marked as stale. Once you approve, I expect that this should merge itself.

dd482IT:
@kelloggm I wasn't able to get RemoveAnnotationsForInference to work for me, it still removed the annotations in my keepFile list from the project. Does the usage below look correct? 

`java -cp "$CHECKERFRAMEWORK/checker/dist/checker.jar" org.checkerframework.framework.stub.RemoveAnnotationsForInference -keepFile /Users/daniel/Documents/WPI-Research/experiments-live/dmn-check-wpi/ignore_annos.txt .`

kelloggm:
@dd482IT that usage does look correct (assuming `.` is the root directory of the target project, and that `/Users/daniel/Documents/WPI-Research/experiments-live/dmn-check-wpi/ignore_annos.txt` is the correct path to the keep file). I confirmed that the current code on this PR's branch is working on simple examples, so if it's not working for you something else must be going on.

In my testing, I've been using a relative path to the keep file (rather than an absolute path), but either should work. Other than that, I don't see a difference.

Do you mind posting the content of the keep file here? Maybe that will help.

dd482IT:
@kelloggm Professor, I figured it out, entirely my mistake, sorry for the false alarm. In my keepFile, I had the "@" prepending the annotations.

framework/src/main/java/org/checkerframework/framework/stub/RemoveAnnotationsForInference.java
```diff
@@ -350,6 +414,17 @@ static boolean isTrustedAnnotation(String name) {
...
 
+  /**
+   * Returns true iff the annotation is present in the user-supplied file of annotations to keep
+   * (via the {@code -keepFile} command-line option).
+   *
+   * @param name the annotation's name (simple or fully-qualified)
+   * @return true if the user requested that this annotation be kept in the source code
+   */
+  static boolean shouldBeKept(String name) {
+    return (annotationsToKeep != null && annotationsToKeep.contains(name));
```
mernst, line 428: Minor: if calling `contains` often on a collection, a `Set` is more appropriate than a `List`.

mernst (dismissed):

mernst (approved):

PR #5669, opened 2023-02-21
Put generated source code into build dir.

smillst:
Also combine `replaceAnnotations` task with `copySource` task.


mernst (approved):
Thanks!

PR #5674, opened 2023-02-27
Use Util methods instead of javac unsupported APIs.

smillst:
Also, deprecate a duplicat util methods.


mernst (dismissed):
Thanks!

mernst:
@smillst This change requires a changlog entry.  Feel free to merge after adding one.

PR #5683, opened 2023-02-28
build.gradle: Fix the scm connection to use https instead of git

sschuberth:
See https://github.blog/2021-09-01-improving-git-protocol-security-github/#no-more-unauthenticated-git


mernst (approved):
Thank you very much for the fix!

PR #5692, opened 2023-03-02
Don't complain about JDK version numbers

mernst:


build.gradle
```diff
@@ -61,20 +61,6 @@ ext {
   localRepo = '.git'
 }
 
-// Keep in sync with check in
-// framework/src/main/java/org/checkerframework/framework/source/SourceChecker.java .
-switch (JavaVersion.current()) {
-  case JavaVersion.VERSION_1_8:
```
smillst, line 67: Once we update the spotless plugin, then the Checker Framework Gradle scripts will not work at all under Java 8, so it would help people building the Checker Framework to get a warning message about build with Java 8

smillst (changes requested):

build.gradle
smillst: Oh right, it's not possible here, but it is in the settings file.  I've done this.   

Note, if you want to do this with another project that doesn't already have a `settings.gradle`, you can add the error in a build script block at the top of `build.gradle`. For the Checker Framework it must go in the `settings.gradle` file, because AFU's `build.gradle` is read before `checker-framework/build.gradle`.

```
buildscript {
  if (JavaVersion.current() == JavaVersion.VERSION_1_8) {
    throw new Error("The Checker Framework can't be build with Java 8.")
  }
}
```

smillst (approved):

smillst (approved):

PR #5709, opened 2023-03-07
Fix isSubtype crash in WPI

### LINKED ISSUE #5708: WPI: isSubtype crash related to type variables
I've observed this problem on two different projects. There is an example on [this branch](https://github.com/kelloggm/checker-framework/tree/isSubtypeWPICrash) (you can observe the failure by running `./gradlew ainferNullnessAjavaTest`). Sorry the test isn't completely minimized, but hopefully it should be minimized enough to get the idea.

The symptom of the problem is a crash in `QualifierHierarchy`:

```
      :-1: other: error: QualifierHierarchy: missing annotation in hierarchy @org.checkerframework.checker.nullness.qual.NonNull. found: @org.checkerframework.checker.initialization.qual.Initialized
```

with the following stacktrace: 
```
        at org.checkerframework.javacutil.BugInCF.<init>(BugInCF.java:34)
        at org.checkerframework.framework.type.QualifierHierarchy.isSubtype(QualifierHierarchy.java:140)
        at org.checkerframework.common.wholeprograminference.WholeProgramInferenceImplementation.updateAnnotationSet(WholeProgramInferenceImplementation.java:876)
...
```

Based on this stacktrace, I suspect that there is something wrong with the fix we merged a few days ago in https://github.com/typetools/checker-framework/pull/5618. @smillst if you have a chance to look at this this morning, I'd appreciate it. I'll take a look in a few hours, but I'm busy until about 2pm eastern time.

kelloggm:
fixes #5708 

I'm not 100% sure that this is the right solution to the problem. It seems to work on the test cases I have (which is good), but it feels a bit hacky. In particular, what bothers me is that this solution effectively forces a type variable's bounds to have been annotated in all relevant hierarchies in order to actually cause the variable to be skipped. That will go fine for most typecheckers (which use just a single hierarchy at a time), but I'm not sure that it is the right choice for typecheckers with multiple interacting hierarchies (like the Nullness Checker). Rather than checking the width, as I do here, it might be better to, for each annotation in `upperAnnos` (i.e., each human-written bound), check if any element of `rhsAtm.getAnnotations()` is a subtype. However, this proposed check would again change the semantics here: we then would no longer need the check that the two annotation sets are of equal size, and I'm not sure exactly how important that actually is.


smillst (approved):
Sorry I didn't have time to look at this yesterday.  The better solution was to fix the TODO.

kelloggm:
@smillst thanks for looking at it now. Your solution looks much cleaner than the hack I came up with to get the test to pass yesterday (and deals with that other TODO, as well). It looks good to me (but I can't approve this PR, since I was the original author).

PR #5728, opened 2023-03-21
Add util methods for MemberReference kinds

smillst:
Remove file SKIP-REQUIRE-JAVADOC after merge.

### DIFF
--- a/framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
+++ b/framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
@@ -3736,8 +3736,7 @@ public class BaseTypeVisitor<Factory extends GenericAnnotatedTypeFactory<?, ?, ?
       // Method type args
       requiresInference = true;
     } else if (memberReferenceTree.getMode() == ReferenceMode.NEW
-        || ((JCTree.JCMemberReference) memberReferenceTree)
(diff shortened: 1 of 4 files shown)


framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -3931,19 +3931,24 @@ private void checkPreAndPostConditions() {
           overriderType, overriddenType, superCPostFalse2, subCPostFalse2, postfalsemsg);
     }
 
+    /**
+     * Checks member reference receivers.
```
mernst, line 3935: What in particular is this checking for?

mernst (changes requested):

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
smillst: I haven't read the method to understand what it is checking.  I only edited the method to us the new util method.  I can just remove the Javadoc if you think that's better.

mernst: I think that's probably better.  Thanks.  You can create the special SKIP-REQUIRE-JAVADOC file.

mernst (dismissed):

mernst (approved):

PR #5733, opened 2023-03-27
RLC should treat calls to methods with @Owning parameters as if they call the argument's must-call methods

kelloggm:


checker/src/main/java/org/checkerframework/checker/resourceleak/ResourceLeakTransfer.java
```diff
@@ -83,6 +85,20 @@ public TransferResult<CFValue, CFStore> visitMethodInvocation(
...
+    // methods have all been called.
+    ExecutableElement methodElt = node.getTarget().getMethod();
+    // Have to use an indexed for-loop, because the loop accesses both the variable element
+    // (to check for @Owning) and the corresponding Node (to call accumulate()).
+    for (int i = 0; i < methodElt.getParameters().size(); i++) {
+      VariableElement param = methodElt.getParameters().get(i);
+      if (rlTypeFactory.hasOwning(param)) {
+        List<String> mcMethods = rlTypeFactory.getMustCallValue(param);
+        Node correspondingArgument = node.getArgument(i);
+        accumulate(correspondingArgument, result, mcMethods.toArray(new String[0]));
```
msridhar, line 98: Sorry I don't fully remember, this will add `mcMethods` to the called methods for both the normal and exceptional successor of this call, correct?

msridhar: Nevermind, answered my own question, it will work for exceptional successors also, via this line https://github.com/typetools/checker-framework/blob/e3a6e65bd65eeb92f4063f975a151d109824b063/checker/src/main/java/org/checkerframework/checker/calledmethods/CalledMethodsTransfer.java#L116-L116

msridhar (dismissed):
The changes LGTM, thanks for the quick fix!!

kelloggm:
@msridhar this needs another review from you after my fix so that it doesn't cause the `-AnoLightweightOwnership` tests to fail.

msridhar (approved):

PR #5737, opened 2023-03-30
Erase types in `super()` signature when the super type is raw

smillst:
Also, clears the `areEqualVisitHistory` when `isSubtype` crashes.


mernst (approved):
Thanks!

PR #5743, opened 2023-03-31
RLC: Resource variables are @Owning for the purposes of issuing reset.not.owning errors

### LINKED ISSUE #5739: Resource Leak Checker suggests `@Owning` on a resource variable
Consider the following code (which is checked in as `ConnectingSockets2.java`):

```java
import java.io.IOException;
import java.net.*;

class ConnectingSockets2 {

  void run(InetSocketAddress isa) {
    try (Socket serverSocket = new Socket()) {
      serverSocket.close();
      serverSocket.connect(isa);
    } catch (IOException e) {
      // do nothing
    }
  }
}
```

The Resource Leak Checker issues this error message:

```
ConnectingSockets2.java:9: error: [reset.not.owning] Calling method connect resets the must-call obligations of the expression serverSocket, which is non-owning. Either annotate its declaration with an @Owning annotation, extract it into a local variable, or write a corresponding @CreatesMustCallFor annotation on the method that encloses this statement.
      serverSocket.connect(isa);
                          ^
```

However, it is illegal to annotated the declaration of `serverSocket` as `@Owning`, and I think the code should be legal.

kelloggm:
Fixes #5739

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/mustcall/MustCallAnnotatedTypeFactory.java
+++ b/checker/src/main/java/org/checkerframework/checker/mustcall/MustCallAnnotatedTypeFactory.java
@@ -184,16 +184,6 @@ public class MustCallAnnotatedTypeFactory extends BaseAnnotatedTypeFactory
     }
   }
 
-  /**
-   * Returns true iff the given element is a resource variable.
-   *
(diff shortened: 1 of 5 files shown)


msridhar (approved):
Thanks for the fix!

mernst (approved):
Thanks!

PR #5755, opened 2023-04-06
WPI: do not infer inconsistent purity annotations in overridden and overriding methods

kelloggm:


framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceImplementation.java
```diff
@@ -743,13 +767,93 @@ public void addMethodDeclarationAnnotation(ExecutableElement methodElt, Annotati
...
+      // be unnecessary if purity implemented as a type system.
+      AnnotationMirror currentPurityAnno = getPurityAnnotation(methodElt);
+      if (currentPurityAnno == null) {
+        annoToAdd = anno;
+      } else {
+        // Clear the current purity annotation, because at this point a new one is definitely
+        // going to be inferred.
+        storage.removeMethodDeclarationAnnotation(methodElt, currentPurityAnno);
+
+        // TODO: is this the best way to do this? Would it be easier to just write a subtype
```
mernst, line 787: I suggest abstracting this lub into a separate routine, which will make this one shorter and easier to understand.

mernst (commented):
One quick comment; I will look more carefully later.

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
```diff
@@ -1050,34 +1054,78 @@ protected void checkPurity(MethodTree tree) {
...
+        // special-cased.
+        Set<? extends ExecutableElement> overriddenMethods =
+            ElementUtils.getOverriddenMethods(methodElt, types);
+        for (ExecutableElement elt : overriddenMethods) {
+          inferPurityAnno(additionalKinds, wpi, elt);
+        }
+        if (overriddenMethods.isEmpty()) {
+          // Handle interfaces. Java overrides by default refer to superclasses,
+          // so only explore the interfaces if no overridden methods were found in superclasses.
+          // Note: ElementUtils#getOverriddenMethods does not return methods in interfaces.
```
mernst, line 1075: This is contrary to the method's documentation.  I looked at the implementation, which calls JDK methods, and as of JDK 17 it does seem to include interfaces.  I commented out the containing `if` test and ran tests under JDK 11 and JDK 17, and they passed.

mernst (changes requested):

framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java
kelloggm: You are right. I think I know what happened: there was a bug in my original implementation of the LUB function over purity annotations, and I must have mis-attributed it to this method not including interfaces (because interface methods were apparently not being included, but it just happened that my test for interfaces was the only one that triggered the LUB bug).

I've removed this code.

kelloggm:
@mernst I also modified the interface for `addMethodDeclarationAnnotation` so that the same abstraction is presented to its other clients, to avoid confusion from this change.

mernst (approved):
Thanks!

mernst (approved):

Issue #4170, opened 2021-01-18
Feature: support JEP 286: Local-Variable Type Inference


vlsi:
JEP 286:

```java
This feature would allow, for example, declarations such as:

var list = new ArrayList<String>();  // infers ArrayList<String>
var stream = list.stream();          // infers Stream<String>
```

Sample use case (see https://github.com/ben-manes/caffeine/blob/fb4eb0b25edf27e7354679c0f5432987b55d1b08/caffeine/src/main/java/com/github/benmanes/caffeine/cache/LocalLoadingCache.java#L65-L73 , which is the current `v3` branch):

```java
  default Map<K, V> loadSequentially(Iterable<? extends K> keys) {
    Map<K, @Nullable V> result = new LinkedHashMap<>();
    for (K key : keys) {
      result.put(key, null);
    }

    int count = 0;
    try {
      for (var iter = result.entrySet().iterator(); iter.hasNext();) { // <== error here
        Map.Entry<K, @Nullable V> entry = iter.next();
```

The current error with `org.checkerframework:checker:3.9.1`:

```
caffeine/caffeine/src/main/java/com/github/benmanes/caffeine/cache/LocalLoadingCache.java:73: error: [assignment.type.incompatible] incompatible types in assignment.
      for (var iter = result.entrySet().iterator(); iter.hasNext();) {
                                                ^
  found   : @Initialized @NonNull Iterator<@Initialized @NonNull Entry<K[ extends @Initialized @Nullable Object super @Initialized @NonNull Void], V[ extends @Initialized @Nullable Object super @Initialized @Nullable Void]>>
  required: @UnknownInitialization @Nullable Iterator<@Initialized @NonNull Entry<K[ extends @Initialized @Nullable Object super @Initialized @NonNull Void], V[ extends @Initialized @Nullable Object super @Initialized @NonNull Void]>>
```

Could the checker framework infer `var iter` into `Iterator<Map.Entry<K, @Nullable V>`? (or something like that)

/cc @ben-manes

mernst:
This should definitely be fixed.
This should probably be fixed after #979.

PR #5770, opened 2023-04-13
Support for `var` annotations from initializer expression

kanaksad:
There are some failing tests in _NullnessTest_ due to `OutOfMemoryError`. It's being caused by `KeyForAnnotatedTypeFactory` while invoking `getAnnotatedType`, seemingly with infinite recursion. 

I was wondering if `PropagationTreeAnnotator.visitVariable` should be invoked at all when `AnnotatedTypeFactory` is of type `KeyForAnnotatedTypeFactory`. And if so, are the test cases in the right directories?
Since the changes seem to be general to all checkers, assumed that they should be placed in the `tests/all-systems` directory.

However, if that's not the case, should there be any constraints on what kind of variableTrees can `KeyForAnnotatedTypeFactory` deal with and add conditions accordingly? Or should the annotation propagation have been done in some other approach?


smillst:
To fix the infinite recursion, in org.checkerframework.checker.nullness.KeyForPropagator#propagateNewClassTree, before  `AnnotatedTypeMirror assignedTo = TypeArgInferenceUtil.assignedTo(atypeFactory, path);` return if `TreePathUtil.getAssignmentContext(path)` is a `var`.  I think that should work.

kanaksad:
> To fix the infinite recursion, in org.checkerframework.checker.nullness.KeyForPropagator#propagateNewClassTree, before `AnnotatedTypeMirror assignedTo = TypeArgInferenceUtil.assignedTo(atypeFactory, path);` return if `TreePathUtil.getAssignmentContext(path)` is a `var`. I think that should work.

thanks @smillst. it fixes the infinite recursion. pushed a commit.

kanaksad:
@smillst some checks failed due to missing javadoc in pre-existing code. could you please have a look?

framework/src/main/java/org/checkerframework/framework/type/treeannotator/PropagationTreeAnnotator.java
```diff
@@ -320,4 +363,49 @@ private void addAnnoOrBound(AnnotatedTypeMirror type, Set<? extends AnnotationMi
...
     type.addMissingAnnotations(annos);
   }
+
+  /**
+   * Propagate annotations from the type arguments of initializer type to var.
+   *
+   * <p>Simple Example:
+   *
+   * <pre>{@code
+   * typeOf(initializerType) = ArrayList<@Untainted String>
```
smillst, line 373: I don't understand this example.   `initializerType` is a type so why is it passed to `typeOf`?

framework/src/main/java/org/checkerframework/framework/type/treeannotator/PropagationTreeAnnotator.java
```diff
@@ -41,9 +49,17 @@
  */
 public class PropagationTreeAnnotator extends TreeAnnotator {
 
+  /** The qualifier hierarchy to use. */
   private final QualifierHierarchy qualHierarchy;
 
-  /** Creates a {@link PropagationTreeAnnotator} for the given {@code atypeFactory}. */
+  /** Used to replace annotations in var type according to initializer annotations. */
+  private final AnnotatedTypeReplacer replacer = new AnnotatedTypeReplacer();
```
smillst, line 56: There's no need to create a new replacer.  Call `atypeFactory.replaceAnnotations(initializerType, variableType);`
 instead.

smillst (changes requested):
Thanks for working on this issue! I have a few comments/questions.

framework/src/main/java/org/checkerframework/framework/type/treeannotator/PropagationTreeAnnotator.java
kanaksad: ah right. the `typeof` is redundant here.

framework/src/main/java/org/checkerframework/framework/type/treeannotator/PropagationTreeAnnotator.java
```diff
@@ -320,4 +363,49 @@ private void addAnnoOrBound(AnnotatedTypeMirror type, Set<? extends AnnotationMi
...
+    final TypeElement variableTypeElement =
+        (TypeElement) variableType.getUnderlyingType().asElement();
+    final Types types = typeFactory.getProcessingEnv().getTypeUtils();
+
+    if (initializerType.getTypeArguments().isEmpty()) {
+      return;
+    }
+
+    Set<Pair<Integer, Integer>> typeParamMappings =
+        TypeArgumentMapper.mapTypeArgumentIndices(
```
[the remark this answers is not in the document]
kanaksad: I don't have a case in mind where the types are not the same. I was under the impression that `aTypeFactory.replaceAnnotations` may not propagate all annotations for cases like `var s = new ArrayList<@Nullable ArrayList<@Nullable String>>()`. If that's not true, we don't need the newly written method `propagateInitializerAnnotationsToVar`at all, which looks like the case. I'm removing it and adding a relevant test case.

framework/src/main/java/org/checkerframework/framework/type/treeannotator/PropagationTreeAnnotator.java
```diff
@@ -289,6 +305,33 @@ public Void visitTypeCast(TypeCastTree tree, AnnotatedTypeMirror type) {
     return null;
   }
 
+  @Override
+  public Void visitVariable(VariableTree tree, AnnotatedTypeMirror type) {
+    if (type.getKind() == TypeKind.DECLARED) {
+      boolean isVar = TreeUtils.isVariableTreeDeclaredUsingVar((JCTree.JCVariableDecl) tree);
+      if (isVar) {
+        ExpressionTree initializer = tree.getInitializer();
+        if (initializer != null) {
```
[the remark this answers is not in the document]
smillst: It's not nesaccary to do that, but a comment here explaining when `getInitializer` might be null is.  Also, please add a test case.

checker/tests/tainting/Issue4170.java
```diff
@@ -0,0 +1,21 @@
+import java.util.ArrayList;
+import org.checkerframework.checker.tainting.qual.Tainted;
+import org.checkerframework.checker.tainting.qual.Untainted;
+
+// @below-java10-jdk-skip-test
+public class Issue4170 {
+  public void method1() {
+    var list = new ArrayList<@Untainted String>();
```
smillst, line 8: Please add tests where the variables declared with `var` are assigned to a variable with a type.  For example, add `ArrayList<@Untainted String> list2 = list` here.

smillst (changes requested):
I have a few more comments.

checker/tests/tainting/Issue4170.java
kanaksad: `ArrayList<@Untainted String> list2 = list` test is failing with an assignment error. the type of `list` stays `ArrayList<String>` in this line. is there any extra step necessary to persist the type of list from the previous line?

smillst: I added that line to this test case, and it is not failing for me.  Though when I run with JDK 11, I get a NoSuchMethodException.

kanaksad: Could you please share your jdk version where it's passing? I'm trying with `openjdk version "17.0.7" 2023-04-18` and it's failing for this version.

kanaksad: I tried again with Eclipse Temurin JDK. It is failing with both JDK 11 and JDK 17 for me. I didn't encounter the NoSuchMethodException with this version of JDK.

msridhar: > I tried again with Eclipse Temurin JDK. It is failing with both JDK 11 and JDK 17 for me. I didn't encounter the NoSuchMethodException with this version of JDK.

FWIW, I see the same result on JDK 11 and 17 on Mac (failed test, no `NoSuchMethodException`).

smillst: So it seem like `isVariableTreeDeclaredUsingVar` isn't working correctly.  I'm guessing on some JDK's it is returning `false` and on others it is crashing.  @kanaksad you can see if it is returning `false` when it should be return `true`? We may need to come up with anther way to check if a variable is declared with `var`.

kanaksad: I have one JDK where it's crashing(_11.0.11 (x86_64) "AdoptOpenJDK"_). But where it doesn't crash(_Eclipse Temurin JDK 11, 17, OpenJDK 17_), the `isVariableTreeDeclaredUsingVar` behavior is as expected, returning `true` where it should return `true`.

kanaksad: is it possible that the annotatedtype calculated here is used for type checking only at current node and doesn't update the tree? asking this because [here](https://github.com/typetools/checker-framework/blob/c1edbbb196decfcfc3bdce3e874bff06bfc56852/framework/src/main/java/org/checkerframework/common/basetype/BaseTypeVisitor.java#L1510) the variable type isn't propagated to the super call and type of tree remains the same.

msridhar: @smillst do you have any further tips here?  I think we're not sure what the next step should be

smillst: I'm still concerned that we are getting crashes on some JDKs, that should be fixed before we merge.  

But, it looks like PropagationTreeAnnotator isn't the right place for this code.  I think it needs to go in `TypeFromMemberVisitor#visitVariable`

kanaksad: @smillst I updated the code in TypeFromMemberVisitor#visitVariable and pushed a commit. This works fine with tainting annotations. However, it's still problematic for nullness annotations. 
To retrieve the annotated initializer type, I'm using `TypeFromTree.fromExpression(f, variableTree.getInitializer())`. 
But when `aTypeFactory` is `KeyForAnnotatedTypeFactory`, we're losing annotations from this API. For example, 
```
new ArrayList<@NonNull String>(); // API returns ArrayList<String>
new ArrayList<@Nullable String>(); // API returns ArrayList<String>
```
Also, the API invocation here suffers from the infinite recursion mentioned previously, so the blocking code in `KeyForPropagator#propagateNewClassTree` needs to be there. 

When `KeyForAnnotatedTypeFactory` is in action, I don't see any way to get the annotated type for `NewClassTree` that preserves the annotations. Is there any alternate way you can think of?
The last commit has a few failing tests demonstrating this problem.

smillst: You should call `f.getAnnotatedType(variableTree.getInitializer())` not `TypeFromTree.fromExpression(f, variableTree.getInitializer())`.   `TypeFromTree.fromExpression(f, variableTree.getInitializer())` does not return the type with all the correct annotations added.

kanaksad: @smillst thanks, it solves the problem. there is one new error reported in the existing `all-systems/VarKeyWord.java` test which looks okay to me. can you please confirm?

```
var s = "Hello!";
s = null; // new error reported on this assignment. 
```

smillst: I just pushed a fix for that testcase.

checker/tests/tainting/Issue4170.java
```diff
@@ -0,0 +1,49 @@
...
+    ArrayList<@Untainted String> list2 = list;
```
smillst, line 9: There are not test case that reassign a variable that was declared with var.   For example:
```
import java.util.List;

public class Var {
  void method(List<@Tainted String> taintedlist, List<@Untainted String> untaintedList ) {
    var list1 = taintedlist;
    List<@Tainted String> l = list1;
    // :: error: (assignment)
    list1 = untaintedList; 
  }
}
```
I think it might be surprising to the user that the reassignment fails, so we may want to add something to the manual to explain this.

msridhar: @smillst do you think a new sub-section of [the type refinement section](https://checkerframework.org/manual/#type-refinement) would be an appropriate place to document this?

smillst: I think so.  I'll assign this to @mernst though to get his opinion.

smillst (approved):
Thanks for implementing this feature.  I think this is ready to merge, but I would like @mernst to review the manual section before merging.

mernst (approved):
This is great.  Thank you very much for implementing this feature.

msridhar:
@mernst sorry, I had forgotten before, but should this new feature also be mentioned in the release notes?  We can open a separate PR for that if it'd be helpful

mernst:
I don't think it's essential to mention it in the release notes.  It is strictly an improvement to functionality that fixes a bug, and #4170 will be mentioned in the release notes (though I admit that is a bit obscure).  So, I think this is OK as is.  Thanks again for the pull request; I'm very happy to have this merged.

PR #5779, opened 2023-04-17
Add diagnostics, and make WPI scripts more consistent

mernst:


checker/bin/wpi-many.sh
```diff
@@ -85,8 +85,8 @@ if [ "${has_java_home}" = "yes" ] && [ ! -d "${JAVA_HOME}" ]; then
     exit 1
 fi
 
+java_version=$("${JAVA_HOME}"/bin/java -version 2>&1 | head -1 | cut -d'"' -f2 | sed '/^1\./s///' | cut -d'.' -f1)
 if [ "${has_java_home}" = "yes" ]; then
-    java_version=$("${JAVA_HOME}"/bin/java -version 2>&1 | head -1 | cut -d'"' -f2 | sed '/^1\./s///' | cut -d'.' -f1)
```
kelloggm, line 89: If this is moved outside the test of `${has_java_home}`, then won't this command always fail when the `$JAVA_HOME` environment variable isn't specified? That seems undesirable, but maybe I just don't understand the motivation for this change.

checker/bin/wpi.sh
```diff
@@ -33,9 +33,9 @@ SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 SCRIPTPATH="${SCRIPTDIR}/wpi.sh"
 
 # Report line numbers when the script fails, from
-# https://unix.stackexchange.com/a/522815
+# https://unix.stackexchange.com/a/522815 .
```
kelloggm, line 36: What is the purpose of this change?

checker/bin/wpi.sh
```diff
@@ -75,8 +75,13 @@ else
   has_java19="yes"
 fi
 
+if [ "${has_java_home}" = "yes" ] && [ ! -d "${JAVA_HOME}" ]; then
+    echo "JAVA_HOME is set to a non-existent directory ${JAVA_HOME}"
+    exit 1
+fi
+
+java_version=$("${JAVA_HOME}"/bin/java -version 2>&1 | head -1 | cut -d'"' -f2 | sed '/^1\./s///' | cut -d'.' -f1)
```
kelloggm, line 83: Same concern as above.

kelloggm (changes requested):

checker/bin/wpi-many.sh
mernst: Good point.  I reverted this change.

checker/bin/wpi.sh
mernst: That period appears in wpi-many.sh but not in this file.  I made the files more consistent, so that their diffs are smaller and it's easier to compare them and to copy changes from one to the other.

checker/bin/wpi.sh
mernst: I reverted this change.

mernst:
@kelloggm That failure was a nondeterministic (but common) problem with CI inference jobs.

PR #5791, opened 2023-04-21
Define `clone()` and `deepCopy()` methods

mernst:


framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
```diff
@@ -1038,6 +1039,29 @@ private static class CompilationUnitAnnos {
...
+      this.types = new ArrayList<>();
+    }
+
+    /**
+     * Private constructor for use by clone().
+     *
+     * @param compilationUnit compilation unit to wrap
+     * @param types wrappers for classes and interfaces in {@code compilationUnit}
+     */
+    public CompilationUnitAnnos(
```
kelloggm, line 1051: Shouldn't this be `private`? That's what the docs say.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeMirror.java
```diff
@@ -73,9 +75,10 @@ public abstract class AnnotatedTypeMirror {
   // AnnotationMirror doesn't override Object.hashCode, .equals, so we use
   // the class name of Annotation instead.
   // Caution: Assumes that a type can have at most one AnnotationMirror for any Annotation type.
-  protected final AnnotationMirrorSet annotations = new AnnotationMirrorSet();
+  // Not final so clone() can assign it; no other method should assign it.
+  protected /*final*/ AnnotationMirrorSet annotations = new AnnotationMirrorSet();
```
kelloggm, line 79: Rather than making this non-final, could `clone()` call `addAll()` on the new `AnnotationMirrorSet` that will be present in the cloned copy? Making things non-final like this seems like a code smell to me.

kelloggm (changes requested):

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeMirror.java
mernst: Unfortunately, `super.clone()` makes a shallow copy, so its `AnnotationMirrorSet` is the same object as in the original object.  Calling a mutator like `addAll()` on it will cause a side effect to both the original object and the clone.

I agree that making the variable non-final is gross, but I don't see a better way to do it.

mernst:
@kelloggm I made significant changes to the CollectionUtils methods since your last review.

kelloggm (dismissed):
These changes look good. I didn't review the new `CollectionUtils` methods as closely as I probably should have, since I assume they'll get a proper review when they're merged in plume-lib.

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeMirror.java
```diff
@@ -93,6 +96,21 @@ private AnnotatedTypeMirror(TypeMirror underlyingType, AnnotatedTypeFactory atyp
...
   }
 
+  @Override
+  public AnnotatedTypeMirror clone() {
+    /// I cannot call a constructor, because AnnotatedTypeMirror is abstract.
+    // return new AnnotatedTypeMirror(
+    //     underlyingType, atypeFactory, underlyingTypeHashCode, annotations.clone());
+
+    try {
+      AnnotatedTypeMirror result = (AnnotatedTypeMirror) super.clone();
```
smillst, line 106: AnnotatedTypeMirrors have to be carefully created so that any recursive types are correctly initialized.  So, I don't think this clone method will work.   Can you just call deepCopy instead?

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
```diff
@@ -1105,6 +1129,23 @@ public ClassOrInterfaceAnnos(@Nullable TypeDeclaration<?> javaParserNode) {
       classDeclaration = javaParserNode;
     }
 
+    // Makes a deep copy
+    @Override
+    public ClassOrInterfaceAnnos clone() {
+      try {
+        ClassOrInterfaceAnnos result = (ClassOrInterfaceAnnos) super.clone();
+        result.callableDeclarations = CollectionUtils.deepCopyValues(result.callableDeclarations);
+        result.fields = CollectionUtils.deepCopyValues(result.fields);
```
smillst, line 1138: I think using `clone` is error prone because it's easy to miss that a class hasn't properly implemented `clone`. For examile, this line (indirectly) calls FieldAnnos#clone, but FieldAnnos#clone hasn't been implemented yet.

smillst (changes requested):

mernst:
> These changes look good. I didn't review the new `CollectionUtils` methods as closely as I probably should have, since I assume they'll get a proper review when they're merged in plume-lib.

FYI, there isn't any review for plume-lib.

javacutil/src/main/java/org/checkerframework/javacutil/CollectionUtils.java
```diff
@@ -23,4 +28,119 @@ protected boolean removeEldestEntry(Map.Entry<K, V> entry) {
       }
     };
   }
+
+  ///
+  /// Temporary utility methods
+  ///
+
+  // TODO: After plume-util 1.6.6 is released, use these methods from it.
```
kelloggm, line 36: @mernst this comment seems to contradict https://github.com/typetools/checker-framework/pull/5791#issuecomment-1520675093

framework/src/main/java/org/checkerframework/framework/type/AnnotatedTypeMirror.java
mernst: Good idea.  I didn't know deepCopy existed.  (I usually define clone and factory methods near the constructor.)

javacutil/src/main/java/org/checkerframework/javacutil/Pair.java
```diff
@@ -22,6 +22,70 @@ public static <V1, V2> Pair<V1, V2> of(V1 v1, V2 v2) {
...
+   *
+   * @param <V1> the type of the first element of the pair
+   * @param <V2> the type of the second element of the pair
+   * @param orig a pair
+   * @return a deep copy of {@code orig}
+   */
+  @SuppressWarnings("nullness") // generics problem with deepCopy()
+  public static <V1 extends DeepCopyable, V2 extends DeepCopyable> Pair<V1, V2> deepCopy(
+      Pair<V1, V2> orig) {
+    // Cannot modify result of super.clone() because fields are final.
```
kelloggm, line 57: `clone()` isn't called in this method (anymore?), so I think this comment should either be deleted or changed to refer to `deepCopy()`

javacutil/src/main/java/org/checkerframework/javacutil/Pair.java
```diff
@@ -22,6 +22,70 @@ public static <V1, V2> Pair<V1, V2> of(V1 v1, V2 v2) {
...
+   * Returns a copy, where the {@code second} element is deep: {@code first} elements are identical
+   * to the argument, and {@code} second elements are deep copies.
+   *
+   * @param <V1> the type of the first element of the pair
+   * @param <V2> the type of the second element of the pair
+   * @param orig a pair
+   * @return a copy of {@code orig}, where the second element is a deep copy
+   */
+  @SuppressWarnings("nullness") // generics problem with deepCopy()
+  public static <V1, V2 extends DeepCopyable> Pair<V1, V2> deepCopySecond(Pair<V1, V2> orig) {
```
kelloggm, line 78: It seems odd to include `deepCopySecond` but not `deepCopyFirst`.

Further, I don't see any calls to this method in this PR. (So maybe this method should be removed and PR'd separately, when it is needed?)

javacutil/src/main/java/org/checkerframework/javacutil/CollectionUtils.java
```diff
@@ -23,4 +28,203 @@ protected boolean removeEldestEntry(Map.Entry<K, V> entry) {
...
+   * corresponding element of {@code orig}.
+   *
+   * @param <T> the type of elements of the collection
+   * @param <C> the type of the collection
+   * @param orig a collection
+   * @return a deep copy of {@code orig}
+   */
+  @SuppressWarnings({
+    "signedness", // problem with clone()
+    "nullness:return" // "return null;"
```
kelloggm, line 50: This seems like a bug in the Nullness Checker: `C` seems clearly nullable, since it extends `@Nullable Collection`. It would be better to open an issue and link that issue here to explain the warning.

javacutil/src/main/java/org/checkerframework/javacutil/CollectionUtils.java
```diff
@@ -23,4 +28,203 @@ protected boolean removeEldestEntry(Map.Entry<K, V> entry) {
...
+  })
+  public static <@Nullable T, C extends @Nullable Collection<T>> C cloneElements(C orig) {
```
kelloggm, line 52: I think, because of implicit bounds defaulting (https://checkerframework.org/manual/#null-defaults), `@Nullable T` and `T` are equivalent here. `@Nullable T` is clearer, I suppose, since we really intend this to be nullable?

kelloggm: Unrelated comment on the same line: should this method be `@PolyNull` (on the return type and the arguments)?

javacutil/src/main/java/org/checkerframework/javacutil/CollectionUtils.java
```diff
@@ -23,4 +28,203 @@ protected boolean removeEldestEntry(Map.Entry<K, V> entry) {
...
+  // After plume-util 1.6.6 is released, use these methods from it.
+
+  /**
+   * Returns a copy of {@code orig}, where each element of the result is a clone of the
+   * corresponding element of {@code orig}.
+   *
+   * @param <T> the type of elements of the collection
+   * @param <C> the type of the collection
+   * @param orig a collection
+   * @return a deep copy of {@code orig}
```
kelloggm, line 46: I'm not sure if it would be better to call this method `deepCopyElements`. This part of the JavaDoc suggests that it would.

In general, though, there seems to be some confusion between the terms "clone" and "deepCopy" - maybe we should define them (perhaps in the `DeepCopyable` JavaDoc?).

javacutil/src/main/java/org/checkerframework/javacutil/CollectionUtils.java
```diff
@@ -23,4 +28,203 @@ protected boolean removeEldestEntry(Map.Entry<K, V> entry) {
...
+   *
+   * @param <K> the type of keys of the map
+   * @param <V> the type of values of the map
+   * @param <M> the type of the map
+   * @param orig a map
+   * @return a deep copy of {@code orig}
+   */
+  @SuppressWarnings({"nullness", "signedness"}) // generics problem with clone
+  public static <K, V, M extends Map<K, V>> M cloneElements(M orig) {
+    return cloneElements(orig, false);
```
kelloggm, line 76: I think this should be `cloneElements(orig, true)`, according to the JavaDoc. As it is, the implementation is also identical to `cloneValues`, which seems wrong.

javacutil/src/main/java/org/checkerframework/javacutil/CollectionUtils.java
```diff
@@ -23,4 +28,203 @@ protected boolean removeEldestEntry(Map.Entry<K, V> entry) {
...
+   *
+   * @param <K> the type of keys of the map
+   * @param <V> the type of values of the map
+   * @param <M> the type of the map
+   * @param orig a map
+   * @param cloneKeys if true, clone keys; otherwise, re-use them
+   * @return a deep copy of {@code orig}
+   */
+  @SuppressWarnings({"nullness", "signedness"}) // generics problem with clone
+  private static <K, V, M extends Map<K, V>> M cloneElements(@PolyNull M orig, boolean cloneKeys) {
```
kelloggm, line 106: Should the return type also be `@PolyNull`? It doesn't make much sense to have only a single parameter be `@PolyNull`.

javacutil/src/main/java/org/checkerframework/javacutil/CollectionUtils.java
```diff
@@ -23,4 +28,203 @@ protected boolean removeEldestEntry(Map.Entry<K, V> entry) {
...
+   * @param <K> the type of keys of the map
+   * @param <V> the type of values of the map
+   * @param <M> the type of the map
+   * @param orig a map
+   * @return a deep copy of {@code orig}
+   */
+  @SuppressWarnings({"nullness", "signedness"}) // generics problem with clone
+  public static <
+          K extends @Nullable DeepCopyable, V extends @Nullable DeepCopyable, M extends Map<K, V>>
+      M deepCopy(M orig) {
```
kelloggm, line 157: Why is `deepCopy` for collections polymorphic, but not this method?

kelloggm (changes requested):

javacutil/src/main/java/org/checkerframework/javacutil/CollectionUtils.java
mernst: > Unrelated comment on the same line: should this method be `@PolyNull` (on the return type and the arguments)?

I think that is not necessary because of the Java polymorphism.

javacutil/src/main/java/org/checkerframework/javacutil/CollectionUtils.java
mernst: > Why is `deepCopy` for collections polymorphic, but not this method?

I think this method is equally polymorphic.  Could you explain what I am missing?

kelloggm (dismissed):
LGTM

javacutil/src/main/java/org/checkerframework/javacutil/CollectionUtils.java
kelloggm: Sorry, my comment was referring to qualifier polymorphism (not Java polymorphism). This has been resolved: you added `@PolyNull` annotations.

mernst:
@smillst Could you take another look when you get a chance?  Thanks.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
```diff
@@ -1038,6 +1040,34 @@ private static class CompilationUnitAnnos {
...
+    }
+
+    /**
+     * Returns a deep copy of this.
+     *
+     * @return a deep copy of this
+     */
+    @Override
+    public CompilationUnitAnnos deepCopy() {
+      /// Calling super.clone() does not work because field `types` is final.
```
smillst, line 1065: Should this be deleted?

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
```diff
@@ -1105,6 +1135,27 @@ public ClassOrInterfaceAnnos(@Nullable TypeDeclaration<?> javaParserNode) {
...
 
+    /**
+     * Returns a deep copy of this.
+     *
+     * @return a deep copy of this
+     */
+    @Override
+    public ClassOrInterfaceAnnos deepCopy() {
+      try {
+        ClassOrInterfaceAnnos result = (ClassOrInterfaceAnnos) super.clone();
```
smillst, line 1146: I don't think I would use clone at all since it's specification is so fuzzy.

smillst (approved):
I think this looks good now, but I would remove all calls to `clone` and not make `DeepCopyable` extends `Cloneable`.  That way we won't be burned by the fuzzy specification of `clone`.  But, if you want to keep it as is, that's ok

javacutil/src/main/java/org/checkerframework/javacutil/DeepCopyable.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * An interface for types that implement the {@code deepCopy()} method.
+ *
+ * @param <T>
+ */
+public interface DeepCopyable<T> extends Cloneable {
+
+  /**
+   * Returns a deep copy of this. The run-time class of the result is identical to the run-time
```
smillst, line 11: `deep copy` needs to be defined.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
mernst: The goal of calling `clone()` is to create a new object with exactly the same class as the receiver, and with fields initialized.  It is safe for that purpose.  An advantage of `clone()` is that if a declaration changes or a new superclass is added, then clone() still works and still produces an object of that type with fields defaulted (here, `classDeclaration` is relevant).  Furthermore, every clone() method definition starts with super.clone() and so doing the same thing in other copy methods feels right to me.

If you don't like its use even in that situation, or think it is too error-prone, I can change it.  In that case, I would create a private constructor that takes enough arguments for all the fields, ensuring that each one is set explicitly (rather than having `deepCopy()` do it.

mernst:
@smillst Could you take another quick look?  Thanks.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
smillst: I also I'm just imagining the nightmare of debugging a problem where the copy returned by `deepCopy` isn't deep enough and so mutations are happening on the original. Or the fact that a new object is created without any constructor being invoked.  Though I don't think either of those things are going to happen in this case.  And we are only talking about a few classes that aren't used through-out the framework.

I think in the end, what was most concerning about marking `AnnotatedTypeMirror`s `Clonable`, and since you changed that, I think this limited use of clone is ok.

smillst (approved):

PR #5794, opened 2023-04-23
Delete `checker/build/` if "clean" task fails to do so

mernst:


build.gradle
```diff
@@ -1034,6 +1034,13 @@ task releaseAndTest {
 // Don't create an empty checker-framework-VERSION.jar
 jar.onlyIf {false}
 
+clean.doLast {
```
smillst, line 1037: This should probably be in `checker/build.gradle` and use `buildDir.exists()` rather that `file("...").exists()`

smillst (changes requested):

PR #5798, opened 2023-04-23
Run CI under JDK 20

mernst:
Merge with:
 * https://github.com/plume-lib/plume-util/pull/288
 * https://github.com/typetools/annotation-tools/pull/575


build.gradle
```diff
@@ -302,37 +301,50 @@ allprojects {
...
-      // named package from the source module to the target module. This is, essentially, a
-      // command-line form of an exports clause in a module declaration[...].
-      // [...]
-      // The --add-exports option enables access to the public types of a specified package.
-      // It is sometimes necessary to go further and enable access to all non-public elements
-      // via the setAccessible method of the core reflection API. The --add-opens option can
-      // be used, at run time, to do this."
+      // "true" is temporary, because omitting it causes wpiManyTest to fail, likely due to a bug in do-like-javac.
+      if (true || JavaVersion.current().compareTo(org.gradle.api.JavaVersion.VERSION_20) < 0) {
+        sourceCompatibility = 8
```
smillst, line 306: In the past we couldn't use `options.release.set(8)`, because it [only works with JDK 10+](https://docs.gradle.org/current/userguide/building_java_projects.html#sec:compiling_with_release). But since we no longer build with JDK 8 should use it instead of  `sourceCompatibility = 8` and  `targetCompatibility = 8`.

build.gradle
```diff
@@ -302,37 +301,50 @@ allprojects {
...
-      // From https://openjdk.org/jeps/261, Section titled: "Breaking encapsulation"
-      // "The effect of each instance [of --add-exports] is to add a qualified export of the
-      // named package from the source module to the target module. This is, essentially, a
-      // command-line form of an exports clause in a module declaration[...].
-      // [...]
-      // The --add-exports option enables access to the public types of a specified package.
-      // It is sometimes necessary to go further and enable access to all non-public elements
-      // via the setAccessible method of the core reflection API. The --add-opens option can
-      // be used, at run time, to do this."
+      // "true" is temporary, because omitting it causes wpiManyTest to fail, likely due to a bug in do-like-javac.
```
smillst, line 304: Removing `true` would make the class files Java 11 and then the Checker Framework would no longer work with JDK 8.  It would be nice if the comment pointed that out.

smillst (changes requested):

smillst (approved):

PR #5811, opened 2023-04-27
Add two formal parameters to `wpiPrepareMethodForWriting()`

mernst:


framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
```diff
@@ -936,30 +936,82 @@ private String addClassesForElement(Element element) {
...
+
+  /**
+   * Return all the CallableDeclarationAnnos for the given signature.
+   *
+   * @param jvmSignature the JVM signature
+   * @param typeNames a collection of type names
+   * @return the CallableDeclarationAnnos for the given signature, in all of the types
+   */
+  private List<CallableDeclarationAnnos> findOverloads(
+      String jvmSignature, Collection<@BinaryName String> typeNames) {
```
kelloggm, line 981: `typeNames` should be `@Nullable`

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
```diff
@@ -936,30 +936,82 @@ private String addClassesForElement(Element element) {
...
    * @param methodAnnos the method or constructor annotations to modify
+   * @param inSupertypes the method or constructor annotations for all overridden methods; not
+   *     side-effected
+   * @param inSubtypes the method or constructor annotations for all overriding methods; not
+   *     side-effected
    */
-  public void wpiPrepareMethodForWriting(CallableDeclarationAnnos methodAnnos) {
-    atypeFactory.wpiPrepareMethodForWriting(methodAnnos);
+  // TODO: The type factory implementation must consider both inferred annotations *and* those
+  // written by the programmer.  Handle annotations in elements later.
```
kelloggm, line 1009: Can you expand on this comment a bit? I think the point it is making is the following, but it's too terse for me to be sure:
* the current implementation of `wpiPrepareMethodForWriting` only considers annotations that were inferred in this round of WPI (because it is based on `CallableDeclarationAnnos`s)
* in theory, a type factory might want to make changes to the inferred annotations on a method based on either the inferred annotations or the programmer written annotations (i.e., "annotations in elements")
* currently, this isn't supported, because there's no way to get from a `CallableDeclarationAnnos` to an element (are we sure about this? `CallableDeclaration`, at least, stores the `ExecutableElement` for the method, so an enterprising type system developer may already be able to accomplish this in a roundabout way)
* it would be desirable to support considering also the annotations written by the programmer (accessed via the relevant elements), but you're leaving it for later

Is that right?

kelloggm (changes requested):

mernst:
@kelloggm This is ready for re-review.

kelloggm (dismissed):

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
```diff
@@ -936,30 +936,86 @@ private String addClassesForElement(Element element) {
...
+  }
+
+  /**
+   * Return all the CallableDeclarationAnnos for the given signature.
+   *
+   * @param jvmSignature the JVM signature
+   * @param typeNames a collection of type names
+   * @return the CallableDeclarationAnnos for the given signature, in all of the types
+   */
+  private List<CallableDeclarationAnnos> findOverloads(
```
smillst, line 980: The method name doesn't match the Javadoc.

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
```diff
@@ -936,30 +936,86 @@ private String addClassesForElement(Element element) {
...
+    if (classAnnos.callableDeclarations.isEmpty()) {
+      return;
+    }
+
     for (Map.Entry<String, CallableDeclarationAnnos> methodEntry :
         classAnnos.callableDeclarations.entrySet()) {
-      wpiPrepareMethodForWriting(methodEntry.getValue());
+      String jvmSignature = methodEntry.getKey();
+      List<CallableDeclarationAnnos> inSupertypes =
+          findOverloads(jvmSignature, supertypesMap.get(classAnnos.className));
```
smillst, line 965: This is collecting overrides not overloads right?

smillst (changes requested):

framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInferenceJavaParserStorage.java
mernst: You are right.  Thanks for catching that.

smillst (approved):

kelloggm (approved):

PR #5832, opened 2023-05-02
Make release email more attractive with formatting

mernst:


docs/developer/release/release_utils.py
```diff
@@ -557,7 +557,7 @@ def get_announcement_email(version):
 
 Changes for Checker Framework version %s:
 
-<<Insert latest Checker Framework changelog entry, omitting the first line with the release version and date, and with hard line breaks removed>>
+<<Insert latest Checker Framework changelog entry from https://github.com/typetools/checker-framework/blob/master/docs/CHANGELOG.md .>>
```
smillst, line 560: Should this be a link to the raw file:
https://raw.githubusercontent.com/typetools/checker-framework/master/docs/CHANGELOG.md

docs/developer/release/release_utils.py
```diff
@@ -557,7 +557,7 @@ def get_announcement_email(version):
 
 Changes for Checker Framework version %s:
```
smillst, line 558: Should this line be removed or changed? Otherwise the email will look like:
```
Changes for Checker Framework version 3.34.0:

Version 3.34.0 (May 2, 2023)
------------------------------

**User-visible changes:**
```

smillst (changes requested):

docs/developer/release/release_utils.py
mernst: Definitely not.  Compare the formatting of the two files.  If you paste the formatted one into your mail client, it should preserve the formatting from the formatted one, which is the whole point of this change.

docs/developer/release/release_utils.py
mernst: I don't think it much matters.

docs/developer/release/release_utils.py
smillst: My mail client (Gmail) is set to plain text mode, so when I copy, the formatting is removed.  So this comment should indicate that the email should be formatted the same as it is at the link. 

When I turn off plain text mode, then the font size of the rest of the email does not match the change log.  So I have to copy the rest of the text into a Google doc to get the font sizes to match.  I've emailed you an example.  Let me know if  that's what you want.

smillst (approved):

PR #5833, opened 2023-05-02
Update the CF Gradle plugin version.

smillst:


checker/build.gradle
```diff
@@ -221,6 +221,7 @@ clean {
 }
 
 clean.doLast {
+  buildDir.deleteDir()
```
kelloggm, line 224: This change should probably be in a separate PR

kelloggm (changes requested):

checker/build.gradle
smillst: Fixed. Thanks!

kelloggm (approved):

PR #5835, opened 2023-05-02
Move `updateAtmWithLub()` into `WholeProgramInference`

mernst:


framework/src/main/java/org/checkerframework/common/wholeprograminference/WholeProgramInference.java
```diff
@@ -180,6 +180,16 @@ void updateContracts(
       ExecutableElement methodElement,
       CFAbstractStore<?, ?> store);
 
+  /**
+   * Updates sourceCodeATM to contain the LUB between sourceCodeATM and ajavaATM, ignoring missing
+   * AnnotationMirrors from ajavaATM -- it considers the LUB between an AnnotationMirror am and a
+   * missing AnnotationMirror to be am. The results are stored in sourceCodeATM.
+   *
```
kelloggm, line 187: This isn't your problem, since you didn't write this code, but would you mind adding a TODO here that says that this Javadoc should explain why this method is in `WholeProgramInference` and not in some `AnnotatedTypeMirror` related class?

kelloggm (approved):

PR #5844, opened 2023-05-09
Permit assignment to fields in a side-effect-free constructor

mernst:
Merge with https://github.com/plume-lib/plume-util/pull/301 .


dataflow/src/main/java/org/checkerframework/dataflow/util/PurityChecker.java
```diff
@@ -345,6 +362,33 @@ protected void assignmentCheck(ExpressionTree variable) {
...
+      TypeElement definesField = ElementUtils.enclosingTypeElement(fieldElt);
+      return currentType.equals(definesField);
+    }
+
+    /**
+     * Returns true if currently processing a constructor.
+     *
+     * @return true if currently processing a constructor
+     */
+    private boolean inConstructor() {
```
smillst, line 385: This method can be replaced with a call to `TreePathUtil#inConstructor`.

smillst (approved):

PR #5846, opened 2023-05-10
Enforce behavioral subtyping for @Owning and @NotOwning

### LINKED ISSUE #5722: Resource Leak Checker: soundness issue with `@Owning` parameters and method overrides
Example:
```java
public class OwningOverride {
  abstract static class A {
    public abstract void closeStream(@Owning InputStream s) throws IOException;
  }
  static class B extends A {
    @Override
    public void closeStream(InputStream s) { }
  }
  static void main(String[] args) throws IOException {
    // no resource leak reported for x
    InputStream x = new FileInputStream("foo.txt");
    A a = new B();
    a.closeStream(x);
  }
}
```
I think having an `@Owning` parameter on a superclass / interface method should force the parameter to be `@Owning` on overriding methods.

kelloggm:
Fixes #5722

### DIFF
--- a/checker/src/main/java/org/checkerframework/checker/resourceleak/ResourceLeakVisitor.java
+++ b/checker/src/main/java/org/checkerframework/checker/resourceleak/ResourceLeakVisitor.java
@@ -17,6 +17,7 @@ import org.checkerframework.checker.mustcall.CreatesMustCallForToJavaExpression;
 import org.checkerframework.checker.mustcall.MustCallAnnotatedTypeFactory;
 import org.checkerframework.checker.mustcall.MustCallChecker;
 import org.checkerframework.checker.mustcall.qual.CreatesMustCallFor;
+import org.checkerframework.checker.mustcall.qual.NotOwning;
 import org.checkerframework.checker.mustcall.qual.Owning;
 import org.checkerframework.common.basetype.BaseTypeChecker;
 import org.checkerframework.dataflow.expression.JavaExpression;
@@ -76,6 +77,7 @@ public class ResourceLeakVisitor extends CalledMethodsVisitor {
       checkCreatesMustCallForOverrides(tree, elt, mcAtf, cmcfValues);
       checkCreatesMustCallForTargetsHaveNonEmptyMustCall(tree, mcAtf);
     }
+    checkOwningOverrides(tree, elt, mcAtf);
     return super.visitMethod(tree, p);
   }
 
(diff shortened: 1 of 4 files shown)


checker/src/main/java/org/checkerframework/checker/resourceleak/ResourceLeakVisitor.java
```diff
@@ -142,6 +144,50 @@ private void checkCreatesMustCallForOverrides(
...
+   * parameter is @Owning in the overrider, and 2) if an overridden method has an @NotOwning return,
+   * then the overrider also has an @NotOwning return.
+   *
+   * @param tree overriding method, for error reporting
+   * @param overrider element for overriding method
+   * @param mcAtf the type factory
+   */
+  private void checkOwningOverrides(
+      MethodTree tree, ExecutableElement overrider, MustCallAnnotatedTypeFactory mcAtf) {
+    for (ExecutableElement overridden : ElementUtils.getOverriddenMethods(overrider, this.types)) {
```
msridhar, line 159: Minor, but I think only the "immediate" or "closest" overridden method needs to be checked, as a small optimization?  We do this in NullAway.  Not sure if there is an easy way to get that method in Checker Framework.  This is not a blocking comment if it's not easy to do

msridhar (approved):
Nice fix!  One non-blocking comment

checker/src/main/java/org/checkerframework/checker/resourceleak/ResourceLeakVisitor.java
kelloggm: I agree that this would be sound. I don't know of an easy way to do it, though (there is no API for it in `ElementUtils` or `Elements`, I think).

@smillst @mernst do either of you know of such an API? If not, I think we can merge this as-is.

msridhar: We implemented this ourselves in Nullaway, borrowing from Error Prone:

https://github.com/uber/NullAway/blob/ffdcf465dfc8276e09d5e77e8f7eb8d7f48c1b55/nullaway/src/main/java/com/uber/nullaway/NullabilityUtil.java#L108-L108

smillst: I don't think the Checker Framework has a method that returns the "closest" overridden method. Note, that  method may have two "closest" overridden methods and both need to be checked.  For example:
```
public class Overrides {

  public static class MyClass {
    public void method() {}
  }

  public interface MyInterface {
    void method();
  }

  public static class MyClassAndInterface extends MyClass implements MyInterface {
    @Override
    public void method() {
      super.method();
    }
  }
}
```

kelloggm: > I don't think the Checker Framework has a method that returns the "closest" overridden method.

In that case, I'm going to merge this PR without making a change.

> Note, that method may have two "closest" overridden methods and both need to be checked

Good point, and definitely relevant to this case. The overhead of checking all the overridden methods isn't that high, and it's definitely safer.

